PR #6, opened 2015-12-14
Fix #4: support bumpversion with .bumpversion.cfg

### LINKED ISSUE #4: Add bumpversion.cfg
There is a nice tool which is named bumpversion. Can be used to raise versions with one command. 

See example: https://github.com/openSUSE/docmanager/blob/develop/.bumpversion.cfg

tomschr:
Add configuration file for [bumpversion](https://pypi.python.org/pypi/bumpversion). Currently, only `kiwi/version.py` is bumped, I couldn't find other version strings.

When bumping, the configuration creates automatically a commit and tag. If you don't like that, change the keywords `commit` or `tag` to `False`.

If you would like to add more files, this is not a problem: amend the config file if necessary. The library provides an extensive setup.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,6 @@
+[bumpversion]
+current_version = 8.10.0
+commit = True
+tag = True
+
+[bumpversion:file:kiwi/version.py]


schaefi:
There is only one version reference, so this should be ok

Thanks

PR #10, opened 2015-12-15
Review and add corrections

tomschr:
* Consistent spelling: kiwi -> KIWI, python -> Python
* Add punctuation
* Use backticks for scripts and dirs


schaefi:
Thanks for the review

PR #11, opened 2015-12-15
Use "next generation" string

tomschr:
This is something really minor, but I think it's a (small) improvement of usability.

To distinguish between old and new KIWI, I would suggest to add a decent version string which advertise the new KIWI correctly (I've used "next generation" as it is consistent with the README). Especially if you have the old and new version on one system.


schaefi:
:1:

PR #22, opened 2016-02-21
Improve consisteny of README

tomschr:
* Consistently use KIWI
* Fixed punctuation
* Added zypper ar for adding KIWI's OBS repo


schaefi:
Very nice, thanks for reading my not so nice docs :smile:

PR #26, opened 2016-02-23
Integrated Tox

tomschr:
This is the first incarnation of Tox for KIWI. :smiley:

## Changes

* First incarnation of `tox.ini`: 
* Extend `MANIFEST.in`
* Add `doc/requirements.txt`
* Extend `conf.py` with 'sphinxcontrib.spelling' extension
* Add wordlist to `doc/source/spelling_wordlist.txt`
* Add flake8 section in `setup.cfg`

## Implementation Details
The current implementation of `tox.ini` does the following:

1. Start test cases with 3.4 then 3.5 (currently that's the only supported Python versions ATM)
2. Spell check the doc
3. Build the doc

If you like, you can insert an additional step regarding code style with flake before step 1. Currently, this is uncommented ad I wasn't sure if you like it.

*IF* you enable flake and you use an invalid code style, the `flake8` command will fail. This may or may not be your intention.

If you like, you can also start separate steps; I tried to modularize the Tox configuration. For example, if you are only interested in building the doc, you can do it with:

```
$ tox -e doc
```

The `doc` step cleans the directory, calls `linkcheck` and builds the HTML documentation. The substeps can also be used independently. If you are only interested in the linkcheck, use:

```
$ tox -e doc.linkcheck
```

There are some open questions were I wasn't exactly sure what you would like.

1. Do you want to call a code style checker? If yes, should invalid code style fail?
2. Do you need a differentiation between with coverage report and without?
3. Should spell checking be more aggressive, for example, make doc checking fail? (Currently it's not)

## Future

It's a first draft, probably there are plenty other options which we could consider:

* Use `prospector` instead of `flake8`.
* Rewrite the independant `spell` step into a substep of doc.
* Add more build targets (Python 3.6?)
* Other ideas.

I've tested it and it works for me. :) Let me know what do you think.


schaefi:
This looks great, I'll merge it and play a little bit to get a better understanding and answers to your questions

tomschr:
Great, thanks a lot! :+1:

PR #30, opened 2016-02-24
Use tox in .travis.yml, remove .travis.script

tomschr:
I've "merged" Tox and Travis. That way, you can use tox in Travis.

Changes:

* `.travis.script` is obsolete with this commit as everything is done by tox now
* Support Python 3.5
* Move most of the content from the `testenv:3.4` section into `testenv`; this helps to overwrite the Python interpreter in `testenv:3.4` and `testenv:3.5`. Makes it a snap and easier to add future interpreters.

Not sure if you like it. Feel free to ignore/close it if the former was better for your needs.

TODO:
* Probably we should adapt the file `.travis.requirements.txt` as some requirements are installed both with `pip` and with `tox`.

### DIFF
--- a/.travis.requirements.txt
+++ b/.travis.requirements.txt
@@ -2,3 +2,4 @@ flake8
 docopt
 coverage
 lxml
+tox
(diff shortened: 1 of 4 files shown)


schaefi:
very cool, let's merge it

PR #32, opened 2016-02-26
Enhance README: improve Contributiong, add new Developing section

tomschr:
* Create a procedure in section Contributing
* Use `pyvenv` as this is already installed with Python 3. No need to use `virtualenv-3.x` anymore which simplifies description (no openSUSE specific hints).
* Add new Developing section to describe how to use tox


schaefi:
ah yes it sounds so much better when you write it :smile:

PR #33, opened 2016-02-26
Fix small Tox issues

tomschr:
* `check` target didn't contain a basepython line, creating therefore a Tox error output.
*  Remove unecessary Travis requirements (will be installed by tox anyway).

=> I could reduce total time from appr. 6min to 3min


schaefi:
thanks, looks good to me

PR #35, opened 2016-02-29
Power

k0da:


kiwi/bootloader/config/grub2.py
```diff
@@ -44,7 +44,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
     def post_init(self, custom_args):
         self.custom_args = custom_args
         arch = platform.machine()
-        if arch == 'x86_64':
+        if arch == 'x86_64' or arch.startswith('ppc64'):
```
schaefi: could we make this more explicit ? like

```python
elif arch == 'ppc64' or arch == 'ppc64le'
```

or are there so many different ppc archs that we better match the way you did it ?

kiwi/bootloader/config/grub2.py
```diff
@@ -64,8 +64,8 @@ def post_init(self, custom_args):
         self.firmware = FirmWare(
             self.xml_state
         )
-
-        self.hybrid_boot = True
+        if arch == 'x86_64':
+            self.hybrid_boot = True
```
schaefi: imho we shouldn't do that. The condition will break hybrid_boot for ix86 and I think there shouldn't be such a condition at all. hybrid_boot just causes a grub.cfg which uses $linux and $initrd grub variables to check for the grub_cpu variable and decide for the right grub command. That should also work on ppc. You might want to change kiwi/bootloader/template/grub2 and there the self.header_hybrid setup

kiwi/bootloader/config/grub2.py
```diff
@@ -127,11 +127,17 @@ def setup_disk_image_config(
...
             log.info('--> Using EFI/BIOS hybrid boot disk template')
             template = self.grub2.get_disk_template(
                 self.failsafe_boot, self.hybrid_boot, self.terminal
             )
+        else:
+            log.info('--> Using regular boot disk template')
+            self.hybrid_boot = False
+            template = self.grub2.get_disk_template(
+                self.failsafe_boot, self.hybrid_boot, self.terminal
+            )
```
schaefi: Follow up with the comment above. This if statement is not needed imho

kiwi/bootloader/install/grub2.py
```diff
@@ -29,7 +29,7 @@
 
 class BootLoaderInstallGrub2(BootLoaderInstallBase):
     """
-        grub2 bootloader installation for x86 bios platform
+        grub2 bootloader installation
```
schaefi: why not state the purpose "for x86 bios and ppc64 platforms" ?

kiwi/bootloader/install/grub2.py
```diff
@@ -42,12 +42,22 @@ def post_init(self, custom_args):
...
+        self.grub2_install_dev = self.device
         self.root_mount = MountManager(
             custom_args['root_device']
         )
         self.boot_mount = MountManager(
             custom_args['boot_device']
         )
+        self.modules =' '.join(Defaults.get_grub_bios_modules())
+        self.grub2_install_args = ['--skip-fs-probe']
+        if 'prep_device' in custom_args:
```
schaefi: the prep_device is required for ppc iirc, it would be nice if this is checked in post_init early as it's done for the other devices

kiwi/builder/disk.py
```diff
@@ -476,7 +486,8 @@ def __write_bootloader_config_to_system_image(self, device_map):
         boot_uuid = self.disk.get_uuid(
             boot_device.get_device()
         )
-        self.bootloader_config.setup_disk_boot_images(boot_uuid)
+        if self.firmware.efi_mode():
+            self.bootloader_config.setup_disk_boot_images(boot_uuid)
```
schaefi: Could you remove this condition ? the bootloader_config object should know if it needs to create images or not.
The more conditions the harder to test and the harder to debug. Thanks

schaefi:
Thanks for the patches, see my review. Apart from that the tests are failing

kiwi/bootloader/config/grub2.py
k0da: we how now ppc64 and ppc64le former is big endian latter is little endian, booting wise those are compatible

kiwi/bootloader/config/grub2.py
k0da: I thought we drop 32bit support no?

kiwi/bootloader/install/grub2.py
k0da: since we moved to grub2-install, it must be multiplatform.

kiwi/bootloader/install/grub2.py
k0da: I wanted to avoid checking for firmware here. Based on if prep is available I modify target, modules and install device accordingly

kiwi/bootloader/config/grub2.py
schaefi: not yet, there are Intel atom systems at 32bit which we still want to support... oh well

kiwi/bootloader/install/grub2.py
schaefi: as long as there are 'if some-arch' in the code... but I don't have strong feelings about it

kiwi/bootloader/config/grub2.py
k0da: I'm confused about we support only SLE12 and Leap on that codebase

kiwi/bootloader/install/grub2.py
k0da: This would require to change the line all the time new architecture added

kiwi/bootloader/config/grub2.py
schaefi: openSUSE 13.2 ist still also in the queue. But even without 32bit support it would be nice to not have this if condition, it would also make your life easier with the test coverage

kiwi/bootloader/install/grub2.py
schaefi: correct and makes the purpose of the class clear and explicit like the code does too. But again you can also leave it as it is.

kiwi/bootloader/install/grub2.py
schaefi: I don't meant checking the firmware, I meant checking for if prep_device exists in custom_args, if not raise. like it's done for boot_device

see the following code from post_init

```python
if not custom_args or 'boot_device' not in custom_args:
            raise KiwiBootLoaderGrubInstallError(
                'boot device node name required for grub2 installation'
            )
```

schaefi: The problem with passing custom args to a class is that you can basically specify anything. That's why I actively check for the required parts in the custom_args before proceeding. According to the changes with grub2-install the boot_device as an example became mandatory in any case. Thus I might refactor the class to pass this information as an argument instead of a custom_arg, but that's for later

kiwi/bootloader/config/grub2.py
k0da: Yes. makes sense, but this is confusing.. 

I see no difference between hybrid  and regular boot entry

kiwi/bootloader/config/grub2.py
```diff
@@ -271,12 +270,14 @@ def setup_disk_boot_images(self, boot_uuid, lookup_path=None):
...
 
-        if self.firmware.efi_mode():
-            self.efi_boot_path = self.create_efi_path()
+        if not self.firmware.efi_mode():
+            return
 
-        self.__copy_theme_data_to_boot_directory(lookup_path)
+        log.info('Creating grub bootloader images')
+
+        self.efi_boot_path = self.create_efi_path()
```
schaefi: sorry I don't understand this change at all

* why did you move the theme copy some lines above, that has no impact and is not related to ppc support ?
* why did you introduce another condition which returns from the function with no effect ? the lines after this return statement are not evaluated if the firmware mode is not efi and it imho is not related to ppc support ?

k0da: we need a font but not modules, fonts are cross platform basically, modules are efi specific

schaefi: ok so with the return you want to skip the creation of the efi image which is architecture specific, got that. But with ppc we set firmware to 'ofw' which means the efi part is not touched. I agree we have to make the creation of the efi module architecture dependant (for e.g arm this has to change) but for the ppc support there is no need to touch it now or am I mistaken ?

I'd like to keep this changeset as clear and small and ppc focused as possible. This is because when I look at the old kiwi code things are so weird and unclear that I'd like to avoid it this time. Thanks

k0da: that path entered always. Thats why I quit everywhere besides efi

schaefi: sorry I don't get it, the current code does

1. if efi mode; create the standard path
2. copy the theme data
3. if efi mode is standard efi; build the efi image (yes that needs adaptions for other archs but not ppc now)
4. if efi mode is secure boot; get the prebuilt secure boot images

For me there is no reason to add another condition which just returns from the function and again why are we messing with the efi setup when we want to support ppc where we set the firmware to 'ofw'

sorry ??

k0da: I'll take a look

k0da: You were right. Changed it back

schaefi:
Thanks this looks good now, I will merge in the changes and will fix the conflict due to the changed namespace

schaefi:
solved conflicts with master and created a new pull request

https://github.com/SUSE/kiwi/pull/41

PR #42, opened 2016-03-08
OPAL POWER firmware

k0da:


kiwi/builder/disk.py
```diff
@@ -258,8 +258,11 @@ def create(self):
             self.__get_exclude_list_for_root_data_sync(device_map)
         )
 
-        # install boot loader
-        self.__install_bootloader(device_map)
+        # OPAL doesn't need a grub2 stage1, just a config
+        # It just kexecs grub2 menu entry
+        if not self.firmware.opal_mode():
+            # install boot loader
+            self.__install_bootloader(device_map)
```
schaefi: As discussed I would be more in favour of

```
diff --git a/kiwi/builder/disk.py b/kiwi/builder/disk.py
index 00aa0b2..d4c1342 100644
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -503,6 +503,11 @@ class DiskBuilder(object):
         )
 
     def __install_bootloader(self, device_map):
+        if arch.startswith('ppc64') and self.firmware.opal_mode():
+            # OPAL doesn't need a bootloader stage1, just a config
+            # It just kexecs the bootloader menu entry in user space
+            return
+
         root_device = device_map['root']
         boot_device = root_device
         if 'boot' in device_map:
```

kiwi/firmware.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, xml_state):
         ]
 
         firmware_types = Defaults.get_firmware_types()
-        if self.host_architecture.startswith('ppc64'):
+        if not self.firmware and self.host_architecture.startswith('ppc64'):
             self.firmware = 'ofw'
```
schaefi: I was thinking about the initial setup of the firmware on the ppc arch. Actually do you think we really should enforce it ?

k0da: would be nice if we'll have a default

schaefi:
The implementation is missing a change in the schema. From kiwi/schema/kiwi.rnc

```
        attribute firmware {
            "bios" | "ec2" | "ec2hvm" | "efi" | "uefi" | "vboot" | "ofw"
        }
```

schaefi:
The tests are failing and the not all code changes are covered. You can call

```
make test
make flake
```

and that will run a tox based test env like the travis checks, also the coding style is checked

kiwi/firmware.py
schaefi: I agree so let us create a map in Defaults which defines a default firmware for each supported arch, it could also be None. I was thinking like

```
diff --git a/kiwi/defaults.py b/kiwi/defaults.py
index e448a6d..9a868d8 100644
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -223,6 +223,27 @@ class Defaults(object):
         }
 
     @classmethod
+    def get_default_firmware(self, arch):
+        default_firmware = {
+            'x86_64': 'bios',
+            'i586': 'bios',
+            'i686': 'bios',
+            'aarch64': None,
+            'arm64': None,
+            'armv5el': None,
+            'armv5tel': None,
+            'armv6l': None,
+            'armv7l': None,
+            'ppc': 'ofw',
+            'ppc64': 'ofw',
+            'ppc64le': 'ofw',
+            's390': None,
+            's390x': None
+        }
+        if arch in default_firmware:
+            return default_firmware(arch)
+
+    @classmethod
     def get_default_boot_timeout_seconds(self):
         return 10
```

and in firmware we just call

```
if not self.firmware:
    self.firmware = Defaults.get_default_firmware(self.host_architecture)
```

Thoughts ?

k0da:
Sure, trying to figure out how to test return from __install_bootloader

schaefi:
I'll took over the task to provide a default firmware. I will open a new pull request to fix the conflict

schaefi:
new pull request created https://github.com/SUSE/kiwi/pull/45

Would be great if you can review the code, thanks

PR #45, opened 2016-03-10
Add handling for OPAL firmware setup on ppc64

schaefi:
This adds a capability of creating an image for Bare metal POWER
platform, where firmware parses grub2.cfg and simply kexecs into
an image kernel


k0da:
+1

Issue #17, opened 2016-02-19
arm support missing


schaefi:
support for the arm architecture has not been added to this version of kiwi. This is a must have since the legacy kiwi version provides support. However there are some technical aspects which needs to be clarified with the arm team at SUSE first

schaefi:
sorry for the german posting, reference information from Dirk Müller

>ARM ist ein bissl unspezifisch, wir haben mehrere arm varianten die
>interessante kiwi cornercases triggern:

>openSUSE Factory für aarch64:
>https://build.opensuse.org/package/show/openSUSE:Factory:ARM/JeOS-efi.aarch64


>dann gibts noch einen spezialfall für chromebook mit dieser extra
>vboot partition:

>https://build.opensuse.org/package/show/devel:ARM:Factory:Contrib:Chromebook/JeOS-chromebook

schaefi:
support for arm image building has been added:

* support for a custom virtual boot partition, required for some boards
* support for the EFI based layout using a firmware which loads a grub2 efi module
* no kiwi internal uboot support. If uboot is the firmware which loads the efi module the setup of uboot needs to be done via the editbootconfig/editbootinstall script hooks and the pre|postSetup|InstallBootLoader initrd hooks. The reason for this is, because each uboot setup is board specific and can't be implemented in a generic way

schaefi:
Integration test setup for building an armv7 panda board image can be found here

https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_arm/test-image-panda-oem

PR #47, opened 2016-03-18
Added support for hybrid GPT

schaefi:
Embedding an MBR into a GPT is required for a collection of
boards, e.g arm rapberry PI. The kiwi configuration provides
a new attribute called

    <type ... gpt_hybrid_mbr="true|false"

which allows to control if the GPT should be hybrid or not.
On build procedures which do not create a GPT the attribute
has no effect. This references Issue #17

### DIFF
--- a/kiwi/partitioner/dasd.py
+++ b/kiwi/partitioner/dasd.py
@@ -15,20 +15,19 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+from tempfile import NamedTemporaryFile
+
 # project
 from ..command import Command
 from ..logger import log
-from tempfile import NamedTemporaryFile
+from .base import PartitionerBase
 
 
-class PartitionerDasd(object):
+class PartitionerDasd(PartitionerBase):
     """
         implement fdasd partition setup
     """
-    def __init__(self, disk_provider):
-        self.disk_device = disk_provider.get_device()
-        self.partition_id = 0
-
+    def post_init(self):
         # fdasd partition type/flag map
         self.flag_map = {
             'f.active': None,
@@ -38,9 +37,6 @@ class PartitionerDasd(object):
             't.efi': None,
             't.csm': None
         }
-
-    def get_id(self):
-        return self.partition_id
 
     def create(self, name, mbsize, type_name, flags=None):
         self.partition_id += 1
@@ -68,7 +64,3 @@ class PartitionerDasd(object):
             # are not able to detect real errors with the fdasd operation at
             # that point.
             log.debug('potential fdasd errors were ignored')
-
-    def set_flag(self, partition_id, flag_name):
-        # on an s390 dasd device there are no such flags
-        pass
--- a/kiwi/partitioner/gpt.py
+++ b/kiwi/partitioner/gpt.py
@@ -18,20 +18,18 @@
 # project
 from ..command import Command
 from ..logger import log
+from .base import PartitionerBase
 
 from ..exceptions import (
     KiwiPartitionerGptFlagError
 )
 
 
-class PartitionerGpt(object):
+class PartitionerGpt(PartitionerBase):
     """
         implement GPT partition setup
     """
-    def __init__(self, disk_provider):
-        self.disk_device = disk_provider.get_device()
-        self.partition_id = 0
-
+    def post_init(self):
         # gdisk partition type/flag map
         self.flag_map = {
             'f.active': None,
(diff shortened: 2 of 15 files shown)


kiwi/partitioner/gpt.py
```diff
@@ -79,3 +79,8 @@ def set_flag(self, partition_id, flag_name):
             )
         else:
             log.warning('Flag %s ignored on GPT', flag_name)
+
+    def set_hybrid_mbr(self):
+        Command.run(
+            ['sgdisk', '-h', '1:2:3', self.disk_device]
```
agraf: What does "1:2:3" do?

kiwi/partitioner/msdos.py
```diff
@@ -106,3 +106,7 @@ def set_flag(self, partition_id, flag_name):
                 )
         else:
             log.warning('Flag %s ignored on msdos', flag_name)
+
+    def set_hybrid_mbr(self):
+        # msdos table is the origin partition table and can't be hybrid
```
agraf: origin?

kiwi/partitioner/dasd.py
```diff
@@ -72,3 +72,7 @@ def create(self, name, mbsize, type_name, flags=None):
     def set_flag(self, partition_id, flag_name):
         # on an s390 dasd device there are no such flags
         pass
+
+    def set_hybrid_mbr(self):
+        # on an s390 dasd device there is no concept of an hybrid MBR
```
agraf: s/an/a/

kiwi/partitioner/gpt.py
schaefi: this are the three partitions to embedd

kiwi/partitioner/dasd.py
agraf: In fact, can you set the function as unimplemented in the parent class? This really *only* makes sense for gpt. ever.

kiwi/partitioner/gpt.py
agraf: And what happens on disk resize at first boot? Does the hybrid GPT get preserved?

agraf: So how do we know we have 3 partitions? Couldn't it be more?

schaefi: sure boot code stays the same as it is right now, if this works for you at the moment it will work afterwards too. actually we just move one part of your custom script into kiwi not more

schaefi:
@agraf updated the code

the partitioner class has been refactored

test/unit/partitioner_gpt_test.py
```diff
@@ -57,3 +54,11 @@ def test_set_flag(self, mock_command):
     def test_set_flag_ignored(self, mock_warn):
         self.partitioner.set_flag(1, 'f.active')
         assert mock_warn.called
+
+    @patch('kiwi.partitioner.gpt.Command.run')
+    def test_set_hybrid_mbr(self, mock_command):
+        self.partitioner.partition_id = 3
+        self.partitioner.set_hybrid_mbr()
+        mock_command.assert_called_once_with(
+            ['sgdisk', '-h', '1:2:3', '/dev/loop0']
```
agraf: Is this test still correct?

schaefi: sure for a test case :)

schaefi: the test mocks the number of partitions to be 3 and expects the sgdisk call in the form above. If you don't see anything wrong in the sgdisk call for the number of 3 partitions we should be fine, shouldn't we ?

agraf: On 21.03.16 10:28, Marcus Schäfer wrote:
> In test/unit/partitioner_gpt_test.py
> <https://github.com/SUSE/kiwi/pull/47#discussion_r56796317>:
> 
>> @@ -57,3 +54,11 @@ def test_set_flag(self, mock_command):
>>      def test_set_flag_ignored(self, mock_warn):
>>          self.partitioner.set_flag(1, 'f.active')
>>          assert mock_warn.called
>> +
>> +    @patch('kiwi.partitioner.gpt.Command.run')
>> +    def test_set_hybrid_mbr(self, mock_command):
>> +        self.partitioner.partition_id = 3
>> +        self.partitioner.set_hybrid_mbr()
>> +        mock_command.assert_called_once_with(
>> +            ['sgdisk', '-h', '1:2:3', '/dev/loop0']
> 
> the test mocks the number of partitions to be 3 and expects the sgdisk
> call in the form above. If you don't see anything wrong in the sgdisk
> call for the number of 3 partitions we should be fine, shouldn't we ?

Yup, all good :). Thanks!


Alex

schaefi: ok, I'll merge it then. There will be a follow up request for the boot code to keep the hybrid setup after repartition. Would be great if you can have a look at that one too. I'll ping you once I'm done

PR #48, opened 2016-04-04
Add Python 3.5 to be consistent with Travis

tomschr:
* Just stated the support of 3.5 in the Trove category list to make it consistent with Travis


schaefi:
Thanks, I also have tested it on 3.5 and could not find a problem so far

Issue #49, opened 2016-04-05
Document Functions, Class, and Methods in Docstrings


tomschr:
## Problem description
*This is probably a long term issue*

Currently, the code base has got different "API documentation": Some functions and methods have docstrings, others comments, and others don't have any documentation at all.

I think, this situation can be improved. :smiley:

## Reasons

I see the following important points why I think this is important:

* It helps contributors to find its way through the code.
* It encourages good coding style.
* It makes interactive development easier (due to introspection).
* It makes it easier to create API documentation automatically (see Sphinx).

## Possible Solution

I think, a good docstring should contain the following parts:

* Short summary of the function 
* Parameters and their datatypes
* Return values and the datatype(s)
* Possible exceptions

In openSUSE/docmanager#7 I've suggested [Sphinx style](http://sphinx-doc.org/domains.html#info-field-lists):

```python
def add(a, b):
    """Adds two numbers

    :param int a: first argument
    :param int b: second argument
    :return:      sum of a and b
    :rtype:       int
    """
    return a+b
```

Here are a few specific ideas, taken from Brett Slatkins book "Effective Python":

* Documenting Modules
  Each module should have a top-level docstring. It should contain a single sentence describing the module's purpose. The paragraph that follow should contain the details that all users of the module should know about its operation.

* Documenting Classes
 Each class should have a class-level docstring. The first line describes the purpose of the class. Paragraphs that follow discuss important details. Also important public attributes and methods should be described in the class-level docstring. If possible or applicable, also provide guidance to subclass.

* Documenting Functions
  Each public function and method should have a docstring. This follows the same pattern as modules and classes. Additionally, there are also some special cases that are important to know:

  * If your function has no arguments and a simple return value, a single sentence description is probably good enough.
  * If your function doesn't return anything, it's better to leave out any mention of the return value instead of saying "returns None".
  * If you don't expect your function to raise an exception during normal operation, don't mention that fact.
  * If your function accepts a variable number of arguments or keyword-arguments, use `*args` and `**kwargs` in the documented list of arguments to describe their purpose.
  * If your function has arguments with default values, those defaults should be mentioned.
  * If your function is a generator, then your docstring should describe what the generator yields when it's iterated.

schaefi:
This make s a lot of sense and should be done

Here is an example for the Cli class

* I don't like the """text syntax, is this somewhat important, or can it be done as suggested below
[...]

PR #51, opened 2016-04-06
Add docstrings for Cli class

schaefi:
Add docstrings for Cli class
    
References Issue #49


schaefi:
@tomschr I have added the basic autodoc structure and tested the generated result

What do you think ?

schaefi:
I'm not sure if we really need the Methods references as part of the class docstring, however it's nice to see them as a reference list

tomschr:
@schaefi Just had a quick look before going back to my workshop... ;)

Actually, after our discussion yesterday I've started also to research a bit about this topic. I'm sorry if I didn't tell you earlier, but it seems most of the tedious work can be done (semi-)automatically. I'm thinking about the [`sphinx-apidoc`](http://www.sphinx-doc.org/en/stable/man/sphinx-apidoc.html)  and the `sphinx-autogen` scripts.

Both seem to do similar tasks. Either way, I guess, the way to go is to use one of these scripts and integrate it into `tox.ini`.

Bascially, the idea is to run one of the scripts over your KIWI Python sources. This creates "stub" files which can then be included into your API reference. I've did it before and it works "somehow".

Unfortunately, I haven't found a good page which explains all the differences, the pros and cons, and what else is needed. Most pages seems a bit outdated on this topic.

Your files look really great and I'm very happy you did such an awesome work! :+1:  It seems to me, most of the work can be done with one of the scripts above.

I would recommend to read http://www.sphinx-doc.org/en/stable/ext/autosummary.html this gives a good overview of the semi-automated way. Hope this helps!

schaefi:
Hi,

>    Your files look really great and I'm very happy you did such an awesome
>    work! :+1: It seems to me, most of the work can be done with one of the
>    scripts above.

:) I have used the scripts, see the commit message:

    Create Initial autodoc structure
    
    Called 'sphinx-apidoc -o source/api ../kiwi'

I have not integrated that into tox, though. My intention is to
first agree on the format of the docstrings to fit well into the
generated documentation according to your thoughts

So what do you think about the docstring format as used in cli.py

* ist that ok, description plus public attributes, plus list of methods ?
* could we live without the method list ?

Thanks

tomschr:
> :) I have used the scripts, see the commit message:

Ok, right. I missed that part, sorry. Note to myself: Don't comment on GH issues before going to a workshop. :grimacing: 

> So what do you think about the docstring format as used in cli.py 
> * is that ok, description plus public attributes, plus list of methods ?

I think this goes a bit over the top if everything is stuffed into a docstring. :) 
I probably failed to address it from my last comment, but usually you create the *description* in your docstring. The list of all methods appears in the RST files (be it manually or automatically created).

That way you don't have too much "doc code" inside your docstrings which just blows up your size.

At least that was my understanding when dealing with this topic. Maybe I'm wrong.

> * could we live without the method list ?

Good question. I think we should answer one of the following questions first:

1. Why do we need it?
2. What purpose does/should it fulfill?

For me (1) would be to have a good API documentation which can be searched and viewed. Help can be "consumed" through the interactive Python shell or via HTML. That depends on what someone prefers.

What the purpose is answers question (2): the purpose (at least IMHO) is to help contributors which works with the source code to find their way through it.

To cut a long story, from a contributors perspective, it would be probably good to have a list of methods and their arguments---at least, when someone is reading the HTML page. However, the methods can be automatically created and doesn't have to appear inside the class docstring.

Let me try it out and I can show you my idea. If you are fine with it, I can commit some of my changes into your branch. Would that be ok?

schaefi:
Thanks Tom, I have included your doc suggestions.

Regarding the docstrings, I think we don't need an extra list of methods, this is documented automatically pretty nicely. I was reading through the generated results and found all information from the classes without extra hyper references

PR #54, opened 2016-04-20
Add QuickStart Guide and Improve Documentation

tomschr:
@schaefi I've looked at the main (HTML) page and think there are some information missing and we can do better. Here is my suggestion! :smiley: 

Add QuickStart Guide and improve documentation

* Added Quick Start Guide, taken from README.md
* Improve `index.rst` file:
  * Reference quickstart in main file
  * Shortend main entry page
  * Make more headings
  * Add feature highlights to draw attention
  * Add sidebar with important KIWI links
  * Use ordered list of KIWI concept (prep and creation step)

Let me know what you think.


tomschr:
Oh, and by the way: I would probably cut the README.md into a very short version or link to the guide.
Does that make sense?

doc/source/quickstart.rst
```diff
@@ -0,0 +1,422 @@
+KIWI Quickstart
+===============
+
```
schaefi: So we call this chapter Quickstart and within the chapter we have a topic also named Quickstart, isn't that confusing ? I think there is much more information in here than just a quick start. It feels more like an introduction. Thoughts ?

doc/source/quickstart.rst
```diff
@@ -0,0 +1,422 @@
...
+|Build Status| |Health|
+
```
schaefi: Does that work in the ReST source, meaning will it show the buildstatus and code health ?

doc/source/quickstart.rst
```diff
@@ -0,0 +1,422 @@
...
+build that with the legacy KIWI, unless the appliance description
+properly encapsulates the differences into a profile.
+
+The next generation KIWI also provides the :option:`--compat` option and
+the :command:`kiwicompat` tool to be able to use the same commandline
+as provided with the legacy KIWI version.
+
+Quick Start
+-----------
+
```
schaefi: Here is the second Quick Start I was talking about in the comment above

schaefi:
I agree if we move everything to the sphinx documentation we should not repeat ourselves and just add a link in the README.md to the docs

doc/source/quickstart.rst
tomschr: Although I reviewed it, I missed that part obviously. I corrected the 2nd Quickstart into "Example Appliance Descriptions".

Well, basically, my idea was to convert the content inside the `README.md` into RST and include it into the other documentation.

doc/source/quickstart.rst
tomschr: I'll remove that.

doc/source/quickstart.rst
tomschr: Yes, replace with "Example Appliance Descriptions".

tomschr:
Ok, I've committed the changes.

Apart from that, probably the quick start can be improved further. My idea was just to present the basic idea and if you like it, I can further improve. :)

tomschr:
@schaefi After reading a bit more, maybe it's better to move the "Motivation" section of the (now) Quick Start into the `index.rst`. I think, that belongs better to an introduction/main page, than into a Quick Start.

Would that make sense?

schaefi:
Makes all sense :smile:

tomschr:
After reading the documentation, I thought it would be better to focus to its target group. As the text was taken from the README, the "Quick Start" guide is therefore renamed and has a more developer focused topic. I guess, it would make sense to add another Quick Start only dedicated just for *building* images (maybe also written as a tutorial).

As such, I've moved some sections from the quick start into the main `index.rst` file like the supported distributions, deprecated features etc. I think they make more sense in the main file.

I guess this can be improved, revised, and reviewed even further, but for the time being I think it's good enough. :smiley: 

Let me know if it's ok for you (should I squash all the commits later?)

schaefi:
ok that looks nice, thanks a ton :+1:

PR #57, opened 2016-04-22
Integrate version string from placeholder

tomschr:
*  Use `|version|` placeholder mostly in titles so no bumpversion etc. is needed.

This makes it clear(er) which version is the current version.

If you don't like the version string in titles, I can move them into a separate block.


schaefi:
yep nice additional information, thanks

PR #58, opened 2016-04-22
Used :returncode: for sphinxcontrib-programoutput

tomschr:
Fixes a warning when trying to display output of `kiwi-ng` command.

* Added `:returncode:` line to avoid failing `kiwi-ng` command.

Fix a warning when calling "kiwi-ng". As the script returns 1, it is considered to have failed by the program-output directive. 

See more info here: https://pythonhosted.org/sphinxcontrib-programoutput/#error-handling


schaefi:
Ah thanks, already was wondering how I could indicate this warning is not a problem :)

PR #59, opened 2016-04-25
Selinux context

### LINKED ISSUE #52: File labels for systems using SELinux missing
## Problem description

Guests which use SELinux (such as Fedora and Red Hat Enterprise Linux) require that each file has a correct SELinux label.

## Actual behaviour

kiwi does not produce images with labels on files

## Solution pointer

http://libguestfs.org/virt-builder.1.html#selinux

fixfiles should be our friend

schaefi:
Added support for SELinux file security contexts
    
Systems using SELinux require the filesystem data to be labeled
according to a security context configuration. kiwi now checks
for the presence of /etc/selinux/targeted/contexts/files/file_contexts
and labels accordingly if it exists. This Fixes #52

### DIFF
--- a/kiwi/boot/image/kiwi.py
+++ b/kiwi/boot/image/kiwi.py
@@ -118,7 +118,7 @@ class BootImageKiwi(BootImageBase):
             # during the process of creating an image but not when the
             # image is actually booting with this initrd
             exclude_from_archive = [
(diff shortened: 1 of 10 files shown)


schaefi:
The changes are two fold, the first commit handles the different grub installation path between e.g SLES and RHEL. I found this when testiing rhel builds which makes use of selinux, suse doesn't

PR #68, opened 2016-04-30
Remove kiwi.logger.init() function

tomschr:
* The `init()` function is mostly unneccessary and can be directly created in the `kiwi.logger` module.


schaefi:
ok :smile:

PR #69, opened 2016-05-01
Use iterator protocol instead of StopIteration

tomschr:
* Replace `while/next()` code with `for line in self.command`

### Explanation
`CommandIterator` contains already the `__iter__` magic method. That means, we can use a `for` loop to iterate over all commands. We don't need to manually use `while` and `next()` in order to advance to the next element.

### Benefits
=> Code refactored to use for loops which does exactly the same job. We gain the following advantages:

* better readable
* avoids the nasty `try...except` block
* uses Python's iterator protocol which is highly "pythonic" :grin:

### DIFF
--- a/kiwi/command_process.py
+++ b/kiwi/command_process.py
@@ -40,16 +40,13 @@ class CommandProcess(object):
         """
         Iterate over process, raise on error and log output
         """
-        try:
-            while True:
-                line = next(self.command)
-                if line:
-                    log.debug('%s: %s', self.log_topic, line)
-        except StopIteration:
-            if self.command.get_error_code() != 0:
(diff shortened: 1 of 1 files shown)


tomschr:
Just for the sake of completeness:
I highly recommend to study the [`itertools`](https://docs.python.org/3.5/library/itertools.html#module-itertools) library. There are many great examples how to deal with iterables. Could also be useful for KIWI in some cases. :smiley:

schaefi:
Thanks while reading your changes, I can't remember exactly the reason why I directly called next()
I think it was because in a former implementation __next__ received additional parameters

anyway your change is of course an improvement :+1: going to merge it soon

PR #71, opened 2016-05-05
Add requirements section in Quick Start

tomschr:
* Describe what KIWI needs to run, further requirements, and for development
* Introduce `ghkiwi` as prefix in "extlinks" to shorten external links and
  to make linking to KIWI's GitHub repository more intuitive and consistent.
  For example, the following string is replaced:
  ```
:ghkiwi:`tox.ini` -> https://github.com/SUSE/kiwi/blob/master/tox.ini
```

@schaefi I wasn't sure if there are more requirements than libxml2/libxslt/libffi. Feel free to amend it, if necessary.

I mainly introduced this section to avoid surprises when a user clones this repo and tries to execute the script without knowing some requirements.


schaefi:
very nice thanks :+1:

PR #81, opened 2016-05-23
Added support for grub2 console switch entry

### LINKED ISSUE #80: grub.cfg should include hidden "t" menu and emit message
## Problem description

We have a new feature in Tumbleweed and SLES12SP2 grub2 which allows us to switch to text output with a simple key press. This is useful for environments, where grub2 runs in uEFI mode, finds graphical output and emits its menu there, leaving the serial console completely unused.

## Actual behaviour

We want to have these bits in the grub.cfg if we're targeting an EFI system:

[before the gfxmenu module gets loaded]
echo "Please press 't' to show the boot menu on this console"

[after all other menu entries]
 # On EFI systems we can only have graphics *or* serial, so allow the user
 # to switch between the two
hiddenentry 'Text mode' --hotkey 't' {
	set textmode=true
	terminal_output console
}

Try to boot with "qemu-system-x86_64 -serial mon:stdio -bios ovmf <disk>" and verify that there's nothing useful on the serial console. With the changes, you should get a message saying "Please press 't' to show the boot menu on this console" and be able to press the "t" key on serial to switch to text mode and thus have the menu shown on both graphical and serial output.

If kiwi doesn't use gfxmenu, this whole thing is moot, since they we already see the menu on both outputs.

schaefi:
An additional hidden entry allows the user to switch between the
graphics and the serial console by pressing the 't' key. The kiwi
grub2 template adds this switch if the terminal mode is set to
graphics. The additional entries are only effective of the
grub_platform is set to efi. Fixes #80

### DIFF
--- a/kiwi/bootloader/template/grub2.py
+++ b/kiwi/bootloader/template/grub2.py
@@ -45,6 +45,9 @@ class BootLoaderTemplateGrub2(object):
         ''').strip() + self.cr
 
         self.header_gfxterm = dedent('''
+            if [ "$${grub_platform}" = "efi" ]; then
+                echo "Please press 't' to show the boot menu on this console"
+            fi
             set gfxmode=${gfxmode}
             insmod all_video
             insmod gfxterm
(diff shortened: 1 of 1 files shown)


agraf:
lgtm

PR #83, opened 2016-05-30
Move all build and install tasks to setup.py

schaefi:
In an effort to distribute kiwi on pypi it should not be
required to call make targets for a complete installation.
Therefore the compilation of the C tools as well as the
installation of the man pages and the bash completion
has been added to setup.py. The spec file to build an rpm
package has been changed to use setup.py exclusively


kiwi/kiwi_compat.py
```diff
@@ -189,22 +187,21 @@ def __repo_argument(self, source, repo_type, alias, priority):
 class Command(object):
     @classmethod
     def execute(self, arguments):
-        # legacy kiwi binary lives in /usr/sbin, the new kiwi
-        # lives in /usr/bin. In order to avoid conflicts the path
-        # to the kiwi we want to use is specificly set here
-        kiwi = '/usr/bin/kiwi'
+        kiwi = '/usr/bin/kiwi-ng'
```
tomschr: Just some nitpicking here: if someone calls `./setup.py --home` or `./setup.py --prefix SOME_PATH` this would probably break.
Not sure how to solve it. *scratching my head*

setup.py
```diff
@@ -2,8 +2,89 @@
...
+from distutils.command import build
+from distutils.command import install
+#from distutils.command import install_scripts
+import distutils
+import subprocess
+
 from kiwi.version import __version__
 
+
+class kiwi_build(build.build):
```
tomschr: This is a minor issue. I've tried `./setup.py build --help` and it gives me this output message (shortend):

```
Global options:
  [...]
Options for 'kiwi_build' command:
```

This could be confusing for some as there is no really a "kiwi_build" command. Changing it to this helps (although it reduces clarity):

```python
class build(build.build):
   # ...
```

setup.py
```diff
@@ -2,8 +2,89 @@
...
+        self.announce(
+            subprocess.check_output(command).decode(),
+            level=distutils.log.INFO
+        )
+
+        # standard build process
+        build.build.run(self)
+
+
+class kiwi_install(install.install):
```
tomschr: Same here: kiwi_install -> install.

setup.py
```diff
@@ -2,8 +2,89 @@
...
+    """
+    sub_commands = [
+        ('install_lib', lambda self:True),
+        ('install_headers', lambda self:False),
+        ('install_scripts', lambda self:True),
+        ('install_data', lambda self:False),
+        ('install_egg_info', lambda self:True),
+    ]
+
+    def run(self):
```
tomschr: I'm not sure if this is a problem or you want to support that use case. I tried to install the package in a different location which is allowed with the "install" subcommand. I tried this:

```
$ ./setup.py install --prefix /tmp/kiwi
running install
Running make install target: ['make', 'install']
install: cannot stat ‘kversion’: No such file or directory
install: cannot stat ‘utimer’: No such file or directory
install: cannot stat ‘dcounter’: No such file or directory
install: cannot stat ‘startshell’: No such file or directory
install: cannot stat ‘isconsole’: No such file or directory
install: cannot stat ‘driveready’: No such file or directory
install: cannot stat ‘setctsid’: No such file or directory
make[1]: *** [install] Error 1
make: *** [install] Error 2
  File "./setup.py", line 130, in <module>
    setup(**config)
  File "/usr/lib64/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib64/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib64/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "./setup.py", line 80, in run
    subprocess.check_output(command).decode(),
  File "/usr/lib64/python2.7/subprocess.py", line 573, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command '['make', 'install']' returned non-zero exit status 2
```

setup.py
schaefi: nope, sorry I had it this way before and that causes an infinite load loop according to the way distutils functions. So I had  a reason for the name difference

setup.py
schaefi: same problem as above

setup.py
schaefi: did you ran build before ? it looks like the C tools were not compiled

kiwi/kiwi_compat.py
schaefi: yes this will be a problem, we could change it to a relative lookup, that would require the path to be in the search path of the shell, if not bad luck but I could live with that. Thoughts ?

setup.py
tomschr: Ahh ok. I've tried it, but it didn't give me invite loops. :) Worked for me. :) But it's only a minor thing.

setup.py
tomschr: Yes, I did, but I get the above message.

schaefi: ok let me try that too

schaefi: Hmm, works for me. I'm really sure there are no tools compiled in your case
could you check if 'setup.py build' has compiled tools below tools/

tomschr: Ok, I've started from scratch. Run "./setup.py build" and after that "./setup.py install --prefix /tmp/kiwi". This gives me a slightly different version (note the absolute paths):

```
$ ./setup.py install --prefix /tmp/kiwi
running install
Running make install target: ['make', 'install']
install: cannot create regular file ‘/usr/bin/kversion’: Permission denied
install: cannot create regular file ‘/usr/bin/utimer’: Permission denied
install: cannot create regular file ‘/usr/bin/dcounter’: Permission denied
install: cannot create regular file ‘/usr/bin/startshell’: Permission denied
install: cannot create regular file ‘/usr/bin/isconsole’: Permission denied
install: cannot create regular file ‘/usr/bin/driveready’: Permission denied
install: cannot create regular file ‘/usr/bin/setctsid’: Permission denied
make[1]: *** [install] Error 1
```

I'm not sure what caused this. I will try it on a different machine and try it there.

setup.py
tomschr: Ah, ok. Let's use your code. :)

setup.py
schaefi: I found a nicer solution to this problem, just a sec need to test it

.gitignore
```diff
@@ -1,8 +1,5 @@
-# compiled tools
-tools_bin
-
 # completion
-completion/kiwi.sh
+completion/kiwi-ng.sh
```
tomschr: Maybe add the `tools/dcounter` and the other binaries to the ignore list too?

setup.py
schaefi: yes this is normal, the install requires permissions to write the C tools to /usr/bin. the prefix setup in this case has no effect on the Makefile

not sure if this needs to be addressed ?

The same applies for the completion, it always lives in /etc/bash_completion.d the prefix will also not apply here

I guess this is the price we have to pay when combining linux targets with pure python code

Thoughts ?

.gitignore
schaefi: yes good idea, done

setup.py
schaefi: Ah wait, I transfer --root to the Makefile, so the following should work for you

setup.py build
setup.py install --root /tmp/kiwi

has worked for me

tomschr: It wooorks!! :+1: Cool, thanks!

schaefi: nice :smile: I think we have fixed all issues or do you see another problem which holds us back from merging this one ?

tomschr: Nothing. That was great! Thanks for this inspiring PR. :smile: Go! :+1:

schaefi:
ok, great works all nicely now, will merge once all tests are green. Thanks much for review and suggestions, really very much appreciated :+1:

PR #89, opened 2016-06-09
Remove xsltproc dependency

isbm:
* Do not call 3rd party XPath processor (`xsltproc`) but use Python library instead
* Allow XMLDescription to be loaded also from the XML in the string, instead of just a physical file

### DIFF
--- a/kiwi/exceptions.py
+++ b/kiwi/exceptions.py
@@ -131,6 +131,10 @@ class KiwiDataStructureError(KiwiError):
     pass
 
 
+class KiwiDescriptionConflict(KiwiError):
+    pass
+
+
 class KiwiDescriptionInvalid(KiwiError):
     pass
 
--- a/kiwi/xml_description.py
+++ b/kiwi/xml_description.py
@@ -22,7 +22,6 @@ from six import BytesIO
 from builtins import bytes
 
 # project
-from .command import Command
 from .defaults import Defaults
 from . import xml_parse
 
@@ -116,4 +115,3 @@ class XMLDescription(object):
         transform = etree.XSLT(etree.parse(Defaults.get_xsl_stylesheet_file()))
         with open(self.description_xslt_processed.name, "wb") as xsltout:
             xsltout.write(etree.tostring(transform(etree.parse(self.description))))
-
--- a/kiwi/xml_description.py
+++ b/kiwi/xml_description.py
@@ -29,7 +29,8 @@ from .exceptions import (
     KiwiSchemaImportError,
     KiwiValidationError,
     KiwiDescriptionInvalid,
-    KiwiDataStructureError
+    KiwiDataStructureError,
+    KiwiDescriptionConflict
 )
 
 
@@ -52,11 +53,19 @@ class XMLDescription(object):
     * :attr:`derived_from`
         path to base XML description file
 
+    * :attr:`xml_content`
+        XML description data as content string
     """
-    def __init__(self, description=None, derived_from=None, content=None):
+    def __init__(self, description=None, derived_from=None, xml_content=None):
+        if description and xml_content:
+            raise KiwiDescriptionConflict(
+                'description and xml_content are mutually exclusive'
+            )
         self.description_xslt_processed = NamedTemporaryFile()
-        self.description = content and BytesIO(bytes(content)) or description
         self.derived_from = derived_from
+
+        self.description = description
+        self.xml_content = xml_content
 
     def load(self):
         """
@@ -86,9 +95,11 @@ class XMLDescription(object):
                 '%s: %s' % (type(e).__name__, format(e))
             )
         if not validation_ok:
-            raise KiwiDescriptionInvalid(
-                'Schema validation for %s failed' % self.description
-            )
+            if self.description:
+                message = 'Schema validation for %s failed' % self.description
+            else:
+                message = 'Schema validation for XML content failed'
+            raise KiwiDescriptionInvalid(message)
         return self.__parse()
 
     def __parse(self):
@@ -96,7 +107,7 @@ class XMLDescription(object):
             parse = xml_parse.parse(
                 self.description_xslt_processed.name, True
             )
-            parse.description_dir = os.path.dirname(
+            parse.description_dir = self.description and os.path.dirname(
                 self.description
             )
             parse.derived_description_dir = self.derived_from
@@ -108,10 +119,21 @@ class XMLDescription(object):
 
     def _xsltproc(self):
         """
-        Transform XML description with the Kiwi rules.
+        Apply XSLT style sheet rules to the XML data
 
-        :return:
+        The result of the XSLT processing is stored in a named
+        temporary file and used for further schema validation
+        and parsing into the data structure classes
         """
-        transform = etree.XSLT(etree.parse(Defaults.get_xsl_stylesheet_file()))
+        xslt_transform = etree.XSLT(
+            etree.parse(Defaults.get_xsl_stylesheet_file())
+        )
+
+        xml_source = self.description
+        if self.xml_content:
+            xml_source = BytesIO(bytes(self.xml_content))
+
         with open(self.description_xslt_processed.name, "wb") as xsltout:
-            xsltout.write(etree.tostring(transform(etree.parse(self.description))))
+            xsltout.write(
+                etree.tostring(xslt_transform(etree.parse(xml_source)))
+            )
--- a/kiwi/xml_description.py
+++ b/kiwi/xml_description.py
@@ -129,9 +129,8 @@ class XMLDescription(object):
             etree.parse(Defaults.get_xsl_stylesheet_file())
         )
 
-        xml_source = self.description
-        if self.xml_content:
-            xml_source = BytesIO(bytes(self.xml_content))
+        xml_source = self.description if self.description else \
+            BytesIO(bytes(self.xml_content))
 
         with open(self.description_xslt_processed.name, "wb") as xsltout:
             xsltout.write(
--- a/kiwi/xml_description.py
+++ b/kiwi/xml_description.py
@@ -18,6 +18,7 @@
 from lxml import etree
 from tempfile import NamedTemporaryFile
 import os
+from six import BytesIO
 
 # project
 from .command import Command
@@ -52,9 +53,9 @@ class XMLDescription(object):
         path to base XML description file
 
     """
-    def __init__(self, description, derived_from=None):
+    def __init__(self, description=None, derived_from=None, content=None):
         self.description_xslt_processed = NamedTemporaryFile()
-        self.description = description
+        self.description = content and BytesIO(content) or description
         self.derived_from = derived_from
 
     def load(self):
@@ -66,7 +67,7 @@ class XMLDescription(object):
         :return: instance of XML toplevel domain (image)
         :rtype: object
         """
-        self.__xsltproc()
+        self._xsltproc()
         try:
             relaxng = etree.RelaxNG(
                 etree.parse(Defaults.get_schema_file())
--- a/kiwi/xml_description.py
+++ b/kiwi/xml_description.py
@@ -107,11 +107,11 @@ class XMLDescription(object):
             )
 
     def _xsltproc(self):
-        '''
+        """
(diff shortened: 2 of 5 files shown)


kiwi/xml_description.py
```diff
@@ -52,9 +53,9 @@ class XMLDescription(object):
         path to base XML description file
 
     """
-    def __init__(self, description, derived_from=None):
+    def __init__(self, description=None, derived_from=None, content=None):
         self.description_xslt_processed = NamedTemporaryFile()
-        self.description = description
+        self.description = content and BytesIO(content) or description
```
aplanas: `self.description = BytesIO(content) if content else description`

But I found this a bit wrong that content have move priority than description when both are passed. I would drop content here

kiwi/xml_description.py
```diff
@@ -105,11 +106,13 @@ def __parse(self):
...
-    def __xsltproc(self):
-        Command.run(
-            [
-                'xsltproc', '-o', self.description_xslt_processed.name,
-                Defaults.get_xsl_stylesheet_file(),
-                self.description
-            ]
-        )
+    def _xsltproc(self):
+        '''
```
aplanas: """?

kiwi/xml_description.py
isbm: You're saying exactly the same thing, just a different syntax. :wink:

aplanas: Ah sure, is a matter of clarity. `and` and `or` was idiomatic like in 2.4 (10 years ago?) Still confusing the precedence of content over description for the value of self.description. This precedence is even more clear with the `if` expression

isbm: Matter of taste.

kiwi/xml_description.py
isbm: Actually, having developing Salt, we hate double-quotes there and using single all everywhere, although PEP8 leans to the double-quotes. Old habits die hard...

While I fixed this, I would suggest not to make it religious, unless @schaefi has a strict PyLint on this.

kiwi/xml_description.py
```diff
@@ -52,9 +54,9 @@ class XMLDescription(object):
         path to base XML description file
 
     """
-    def __init__(self, description, derived_from=None):
+    def __init__(self, description=None, derived_from=None, content=None):
         self.description_xslt_processed = NamedTemporaryFile()
-        self.description = description
+        self.description = content and BytesIO(bytes(content)) or description
```
aplanas: What do you want to accomplish here? I see three problems

* hard to read because Boolean circuits instead of using Python
* `content` over `description` (why can't you simply not use description instead of content? or if content is a better name, why not to refactor and change the name of description to content?)
* type mangling, why force it to byte instead of using string? Some python 3 issue? Why not to fix the type issue in the caller and not in the constructor?

isbm: > What do you want to accomplish here?

What does it do, you think?

aplanas: @isbm I am sorry if I am upsetting you, not my intention.

Code is to be read so others can understand it. You are submitting the code, and I am this 'other' that is trying to understand it, and to do that I am making some reasonable and honest technical questions. You can decide to answer it or not, but in any case, please, avoid this kind of bullshit.

isbm: > hard to read
That's [normal](http://gamedevwithoutacause.com/?p=1329). Also could you please watch your tone? If you don't like it, PR please. Bye.

schaefi: I only see a problem with the use of self.description. The scope of it is now two fold, it could be either the path to a file name or raw content which is not a problem for etree.parse but at other places e.g if an exception occurs we use self.description in the exception message and it would make only sense if it would be in the scope of a filename. In addition the member variable parse.description_dir should be None if raw content will be used and would be set to the content which might be confusing

How about dealing with self.description and self.content instead of merging it at the object construction level
e.g

```python
self.description = description
self.content = content

...

etree.parse(self.content and BytesIO(bytes(self.content)) or self.description)

...handle output of KiwiDescriptionInvalid exception accordingly

```

I think that would put the logic at the place where it is needed and follows more the explicit is better than implicit python advise

Thoughts ?

isbm: OK, the exception handling issue makes perfect sense as well as the directory variable. Will change that accordingly.

aplanas: @schaefi Aaah more clear IMHO. There are some bits that can help too:

* Raise an exception when both description and content are provided. The prevalence of one over another at least needs to be documented in the header (also missing)
* I am a bit worried with the type mangling. `content` now expect bytes and can misbehave in unicode. The use of `six` indicates me that is expected to work on Py2 and Py3 so this can be a problem. Can be helpful to add test cases for unicode and str (py2) and str (py3)

I'm sure that @isbm had good reasons to use bytes (I asked to understand them), but I am curious why not use `ET.fromstring` or `ET.parse` directly to manage file and string cases.

One note also, `from builtins import bytes` is not going to work in py2, but maybe this test is only executed in Py3

isbm: @aplanas tests are passing on both sides. In Salt we use `ext` from the beginning. Yet we have way less noise in the comments. :wink:

schaefi:
@isbm do you plan to work on this request next week ? if not it's ok and if ok with you I would take over the branch because it's a great idea to get rid of the external xsltproc dependency for the next release which is scheduled for next week. Just let me know, thanks

isbm:
@schaefi Yes, in general I plan. But if you have a time, you could also push changes to this branch, of course!

kiwi/xml_description.py
```diff
@@ -86,17 +95,19 @@ def load(self):
                 '%s: %s' % (type(e).__name__, format(e))
             )
         if not validation_ok:
-            raise KiwiDescriptionInvalid(
-                'Schema validation for %s failed' % self.description
-            )
+            if self.description:
+                message = 'Schema validation for %s failed' % self.description
```
isbm: Nitpick of formatting: lately we use in Python `.format(...)` instead of this way. Example:
```python
message = 'Schema validation of {0} had failed'.format(self.description)
```
Or with parameters:
```python
message = 'Schema validation of {descr} had failed'.format(descr=self.description)
```
JFYI.

test/unit/xml_description_test.py
```diff
@@ -27,35 +47,49 @@ def test_load_schema_import_error(self, mock_xslt, mock_relax):
...
+        self.description_from_file.load()
 
     @raises(KiwiValidationError)
     @patch('lxml.etree.RelaxNG')
     @patch('lxml.etree.parse')
     @patch.object(XMLDescription, '_xsltproc')
-    def test_load_schema_validation_error(
+    def test_load_schema_validation_error_from_file(
         self, mock_xslt, mock_parse, mock_relax
     ):
```
isbm: This is compilable but I would suggest not to format it that way due to [indentation as per PEP8](https://www.python.org/dev/peps/pep-0008/#indentation). I would offer several ways.

As a member of Emacs religion:
```python
    def test_load_schema_validation_error_from_file(self, mock_xslt,
                                                    mock_parse, mock_relax):
    # Code here
```
Or:
```python
def test_load_schema_validation_error_from_file(
        self, mock_xslt, mock_parse, mock_relax):
    # Code here
```
Or:
```python
def test_load_schema_validation_error_from_file(self,
                                                mock_xslt,
                                                mock_parse,
                                                mock_relax):
    # Code here
```
In all cases the code starts from the left side from `):` thing.

kiwi/exceptions.py
```diff
@@ -131,6 +131,10 @@ class KiwiDataStructureError(KiwiError):
     pass
 
 
+class KiwiDescriptionConflict(KiwiError):
+    pass
```
isbm: Instead of `pass` I would just add a documentation here. Something like:
```python
# No
class KiwiDescriptionConflict(KiwiError):
    pass

# Yes
class KiwiDescriptionConflict(KiwiError):
    """
    Documentation, that describes a purpose of this class.
    Below of this documentation string there is nothing else.
    """
```

kiwi/xml_description.py
```diff
@@ -52,12 +53,20 @@ class XMLDescription(object):
     * :attr:`derived_from`
         path to base XML description file
 
+    * :attr:`xml_content`
+        XML description data as content string
     """
-    def __init__(self, description=None, derived_from=None, content=None):
+    def __init__(self, description=None, derived_from=None, xml_content=None):
```
isbm: Well, if we are renaming `content` to `xml_content` (are we expect something like `json_content` or `yaml_content` too?) then the `description` is actually just a path to the file. Maybe it can be `path` now?

I would keep `content` instead.

kiwi/xml_description.py
```diff
@@ -108,10 +119,21 @@ def __parse(self):
...
         """
-        transform = etree.XSLT(etree.parse(Defaults.get_xsl_stylesheet_file()))
+        xslt_transform = etree.XSLT(
+            etree.parse(Defaults.get_xsl_stylesheet_file())
+        )
+
+        xml_source = self.description
+        if self.xml_content:
+            xml_source = BytesIO(bytes(self.xml_content))
+
```
isbm: I always try to avoid re-assigning and thus I would write it:
```python
if self.xml_content:
    xml_source = BytesIO(bytes(self.xml_content))
else:
    xml_source = self.description
```
But this is looks like a Pascal in my student's ages. Therefore use ternary operator:
```python
xml_source = BytesIO(bytes(self.xml_content)) if self.xml_content else self.description
```
Or 1 char shorter:
```python
xml_source = self.xml_content and BytesIO(bytes(self.xml_content)) or self.description
```

test/unit/xml_description_test.py
```diff
@@ -3,22 +3,42 @@
 
 from .test_helper import *
 
-from kiwi.exceptions import (
-    KiwiSchemaImportError,
-    KiwiValidationError,
-    KiwiDescriptionInvalid,
-    KiwiDataStructureError
-)
+from kiwi.exceptions import *
```
isbm: Please don't. Stars are _very bad_ style and [PEP8 hates that](https://www.python.org/dev/peps/pep-0008/#imports).

schaefi: we "allow" the wildcard import only in the tests, wouldn't that be fair enough ?
exceptions not handled in a test will result in a coverage error, thus I thought it is ok to have a wildcard import for exceptions in tests

kiwi/exceptions.py
schaefi: I agree, I will add doc strings explaining the exceptions in an extra pull request ok ?

test/unit/xml_description_test.py
isbm: Test is also code. :wink: Personally I don't do it anywhere and I always want to know what is imported. But your call...

kiwi/xml_description.py
schaefi: nice, that is easier to read, will make use of it

kiwi/exceptions.py
schaefi: opened branch exceptions-doc-strings to track this one

schaefi:
if you think we are good to go, give me a thumbs up

I'll fix the doc strings for exceptions next... boring but must be done :smile:

test/unit/xml_description_test.py
```diff
@@ -1,67 +1,109 @@
...
+        self.description_from_file = XMLDescription(description='description')
+        self.description_from_data = XMLDescription(xml_content=test_xml)
+
+    @raises(KiwiDescriptionConflict)
+    def test_constructor_conflict(self):
+        XMLDescription(description='description', xml_content='content')
+
+    def test_load_schema_from_xml_content(self):
+        parsed = self.description_from_data.load()
+        assert parsed.get_schemaversion() == '6.3'
```
isbm: Maybe the schema version can be picked up from somewhere, so we don't need to fix this test all the time XSL template changes? (Just a thought)

schaefi: good idea

kiwi/exceptions.py
schaefi: request #97 open for review

isbm:
Perfect. Go for it. :+1:

PR #90, opened 2016-06-09
Debian based distro support

schaefi:
Add support for building Debian based Appliances

Along with the change an example image description for Ubuntu Xenial can be found here
https://github.com/SUSE/kiwi-descriptions/tree/master/ubuntu/x86_64/ubuntu-xenial-JeOS

So far simple virtual disk images, vmx type together with the dracut initrd system can be build
support for oem, pxe and live iso types are still about to be added


kiwi/bootloader/config/grub2.py
```diff
@@ -461,7 +461,25 @@ def __get_module_path(self, format_name, lookup_path=None):
             ]
         )
 
+    def __fixup_legacy_grub_location(self):
+        legacy_grub_theme_dir = self.root_dir + '/boot/grub/themes/'
+        legacy_font = self.root_dir + '/boot/grub/unicode.pf2'
+        grub_dir = self.root_dir + \
+            '/boot/' + self.boot_directory_name + '/'
```
isbm: Should be `os.path.join` here

kiwi/bootloader/config/grub2.py
```diff
@@ -461,7 +461,25 @@ def __get_module_path(self, format_name, lookup_path=None):
...
+        if os.path.exists(legacy_grub_theme_dir):
+            # found grub2 theme directory in legacy grub directory
+            Path.create(grub_dir)
+            Command.run(
+                ['cp', '-a', legacy_grub_theme_dir, grub_dir]
+            )
```
isbm: I would avoid `Command.run` at all costs. Especially if there is [`shutil`](https://docs.python.org/3/library/shutil.html) for this.

kiwi/bootloader/config/grub2.py
```diff
@@ -461,7 +461,25 @@ def __get_module_path(self, format_name, lookup_path=None):
...
+        if os.path.exists(legacy_font):
+            # found grub2 unicode font in legacy grub directory
+            Command.run(
+                ['cp', legacy_font, grub_dir]
+            )
```
isbm: See above about [`shutil`](https://docs.python.org/3/library/shutil.html).

kiwi/package_manager/apt.py
```diff
@@ -0,0 +1,260 @@
...
+        chroot_apt_get_args = self.root_bind.move_to_root(
+            self.apt_get_args
+        )
+        return Command.call(
+            [
+                'chroot', self.root_dir, 'apt-get'
+            ] + chroot_apt_get_args + self.custom_args + [
+                'remove'
+            ] + self.__package_requests(), self.command_env
+        )
```
isbm: I would write it more readable.

```python
def process_delete_requests(self, force=False):
    """
    Process package delete requests (chroot)

    :param bool force: unused
    """
    cmd = ['chroot', self.root_dir, 'apt-get']
    cmd.extend(self.root_bind.move_to_root(self.apt_get_args))
    cmd.extend(self.custom_args)
    cmd.append('remove')
    cmd.extend(self.__package_requests())

    return Command.call(cmd, self.command_env)
```

And yes. Why param `force` is unused? :wink:

kiwi/package_manager/apt.py
```diff
@@ -0,0 +1,260 @@
...
+
+        :param string version: unused
+        """
+        pass
+
+    def __package_requests(self):
+        items = []
+        items += self.package_requests
+        self.cleanup_requests()
+        return items
```
isbm: You want to make a copy here, so the `package_requests` modification won't affect `items`? Then:
```python
items = self.package_requests[:]
self.cleanup_requests()
return items
```

kiwi/repository/apt.py
```diff
@@ -0,0 +1,200 @@
...
+            'apt_get_args': self.apt_get_args,
+            'command_env': self.command_env,
+            'distribution': self.distribution,
+            'distribution_path': self.distribution_path
+        }
+
+    def add_repo(
+        self, name, uri, repo_type='deb',
+        prio=None, dist=None, components=None
+    ):
```
isbm: Nope. Either all one line (best in this case) or vertical, one param per a time, as per PEP8.

kiwi/repository/apt.py
```diff
@@ -0,0 +1,200 @@
...
+
+        :param string name: repository base file name
+        :param string uri: repository URI
+        :param repo_type: unused
+        :param int prio: unused
+        :param dist: distribution name for non flat deb repos
+        :param components: distribution categories
+        """
+        list_file = self.shared_apt_get_dir['sources-dir'] + \
+            '/' + name + '.list'
```
isbm: Actually you don't have to stick to 79 per PEP8. :wink: Especially when `\` comes to the deal.

kiwi/repository/apt.py
```diff
@@ -0,0 +1,200 @@
...
+        Repository configurations which are not used for this build
+        must be removed otherwise they are taken into account for
+        the package installations
+        """
+        repos_dir = self.shared_apt_get_dir['sources-dir']
+        for elements in os.walk(repos_dir):
+            for repo_file in list(elements[2]):
+                if repo_file not in self.repo_names:
+                    Path.wipe(repos_dir + '/' + repo_file)
+            break
```
isbm: This will always stop after the 1st element. Is this is what you want?

kiwi/repository/template/apt.py
```diff
@@ -0,0 +1,84 @@
...
+from string import Template
+from textwrap import dedent
+
+
+class PackageManagerTemplateAptGet(object):
+    """
+    apt-get configuration file template
+    """
+    def __init__(self):
+        self.cr = '\n'
```
isbm: `os.linesep`

kiwi/repository/template/apt.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        apt-get package manager template for apt-get called
+        outside of the image, not chrooted
+
+        :rtype: Template
+        """
+        template_data = self.host_header
+        template_data += self.apt
+        template_data += self.dpkg
+        return Template(template_data)
```
isbm: No. Don't. Strings are immutable, concatenation of them is painful. Instead:
```python
return Template(''.join([self.host_header, self.apt, self.dpkg]))
```

kiwi/repository/template/apt.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        apt-get package manager template for apt-get called
+        inside of the image, chrooted
+
+        :rtype: Template
+        """
+        template_data = self.image_header
+        template_data += self.apt
+        template_data += self.dpkg
+        return Template(template_data)
```
isbm: See above.

kiwi/repository/yum.py
```diff
@@ -104,14 +104,19 @@ def runtime_config(self):
             'command_env': self.command_env
         }
 
-    def add_repo(self, name, uri, repo_type='rpm-md', prio=None):
+    def add_repo(
+        self, name, uri, repo_type='rpm-md',
+        prio=None, dist=None, components=None
+    ):
```
isbm: Oneliner is just fine.

test/unit/package_manager_apt_test.py
```diff
@@ -0,0 +1,170 @@
...
+                ], ['env']
+            )
+        ]
+        mock_call.assert_called_once_with(
+            [
+                'chroot', 'root-dir', 'apt-get',
+                'root-moved-arguments', 'install', 'vim'
+            ], ['env']
+        )
+
```
isbm: You can contract all of that above to a bit more sparse code. :wink: Up to you though, but I would prefer less scrolling over ```[
     [ 
               {
        [
                      (
           [
                        (
               [
                   (
                              (
           [
                    (
                       )
              ]
               ,
           ]
      )
                  ]
         }
     ]
```
:wink:

kiwi/repository/apt.py
schaefi: It is what I want but it is bad style, I'm going to change that

schaefi:
Thanks for the review, I think we are good to go now

Thoughts ?

PR #93, opened 2016-06-13
Optimize RootInit.create() from kiwi/system/root_init.py

tomschr:
* Use a tuple in `base_system_paths` instead of a list
* Optimize "mknod" call by using for loop
* Optimize "ln -s" call by using for loop

Hope you like it. ;-)

### DIFF
--- a/kiwi/system/root_init.py
+++ b/kiwi/system/root_init.py
@@ -15,6 +15,8 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+import stat
+from pwd import getpwnam
 from tempfile import mkdtemp
 from shutil import rmtree
 import os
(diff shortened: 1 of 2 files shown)


isbm:
I would suggest not to. This is like pouring from hollow to empty. :wink: If you are re-factoring, then what is wrong with [`os.mknod`](https://docs.python.org/2/library/os.html#os.mknod) and [`os.symlink`](https://docs.python.org/2/library/os.html#os.symlink) respectively?

kiwi/system/root_init.py
```diff
@@ -72,60 +73,42 @@ def create(self):
...
+                          "mknod -m 666 %s/dev/ptmx c 5 2",
+                          "mknod -m 640 %s/dev/loop0 b 7 0",
+                          "mknod -m 640 %s/dev/loop1 b 7 1",
+                          "mknod -m 640 %s/dev/loop2 b 7 2",
+                          "mknod -m 666 %s/dev/loop3 b 7 3",
+                          "mknod -m 666 %s/dev/loop4 b 7 4",
+                          )
+            for cmd in map(shlex.split, mknod_data):
+                cmd[3] = cmd[3] % root
+                Command.run(cmd)
```
isbm: Use `os.mknod` instead, please.

kiwi/system/root_init.py
```diff
@@ -72,60 +73,42 @@ def create(self):
...
-            Command.run(['ln', '-s', 'fd/0', root + '/dev/stdin'])
-            Command.run(['ln', '-s', 'fd/1', root + '/dev/stdout'])
-            Command.run(['ln', '-s', '/run', root + '/var/run'])
+            for src, target in (('/proc/self/fd', '%s/dev/fd'),
+                                ('fd/2', '%s/dev/stderr'),
+                                ('fd/0', '%s/dev/stdin'),
+                                ('fd/1', '%s/dev/stdout'),
+                                ('/run', '%s/var/run'),
+                                ):
+                Command.run(['ln', '-s', src, target % root])
```
isbm: Use `os.symlink` instead, please.

kiwi/system/root_init.py
```diff
@@ -72,60 +73,42 @@ def create(self):
...
+                          "mknod -m 666 %s/dev/tty c 5 0",
+                          "mknod -m 666 %s/dev/ptmx c 5 2",
+                          "mknod -m 640 %s/dev/loop0 b 7 0",
+                          "mknod -m 640 %s/dev/loop1 b 7 1",
+                          "mknod -m 640 %s/dev/loop2 b 7 2",
+                          "mknod -m 666 %s/dev/loop3 b 7 3",
+                          "mknod -m 666 %s/dev/loop4 b 7 4",
+                          )
+            for cmd in map(shlex.split, mknod_data):
+                cmd[3] = cmd[3] % root
```
isbm: Please use `os.path.join` instead.

kiwi/system/root_init.py
```diff
@@ -72,60 +73,42 @@ def create(self):
...
+                          "mknod -m 644 %s/dev/urandom c 1 9",
+                          "mknod -m 666 %s/dev/tty c 5 0",
+                          "mknod -m 666 %s/dev/ptmx c 5 2",
+                          "mknod -m 640 %s/dev/loop0 b 7 0",
+                          "mknod -m 640 %s/dev/loop1 b 7 1",
+                          "mknod -m 640 %s/dev/loop2 b 7 2",
+                          "mknod -m 666 %s/dev/loop3 b 7 3",
+                          "mknod -m 666 %s/dev/loop4 b 7 4",
+                          )
+            for cmd in map(shlex.split, mknod_data):
```
isbm: Shlex!! Why you would need that? What is wrong with somthing like:
```python
mknod_data = ((0o666, '/dev/null', stat.S_IFCHR, os.makedev(1, 3)),
              (0o666, '/dev/zero', stat.S_IFCHR, os.makedev(1, 5)),)
for data in mknod_data:
    d_mode, d_path, d_type, d_dev = data
    os.mknod(os.path.join(roo, d_path), d_mode | d_type, d_dev)
```

?

tomschr: @isbm I know os.mknod, os.path.join, and the like. :-)

I wanted to start with a _minimal_ change. Without disrupting everything. I thought of using the Python functions from the standard library too. However, I decided to contribute a _minimal_ change first and optimize it later.

If you like, you can try to optimize my PR. Don't forget the test cases. ;-)

isbm: I think it is a good start, but makes no much sense to  be merged. What exactly does it fix? Nothing really. So it would be much more helpful if we could get rid of all `Command.run` stuff whenever it is possible and replace with the equivalent Python library. This is why I would not recommend to merge this.

tomschr: Well, I guess, Marcus implemented the Command class for some reason. As far as I can see, it logs things. Which isn't possible with the native commands.

> What exactly does it fix?

Style & optimization. ;-P

isbm: `logger`.

schaefi: I did the Command class basically for external calls where we do not have a native python call for. The reason why I did not use os.symlink, os.mknod was just lack of knowledge when I started with the root init code

So I would be perfectly fine if we handle the move from the external calls into an os. variant
and merge that as a refactoring commit.

@tomschr If you could do that as part of this branch it would be really great, thanks for the effort

tomschr: > If you could do that as part of this branch it would be really great, thanks for the effort

I'll do that, but I need some more time as I'm busy with other stuff.

isbm: Thanks @tomschr ! No hurry here.

schaefi:
Thanks Tom for your effort, as discussed on the phone I added your work to the branch and continue the refactoring

tomschr:
@schaefi Thanks! Great work!

schaefi:
cool, let's merge

PR #97, opened 2016-06-15
Exceptions doc strings

schaefi:
Add doc information for Exceptions

### DIFF
--- a/test/unit/package_manager_apt_test.py
+++ b/test/unit/package_manager_apt_test.py
@@ -7,9 +7,7 @@ import re
 from .test_helper import *
 
 from kiwi.package_manager.apt import PackageManagerApt
-from kiwi.exceptions import (
-    KiwiDebootstrapError
-)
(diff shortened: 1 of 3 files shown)


test/unit/package_manager_apt_test.py
```diff
@@ -7,7 +7,9 @@
 from .test_helper import *
 
 from kiwi.package_manager.apt import PackageManagerApt
-from kiwi.exceptions import *
+from kiwi.exceptions import (
+    KiwiDebootstrapError
+)
```
isbm: No need of parenthesis in this case. Only when multiple imports.

isbm:
Is this intentional having no "." at the end of every sentence? Normally, this documentation you see in IDEs, it is generated etc. So typically it supposed to be just a proper grammar.

schaefi:
No this is just me being lazy, I'll fix that... duh you might see this at several other doc strings *urks*

isbm:
Done. :+1:

Issue #40, opened 2016-03-06
Docker images doesn't require lxc configuration


schaefi:
---mail snippet Aleksa Sarai <asarai@suse.de> ---

I just want to point out that I'm not entirely sure why we generate LXC
configurations (Docker hasn't used the LXC execdriver as the default
for about 2 years, and support for it was entirely dropped in Docker
1.10 -- a month ago).

ChrisBr:
Removed the LXC configuration files and the container still works.
@schaefi what about the LXC folder? Do we want to remove the whole folder also?

schaefi:
yes I think we don't need the directory, but I can't tell for sure

PR #99, opened 2016-06-20
Add support LXD container image format

matt4d617474:
This is my first GitHub pull request.  Please be gentle.  :-)

I'm happy to change things if needed.

Thanks!
Matt


kiwi/container/lxd.py
```diff
@@ -0,0 +1,268 @@
...
+        else:
+            return 'legacy'
+
+    def _has_lzma(self):  # pragma: no cover
+        try:              # ... bit tricky to cover missing imports
+            import lzma
+            lzma.is_check_supported(lzma.CHECK_NONE)  # to remedy "unused import" chirp
+            return True
+        except:
+            return False
```
isbm: It can be shorter.

```python
try
    import lzma
except ImportError as err:
    lzma = None

...

# Remove this:
def _get_creation_method(self):
    pass

def create(self, filename):
    getattr(self, "_create_{0}".format(sys.version_info > (3, 3) and lzma is not None and 'new' or 'legacy'))(filename)
```

isbm:
@matt4d617474 Please also make sure Python 2.7 is passing. :wink:

matt4d617474:
Doh!  Re: py27 failing, it was caused from the TravisCI underlying OS (for the v27 builds, at least) having an older GNU tar.  So that issue has always been there, independent of my code change, it was just that my tests were of sufficient integration-style to exercise the kiwi.archive.tar.ArchiveTar functionality requiring extended attributes support (in v1.27+ of GNU tar).  I've since patched kiwi.archive.tar.ArchiveTar to check for the version and send more appropriate arguments.

But another approach would be to require v1.27+ of GNU tar if we deem --xattr* option support critical.

*shrugs*

schaefi:
Hi Matt, thanks for your effort this is very much appreciated

I had a look to the changes and before we continue with a closer look to the code I have a few questions and suggestions.

* You distinguish the capabilities to create a tarball as part of the kiwi/container/lxd.py code.
  It would be nice if this can be done in the class which imho should deal with it. If you could extend
  kiwi/archive/tar.py in a way that it decides for the tarfile module or the call of the external tar
  command this would be imho more general and re-usable at any place in the code.
  In addition this change could be sent in as an extra first pull request. Would be great if you
  can do that. In general a class should implement a single responsibility, it's not the task of the
  Lxd class to care for which tar can do what I need

* I did not understand for what use case you added the _import_custom_directories() functionality
  At the moment I don't see why we need this. Maybe you could explain this part ? In any case it
  would be an additional functionality which should be an extra pull request like the tar extensions.
  Consider it preparation changes for Lxd support

Overall it looks nice to me, we just need to move a few parts around

Thoughts ?

schaefi:
With regards to tar; I don't have any objections if the use of the tarfile module will become the preferred method if possible

schaefi:
Ah and a very small part you might want to change is the initial description :smile: 

Thanks

matt4d617474:
Hey there, Marcus!  Thanks very much for the time and feedback.  Apologies for the crazy delay in response (been more than swamped at work sadly).

1) Agree that I should put that determination with the other tar code, Principle of Least Astonishment and all.  My only question is whether or not I should also extend ArchiveTar to handle my use case entirely (again, to have all tar support in one spot).  Initially, I wanted to make as surgical a change as possible.  And I kind of wanted to perhaps close https://github.com/SUSE/kiwi/issues/40 (looks already closed) in supporting the OCF (libcontainer/opencontainers.org format) which feels very similar to LXD (config.json and a rootfs).  But looks someone (today?! <grin/>) already started in that?  I'll probably stay surgical in the follow-up pull request re: tar code unless you say otherwise.

2) The reason I wanted an ability to copy paths/directories (in addition to the preexisting scripts/archives) is that I need a place to put the LXD metadata that moves from the prepare step to the create step.  The lxd.py code looks for <root_dir>/image/lxd_metadata/{metadata.yaml,templates/}.  LXD's metadata isn't placed in the rootdir proper.  Is there a more idiomatic way to pass out-of-band data between the phases?  For what it's worth, using what I did above felt pretty natural as I've been using it quite a lot these last few weeks.

Thanks again!

schaefi:
yes Christian worked on the docker container support Issue #40 and fixed that. I'm fine if you continue to extend ArchiveTar in a way that it will become useful for your use case and generic enough for all others. An extra pull request would be great, that will be the first part we can merge then

schaefi:
with regards to the metadata you need to create I have a couple questions:

1.  metadata.yaml seems to be a mandatory setup file. The contents according to the test
     are: arch, date and description which can all be found in the kiwi XML description or generated.
     Thus I think this file should be generated as part of the ContainerSetupLxd class rather than
     statically copied from <description_dir>/lxd_metadata

2. templates/  is marked as optional data. I do not understand why this information can not
    be put into <description_dir>/root/image/templates. All data in the __root__ overlay directory 
    is automatically synced to the root system. Your code expects the templates below /image
    so we can just put it as such below <description_dir>/root/. An extra <description_dir>/lxd_metadata/
    directory seems not needed and along with it the copy and import functions you wrote

Thoughts ?

schaefi:
@matt4d617474 Would you mind if I continue with the extensions to tar ? I need parts of the functionality started here for other code in kiwi ?

matt4d617474:
Hey there Marcus.  :-)

Thanks for being so kind and asking!  I think I can have something for you in a day so as to not duplicate work if you're willing to wait that long?  Apologies for catching me at a stupid busy time in life.

schaefi:
sure, I'll wait, thanks for looking into this :+1:

davidcassany:
@matt4d617474 , I'm just letting you know we will continue with the extensions to tar commented above from our side. Of course will help with any resulting conflict that may arise with the refactored kiwi/archive/tar

matt4d617474:
David/Marcus - You all have been more than patient.  Apologies for the unanticipated delay.  After your changes land, I will work to get things worked together.  Thanks!

davidcassany:
@matt4d617474  I have done already a pull request to update the kiwi/archive/tar class. Finally we decided not to use the tarfile python module, since this module does not support xz compression in Python v2.7 and neither support for some important tar options like _numeric-owner_, which are used in kiwi code. However I have included your tar version control and xattrs options handling. Just a couple of things of how your code could be updated not no use the tarfile:

- I noticed you are appending files into already existing tarball (e.g. the metadata.yaml file). In order to do so I have suggested a ArchiveTar.append_files method to provide this functionality.
- I also noticed you are using the arcname option of tarfile in order to set the top directory name of your tarball to rootfs. This is something that can already be done using the appropriate parameters in ArchiveTar.create, see the following create call:

`archive.create('source_dir', options=['--transform', 's|^|roofs/|]')`

It will place all the items inside _source_dir_ inside a top directory called roofs. See _transform_ docs and examples [here](https://www.gnu.org/software/tar/manual/html_section/tar_52.html#transform).

schaefi:
Matt, hope you don't mind but I'll close this pull request now because in the current form it won't merge. You can rebase / solve conflicts of your branch and send a new request once you are done with your work.

Thanks

PR #102, opened 2016-06-29
Remove lxc configuration from docker images

### LINKED ISSUE #40: Docker images doesn't require lxc configuration

ChrisBr:
Fixes #40.
Docker does not use LXC execdriver anymore.

* Removes create_lxc_fstab function
* Removes create_lxc_config function

### DIFF
--- a/kiwi/container/setup/docker.py
+++ b/kiwi/container/setup/docker.py
@@ -17,7 +17,6 @@
 #
 
 # project
-from ...path import Path
 from .base import ContainerSetupBase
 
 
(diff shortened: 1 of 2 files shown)


schaefi:
:+1:

PR #107, opened 2016-07-05
Volume name limitations

### LINKED ISSUE #39: No name limitations for volumes by kiwi
kiwi uses the underscore (_) as directory separator because slash (/) is an invalid character for a shell variable. The volume information is turned into shell variables of the profile environment which introduces that limitation.

This means users can't specify an underscore in the mountpoint= and/or name= attributes. This limitation is bad because the underscore is a quite common character for volume names and allowed in all supported volume management systems.

In order to fix this limitation the way the kiwi initrd code handles the volume information needs to change first

For details watch:

* kiwi/xml_state::get_volumes()
* search for kiwi_LVM_ in kiwi/boot/functions.sh
* search for kiwi_LVM_ in kiwi/boot/arch/*/oemboot/suse-repart

schaefi:
Fixes #39

get rid of volume name limitations

* Refactor volume setup in image .profile
* Refactor use of volume profile information in boot code


isbm:
Just a nitpick. It is better to avoid permanent scope under condition, if it takes the whole function body. Therfore in the `xml_state.py` at this snippet:
```python
volume_type_list = []
systemdisk_section = self.get_build_type_system_disk_section()
if systemdisk_section:
    ...
    # All code here
```
...I would rather do this:
```python
volume_type_list = []
systemdisk_section = self.get_build_type_system_disk_section()
if not systemdisk_section:
    return ...
# All code here
```

Issue #106, opened 2016-07-04
Remove dependency on genisoimage


plater:
## Steps to reproduce the behaviour

plater:
Somehow kiwi has an entwined dependency on a package that is dead in the water and it results in this package being installed by default in every new openSUSE installation regardless of whether kiwi is installed.

schaefi:
I'm sorry I can't see where kiwi depends on cdrkit-cdrtools-compat, looked up package/python3-kiwi-spec-template and did not find a requires. Could you provide more details ?

Thanks

plater:
I want to replace the two lines in the openSUSE package : 
BuildRequires:  cdrkit-cdrtools-compat
BuildRequires:  genisoimage
with BuildRequires:  cdrtools to remove the dependency on the wodim package.
This will enable me to replace the dependency on wodim from patterns-openSUSE with cdrtools and stop wodim from being installed by default on new systems.

plater:
I've had a chance to look at python3-kiwi-spec-template, genisoimage was made as a drop in replacement for mkisofs but has been unmaintained for years. What I need to know is if it is still replaceable. I don't want to break Kiwi.

schaefi:
ok, so the never ending story about this topic. It seemed to be so easy but it isn't. For the kiwi v7 version which you referred to with ```Requires: cdrkit-cdrtools-compat``` see the following thread where we already tried to get rid of it:

* https://github.com/openSUSE/kiwi/pull/509

I can't tell you the exact reasons but it has something to do with the fact that the kiwi v7 version supports old (sle11) and new (up to SLE12, Leap42) distributions. I'm pretty sure people realized that changing the working version will cause some problems down the decade of distributions we support with kiwi v7. Therefore my recommendation; don't touch it :)

The next generation of kiwi (kiwi v8) which should be the reference for issues reported here, supports distributions from SLE12, Leap42 or newer. Therefore I don't had to care for anything older and that made the requirements a little bit easier. However, I choose ```Requires: genisoimage```
for the following reason:

```
commit 47e83f7dcfbe36cde3df3132a723dc5e425fedb2
Author: Marcus Schäfer <ms@suse.com>
Date:   Tue Apr 19 10:09:20 2016 +0200

    Added genisoimage requirement to kiwi spec file
    
    On SLE12 we only have genisoimage, on Leap and Tumbleweed we
    have both genisoimage and mkisofs. For the moment both toolkits
    produces working iso images with kiwi, thus I chose the save
[...]

PR #108, opened 2016-07-06
Lookup iso creation tool

schaefi:
Related to #106

__dynamically lookup for an iso creation tool__

Instead of strictly using genisoimage allow for using
either the community variant or mkisofs as provided by
J.Schilling

### DIFF
--- a/kiwi/filesystem/isofs.py
+++ b/kiwi/filesystem/isofs.py
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
-import os
+from distutils.spawn import find_executable
 
 # project
 from .base import FileSystemBase
@@ -71,12 +71,12 @@ class FileSystemIsoFs(FileSystemBase):
         There are tools by J.Schilling and tools from the community
         Depending on what is installed a decision needs to be made
         """
-        iso_creation_tools = [
-            '/usr/bin/mkisofs', '/usr/bin/genisoimage'
-        ]
+        iso_creation_tools = ['mkisofs', 'genisoimage']
         for tool in iso_creation_tools:
-            if os.path.exists(tool):
-                return tool
+            tool_found = find_executable(tool)
+            if tool_found:
+                return tool_found
+
         raise KiwiIsoToolError(
             'No iso creation tool found, searched for: %s' %
             iso_creation_tools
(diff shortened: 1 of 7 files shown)


kiwi/filesystem/isofs.py
```diff
@@ -61,3 +67,19 @@ def create_on_file(self, filename, label=None, exclude=None):
...
+        iso_creation_tools = [
+            '/usr/bin/mkisofs', '/usr/bin/genisoimage'
+        ]
+        for tool in iso_creation_tools:
+            if os.path.exists(tool):
+                return tool
+        raise KiwiIsoToolError(
+            'No iso creation tool found, searched for: %s' %
+            iso_creation_tools
+        )
```
isbm: Umhmh... While it is probably always in `/usr/bin`, I wouldn't do that. Instead, I would look for `PATH` variable and then look if the _executable_ is there. Something (:boom: **Untested!**) like:
```python
def _find_isomake(self):
    binaries = ["mkisofs", "genisoimage"]
    paths = os.environ.get("PATH").split(os.pathsep)
    for executable in binaries:
        for path in paths:
            location = os.path.join(path, executable)
            if os.path.exists(location):
                return location
    raise KiwiIsoToolError("Karamba! :-(")
```

schaefi: yes actually it would be against FHS if those tools are not in /usr/bin, but I agree a path search is the way to go. I'm going to change it

kiwi/filesystem/isofs.py
```diff
@@ -15,7 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
-import os
+from distutils.spawn import find_executable
```
isbm: Uhm, don't. :laughing:  Distutils aren't always installed. If you _package_ Kiwi, you might find out that this will fail "module not found".

schaefi: Hmm, I have checked python-base and python3-base down to SLE11 SP4 and distutils.spawn is part of the core python. In other words from 2.6 to 3.5 it is available. That imho spans a big enough range.

schaefi: The used find_executable method seems to be present for the core distutils, I also see extra distutils packages but we don't need this functionality

isbm: Then make sure it is everywhere 100% on each distro Kiwi is ran. :wink:

schaefi: ok, I compared the python packaging in RH and others and you are right there are big differences. I thought there is a global definition what is covered by python core but it seems every distro packages it differently. Looks like we need our own method to do the search as you suggested

PR #109, opened 2016-07-07
Fix VMWare disk tagging

bear454:
Fixes [BSC#988086](https://bugzilla.suse.com/show_bug.cgi?id=988086), [BSC#988087](https://bugzilla.suse.com/show_bug.cgi?id=988087)

The current implementation had a few issues resulting in nonfunctional VMDKs.

Changes proposed in this PR:

* strip null characters off the `dd` output.
* add the encoding prefix only if it's not already present


schaefi:
great, thanks much :+1:

PR #115, opened 2016-07-22
Archive tar extension

davidcassany:
**Changes proposed in this pull request:**

From one side it includes the tar version validation already posted in #99, just with some minor changes in order to test the version just once in constructor and not in every command run. It also includes some small extension to provide files appending features that might handy to support lxd container format as seen in #99. Finally a comment is included in class description in order to justify why kiwi is no using tarfile python module.


schaefi:
looks good to me :+1:

PR #118, opened 2016-07-26
Fix user schema layout

davidcassany:
__Changed user and group schema layout__

The user schema layout has been changed in order to make groups management easier and more flexible. The current schema is like the one below:

```
<users>
    <user group="root" secgroups="testgrp,anothergrp" password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root"/>
    <user group="tux" password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/home/tux" name="tux"/>
    <user group="users" password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/home/custom_dir" name="oracle"/>
</users>
```

The **group** attribute is mandatory and it is the primary group of the user. The **secgroups** attribute is optional and it is a comma separated value list of different secondary groups, if not present the user is no included in any secondary group.

In order to ensure the comma separated normalization for the secondary groups list, a pattern has been included in schema to restrict the naming possibilities of groups. Now any group name must be format according to the _POSIX portable filename character set_. Finally, due to this new schema layout the feature of setting the group ID has been dropped.

This pull request will require updating the the descriptions in [kiwi-descriptions](https://github.com/SUSE/kiwi-descriptions).

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -279,9 +279,9 @@ class SystemSetup(object):
                 if user.get_group() and not system_users.group_exists(user.get_group()):
                     log.info('Adding group %s', user.get_group())
                     system_users.group_add(user.get_group(), [])
-                if not user.get_sec_groups():
+                if not user.get_secgroups():
                     continue
-                for group in user.get_sec_groups():
+                for group in user.get_secgroups().split(','):
                     if not system_users.group_exists(group):
                         log.info('Adding group %s', group)
                         system_users.group_add(group, [])
@@ -302,7 +302,7 @@ class SystemSetup(object):
                 user_id = user.get_id()
                 user_realname = user.get_realname()
                 user_shell = user.get_shell()
-                user_sec_groups = user.get_sec_groups()
+                user_secgroups = user.get_secgroups()
                 user_group = user.get_group()
 
                 user_exists = system_users.user_exists(user_name)
@@ -319,9 +319,9 @@ class SystemSetup(object):
                 if user_group:
                     options.append('-g')
                     options.append(user_group)
-                if user_sec_groups and len(user_sec_groups) > 0:
+                if user_secgroups:
                     options.append('-G')
-                    options.append(','.join(user_sec_groups))
+                    options.append(user_secgroups)
                 if user_id:
                     options.append('-u')
                     options.append(user_id)
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -274,16 +274,17 @@ class SystemSetup(object):
         """
         system_users = Users(self.root_dir)
 
-        for users in self.xml_state.get_users():
-            if not system_users.group_exists(users.group_name):
-                options = []
-                if users.group_id:
-                    options.append('-g')
-                    options.append(users.group_id)
-                log.info('Adding group %s', users.group_name)
-                system_users.group_add(
-                    users.group_name, options
-                )
+        for users_set in self.xml_state.get_users():
+            for user in users_set:
+                if user.get_group() and not system_users.group_exists(user.get_group()):
+                    log.info('Adding group %s', user.get_group())
+                    system_users.group_add(user.get_group(), [])
+                if not user.get_sec_groups():
+                    continue
+                for group in user.get_sec_groups():
+                    if not system_users.group_exists(group):
+                        log.info('Adding group %s', group)
+                        system_users.group_add(group, [])
 
     def setup_users(self):
         """
@@ -291,8 +292,8 @@ class SystemSetup(object):
         """
         system_users = Users(self.root_dir)
 
-        for users in self.xml_state.get_users():
-            for user in users.user_sections:
+        for users_set in self.xml_state.get_users():
+            for user in users_set:
                 log.info('Setting up user %s', user.get_name())
                 password = user.get_password()
                 password_format = user.get_pwdformat()
@@ -301,6 +302,8 @@ class SystemSetup(object):
                 user_id = user.get_id()
                 user_realname = user.get_realname()
                 user_shell = user.get_shell()
+                user_sec_groups = user.get_sec_groups()
+                user_group = user.get_group()
 
                 user_exists = system_users.user_exists(user_name)
 
@@ -313,12 +316,12 @@ class SystemSetup(object):
                 if user_shell:
                     options.append('-s')
                     options.append(user_shell)
-                if users.group_id or users.group_name:
+                if user_group:
                     options.append('-g')
-                if users.group_id:
-                    options.append(users.group_id)
-                else:
-                    options.append(users.group_name)
+                    options.append(user_group)
+                if user_sec_groups and len(user_sec_groups) > 0:
+                    options.append('-G')
+                    options.append(','.join(user_sec_groups))
                 if user_id:
                     options.append('-u')
                     options.append(user_id)
@@ -333,13 +336,15 @@ class SystemSetup(object):
                 if user_exists:
                     log.info(
                         '--> Modifying user: %s [%s]',
-                        user_name, users.group_name
+                        user_name,
+                        user_group
                     )
                     system_users.user_modify(user_name, options)
                 else:
                     log.info(
                         '--> Adding user: %s [%s]',
-                        user_name, users.group_name
+                        user_name,
+                        user_group
                     )
                     system_users.user_add(user_name, options)
                     if home_path:
@@ -347,7 +352,9 @@ class SystemSetup(object):
                             '--> Setting permissions for %s', home_path
                         )
                         system_users.setup_home_for_user(
-                            user_name, users.group_name, home_path
+                            user_name,
+                            user_group,
+                            home_path
                         )
 
     def import_image_identifier(self):
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -274,14 +274,11 @@ class SystemSetup(object):
         """
         system_users = Users(self.root_dir)
 
-        for users_set in self.xml_state.get_users():
-            for user in users_set:
-                if user.get_groups():
-                    user_groups = user.get_groups().split(',')
-                    for group in user_groups:
-                        if not system_users.group_exists(group):
-                            log.info('Adding group %s', group)
-                            system_users.group_add(group, [])
+        for user in self.xml_state.get_users():
+            for group in self.xml_state.get_user_groups(user.get_name()):
+                if not system_users.group_exists(group):
+                    log.info('Adding group %s', group)
+                    system_users.group_add(group, [])
 
     def setup_users(self):
         """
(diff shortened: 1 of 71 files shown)


isbm:
Uhm... Too much ad-hoc non-obvious keywords and it is still cumbersome. How about:
```xml
<users default="users">
      <user password="...(cipher)..." home="/root" name="root"/>
      <user groups="users" password="...(cipher)..." home="/home/foo" name="foo"/>
      <user groups="users,tux,blah" password="...(cipher)..." home="/home/bar" name="bar"/>
</users>
```

That is, if you do not specify `groups`, it sets `groups="users"` as per `default` value. Otherwise, `default` is ignored.

Result:
1. Less keywords
2. Obvious use

schaefi:
good point

davidcassany:
@isbm Agree with your comment, however I could not find a way to clearly distinguish the primary group from the others. My first though was to do it as you suggest and assume the first group in the list (or the default if none) would be the primary one, but this is looked to me also non-obvious. Maybe is better to not clearly specify this groups nuance to the user, since having to distinguish between primary and secondary groups is a niche use case.

About the default group I would not assume any value, since it may differ in different distros and let _useradd_ do the job according to its environment configuration.

schaefi:
> About the default group I would not assume any value, since it may differ in different distros and let  > useradd do the job according to its environment configuration.

I agree a fixed value might not match other distros, thanks for pointing that out

With regards to primary and secondary group I think it would be pretty ok to document that in the schema. I agree with lsbm that we should not introduce different attributes for the group list.
One of the open issues is also to create good user documentation from the schema. Therefore I think it is quite ok to just have that part documented

schaefi:
What is missing here however, is the xslt style sheet which transfers the old schema to the new layout. That also implies a schema version change which is needed in kiwi.rnc

You can find the style sheets in kiwi/xsl/ and we need a __convert63to64.xsl__ along with the proposed changes.

Once the style sheet is in place you also don't have to manually touch any of the existing XML descriptions and just run:

```
$ make valid
```

This will auto move the XML descriptions to the new layout and is a good test if everything worked as we want it.

Once we agreed on the final changes in the schema I can create the xslt style sheet if you like ?

davidcassany:
@schaefi  agree, I am on it then, those are small changes.

About generating documentation from the schema I have been googling around to find an opensource tool to auto generate the documentation from the XSD, but I just found a couple of abandoned projects (xs3p and xsddoc)

schaefi:
yep I also was searching together with the doc team and we also did not find anything good, which was the reason why we opened an issue. So far all docs in kiwi using ReST so one idea was to extract the docs from the schema and create a ReST document which is then normally handled like all other docs using sphinx.

schaefi:
for I think it's ok to just have the information in the schema like it exists for other attributes and how we extract it for the docs is then another topic we can work on next

davidcassany:
@schaefi, user I think it a good idea you do the xslt style sheet if you wish, this is might be a good way to involve someone else in the review and testing of the schema changes.

schaefi:
yep I'm happy to contribute that part, we just need to agree to the final schema change

schaefi:
If we go with a configurable default group I vote for:

```xml
<users default_group="users">
      <user password="...(cipher)..." home="/root" name="root"/>
      <user groups="users" password="...(cipher)..." home="/home/foo" name="foo"/>
      <user groups="users,tux,blah" password="...(cipher)..." home="/home/bar" name="bar"/>
</users>
```

If we let the toolchain decide for the default group I vote for

```xml
<users>
      <user password="...(cipher)..." home="/root" name="root"/>
      <user groups="users" password="...(cipher)..." home="/home/foo" name="foo"/>
      <user groups="users,tux,blah" password="...(cipher)..." home="/home/bar" name="bar"/>
</users>
```

The race is open :smile:

davidcassany:
I like isbm suggestion, single groups attribute which would be optional. I prefer the second option :smile: 

Then in would lead to a useradd command as it follows:

`useradd -p whatever -s /bin/bash -m -d /home/tux tux`

Without any _-g group_ option. According to the useradd man pages this will be the default group:
> If not specified, the behavior of useradd will depend on the USERGROUPS_ENAB variable in /etc/login.defs. If this variable is set to yes (or -U/--user-group is specified on the command line), a group will be created for the user, with the same name as her loginname. If the variable is set to no (or -N/--no-user-group is specified on the command line), useradd will set the primary group of the new user to the value specified by the GROUP variable in /etc/default/useradd, or 100 by default.>

schaefi:
Fine with me :+1:

isbm:
In any case this sucks: `group="root" secgroups="testgrp,anothergrp"`, frankly speaking. :wink:

davidcassany:
Yep I know :sweat_smile: Thanks for having a look at it :wink:

davidcassany:
@schaefi I have updated the schema according our latest discussion, so now will you contribute to the xslt stylesheet? I have not done anything related to this update.

schaefi:
yep I will work on the xslt, this will not conflict with your work

I'll start on that tomorrow hopefully

schaefi:
Added style sheet for auto conversion from schema 6.3 to 6.4

kiwi/xml_parse.py
```diff
@@ -3553,18 +3540,12 @@ def build(self, node):
...
-        if value is not None and 'secgroups' not in already_processed:
-            already_processed.add('secgroups')
-            self.secgroups = value
-            self.secgroups = ' '.join(self.secgroups.split())
-            self.validate_secgroups_list(self.secgroups)    # validate type secgroups-list
+        value = find_attr_value_('groups', node)
+        if value is not None and 'groups' not in already_processed:
+            already_processed.add('groups')
+            self.groups = value
+            self.groups = ' '.join(self.groups.split())
```
isbm: Why not just below instead?
```python
self.groups = ' '.join(value.split())
```

kiwi/system/setup.py
```diff
@@ -276,15 +276,12 @@ def setup_groups(self):
...
-                    log.info('Adding group %s', user.get_group())
-                    system_users.group_add(user.get_group(), [])
-                if not user.get_secgroups():
-                    continue
-                for group in user.get_secgroups().split(','):
-                    if not system_users.group_exists(group):
-                        log.info('Adding group %s', group)
-                        system_users.group_add(group, [])
+                if user.get_groups():
+                    user_groups = user.get_groups().split(',')
```
isbm: This is something essentially wrong. I would definitely expect that `user.get_groups()` _already_ splitted AND verified.

kiwi/system/setup.py
```diff
@@ -276,15 +276,12 @@ def setup_groups(self):
...
+                    for group in user_groups:
+                        if not system_users.group_exists(group):
```
isbm: And here is where it should be like `group.exists()` instead.

kiwi/system/setup.py
```diff
@@ -302,8 +299,10 @@ def setup_users(self):
                 user_id = user.get_id()
                 user_realname = user.get_realname()
                 user_shell = user.get_shell()
-                user_secgroups = user.get_secgroups()
-                user_group = user.get_group()
+                if user.get_groups():
+                    user_groups = user.get_groups().split(',')
+                else:
+                    user_groups = None
```
isbm: This can be oneliner:
```python
user_groups = user.get_groups().split(",") if user.get_groups() else None
```
But as I already said, this code even should not exist. You are splitting groups all around the place. Instead, the `user.get_groups()` should either return `None` or a list of groups, preferably objects (see above the case what for).

kiwi/xml_parse.py
```diff
@@ -3459,10 +3458,8 @@ def factory(*args_, **kwargs_):
         else:
             return user(*args_, **kwargs_)
     factory = staticmethod(factory)
-    def get_group(self): return self.group
-    def set_group(self, group): self.group = group
-    def get_secgroups(self): return self.secgroups
-    def set_secgroups(self, secgroups): self.secgroups = secgroups
+    def get_groups(self): return self.groups
+    def set_groups(self, groups): self.groups = groups
```
isbm: This is unacceptably non-pythonic and is just horrid. Please use [property decorator](http://www.programiz.com/python-programming/property), instead.

kiwi/xml_parse.py
```diff
@@ -3516,12 +3506,9 @@ def export(self, outfile, level, namespace_='', name_='user', namespacedef_='',
         else:
             outfile.write('/>%s' % (eol_, ))
     def exportAttributes(self, outfile, level, already_processed, namespace_='', name_='user'):
-        if self.group is not None and 'group' not in already_processed:
-            already_processed.add('group')
-            outfile.write(' group=%s' % (quote_attrib(self.group), ))
-        if self.secgroups is not None and 'secgroups' not in already_processed:
-            already_processed.add('secgroups')
-            outfile.write(' secgroups=%s' % (quote_attrib(self.secgroups), ))
+        if self.groups is not None and 'groups' not in already_processed:
```
isbm: Again, if your `user.get_groups()` would return a list or None, the first check would not be needed. Note, here is a bug: if `self.groups` is an empty string or an empty list, you are in a positive state!

kiwi/xml_parse.py
```diff
@@ -3553,18 +3540,12 @@ def build(self, node):
...
-            self.group = ' '.join(self.group.split())
-            self.validate_group_name(self.group)    # validate type group-name
-        value = find_attr_value_('secgroups', node)
-        if value is not None and 'secgroups' not in already_processed:
-            already_processed.add('secgroups')
-            self.secgroups = value
-            self.secgroups = ' '.join(self.secgroups.split())
-            self.validate_secgroups_list(self.secgroups)    # validate type secgroups-list
+        value = find_attr_value_('groups', node)
+        if value is not None and 'groups' not in already_processed:
```
isbm: So it is OK if `value` equals to `""`?

kiwi/xml_parse.py
schaefi: xml_parse.py is autogenerated code from generateDS.py version 2.22a

schaefi:
@isbm I agree to the review on xml_parse.py, but as I said this is not manually written code, thus we can't do much here, except for contacting the author of the generateDS module. From a functional point of view the generated data structures matches the schema and it works pretty well. Thus I'm ok with the "non standard code" it produces

isbm:
@schaefi Ah, I forget that part it is auto-baked code. :sob: Then ignore me.

davidcassany:
@isbm I tried to find a way to get users directly splitted from xml_parse, but I think the generator can only handle space separated lists. I dislike space separated lists `groups="root users blah"` (to my eyes this is less obvious and error prone), so that I didn't even consider that option, however we might consider it, even I don't see a clear benefit since xml_parse will still be some sort of _ugly_ code.

isbm:
@davidcassany @schaefi However, with the auto-dumped code I always do subclassing from PyQT times. That is, you don't _modify_ the autogenerated code, but you _override_ it. And you do not use autogenerated code directly. Could be an option in order to keep your code clean.

isbm:
An example:
```python
class MyUglyAutogeneratedDump:
    def get_users(self):
        return "only space broken stuff"

class MyCleanStuff(object, MyUglyAutogeneratedDump):
    def get_users(self):
        return MyUglyAutogeneratedDump.get_users(self).split(" ")

class MyUsage(object):
    def main(self):
        clean = MyCleanStuff()
        nice_list = clean.get_users()
```

Something like that.

kiwi/system/setup.py
davidcassany: @isbm I tend not to use long oneliners, since IMHO they tend to be difficult to read. However you and @schaefi have a higher experience in python than me, so I am totally fine using such a ternary operator, do we have already an style criteria on how to code this situations?

Just for me to know for the future  and be able to code in a coherent and consistent style.

isbm: This "hard to read" is more than excuse everywhere. It is not about reading the code, but _understanding_ it. In both cases, if you do:
```python
if foo is not None:
   result = foo[0]
else:
   result = foo
```
...or:
```python
result = foo[0] if foo is not None else foo
```

In both cases I would still question why you would wrote that bug? :wink: Because in both cases you can read, that `foo` can be also `[]`.

davidcassany: Good point ;) that's why I included the groups_list type in the schema with the appropriate pattern, so I does not allow an empty string thus split won't return an empty list.

However, once again, the good option would be to directly get a validated list from xml_parse, as you said.

davidcassany:
@schaefi I have just added the method to get user groups directly as a merged list in xml_state and modified the users list in order they also get merged when multiple <users> sections are present. I also added an example configuration file to test the code with multiple <users> sections.

davidcassany:
Reopening, I missed clicked "Close and comment"

kiwi/xml_state.py
```diff
@@ -522,33 +522,52 @@ def get_users(self):
...
-                    users_by_group.append(
-                        users_by_group_type(
-                            group_name=users.get_group(),
-                            group_id=users.get_id(),
-                            user_sections=user_sections
-                        )
-                    )
-        return users_by_group
+        if len(users_sections):
+            for users_section in users_sections:
```
schaefi: you can rely on get_users_sections to return a list, so this can be simplified to

```python
for users_section in self.get_users_sections():
     ...
```

kiwi/xml_state.py
```diff
@@ -522,33 +522,52 @@ def get_users(self):
...
+                for user in users_section.get_user():
+                    match = [i for i, u in enumerate(users_list) if u.get_name() == user.get_name()]
```
schaefi: could you put that into a private method so that we have a name, e.g

```python
def _lookup_user_index(self, users_list, user):
    return [
        user_index for user_index, current_user in enumerate(users_list) if
        current_user.get_name() == user.get_name()
    ]
```

kiwi/xml_state.py
```diff
@@ -522,33 +522,52 @@ def get_users(self):
...
+                    for idx in match:
```
schaefi:
```python
for user_index in self._lookup_user_index(users_list, user):
```

kiwi/xml_state.py
```diff
@@ -522,33 +522,52 @@ def get_users(self):
...
+        for user in self.get_users():
+            if user.get_name() == user_name and user.get_groups():
+                seen = set()
+                groups_list = [
+                    grp for grp in user.get_groups().split(',')
+                    if not (grp in seen or seen.add(grp))
+                ]
+                break
+
+        return groups_list
```
schaefi: You can also write

```python
groups_list = []
for user in self.get_users():
    if user.get_name() == user_name and user.get_groups():
        groups_list += user.get_groups().split(',')

return list(set(groups_list))
```

kiwi/xml_state.py
```diff
@@ -522,33 +522,52 @@ def get_users(self):
...
+            for users_section in users_sections:
+                for user in users_section.get_user():
+                    match = [i for i, u in enumerate(users_list) if u.get_name() == user.get_name()]
+                    for idx in match:
+                        if users_list[idx].get_groups() and user.get_groups():
+                            users_list[idx].set_groups(
+                                ','.join([users_list[idx].get_groups(), user.get_groups()])
+                            )
+                        elif not users_list[idx].get_groups() and user.get_groups():
+                            users_list[idx].set_groups(user.get_groups())
```
schaefi: I'm a little bit concerned that get_users changes the groups attribute, why is this needed ?

schaefi: I committed an alternative way, feel free to revert if you see issues

kiwi/xml_state.py
davidcassany: The problem here is that if we assume to get a list of unique users from `self.get_users()` then we need to have the groups appended in advance (that the motivation to modify groups attribute inside get_users).

Also using `set(groups_list)` does not keep the order groups, so we have no clear criteria on which group is going to be to login group.

Maybe we can assume self.get_users() to return more than one instance of the same user, that would lead to a  user modification so the user would be configured as it is in the last appearance. That could simplify this code.

kiwi/xml_state.py
davidcassany: Yep, that looks good

davidcassany:
@schaefi That looks much better

kiwi/xml_state.py
davidcassany: @schaefi Your last code suggestion is better, not use get_users inside get_user_groups :wink:

PR #119, opened 2016-08-05
Detailed schema validation report

davidcassany:
This pull request adds the execution of a **jing** process if the validation
fails, that way the user has more information about what is
actually failing during the validation process. If **jing** command
is not found, it just produces an info message to warn the user
and kiwi finalizes as expected.

### DIFF
--- a/kiwi/command.py
+++ b/kiwi/command.py
@@ -65,9 +65,9 @@ class Command(object):
         if custom_env:
             environment = custom_env
         if not Path.which(
-                command[0],
-                custom_env=environment,
-                access_mode=os.X_OK
+            command[0],
+            custom_env=environment,
+            access_mode=os.X_OK
         ):
             raise KiwiCommandNotFound(
                 'Command %s not found in the environment' % command[0]
@@ -131,9 +131,9 @@ class Command(object):
         if custom_env:
             environment = custom_env
         if not Path.which(
-                command[0],
-                custom_env=environment,
-                access_mode=os.X_OK
+            command[0],
+            custom_env=environment,
+            access_mode=os.X_OK
         ):
             raise KiwiCommandNotFound(
                 'Command %s not found in the environment' % command[0]
--- a/kiwi/command.py
+++ b/kiwi/command.py
@@ -26,7 +26,10 @@ from collections import namedtuple
 from builtins import bytes
 
 # project
-from .exceptions import KiwiCommandError, KiwiCommandNotFound
+from .exceptions import (
+    KiwiCommandError,
+    KiwiCommandNotFound
+)
 
 
 class Command(object):
@@ -56,13 +59,18 @@ class Command(object):
         :rtype: tuple
         """
         from .logger import log
+        from .path import Path
         log.debug('EXEC: [%s]', ' '.join(command))
         environment = os.environ
         if custom_env:
             environment = custom_env
-        if not self._command_exists(command[0], environment):
+        if not Path.which(
+                command[0],
+                custom_env=environment,
+                access_mode=os.X_OK
+        ):
             raise KiwiCommandNotFound(
-                'Command {} not found in the environment'.format(command[0])
+                'Command %s not found in the environment' % command[0]
             )
         try:
             process = subprocess.Popen(
@@ -117,13 +125,18 @@ class Command(object):
         :rtype: tuple
         """
         from .logger import log
+        from .path import Path
         log.debug('EXEC: [%s]', ' '.join(command))
         environment = os.environ
         if custom_env:
             environment = custom_env
-        if not self._command_exists(command[0], environment):
+        if not Path.which(
+                command[0],
+                custom_env=environment,
+                access_mode=os.X_OK
+        ):
             raise KiwiCommandNotFound(
-                'Command {} not found in the environment'.format(command[0])
+                'Command %s not found in the environment' % command[0]
             )
         try:
             process = subprocess.Popen(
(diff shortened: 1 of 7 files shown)


kiwi/xml_description.py
```diff
@@ -108,6 +112,25 @@ def load(self):
...
+    def _get_validation_details(self):
+        """
+        Run jing process to validate description against the schema
+
+        Jing command provides detailed feedback in case of schema
+        validation failure
+        """
+        try:
+            cmd = Command.run(
+                ['jing', Defaults.get_schema_file(), self.description_xslt_processed.name],
```
schaefi: I think we should have test if jing is installed prior to calling it. We can't require jing in the spec of the kiwi package because it will pull in the entire java stack. Thus an information to the user that kiwi would provide better error messages if jing is installed and from where it can be installed would be good

davidcassany: Yep, I though it was already done by Command, as I got a FileNotFoundError message while trying silly commands. But yep, now I checked and it is just the forwarded exception message from subprocess. Wouldn't it make sense to test if the command is available in Command.run and raise KiwiCommandNotFoundError (subclass of KiwiCommandError) exception or something like? This way we would just have to distinguish the given exception from Command and we would not modify the behavior of current KiwiCommandError try/catch bocks.

What you think?

schaefi: sounds good to me, you can raise a KiwiCommandNotFound exception from the Command class and catch that in your code. so all others benefit too, that's nice

kiwi/command.py
```diff
@@ -26,7 +26,7 @@
 from builtins import bytes
 
 # project
-from .exceptions import KiwiCommandError
+from .exceptions import KiwiCommandError, KiwiCommandNotFound
```
schaefi: This should be in the form

```
from .exceptions import (
    KiwiCommandError,
    KiwiCommandNotFound
)
```

just to be consistent with the rest of the code

davidcassany: Right, I will update it

kiwi/command.py
```diff
@@ -165,3 +173,24 @@ def _select():
...
+
+        :return: True if the command is found, False otherwise
+        :rtype: bool
+        """
+        if 'PATH' not in env:
+            raise KiwiCommandError('No PATH variable in environment')
+        return any(
+            os.access(os.path.join(path, command), os.X_OK)
+            for path in env['PATH'].split(os.pathsep)
+        )
```
schaefi: kiwi has an implementation of which in Path. I think you don't need a _command_exists method

```
from .path import Path

if Path.which(command):
    ...
```

kiwi/command.py
```diff
@@ -60,6 +60,10 @@ def run(self, command, custom_env=None, raise_on_error=True):
         environment = os.environ
         if custom_env:
             environment = custom_env
+        if not self._command_exists(command[0], environment):
+            raise KiwiCommandNotFound(
+                'Command {} not found in the environment'.format(command[0])
```
schaefi: I've used python's string format, but that's personal taste

```python
command_name = command[0]
if not Path.which(command_name):
    raise KiwiCommandNotFound(
        'Command %s not found in environment' % command_name
    )
```
```

File: kiwi/command.py (hunk 3 above)
Comment (schaefi) [inline kiwi/command.py]: The X_OK is however not checked with which, but I think it would make sense to have these kind of checks in Path not here

Comment (davidcassany) [inline kiwi/command.py]: I did not see that which implementation, however I am not sure to use that Path.which method here, at least as it is now. I think the environment management is done in a different way in command than how it is done in  Path.which. I am not sure how the custom env you can give to a Command.run is actually used, but the command is not checked against the host environment if a custom one is provided. However, the Path.which, always includes the os.environ in the search. Is that the case or I am missing something?

If so I would suggest to slightly modify Path.which, even I am not entirely sure how to do that without altering the actual behavior of which and without adding non obvious extra parameters

Comment (schaefi) [inline kiwi/command.py]: right, so I suggest to exend Path.which by a custom_env and a flags parameter. If custom_env is provided it is used instead of os.environ. Does that makes sense ?

Comment (davidcassany) [inline kiwi/command.py]: Yes I think so. So, I will include X_OK in which and add also a custom_env with a default None. I haven't understood that _flags parameter_ thing... :sweat_smile:  Do we need extra flags?

File: kiwi/command.py (hunk 4 above)
Comment (davidcassany) [inline kiwi/command.py]: Sure I will use that to make the code consistent in style.

File: kiwi/command.py (hunk 3 above)
Comment (schaefi) [inline kiwi/command.py]: I meant checking if the command is executable needs an extra "flag" parameter, maybe file_attributes matches it better

Comment (davidcassany) [inline kiwi/command.py]: Oh I see. Yes, checking if it is executable or not needs to be in a flag.

File: kiwi/command.py [hunk 5]
```diff
@@ -56,10 +59,19 @@ def run(self, command, custom_env=None, raise_on_error=True):
         :rtype: tuple
         """
         from .logger import log
+        from .path import Path
         log.debug('EXEC: [%s]', ' '.join(command))
         environment = os.environ
         if custom_env:
             environment = custom_env
+        if not Path.which(
+                command[0],
```
schaefi: nit pick, 4 spaces are enough to indent that block :)

schaefi:
looks great to me now, just the small indentation thing and we are done

kiwi/command.py
davidcassany: Yes right. I missed copied and pasted that part :stuck_out_tongue:

PR #123, opened 2016-08-18
Wait for storage device patch

davidcassany:
Fixes bugzilla issue 992992

* Added getDeviceTransportType method in functions.sh
* Modified waitForStorageDevice method. Now it reacts depending on the device type.

### DIFF
--- a/kiwi/boot/functions.sh
+++ b/kiwi/boot/functions.sh
@@ -5546,7 +5546,7 @@ function waitForStorageDevice {
     local IFS=$IFS_ORIG
     local device=$1
     local check=0
-    local transport=`getDeviceTransportType $device`
+    local transport="`getDeviceTransportType $device`"
     if [[ $transport == *"usb"* ]];then
         limit=2
     else
--- a/kiwi/boot/functions.sh
+++ b/kiwi/boot/functions.sh
@@ -5539,9 +5539,8 @@ function waitForStorageDevice {
     # function to check access on a storage device
     # which could be a whole disk or a partition.
     # the function will wait until the size of the
-    # storage device could be obtained or the check
-    # counter equals 30. For USB devices it waits
-    # only until check counter equals 2.
+    # storage device could be obtained or the timeout
+    # is reached.
     # ----
     local IFS=$IFS_ORIG
     local device=$1
@@ -5549,9 +5548,11 @@ function waitForStorageDevice {
     local transport=$(getDeviceTransportType $device)
     case "$transport" in
 	"usb")
+	    # Counter limit=2 means 4 seconds timeout
 	    limit=2
 	;;
 	*)
+	    # Counter limit=30 means 60 seconds timeout
 	    limit=30
 	;;
     esac
--- a/kiwi/boot/functions.sh
+++ b/kiwi/boot/functions.sh
@@ -5517,22 +5517,19 @@ function killShell {
     fi
 }
 #======================================
-# isUSBDevice
+# getDeviceTransportType
 #--------------------------------------
-function isUSBDevice {
+function getDeviceTransportType {
     # /.../
-    # function to check if a device is a USB connected
-    # device or not. It uses lsblk command.
+    # function to get the transport type of the
+    # given device. It uses lsblk command.
     # ----
     local device=$1
     if [ -z "$device" ] || [ ! -e "$device" ];then
-        echo 1 ; return 1; 
+        echo 1 ; return 1;
     fi
     local disk=${device%%[0-9]*}
-    ret=`lsblk -dno tran $disk`
-    if [ "$ret" != "usb" ];then
-        echo 1; return 1;
-    fi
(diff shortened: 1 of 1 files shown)


kiwi/boot/functions.sh
```diff
@@ -5517,6 +5517,21 @@ function killShell {
...
+    # /.../
+    # function to get the transport type of the
+    # given device. It uses lsblk command.
+    # ----
+    local device=$1
+    if [ -z "$device" ] || [ ! -e "$device" ];then
+        echo 1 ; return 1;
+    fi
+    local disk=${device%%[0-9]*}
+    echo "`lsblk -dno tran $disk`"
```
schaefi: could you use the $() syntax

```bash
echo $(lsblk -dno tran $disk)
```

I switched to this invocation type at some point in the past, there are still some backticks call left but I'd like to get rid of them. Thanks

kiwi/boot/functions.sh
```diff
@@ -5525,18 +5540,25 @@ function waitForStorageDevice {
...
     # storage device could be obtained or the check
-    # counter equals 4
+    # counter equals 30. For USB devices it waits
+    # only until check counter equals 2.
     # ----
     local IFS=$IFS_ORIG
     local device=$1
     local check=0
+    local transport="`getDeviceTransportType $device`"
+    if [[ $transport == *"usb"* ]];then
```
schaefi: I'm not sure how many usb transport type exists, but is it a safe regular expression to match any string that has the three letters 'usb' in its name ? Just asking, maybe a simple string match works for usb devices ?

kiwi/boot/functions.sh
davidcassany: Sure $() looks nicer :smile:

kiwi/boot/functions.sh
davidcassany: I actually included the regular expression because I could not find a list of possible device transports. So I could not see anywhere the possible values I can get out of it and I though that maybe could be enough if the transport contains usb, however it is true that being more restrictive we are being also safer. So thinking it twice I  think you got a point on that, I prefer just a simple match, better to experience a false negative than a false positive in this case.

kiwi/boot/functions.sh
isbm: @davidcassany Graveticks are deprecated and is limited and the `$()` is new. Whilst graveticks can be nested with a back-slash form, multi-nesting gets very bad and overall limited. The new `$()` form can be easily nested and can also contain a script. You can also put an arithmetic expansions, several sub-shells, complex relational expressions, like:
```bash
while [ $(( (($x + $y)/($a * $b)) < ($foo*$bar) )) -ne 0 ]
```
...and so on. Graveticks is an old thing that basically not used nowadays anymore.

davidcassany: @isbm that's cool, I was not aware of the possibility to use arithmetic in such a way :+1:  I though it was mostly a matter of getting rid of graveticks mixed with simple or double quotes, which can be tricky.

thanks :wink:

kiwi/boot/functions.sh
schaefi: ok so let's change this to a simple 

```bash
if [ "$transport" = "usb" ];then
```

in the worst case it will result in the default wait period which is ok for me. But if we wait too short for a device it would have a bad impact

schaefi: you could also consider to code a case statement even though we will match only one case it might be nicer to maintain when more transport types follows

```bash

case "$transport" in
    "usb")
        # handle usb transport type
    ;;
    *)
        # handle default transport
    ;;
    esac
```

I leave that up to you

isbm: @davidcassany @schaefi And you can also use a regexp:
```bash
if [[ $transport =~ "usb" ]]; then
...
fi
```

P.S. Please keep a whitespace after `;` and before `then` :wink:

kiwi/boot/functions.sh
```diff
@@ -5525,18 +5540,28 @@ function waitForStorageDevice {
...
     local device=$1
     local check=0
+    local transport=$(getDeviceTransportType $device)
+    case "$transport" in
+	"usb")
+	    limit=2
+	;;
+	*)
+	    limit=30
+	;;
```
schaefi: nit picking sorry, could you move the comment information from the function body into the corresponding case blocks so that we know limit=2 means a 4 sec wait period, limit=30 means a 60sec wait period

davidcassany: @schaefi  Not sure I got it, you mean something like: 

```
    case "$transport" in
        "usb")
            # Counter limit=2 means 4 seconds timeout
            limit=2
        ;;
        *)
            # Counter limit=30 means 60 seconds timeout
            limit=30
        ;;
    esac
````

schaefi:
great, let's merge this :+1:

PR #125, opened 2016-08-19
Fixes #124

### LINKED ISSUE #124: Setting the user ID fails
## Problem description
Inclucing a user id attribute in `<user>` section causes a KiwiCommandError exception.  For instance, a user definition like: 

`<user name="vagrant" password="vagrant" pwdformat="plain" home="/home/vagrant" id="1000"/>`

In kiwi/system/setup.py the user id given by user.get_id() if not None should be converted into string, as options in users.py are expected to be a string list.

davidcassany:
Fixes #124

user.get_id() returns int type and string type is required in the options list the finally is handled by Command.

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -316,7 +316,7 @@ class SystemSetup(object):
                     options.append(','.join(user_groups[1:]))
             if user_id:
                 options.append('-u')
-                options.append(user_id)
+                options.append(str(user_id))
             if user_realname:
                 options.append('-c')
                 options.append(user_realname)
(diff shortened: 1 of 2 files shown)


kiwi/system/setup.py
```diff
@@ -316,7 +316,7 @@ def setup_users(self):
                     options.append(','.join(user_groups[1:]))
             if user_id:
                 options.append('-u')
-                options.append(user_id)
+                options.append(str(user_id))
```
schaefi: please use format() instead of str()

schaefi:
:+1:

PR #126, opened 2016-08-23
travis: Next try to fix automatic pypi deployment

toabctl:
Also the travis command which added the pypi deployment part formatted
the yaml file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,30 @@
 language: python
-
 branches:
-    except:
-        - gh-pages
-
+  except:
+  - gh-pages
 matrix:
-    include:
-        - python: 3.4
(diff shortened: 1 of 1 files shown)


toabctl:
@schaefi I recognized that you are still uploading the tarballs manually. So I used now the ```travis``` CLI to create the config. This worked at least for other projects for me. So hopefully this PR fixes the problem now...

schaefi:
Thanks much, let's give it a try. However I'm still concerned with regards to this issue
https://github.com/travis-ci/travis-ci/issues/1675

I followed the suggestions there and it never worked, let's hope this one fixes it even though I don't understand why it should make a difference.

I'll just try

PR #127, opened 2016-08-23
Add evaluation of machine's vmconfig_entry

### LINKED ISSUE #122: preferences/type/machine/vmconfig-entry values are ignored
In Kiwi8, it appears that the vmconfig-entries are ignored.

The following works in Kiwi7:
`			<machine memory="8192" guestOS="centos7-64" ovftype="vmware">
				<vmdisk id="0" controller="ide"/>
				<vmnic driver="e1000" interface="0" mode="bridged"/>

				
				<vmconfig-entry>numvcpus = "4"</vmconfig-entry>
				<vmconfig-entry>cpuid.coresPerSocket = "2"</vmconfig-entry>
			</machine>`

This results in the following in the output VMX file of Kiwi8:
`numvcpus = "4"
cpuid.coresPerSocket = "2"`

In Kiwi8, these values aren't present.

Mark Scott.

schaefi:
As part of the machine section it is possible to setup a custom
entry which is stored in the machine configuration file. The
evaluation of such an entry for the vmdk (.vmx) config file
was still missing. This Fixes #122

### DIFF
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -470,6 +470,18 @@ class XMLState(object):
             if vmdvd_sections:
                 return vmdvd_sections[0]
 
+    def get_build_type_vmconfig_entries(self):
+        """
+        List of vmconfig-entry section values from the first
+        machine section in the build type section
+
+        :return: vmconfig_entry values
+        :rtype: list
+        """
+        machine_section = self.get_build_type_machine_section()
+        if machine_section:
+            return machine_section.get_vmconfig_entry()
+
     def get_build_type_oemconfig_section(self):
         """
         First oemconfig section from the build type section
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -480,7 +480,11 @@ class XMLState(object):
         """
         machine_section = self.get_build_type_machine_section()
         if machine_section:
-            return machine_section.get_vmconfig_entry()
+            vmconfig_entries = machine_section.get_vmconfig_entry()
+            if vmconfig_entries:
+                return vmconfig_entries
+
+        return []
 
     def get_build_type_oemconfig_section(self):
         """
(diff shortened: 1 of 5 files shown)


kiwi/xml_state.py
```diff
@@ -470,6 +470,18 @@ def get_build_type_vmdvd_section(self):
             if vmdvd_sections:
                 return vmdvd_sections[0]
 
+    def get_build_type_vmconfig_entries(self):
```
davidcassany: @schaefi shouldn't it return an empty list in case there is no machine section defined? Then we are sure it is iterable.

schaefi: it does that the auto generated code from xml_parse::get_vmconfig_entry() returns an empty list. This is also proven by the test in test_get_build_type_vmconfig_entries

kiwi/xml_state.py
```diff
@@ -470,6 +470,18 @@ def get_build_type_vmdvd_section(self):
...
+        """
+        List of vmconfig-entry section values from the first
+        machine section in the build type section
+
+        :return: vmconfig_entry values
+        :rtype: list
+        """
+        machine_section = self.get_build_type_machine_section()
+        if machine_section:
+            return machine_section.get_vmconfig_entry()
```
isbm: This is auto-generated code, or? :wink: I forgot again. :laughing:

kiwi/xml_state.py
davidcassany: I think get_build_type_vmconfig_entries returns None if the whole <machine> section is missing, in that case xml_parse::get_vmconfig_entry() is not called and then in vmdk.py custom_entries is None. The current description files I am using to build terraform images fails with that code, as they don't have <machine> section.

kiwi/xml_state.py
davidcassany: @isbm No it is not, I think it should be something like:
```

if machine_section:
    return machine_section.get_vmconfig_entry()
return []
```

schaefi: ok good point, let me check that in a test

schaefi: you're right, thanks man

davidcassany: Welcome :+1:

PR #131, opened 2016-08-25
Adding newline separator in vmconfig entries

### LINKED ISSUE #130: Builds with vmconfig-entries produce invalid VMX files
It appears that newlines are not created in the VMX if you have more than one entry, which fails to parse properly by ovftool later, to create an OVF.

`#!/usr/bin/env vmware
# kiwi generated VMware settings file
config.version = "8"
tools.syncTime = "true"
uuid.action = "create"
virtualHW.version = "9"
displayName = "enterprise_lexington"
guestOS = "centos7-64"
priority.grabbed = "normal"
priority.ungrabbed = "normal"
powerType.powerOff = "soft"
powerType.powerOn  = "soft"
powerType.suspend  = "soft"
powerType.reset    = "soft"
memsize = "8192"
ide0:0.present = "true"
ide0:0.fileName= "/home/jenkins/jenkins_slave_builds_prod/workspace/enterprise_lexington_appliance/appliance/target/output/vmx/ovf/enterprise_lexington.x86_64-0.0.1.vmdk"
ide0:0.redo = ""
ethernet0.present = "true"
ethernet0.allow64bitVmxnet = "true"
ethernet0.addressType = "generated"
ethernet0.virtualDev = "e1000"
ethernet0.connectionType = "bridged"
usb.present = "true"
**numvcpus = "4"cpuid.coresPerSocket = "2"**`

davidcassany:
Fixes #130

This commit fixes #130, vmconfig custom entries were not separated
by newline character, so '\n' appended in each custom entry string.

### DIFF
--- a/kiwi/storage/subformat/vmdk.py
+++ b/kiwi/storage/subformat/vmdk.py
@@ -188,7 +188,7 @@ class DiskFormatVmdk(DiskFormatBase):
             with open(settings_file, 'w') as config:
                 config.write(settings_template.substitute(template_record))
                 for custom_entry in custom_entries:
-                    config.write(custom_entry + '\n')
+                    config.write(custom_entry + os.linesep)
         except Exception as e:
             raise KiwiTemplateError(
                 '%s: %s' % (type(e).__name__, format(e))
--- a/test/unit/storage_subformat_vmdk_test.py
+++ b/test/unit/storage_subformat_vmdk_test.py
@@ -2,6 +2,7 @@
 from mock import patch
 from mock import call
 import mock
+import os
 
 from .test_helper import *
 
@@ -308,10 +309,10 @@ class TestDiskFormatVmdk(object):
             self.vmdk_settings
         )
         assert self.file_mock.write.call_args_list[2] == call(
-            'custom entry 1\n'
+            'custom entry 1' + os.linesep
         )
         assert self.file_mock.write.call_args_list[3] == call(
-            'custom entry 2\n'
+            'custom entry 2' + os.linesep
         )
         assert self.file_mock.seek.call_args_list == [
             call(512, 0), call(0, 2)


kiwi/storage/subformat/vmdk.py
```diff
@@ -188,7 +188,7 @@ def _create_vmware_settings_file(self):
             with open(settings_file, 'w') as config:
                 config.write(settings_template.substitute(template_record))
                 for custom_entry in custom_entries:
-                    config.write(custom_entry)
+                    config.write(custom_entry + '\n')
```
isbm:
```python
import os
...
config.write(custom_entry + os.linesep)
```

test/unit/storage_subformat_vmdk_test.py
```diff
@@ -308,10 +308,10 @@ def side_effect(arg):
             self.vmdk_settings
         )
         assert self.file_mock.write.call_args_list[2] == call(
-            'custom entry 1'
+            'custom entry 1\n'
```
isbm: `os.linesep`

test/unit/storage_subformat_vmdk_test.py
```diff
@@ -308,10 +308,10 @@ def side_effect(arg):
...
         )
         assert self.file_mock.write.call_args_list[3] == call(
-            'custom entry 2'
+            'custom entry 2\n'
```
isbm: `os.linesep`

schaefi:
I'm sorry I did something stupid this morning, will resolve the conflict

kiwi/storage/subformat/vmdk.py
davidcassany: I think this is not safe, actually this was my first attempt, but after reading the docs and googling a little I switched to '\n'. `open('settings_file', 'w')` opens the file in text mode which, according to the docs, it is converted automatically to the platform newline character. This is some sort of write() black magic. Note that if using 'os.linesep' in windows it would fail, sinse '\r\n' would be translated to '\r\r\n'. From [here](https://docs.python.org/2/library/functions.html#open):

>The default is to use text mode, which may convert '\n' characters to a platform-specific representation on writing and back on reading.

I do not think we might having an issue because of that, however I think the right way to write a line is using '\n'. I would prefer os.linesep also, but looks like write() is doing some black magic behind the scenes :sweat:

PR #134, opened 2016-08-25
Activate partition using parted instead of sfdisk

### LINKED ISSUE #129: partition activation (sfdisk) fails with util-linux => 2.26
## Problem description

sfdisk changed '--activate<[=device]>' parameter to '--activate <dev>' in util-linux 2.26

[  188s] [ DEBUG   ]: 09:20:01 | EXEC: [sfdisk --activate=1 /dev/loop0]
[  188s] [ DEBUG   ]: 09:20:01 | EXEC: Failed with stderr: sfdisk: option '--activate' doesn't allow an argument

schaefi:
sfdisk changes their caller semantics incompatible. Therefore
we move to a tool which is still stable in the caller options
Fixes #129

### DIFF
--- a/kiwi/boot/functions.sh
+++ b/kiwi/boot/functions.sh
@@ -9322,7 +9322,7 @@ function activateBootPartition {
     fi
     local bootID=$(nd $device)
     local diskID=$(dn $device)
-    sfdisk $diskID --force -A $bootID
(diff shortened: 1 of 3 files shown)


schaefi:
real live test, showed good results

PR #135, opened 2016-08-25
Move to tar.gz as sdist target

### LINKED ISSUE #132: Use .tar.gz as the sdist format
It looks like you are using bztar for your sdist format; PyPI is planning to support only .tar.gz in the near future. See https://www.python.org/dev/peps/pep-0527/

schaefi:
PyPI is planning to support only .tar.gz in the near future.
See https://www.python.org/dev/peps/pep-0527. This Fixes #132


schaefi:
right but I like to see the configured target because I can't remember the default in a while :smiley:

PR #136, opened 2016-09-06
Hanlde rpm excludedocs entry

### LINKED ISSUE #133: rpm-excludedocs handling missing
## Problem description

The XML setup for rpm-excludedocs is not taken into account by the code

## Actual behaviour

package installation happens without docs excluded

davidcassany:
Fixes #133

This pull request only handles the rpm-excludedocs for zypper.

* Added get_rpm_excludedocs method in xml_state to get the value from parsed data.
* Added exclude_docs flag in RepositoryBase.
* Updated RepositoryZypper to make use of the exclude_docs flag and include `rpm.install.excludedocs` parameter accordingly in the libzypp configuration file. 
* Removed paths in the libzypp configuration files, as they are already provided in the command line as parameters and later on the are inconsistent in chrooted operations.
* Updated PackageMangerZypper to export a correct ZYPP_CONF environment variable in chrooted operations.
* Updated RootBind to provide a move_path_to_root method that works for a single path and makes sure it only replaces the first ocurrence of root_dir.

### DIFF
--- a/kiwi/boot/image/base.py
+++ b/kiwi/boot/image/base.py
@@ -180,6 +180,7 @@ class BootImageBase(object):
             'locale',
             'packagemanager',
             'rpm_check_signatures',
+            'rpm_excludedocs',
             'showlicense'
         ]
         self.xml_state.copy_preferences_subsections(
--- a/kiwi/repository/zypper.py
+++ b/kiwi/repository/zypper.py
@@ -63,8 +63,14 @@ class RepositoryZypper(RepositoryBase):
         :param list custom_args: zypper arguments
         """
         self.custom_args = custom_args
+        self.exclude_docs = False
         if not custom_args:
             self.custom_args = []
+
+        # extract custom arguments used for zypp config only
+        if 'exclude_docs' in self.custom_args:
+            self.custom_args.remove('exclude_docs')
+            self.exclude_docs = True
 
         self.repo_names = []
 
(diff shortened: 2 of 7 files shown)


davidcassany:
@isbm Bo, if you got the time, I would appreciate you to make a review on this PR as Marcus is on vacation.

kiwi/package_manager/zypper.py
```diff
@@ -149,11 +157,15 @@ def update(self):
         chroot_zypper_args = self.root_bind.move_to_root(
             self.zypper_args
         )
+        chroot_command_env = self.command_env
+        chroot_command_env['ZYPP_CONF'] = self.root_bind.move_path_to_root(
+            self.command_env['ZYPP_CONF']
+        )
```
schaefi: This should not be needed, ZYPP_CONF should already be part of the environment because the PackageManager instance get this information from self.repository.runtime_config() in the post_init constructor

kiwi/repository/__init__.py
```diff
@@ -41,9 +41,9 @@ class Repository(object):
         list of custom package manager arguments
         to setup the repository
     """
-    def __new__(self, root_bind, package_manager, custom_args=None):
+    def __new__(self, root_bind, package_manager, custom_args=None, exclude_docs=False):
```
schaefi: for passing additional arguments the class provides the custom_args parameter. I think exclude_docs should be part of custom_args not an additional argument to the constructor. We might end up with a continuously growing param list which I'd like to avoid especially because we try to provide a common interface with Repository and PackageManager for different implementations (zypper, yum, apt... maybe more) and so I just like to pass in a dictionary with customization values and keep the class design as stable as possible.

That way the class implementing the functionality can work over the custom_args and provide information if the data is used, ignored, not implemented by us, or by the tool without changing the interface all the time we add new values

kiwi/package_manager/zypper.py
davidcassany: I think this is needed, this is to update ZYPP_CONF path to the chrooted env. It is true that the var is passed, but it is passed with a wrong path (i.e. `/tmp/image/build/image-root/tmp_config_file` should be converted to something like `/tmp_config_file`). It happens that exclude docs (and any other configuration in the conf file) is not applied in image packages, since zypper is executed in a chrooted operation and cannot find the conf file, so all defaults are applied.

PR #137, opened 2016-09-07
travis: Update pypi password

toabctl:
Fixes # .

*
*

Next try to get the automatic deployment on pypi fixed.


toabctl:
@schaefi I updated the password for the ```suse``` user. Let's see if that helps. The same config works for a private project so that's the last thing I can try (or maybe use my private account if that now does not work)....

schaefi:
Thanks again :+1:

PR #139, opened 2016-09-09
Schema documentation

davidcassany:
Added Schema documentation helper script

The schema_parser.py helper script is called within the tox documentation generation in order to create new schema documentation from the RELAXNG schema (kiwi/schema/kiwi.rng). The script produces warning for each element or attribute that has no documentation.


schaefi:
This is great work :+1:  Thanks

PR #143, opened 2016-09-21
Custom device timeout variable added

davidcassany:
#123 PR did not solve the problem and had to be reverted. A more simple solution is now proposed.

This PR addresses the issue bnc#992992. Now DEVICE_TIMEOUT variable can be used in order to set the device waiting timeout in watForStorageDevice function. Must be a numeric value expressed in seconds.

### DIFF
--- a/kiwi/boot/functions.sh
+++ b/kiwi/boot/functions.sh
@@ -5534,8 +5534,8 @@ function waitForStorageDevice {
     local device=$1
     local check=0
     local limit=30
-    if [ -n ${DEVICE_TIMEOUT##*[!0-9]*} ]; then
-	limit=$((DEVICE_TIMEOUT / 2))
+    if [[ $DEVICE_TIMEOUT =~ ^[0-9]+$ ]]; then
+	limit=$(((DEVICE_TIMEOUT + 1)/ 2))
     fi
     udevPending
     while true;do
--- a/kiwi/boot/functions.sh
+++ b/kiwi/boot/functions.sh
@@ -5524,19 +5524,26 @@ function waitForStorageDevice {
     # function to check access on a storage device
     # which could be a whole disk or a partition.
     # the function will wait until the size of the
-    # storage device could be obtained or the check
-    # counter equals 4
+    # storage device could be obtained or the timeout
+    # is reached. Default timeout is 60 seconds, however
+    # it can be set to different value by setting the
+    # DEVICE_TIMEOUT variable in the kernel command
+    # line.
     # ----
     local IFS=$IFS_ORIG
     local device=$1
     local check=0
+    local limit=30
+    if [ -n ${DEVICE_TIMEOUT##*[!0-9]*} ]; then
+	limit=$((DEVICE_TIMEOUT / 2))
+    fi
     udevPending
     while true;do
         partitionSize $device &>/dev/null
-        if [ $? = 0 ];then
+        if [ $? = 0 ]; then
             sleep 1; return 0
         fi
-        if [ $check -eq 30 ];then
+        if [ $check -eq $limit ]; then
             return 1
         fi
         Echo "Waiting for storage device $device to settle..."


davidcassany:
@schaefi I could not find a better way to document the usage of the variable DEVICE_TIMEOUT in kernel command line (used to set the waitForStorage timeout). Is there anywhere else it should be listed or added?

schaefi:
There is documentation in the doc team for kiwi v7 where you would find a place to add this information. For kiwi v8 (this pull request) we would need to add a section about boot variable for the kiwi initrd which would be a very good idea but should be done as an extra task.

Could you open a card in trello to address this ? Would be great thanks

kiwi/boot/functions.sh
```diff
@@ -5524,19 +5524,26 @@ function waitForStorageDevice {
...
+    # is reached. Default timeout is 60 seconds, however
+    # it can be set to different value by setting the
+    # DEVICE_TIMEOUT variable in the kernel command
+    # line.
     # ----
     local IFS=$IFS_ORIG
     local device=$1
     local check=0
+    local limit=30
+    if [ -n ${DEVICE_TIMEOUT##*[!0-9]*} ]; then
```
schaefi: Stupid question, what is this construct good for ?

schaefi:
Apart from the review, we need this fix for the kiwi v7 version too.

kiwi/boot/functions.sh
davidcassany: I included it just to verify DEVICE_TIMEOUT is a number, this results in an empty string if DEVICE_TIMEOUT has any other character which is not included in [0-9]

davidcassany:
Yes I will handle Kiwi v7 also (including the documentation), however I wanted to solve it Kiwi v8 since it is easier for me to debug. Also I will add a card for documenting initrd in Kiwi v8.

kiwi/boot/functions.sh
schaefi: Are you sure ? That's what I thought too and wrote the following test

```bash
DEVICE_TIMEOUT=fdfd

if [ -n ${DEVICE_TIMEOUT##*[!0-9]*} ]; then
  echo $DEVICE_TIMEOUT
fi
```

schaefi: This one works for me

```bash
if [[ ${DEVICE_TIMEOUT} =~ [0-9]+ ]];then
    echo $DEVICE_TIMEOUT
fi
```

schaefi: To be honest, I'm with bash for some time but never saw the syntax you used :smile: that's why I asked if bash can do the number check in another way than a regexp too

schaefi:
great thanks

kiwi/boot/functions.sh
schaefi: The regexp above is most probably also not enough because we should have a min timeout which is not below 2sec

davidcassany: Yes you are right, there is a mistake here :disappointed:  I missed to add double quotes so it is evaluated as string. The following works:

```bash
if [ -n "${DEVICE_TIMEOUT##*[!0-9]*}" ]; then
    echo $DEVICE_TIMEOUT
fi
```
I have not used double brackets and regexp since I could not easily figure out it is not Bash specific syntax and I was not sure this script is running under a bash shell. I am totally fine with regexp, we can change it if is it going to be more consistent with the rest of the code.

About not being under 2 seconds... I think I would not care, it would result in 0 seconds timeout instead of 1 second (something that we can't achieve unless we modify our waiting steps to `sleep 1`). Moreover someone that uses DEVICE_TIMEOUT=1 I think it is going to be aware enough to increase it if something fails due to not waiting enough. Or, on the other hand, we could ceil the number with something like:

```bash
limit=$(((DEVICE_TIMEOUT + 1) / 2))
```

davidcassany:
@scaefi just added the regexp validation and ceiling the number in order to get always an integer result after dividing by 2.

By the way I noticed sometimes landscape checker get stuck with the validation, do you have an idea why it could happen?

kiwi/boot/functions.sh
schaefi: * ceil the number sounds good to me
* regarding '[[' I use it all over the initrd code, thus if the shell evaluating this can't do it... bad luck. So I leave it up to you what you like better, if it works ;)

schaefi:
the landscape service is sometimes very slow or under heavy load. If it does not finish I don't care much since tests and flake8 ensures an adequate quality. I normally look over landscape once a month and then do a "landscape day" and fix the complains :)

schaefi:
looks good to me now :+1:

PR #146, opened 2016-09-26
grub hotfixes

schaefi:
* Only sync .mod grub2 module files
* Mount boot volumes on grub install
    
    If there are volumes below /boot they need to be mounted before
    grub2-install / shim-install is called in order to make sure all
    data is available in the volume


schaefi:
sorry I had to merge this today in order to continue another task, hope it will not strike back

PR #148, opened 2016-09-29
Exclude docs yum

### LINKED ISSUE #133: rpm-excludedocs handling missing

schaefi:
Added rpm_excludedocs handling for yum
    
rpm supports the --excludepath option. However, yum can not be configured to pass along options to rpm or the python interface  it uses. Thus only a warning about excludedocs not being  supported by kiwi for yum is issued. Fixes #133

### DIFF
--- a/kiwi/repository/apt.py
+++ b/kiwi/repository/apt.py
@@ -55,8 +55,14 @@ class RepositoryApt(RepositoryBase):
         :param list custom_args: apt-get arguments
         """
         self.custom_args = custom_args
+        self.exclude_docs = False
         if not custom_args:
             self.custom_args = []
+
+        # extract custom arguments used for apt config only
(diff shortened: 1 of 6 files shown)


schaefi:
the git fufu failed

davidcassany:
I agree that we might not support for yum. There was a discussion about that in Yum mailing list some several years ago (so it might me totally oudated), see the following:

- http://lists.baseurl.org/pipermail/yum/2004-May/015483.html

Apparently it is possible to exlude docs with yum using rpmmacros. The problem is that I could not find any documentation on how modify rpmmacros file path, always located in user home, this is a blocking issue for a non chrooted operations from my point of view.

Adding the comment just for the record.

PR #151, opened 2016-09-30
Added support for mkisofs

davidcassany:
Support for mkisofs tool (instead of genisoimage) was not properly implemented, since these tools do not support exactly the same flags. It can be solved by making sure we use only common flags, this way is transparent to kiwi which is the underlying tool installed in the platform.

### DIFF
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -243,6 +243,7 @@ class LiveImageBuilder(object):
             log.info('ISO exceeds 4G size, using UDF filesystem')
             custom_iso_args['create_options'].append('-iso-level')
             custom_iso_args['create_options'].append('3')
+            custom_iso_args['create_options'].append('-udf')
 
         # create iso filesystem from media_dir
         log.info('Creating live ISO image')
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -241,8 +241,8 @@ class LiveImageBuilder(object):
         # calculate size and decide if we need UDF
         if rootsize.accumulate_mbyte_file_sizes() > 4096:
             log.info('ISO exceeds 4G size, using UDF filesystem')
-            custom_iso_args['create_options'].append('-allow-limited-size')
-            custom_iso_args['create_options'].append('-udf')
+            custom_iso_args['create_options'].append('-iso-level')
+            custom_iso_args['create_options'].append('3')
 
         # create iso filesystem from media_dir
         log.info('Creating live ISO image')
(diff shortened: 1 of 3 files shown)


kiwi/builder/live.py
```diff
@@ -241,8 +241,8 @@ def create(self):
         # calculate size and decide if we need UDF
         if rootsize.accumulate_mbyte_file_sizes() > 4096:
             log.info('ISO exceeds 4G size, using UDF filesystem')
-            custom_iso_args['create_options'].append('-allow-limited-size')
-            custom_iso_args['create_options'].append('-udf')
+            custom_iso_args['create_options'].append('-iso-level')
+            custom_iso_args['create_options'].append('3')
```
schaefi: just for my own understanding; does iso-level 3 imply that the UDF filesystem is used ? This is important for the kiwi boot code in order to work properly

davidcassany: No, it does not imply udf filesystem. As those flags are applied only on images bigger than 4Gb I just tested that iso-level 3 also supports such a big image, I have built it and dd'ed it to a USB, but I have not done it in real DVD, if that is the case you have in mind. Some days before I tested as CD-ROM input in a libvirt virtual machine.

Do you have in mind somewhere it could fail? I have no included this flag because is different in mkisofs (-UDF) and in genisoimage (-udf)

davidcassany: mkisofs command has -udf and -UDF flags. genisoimage equivalent to -udf is -UDF, however in mkisofs -udf is a sort of flavour of UDF, so I guess it implies it, I am checking it right now. If that's the case then we could test it by adding this flag as it was before and see if the ouput is UDF. If that's the case we would be in a safer position I guess

schaefi: The reason for udf as filesystem was this bug: https://bugzilla.suse.com/show_bug.cgi?id=770703

schaefi: It is quite some time ago, but support for >4g iso images was implemented with iso-level 3 plus UDF. I don't know if this is really required from a todays perspective. If you can correctly boot the iso in full speed mode without UDF we are fine, but I think this is still a valid point

davidcassany: I am not sure how to verify I am actually booting in full speed mode, however I booted without experimenting performance issues, it booted quickly (but this might be due to using a JeOS image).

On the other hand I just built an image with mkisofs using '-iso-level 3' and '-udf'  flags and I can verify it is in UDF files system, now copying it in a USB to be sure it boots in hybrid mode. Now is being build with genisoimage to test it also works with this flags.

davidcassany: Apparently it works fine using -udf in genisoimage and in mkisofs. In both cases I could boot a virtual machine from the generated ISO and I could also boot from a USB device using that ISO.

schaefi:
:+1:

PR #156, opened 2016-10-06
Added btrfs_root_is_readonly_snapshot attribute

schaefi:
The attribute allows to specify if the root filesystem should be set to read-only if it is created as a btrfs snapshot. The option only has an effect if a btrfs snapshot is used as root filesystem. Fixes bnc#1000080


schaefi:
re-activate travis by closing and re-open the request

PR #157, opened 2016-10-06
Support for btrfs read-only root snapshot

schaefi:
Added btrfs_root_is_readonly_snapshot attribute
    
The attribute allows to specify if the root filesystem should be set to read-only if it is created as a btrfs snapshot. The option only has an effect if a btrfs snapshot is used as root filesystem. Fixes bnc#1000080


schaefi:
Note: kiwi.rng and xml_parse.py are auto generated code files and can be skipped from review

schaefi:
Thanks, yes the project making use of this feature will do some other crazy stuff. I'll let you know when I know all the details :)

PR #161, opened 2016-10-10
Bundler setup changes

### LINKED ISSUE #159: kiwi bundle procedure compress all files
## Problem description

kiwi bundle compress and keeps all image files it produces (including the raw image), however in the previous kiwi version no compression was performed and the raw image was not included if it was not the target format.

Having the raw image (even if requested another format) can be handy and it looks like a must to compress it (as it is likely to be in the order of GBs and compressed is likely to be in the order of hundreds of MBs), however having the requested format without compression should be taken into consideration in order to keep compatibility with previous kiwi versions and also to facilitate integration with repository systems (i.e. OBS). 

Also, compressing the raw image might take quite long (a raw image might huge, compressing a 32GB takes quite long), so I would suggest to make the raw image provisioning as an optional feature, as an additional flag of kiwi bundle.

schaefi:
Fixes #159 

Do not bundle the raw disk if a format is setup
    
* Only bundle the compressed version of the .raw disk image if no disk format like qcow2, vmdk, etc... is specified.

Do not compress disk formats
    
* Stay compatible with the former version of kiwi and do not compress disk formats like qcow2, vmdk, etc... It also does not make much sense since the disk formats itself are using a compression algorithm.

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -494,7 +494,7 @@ class DiskBuilder(object):
         self.result.add(
             key='disk_image',
             filename=self.diskname,
-            use_for_bundle=True,
+            use_for_bundle=True if not self.image_format else False,
             compress=True,
             shasum=True
         )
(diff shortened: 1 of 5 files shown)


schaefi:
any disk format won't be compressed correct, that should be the same compared to v7

PR #163, opened 2016-10-13
Create info.xml file for btrfs snapshot

schaefi:
If the system is installed into a btrfs snapshot a metadata file called info.xml is created which is used by tools like snapper. Fixes bnc#1000117

### DIFF
--- a/kiwi/volume_manager/btrfs.py
+++ b/kiwi/volume_manager/btrfs.py
@@ -17,6 +17,9 @@
 #
 import re
 import os
+import datetime
+from xml.etree import ElementTree
+from xml.dom import minidom
 
 # project
 from ..command import Command
@@ -267,6 +270,9 @@ class VolumeManagerBtrfs(VolumeManagerBase):
             sync_target = self.mountpoint + '/@'
             if self.custom_args['root_is_snapshot']:
                 sync_target = self.mountpoint + '/@/.snapshots/1/snapshot'
+                self._create_snapshot_info(
+                    self.mountpoint + '/@/.snapshots/1/info.xml'
+                )
             data = DataSync(self.root_dir, sync_target)
             data.sync_data(
                 options=['-a', '-H', '-X', '-A', '--one-file-system'],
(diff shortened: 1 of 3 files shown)


kiwi/volume_manager/btrfs.py
```diff
@@ -267,6 +270,9 @@ def sync_data(self, exclude=None):
             sync_target = self.mountpoint + '/@'
             if self.custom_args['root_is_snapshot']:
                 sync_target = self.mountpoint + '/@/.snapshots/1/snapshot'
+                self._create_snapshot_info(
+                    self.mountpoint + '/@/.snapshots/1/info.xml'
+                )
```
isbm: Guys-guys. Please remember [this](http://kotsf.com/gdarchive/7/), [this](https://wiki.python.org/moin/PythonSpeed/PerformanceTips) and [this](http://www.wikihow.com/Avoid-Repeating-the-Old-Same-Mistakes-All-over-Again).

test/unit/volume_manager_btrfs_test.py
```diff
@@ -1,7 +1,10 @@
-
 from mock import patch
 from mock import call
 import mock
+import os
+
+from textwrap import dedent
+import datetime
 
 from .test_helper import *
```
isbm: Let me [save 128K of words](https://pythonconquerstheuniverse.wordpress.com/2011/03/28/why-import-star-is-a-bad-idea/) for this. :smile:

test/unit/volume_manager_btrfs_test.py
```diff
@@ -13,6 +16,22 @@
...
+        setattr(self.context_manager_mock, '__exit__', self.exit_mock)
+        self.xml_info = dedent('''
+        <?xml version="1.0" ?>
+        <snapshot>
+            <type>single</type>
+            <num>1</num>
+            <description>first root filesystem</description>
+            <date>2016-01-01 00:00:00</date>
+        </snapshot>
+        ''').strip() + os.linesep
```
isbm: Maybe it will work for you too: in Salt unit tests we are trying remove these kind of data outside into a text files and then load on demand.

schaefi: yes I think that is better

schaefi:
@isbm hope that's better now :smile:

test/unit/volume_manager_btrfs_test.py
```diff
@@ -6,7 +6,10 @@
 from textwrap import dedent
 import datetime
 
-from .test_helper import *
+from .test_helper import (
+    raises, patch_open
+)
```
isbm: @schaefi I am not klugscheissing, but when I do...
The `(` and `)` parenthesis in imports are _only_ needed for a [multi-line imports](https://www.python.org/dev/peps/pep-0328/#rationale-for-parentheses). In this case you are doing an overkill. It is enough:
```python
from .test_helper import raises, patch_open
```

isbm: Parenthesis would be correct if you would:
```python
from .test_helper import (
    raises,
    patch_open
)
```
Sure you can do that too, but [don't](https://medium.com/@webseanhickey/the-evolution-of-a-software-engineer-db854689243#.qxgpwbfak).

isbm:
XML [load from file](http://imgur.com/gallery/CnGdhOd)? :laughing:

isbm:
LGTM :+1:

PR #164, opened 2016-10-14
Efi setup for removable devices

schaefi:
Added target_removable attribute
    
Indicate if the target disk for oem images is deployed to a removable device e.g a USB stick or not. This only affects the EFI setup if requested and in the end avoids the creation of a custom boot menu entry in the firmware of the target machine on first boot. This is related to bnc#993130

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -1614,7 +1614,7 @@ div {
         ## affects the EFI setup if requested and in the end avoids
         ## the creation of a custom boot menu entry in the firmware
         ## of the target machine. By default the target disk is
-        ## expected to be solid
+        ## expected to be non-removable
         attribute target_removable { xsd:boolean }
     k.type.zipl_targettype.attribute =
        ## The device type of the disk zipl should boot. On zFCP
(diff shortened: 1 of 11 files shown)


schaefi:
Autogenerated code:

* kiwi/schema/kiwi.rng 
* kiwi/xml_parse.py

please do not review those two, thanks

kiwi/schema/kiwi.rnc
```diff
@@ -1608,6 +1608,14 @@ div {
...
         ## desired target by calling: blockdev --report device
         attribute target_blocksize { xsd:nonNegativeInteger }
+    k.type.target_removable =
+        ## Indicate if the target disk for oem images is deployed
+        ## to a removable device e.g a USB stick or not. This only
+        ## affects the EFI setup if requested and in the end avoids
+        ## the creation of a custom boot menu entry in the firmware
+        ## of the target machine. By default the target disk is
+        ## expected to be solid
+        attribute target_removable { xsd:boolean }
```
schaefi: makes sense, I updated it

schaefi:
@davidcassany guess we are good to go ?

davidcassany:
@schaefi Yes I think we are good :smiley:

PR #165, opened 2016-10-18
Disk resize

schaefi:
Added kiwi image resize command
    
The image resize command allows to resize a disk image and its optional disk format to a new disk geometry

### DIFF
--- a/kiwi/cli.py
+++ b/kiwi/cli.py
@@ -17,6 +17,12 @@
 #
 """
 usage: kiwi -h | --help
+       kiwi [--profile=<name>...]
+            [--type=<build_type>]
+            [--logfile=<filename>]
+            [--debug]
+            [--color-output]
+           image <command> [<args>...]
        kiwi [--debug]
             [--color-output]
            result <command> [<args>...]
@@ -43,7 +49,7 @@ global options:
     help
         show manual page
 
-global options for service: system
+global options for services: image, system
     --logfile=<filename>
         create a log file containing all log information including
         debug information even if this is was not requested by the
@@ -119,7 +125,9 @@ class Cli(object):
         :return: service name
         :rtype: string
         """
-        if self.all_args['system']:
+        if self.all_args['image']:
+            return 'image'
+        elif self.all_args['system']:
             return 'system'
         elif self.all_args['result']:
             return 'result'
--- a/kiwi/cli.py
+++ b/kiwi/cli.py
@@ -125,13 +125,13 @@ class Cli(object):
         :return: service name
         :rtype: string
         """
-        if self.all_args['image']:
+        if self.all_args.get('image') is True:
             return 'image'
-        elif self.all_args['system']:
+        elif self.all_args.get('system') is True:
             return 'system'
-        elif self.all_args['result']:
+        elif self.all_args.get('result') is True:
             return 'result'
-        elif self.all_args['--compat']:
+        elif self.all_args.get('--compat') is True:
             return 'compat'
         else:
             raise KiwiUnknownServiceName(
--- a/kiwi/storage/subformat/base.py
+++ b/kiwi/storage/subformat/base.py
@@ -20,9 +20,11 @@ import platform
 from collections import OrderedDict
 
 from ...exceptions import (
-    KiwiFormatSetupError
+    KiwiFormatSetupError,
+    KiwiResizeRawDiskError
 )
 
+from ...command import Command
 from ...defaults import Defaults
 from ...path import Path
 from ...logger import log
@@ -79,6 +81,24 @@ class DiskFormatBase(object):
         :param list custom_args: unused
         """
         pass
+
+    def has_raw_disk(self):
+        return os.path.exists(self.diskname)
+
+    def resize_raw_disk(self, size_bytes):
+        """
+        Resize raw disk image to specified size
+        """
+        current_byte_size = os.path.getsize(self.diskname)
+        size_bytes = int(size_bytes)
+        if size_bytes <= current_byte_size:
+            raise KiwiResizeRawDiskError(
+                'Can not shrink %s disk to %d bytes without data corruption' %
+                (self.diskname, size_bytes)
+            )
+        Command.run(
+            ['qemu-img', 'resize', self.diskname, format(size_bytes)]
+        )
 
     def create_image_format(self):
         """
--- a/kiwi/storage/subformat/base.py
+++ b/kiwi/storage/subformat/base.py
@@ -105,8 +105,9 @@ class DiskFormatBase(object):
         size_bytes = int(size_bytes)
         if size_bytes < current_byte_size:
             raise KiwiResizeRawDiskError(
-                'Can not shrink %s disk to %d bytes without data corruption' %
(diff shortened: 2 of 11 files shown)


schaefi:
[kiwi image resize](https://trello.com/c/EQsllK3y/197-kiwi-image-resize)

kiwi/storage/subformat/base.py
```diff
@@ -80,6 +82,24 @@ def post_init(self, custom_args):
...
+    def has_raw_disk(self):
+        return os.path.exists(self.diskname)
+
+    def resize_raw_disk(self, size_bytes):
+        """
+        Resize raw disk image to specified size
+        """
+        current_byte_size = os.path.getsize(self.diskname)
+        size_bytes = int(size_bytes)
+        if size_bytes <= current_byte_size:
```
davidcassany: I am not sure the equality is strictly needed here. If the raw disk has the same size as requested I think it should just skip this step, but no fail. Simple example that happened to me while testing it: I forgot to include the type and profile I used to build the image (`image="oem"` and `profile="vmxFlavour"`), so the command resize selected the default ones (`image="iso"` and `profile="vmxFlavour"`), so it resized the raw disk but it did not perform any format process because in iso type there is no format specified (I did the previous oem build with `format="qcow2"`). The I run again the same command including the oem type expecting it to not do any resize (or do a transparent/dummy resize) and also include the image formatting, however it failed with this KiwiResizeRawDiskError. In order to get the formatted qcow2 disk I had to resize again to a bigger size than I initially wanted.

Maybe if equal size is requested the command could be skipped, or even run it since running a resize process with `qemu-img` twice (with the same size value) over the same file does not change the initial result (I validated it with diff)

kiwi/cli.py
```diff
@@ -119,7 +125,9 @@ def get_servicename(self):
         :return: service name
         :rtype: string
         """
-        if self.all_args['system']:
+        if self.all_args['image']:
+            return 'image'
```
isbm: Scary. :laughing: It is generally a bad idea to hammer the key directly on _testing_ it. I would prefer:
```python
if self.all_args.get('anykey') is not None:
    return 'whatever needs to be here'
elif ...
```

kiwi/storage/subformat/base.py
```diff
@@ -80,6 +82,24 @@ def post_init(self, custom_args):
...
+            raise KiwiResizeRawDiskError(
+                'Can not shrink %s disk to %d bytes without data corruption' %
+                (self.diskname, size_bytes)
```
isbm: I remember you were switching to the `"foo {0}".format("bar")` way.

kiwi/tasks/image_resize.py
```diff
@@ -0,0 +1,135 @@
...
+        Instance of Help
+    """
+    def process(self):
+        """
+        reformats raw disk image and its format to a new disk
+        geometry using the qemu tool chain
+        """
+        self.manual = Help()
+        if self._help():
+            return
```
isbm: Meh. Why call the whole thing from man, instead of just:
```python
if self.command_args.get(''help'):
    return None
```
BTW "explicit is better than implicit" (`import this`), thus `return None` is better than just `return`.

kiwi/tasks/image_resize.py
```diff
@@ -0,0 +1,135 @@
...
+                (size_value, size_format)
+            )
+        size_base = int(size.group(1))
+        size_unit = size.group(2).lower()
+        if size_unit == 'g':
+            return size_base * math.pow(1024, 3)
+        elif size_unit == 'm':
+            return size_base * math.pow(1024, 2)
+        else:
+            return size_base
```
isbm: Can be shorter:
```python
size_base = int(size.group(1))
size_unit = {'g': 3, 'm': 2}.get(size.group(2).lower())
return size_unit and size_base * math.pow(0x400, size_unit) or size_base
```

kiwi/storage/subformat/base.py
schaefi: makes sense I will change this

schaefi:
Thanks for the review, I have changed the mentioned code parts. However I fear the tests will fail due to an unrelated issue. I found that the xattr module does not play well with the new release of pycparser 2.15. That results in very strange error messages at travis. I have tried to fix this in the pip requirements file by explicitly installing pycparser 2.14 but while that works locally I see i the travis build that the pip install of xattr still pulls in pycparser 2.15 and then fails the install of xattr. Any idea how to fix this would be very much appreciated

schaefi:
Hmm, I have no clue why the travis builds for master currently fails, while the pull requests work ??? I just see that pycparser 2.15 vs 2.14 causing the trouble when the required xattr module is installed

schaefi:
Code adapted, thanks for the review

kiwi/tasks/image_resize.py
```diff
@@ -0,0 +1,127 @@
...
+
+        disk_format = self.xml_state.build_type.get_format()
+
+        image_format = DiskFormat(
+            disk_format or 'raw', self.xml_state, image_root,
+            self.command_args['--target-dir']
+        )
+        if not image_format.has_raw_disk():
+            raise KiwiImageResizeError(
+                'no raw disk image %s found in build results' %
```
isbm: Seems

kiwi/tasks/image_resize.py
```diff
@@ -0,0 +1,127 @@
...
+                image_format.diskname
+            )
+
+        new_disk_size = self._to_bytes(self.command_args['--size'])
+
+        log.info(
+            'Resizing raw disk to %d bytes', new_disk_size
```
isbm: like

kiwi/tasks/image_resize.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+        resize_result = image_format.resize_raw_disk(new_disk_size)
+        if disk_format and resize_result is True:
+            log.info(
+                'Creating %s disk format from resized raw disk', disk_format
```
isbm: it (%s thing)

kiwi/tasks/image_resize.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+            image_format.create_image_format()
+        elif resize_result is False:
+            log.info(
+                'Raw disk is already at %d bytes', new_disk_size
```
isbm: is

kiwi/tasks/image_resize.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+
+    def _to_bytes(self, size_value):
+        size_format = '^(\d+)([gGmM]{0,1})$'
+        size = re.search(size_format, size_value)
+        if not size:
+            raise KiwiImageResizeError(
+                'unsupported size format %s, must match %s' %
```
isbm: everywhere...

schaefi:
@isbm there are more places where I'd like to cleanup the % vs .format messaging. So far it should be clean in the scope of this pull request. Thanks

schaefi:
@isbm are you ok with the changes ? if so would be great to get your approval so I can submit a new release. Thanks much

PR #168, opened 2016-10-20
Fixed a rounding error in partedMBToCylinder method

davidcassany:
bc truncates number of cylinders to zero decimal places, which results
in a partition that is slightly smaller than the requested size. Add one
cylinder to compensate.

This is a port form the patch proposed in former kiwi versions here: openSUSE/kiwi#605


schaefi:
:+1:

PR #173, opened 2016-10-25
Attribute co-constraints based on image type

davidcassany:
This PR adds co-constraints in the kiwi schema using the Schematron syntax, which can nicely co-exist with the RelaxNG syntax and it is also supported by the underlying python module lxml. Now each `<type>` section attribute is assigned to an image type, this way if any attribute is incoherent with the selected image type the validation fails.


schaefi:
This is great work :+1: I will merge it now. Thanks

PR #174, opened 2016-10-25
Cache location configurable

### LINKED ISSUE #92: Request to add rpm cache location configurable
## Problem description
Currently kiwi 8 creates a directory in /var/cache and puts downloaded rpms in there. We are wanting to contain all the files that gets created during the kiwi appliance creation process within the targetdir so that we manage them appropriately. Please provide an option to override the default behavior.

verbatim from kiwi google groups:
==========================

Actually the package manager cache is to share data for all builds on 
the building host, thus I had planned to keep this as a fixed location. 
However it is possible to use another location for it, there is just 
no switch to change from the default place implemented 

If you need that please open an issue on the github issue page 

Regards, 
Marcus

schaefi:
Added global option --shared-cache-dir
    
The option allows to specify an alternative shared host_to_image cache directory. The default location is /var/cache/kiwi. Fixes #92

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -65,13 +65,15 @@ class Defaults(object):
         """
         The shared location is a directory which shares data from
         the image buildsystem host with the image root system.
+        The location is returned as an absolute path stripped off
+        by the leading '/'. This is because the path is transparently
+        used on the host /<cache-dir> and inside of the image
+        imageroot/<cache-dir>
         """
         from .cli import Cli
-        global_args = Cli().get_global_args()
-        if global_args['--shared-cache-dir']:
-            return global_args['--shared-cache-dir']
-        else:
-            return 'var/cache/kiwi'
+        return os.path.abspath(os.path.normpath(
+            Cli().get_global_args().get('--shared-cache-dir')
+        )).lstrip(os.sep)
 
     @classmethod
     def get_failsafe_kernel_options(self):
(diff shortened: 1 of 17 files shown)


schaefi:
Along with the option a cleanup of the tests were required. I'm sorry for the mixup, hope a review is still possible. Thanks

kiwi/defaults.py
```diff
@@ -66,7 +66,12 @@ def get_shared_cache_location(self):
         The shared location is a directory which shares data from
         the image buildsystem host with the image root system.
         """
-        return 'var/cache/kiwi'
+        from .cli import Cli
+        global_args = Cli().get_global_args()
+        if global_args['--shared-cache-dir']:
+            return global_args['--shared-cache-dir']
+        else:
+            return 'var/cache/kiwi'
```
isbm: Nope. Double returns are very rarely correct. The whole thing is oneliner:

```python
return Cli().get_global_args().get('--shared-cache-dir', 'var/cache/kiwi')
```

Also, if `var/cache/kiwi` is correct to be relative for some very odd reasons.

schaefi: which reminds me that I need to care for this when the user specifies the location

schaefi:
@isbm do you give us your ok wrt the latest update ? Thanks

schaefi:
[configurable cache location](https://trello.com/c/KwBsJfBU/195-configurable-cache-location)

schaefi:
Tested several combinations for cache-dirs and it worked nicely.

@isbm mind if I merge it ?

schaefi:
guess it's ok to move on ;)

PR #176, opened 2016-10-26
SystemPrepare instance delete before runnig SystemSetup

### LINKED ISSUE #175: In kiwi boot initrd images SystemPrepare is cleaned up after SystemSetup
## Problem description

When using local repositories (i.e. in OBS) they can be deleted during the SystemSetup procedure while creating kiwi initrd image. The problem is that SystemSetup runs before the SystemPrepare is freed.

Use multiple profiles in OBS to create multiple vmx images using the kiwi initrd.

davidcassany:
This PR fixes #175. Now the SystemPrepare instance is cleaned before any other change over the build root is performed by a SystemSetup instance.

### DIFF
--- a/kiwi/boot/image/kiwi.py
+++ b/kiwi/boot/image/kiwi.py
@@ -78,6 +78,8 @@ class BootImageKiwi(BootImageBase):
         system.pinch_system(
             manager=manager, force=True
         )
+        # make sure system instance is cleaned up before setting up
+        del system
 
         self.setup.call_image_script()
         self.setup.create_init_link_from_linuxrc()


schaefi:
Hmm with the deletion of the system instance also the repository setup will be cleaned up. By definition of the custom scripts, in this case imges.sh we tell users that they still have access to the repo when the script runs. Thus the call

```python
 self.setup.call_image_script()
```

should be done with an active system.

Can you explain why exactly we need the cleanup prior to the setup tasks ? Would it also solve the problem if you cleanup after the script call ?

davidcassany:
Hmmm interesting, I see in images.sh some packages could be deleted or something like, right? 

Well the case is simple (however it took me some hours to get it :disappointed:), in `strip.xml` there is a directory to delete, `/usr/src/pacakges/`, which includes the mount point of all the OBS repositories, thus, all the repositories are deleted when the strip.xml rules are applied in SystemSetup. Again it only happens in OBS (also using osc locally). If using multiple profiles it happens that the second profile doesn't have any repository in place (OBS creates the local repositories only once). If only a single profile is used it doesn't lead to any errors since the repositories aren't needed anymore after stripping the image (at least in the image I tested which is very simple and without any custom configuration).

Finally, note that in kiwi v7 the system equivalent is cleaned up before the image is stripped.

davidcassany:
The images.sh script stderr outputs something like this: 

```
[  460s] [ DEBUG   ]: 23:25:51 | --------------err start-------------
[  460s] [ DEBUG   ]: 23:25:51 | rm: cannot remove 'usr/src/packages/SOURCES/repos/Virtualization:Appliances:Builder/openSUSE_Leap_42.2': Device or resource busy
[  460s] rm: cannot remove 'usr/src/packages/SOURCES/repos/openSUSE:Leap:42.2/standard': Device or resource busy
[  460s] rm: cannot remove 'usr/src/packages/SOURCES/repos/openSUSE:Leap:42.2:NonFree:Update/standard': Device or resource busy
[  460s] rm: cannot remove 'usr/src/packages/SOURCES/repos/openSUSE:Leap:42.2:Update/standard': Device or resource busy
[  460s] rm: cannot remove 'usr/src/packages/SOURCES/repos/openSUSE:Leap:42.2:NonFree/standard': Device or resource busy
[  460s] 
[  460s] [ DEBUG   ]: 23:25:51 | --------------err stop--------------
```

Those directories are not deleted because they are a mount point, this is the result of `rm -rf /usr/src/packages`, so all the repo contents are lost.

davidcassany:
Also, we could just remove `/usr/src/packages` from strip.xml, but I thought there might be a good reason to include it in the list. This is the first thing I tested and it worked out, but I did not check if the resulting image was significantly bigger or not, as I was only trying to verify that this was the actual problem.

schaefi:
Thanks for the clarification, that makes a lot sense. You are right local repos are bind mounted and thus gets removed in this case. I'm ok with your change, let's go for it

PR #177, opened 2016-10-27
Do not allow imageinclude for images build in obs

schaefi:
When building in the buildservice the repos are moved into a local source path which only exists on that worker. Using the imageinclude attribute to setup that repo as a repo in the image doesn't make sense and should result in a runtime exception

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -61,6 +61,12 @@ class Defaults(object):
         ]
 
     @classmethod
+    def is_obs_worker(self):
+        # the presence of /.buildenv on the build host indicates
+        # we are building inside of the open buildservice
+        return os.path.exists('/.buildenv')
+
+    @classmethod
     def get_shared_cache_location(self):
         """
         The shared location is a directory which shares data from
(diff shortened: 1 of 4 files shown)


kiwi/defaults.py
```diff
@@ -61,6 +61,12 @@ def __init__(self):
         ]
 
     @classmethod
+    def is_obs_worker(self):
+        # the presence of /.buildenv on the build host indicates
+        # we are building inside of the open buildservice
+        return os.path.exists('/.buildenv')
+
```
davidcassany: Wouldn't it be better to test it against the usage of `suse://` schema in the uri? This way all the ad-hoc code relies on the same entry point, and we have only the path of the repo hardcoded (even `.buildenv` doesn't look like is going to change).

However I have to admit that I do not have a clear idea where we could store the `is_obs_worker` flag. Maybe in xml_state when parsing the uri?

davidcassany:
@schaefi  that PR can be related to #178, so it get closed when merging.

kiwi/defaults.py
schaefi: at the point when the runtime checks run we have not yet adapted the obs:// style to suse:// style. The check in which environment we are running happens in the later in the builder. That's why I choose to provide a method which returns as where we are. Let's go with that for the moment because otherwise we have to put the runtime check later in the process which I'd like to avoid. Thoughts ?

davidcassany: Sounds reasonable to me. I don't really know how obs:// is converted to suse://, for some reason I though it was done by OBS before running kiwi, so that I considered suse:// to be our preferred indicator.

I agree the runtime checks should be as early as possible and centralized as possible.

PR #180, opened 2016-11-03
Load the default unicode font in grub2

### LINKED ISSUE #179: Grub does not load the default font if no themes with further fonts are present
## Problem description

In kiwi v8 grub menu appears with wrong characters when no theme is selected/installed. The frame border is done with `@` instead of the appropriate symbols (i.e simple lines `|`). Grub is still completely functional, however the menu is not rendered as it should.

davidcassany:
This commit fixes #179. The default font was never loaded,
the loadfont call for the default unicode font was missing.

### DIFF
--- a/kiwi/bootloader/template/grub2.py
+++ b/kiwi/bootloader/template/grub2.py
@@ -73,6 +73,9 @@ class BootLoaderTemplateGrub2(object):
         ''').strip() + os.linesep
 
         self.header_theme = self.fonts + dedent('''
+            if [ -f $${font} ];then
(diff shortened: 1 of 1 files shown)


schaefi:
Thanks for the patch, it looks good to me, however you should merge the current master into your branch. I had to hotfix master with regards to the sphinxcontrib-programoutput module not longer available on pypi

davidcassany:
@scheafi sure no problem

schaefi:
Thanks for finding this :+1:

PR #181, opened 2016-11-03
updateOtherDeviceFstab did not get the fstab location parameter

davidcassany:
With this PR the fstab file gets updated as it should in netboot deployments that had defined one or more non root or swap partitions.

This is port from the kiwi v7 code base openSUSE/kiwi#610


schaefi:
Thanks for the patch, it looks good to me, however you should merge the current master into your branch. I had to hotfix master with regards to the sphinxcontrib-programoutput module not longer available on pypi

PR #183, opened 2016-11-07
Don't prefix bootpath with snapshot path

thkukuk:
We call set-default to make the snapshot subvolume the
default root. In this case, grub2 and the kernel will
always use the correct snapshot. Prefixing it with the
snapshot path will only work, if /.snapshots is accessible
and as long as you don't do a rollback. In rollback case,
only the filesystem default is adjusted, but not grub.cfg.
So grub2 would try to load a kernel from an outdated
snapshot and in worst case, the snapshot is already
deleted and no kernel found anymore.

### DIFF
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -300,9 +300,7 @@ class BootLoaderConfigBase(object):
                     # setup and no extra boot partition we have to care for
                     # the layout of the system which places all volumes below
                     # a topleve volume or snapshot
-                    if not boot_is_on_volume and root_is_snapshot:
-                        bootpath = '/@/.snapshots/1/snapshot' + bootpath
-                    else:
+                    if not boot_is_on_volume and not root_is_snapshot:
                         bootpath = '/@' + bootpath
 
         return bootpath


schaefi:
I'm fine with the change, could you adapt the test for it too

Thanks

schaefi:
I understand the reason why this is an important change but when I test this with a local build
the result image does not boot anymore. Have you tested the change with your casp build, did it work
for you ?

All I got from grub is "error: file '/boot/vmlinuz-4.4.21-81-default' not found"

If I manually change the path in the grub shell to point to /@/.snapshots/1/snapshot it boots again.

from the kiwi log I see:

```
INFO: Installing grub2 on disk /dev/loop0
DEBUG: EXEC: [mountpoint /tmp/kiwi_mount_manager.co8dw3mr]
DEBUG: EXEC: [mount /dev/mapper/loop0p3 /tmp/kiwi_mount_manager.co8dw3mr]
```

* this correctly mounted the snapshot as it was set as default. I have checked that
* next all the subvol mounts follows which looks correct to me too
* next grub2-install is called in the chroot as follows

```
DEBUG: EXEC: [chroot /tmp/kiwi_mount_manager.co8dw3mr grub2-install --skip-fs-probe --directory /usr/lib/grub2/i386-pc --boot-directory /boot --target i386-pc --modules ext2 iso9660 linux echo configfile search_label search_fs_file search search_fs_uuid ls normal gzio png fat gettext font minicmd gfxterm gfxmenu video video_fb xfs btrfs lvm test multiboot part_gpt part_msdos biosdisk vga vbe chain boot /dev/loop0]
```

The version of grub2 in the chroot was: grub2-2.02~beta2-104.16.x86_64

Do you have any idea why grub2 does not know that it should take the data from the snapshot by default ?

thkukuk:
Hi,

On Wed, Nov 09, Marcus Schäfer wrote:

> I understand the reason why this is an important change but when I test this
> with a local build
> the result image does not boot anymore. Have you tested the change with your
> casp build, did it work
> for you ?

Yes, that works fine.

> All I got from grub is "error: file '/boot/vmlinuz-4.4.21-81-default' not
> found"

> If I manually change the path in the grub shell to point to /@/.snapshots/1/
> snapshot it boots again.

Which means btrfs subvolume set-default was not called in your
case on the right image.

If you even need /@/, this sounds like setup in kiwi/volume_manager/btrfs.py
is not called at all. 
If you mount the not bootable image on your system or boot into the
system by giving the full path, what does 
"btrfs subvolume get-default <mountpoint>" or
"btrfs subvolume get-default /" return?

It should be something like
ID XXX gen XXXXXX top level XXX path @/.snapshots/1/snapshot


  Thorsten

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CASP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

schaefi:
I added a break point before grub2-install is called and run

btrfs subvol get-default /tmp/kiwi_mount_manager.vpxsuvz3

==> ID 259 gen 20 top level 258 path @/.snapshots/1/snapshot

This is the correct path

schaefi:
Is /etc/default/grub relevant at the time of calling grub2-install ?

schaefi:
I included your changes together with my findings in a new pull request:

* https://github.com/SUSE/kiwi/pull/192

PR #185, opened 2016-11-08
ActivateImage links correction

davidcassany:
Loading the iso image into a read-only ramdisk caused some issues
due to the activatImage method cp commands. With this commit the
symlinks are consistent during the preinit phase and also before
running the cp command it is tested we are not in a read-only
filesystem.

This pull request is a port form openSUSE/kiwi#613 PR


schaefi:
:+1:

PR #187, opened 2016-11-08
vhdfixed format naming inconsistency

### LINKED ISSUE #184: Naming mismatch in vhd [-] fixed storage format
## Problem description
It seems that kiwi.rng does not recognize renamed "vhd-fixed" -> "vhdfixed" format

## Actual behaviour
I reused my kiwi profile for HyperV from v7, but got an error:
  "Support for vhd-fixed disk format not implemented"
When I change format to vhdfixed (as defined in kiwi/storage/subformat/__init__.py), I get an error:
  "EXEC: Command "jing" not found in the environment"

davidcassany:
When selecting vhd-fixed disk format the schema expects `vhd-fixed` value in `format` attribute, however the subformat initializer was expecting `vhdfixed` literal to initialize the appropriate DiskFormatVhdFixed class, thus it was leading to an unknown format error.

This PR fixes #184

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -1823,7 +1823,7 @@ div {
         ## To be remove from here by the end of 2014
         attribute format {
             "ec2" | "gce" | "ovf" | "ova" | "qcow2" | "vagrant" | "vmdk" |
-            "vdi" | "vhd" | "vhdfixed"
+            "vdi" | "vhd" | "vhd-fixed"
         }
         >> sch:pattern [ id = "format" is-a = "image_type"
             sch:param [ name = "attr" value = "format" ]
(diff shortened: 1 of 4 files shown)


schaefi:
I think we can't do that. in the v7 schema it is listed as "vhd-fixed" and that should stay compatible

schaefi:
For example see: Devel:PubCloud:Azure:ImagesSLE12/SLES12-SP2-Azure that kiwi description should stay valid with v8

davidcassany:
@schaefi sure you are right, I haven't though about this incompatibility. Then I think we must do it the other way round, update `kiwi/storage/subformat/__init__.py` file and translate any other 'vhd-fixed' literal in the code if needed. I am on it

schaefi:
yes that's the way to go. Thanks

schaefi:
:+1:

PR #189, opened 2016-11-08
Insufficient disk size for xfs

### LINKED ISSUE #186: Build fails when image size is not set
## Problem description
Element image->preferences->type->size default value is not set according to v7 documentation:
"If the size element is not used, KIWI will create an image containing approximately 30 % free space."

## Actual behaviour
Image without size element fails to build with error:
KiwiCommandError: rsync: stderr: rsync: write failed on "/tmp/kiwi_mount_manager.4qy_vkin/var/lib/rpm/Packages": No space left on device (28)

The same project builds properly in kiwi v7, but fails in v8.

    <preferences>
        <version>1.1.0</version>
        <bootsplash-theme>openSUSE</bootsplash-theme>
        <bootloader-theme>openSUSE</bootloader-theme>
        <type
            image="vmx"
            filesystem="xfs"
            boot="vmxboot/suse-leap42.2"
            format="qcow2"
            vga="normal"
            boottimeout="1"
            bootloader="grub2"
            firmware="uefi"
            kernelcmdline="plymouth.enable=0 console=tty1 console=ttyS0 net.ifnames=0"
            bootpartition="false"
            bootkernel="custom"
            devicepersistency="by-label">
        
        </type>
    </preferences>

davidcassany:
When no size is specified in type section the resulting image size is calculated using the size of the whole build-root increased by an empiric factor. Some tests revealed that this factor was not enough for XFS filesystems. The empiric factor has been increased.

This pull request fixes #186

### DIFF
--- a/kiwi/system/size.py
+++ b/kiwi/system/size.py
@@ -58,7 +58,7 @@ class SystemSize(object):
         elif requested_filesystem == 'btrfs':
             size *= 1.5
         elif requested_filesystem == 'xfs':
-            size *= 1.2
+            size *= 1.5
 
         return int(size)
 
--- a/test/unit/system_size_test.py
+++ b/test/unit/system_size_test.py
@@ -23,7 +23,7 @@ class TestSystemSize(object):
         assert self.size.customize(42, 'btrfs') == 63
 
     def test_customize_xfs(self):
-        assert self.size.customize(42, 'xfs') == 50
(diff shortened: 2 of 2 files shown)


schaefi:
ok :) I was struggling with this empiric factors since the beginning of the appliance builder. I think it never will be correct :)

PR #190, opened 2016-11-09
No need to mock Path.which

davidcassany:
This PR makes sure the iso unit tests can run in a platform where `isoinfo` is not present in PATH. Concretely, this is the case in an openSUSE platform that has no installation of `mkisofs` or `cdrkit-cdrtools-compat` packages. Note that none of those packages are KIWI requirements.

### DIFF
--- a/test/unit/iso_test.py
+++ b/test/unit/iso_test.py
@@ -200,9 +200,7 @@ class TestIso(object):
         assert result[2158].name == 'header_end'
 
 
-    @patch('kiwi.path.Path.which')
-    def test_create_header_end_block(self, mock_which):
-        mock_which.return_value = 'isoinfo'
+    def test_create_header_end_block(self):
         temp_file = NamedTemporaryFile()
         self.iso.header_end_file = temp_file.name
         assert self.iso.create_header_end_block(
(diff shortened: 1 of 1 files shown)


schaefi:
Don't get it, the mocking of Path.which is there to precisely define that we want to run this test under the assumption that 'isoinfo' was found. I'd like to prevent this test to depend on what the environment really provides

?

davidcassany:
@schaefi the problem here is that the isoinfo is actually called during the test. Mocking it this way we presume isoinfo is present and present in $PATH. In my host, currently, I neither have `mkisofs` or `cdrkit-cdrtools-compat` installed, also they are not a requirement of KIWI and also they are the package providing `/urs/bin/isoinfo`. So this that case the test fails trying to run the command `isoinfo`. Note that the code of `isols()` method tests `isoinfo` in PATH and also in an alternate location `/usr/lib/genisoimage/`, which, in fact, contains `isoinfo`. The point is that genisoimage contains provides isoinfo, but it is not included in path, and `genisoimage` is the actual kiwi requirement.

I don't know maybe we can find another way to work around the test, however I would not leave as is now, since it won't pass in a clean o JeOS like environment or OBS. Also I don't think it really makes sense to change requirements only for that.

schaefi:
ah, I see yes it is really called on a dummy iso I provided in the test data... you are right

PR #191, opened 2016-11-09
Do not delete all repositories if there aren't repositories to import

### LINKED ISSUE #182: Repository files added in by the user in root directory are deleted and ignored.
## Problem description

In OBS `imageinclude` attribute is not supported because the repositories that kiwi uses in an OBS process are translated to some temporary local ones that OBS creates. So another option would be to include repository files in `/etc/zypp/repos.d` within the user defined root folder, however it doesn't work because the content of this folder is always deleted in `import_repositories_marked_as_imageinclude`. 

I think `delete_all_repos()` shouldn't be called at that setup stage, as it can delete user custom configuration files. What about _cleaning_ the repositories right after the setup prepare stage?

davidcassany:
The proposed patch forces import_repositories_marked_as_imageinclude() to return immediately if there are no repositories to import. This way delete_all_repos() method is not called and any repository configuration file present in the default location is kept.

This PR fixes #182.

### DIFF
--- a/kiwi/package_manager/base.py
+++ b/kiwi/package_manager/base.py
@@ -15,7 +15,6 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
-from ..logger import log
 
 
 class PackageManagerBase(object):
@@ -183,8 +182,3 @@ class PackageManagerBase(object):
         del self.package_requests[:]
         del self.collection_requests[:]
         del self.product_requests[:]
-
-    def __del__(self):
-        log.info('Cleaning up all repos from %s instance', type(self).__name__)
-        if self.repository is not None:
-            self.repository.delete_all_repos()
--- a/kiwi/package_manager/base.py
+++ b/kiwi/package_manager/base.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+from ..logger import log
 
 
 class PackageManagerBase(object):
@@ -182,3 +183,8 @@ class PackageManagerBase(object):
         del self.package_requests[:]
         del self.collection_requests[:]
         del self.product_requests[:]
+
+    def __del__(self):
+        log.info('Cleaning up all repos from %s instance', type(self).__name__)
+        if self.repository is not None:
+            self.repository.delete_all_repos()
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -115,16 +115,7 @@ class SystemSetup(object):
         imageinclude attribute should be permanently added to
         the image repository configuration
         """
-        has_imageinclude = False
         repository_sections = self.xml_state.get_repository_sections()
-        for xml_repo in repository_sections:
-            if xml_repo.get_imageinclude():
-                has_imageinclude = True
-                break
-
-        if not has_imageinclude:
-            return
-
         root = RootInit(
             root_dir=self.root_dir, allow_existing=True
         )
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -115,7 +115,16 @@ class SystemSetup(object):
         imageinclude attribute should be permanently added to
         the image repository configuration
         """
+        has_imageinclude = False
         repository_sections = self.xml_state.get_repository_sections()
+        for xml_repo in repository_sections:
+            if xml_repo.get_imageinclude():
+                has_imageinclude = True
+                break
+
+        if not has_imageinclude:
+            return
+
         root = RootInit(
             root_dir=self.root_dir, allow_existing=True
         )
@@ -123,6 +132,7 @@ class SystemSetup(object):
             RootBind(root), self.xml_state.get_package_manager()
         )
         repo.use_default_location()
+        repo.delete_all_repos()
         for xml_repo in repository_sections:
             repo_marked_for_image_include = xml_repo.get_imageinclude()
 
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -123,7 +123,6 @@ class SystemSetup(object):
             RootBind(root), self.xml_state.get_package_manager()
         )
         repo.use_default_location()
-        repo.delete_all_repos()
         for xml_repo in repository_sections:
             repo_marked_for_image_include = xml_repo.get_imageinclude()
 
--- a/kiwi/tasks/system_build.py
+++ b/kiwi/tasks/system_build.py
@@ -189,8 +189,7 @@ class SystemBuildTask(CliTask):
         del manager
 
         # setup permanent image repositories after cleanup
-        repositories = self.xml_state.get_repository_sections()
-        if bool(list(xr for xr in repositories if xr.get_imageinclude())):
+        if self.xml_state.has_repositories_marked_as_imageinclude():
             setup.import_repositories_marked_as_imageinclude()
 
         setup.call_image_script()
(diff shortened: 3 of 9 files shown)


kiwi/package_manager/base.py
```diff
@@ -182,3 +183,8 @@ def cleanup_requests(self):
         del self.package_requests[:]
         del self.collection_requests[:]
         del self.product_requests[:]
+
+    def __del__(self):
+        log.info('Cleaning up all repos from %s instance', type(self).__name__)
+        if self.repository is not None:
```
isbm: I would suggest:
```python
if self.repository:
    self.repository.delete_all_repos()
```
Because `self.repository` supposed to be an object, specifig to repos. Although this is still pretty odd, because normally it would be:
```python
self.repository.delete()
```
...or:
```python
for repo in self.repositories:
    repo.delete()
```

test/unit/package_manager_base_test.py
```diff
@@ -14,6 +14,9 @@ def setup(self):
         repository.root_dir = 'root-dir'
         self.manager = PackageManagerBase(repository)
 
+    def teardown(self):
+        del self.manager
```
isbm: Actually, this should not be used. Usage of `del` is a code smell. Like very stinky smell, actually. This op should be used only preventing circular references, like in case of acessing traceback frame via variable and some very rare cases when typical GC is not enough (high-density memory calculations etc).

But in your case it is clearly code smell. I would revisit the design.

davidcassany: Yep agree that `del` should be avoided if possible. I will dig if I can manage a better test for that, in fact I expected that GC would do that for me in the test, maybe I can rearrange the scope.

kiwi/package_manager/base.py
davidcassany: Yes this Repository class that can handle multiple repositories used to confuse me a little also. I take your suggestion.

davidcassany:
I have been thinking a bit more about this and I think that maybe a good option to improve this instance cleaning of some classes (concretely PackageManager, SystemPrepare and SystemSetup) by encasulating them in a context manager, this way we could use them in a `with` statement and handle the cleaning in  a nicer way than the current `del` calls (as in `SystemBuildTask`). 

Note that those classes instances are only scoped inside system tasks (and boot/image/kiwi), so it would be quite a simple refactor.

@schaefi what you think? Does it makes sense to you?

schaefi:
I'm a bit concerned, most probably because I don't understand the problem.

The repository interface defines among others the three methods

* add_repo
* delete_repo
* delete_all_repos

- As they belong to a RepositorySomething class one could argue that the short form add, delete and   delete_all would have been enough, ok

- The delete_all just exists because repos are always a bunch of files in a directory and instead of itarating over the instances of RepositorySomething and call delete you have the delete_all joker on any of these instances. It's a nice to have thing to save two lines of iteration code... also questionable but I like it :)

With regards to the original issue, and I think this pull request should focus on getting that fixed first before mixing it with refactoring, I think your problem is that you can't add obs repos as imageinclude because we can't really resolve them into an external location

Actually the import_repositories_marked_as_imageinclude() method should only delete all repos if there are repos marked imageinclude in the image description. If no such repo definition exists it should not delete anything. This imho needs to be fixed.

However _if_ there are repos added and marked as imageinclude we want the repo setup to consist exclusively out of these entries and nothing else. Therefore the method calls the delete_all

When building in obs imageinclude doesn't make sense, so no repo should have that flag set

Let's fix import_repositories_marked_as_imageinclude() first

I don't want to have the repository destructor to delete all repos that is not the intended behavior

davidcassany:
@schaefi you are right, that refactor suggestion and the original issue are two different matters, I never thought to include such a change in this pull request, it just came to my mind while thinking how solve the initial issue.

At first I imagined a different behavior of import_repositories_marked_as_imageinclude() method (i.e append import_repositories_marked_as_imageinclude), but I also totally fine with your idea to only include those which are marked with `imageinclude`. Then if no repo has the flag the behavior would be to leave as is, without deleting anything. I'll do that.

Between, about the _repository_ vs _repositories_ comment I did not have the intention to modify anything, when i said _I take your suggestion_ I meant to avoid testing an object against `None` :)

davidcassany:
@scheafi I think this is what you suggested

schaefi:
yes that looks good to me :+1: 

maybe the nested python loop to check if any of the repos has an imageinclude attribute can be done differently, thus let's wait if @isbm has read it :)

kiwi/system/setup.py
```diff
@@ -115,14 +115,24 @@ def import_repositories_marked_as_imageinclude(self):
...
         """
+        has_imageinclude = False
         repository_sections = self.xml_state.get_repository_sections()
+        for xml_repo in repository_sections:
+            if xml_repo.get_imageinclude():
+                has_imageinclude = True
+                break
+
+        if not has_imageinclude:
+            return
```
isbm: Tip: This all can be written as one-liner:
```python
if not bool(next(xr for xr in self.xml_state.get_repository_sections() if xr.get_imageinclude())):
    return
```

But in general I am not fan of returning in the middle of the method. That means it is double-checking something inside that can be checked outside. I would rather check if this `has_imageinclude` and _only then_ would call this whole method _without_ double-checking things, assuming that the repos has included image.

davidcassany: Yes, I also prefer to test if the method needs to be executed or not outside.

davidcassany:
@isbm I think that now definitely looks better, thanks :wink:

kiwi/tasks/system_build.py
```diff
@@ -189,7 +189,9 @@ def process(self):
         del manager
 
         # setup permanent image repositories after cleanup
-        setup.import_repositories_marked_as_imageinclude()
+        repositories = self.xml_state.get_repository_sections()
+        if bool(list(xr for xr in repositories if xr.get_imageinclude())):
```
schaefi: wouldn't it be nice to have this as a method in xml_state.py e.g

```python
def get_repositories_marked_as_imageinclude(self):
    repositories = self.get_repository_sections()
    return bool(list(repo for repo in repositories if repo.get_imageinclude()))
```

* That's what xml_state is for, statefull information about the current description.
* We would also avoid duplicating the code as now done in system_prepare

davidcassany: @schaefi yes that's nice, but if we return a boolean I would just call it a bit different, maybe something like `has_repositories_marked_as_imageinclude`, as it does not return a list of anything.

schaefi: yes that's fine

davidcassany:
@schaefi I have added a new xml file for testing a description without any imageinclude. There are others (used in other tests) I could have used, but their names make clear references to other characteristics and tests, so that I preferred to use a new one with a more suited name.

schaefi:
this is great :+1: could squash the 5 commits into one and we can merge it :smile:

kiwi/tasks/system_build.py
```diff
@@ -190,7 +190,7 @@ def process(self):
 
         # setup permanent image repositories after cleanup
         repositories = self.xml_state.get_repository_sections()
-        if bool(next(xr for xr in repositories if xr.get_imageinclude())):
+        if bool(list(xr for xr in repositories if xr.get_imageinclude())):
```
isbm: This is not entirely correct. Your previous code was to break the loop as soon as there is anything. So I rewrote this into oneliner. Your version says "if there is built list that contains some elements". I.e. you loop until the very end. Also `next` returns only an iterator while `list` builds that fully in the memory. If you want go that way, than you could just:
```python
if [x for x in repositories if x.get_imageinclude()]:
   ...
```
But this is just not as efficient as my original example.

isbm: @davidcassany ^^^

PR #194, opened 2016-11-12
Move source of grub-snapshot.cfg to the right position

thkukuk:
grub-snapshot.cfg should not be sourced at the beginning, but
at the end of grub.cfg. Else with creating the first snapshot
this entry is getting the default boot target and the system
does not boot automatically anymore.

### DIFF
--- a/kiwi/bootloader/template/grub2.py
+++ b/kiwi/bootloader/template/grub2.py
@@ -32,7 +32,7 @@ class BootLoaderTemplateGrub2(object):
             search ${search_params}
             set default=${default_boot}
             set timeout=${boot_timeout}
-	    if [ -n "$$extra_cmdline" ]; then
+            if [ -n "$$extra_cmdline" ]; then
               submenu "Bootable snapshot $$snapshot_num" {
                 menuentry "If OK, run 'snapper rollback' and reboot." { true; }
               }
--- a/kiwi/bootloader/template/grub2.py
+++ b/kiwi/bootloader/template/grub2.py
@@ -30,9 +30,6 @@ class BootLoaderTemplateGrub2(object):
             search ${search_params}
             set default=${default_boot}
             set timeout=${boot_timeout}
-            if [ -f "/.snapshots/grub-snapshot.cfg" ]; then
-                source "/.snapshots/grub-snapshot.cfg"
-            fi
         ''').strip() + os.linesep
 
         self.header_hybrid = dedent('''
@@ -116,6 +113,13 @@ class BootLoaderTemplateGrub2(object):
                 set theme=($$root)/boot/grub2/themes/${theme}/theme.txt
             fi
         ''').strip() + os.linesep
+
+        self.menu_entry_boot_snapshots = dedent('''
+            if [ -f "/.snapshots/grub-snapshot.cfg" ]; then
+                source "/.snapshots/grub-snapshot.cfg"
+            fi
+        ''').strip() + os.linesep
+
 
         self.menu_entry_console_switch = dedent('''
             if [ "$${grub_platform}" = "efi" ]; then
@@ -291,6 +295,7 @@ class BootLoaderTemplateGrub2(object):
             template_data += self.menu_entry
             if failsafe:
                 template_data += self.menu_entry_failsafe
+        template_data += self.menu_entry_boot_snapshots
         if terminal == 'gfxterm':
             template_data += self.menu_entry_console_switch
         return Template(template_data)
@@ -318,6 +323,7 @@ class BootLoaderTemplateGrub2(object):
         template_data += self.menu_entry_multiboot
         if failsafe:
             template_data += self.menu_entry_failsafe_multiboot
+        template_data += self.menu_entry_boot_snapshots
         if terminal == 'gfxterm':
             template_data += self.menu_entry_console_switch
         return Template(template_data)
@@ -353,6 +359,7 @@ class BootLoaderTemplateGrub2(object):
             if failsafe:
                 template_data += self.menu_entry_failsafe
         template_data += self.menu_iso_harddisk_entry
+        template_data += self.menu_entry_boot_snapshots
         if terminal == 'gfxterm':
             template_data += self.menu_entry_console_switch
         return Template(template_data)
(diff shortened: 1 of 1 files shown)


kiwi/bootloader/template/grub2.py
```diff
@@ -291,6 +295,7 @@ def get_disk_template(
             template_data += self.menu_entry
             if failsafe:
                 template_data += self.menu_entry_failsafe
+        template_data += self.menu_entry_boot_snapshots
```
isbm: When somebody is doing `+=` on Python strings and it is over templates, then I am like:
[Please do not](https://www.youtube.com/watch?v=JncgoPKklVE)
https://www.youtube.com/watch?v=JncgoPKklVE

thkukuk:
On Sat, Nov 12, Bo Maryniuk wrote:

> @isbm requested changes on this pull request.
> 
> Please do not use +=. No. Noooo. :-)

The whole code is doing that. Why should this single
case be different?
But feel free to rewrite the whole code to not doing it.


-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CASP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

isbm:
@thkukuk This is a bad idea and continue that is also bad idea.

thkukuk:
On Sat, Nov 12, Bo Maryniuk wrote:

> @thkukuk This is a bad idea and continue that is also bad idea.

Sorry, but this is no argument.

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CASP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

isbm:
@thkukuk String concatenation [should be avoided](https://wiki.python.org/moin/PythonSpeed/PerformanceTips#String_Concatenation). Period. As of your argument "I don't care fixing wrong, because the whole code doing wrong", is something that is foreign to me.

@schaefi you decided merge this or not.

thkukuk:
On Sat, Nov 12, Bo Maryniuk wrote:

> @thkukuk String concatenation should be avoided. Period. As of your argument "I
> don't care fixing wrong, because the whole code doing wrong", is something that
> is foreign to me.

In all upstream projects I'm involved in, you have extactly one coding
style. If a patch, which uses another coding style, even if this is
"more correct", is submitted, this is a direct reject reason.
I'm really surprised that this is foreign to you. Doesn't speak
for good code readability.

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CASP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

isbm:
It is not about _style_. It is about how Python interprets the strings in memory. Code readability has nothing to do here. Besides, we here had already these discussions and decided to remove `+=` step-by-step.

I just wrote my opinion on your PR and it is up to you fix it or not.

schaefi:
The reason why I have used += in the templates was basically lack of knowledge. Bo told us recently not to do that including a good explanation why we should change it. I agreed to that and we will refactor those places step by step.

However in this particular case I would first go with Thorsten's implementation and after that refactor the complete template in an extra pull request. That's because the template class is 99% string operations and serves as a good candidate to become refactored in an extra effort

What's really missing in this pull request is the adaption of the test. @thkukuk in this version of kiwi we apply the 100% test coverage rule which is the reason why the tests currently fail. Could you adapt the test too ?

Thanks

PS: I'm currently ill and need some rest, thus I'll come back to this one when I'm back at work, probably Tuesday.... back in bed

thkukuk:
On Sun, Nov 13, Marcus Schäfer wrote:

> What's really missing in this pull request is the adaption of the test.
> @thkukuk in this version of kiwi we apply the 100% test coverage rule which is
> the reason why the tests currently fail. Could you adapt the test too ?

Which test? I cannot find any test containing grub-snapshot.cfg.
And "make test" aborts with an error even for the unpatched kiwi
version:
tox.ConfigError: ConfigError: substitution key '3.4,check,doc,doc_travis,doc_travis_deploy' not found

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CASP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

schaefi:
Well sorry not the test but python flake8 code style checks fails with

```
/home/travis/build/SUSE/kiwi/kiwi/bootloader/template/grub2.py:35:1: W191 indentation contains tabs
/home/travis/build/SUSE/kiwi/kiwi/bootloader/template/grub2.py:35:1: E101 indentation contains mixed spaces and tabs
/home/travis/build/SUSE/kiwi/kiwi/bootloader/template/grub2.py:36:1: E101 indentation contains mixed spaces and tabs
2     E101 indentation contains mixed spaces and tabs
1     W191 indentation contains tabs
```

An easy way to setup a development environment from which you can run all tests and checks is to use the standard python virtual environment. I have documented all steps here:

* http://suse.github.io/kiwi/contributing.html

aplanas:
@isbm Can I ask about the problem with `+=` in Python with `str` objects? I found it misleading, because in some languages can be inferred that the original object is modified, when in reality a new string is constructed (`str` is immutable)

Before Python 2.5 string concatenation can show some speed issue, but this was fixed 10 years ago[1]. As today this operation is really fast:

```
> python -m timeit 'a="a"*1000; a+="extra"'
1000000 loops, best of 3: 0.2 usec per loop

> python -m timeit 'a="a"*1000; a=a+"extra"'
1000000 loops, best of 3: 0.2 usec per loop

> python -m timeit 'a="a"*1000; a="%s%s"%(a,"extra")'
1000000 loops, best of 3: 0.495 usec per loop

> python -m timeit 'a="a"*1000; a="{}{}".format(a,"extra")'
1000000 loops, best of 3: 0.55 usec per loop
```

In all the examples a new string object is build, but `+=` is the fasted one (together with the unsugarized version `a=a+b`)

But I agree that if we are concatenating tons of string, `''.join()` is better, or `template.format(**kwargs)`, in performance and in clarity. But this is not the case AFAIK.

(Just asking, I am personally +1 in favor of using `join`, `%` or `format`)

[1] https://www.python.org/download/releases/2.5/highlights/ (Faster section)

PR #196, opened 2016-11-16
vhdfixed images compression

davidcassany:
This pull request is related to #184 and #187. With these
changes the bundle procedure for vhdfixed images produces
compressed files, as it used to do in previous kiwi versions. 
This way the current kiwi version is still compatible with other
related services.


schaefi:
:+1: thanks

PR #197, opened 2016-11-18
Fixed kernel name lookup

schaefi:
Complete the list of kernel names for the lookup. Normally
the kernel package provides a symlink to the actual kernel
image file. However if the link does not exist we extend the
search to a collection of names for possible kernel images.
The new list now also covers kernel names as used for arm

Along with the change this patch also provides an option to
raise an exception if the kernel lookup did not find any
kernel, which is used for the get_kernel() request in the
dracut initrd system setup where it is mandatory to find
a kernel image.

This fixes bnc#1010874

### DIFF
--- a/kiwi/boot/image/dracut.py
+++ b/kiwi/boot/image/dracut.py
@@ -46,9 +46,7 @@ class BootImageDracut(BootImageBase):
         if self.is_prepared():
             log.info('Creating generic dracut initrd archive')
             kernel_info = Kernel(self.boot_root_directory)
-            kernel_details = kernel_info.get_kernel(
-                raise_on_not_found=True
-            )
+            kernel_details = kernel_info.get_kernel(raise_on_not_found=True)
             Command.run(
                 [
                     'chroot', self.boot_root_directory,
--- a/kiwi/boot/image/dracut.py
+++ b/kiwi/boot/image/dracut.py
@@ -46,7 +46,9 @@ class BootImageDracut(BootImageBase):
         if self.is_prepared():
             log.info('Creating generic dracut initrd archive')
             kernel_info = Kernel(self.boot_root_directory)
-            kernel_version = kernel_info.get_kernel().version
+            kernel_details = kernel_info.get_kernel(
+                raise_on_not_found=True
+            )
             Command.run(
                 [
                     'chroot', self.boot_root_directory,
@@ -55,7 +57,7 @@ class BootImageDracut(BootImageBase):
                     '--no-hostonly-cmdline',
                     '--no-compress',
                     os.path.basename(self.initrd_file_name),
-                    kernel_version
+                    kernel_details.version
                 ]
             )
             Command.run(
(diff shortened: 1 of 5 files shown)


kiwi/boot/image/dracut.py
```diff
@@ -46,7 +46,9 @@ def create_initrd(self, mbrid=None):
         if self.is_prepared():
             log.info('Creating generic dracut initrd archive')
             kernel_info = Kernel(self.boot_root_directory)
-            kernel_version = kernel_info.get_kernel().version
+            kernel_details = kernel_info.get_kernel(
+                raise_on_not_found=True
+            )
```
isbm: That usually _has_ to be one line, unless multi-params. Otherwise not really "pythonic" code.

kiwi/system/kernel.py
```diff
@@ -21,6 +21,8 @@
 # project
 from ..command import Command
 
+from ..exceptions import KiwiKernelLookupError
```
isbm: I am honestly still not fan of those relative imports. It somehow needs to be done like:
```python
from full.path.to.the.module import myfunction, myclass
```
But this is a next PR, I guess.

kiwi/system/kernel.py
```diff
@@ -127,3 +129,30 @@ def copy_xen_hypervisor(self, target_dir, file_name=None):
...
+                # not found, then lookup the real kernel image names
+                # depending on the arch and os they are different
+                ''.join(['uImage-', kernel_dir]),
+                ''.join(['Image-', kernel_dir]),
+                ''.join(['zImage-', kernel_dir]),
+                ''.join(['vmlinuz-', kernel_dir, '.gz']),
+                ''.join(['vmlinux-', kernel_dir]),
+                ''.join(['image-', kernel_dir]),
+                ''.join(['vmlinuz-', kernel_dir])
+            ]
```
isbm: Why "for loop" if it will work only once? — the `return` will kick first in line and that's it. Additionally, we could make it not like Assembler programming :smile: :
```python
patt = '{prefix}-{kdir}'
kdir = os.listdir(''.join([self.root_dir, '/lib/modules']))
out = ['vmlinux', 'vmlinuz', 'zImage']
if kdir:
    for prefix in ['uImage', 'Image', 'zImage', 'vmlinuz', 'vmlinux', 'image']:
        out.append(patt.format(prefix=prefix, kdir=kdir[0]))  # Same functionality as above
        if prefix == 'vmlinuz':
            out.append(out[-1] + '.gz')  # Repeat it again with the ".gz" suffix

return out
```

Not tested, but something like this.

kiwi/system/kernel.py
schaefi: yes, what's interesting here is that I had that in the early days of this code when another python Guru told me it should be better changed to relative imports. Apart from personal taste is there also a technical reason why relative imports sucks ? Thanks

kiwi/system/kernel.py
schaefi: :) all right let's de-assemble it a bit

schaefi:
I intentionally left out the ".gz" name, actually that should never be a valid kernel image name and was only a nasty workaround for a kernel packaging bug on arm in the past

schaefi:
guess we are ok to merge this, casp team is waiting for the fix

PR #203, opened 2016-11-24
Parse arch attribute as a comma separated list

### LINKED ISSUE #202: packages/arch: comma seperated list not working
## Problem description
https://doc.opensuse.org/projects/kiwi/doc/ states in 4.1.8.2:
"To restrict a package to a specific architecture, use the arch attribute to specify a comma separated list of allowed architectures. "
## Actual behaviour
If I use a comma separated list of architectures for a package, it will be installed on no architecture.

Create an entry like "<package name="XX" arch="x86_64,aarch64". The package will not installed on any architecture.

davidcassany:
This commit ensure that the arch attribute of each package section
is parsed as a comma separated list. This way, as in previous kiwi
versions, each package might be suitable for one or more specified
architectures.

Fixes #202

### DIFF
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -151,6 +151,23 @@ class XMLState(object):
                 result.append(packages)
         return result
 
+    def package_matches_host_architecture(self, package):
+        """
+        Tests if the given package is applicable for the current host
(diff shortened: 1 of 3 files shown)


schaefi:
:+1:

PR #206, opened 2016-11-30
LVM volumes management during grub2 install

### LINKED ISSUE #205: Grub2 install fails when LVM volumes with nested mountpoints are defined
## Problem description

Grub2 install procedure fails when attempting to mount the nested volumes of the resulting VMX image.

## Actual behaviour

KIWI tries to mount all volumes from the loop device corresponding the disk image, however it fails to mount a volume that has a nested mountpoint with another volume.

Use this ```systemdisk``` configuration:
```xml
<systemdisk name="vg0">
  <volume name="@root" size="4G"/>
  <volume name="LVhome" size="1G" mountpoint="home"/>
  <volume name="LVtmp" size="2G" mountpoint="tmp"/>
  <volume name="LVopt" size="1G" mountpoint="opt"/>
  <volume name="LVvar" size="2G" mountpoint="var"/>
  <volume name="LVlog" size="2G" mountpoint="var/log"/>
</systemdisk>
```

davidcassany:
Grub installation procedure have been updated to take into account
LVM volumes mountpoint and ensure they are mounted with a
coherent order.

Fixes #205

### DIFF
--- a/kiwi/bootloader/install/grub2.py
+++ b/kiwi/bootloader/install/grub2.py
@@ -24,6 +24,7 @@ from ...command import Command
 from ...logger import log
 from ...defaults import Defaults
 from ...mount_manager import MountManager
+from ...path import Path
 
 from ...exceptions import (
     KiwiBootLoaderGrubInstallError,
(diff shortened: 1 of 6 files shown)


schaefi:
looks good to me 👍

PR #210, opened 2016-12-02
Including minimum volume size offset in disk size calculation

### LINKED ISSUE #207: Storage calculation is wrong for LVM volumes and VMX images.
## Problem description
When creating a VMX with LVM custom partitions of specific size the virtual disk that is created to hold the system is not big enough. When creating the volumes KIWI fails due to lack of space.

## Actual behaviour
`DiskSetup.get_disksize_mbytes()` method for VMX images makes a sum of all defined volume sizes and creates a raw disk with this exact value, however there are a couple of overheads that are not taken into consideration. First, modern `fdisk` implementations assume the first partition to start at sector 2048 (1MB with 512 bytes per sector); second, LVM metadata causes an offset of one Physical Extent (4MB by default in current `lvm2` implementations). Given the fact that volumes are aligned at PEs boundaries it would be enough to include one extra PE to hold both, the `fdisk` offest and LVM metadata offset.

Use a systemdisk configuration like the following in a VMX image:

```xml
<systemdisk name="vg0">
  <volume name="@root" size="4G"/>
  <volume name="LVhome" size="1G" mountpoint="home"/>
  <volume name="LVtmp" size="1G" mountpoint="tmp"/>
</systemdisk>
```

davidcassany:
Adding an extra space on disk calculation in order to hold `fdisk` offset, LVM metadata or any other kind of data that is needed for volume and partitions handling which not part of the partition's or volume's size.

Fixes #207

### DIFF
--- a/kiwi/storage/setup.py
+++ b/kiwi/storage/setup.py
@@ -312,13 +312,13 @@ class DiskSetup(object):
                 [size_type, req_size] = volume.size.split(':')
                 disk_add_mbytes = 0
                 if size_type == 'freespace':
-                    disk_add_mbytes += int(req_size) + \
-                        Defaults.get_min_volume_mbytes()
+                    disk_add_mbytes += int(req_size)
                 else:
                     disk_add_mbytes += int(req_size) - \
                         data_volume_mbytes.volume[volume.realpath]
                 if disk_add_mbytes > 0:
-                    disk_volume_mbytes += disk_add_mbytes
+                    disk_volume_mbytes += disk_add_mbytes + \
+                        Defaults.get_min_volume_mbytes()
                 else:
                     log.warning(
                         'volume size of %s MB for %s is too small, skipped',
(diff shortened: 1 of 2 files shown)


kiwi/storage/setup.py
```diff
@@ -312,13 +312,13 @@ def _accumulate_volume_size(self, root_mbytes):
                 [size_type, req_size] = volume.size.split(':')
                 disk_add_mbytes = 0
                 if size_type == 'freespace':
-                    disk_add_mbytes += int(req_size) + \
-                        Defaults.get_min_volume_mbytes()
+                    disk_add_mbytes += int(req_size)
```
schaefi: why did you remove the additional min bytes from the freespace setup ? The current use in bytes plus the requested free size will not lead the requested size being really free due to the volume overhead. I rather have a little bit more free than less than requested. Thoughts ?

davidcassany:
@schaefi I just saw I left my comments of today morning in pending state 😢 and now one appears duplicated. 
Still confused with the review feature and having an extra submit step...

schaefi:
this review thing tricks us all 😃

PR #215, opened 2017-01-09
Docker type attributes

schaefi:
Update schema for container setup
    
In preparation to support creation of native docker containers the way docker likes it, the kiwi schema has to be adapted. So far there  was no way to specify additional metadata information for containers.
With this commit a new section called containerconfig is provided.  containerconfig is a subsection of the type and only allowed for  the docker image type. This constraint is checked via a schematron rule. In addition an xslt stylesheet moves the currently existing  container attribute, which was used to set a name for the container,  into the new containerconfig section as name attribute

An example type definition for docker would look like this

```xml
<type image="docker">
    <containerconfig name="foo" entry_point="xxx"/>
</type>
```

### DIFF
--- a/doc/source/schema.rst
+++ b/doc/source/schema.rst
@@ -271,7 +271,7 @@ Parents:
    These elements contain ``type``: :ref:`k.image.preferences`
 
 Children:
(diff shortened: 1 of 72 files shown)


schaefi:
Please be aware many of the files changed here are auto generated code

schaefi:
Review is appreciated for:

* kiwi/builder/container.py
* kiwi/schema/kiwi.rnc
* kiwi/xml_state.py
* kiwi/xsl/convert64to65.xsl
* kiwi/xsl/master.xsl
* test/unit/*

Thanks

schaefi:
[complete creation of docker base images](https://trello.com/c/MBG8pM4L/199-complete-creation-of-docker-base-images)

PR #216, opened 2017-01-09
Refactor docker container creation

schaefi:
Instead of creating a simple tarball the tools umoci and skopeo from the docker tool chain are used to create official docker images. The way to specify a docker image type has changed as follows

```xml
<type image="docker">
        <containerconfig name="..." maintainer="..." user="..." workingdir="...">
            <entrypoint execute="command">
                <argument name="option"/>
                ...
            </entrypoint>
            <subcommand execute="command">
                <argument name="option"/>
                ...
            </subcommand>
            <expose>
                <port number="..."/>
                ...
            </expose>
            <volumes>
                <volume name="..."/>
                ...
            </volumes>
            <environment>
                <env name="variable" value="value"/>
                ...
            </environment>
            <labels>
                <label name="..." value="..."/>
                ...
            </labels>
        </containerconfig>
    </type>
```

### DIFF
--- a/kiwi/container/__init__.py
+++ b/kiwi/container/__init__.py
@@ -30,9 +30,9 @@ class ContainerImage(object):
     :param string name: container system name
     :param string root_dir: root directory path name
     """
-    def __new__(self, name, root_dir):
+    def __new__(self, name, root_dir, custom_args=None):
         if name == 'docker':
-            return ContainerImageDocker(root_dir)
+            return ContainerImageDocker(root_dir, custom_args)
         else:
             raise KiwiContainerImageSetupError(
                 'Support for %s container not implemented' % name
--- a/kiwi/container/docker.py
+++ b/kiwi/container/docker.py
@@ -15,9 +15,15 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+import os
+
 # project
-from ..archive.tar import ArchiveTar
 from ..defaults import Defaults
+from ..path import Path
+from ..command import Command
+from tempfile import mkdtemp
+from ..utils.sync import DataSync
+from ..utils.compress import Compress
 
 
 class ContainerImageDocker(object):
@@ -28,9 +34,60 @@ class ContainerImageDocker(object):
 
     * :attr:`root_dir`
         root directory path name
+
+    * :attr:`custom_args`
+        representation of the containerconfig and its subsections
     """
-    def __init__(self, root_dir):
+    def __init__(self, root_dir, custom_args=None):
         self.root_dir = root_dir
+        self.docker_dir = None
+        self.docker_root_dir = None
+
+        self.container_name = []
+        self.container_tag = 'latest'
+        self.entry_command = ['--config.cmd=/bin/bash']
+        self.entry_subcommand = []
+        self.maintainer = []
+        self.user = []
+        self.workingdir = []
+        self.expose_ports = []
+        self.volumes = []
+        self.environment = []
+        self.labels = []
+
+        if custom_args:
+            if 'container_name' in custom_args:
+                self.container_name = custom_args['container_name']
+
+            if 'container_tag' in custom_args:
+                self.container_tag = custom_args['container_tag']
+
+            if 'entry_command' in custom_args:
+                self.entry_command = custom_args['entry_command']
+
+            if 'entry_subcommand' in custom_args:
+                self.entry_subcommand = custom_args['entry_subcommand']
+
+            if 'maintainer' in custom_args:
+                self.maintainer = custom_args['maintainer']
+
+            if 'user' in custom_args:
+                self.user = custom_args['user']
+
+            if 'workingdir' in custom_args:
+                self.workingdir = custom_args['workingdir']
+
+            if 'expose_ports' in custom_args:
+                self.expose_ports = custom_args['expose_ports']
+
+            if 'volumes' in custom_args:
+                self.volumes = custom_args['volumes']
+
+            if 'environment' in custom_args:
+                self.environment = custom_args['environment']
+
+            if 'labels' in custom_args:
+                self.labels = custom_args['labels']
 
     def create(self, filename):
         """
@@ -39,14 +96,82 @@ class ContainerImageDocker(object):
         :param string filename: archive file name
         """
         exclude_list = [
-            'image', '.profile', '.kconfig', 'boot',
+            'image', '.profile', '.kconfig', 'boot', 'dev', 'sys', 'proc',
             Defaults.get_shared_cache_location()
         ]
-        # replace potential suffix from filename because
-        # it is added by the archive creation call
-        archive = ArchiveTar(
-            filename.replace('.xz', '')
+
+        self.docker_dir = mkdtemp(prefix='kiwi_docker_dir.')
+        self.docker_root_dir = mkdtemp(prefix='kiwi_docker_root_dir.')
+
+        container_dir = os.sep.join(
+            [self.docker_dir, 'umoci_layout']
         )
-        archive.create_xz_compressed(
-            source_dir=self.root_dir, exclude=exclude_list
+        container_name = ':'.join(
+            [container_dir, self.container_name]
         )
+
+        Command.run(
+            ['umoci', 'init', '--layout', container_dir]
+        )
+        Command.run(
+            ['umoci', 'new', '--image', container_name]
+        )
+        Command.run(
+            ['umoci', 'unpack', '--image', container_name, self.docker_root_dir]
+        )
+        docker_root = DataSync(
+            ''.join([self.root_dir, os.sep]),
+            os.sep.join([self.docker_root_dir, 'rootfs'])
+        )
+        docker_root.sync_data(
+            options=['-a', '-H', '-X', '-A'], exclude=exclude_list
+        )
+        Command.run(
+            ['umoci', 'repack', '--image', container_name, self.docker_root_dir]
+        )
+        Command.run(
+            [
+                'umoci', 'config'
+            ] +
+            self.maintainer +
+            self.user +
+            self.workingdir +
+            self.entry_command +
+            self.entry_subcommand +
+            self.expose_ports +
+            self.volumes +
+            self.environment +
+            self.labels +
+            [
+                '--image', container_name,
+                '--tag', self.container_tag
+            ]
+        )
+        Command.run(
+            ['umoci', 'gc', '--layout', container_dir]
+        )
+
+        docker_tarfile = filename.replace('.xz', '')
+
+        # make sure the target tar file does not exist
+        # skopeo doesn't support force overwrite
+        Path.wipe(docker_tarfile)
+
+        Command.run(
+            [
+                'skopeo', 'copy', 'oci:{0}'.format(
+                    container_name
+                ),
+                'docker-archive:{0}:{1}:{2}'.format(
+                    docker_tarfile, self.container_name, self.container_tag
+                )
+            ]
+        )
+        compressor = Compress(docker_tarfile)
+        compressor.xz()
+
+    def __del__(self):
+        if self.docker_dir:
+            Path.wipe(self.docker_dir)
+        if self.docker_root_dir:
+            Path.wipe(self.docker_root_dir)
--- a/doc/source/schema.rst
+++ b/doc/source/schema.rst
@@ -341,10 +341,196 @@ Provides metadata information for containers
 Parents:
    These elements contain ``containerconfig``: :ref:`k.image.preferences.type`
 
+Children:
+   The following elements occur in ``containerconfig``: :ref:`entrypoint <k.image.preferences.type.containerconfig.entrypoint>` `[?]`_, :ref:`subcommand <k.image.preferences.type.containerconfig.subcommand>` `[?]`_, :ref:`expose <k.image.preferences.type.containerconfig.expose>` `[?]`_, :ref:`volumes <k.image.preferences.type.containerconfig.volumes>` `[?]`_, :ref:`environment <k.image.preferences.type.containerconfig.environment>` `[?]`_, :ref:`labels <k.image.preferences.type.containerconfig.labels>` `[?]`_
+
 List of attributes for ``containerconfig``:
 
-* ``name`` : Specifies a name for the container. This is usually the the tag name of the container as read if the container image is imported via the docker load command
-* ``entry_command`` `[?]`_: Specifies the default command to run if the container is called via the docker run command.
+* ``name`` : Specifies a name for the container. This is usually the the repository name of the container as read if the container image is imported via the docker load command
+* ``tag`` `[?]`_: Specifies a tag for the container. This is usually the the tag name of the container as read if the container image is imported via the docker load command
+* ``maintainer`` `[?]`_: Specifies a maintainer for the container.
+* ``user`` `[?]`_: Specifies a user for the container.
+* ``workingdir`` `[?]`_: Specifies the default working directory of the container
+
+.. _k.image.preferences.type.containerconfig.entrypoint:
+
+entrypoint
+::::::::::
+
+Provides details for the entry point command. This includes the execution name and its parameters. Arguments can be optionally specified
+
+Parents:
+   These elements contain ``entrypoint``: :ref:`k.image.preferences.type.containerconfig`
+
+Children:
+   The following elements occur in ``entrypoint``: :ref:`argument <k.image.preferences.type.containerconfig.entrypoint.argument>` `[*]`_
+
+List of attributes for ``entrypoint``:
+
+* ``execute`` : Specifies the entry point program name to execute
+
+.. _k.image.preferences.type.containerconfig.entrypoint.argument:
+
+argument
+;;;;;;;;
+
+Provides details about a command argument
+
+Parents:
+   These elements contain ``argument``: :ref:`k.image.preferences.type.containerconfig.entrypoint`, :ref:`k.image.preferences.type.containerconfig.subcommand`
+
+List of attributes for ``argument``:
+
+* ``name`` : Specifies a command argument name
+
+.. _k.image.preferences.type.containerconfig.subcommand:
+
+subcommand
+::::::::::
+
+Provides details for the subcommand command. This includes the execution name and its parameters. Arguments can be optionally specified. The subcommand is appended the command information from the entrypoint. It is in the responsibility of the author to make sure the combination of entrypoint and subcommand forms a valid execution command
+
+Parents:
+   These elements contain ``subcommand``: :ref:`k.image.preferences.type.containerconfig`
+
+Children:
+   The following elements occur in ``subcommand``: :ref:`argument <k.image.preferences.type.containerconfig.subcommand.argument>` `[*]`_
+
+List of attributes for ``subcommand``:
+
+* ``execute`` : Specifies the subcommand program name to execute
+
+.. _k.image.preferences.type.containerconfig.subcommand.argument:
+
+argument
+;;;;;;;;
+
+Provides details about a command argument
+
+Parents:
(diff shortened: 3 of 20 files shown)


schaefi:
[complete creation of docker base images](https://trello.com/c/MBG8pM4L/199-complete-creation-of-docker-base-images)

schaefi:
__NOTE:__ this can be merged only if skopeo and umoci tools are present in Factory and SLE12

kiwi/container/__init__.py
```diff
@@ -30,9 +30,9 @@ class ContainerImage(object):
     :param string name: container system name
     :param string root_dir: root directory path name
     """
-    def __new__(self, name, root_dir):
+    def __new__(self, name, root_dir, custom_args=None):
```
davidcassany: `custom_args` are not used in the constructor so they are always `None,` see `ContainerBuilder.create` method. I guess a map of all `containerconfig` attributes should be included as `custom_args.`

schaefi: This was intentional as long as I was not 100% sure which parameters we need for umoci and skopeo. But I will adapt it. Thanks

schaefi: ok done, see kiwi/builder/container.py, there all the data gets passed. I will now add the runtime check for umoci and skopeo so that we don't need to add them as requires in the spec file for the first shot

schaefi: runtime check added

schaefi:
This is now on hold until skopeo has the patch required and used by our code.
@davidcassany just let me know if all that makes sense so far

Thanks

schaefi:
For reference here the error message from skopeo

```
[ ERROR   ]: 16:21:58 | KiwiCommandError: skopeo: stderr: time="2017-01-24T16:21:58+01:00" level=fatal msg="Invalid destination name docker-archive:/tmp/mytest//LimeJeOS-Leap-42.2.x86_64-1.42.2.docker.tar:/tmp/kiwi_docker_dir.8joqz0qn/umoci_layout:latest: Invalid image name \"docker-archive:/tmp/mytest//LimeJeOS-Leap-42.2.x86_64-1.42.2.docker.tar:/tmp/kiwi_docker_dir.8joqz0qn/umoci_layout:latest\", unknown transport \"docker-archive\"" 
```

schaefi:
great with the updated skopeo all works fine, tested the image with a docker load and all went good

schaefi:
@davidcassany could you complete your review ? if there aren't any issues from your side it would be nice if you can give your ok ;)

After that I'll just wait until the skopeo patch goes upstream

We are getting closer

davidcassany:
@schaefi so good it worked for you. However it did not for me, we just found an issue in umoci with @cyphar, so we can't go on until fixed 😞 There is a problem with UTF-8 characters handling, go-mtree relies on some old BSD code that does no properly handle special characters. See opensuse/umoci#80 In short, files named with special characters make `umoci repack` fail,  for example, `ca-certificates` package includes files with non ascii names which produce such a failure.

schaefi:
oh interesting, looks like I was lucky not stumbling over that one

schaefi:
nevertheless you can complete your review regarding this code

davidcassany:
@schaefi sure you are right. There are some issues about how parameters are passed to the umoci tool in the `umoci config` subcommand. Specially the ones that are considered an array of strings. We already have a list of the configuration parameters umoci supports that should be configurable from the xml file. I have been checking how they are mapped to the umoci command line tool. Below I include an example of how a complete (including all options) containerconfig tag could look like (obviously we could agree on different ways of representing it, I am totally opened to that):

```xml
<containerconfig name="mynewcontainer" maintainer="tux" user="root" workingdir="/root">
  <!-- maps to: --author="tux" --config.user="root" --config.workingdir="/root" -->
  <containercmd> <!-- maps to: --config.cmd='/bin/bash' --config.cmd='somescript' ...-->
    <cmd>/bin/bash</cmd>
    <cmd>somescript</cmd>
    <cmd>scriptargument</cmd>
  </containercmd>
  <containerentrypoint> <!-- maps to: --config.entrypoint='/bin/bash' --config.entrypoint... -->
    <entrypoint>/bin/bash</entrypoint>
    <entrypoint>somescript</entrypoint>
    <entrypoint>scriptargument</entrypoint>
  </containerentrypoint>
  <containerexpose> <!-- maps to: --config.exposedports="80" --config.exposedports="8080" -->
    <port>80</port>
    <port>8080</port>
  </containerexpose>
  <containervolumes> <!-- maps to: --config.volume="/var/log" --config.volume="/tmp" -->
    <volume>/tmp</volume>
    <volume>/var/log</volume>
  </containervolumes>
  <containerenvs> <!-- maps to: --config.env="PATH=/bin:/usr/bin:/home/user/bin" -->
    <envvar name="PATH" value="/bin:/usr/bin:/home/user/bin"/>
    <envvar name="SOMEVAR" value="somevalue"/>
  </containerenvs>
  <containerlabel> <!-- maps to: --config.label="somelabel=labelvalue" -->
    <label name="somelabel" value="labelvalue"/>
    <label name="someotherlabel" value="anotherlabelvalue"/>
  </containerlabel>
</containerconfig>
```

kiwi/container/docker.py
```diff
@@ -28,9 +34,26 @@ class ContainerImageDocker(object):
...
+        self.docker_dir = None
+        self.docker_root_dir = None
+
+        if custom_args and 'container_name' in custom_args:
+            self.container_name = custom_args['container_name']
+        else:
+            self.container_name = 'latest'
+
+        if custom_args and 'entry_command' in custom_args:
+            self.entry_command = custom_args['entry_command']
```
davidcassany: note that entry_command may not be the best variable name since umoci supports `entry_point` and `cmd` configuration tags. I would go for direct mapping, lets say 'self.entry_point' and `self.cmd`.

kiwi/container/docker.py
```diff
@@ -39,14 +62,65 @@ def create(self, filename):
...
+        docker_root.sync_data(
+            options=['-a', '-H', '-X', '-A'], exclude=exclude_list
+        )
+        Command.run(
+            ['umoci', 'repack', '--image', container_tag, self.docker_root_dir]
+        )
+        Command.run(
+            [
+                'umoci', 'config',
+                '='.join(['--config.cmd', self.entry_command]),
```
davidcassany: This is wrong, since there could be multiple `entry_points` and umoci does not understand the space as a parameter separator, some like `command arg1 arg2` should be passed three different `--config.cmd` parameters: `--config.cmd='command' --config.cmd='arg1' --config.cmd='arg2'`. Thats the main reason I suggested to go for a nested structure in the XML, this way looks to me easier to loop over multiple values.

schaefi: That's a very strange handling for command parameters, anyway we have to cope with it

schaefi:
Thanks for the details, I'll extend the schema and let you know the result

schaefi:
Can you explain why we distinguish between 

```xml
<containercmd>
<containerentrypoint>
```

looks the same to me, what is the difference ?

schaefi:
given containercmd and containerentrypoint is the same set of information I suggest the following XML structure

```xml
<containerconfig name="mynewcontainer" maintainer="tux" user="root" workingdir="/root">
    <entrypoint>
        <command name="/bin/bash">
            <argument name="arg1"/>
            <argument name="arg2"/>
        </command>
    </entrypoint>

    <expose>
        <port number="80"/>
        <port number="8080"/>
    </expose>

    <volumes> 
        <volume name="/tmp"/>
        <volume name="/var/log"/>
    </volumes>

    <environment> 
        <env name="PATH" value="/bin:/usr/bin:/home/user/bin"/>
        <env name="SOMEVAR" value="somevalue"/>
    </environment>

    <label> 
        <label name="somelabel" value="labelvalue"/>
        <label name="someotherlabel" value="anotherlabelvalue"/>
    </label>
</containerconfig>

Thoughts ?
```

kiwi/container/docker.py
davidcassany: Yes this is quite strange to my eyes also. I assume it is the same as using the CMD statement of the Dockerfile using the brackets syntax (i.e CMD ['/usr/bin/cowsay', 'have', 'fun']), in that case every parameter is passed to the `exec` function instead to a shell.

davidcassany:
To my eyes there is also no real difference between `entry_point` and `cmd`, apart of some minor features to handle them form the docker run command line. However since docker supports both and it commonly uses both I guess it is expected to support both. @cyphar, @flavio do you have any thoughts about that topic, is it a real need to support both? Is there a real difference?

davidcassany:
@schaefi I definitely like your proposal of XML, I think  we just have to agree how use entry_point and cmd. I would not have a child tag called `<command>` inside `entry_point`, just not to get confused with the Dockerfile syntax where `CMD` and `ENTRY_POINT` are two different statements.

schaefi:
agreed, Flavio should clarify if we have to differentiate and once we know that we can find an expressive name

schaefi:
Ok I have started to add the schema extensions as discussed. This is work in progress and also the test is not complete

cyphar:
@davidcassany openSUSE/umoci#80 has been fixed, and I have added integration tests to make sure that unicode character file names don't cause further issues in the future.

cyphar:
@schaefi Can you explain what `command=` is meant to mean here? Is it just the first argument in `argv`?

```
    <entrypoint command="/bin/bash">        
         <argument name="arg1"/>
         <argument name="arg2"/>        
    </entrypoint>
```

And please note that `CMD` and `ENTRYPOINT` both need to be supported -- `ENTRYPOINT` is used to be the hardcoded set of arguments and commands, and `CMD` is the default arguments (which can be overridden easily on the commandline).

I don't like the way they did that split, but we need it for our images.

schaefi:
@cyphar  yes with command I meant the command :) argv[0] if you like

I understand that we need the hardcoded entrypoint, but I'm confused about the purpose of config.cmd ?
So what happens if the following information exists

```
config.entrypoint="/bin/bash" config.entrypoint="-c" config.entrypoint="ls -l"
config.cmd="ssh" config.cmd="root@foo" config.cmd="ls -l"
```

two totally different set of specifying a command call, but what should that tell us ? entrypoint is used of no other one is given on docker run, but what is the purpose of config.cmd ? Sorry that confuses me

Thanks

cyphar:
@schaefi In Docker this is how it's used. Imagine that you have an image with `config.Entrypoint = ["/bin/echo", "entrypoint-argument"]` and `config.Cmd = ["cmd", "arguments", "here"]`.

```
% docker run image
entrypoint-argument cmd arguments here
% docker run image other arguments
entrypoint-argument other arguments
```

In effect, Docker uses `Entrypoint` as the "mandatory" part of `argv`, and `Cmd` as the "user defined" part of `argv` (with a set default). **However** note that currently this is not part of the OCI specification. The only thing you can currently say with certainty is that both fields **must** be implemented in order to be compliant with the spec.

schaefi:
Ah now I understand how this fits together, thanks much

schaefi:
ok so given the following entrypoint description

```xml
<entrypoint command="/bin/bash">        
     <argument name="arg1"/>
     <argument name="arg2"/>        
</entrypoint>
```

we will pass the following commandline

```bash
--config.entrypoint='/bin/bash' --config.entrypoint='arg1' --config.entrypoint='arg2' --config.cmd='/bin/bash' --config.cmd='arg1' --config.cmd='arg2'
```

Does this make sense or do we need a more fine grained setup from an XML perspective ?

schaefi:
next proposal for setting up entrypoint and cmd

```xml
<entrypoint execute="/bin/bash">
    <argument name="-x"/>
    <subcommand execute="somebin">
        <argument name="arg1"/>
        <argument name="arg2"/>
    </subcommand>
</entrypoint>
```

This would result in

```bash
--config.entrypoint="/bin/bash" --config.entrypoint="-x" --config.cmd="somebin" --config.cmd="arg1" --config.cmd="arg2"
```

Thoughts

davidcassany:
@schaefi I like this nested structure, feels more natural that having completely independent entry_point and command sections.

cyphar:
The problem is that you can have an empty (even without `argv[0]`!) `Entrypoint` -- and the same with `Cmd`. While they are related in Docker's use, I wouldn't really want to make it so that someone who has to specify an XML has to do it this way. What if we went with this:

```
<entrypoint>
  <argument>...</argument>
</entrypoint>

<cmd>
  <argument>...</argument>
</cmd>
```

Without the `command=` and other stuff? I understand it's a bit ugly, but the proposed syntax means that it's not really possible to create an empty `Entrypoint`...

schaefi:
why should that not be possible ?

```xml
<entrypoint/>  
```

empty entrypoint, empty command, or you don't specify an entrypoint section at all

```xml
<entrypoint>
    <subcommand execute="somebin">
        <argument name="arg1"/>
        <argument name="arg2"/>
    </subcommand>
</entrypoint>
```

empty entrypoint, but some command. Only God knows why that should be allowed when you said entrypoint is a mandatory information :) but the xml structure can allow for all that

I dislike having two independent sections for something which is intentionally put together by docker

schaefi:
I forgot ```<argument>``` is of course optional too

flavio:
I would prefer to have `cmd` and `entrypoint` clearly separated like @cyphar proposed:

```xml
<entrypoint>
  <argument>...</argument>
</entrypoint>

<cmd>
  <argument>...</argument>
</cmd>
```

This is why:
  * the way `cmd` and `entrypoint` are handled by docker is a confusing for the users, but we cannot do anything in that regard
  * by providing a *"1:1"* mapping between Dockerfile and KIWI's xml users will be less confused -> fewer bugs/questions -> it will be easier to convert an existing Dockerfile to a xml

schaefi:
sorry but the fact that everybody seem to agree it's confusing what docker did does not mean we have to do the same for the kiwi XML description. afaik a DockerFile has no semantic structure nor is it connected to a schema, it's not even using a markup language. So this file feels like an environment file specifying variables like CMD or ENTRYPOINT. A conversion between the container metadata part of an image description and the relevant parts for the dockerfile would require a transformation code anyway, which can be easily done by an xslt stylesheet.

We really don't have to care for a mapping, but we should care for a better (more user friendly) and clear definition of this command string. In the end its not more than a command splitted up in two portions and appended together: ENTRYPOINT + CMD

So let me ask:

* is entrypoint mandatory yes|no ? according to one of the last messages here it is, but it also can be empty which feels not really mandatory

* It's always entrypoint + cmd and not any other order yes|no ? if yes that means a nested structure is useful to indicate that, more userfriendly, more clear, and exactly tells this information belongs together.

I can add any kind of constraint to the suggested structure from being open to all forms and also being more restrictive, but maintaining two independent toplevel sections seems pretty wrong to me from a schema specification point of view

cyphar:
@schaefi 

> is entrypoint mandatory yes|no ? according to one of the last messages here it is, but it also can be empty which feels not really mandatory

No, it isn't. It can be empty, both in Docker and OCI images. Same goes for `Cmd`.

> It's always entrypoint + cmd and not any other order yes|no ? if yes that means a nested structure is useful to indicate that, more userfriendly, more clear, and exactly tells this information belongs together.

While this is how Docker uses it, this is not currently defined as part of the OCI spec. This discussion is currently happening inside opencontainers/image-spec#492. For the moment, coupling it in this way is not guaranteed to be correct.

schaefi:
Hmm, ok in this case it might be in fact better to have two separate sections. Could you give me a time frame when the OCI spec will come to a conclusion in this regard ? Does it make sense to wait ?

I'm asking that because changing a schema definition in a compatible way is always much more work than setting up the best structure in the first shot. I'm still convinced if docker handles ENTRYPOINT and CMD together it makes a lot sense to couple this information, but I can't tell under which circumstances OCI is discussing this part

Thanks

schaefi:
ok, let's go with two separate sections for the moment

schaefi:
ok the umoci config implementation is done with all the parameters we need. During testing I stumbled over another skope problem which I can't explain:

```
DEBUG: EXEC: [umoci config --author="tux" --config.user="root" --config.workingdir="/root" --config.entrypoint="/bin/bash" --config.entrypoint="-x" --config.cmd="ls" --config.cmd="-l" --config.exposedports="80" --config.exposedports="8080" --config.volume="/tmp" --config.volume="/var/log" --config.env="PATH=/bin:/usr/bin:/home/user/bin" --config.env="SOMEVAR=somevalue" --config.label="somelabel=labelvalue" --config.label="someotherlabel=anotherlabelvalue" --image /tmp/kiwi_docker_dir.ux3u4xla/umoci_layout:container_name]
DEBUG: EXEC: [umoci gc --layout /tmp/kiwi_docker_dir.ux3u4xla/umoci_layout]
DEBUG: EXEC: [rm -r -f /tmp/mytest//LimeJeOS-Leap-42.2.x86_64-1.42.2.docker.tar]
DEBUG: EXEC: [skopeo copy --remove-signatures oci:/tmp/kiwi_docker_dir.ux3u4xla/umoci_layout:container_name docker-archive:/tmp/mytest//LimeJeOS-Leap-42.2.x86_64-1.42.2.docker.tar:container_name]
DEBUG: EXEC: Failed with stderr: time="2017-02-15T11:13:49+01:00" level=fatal msg="Error writing manifest: Unsupported manifest type, need a Docker schema 2 manifest" 
, stdout: Copying blob sha256:9881f50d7cd5eb8372de8edb36a95356bca60787dec8cf8b123e3501a0989ce3
 215.06 MB / 238.89 MB [=================================================>-----]
Copying config sha256:f653886ed29339ad7d78c81393aecf07df8fc3e72bf5c02f9f06153502a3bcb9
 0 B / 785 B [-----------------------------------------------------------------]
Writing manifest to image destination
```

Unsupported manifest ?

Any help is appreciated. I have used

```
umoci-0.1.0-17.1.x86_64
skopeo-0.1.18-11.1.x86_64
```

Thanks

davidcassany:
@schaefi, @cyphar I also faced that issue and I am currently having a look, apparenlty the MediaType of the manifest is not properly set within OCI<->Docker convertion. I just did a quick hack to skopeo to work around that and I am on the way to test it.

cyphar:
Damn. I fixed this issue several months ago, it's possible upstream changed something and they broke it again (I've been trying to get tests into skopeo to avoid these sorts of breakages).

cyphar:
The fix for this is in containers/image#237 and I'll add the patch to openSUSE's `skopeo`. I'm also going to add more tests to `skopeo` upstream to avoid this issue, but the current testing system of `skopeo` is questionable at best...

cyphar:
Alright, I've submitted this SR to fix the issue in `Virtualization:containers`: https://build.opensuse.org/request/show/457418

schaefi:
Thanks much for taking care

schaefi:
I can confirm skopeo-0.1.18-13.1.x86_64 works great, thanks

davidcassany:
@schaefi I hope you don't mind I pushed some minor changes to remove the quotes from the umoci command.

All the tests I have done so far are all green 😃 I also tested some values with spaces to make sure it was correctly handled.

@schaefi looks all nice to me 👍

schaefi:
great 😄 Thanks much, of course you can push fixes as often as you like 😄 

So we are basically ready to merge but we still should wait for skopeo and umoci to show up in a distrobution repository.

@flavio can you give us some details what the plan is with those tools ?

* will they be released as a maintenance update to SLE12 ?
* will they be available on Leap

Thanks

cyphar:
@schaefi Currently `umoci` has an open SR to be in `openSUSE:Factory` (https://build.opensuse.org/request/show/456512). `skopeo` doesn't at the moment, mainly because of recent issues. I'm currently working upstream to help make their integration testing for OCI image handling much better so that I have a high enough confidence to push it into `openSUSE:Factory`.

flavio:
> @flavio can you give us some details what the plan is with those tools ?
> 
>     will they be released as a maintenance update to SLE12 ?
>     will they be available on Leap

As @cyphar said they are both going to be part of Factory. Hence they will appear inside of Leap too.

As for SLE12 I doubt they will be part of it. I would add them to the containers module together with all the other container-related tools we maintain.

@eotchi: what do you think about my SLE proposal?

schaefi:
All right, let's merge this once the tests are green

cyphar:
Yay!

PR #217, opened 2017-01-09
Running config.sh after including the image repositories

### LINKED ISSUE #214: config.sh is executing before repository setup
## Problem description
If I'm not mistaken in kiwi v7 the flow was that it first initialized repositories inside an image and than call config.sh inside the image. Right now it is the other way around, so If i try to run `zypper --non-interactive --gpg-auto-import-keys refresh` the keys are not imported because the repos were not set up when script run.

## Actual behaviour
See [log](http://pastebin.com/7ZDCKhMv) of the build

davidcassany:
This commit rearranges some code lines in order to run the config.sh
script after the repositories of the image have been included, this
way repositories can be trusted and refreshed in config.sh script.

Fixes #214

### DIFF
--- a/kiwi/tasks/system_build.py
+++ b/kiwi/tasks/system_build.py
@@ -174,7 +174,6 @@ class SystemBuildTask(CliTask):
 
         setup.import_description()
         setup.import_overlay_files()
-        setup.call_config_script()
         setup.import_image_identifier()
         setup.setup_groups()
         setup.setup_users()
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi do you remember why the system instance was deleted before including the image repositories? I bet there was a good reason, but I couldn't see it now... I just opened the Pull Request to share the code and start the discussion. 

Finding a way to include trusted repositories is something that has been also requested by the Docker team.

schaefi:
I guess the reason was to have all connections to the host system (bind mounts, etc) to be released before custom scripts come into play. However if that prevents us from providing required functionality I'm fine with the change

PR #220, opened 2017-01-16
Adapt mountpoint to the default toplevel subvolume

davidcassany:
If using root btrfs snapshot, the default toplevel subvolume is set to
`/@/.snapshots/1/snapshot`, thus all defined subvolumes are mounted
under that customized default subvolume. For the first time
subvolumes are mounted it is fine to include `/@/.snapshots/1/snapshot`
prefix as root is not yet set to that specific path, however in any
future mount this path prefix is not needed any more, as the root
gets mounted in` /@/.snapshots/1/snapshot` by default.

This PR handles this behavior by testing the subvolume path
of the specified root mountpoint in order to verify if the
snapshots path prefix should be included or not.

This PR fixes bnc#1015549

### DIFF
--- a/kiwi/volume_manager/btrfs.py
+++ b/kiwi/volume_manager/btrfs.py
@@ -70,6 +70,7 @@ class VolumeManagerBtrfs(VolumeManagerBase):
 
         self.subvol_mount_list = []
         self.toplevel_mount = None
+        self.toplevel_volume = None
 
     def setup(self, name=None):
         """
@@ -232,11 +233,16 @@ class VolumeManagerBtrfs(VolumeManagerBase):
         """
         Mount btrfs subvolumes
         """
-        if not self.toplevel_mount.is_mounted():
-            self.toplevel_mount.mount(
-                self.custom_filesystem_args['mount_options']
-            )
+
+        self.toplevel_mount.mount(
+            self.custom_filesystem_args['mount_options']
+        )
+
         for volume_mount in self.subvol_mount_list:
+            if self.volumes_mounted_initially:
+                volume_mount.mountpoint = os.path.normpath(
+                    volume_mount.mountpoint.replace(self.toplevel_volume, '', 1)
+                )
             if not os.path.exists(volume_mount.mountpoint):
                 Path.create(volume_mount.mountpoint)
             subvol_name = self._get_subvol_name_from_mountpoint(volume_mount)
@@ -248,6 +254,8 @@ class VolumeManagerBtrfs(VolumeManagerBase):
             volume_mount.mount(
                 options=[subvol_options]
             )
+
+        self.volumes_mounted_initially = True
 
     def umount_volumes(self):
         """
@@ -313,6 +321,7 @@ class VolumeManagerBtrfs(VolumeManagerBase):
                             volume_id, self.mountpoint
                         ]
                     )
+                    self.toplevel_volume = default_volume
                     return
 
         raise KiwiVolumeRootIDError(
@@ -347,9 +356,9 @@ class VolumeManagerBtrfs(VolumeManagerBase):
 
     def _get_subvol_name_from_mountpoint(self, volume_mount):
         subvol_name = '/'.join(volume_mount.mountpoint.split('/')[3:])
-        if self.custom_args['root_is_snapshot']:
-            subvol_name = subvol_name.replace('.snapshots/1/snapshot/', '')
-        return os.path.normpath(subvol_name)
+        if self.toplevel_volume and self.toplevel_volume in subvol_name:
+            subvol_name = subvol_name.replace(self.toplevel_volume, '')
+        return os.path.normpath(os.sep.join(['@', subvol_name]))
 
     def __del__(self):
         if self.toplevel_mount:
--- a/kiwi/volume_manager/base.py
+++ b/kiwi/volume_manager/base.py
@@ -83,6 +83,10 @@ class VolumeManagerBase(DeviceProvider):
         # the device should be released.
         self.device_provider = device_provider
 
+        # An indicator for the mount of the filesystem and its volumes
+        # when mounted for the first time
+        self.volumes_mounted_initially = False
+
         self.root_dir = root_dir
         self.volumes = volumes
         self.volume_group = None
--- a/kiwi/volume_manager/lvm.py
+++ b/kiwi/volume_manager/lvm.py
@@ -211,6 +211,7 @@ class VolumeManagerLVM(VolumeManagerBase):
             volume_mount.mount(
                 options=[self.mount_options]
             )
+        self.volumes_mounted_initially = True
 
     def umount_volumes(self):
         """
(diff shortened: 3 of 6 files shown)


schaefi:
Hmm, I understand the problem but I think we could fix this more explicit. There is the method:

```python
_set_default_volume()
```

once this is called we know the root of the btrfs and don't need to call an extra program to read the mount flags. I would be more in favour to let _set_default_volume store the value of the default root in the object

Assuming the object knows the root the method

```python
_get_subvol_name_from_mountpoint()
```

could be smart enough to return the correct subvol_name

imho this should lead to a less complex patch

Thougths ?

davidcassany:
Yes that might be a simpler option, you are right. The only potential issue I see (if it can be considered as so) is that running multiple times moult_volumes method without umounting first may lead to some inconsistency, it could mount twice in different locations. However that could be considered a bug in the code to run mount_volumes method several times without umount first.

schaefi:
good point, we can handle that case as well and allow mounting only once. if an attempt is made to mount when the target is already mounted we issue a warning that we will not mount again. imho it can be done by using a mount counter between mount_volumes and umount_volumes. Thoughts ?

davidcassany:
Yes I am already exploring it. Another little issue is that we need to identify first mount, so we need an state flag (set_default is called right before mounting for the first time), which leads me to another some thoughts: I guess there isn't more than a single instance of VolumeManager in each kiwi run, we would have issues if the object is destroyed and created again.

I am already rewriting the patch, I will push it here in short.

schaefi:
there should only be only one VolumeManager instance to solve one task (builder/disk.py) afaik this should be safe

schaefi:
yes you can introduce state flags in the constructor of a VolumeManager

davidcassany:
See the updated patch, it simplifies the `_get_subvol_name_from_mountpoint()` method a little and doesn't need of an extra tool to test mountpoints. I have not included any counter between mount_volumes and umount_volumes yet, as to me it is a bit weird to implement it only for btrfs, that logic imho should be independent of the filesystem. Then we have two options, in case we want to implement it, place the logic on the VolumeManagerBase or place it on the MountManager. I'd vote for VolumeManagerBase.

kiwi/volume_manager/btrfs.py
```diff
@@ -70,6 +70,8 @@ def post_init(self, custom_args):
 
         self.subvol_mount_list = []
         self.toplevel_mount = None
+        self.first_mount = True
```
schaefi: Can we change the logic here ? I think if you construct a volume manager no volumes are mounted,
thus the following feels more natural to me

```python
self.volumes_mounted = False
```

sorry if that's nit picking

kiwi/volume_manager/btrfs.py
```diff
@@ -70,6 +70,8 @@ def post_init(self, custom_args):
...
+        self.toplevel_volume = ''
```
schaefi: Can we initialize that with

```python
self.toplevel_volume = None
```

schaefi:
I agree the initialization code added to the btrfs constructor should be in base.

kiwi/volume_manager/btrfs.py
```diff
@@ -347,9 +355,9 @@ def _create_snapshot_info(self, filename):
 
     def _get_subvol_name_from_mountpoint(self, volume_mount):
         subvol_name = '/'.join(volume_mount.mountpoint.split('/')[3:])
-        if self.custom_args['root_is_snapshot']:
-            subvol_name = subvol_name.replace('.snapshots/1/snapshot/', '')
-        return os.path.normpath(subvol_name)
+        if self.toplevel_volume and self.toplevel_volume in subvol_name:
+            subvol_name = subvol_name.replace(self.toplevel_volume, '')
+        return os.path.normpath('/'.join(['@', subvol_name]))
```
schaefi: use os.sep instead of '/'

davidcassany: yep sure

schaefi: Hmm, and I wonder why "return os.path.normpath(subvol_name)" has changed to "return os.path.normpath('/'.join(['@', subvol_name]))"

?

kiwi/volume_manager/btrfs.py
```diff
@@ -347,9 +355,9 @@ def _create_snapshot_info(self, filename):
 
     def _get_subvol_name_from_mountpoint(self, volume_mount):
         subvol_name = '/'.join(volume_mount.mountpoint.split('/')[3:])
-        if self.custom_args['root_is_snapshot']:
-            subvol_name = subvol_name.replace('.snapshots/1/snapshot/', '')
-        return os.path.normpath(subvol_name)
+        if self.toplevel_volume and self.toplevel_volume in subvol_name:
+            subvol_name = subvol_name.replace(self.toplevel_volume, '')
+        return os.path.normpath(os.sep.join(['@', subvol_name]))
```
schaefi: why has the return line changed from

```
return os.path.normpath(subvol_name)
```

to

```
return os.path.normpath(os.sep.join(['@', subvol_name]))
```

kiwi/volume_manager/btrfs.py
```diff
@@ -347,9 +355,9 @@ def _create_snapshot_info(self, filename):
 
     def _get_subvol_name_from_mountpoint(self, volume_mount):
         subvol_name = '/'.join(volume_mount.mountpoint.split('/')[3:])
-        if self.custom_args['root_is_snapshot']:
-            subvol_name = subvol_name.replace('.snapshots/1/snapshot/', '')
-        return os.path.normpath(subvol_name)
+        if self.toplevel_volume and self.toplevel_volume in subvol_name:
```
schaefi: and is that not also depending on if self.volumes_mounted is true or false ?

schaefi: ah I see you are changing the mountpoint in volume_mount.mountpoint before

kiwi/volume_manager/btrfs.py
schaefi: and maybe the answer is because of the changed mountpoint in volume_mount.mountpoint.

kiwi/volume_manager/btrfs.py
davidcassany: Yes I was in doubt about that also. The '@' is included because the toplevel_volume always includes it (in fact it can only be '@' or '@/.snapshots/1/snapshot'), so it gets replaced if the mountpoint already includes the toplevel subvolume path. Then '@' is included out of the if clause because in case the mountpoint does not include the toplevel subvolume it means it in the shape of something line `/tmp/tmpdirxxx/var/cache`, so it that case it also needs to be appended.

I know this lines are quite ofuscated, I don't like them either, I tried to keep them as simple as possible, not to make a complex  `_get_subvol_name_from_mountpoint` method.

schaefi:
I have done several build tests, this looks pretty good now. I will do a last build with the type definition of the casp project and then this can be merged

@davidcassany what do you think ?

schaefi:
I think we are good to go, all build tests worked fine for me. Thanks David for your time and effort

PR #230, opened 2017-01-31
Components of an Image Description section

davidcassany:
This commit extends the `doc/sources/workflow.rst` documentation
by including details about `config.sh` and `images.sh` files.

### DIFF
--- a/doc/source/workflow.rst
+++ b/doc/source/workflow.rst
@@ -33,10 +33,13 @@ images that:
 
 The image creation process with KIWI is automated and does not require any
 user interaction. The information required for the image creation process is
-provided by the primary configuration file named :file:`config.xml`. In
-addition, the image can optionally be customized using the :file:`config.sh`
-and :file:`images.sh` scripts and by using an *overlay tree (directory)*
-called :file:`root`.
+provided by the primary configuration file named :file:`config.xml`. 
+This file is validated against the schema documented in
+:ref:`Schema Documentation <schema-docs>` section.
+In addition, the image can optionally be customized
+using the :file:`config.sh` and :file:`images.sh` scripts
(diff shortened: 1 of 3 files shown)


doc/source/workflow.rst
```diff
@@ -207,6 +214,474 @@ KIWI:
 
    #. **Container Image**: Base images for the Docker engine.
 
+Components of an Image Description
+----------------------------------
+
+A KIWI image description can composed by several parts. The main part is
+the KIWI description file itself, the :file:`config.xml` file. The
```
schaefi: alternatively user can also name the XML file "something.kiwi"

doc/source/workflow.rst
```diff
@@ -207,6 +214,474 @@ KIWI:
...
+   create step. So it is expected to be used to handle image type specific
+   tasks.
+
+#. Overlay tree directory
+
+   The *overlay tree* is a folder, called :file:`root`, that contains
+   files and directories that will be copied to the target image build tree
+   during the :ref:`prepare step <prepare-step>`. It is executed
+   after all the packages included in the :file:`config.xml` file
+   have been installed. Any already present file is overwritten.
```
schaefi: The overlay tree can also be provided as tarball "root.tar.gz"

schaefi:
apart from the two small addition it looks great 👍

doc/source/workflow.rst
davidcassany: @schaefi yes, this is explained in a note in `Building Image` section. But you are right it makes a lot of sense to remind it here. I'll place a link or add a comment here.

doc/source/workflow.rst
davidcassany: @schaefi, right, needs to be specified.

PR #231, opened 2017-02-01
Added support for package locking for zypper

schaefi:
Zypper supports the al (add lock) command which allows to ignore a package in the dependecny resolution process. This is useful to prevent installation of a package which was pulled in by e.g a recommendation flag from the spec file. Packages marked to be ignored are not handled for apt and yum right now. Using this feature together with an unsupported package manager backend results in a warning to the user

A package lock can be set via the XML description as follows:

```xml

<packages type="image">
    <ignore name="package"/>
</packages>
```

The XML description supported that already but kiwi v8 did not had support for it. Thus this completes the zypper package manager support with regards to what the kiwi v7 version could do

### DIFF
--- a/doc/source/schema.rst
+++ b/doc/source/schema.rst
@@ -835,7 +835,7 @@ List of attributes for ``configuration``:
 
 * ``source`` : A location where packages can be found to build an installation source from
 * ``dest`` : Destination of a resource
-* ``arch`` `[?]`_: An architecture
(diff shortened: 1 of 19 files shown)


davidcassany:
All the code is fine for me. I just have a related thought: shouldn't include the `ignore` tag include the `arch` attribute? I think it would make sense, I've seen it is not in the schema, so maybe there was a motivation not to include it, but IMHO it could be there and it would also make sense.

schaefi:
good point, I'm going to add that part

davidcassany:
@schaefi looks fine to me 👍

schaefi:
Thanks :)

PR #232, opened 2017-02-02
Eliminate instsource element and requirements

schaefi:
The schema defintions for kiwi v8 still contained the product schema definitions aka instsource which is no longer supported by this version of kiwi. Building SUSE products is done with the kiwi v7 code base and will most probably be handled by an extra product builder tool in the future maintained and developed by the buildservice team. We would be proud if the api of the kiwi v8 code base can be helpful for the product builder in the future


davidcassany:
Great clean up 👍

schaefi:
thanks 😄

PR #233, opened 2017-02-07
Added kiwi image info command

schaefi:
Allow to list information about the image description. In a first implementation one can show the image name as well as run a dependency resolution for the packages listed in the XML description


schaefi:
solver error handling is missing... stay tuned

PR #243, opened 2017-02-20
Include compression optimization flags for mksquashfs command

### LINKED ISSUE #242: Live images increase in size by ~ 5% when building using kiwi8 compared to kiwi7
In an attempt to convert the openSUSE Factory Live images to kiwi8 I ran into a (for openSUSE:Factory) enormous issue:

openSUSE:Factory:Live uses kiwi out of a spec file (as there are a couple safeguards in place which kiwi/OBS does not offer on its own - but that's out of scope for this issue)

so, I converted the .spec file to no longer use kiwi, but actually run kiwi-ng (converted command lines)
the config.xml fed to kiwi7 and kiwi-ng is the same file (comes from openSUSE:Factory:Live/kiwi-config-openSUSE)

The resulting iso file as produced by kiwi8 is roughly 5 - 7% larger than the one produced by kiwi7.

this is insofar a problem as that we declared the Live image to be a maximum of 1 000 000 000 bytes in size and the current kiwi7 builds of it are already very close to that limit (it's a constant battle to stay under the limit)

with kiwi8, having the image grow by 66MB made it go way above the limit - which means I would have to sacrifice software on the media in order to allow building with kiwi8

davidcassany:
This commit includes some flags for mksquashfs command in other to
achieve higher compression rates. Also note that those flags were
already present in KIWI former versions, thus they have been
included again for compatibility reasons.

Fixes #242

### DIFF
--- a/kiwi/filesystem/squashfs.py
+++ b/kiwi/filesystem/squashfs.py
@@ -14,7 +14,9 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
-#
+
+import platform
+
 # project
 from .base import FileSystemBase
 from ..command import Command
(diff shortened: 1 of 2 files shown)


schaefi:
👍

PR #244, opened 2017-02-24
Added DNF package management implementation

schaefi:
DNF is the package manager used on Fedora systems and will also most likely replace yum for the next generation of RHEL. Along with the implementation here an example JeOS image description for Fedora can be found on: 

* https://github.com/SUSE/kiwi-descriptions

### DIFF
--- a/kiwi/repository/dnf.py
+++ b/kiwi/repository/dnf.py
@@ -1,0 +1,238 @@
+# Copyright (c) 2015 SUSE Linux GmbH.  All rights reserved.
+#
+# This file is part of kiwi.
+#
+# kiwi is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# kiwi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kiwi.  If not, see <http://www.gnu.org/licenses/>
+#
+import os
+from six.moves.configparser import ConfigParser
+from tempfile import NamedTemporaryFile
+
+# project
+from ..logger import log
+from .base import RepositoryBase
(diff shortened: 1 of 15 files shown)


kiwi/repository/dnf.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def use_default_location(self):
+        """
+        Setup dnf repository operations to store all data
+        in the default places
+        """
+        self.shared_dnf_dir['reposd-dir'] = \
+            self.root_dir + '/etc/yum/repos.d'
+        self.shared_dnf_dir['cache-dir'] = \
+            self.root_dir + '/var/cache/yum'
```
Conan-Kudo: Default cache dir is `/var/cache/dnf`

kiwi/repository/dnf.py
```diff
@@ -0,0 +1,238 @@
...
+        self.shared_dnf_dir['pluginconf-dir'] = \
+            self.root_dir + '/etc/yum/pluginconf.d'
```
Conan-Kudo: Default plugin config dir is `/etc/dnf/plugins`

kiwi/repository/dnf.py
```diff
@@ -0,0 +1,238 @@
...
+        :param list custom_args: dnf arguments
+        """
+        self.custom_args = custom_args
+        if not custom_args:
+            self.custom_args = []
+
+        # extract custom arguments not used in dnf call
+        if 'exclude_docs' in self.custom_args:
+            self.custom_args.remove('exclude_docs')
+            log.warning('rpm-excludedocs not supported for dnf: ignoring')
```
Conan-Kudo: You can do this by doing `--setopt='tsflags=nodocs'`. There is a [pending PR for adding `--nodocs`](https://github.com/rpm-software-management/dnf/pull/748).

schaefi: Thanks, can this option also be part of the dnf config file, e.g

```
[main]
tsflags=nodocs
```

or would that not work ?

schaefi: Looking at https://github.com/fedora-cloud/docker-brew-fedora/issues/20 this should work for dnf and yum.

Conan-Kudo: That will work too!

schaefi:
Thanks for the review 😄

PR #247, opened 2017-02-28
Fix default reposdir path for DNF

Conan-Kudo:
The default reposdir path for DNF is wrong (subtle one-character change that I missed during initial review).

* `/etc/yum/repos.d` -> `/etc/yum.repos.d`


schaefi:
Thanks much 👍

PR #250, opened 2017-03-01
Fix tagging for OCI images

### LINKED ISSUE #249: Docker configurations are not imported to the resulting image
## Problem description

Docker images have no command, entrypoint and other configuration parameters set, even specified in KIWI description file.

`docker run -ti buildimage` always fails due to not having any command specified to run.
Also running `docker inspect buildimage` show all the configuration parameters are empty or null.

davidcassany:
This commit fixes the tagging schema for umoci. An OCI image
name is path[:tag], this commit rearranges some variable names to
avoid confusions between names and tags.

Fixes #249


schaefi:
Thanks that's much better to read 👍

PR #252, opened 2017-03-01
Use default command only if no entrypoit and subcommand aren't set

### LINKED ISSUE #251: Docker default command
## Problem description

Images with a only a subcommand specified have a /bin/bash pre-appended. Default command should only be set if neither entrypoint and subcommand are specified.

## Actual behaviour

Images with only a subcommand, without entrypoint, fail to run when no command arguments are provided at runtime.

Use a config file with only a subcommand configuration e.g. `<subcommand execute="fortune"/>`

davidcassany:
Set the default command only if no `subcommand` and `entrypoint` sections are missing in config.xml

Fixes #251

### DIFF
--- a/kiwi/container/docker.py
+++ b/kiwi/container/docker.py
@@ -43,9 +43,9 @@ class ContainerImageDocker(object):
         self.docker_dir = None
         self.docker_root_dir = None
 
-        self.container_name = []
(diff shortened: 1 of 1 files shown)


schaefi:
Thanks 👍

PR #254, opened 2017-03-02
Support container names with slashes

### LINKED ISSUE #253: Docker images names containing slashes
## Problem description

KIWI docker images do not support the common naming convention used by the docker registries, e.g. `opensuse/dice:latest`.

## Actual behaviour

Including `/` symbols in `container_name` attribute leads to a failure during the image creation procedure.

davidcassany:
This commit provides a patch in order to enable container_name
attribute values with slashes. Given that OCI images names are in
the shape of path[:tag], appropriate subdirectories need to be created
if the container name contains slashes.

Fixes #253

### DIFF
--- a/kiwi/container/docker.py
+++ b/kiwi/container/docker.py
@@ -107,18 +107,11 @@ class ContainerImageDocker(object):
         self.docker_root_dir = mkdtemp(prefix='kiwi_docker_root_dir.')
 
         container_dir = os.sep.join(
-            [self.docker_dir, self.container_name]
+            [self.docker_dir, 'umoci_layout']
         )
         container_name = ':'.join(
             [container_dir, self.container_tag]
         )
-
-        # Create container subdirectories if needed
-        # umoci is not capable to create subdirectories e.g. opensuse/leap
-        if os.sep in self.container_name:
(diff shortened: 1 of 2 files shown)


cyphar:
Uhm, I'm not sure what the intention is of making subdirectories in this way -- there is no difference between an image in a directory called `foo` or `foo/bar`. This only /really/ matters for `skopeo` when exporting.

Note however that in the next update of `umoci` some aspects of this UX will have to change unfortunately, because we can no longer represent everything as simple tags. I'll talk to you about this on IRC or mumble to come to agreement about what UX you would like to see.

davidcassany:
Well, the intention is just to call skopeo and umoci in a parallel way and not having to handle different suffixes for each tool. 

Anyway, you are right, it can be simplified without having to create dummy directories.

davidcassany:
@schaefi, @cyphar I think now is a nice way to leave it meanwhile, simple and using container names like `opensuse/amd64` don't fail and are correctly loaded by docker.

@schaefi if no objections I will squash commits to write a more convenient log.

schaefi:
I'm ok with the change 👍

PR #255, opened 2017-03-02
Add support for repository credentials

### LINKED ISSUE #246: repositories with username and password not handled
## Problem description

Repositories with basic authentication credentials are not handled. For e.g zypper data in etc/zypp/credentials must be added, other package managers e.g dnf, yum and apt are also not supported

## Actual behaviour

username and password attributes can be set but are not used

Use a repository with credentials

schaefi:
The package and solver repository classes did not provide an
interface to deal with repository credentials. This commit
add support for the zypper package manager and the generic
urlopen based download method of the solver class. This
Fixes #246

### DIFF
--- a/kiwi/repository/apt.py
+++ b/kiwi/repository/apt.py
@@ -113,7 +113,9 @@ class RepositoryApt(RepositoryBase):
         }
 
     def add_repo(
-        self, name, uri, repo_type='deb', prio=None, dist=None, components=None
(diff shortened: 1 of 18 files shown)


schaefi:
Credentials handling for other than the zypper package manager has been prepared but not yet implemented. For dnf, yum and apt this can be done as follow up pull requests and should not hold us back from merging this one

PR #257, opened 2017-03-03
Use kernel alias name for default kernel

schaefi:
Instead of a hard requirement to kernel-default we should
better use the kernel-base alias in the boot image
descriptions. This Fixes (bsc#1027610)


schaefi:
ongoing discsussion in https://bugzilla.suse.com/show_bug.cgi?id=1027610 points out this one seems not the right solution. Thus don't merge !

schaefi:
we are not approaching this as the correct soloution, closing

PR #259, opened 2017-03-03
Derived docker images

davidcassany:
This PR extends KIWI to support derived docker images. Now a layered docker image can be built on top of a given compressed docker images file. Attribute `derived_from` has been added for docker image type to reference the base image.

Type section for derived images looks like:
```xml
<type image="docker" derived_from="file:///someimagefile.tar.xz">
```

### DIFF
--- a/kiwi/builder/container.py
+++ b/kiwi/builder/container.py
@@ -16,6 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import platform
+import os
 
 # project
 from ..container import ContainerImage
@@ -23,6 +24,7 @@ from ..container.setup import ContainerSetup
 from ..system.setup import SystemSetup
 from ..logger import log
 from ..system.result import Result
+from ..system.uri import Uri
 
 
 class ContainerBuilder(object):
@@ -69,6 +71,14 @@ class ContainerBuilder(object):
                 '.', self.requested_container_type, '.tar.xz'
             ]
         )
+        self.base_image = None
+
+        base_image = xml_state.build_type.get_derived_from()
+        if base_image:
+            uri = Uri(base_image, 'images')
+            self.base_image = os.sep.join(
+                [self.root_dir, 'image', os.path.basename(uri.translate())]
+            )
         self.result = Result(xml_state)
 
     def create(self):
@@ -80,22 +90,25 @@ class ContainerBuilder(object):
 
         * image="docker"
         """
-        log.info(
-            'Setting up %s container', self.requested_container_type
-        )
-        container_setup = ContainerSetup(
-            self.requested_container_type, self.root_dir, self.container_config
-        )
-        container_setup.setup()
+        if not self.base_image:
+            log.info(
+                'Setting up %s container', self.requested_container_type
+            )
+            container_setup = ContainerSetup(
+                self.requested_container_type, self.root_dir,
+                self.container_config
+            )
+            container_setup.setup()
 
         log.info(
             '--> Creating container image'
         )
         container_image = ContainerImage(
-            self.requested_container_type, self.root_dir, self.container_config
+            self.requested_container_type, self.root_dir,
+            self.container_config
         )
         container_image.create(
-            self.filename
+            self.filename, self.base_image
         )
         self.result.add(
             key='container',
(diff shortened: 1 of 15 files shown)


schaefi:
looks like the branch can't be merged

davidcassany:
@schaefi yes this is because this branch does not include "Support container names with slashes" commit in history, I tried to include it with a git cherry-pick, but it got renamed as a new commit, so that the conflict still persists, I should have merged it instead. I did not care much about that as I thought we might have a close look to the code and possibly change something before merging.

davidcassany:
yes, you won't miss anything, sorry for the confusion, I did something stupid not porting the mentioned commit to this branch before pushing (the code has been always there, but not as a commit).

kiwi/builder/container.py
```diff
@@ -69,6 +71,14 @@ def __init__(self, xml_state, target_dir, root_dir):
...
             ]
         )
+        self.base_image = None
+
+        base_image = xml_state.build_type.get_derived_from()
+        if base_image:
+            uri = Uri(base_image, 'images')
+            self.base_image = os.sep.join(
+                [self.root_dir, 'image', os.path.basename(uri.translate())]
+            )
```
schaefi: Can we make the variable naming more clear ? imho base_image is used to hold just the value of the derived_from attribute and after that its meaning changes to a local path from an uri translate

Second an uri describes a resource which must not necessarily be a local path. Thus I think an uri.is_remote() check seems to be needed here too

Thoughts ?

kiwi/builder/container.py
```diff
@@ -80,22 +90,25 @@ def create(self):
...
+            container_setup.setup()
 
         log.info(
             '--> Creating container image'
         )
         container_image = ContainerImage(
-            self.requested_container_type, self.root_dir, self.container_config
+            self.requested_container_type, self.root_dir,
+            self.container_config
```
schaefi: seems to be an unneeded change, why the line split ?

schaefi:
I'm a bit concerned about the code flow. Let me try to explain what I mean.

You introduced a new class called RootImport which is nice. But it's not a generic approach. It's tight to an instance of ContainerImage and it calls a method unpack_to_root_dir from there where it imho does not belong to. So from an OO perspective the design is not based on single responsibility. The task of RootImport is:

* provide me a root directory imported from somewhere. It should be as generic as this

The task of the container/docker.py:

* build a docker image from a given path, this class has no responsibility to provide any root data. The method unpack_to_root_dir imho does not belong there

I think we should change the design

Thoughts ?

schaefi:
I also think it would be nice to have a first PR which only covers the RootImport from somewhere. We would be able to merge this without breaking anything. After that a subsequent pull request can make use of RootImport for derived docker images.

Imho that makes the review easier too

davidcassany:
Frist, I agree with splitting the PR, that makes sense to me also, as RootImport concept should be independent of any docker related stuff. Now it is tied to ContainerImage since is the only format supported, however note I already included an image type variable that defaults to docker. 

One option would be to create a factory for RootImport class and then have dedicated RootImport(format) classes. Something like RootImport as the factory, RootImportBase as the base class and RootImportDocker for docker images. Initially we could take the format from the image attribute of the type section, so we ensure there are no crossed formats within the root import and the build type. This would still be complete PR, as it would enable to build derived images as a new base image (assuming the kiwi/container/docker changes go to a future PR). If that's OK for you I can make such a PR (using a new clean branch).

Then, about the method unpack_to_root_dir, initially I had a separate class to unpack images, but I had the feeling it was quite similar to the kiwi/container/docker and that I was distributing umoci and skopeo calls (or docker specific stuff) too much over the code. So having all docker specific commands to manage docker images within a single class that only pack/unpack images seamed to me a reasonable approach.

kiwi/builder/container.py
davidcassany: Sure the variable naming should be improved. Then about the uri.is_remote() check I did not include it here because at this stage the base_image is always local, concretely the file that RootImport stores in `root_dir/image` folder.

Anyway, as you said, I agree to divide the PR in two, it will be easier to review and refine, and part would be for the second one.

Thanks for the comments so far

kiwi/builder/container.py
davidcassany: mistake of mine, that shouldn't be changed.

schaefi:
I like the idea of a root import factory 👍 

I think if there is RootImportDocker class it could also hold the code of unpack_to_root_dir which does skopeo/umoci calls. Actually I don't see it as a problem if there is more than one class which calls docker tools if it fits into the responsibility of the class.

So yes an import factory brings everything into good shape

davidcassany:
Closing this PR since it is superseded by #262.

PR #262, opened 2017-03-07
Import root system from a given image

davidcassany:
This commit includes the root import feature. A `derived_from`
attribute has been included with the `<type>` section to make
reference to the image file to import. The image format to import
is assumed to be the same as the build type to import.

Currently, the only format supported is docker


davidcassany:
Note that following files are autogenerated and could be omited from the review:

- doc/source/schema.rst 
-  kiwi/schema/kiwi.rng
- kiwi/xml_parse.py

schaefi:
great work 👍 let's merge it

PR #265, opened 2017-03-08
Support for layered docker images

davidcassany:
This commit includes support for building layered docker. A new
layer is added on top of the base image referenced by `derived_from`
attribute.

### DIFF
--- a/kiwi/builder/container.py
+++ b/kiwi/builder/container.py
@@ -25,6 +25,7 @@ from kiwi.system.setup import SystemSetup
 from kiwi.logger import log
 from kiwi.system.result import Result
 from kiwi.utils.checksum import Checksum
+from kiwi.defaults import Defaults
 from kiwi.exceptions import KiwiContainerBuilderError
 
 
@@ -60,31 +61,24 @@ class ContainerBuilder(object):
         self.target_dir = target_dir
         self.container_config = xml_state.get_container_config()
         self.requested_container_type = xml_state.get_build_type_name()
-        self.base_image_uri = xml_state.get_derived_from_image_uri()
         self.base_image = None
         self.base_image_md5 = None
 
-        if self.base_image_uri:
-            base_image_origin = self.base_image_uri.translate()
+        if xml_state.get_derived_from_image_uri():
             # The base image is expected to be unpacked by the kiwi
             # prepare step and stored inside of the root_dir/image directory.
             # In addition a md5 file of the image is expected too
-            self.base_image = os.path.normpath(
-                os.sep.join(
-                    [
-                        root_dir, 'image',
-                        base_image_origin.replace('.tar.xz', '')
-                    ]
-                )
+            self.base_image = Defaults.get_imported_root_image(
+                self.root_dir
             )
+            self.base_image_md5 = ''.join([self.base_image, '.md5'])
+
             if not os.path.exists(self.base_image):
                 raise KiwiContainerBuilderError(
                     'Unpacked Base image {0} not found'.format(
                         self.base_image
                     )
                 )
-
-            self.base_image_md5 = ''.join([self.base_image, '.md5'])
             if not os.path.exists(self.base_image_md5):
                 raise KiwiContainerBuilderError(
                     'Base image MD5 sum {0} not found at'.format(
--- a/kiwi/builder/container.py
+++ b/kiwi/builder/container.py
@@ -60,10 +60,25 @@ class ContainerBuilder(object):
         self.target_dir = target_dir
         self.container_config = xml_state.get_container_config()
         self.requested_container_type = xml_state.get_build_type_name()
-        if not xml_state.build_type.get_derived_from():
-            self.base_image = None
-        else:
-            self.base_image = os.sep.join([root_dir, 'image', 'image_file'])
+        self.base_image_uri = xml_state.get_derived_from_image_uri()
+        self.base_image = None
+        self.base_image_md5 = None
+
+        if self.base_image_uri:
+            base_image_origin = self.base_image_uri.translate()
+
+            # we expect the base image to be unpacked by the kiwi prepare step
+            # and stored inside of the root_dir/image directory. In addition
+            # a md5 file of the image is expected there too
+            self.base_image = os.path.normpath(
+                os.sep.join(
+                    [
+                        root_dir, 'image',
+                        base_image_origin.replace('.tar.xz', '')
+                    ]
+                )
+            )
+            self.base_image_md5 = ''.join([self.base_image, '.md5'])
 
         self.system_setup = SystemSetup(
             xml_state=xml_state, root_dir=self.root_dir
@@ -93,23 +108,24 @@ class ContainerBuilder(object):
                 'Setting up %s container', self.requested_container_type
             )
             container_setup = ContainerSetup(
-                self.requested_container_type, self.root_dir, self.container_config
+                self.requested_container_type, self.root_dir,
+                self.container_config
             )
             container_setup.setup()
         else:
-            base_image_md5 = ''.join([self.base_image, '.md5'])
-            data = None
             try:
                 checksum = Checksum(self.base_image)
-                with open(base_image_md5, 'r') as md5_file:
-                    data = md5_file.read()
+                with open(self.base_image_md5) as md5_file:
+                    base_image_md5 = md5_file.read()
             except Exception as e:
                 raise KiwiContainerBuilderError(
                     '%s: %s' % (type(e).__name__, format(e))
                 )
-            if data.split(' ')[0] != checksum.md5():
+            if base_image_md5.split(' ')[0] != checksum.md5():
                 raise KiwiContainerBuilderError(
-                    'base image file {0} checksum validation failed'.format(self.base_image)
+                    'base image file {0} checksum validation failed'.format(
+                        self.base_image
+                    )
                 )
 
         log.info(
--- a/kiwi/builder/container.py
+++ b/kiwi/builder/container.py
@@ -24,7 +24,7 @@ from kiwi.container.setup import ContainerSetup
 from kiwi.system.setup import SystemSetup
 from kiwi.logger import log
 from kiwi.system.result import Result
-from kiwi.system.uri import Uri
+from kiwi.utils.checksum import Checksum
 from kiwi.exceptions import KiwiContainerBuilderError
 
 
@@ -60,17 +60,10 @@ class ContainerBuilder(object):
         self.target_dir = target_dir
         self.container_config = xml_state.get_container_config()
         self.requested_container_type = xml_state.get_build_type_name()
-        self.base_image = None
-        image_uri = xml_state.build_type.get_derived_from()
-        if image_uri:
-            uri = Uri(image_uri, 'images')
-            if uri.is_remote():
-                raise KiwiContainerBuilderError(
-                    'Only local base images are supported'
-                )
-            self.base_image = os.sep.join(
-                [root_dir, 'image', os.path.basename(uri.translate())]
-            )
+        if not xml_state.build_type.get_derived_from():
+            self.base_image = None
+        else:
+            self.base_image = os.sep.join([root_dir, 'image', 'image_file'])
 
         self.system_setup = SystemSetup(
             xml_state=xml_state, root_dir=self.root_dir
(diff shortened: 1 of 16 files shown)


kiwi/builder/container.py
```diff
@@ -57,6 +60,18 @@ def __init__(self, xml_state, target_dir, root_dir):
         self.target_dir = target_dir
         self.container_config = xml_state.get_container_config()
         self.requested_container_type = xml_state.get_build_type_name()
+        self.base_image = None
+        image_uri = xml_state.build_type.get_derived_from()
+        if image_uri:
+            uri = Uri(image_uri, 'images')
```
schaefi: I was looking which repo_type 'images' classifies and found that we actually don't have a handling for it. What do you think about the idea to allow repo_type=None in the Uri constructor ? I have prepared a pull request to address this

kiwi/system/prepare.py
```diff
@@ -73,6 +73,7 @@ def __init__(
                 root_dir, image_uri, xml_state.build_type.get_image()
             )
             root_import.sync_data()
+            root_import.copy_image_file()
```
schaefi: why do we need to copy the base image file into the unpacked root ?

kiwi/container/docker.py
```diff
@@ -113,12 +116,24 @@ def create(self, filename):
...
+        if base_image:
+            uncompressor = Compress(base_image)
+            uncompressor.uncompress(True)
+            self.uncompressed_image = uncompressor.uncompressed_filename
+
+            Command.run([
+                'skopeo', 'copy',
+                'docker-archive:{0}'.format(self.uncompressed_image),
+                'oci:{0}'.format(container_name)
+            ])
```
schaefi: I don't understand why this step is needed ? iirc we unpack the base image, then the base image is copied as file into this unpacked directory and then it is uncompressed and turned into an oci image. Is that correct, and if so I'm clueless why we do that :) can you enlight me ?

schaefi:
Apart from the small Uri issue the code looks good. I just don't understand the workflow

kiwi/builder/container.py
schaefi: ok thanks for the review, it's merged you can change that code into

```python
uri = Uri(image_uri)
```

kiwi/container/docker.py
davidcassany: This is done to make sure we keep all the layers. We need to make a new layer on top of an existing image, so in that case the `umoci init --layout` and `umoci new --image` are not required. The sequence is `umoci unpack` the existing one, synchronize the modified rootfs, and finally  `umoci repack`. The `RootImport` unpacked that image before, but it doesn't exist anymore at the time, as it is unpacked to a temp folder.

Then I decided to keep the original image inside the /image because I wanted to make sure that the same image file is used in both cases. Imagine we run `system prepare` and `system create` as separate steps and the original image gets modified in between, because is remote, because the user deleted it... I wanted to leave an self-sufficient structure after `system prepare`

davidcassany: Actually this was the mentioned drawback of my proposal in the trello card, having to do such a strange movement, I am in favor of finding a nicer solution, but no ideas so far.

schaefi: ok so how about the following:

- RootImportDocker::sync_data()

   takes care to embedd the derived image as uncompressed file and also creates a checksum
   from that file

- ContainerImageDocker::create()

   checks if the checksum matches the derived image file and only operates on it if this is the
   case. This will not prevent possible inconsistencies between a prepare and create call but
   it offers enough protection in my eyes. I understand that we want an atomic operation, thanks
   for the clarification

After all I think we should write some code comments on the workflow, why we keep the derived image in the tree, basically because I think this information can not be taken from just reading the code

Thoughts ?

davidcassany: Ok, I like that checksum idea. Sure, some comments on why we keep the image file will be handy.

davidcassany:
The changes now should reflect your suggestion of using a checksum to validate the file, moreover, the image is kept uncompressed.

schaefi:
Thanks David, I did not had time to review your changes, many tel call meetings today. I'll take a look over the weekend or on Monday latest. Hope that's ok

davidcassany:
sure, next week is fine, no hurries

davidcassany:
[creation of derived docker images](https://trello.com/c/8pvTVbMX/200-creation-of-derived-docker-images)

kiwi/system/root_import/docker.py
```diff
@@ -63,10 +63,19 @@ def sync_data(self):
...
-            Path.wipe(self.uncompressed_image)
+        try:
+            if self.oci_layout_dir:
+                Path.wipe(self.oci_layout_dir)
+            if self.oci_unpack_dir:
+                Path.wipe(self.oci_unpack_dir)
+            if self.uncompressed_image:
+                Path.wipe(self.uncompressed_image)
+        except Exception:
+            pass
```
schaefi: A destructor should not raise an exception because it can't be handled. In this case we would just pass on which is kiwi v7 style ;) Actually you are calling Path.wipe which is an 'rm -f' which is call that normally never fails. Did you have a specific reason for the try block here ? If not let's remove that

kiwi/builder/container.py
```diff
@@ -57,6 +60,11 @@ def __init__(self, xml_state, target_dir, root_dir):
         self.target_dir = target_dir
         self.container_config = xml_state.get_container_config()
         self.requested_container_type = xml_state.get_build_type_name()
+        if not xml_state.build_type.get_derived_from():
+            self.base_image = None
+        else:
+            self.base_image = os.sep.join([root_dir, 'image', 'image_file'])
+
```
schaefi: I think we should change the way we retrieve the derived_from information. Actually the information in derived_from is an uri. Thus I suggest to write a method in xml_state.py which returns an uri object. e.g

```python
def get_derived_from_image_uri(self):
     ....code
```

* That method is then used in system/prepare.py and RootImport actually receives the Uri object
* That method is then called here and self.base_image is set via the Uri.translate()

I think that will make it easier for us to deal with different base image locations in the future. Additionally there will be only one place where we handle the result of derived_from

Thoughts ?

schaefi:
All the rest looks great to me 👍

kiwi/builder/container.py
```diff
@@ -57,6 +60,23 @@ def __init__(self, xml_state, target_dir, root_dir):
...
+            # and stored inside of the root_dir/image directory
+            self.base_image = os.path.normpath(
+                os.sep.join(
+                    [
+                        root_dir, 'image',
+                        base_image_origin.replace('.tar.xz', '')
+                    ]
+                )
+            )
+
```
schaefi: when I realized we do some path and file name assumptions here, I was worried if we are not creating a race condition under certain circumstances

kiwi/builder/container.py
```diff
@@ -57,6 +60,23 @@ def __init__(self, xml_state, target_dir, root_dir):
...
+        if self.base_image_uri:
+            base_image_origin = self.base_image_uri.translate()
+
+            # we expect the base image to be unpacked by the kiwi prepare step
+            # and stored inside of the root_dir/image directory
+            self.base_image = os.path.normpath(
+                os.sep.join(
+                    [
+                        root_dir, 'image',
+                        base_image_origin.replace('.tar.xz', '')
```
schaefi: I realized we do some path and file name assumptions here, and was a bit concerned if we are not creating unstable code under certain circumstances. Thoughts ?

isbm: How do you know it is `.xz`, not `.gz` or .`bz2`?

kiwi/system/root_import/docker.py
isbm: https://realpython.com/blog/python/the-most-diabolical-python-antipattern

kiwi/builder/container.py
```diff
@@ -57,6 +60,26 @@ def __init__(self, xml_state, target_dir, root_dir):
...
+            # a md5 file of the image is expected there too
+            self.base_image = os.path.normpath(
+                os.sep.join(
+                    [
+                        root_dir, 'image',
+                        base_image_origin.replace('.tar.xz', '')
+                    ]
+                )
+            )
+            self.base_image_md5 = ''.join([self.base_image, '.md5'])
```
isbm: Hmmm... this is so conceptual spaghetti! I would ask myself: if I already have a future image name in some sort of config and I end up with `base_image_origin` that contains some extension like `.tar.xz`, then why I am unable to do something like:
```python
# Somewhere earlier in your Universe
if not config.get('image_name'):
    raise Exception('Hey, please tell me where to generate the whole thing!')
# Construct the base_image_origin, based on configured archive
# I am assuming everything goes to the TAR first.
base_image_origin = '{0}.tar.{1}'.format(config.get('image_name'), config.get('archiver_of_the_day'))
# OK so far?

...

# Then here
self.base_image = os.path.join(root_dir, 'image', config.get('image_name'))
self.base_image_chksm = '{0}.{1}'.format(self.base_image, config.get('hashing_algorithm'))
```
I would not assume it is _always_ MD5. What if I want to use SHA256?.. I would not assume it is always one type of compression. What if I want to use BZip2? And so on.

kiwi/builder/container.py
```diff
@@ -80,13 +103,30 @@ def create(self):
...
+            container_setup = ContainerSetup(
+                self.requested_container_type, self.root_dir,
+                self.container_config
+            )
+            container_setup.setup()
+        else:
+            try:
+                checksum = Checksum(self.base_image)
+                with open(self.base_image_md5) as md5_file:
+                    base_image_md5 = md5_file.read()
```
isbm: I am Morgan Stanley. Or Nomura. So I want my checksums in SHA256. Or CosmicRay13. Or Nirvana777.

kiwi/builder/container.py
```diff
@@ -80,13 +103,30 @@ def create(self):
...
+            except Exception as e:
+                raise KiwiContainerBuilderError(
+                    '%s: %s' % (type(e).__name__, format(e))
```
isbm: We should stop using old formatters. They are gone with 2.6 version and we have `.format(...)` instead. Let's do this:
```python
raise KiwiContainerBuilderError('{name}: {reason}'.format(name=type(e).__name__,
                                                          reason=format(e)))

File: kiwi/builder/container.py [hunk 11]
```diff
@@ -80,13 +103,30 @@ def create(self):
...
+                )
+            if base_image_md5.split(' ')[0] != checksum.md5():
+                raise KiwiContainerBuilderError(
+                    'base image file {0} checksum validation failed'.format(
+                        self.base_image
```
isbm: And you _are_ using new style formatting. So we should be consistent then.

kiwi/system/root_import/docker.py
isbm: `except/pass` sucks, even you suppressing something. In your specific case this _has_ to fail so you will find out why: likely you screwed up on permissions in between, so you'd better make sure it doesn't really happen. At least this has to be quacked in the logs.

kiwi/builder/container.py
schaefi: yes you are right, I'm concerned too, we need to discuss this part

davidcassany: @schaefi in my first commit that name was forced, since to my eyes this is something it should not depend on any user choice or configurable parameter, I named that file with a constant name (`image_file`) to avoid any kind of name assumption. Maybe I should have included some comment as it was not obvious reading the code why it was named like that. In my initial commit the file were forced by code to be called `root_dir/image/image_file`  and `root_dir/image/image_file.md5`, in build step I was just checking the `derived_from` attribute to validate if those files were expected to be found or not.

I think this could solve the naming assumption. What you think?

schaefi: ah ok, so the prepare step always leaves the same file and that is a fix name we can rely on ? If that's the case I'm ok with it. But I would prefer if that name will be part of a method in the Defaults class, including the mentioned comment. Any code dealing with the file should take the name from this unique place. Would that be ok ?

kiwi/system/root_import/docker.py
davidcassany: All right `except/pass` this needs to be changed, as it can miss behave under some circumstances.

Just as side comment, note that `try/except/pass` block has nothing to do with the `Path.wipe`, I included it because this delete code gets executed when an Exception is thrown during the __init__ of the base class, in which case the `post_init` method of the derived class doesn't get executed meaning that the attributes evaluated on that method (`self.oci_unpack_dir` and so on) are not initialized at that point.

Just explaining it because you'll see that arranging that I will also refactor a little the root_import init method of the base classe, just in case you wonder why.

schaefi: good catch, I just wait for your refactoring to finish, hope I did not disturb your coding with my additional commits

kiwi/builder/container.py
davidcassany: @schaefi sure, totally agree, that file name should be in Defaults.

kiwi/system/root_import/docker.py
davidcassany: no problem at all, is good that all those questions arise

schaefi:
looks great to me now 👍

PR #267, opened 2017-03-08
Added Leap42.3 boot descriptions

schaefi:
Along with support for Leap42.3, openSUSE 13.2 is gone


schaefi:
[Create boot descriptions for Leap42.3](https://trello.com/c/iF8SBF3F/203-create-boot-descriptions-for-leap42-3)

schaefi:
correct, it was git which did that for me. It found out that some files are the same and marked them as renamed. I should have created extra commits but missed that step. Maybe I can revert that, let me check

schaefi:
Should be better now :)

PR #269, opened 2017-03-14
Update distribution matcher in spec file

schaefi:
Updates the distribution matchers as needed for the open buildservice

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -29,6 +29,11 @@
 %define distro suse-SLES12
 %endif
 
+# SLE13:
+%if 0%{?sle_version} == 130000 && !0%{?is_opensuse}
+%define distro suse-SLES13
+%endif
+
 # Leap 42.1(openSUSE product based on SLE 12SP1):
 %if 0%{?is_opensuse} && 0%{?sle_version} == 120100
 %define distro suse-leap42.1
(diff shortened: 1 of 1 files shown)


schaefi:
I'm missing SLES13

I'd appreciate a review from the obs team

package/python-kiwi-spec-template
```diff
@@ -19,27 +19,41 @@
...
 # SLE12:
 %if 0%{?suse_version} == 1315 && !0%{?is_opensuse}
 %define distro suse-SLES12
 %endif
-# Leap 42.1:
-%if 0%{?suse_version} == 1315 && 0%{?is_opensuse}
+
+# SLE13:
+%if 0%{?sle_version} == 130000
+%define distro suse-SLES13
```
DimStar77: Careful here - this is likely also to catch Leap 43.0

schaefi: ok how about it now ?

package/python-kiwi-spec-template
```diff
@@ -30,7 +30,7 @@
 %endif
 
 # SLE13:
-%if 0%{?sle_version} == 130000
+%if 0%{?sle_version} == 130000 && !0%{?is_opensuse}
```
DimStar77: LGTM

PR #270, opened 2017-03-15
Allow https location as repository source

schaefi:
a repository of the form

```xml
<repository type="yast2" imageinclude="true">
    <source path="https://download.opensuse.org/distribution/leap/42.2/repo/oss"/>
</repository>
```

was not allowed because the https locator was not know to kiwi's uri style

### DIFF
--- a/kiwi/system/uri.py
+++ b/kiwi/system/uri.py
@@ -113,7 +113,7 @@ class Uri(object):
             return self._suse_buildservice_path(
                 ''.join([uri.netloc, uri.path])
             )
-        elif uri.scheme == 'http' or uri.scheme == 'ftp':
+        elif uri.scheme == 'http' or uri.scheme == 'https' or uri.scheme == 'ftp':
             return ''.join([uri.scheme, '://', uri.netloc, uri.path])
         else:
             raise KiwiUriStyleUnknown(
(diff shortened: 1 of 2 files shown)


schaefi:
@davidcassany I tried building an image with the above repo setup but when package are installed in the second phase of the prepare step, meaning from within the so far created root system via chroot, zypper complains with

```
[ ERROR   ]: 17:25:58 | KiwiInstallPhaseFailed: System package installation failed: Media source 'https://download.opensuse.org/distribution/leap/42.2/repo/oss' does not contain the desired medium
```

Do you have an idea what might be missing in the chroot in order to allow zypper to work from within the chroot ? I couldn't find a solution, added ca-certificates and openssl in the bootstrap packages but couldn't came across a working solution

davidcassany:
@schaefi ca-certificates is not sufficient apparently, the package that actually contains all the certificates is `ca-certificates-mozilla`, including that one in the bootstrap solves the problem. In fact `ca-certificates-mozilla` requires `ca-certificates` which requires `openssl`, so it should be sufficient to only include the `ca-certificates-mozilla`.

schaefi:
Ah good to know, thanks, I'll report that on the ML too

PR #271, opened 2017-03-21
Convert root_dir passed to zypper/rpm --root to absolute path

Conan-Kudo:
Zypper on openSUSE 42.2 barfs when passed in root_dir as an argument for `--root` because it is always a relative path and Zypper doesn't like that. So, we convert it to an absolute path.

RPM also requires a similar change.

P.S.: @msrex sends his regards (we spent some time tracking this down while building images on openSUSE Leap 42.2)

* Use `os.path.abspath()` on `root_dir` being passed to `zypper --root`
* Use `os.path.abspath()` on `root_dir` being passed to `rpm --root`

### DIFF
--- a/kiwi/package_manager/zypper.py
+++ b/kiwi/package_manager/zypper.py
@@ -104,7 +104,7 @@ class PackageManagerZypper(PackageManagerBase):
         Process package install requests for bootstrap phase (no chroot)
         """
         command = ['zypper'] + self.zypper_args + [
-            '--root', self.root_dir,
+            '--root', os.path.abspath(self.root_dir),
             'install', '--auto-agree-with-licenses'
         ] + self.custom_args + self._install_items()
         return Command.call(
--- a/kiwi/repository/zypper.py
+++ b/kiwi/repository/zypper.py
@@ -205,7 +205,7 @@ class RepositoryZypper(RepositoryBase):
         self._backup_package_cache()
         Command.run(
             ['zypper'] + self.zypper_args + [
-                '--root', self.root_dir,
+                '--root', os.path.abspath(self.root_dir),
                 'addrepo',
                 '--refresh',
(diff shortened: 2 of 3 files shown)


Conan-Kudo:
@davidcassany There is no way for the user to provide paths for `--root`. These things that I'm forcing the paths to become absolute are generated by KIWI itself. Frankly, we're wondering how in the world it ever worked since we spent four hours debugging this to get it to work, leading to this PR and #272.

I don't know how to fix the unit-tests for either PR.

davidcassany:
@Conan-Kudo right, let me make some trials about it, I think that forcing absolute paths for prepare, create or build tasks should be sufficient. 

You using relative paths in the command line, right? Does it work for you if using absolute paths?

msrex:
@davidcassany We used relative paths in the kiwi command line.

schaefi:
It should definitely work with relative and absolute paths, turning any path into its abspath seems a good solution if there are also tests covering both cases

davidcassany:
@scahefi right, then I would go for turning relative paths to absolute just at the time of parsing the user input flags. I mean inside `kiwi/tasks/*` to make sure they are absolute when they are passed to `kiwi/system/*` classes or any other class that actually performs the work. 

What you think?

schaefi:
sounds good to me :+1:

msrex:
I experimented with building yum-based images. There we have the same issue with absolute paths being required for --root.

davidcassany:
@schaefi ok, I am on it. 

@Conan-Kudo, @msrex with you permit I'll provide a patch for that as discussed above, this way it will be also easier to make the unit tests.

schaefi:
@msrex yes the issue should be the same for all targets, David will create a proper fix. Handling that at the entry point level will most probably also fix #272

Conan-Kudo:
@schaefi Sounds good to me.

PR #272, opened 2017-03-21
Pass full path to initrd file for compressing

Conan-Kudo:
At initrd compress time, host xz is being used, so the path expected is relative to the host filesystem structure. Thus, we pass the full path.

* Use full path for initrd file path to pass to xz for compressing initrd

### DIFF
--- a/kiwi/boot/image/builtin_kiwi.py
+++ b/kiwi/boot/image/builtin_kiwi.py
@@ -142,6 +142,6 @@ class BootImageKiwi(BootImageBase):
             log.info(
                 '--> xz compressing archive'
             )
-            compress = Compress(self.initrd_file_name)
+            compress = Compress(temp_boot_root_directory + "/" + self.initrd_file_name)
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@davidcassany Well, we can't get far enough to reach this point without the fixes in #271, so I'd have to say no, I didn't encounter this until after applying the fixes from #271.

msrex:
@davidcassany This is not related to #271 . The compression stage for the boot image simply looks in the wrong spot: the boot image is created in temp_boot_directory, and the compression looks in the final image path. No absolute paths to external commands are used, which is what #271 is about.

schaefi:
imho with the Fix from David in #278 this one is fixed too. self.initrd_file_name is build from target_dir which is now handed over as abs_path to all objects using it

Could you confirm ?

Conan-Kudo:
@schaefi The merged fix from #278 also fixes this issue, so closing.

PR #273, opened 2017-03-22
Update flags attribute and its documentation

schaefi:
Delete the seed value as it is not implemented and won't be
supported in the future. Properly document the existing
overlay methods


schaefi:
yes :)

PR #278, opened 2017-03-24
Convert user input to absolute paths

davidcassany:
This commit converts the paths provided by the user with the
command line to absolute paths. The effected arguments are:
* --root
* --target-dir
* --bundle-dir

This supersedes and fixes #271


schaefi:
great :+1:

PR #280, opened 2017-03-27
Refactor method name of package manager interface

### LINKED ISSUE #248: cleanup package manager lock requests
## Problem description

The package manager interface provides a method called __request_package_lock__ which is called that way because of zypper's add-lock command. However the interface should be generic over all package manager implementation and the concept of a lock is only available to zypper. In fact such a lock is not more than excluding the requested packages from the solver operation (with all its consequences)

Thus this issue should solve:

* better name for request_package_lock e.g request_package_exclusion
* better name for self.lock_requests e.g self.exclude_requests
* fixup of the comment in process_install_requests which describes what a lock request does

## Actual behaviour

All works fine, but the implementation is inconsistent

schaefi:
The request_package_lock was renamed into request_package_exclusion
because that is the goal, to actually exclude(skip) a package. From
an implementation point of view this is done to set a lock in the
zypper case. However other package managers might do it differently.
The interface should stay consistent with regards to the user goal
and not with the package manager specific implementation. This
Fixes #248

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -93,16 +93,16 @@ class PackageManagerApt(PackageManagerBase):
             'Product(%s) handling not supported for apt-get', name
         )
 
-    def request_package_lock(self, name):
+    def request_package_exclusion(self, name):
         """
-        Queue a package lock(ignore) request
+        Queue a package exclusion(skip) request
 
-        There is no locking mechanism for the apt package manager
+        Package exclusion for apt package manager not yet implemented
 
         :param string name: unused
         """
         log.warning(
-            'Package locking for (%s) not supported for apt-get', name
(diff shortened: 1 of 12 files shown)


Conan-Kudo:
As an additional note, both Yum and DNF implement excludes with `--exclude <pkgname>` with `<pkgname>` being a name or a glob. You can stack multiple `--exclude` options and they'll be processed in an additive manner. That is, `--exclude pkgA --exclude pkgB*` will exclude both `pkgA` and all packages matching `pkgB*` glob. If you want to, you can implement them as part of this PR.

schaefi:
Thanks I'll put that information into a new issue, so we can add it asap

schaefi:
added an issue:

* https://github.com/SUSE/kiwi/issues/282

PR #283, opened 2017-03-30
Delete obsolete ec2 disk format

schaefi:
The ec2 value in the format attribute is deprecated and no
longer supported It remained in the schema for a while and
has now reached EOL


schaefi:
With the introduction of the 4.4 kernel in openSUSE and SLE the SUSE Xen patches were all upstremead and we enabled the so called "pvops kernel" as the -default kernel. With this the behavior for PV images in EC2 changes and we can now treat PV and HVM as equals.

Thus only one ec2 firmware type will stay in the schema for this next generation of kiwi

PR #284, opened 2017-03-30
Include SLE13 boot descriptions for x86_64 and arm

davidcassany:
This commit includes boot descriptions for SLE13, the descriptions
are based on leap 42.3.


kiwi/boot/arch/arm/oemboot/suse-SLES13/config.sh
```diff
@@ -0,0 +1,51 @@
...
+
+#======================================
+# SuSEconfig
+#--------------------------------------
+suseConfig
+
+#==========================================
+# setup gfxboot
+#------------------------------------------
+suseGFXBoot openSUSE grub
```
davidcassany: I realized that on PPC and s390 this part is `suseGFXBoo SLES grub` instead of openSUSE. I am just wondering if this is done in purpose or is a simple copy and paste mistake.

schaefi: This is on purpose, because for s390 and ppc only SLES is relevant

schaefi: The theme name provided as argument to the method is just a default theme name. The value only applies if the XML config does not provide a theme name. However if no theme name is provided the next generation kiwi also generates a boot setup based on textmode and thus the value has no effect. So this is basically for historical reasons and to be sure there is always a default theme name.

The code around suseGFXBoot could be refactored but so far I just left it untouched because it just works :)

schaefi:
Overall looks good to me, except for the little typo in the commit message "fro" vs. "for"

maybe you can reword that as it will appear in the package changelog later

Thanks

davidcassany:
Right, just amended the commit comment.

schaefi:
nice 👍

PR #288, opened 2017-04-03
Fixup grub2 setup and tool calls

schaefi:
Depending on the distribution the grub tools are either named
grub2-tool or grub-tool. Additionally the grub configuration
data is expected to live in boot/grub2 or boot/grub. This commit
handles the tool calls and also the grub boot directory setup
in a generic way

### DIFF
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -298,12 +298,15 @@ class BootLoaderConfigBase(object):
                 filesystem = self.xml_state.build_type.get_filesystem()
                 volumes = self.xml_state.get_volumes()
                 if filesystem == 'btrfs' and volumes:
-                    # boot or boot/grub2 on a subvolume prevents grub from
-                    # finding its config file
+                    # grub boot data paths must not be in a subvolume
+                    # otherwise grub won't be able to find its config file
+                    grub2_boot_data_paths = ['boot', 'boot/grub', 'boot/grub2']
                     for volume in volumes:
-                        if volume.name == 'boot' or volume.name == 'boot/grub2':
+                        if volume.name in grub2_boot_data_paths:
                             raise KiwiBootLoaderTargetError(
-                                'boot or boot/grub2 must not be a subvolume'
+                                '{0} must not be a subvolume'.format(
+                                    volume.name
+                                )
                             )
 
         return bootpath
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -17,7 +17,6 @@
 #
 import os
 import platform
-import shutil
 from collections import OrderedDict
 
 # project
@@ -31,7 +30,7 @@ from kiwi.path import Path
 from kiwi.utils.sync import DataSync
 from kiwi.utils.sysconfig import SysConfig
 
-from ...exceptions import (
+from kiwi.exceptions import (
     KiwiTemplateError,
     KiwiBootLoaderGrubPlatformError,
     KiwiBootLoaderGrubModulesError,
@@ -145,7 +144,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         self.grub2 = BootLoaderTemplateGrub2()
         self.config = None
         self.efi_boot_path = None
-        self.boot_directory_name = 'grub2'
+        self.boot_directory_name = self._get_grub2_boot_directory_name()
         self.cmdline_failsafe = None
         self.cmdline = None
         self.iso_efi_boot = False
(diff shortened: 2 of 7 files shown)


schaefi:
Needs further integration testing, please do not yet merge

kiwi/bootloader/config/grub2.py
```diff
@@ -716,3 +719,24 @@ def _find_grub_data(self, lookup_path):
...
+
+        Depending on the distribution the grub boot path could be
+        either boot/grub2 or boot/grub. The method will decide for
+        the correct base directory name according to the name pattern
+        of the installed grub tools
+        """
+        chroot_env = {
+            'PATH': os.sep.join([self.root_dir, 'usr', 'sbin'])
+        }
+        if Path.which(filename='grub-install', custom_env=chroot_env):
```
Conan-Kudo: You'll want to test for grub2-* first, since some distros ship both grub legacy and grub2, which would confuse this test

kiwi/bootloader/config/base.py
```diff
@@ -298,12 +299,17 @@ def get_boot_path(self, target='disk'):
                 filesystem = self.xml_state.build_type.get_filesystem()
                 volumes = self.xml_state.get_volumes()
                 if filesystem == 'btrfs' and volumes:
-                    # boot or boot/grub2 on a subvolume prevents grub from
-                    # finding its config file
+                    # grub boot data paths must not be in a subvolume
+                    # otherwise grub won't be able to find its config file
+                    grub_boot_data_paths = [
+                        'boot', os.sep.join(['boot', 'boot/grub', 'boot/grub2'])
```
Conan-Kudo: This doesn't make sense? The `os.sep.join()` call makes nonsense output:

```
Python 3.5.3 (default, Mar 21 2017, 17:21:33) 
[GCC 6.3.1 20161221 (Red Hat 6.3.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.sep.join(['boot', 'boot/grub', 'boot/grub2'])
'boot/boot/grub/boot/grub2'

```

schaefi: right, that happens if two thoughts come together where only one was correct :) Thanks for finding this

schaefi:
Successfully build and booted test images for ubuntu, centos, fedora and suse from here:
https://github.com/SUSE/kiwi-descriptions

kiwi/bootloader/config/base.py
```diff
@@ -298,12 +298,15 @@ def get_boot_path(self, target='disk'):
                 filesystem = self.xml_state.build_type.get_filesystem()
                 volumes = self.xml_state.get_volumes()
                 if filesystem == 'btrfs' and volumes:
-                    # boot or boot/grub2 on a subvolume prevents grub from
-                    # finding its config file
+                    # grub boot data paths must not be in a subvolume
+                    # otherwise grub won't be able to find its config file
+                    grub_boot_data_paths = ['boot', 'boot/grub', 'boot/grub2']
```
Conan-Kudo: I'm not sure if this test is ordered in the way it is written, but if it is, you should always test for `boot/grub2` before `boot/grub`, because it's possible for distros to ship grub legacy and grub2 (Debian, Mageia, and a few others do).

I would also suggest naming all variables `grub2_<foo>` to make it clear grub2 is the only thing supported.

kiwi/bootloader/config/base.py
```diff
@@ -298,12 +298,15 @@ def get_boot_path(self, target='disk'):
...
                     for volume in volumes:
-                        if volume.name == 'boot' or volume.name == 'boot/grub2':
+                        if volume.name in grub_boot_data_paths:
```
Conan-Kudo: I would also suggest naming the variable `grub2_boot_data_paths` to make it clear grub2 is the only thing supported.

kiwi/bootloader/config/grub2.py
```diff
@@ -145,7 +144,7 @@ def post_init(self, custom_args):
         self.grub2 = BootLoaderTemplateGrub2()
         self.config = None
         self.efi_boot_path = None
-        self.boot_directory_name = 'grub2'
+        self.boot_directory_name = self._get_grub_boot_directory_name()
```
Conan-Kudo: Likewise here, `grub2` instead of `grub`.

kiwi/bootloader/config/base.py
schaefi: The order in this case is immaterial as we are just looking that a subvolume in the btrfs case is not pointing to any possible grub2 boot path. So this is basically to check for a limitation of grub2's btrfs support and not for the grub2 installation in the system

I agree to the variable naming and will adapt it accordingly

schaefi:
all right I think we are good to go, thoughts ?

PR #292, opened 2017-04-06
Keep imported image in OCI format instead of docker

davidcassany:
Kiwi always uses OCI format for container manipulations, so it is
easier to assume the image kept between prepare and create step
is also in OCI format, this way less format conversions are needed.


schaefi:
That makes a lot sense 👍

PR #294, opened 2017-04-06
Added support for unknown uri types for base image references

davidcassany:
This commits bypasses any URI check if the uri schema is unknown in
RootImport class. This way the URI is bypassed to skopeo if it couldn't
be translated to any known type. That enables referencing images with
any URI supported by skopeo, i.e. DockerHub images.

### DIFF
--- a/kiwi/system/root_import/base.py
+++ b/kiwi/system/root_import/base.py
@@ -19,7 +19,10 @@ import os
 
 # project
 from kiwi.utils.checksum import Checksum
-from kiwi.exceptions import KiwiRootImportError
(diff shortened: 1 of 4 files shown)


schaefi:
I'm ok with the change, but I think we should show a warning message that kiwi is not able to cope with this uri and that we pass it along to another tool and hope that one can cope with it. Thoughts ?

davidcassany:
I agree, some logging would improve the UX.

davidcassany:
I have included a generic warning message in the base class and another one in docker class stating the URI is direclty bypassed to skopeo tool.

PR #295, opened 2017-04-10
Add method to setup installation of recommended packages

### LINKED ISSUE #285: Kiwi doesn't respect patternType="plusRecommended"
## Problem description
I marked packages section with `patternType="plusRecommended"` what I would expect to happen is that Kiwi installs pattern-related recommends while still ignoring package-related recommends. 

## Additional info
Zypper backend

## Actual behaviour
Kiwi ignores `plusRecommended` and installs only required pattern packages.

1. Create pattern package spec file with a few `Recommends:`
2. Mark `<packages type="image" patternType="plusRecommended">`
3. Build

davidcassany:
With this commit package manager are always set to include only
required packages or set to include required and recommended
packages.

This commit fixes #285

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -214,7 +214,15 @@ class PackageManagerApt(PackageManagerBase):
         """
         Setup package processing only for required packages
 
-        There is no required/recommends information in the rhel repo data
+        Kiwi has no required/recommends support for apt manager yet
+        """
+        pass
+
+    def process_plus_recommended(self):
+        """
+        Setup package processing to also include recommended dependencies
+
+        Kiwi has no required/recommends support for apt manager yet
         """
         pass
 
--- a/kiwi/package_manager/dnf.py
+++ b/kiwi/package_manager/dnf.py
@@ -191,7 +191,15 @@ class PackageManagerDnf(PackageManagerBase):
         """
         Setup package processing only for required packages
 
-        There is no required/recommends information in the rhel repo data
+        Kiwi has no required/recommends support for dnf manager yet
+        """
+        pass
+
+    def process_plus_recommended(self):
+        """
+        Setup package processing to also include recommended dependencies
+
+        Kiwi has no required/recommends support for dnf manager yet
         """
         pass
 
--- a/kiwi/package_manager/base.py
+++ b/kiwi/package_manager/base.py
@@ -154,6 +154,14 @@ class PackageManagerBase(object):
         """
         raise NotImplementedError
 
+    def process_plus_recommended(self):
+        """
+        Setup package processing to also include recommended dependencies
+
+        Implementation in specialized package manager class
+        """
+        raise NotImplementedError
+
     def match_package_installed(self, package_list, log_line):
         """
         Match expression to indicate a package has been installed
--- a/kiwi/package_manager/yum.py
+++ b/kiwi/package_manager/yum.py
@@ -193,6 +193,14 @@ class PackageManagerYum(PackageManagerBase):
         """
         pass
 
+    def process_plus_recommended(self):
+        """
+        Setup package processing to also include recommended dependencies
+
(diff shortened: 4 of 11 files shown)


davidcassany:
Since package manager instance is the same for bootstrapping and chrooted installs, the default configuration for chrooted install is the same as the bootstrap packages configuration, as nothing reverts it.  Since there was not method or mechanism to remove the `--no-recommends` configured during bootstrap it was also included in chrooted installations (even setting plusRecommended).

schaefi:
Ah yes good catch, thanks 👍

Conan-Kudo:
@davidcassany There is, in fact, recommends/supplements/etc for DNF. I think it might be currently enabled. The way to switch it off via CLI is `--setopt=install_weak_deps=False` and it is possible to set in the`dnf.conf` using the same thing in `dnf.conf`.

More info: http://dnf.readthedocs.io/en/latest/conf_ref.html

kiwi/package_manager/apt.py
```diff
@@ -218,6 +218,14 @@ def process_only_required(self):
         """
         pass
 
+    def process_plus_recommended(self):
+        """
+        Setup package processing to also include recommended dependencies
+
+        There is no required/recommends information in the rhel repo data
```
Conan-Kudo: apt is Debian, not RHEL. :)

kiwi/package_manager/dnf.py
```diff
@@ -195,6 +195,14 @@ def process_only_required(self):
         """
         pass
 
+    def process_plus_recommended(self):
+        """
+        Setup package processing to also include recommended dependencies
+
+        There is no required/recommends information in the rhel repo data
```
Conan-Kudo: It's Fedora, not RHEL, and yes, there is Recommends/Supplements, etc..

davidcassany:
@conan-kudo, yes I am aware of it, also it exists in apt and possibly with yum (I haven't checked), however I though it should be another PR, since it would not be a bug fix, but an enhancement, as there was also no support for required only for yum, apt and dnf.

kiwi/package_manager/dnf.py
davidcassany: Ups you are right, I am on my way to correct it. Copy & Paste is evil...

Conan-Kudo:
@davidcassany Here I disagree. When you implement *new* functionality (as this is), you should always ensure parity whenever possible. Otherwise, what's the point?

davidcassany:
@Conan-Kudo, I'll make a quick check to see if I can implement required/recommends for dnf/apt/yum as simple as in zypper (I think it is for apt/dnf, not sure about yum) and if so include it. I don't wanna block #285.

Conan-Kudo:
@davidcassany Yum controls it with a `requires_policy` flag. The value `strong` does only Requires, `weak` does Requires and Recommends. It can be set via CLI using `--setopt=requires_policy=weak` to turn on weak deps.

More info: http://man7.org/linux/man-pages/man5/yum.conf.5.html

schaefi:
I like small steps, we are not in a hurry to implement recommends support for all package managers in one pull request. Regarding this one it fixes a bug not correctly handling recommends from an interface point of view which was imho good and correctly handled. That state could be merged to cleanup the existing interface issue. A clean implementation for the missing recommends handling could follow one after the other. Even better this effort could be shared to more than one developer ;)

I'm ok if David wants to nail it all down in one blob but I think small steps reduces the risk to overlook errors

davidcassany:
@schaefi I agree, I will add enhancement issues for recommends in other package managers in other to track it and link with that dnf/yum management explanations from @Conan-Kudo.

Then, unless there is something you think it can be improved, I'd say we are fine here.

schaefi:
looks good to me, we have those kind of issue reports for the package exclude support as well. imho those should be addressed in the next weeks

PR #296, opened 2017-04-10
Uefi live builder fix

schaefi:
The live iso builder still used a fixed boot/grub2 path  but due to the refactoring this path is now distro dependant and should be obtained dynamically

Additionally some cleanup in the boot code regarding the use of the bootloader path has been done.
This is just a start and could be extended with regards to other distros in the future

### DIFF
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -18,6 +18,8 @@
 from tempfile import mkdtemp
 import platform
 import shutil
+import glob
+import os
 
 # project
 from kiwi.bootloader.config import BootLoaderConfig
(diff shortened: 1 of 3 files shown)


kiwi/builder/live.py
```diff
@@ -240,10 +242,11 @@ def create(self):
             if self.firmware.efi_mode() == 'uefi':
                 # write bootloader config to EFI directory in order to allow
                 # grub loaded by shim to find the config file
-                shutil.copy(
-                    self.media_dir + '/boot/grub2/grub.cfg',
-                    self.media_dir + '/EFI/BOOT'
+                grubconf_glob = os.sep.join(
+                    [self.media_dir, 'boot/grub*/grub.cfg']
```
Conan-Kudo: What happens if both paths exist, which can happen for a variety of reasons? This seems like it'd work out pretty badly. You should really test for the versioned one first, and then the unversioned one if the former doesn't exist.

schaefi: This can not happen here, because the glob search runs on the self created temporary media_dir. The decision for grub2 or grub has been made in the prior bootloader class and there we already respect the order as you suggested in a former conversation

Conan-Kudo: Okay, then. Just checking. :)

👍

PR #297, opened 2017-04-10
Rename SLE13 to SLE15

davidcassany:
This PR renames SLE13 to SLE15, as it has been decided that there won't be any SLE13 or SLE14 due to their meanings or connotations in some cultures.


schaefi:
We also need to adapt the spec file template in kiwi/package/python-kiwi-spec-template

davidcassany:
I missed that one, now should be fine

schaefi:
all right let's hope those people doesn't come up with new bright name ideas ;)

PR #299, opened 2017-04-11
Bypass bootstrap procedure if no bootstrap packages section

davidcassany:
This commit updates kiwi to bypass bootstrap procedure in case
there is no bootstrap packages section.

It also fixes little related issues in root_bind and apt manager.

### DIFF
--- a/kiwi/system/root_bind.py
+++ b/kiwi/system/root_bind.py
@@ -88,9 +88,14 @@ class RootBind(object):
         """
         try:
             for location in self.bind_locations:
-                if os.path.exists(location):
+                location_mount_target = os.path.normpath(os.sep.join([
+                    self.root_dir, location
+                ]))
+                if os.path.exists(location) and os.path.exists(
+                    location_mount_target
+                ):
                     shared_mount = MountManager(
-                        device=location, mountpoint=self.root_dir + location
+                        device=location, mountpoint=location_mount_target
                     )
                     shared_mount.bind_mount()
                     self.mount_stack.append(shared_mount)
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -155,6 +155,11 @@ class SystemPrepare(object):
         Install system software using the package manager
         from the host, also known as bootstrapping
         """
+        if not self.xml_state.get_bootstrap_packages_sections():
+            log.warning('No <packages> sections marked as "bootstrap" found')
+            log.info('Processing of bootstrap stage skipped')
+            return
+
         log.info('Installing bootstrap packages')
         bootstrap_packages = self.xml_state.get_bootstrap_packages()
         bootstrap_packages.append(
--- a/test/unit/system_prepare_test.py
+++ b/test/unit/system_prepare_test.py
@@ -267,6 +267,16 @@ class TestSystemPrepare(object):
         mock_tar.assert_called_once_with('../data/bootstrap.tgz')
         tar.extract.assert_called_once_with('root_dir')
 
+    @patch('kiwi.logger.log.warning')
+    @patch('kiwi.xml_state.XMLState.get_bootstrap_packages_sections')
+    def test_install_bootstrap_skipped(
+        self, mock_bootstrap_section, mock_log_warning
+    ):
+        mock_bootstrap_section.return_value = []
+        self.system.install_bootstrap(self.manager)
+        mock_bootstrap_section.assert_called_once_with()
+        assert mock_log_warning.called
+
     @patch('kiwi.xml_state.XMLState.get_bootstrap_collection_type')
     @patch('kiwi.system.prepare.CommandProcess.poll_show_progress')
     @patch('kiwi.system.prepare.ArchiveTar')
--- a/test/unit/system_root_bind_test.py
+++ b/test/unit/system_root_bind_test.py
@@ -70,8 +70,10 @@ class TestRootBind(object):
         self.bind_root.setup_intermediate_config()
         mock.cleanup.assert_called_once_with()
 
+    @patch('kiwi.system.root_bind.os.path.exists')
     @patch('kiwi.system.root_bind.MountManager')
-    def test_mount_kernel_file_systems(self, mock_mount):
+    def test_mount_kernel_file_systems(self, mock_mount, mock_exists):
+        mock_exists.return_value = True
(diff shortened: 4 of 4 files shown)


davidcassany:
With this change derived images are more consistent, since they might not require any kind of bootstrapping. With this PR derived images from oficial DockerHub images of openSUSE, Fedora, Ubuntu and Centos are possible.

kiwi/package_manager/apt.py
```diff
@@ -145,28 +145,24 @@ def process_install_requests_bootstrap(self):
...
     def process_install_requests(self):
         """
         Process package install requests for image phase (chroot)
         """
+        update_command = ['chroot', self.root_dir, 'apt-get']
+        update_command.extend(self.root_bind.move_to_root(self.apt_get_args))
+        update_command.extend(self.custom_args)
+        update_command.append('update')
+        Command.run(update_command, self.command_env)
+
```
schaefi: why has the bootstrap update call moved into the install procedure ? The reason why 'update' was called once in the bootstrap phase is because after debootstrap you need to update the apt cache before installing further packages from within the chroot. That call happens only once, now that it is part of the install requests it happens every time a package install request is processed inside of the chroot.

* I'd like to understand the motivation of the change
* I'd prefer an extra commit for this change

Could you share those details ? Thanks

kiwi/system/root_bind.py
```diff
@@ -88,9 +88,12 @@ def mount_kernel_file_systems(self):
         """
         try:
             for location in self.bind_locations:
-                if os.path.exists(location):
+                mount_point = os.path.normpath(os.sep.join([
```
schaefi: I'm not sure under which circumstances one of self.bind_locations does not exist in the image root directory but I'm ok with the extra check. For a better understanding what mount_point is could we use a more expressive name, e.g

```python
location_mount_target = os.path.normpath(...
```

Thoughts ?

kiwi/tasks/system_build.py
```diff
@@ -155,11 +155,12 @@ def process(self):
...
+
         manager = system.setup_repositories()
-        system.install_bootstrap(manager)
-        system.install_system(
-            manager
-        )
+        if self.xml_state.get_bootstrap_packages_sections():
+            system.install_bootstrap(manager)
+        system.install_system(manager)
+
```
schaefi: I think it would be better to handle the condition if bootstrapping should happen or not in the method install_bootstrap in system/prepare.py and not at several places in the code. I also think a warning or info message that bootstrapping will be skipped because no bootstrap packages were specified is important. We also should care for the fact that packages are not the only part of a bootstrap section:

* there could be bootstrap archives
* there could be bootstrap collections
* there could be bootstrap products

It should also be clear that bootstrap is only optional if a base image is given which contains what bootstrap normally does. So just skipping the entire step by the condition of an empty bootstrap package list is too short
thinking in my eyes.

kiwi/tasks/system_prepare.py
```diff
@@ -148,11 +148,12 @@ def process(self):
...
+
         manager = system.setup_repositories()
-        system.install_bootstrap(manager)
-        system.install_system(
-            manager
-        )
+        if self.xml_state.get_bootstrap_packages_sections():
+            system.install_bootstrap(manager)
+        system.install_system(manager)
+
```
schaefi: see above comment

kiwi/system/root_bind.py
davidcassany: It turned out that some images did not have all the default kernel file systems folders (now I don't remember if it was CentOS or Fedora), thus is was breaking due to a non existing mountpoint.

I agree with the mountpoint name change. I am gonna change that.

kiwi/package_manager/apt.py
davidcassany: Ok, I agree this can be sorted in a separate commit. About the motivation, apt doesn't have an automatic refresh system as it happens in zypper, dnf or yum. So, if the boostrap is not executed it turns that packages are not installable because the debian repository database is not populated yet.

Maybe I missed something but I believe this is the same behavior as we have with zypper using `--refresh,` a repository database update is run before every zypper install. I checked that `apt-get update` doesn't re-download data if it is already up to date, like zypper does.

kiwi/tasks/system_build.py
davidcassany: My initial intention was to let bootstrap run normally as we have been doing up to now (if not boostrap section it shouldn't hurt, just verify package manager is there and not much more), however it turned out that the debian boostraping tool is not really re-entrant and I didn't want to code any strange check to guess if the bootstrap was already there or not. That's why I came with the idea of not even running bootstrap if not stated in the description file.

I think all three cases you noted above are covered with the condition I wrote, since `xml_state.get_bootstrap_packages_sections` returns any `<packages type="bootstrap">`.

Said that I am totally opened to modify that condition or find another solution.

kiwi/package_manager/apt.py
schaefi: ok if we don't create a performance issue I'm fine with it, thanks for the clarification

kiwi/tasks/system_build.py
schaefi: I agree the condition check is correct, did not look close enough to see that this is checking for all <packages> section marked as bootstrap. I still think the check should be done once in the SystemPrepare class: e.g

```
diff --git a/kiwi/system/prepare.py b/kiwi/system/prepare.py
index fa5a2730..63f99c31 100644
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -155,6 +155,11 @@ class SystemPrepare(object):
         Install system software using the package manager
         from the host, also known as bootstrapping
         """
+        if not self.xml_state.get_bootstrap_packages_sections():
+            log.warning('No <packages> sections marked as "bootstrap" found')
+            log.info('Processing of bootstrap stage skipped')
+            return
+
         log.info('Installing bootstrap packages')
         bootstrap_packages = self.xml_state.get_bootstrap_packages()
         bootstrap_packages.append(
```

schaefi: as result no tasks code needs to be changed

davidcassany:
Now it should include all the amendments discussed above.

Apt package manager changes have been ported to another PR.

schaefi:
👍 thanks

PR #300, opened 2017-04-11
Make sure debian repositories database is populated before install

davidcassany:
This commit includes an `apt-get update` call before any 
`apt-get install` command. This way the packages database is
always ready, even if no bootstrap procedure has been executed.


schaefi:
great thanks 👍

PR #301, opened 2017-04-12
Add clear attribute for entrypoint and subcommand sections

davidcassany:
This commit adds the possibility of clearing any subcommand or
entrypoint. This is relevant for docker derived images, as they
inherit the configuration and it might lead to some bad behavior.

See how it is used [here](https://github.com/davidcassany/kiwi-derived-images)

### DIFF
--- a/doc/source/schema.rst
+++ b/doc/source/schema.rst
@@ -369,6 +369,7 @@ Children:
 List of attributes for ``entrypoint``:
 
 * ``execute`` : Specifies the entry point program name to execute
+* ``clear`` : Specifies to clear or not some data or configurations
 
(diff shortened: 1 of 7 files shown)


schaefi:
:+1: thanks 

I was worrying if the name of the attribute is kind of self explaining but the docker tooling uses the same option name so I think it's ok to go with it

schaefi:
One thing I noticed was that the schema docs was not updated regarding the schema change

davidcassany:
@schaefi you are absolutely right, I forgot to re run the docs generator, I will include it

davidcassany:
@schaefi, now should be fine

schaefi:
yep, thanks

PR #305, opened 2017-04-14
Do not purge the repositories before inserting them

### LINKED ISSUE #304: yum repo, dnf repo and apt repo-files are deleted in image if imageinclude=true is used
## Problem description

This happens only when using yum, dnf or apt as repo-type (basically everybody besides zypper I guess...).

If you include any repo and mark it with imageinclude=true, all default repos in `/etc/yum.repos.d/` and `/etc/apt-get/sources-list.d/` are deleted. This includes the main repo and the updates repo etc. - the only one remaining is the repo marked with imageinclude=true.

**THIS IS SERIOUSLY NOT GOOD!**

It is especially bad as Fedora/RHEL/Centos include their repo-definition files as a file inside an rpm (check the contents package `epel-release.rpm`). That leads to package verification failures as the repo definition has changed or is missing. Also some rpms contain more than one repo (e.g. the Centos main repo-definition rpm). In the middle of #303 @Conan-Kudo explains a bit more.

Actually this issue only came to light after fixing #303 - before then the repo-definitions in the wrong place were deleted so all appeared fine.

The different behaviour of `yum`/`dnf` and `zypper` is that `yum`/`dnf` simply overwrite any existing repo whereas `zypper` complains, hence we are very aggressive in deleting all repos (class `SystemSetup` function `import_repositories_marked_as_imageinclude`).

Looking at `RepositoryZypper` and the `addrepo`-function this deletion seems to be unnecessary as it removes an existing file there anyways, if `zypper` is used. For the other package managers the deletion is not needed as they don't complain.

## Actual behaviour

See above.

## Reproduction
-  set up a Kiwi builder with v9.4.11 WITHOUT applying #303
1. create an image (Centos in this example) with @schaefi 's JeOS definition
2. go in the image-root (no need to boot) and check the files in `/etc/yum.repos.d`. Those files are the definitions for binary, update and a few other repos
3. if you want to go fancy verify them with `rpm -V centos-release` (chroot!)

- Now apply #303 and repeat. You will still see all files there as long as you have no repos marked with `imageinclude="true"`

- Next use a Kiwi builder **without** #303
1. change @schaefi 's Centos JeOS definition and give any repo the flag `imageinclude="true"` and build
2. go into the image-root and check that the files in `/etc/yum.repos.d` are still there. Also you will now find a `.repo` file with the contents you marked as `imageinclude="true"` in `/etc/yum/repos.d` (the wrong location per #303!).

- Finally apply #303 again
1. build the JeOS from above with a repo having the flag `imageinclude="true"`
2. go into the image-root and discover to your horror that `/etc/yum.repos.d` is empty except for the one repo-file that you specified
3. be fancy in your horror and verify the rpm with `rpm -V centos-release`

Conan-Kudo:
There are no good reasons to be purging the repo directories, especially when it is common for some distributions (Red Hat/CentOS/Fedora, for example) to ship repository configuration as packages. Deleting them puts the package manager in the system into a weird state, so we want to avoid this.

Fixes #304.

* Remove `repo.delete_all_repos()` call in `import_repositories_marked_as_imageinclude()` in `SystemSetup()` class.

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -123,7 +123,6 @@ class SystemSetup(object):
             RootBind(root), self.xml_state.get_package_manager()
         )
         repo.use_default_location()
-        repo.delete_all_repos()
         for xml_repo in repository_sections:
             repo_marked_for_image_include = xml_repo.get_imageinclude()
 
--- a/test/unit/system_setup_test.py
+++ b/test/unit/system_setup_test.py
@@ -686,7 +686,6 @@ class TestSystemSetup(object):
         repo = mock.Mock()
         mock_repo.return_value = repo
         self.setup_with_real_xml.import_repositories_marked_as_imageinclude()
-        repo.delete_all_repos.assert_called_once_with()
         repo.add_repo.assert_called_once_with(
             '95811799a6d1889c5b2363d3886986de',
             'http://download.opensuse.org/repositories/Devel:PubCloud:AmazonEC2/SLE_12_GA',


schaefi:
ok, I thought it was a good idea to cleanup before adding the configured repos, but it seems it's better to keep what the distro packages added

Thanks

PR #308, opened 2017-04-19
Add support for OCI images

davidcassany:
This Pull Request adds support for OCI images. As Docker and OCI images are managed in parallel by KIWI, Docker support has been refactored, now Docker classes are a specialization of OCI classes and OCI classes borrowed most of its code from the previous Docker implementation.

Derived OCI images require to provide a URI for the image file which should include also the tag that is going to be imported (unlike Docker image tarballs, OCI images can hold multiple tags). The suggested implementation relies on URI metadata to define the tag, concretely URI fragment is used to specify the tag. See the example:

```xml
<type image="oci" derived_from="file:///some_oci_image.tar.xz#some_tag">
```


davidcassany:
[Support for OCI images](https://trello.com/c/hWOu8EzD/201-support-for-oci-images)

schaefi:
I'm ok with the change, I just did not understand what a "fragment" part of the uri is, it feels more like we need a generic way to extract uri parameters. But this was only on a short look. I need to read the changed to uri more closely once the conflict was resolved

davidcassany:
The fragment part of the URI makes reference to the tag to import. Unlike docker images (tarball including a single image), OCI images are a tarball of a folder that can hold several tags, by default, _umoci_, if no tag is provided (`path[:tag]`) it tries to use the `latest` tag. So when building a base OCI image with some custom `tag` the resulting image will not have a `latest` tag defined, because of that, if you want to make a derived image from that base image, the call `umoci unpack --image some_temp_path` will fail, because there is no `latest` tag. That's why I have included that fragment part in order to state the tag you are aiming to import. Something like:

`derived_from="file:///some/path/image.tar.xz#custom_tag"`

I did not want to include a new type attribute just for that, specially because it is something that might change. Maybe @cyphar has some better advice about how that could be handled, but I don't see many options.

cyphar:
@davidcassany After some recent work I've been doing to implement distribution, I've been reintroducing myself to [the actual RFC for URIs](https://tools.ietf.org/html/rfc3986) and I've realised that `umoci` probably shouldn't be using Docker's brain-dead `:` style for tags. But I'll definitely conflate this with other "breaking changes" in future versions.

The whole `latest` tag thing is to make "the Docker workflow" seem more natural but I'm starting to wonder whether it's a good idea to be honest (it makes `--layout` and `--image` seem too similar to the untrained eye). Overall I'm +1 to using URI fragments to represent tags, though you should note that there is still discussion in the OCI image spec as to what the charset for tags should be (previously it was "whatever the filesystem let you do", but now we have to add some restrictions since they are now defined in JSON as annotations). So it's possible that fragments will restrict tag naming (but we'll see).

schaefi:
ok thanks for the clarification, makes all sense to me

davidcassany:
Looks we are on the same page then, I choosed to use URI fragment because I did not wanted to code a custom error prone `:` parser for container URIs, specially if we plan to support OBS URI types. Also I did not wanted to split in two different portions the base image reference.

@schaefi I will handle the conflict due to the inclusing of the `--delete` flag in DataSync.

schaefi:
Thanks

davidcassany:
Conflict solved

schaefi:
ok, let's go for it :) Thanks

PR #310, opened 2017-04-20
Include --delete flag in DataSync for docker images

### LINKED ISSUE #309: Delete packages from the base image in derived images
## Problem description

Cannot successfully delete packages from the base image.

## Actual behaviour

KIWI is not capable to delete files from the base image when building derived images.

If including `<packages type="delete">` section in the derived description file packages are deleted (removed from rpm database), but the resulting image still contains deleted packages files.

davidcassany:
This commit includes the --delete flag in order to synchronize the
docker images. This is relevant for derived images where the new
layer might not only add files, but also remove something from the
base image.

Fixes #309


schaefi:
:+1:

PR #312, opened 2017-04-25
Derived container source locators

schaefi:
Regarding a discussion with the buildservice team we agreed on an uri schema to allow the use of obs:/ and obsrepositories:/ on the derived_from attribute in a docker type setup. This pull request implements the parts required for KIWI

There will be follow up pull requests which will allow to set the derived_from uri and the container tag also from the commandline

### DIFF
--- a/kiwi/system/uri.py
+++ b/kiwi/system/uri.py
@@ -73,7 +73,8 @@ class Uri(object):
             'iso': True,
             'dir': True,
             'file': True,
-            'suse': True
+            'suse': True,
+            'obsrepositories': True
         }
 
     def translate(self):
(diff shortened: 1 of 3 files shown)


kiwi/system/uri.py
```diff
@@ -210,15 +219,29 @@ def _obs_distribution(self, name):
...
+            if urischeme == 'obsrepositories':
+                local_path = os.sep.join(
+                    [bs_source_dir, 'containers/_obsrepositories', name]
+                )
+            else:
+                local_path = os.sep.join(
+                    [bs_source_dir, 'containers', name]
+                )
+            if fragment:
+                local_path = ''.join([local_path, '#', fragment])
```
davidcassany: All looks great to me :+1 I am just wondering if OBS backend is going to be the responsible of putting in place the appropriate file with the exact name (e.g. `/usr/src/packages/SOURCES/containers/opensuse`) or it is KIWI going to use some kind of regexp match on any file inside the specified folder (e.g. first match of `opensuse*.docker.tar.xz` could be something like `opensuse.x86_64-1.0.0.docker.tar.xz`).

schaefi: The buildservice people told me they will provide it with the name as set in the uri. How they do the mapping to the full name is in their responsibility. So we we should be fine

davidcassany: That's great, it might also facilitate to extend such integration for remote OBS containers repos, like if we were using OBS URIs to build derived images using kiwi outside of OBS. If on their side it is also possible to download the image using this naming convention.

PR #313, opened 2017-04-26
Additional container commandline options

schaefi:
Added --set-container-derived-from and --set-container-tag
commandline options which allows to overwrite the data set
in the XML configuration

This completes the buildservice requirements for building docker images in OBS

### DIFF
--- a/kiwi/tasks/system_build.py
+++ b/kiwi/tasks/system_build.py
@@ -24,6 +24,8 @@ usage: kiwi system build -h | --help
            [--obs-repo-internal]
            [--add-package=<name>...]
            [--delete-package=<name>...]
+           [--set-container-derived-from=<uri>]
+           [--set-container-tag=<name>]
        kiwi system build help
 
 commands:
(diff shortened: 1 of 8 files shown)


kiwi/tasks/system_build.py
```diff
@@ -49,6 +51,15 @@
         when using obs:// repos resolve them using the SUSE internal
         buildservice. This only works if access to SUSE's internal
         buildservice is granted
+    --set-container-derived-from=<uri>
+        overwrite the source location of the base container
+        for the selected image type. The setting is only effective
+        if the configured image type is setup with an initial
+        derived_from reference
+    --set-container-tag=<name>
+        overwrite the container tag in the container configuration.
```
davidcassany: This is because it is expected to be a configurable value form the OBS project configuration? I have difficulties to see the use case.

schaefi: yes see my mail on this topic which should have landed in your inbox now

with every build in the buildservice a new release name which is set as a tag needs to be setup. Thus the static tag value in the containerconfig is not enough when building in the buildservice

As we can set any repo uri from the commandline this should also work for the derived_from uri. The buildservice might also make use of it in case obsrepositories:/... style is used

davidcassany: Yes thanks for the clarification, I see.

PR #314, opened 2017-04-27
Fix comment about Zypper locks to reflect reality

Conan-Kudo:
* Adjust comment to reflect what `zypper al` actually does.
* No code changes.


schaefi:
ok thanks, I can confirm the transaction is broken if one rips out a hard required package from the operation

PR #315, opened 2017-04-27
Add support for excluding packages when using Yum or DNF

### LINKED ISSUE #282: Package exclusion for yum and dnf missing
Excluding a package works as follows

As an additional note, both Yum and DNF implement excludes with --exclude <pkgname> with <pkgname> being a name or a glob. You can stack multiple --exclude options and they'll be processed in an additive manner. That is, --exclude pkgA --exclude pkgB* will exclude both pkgA and all packages matching pkgB* glob. This is a global option, and precedes the action subcommand (e.g. install, etc.).

* Yum:

```
-x, --exclude=package
Exclude a specific package by name or glob from updates on all repositories. Configuration Option: exclude

Specifying Package Names

A package can be referred to for install, update, remove, list, info etc with any of the following as well as globs of any of the following:
name

name.arch
name-ver
name-ver-rel
name-ver-rel.arch
name-epoch:ver-rel.arch
epoch:name-ver-rel.arch
```

* DNF:

```
-x <package-spec>, --exclude=<package-spec>
Exclude packages specified by <package-spec> from the operation.


Specifying Packages
Many commands take a <package-spec> parameter that selects a package for the operation. DNF looks for interpretations of the parameter from the most commonly used meanings to the least, that is it tries to see if the given spec fits one of the following patterns (in decreasing order of priority):

name.arch
name
name-[epoch:]version-release.arch
name-[epoch:]version-release
name-[epoch:]version
```

Conan-Kudo:
This pull request implements support for excluding packages from installation onto the image being built when the selected package manager is Yum or DNF.

Fixes #282.

* Add support for excluding packages when using Yum
* Add support for excluding packages when using DNF

### DIFF
--- a/kiwi/package_manager/dnf.py
+++ b/kiwi/package_manager/dnf.py
@@ -18,7 +18,6 @@
 import re
 
 # project
-from kiwi.logger import log
 from kiwi.command import Command
 from kiwi.package_manager.base import PackageManagerBase
 from kiwi.exceptions import (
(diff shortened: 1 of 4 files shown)


schaefi:
yep this looks great to me too, thanks

PR #317, opened 2017-04-28
Evaluate UNIONFS_CONFIG filesystem

### LINKED ISSUE #316: UnionFST variable is not set in linuxrc in some cases
## Problem description

PXE imges fail to boot using squashfs filesystem and using the following UnionFS configs:

`UNIONFS_CONFIG=/dev/sda3,/dev/sda2,overlay`
`UNIONFS_CONFIG=tmpfs,/dev/sda2,overlay`

## Actual behaviour

Step 18 of linuxrc (check filesystem) fails, as the $UnionFST variable has not been set anywere, which should include the `overlay` filesystem stated in UNIONFS_CONFIG variable.

Build the a pxe image: 

`$ sudo kiwi-ng-3 --type pxe --profile netboot system build --description kiwi-descriptions/suse/x86_64/suse-SLE12-community-JeOS --target-dir /tmp/pxe-image`

Boot it with `/srv/tftpboot/KIWI/config.default` being:

```bash
DISK=/dev/sda
PART='5;S;x,400;L;/,x;L;x'
IMAGE='/dev/sda2;LimeJeOS-SLE12-Community.x86_64;1.13.1;192.168.122.164;4096'
UNIONFS_CONFIG=/dev/sda3,/dev/sda2,overlay
KIWI_INITRD=/boot/initrd
```

davidcassany:
This commit makes sure the filesystem is parsed from the
UNIONFS_CONFIG configuration variable when the PXE image is
loaded to a block device.

This commit fixes #316

### DIFF
--- a/kiwi/boot/arch/ppc/netboot/linuxrc
+++ b/kiwi/boot/arch/ppc/netboot/linuxrc
@@ -604,8 +604,9 @@ then
         done
         imageRootDevice=$imageDevice
         if [ ! -z "$UNIONFS_CONFIG" ];then
-            imageRWDevice=`echo $UNIONFS_CONFIG | cut -d , -f 1`
-            imageRODevice=`echo $UNIONFS_CONFIG | cut -d , -f 2`
+            imageRWDevice=$(echo $UNIONFS_CONFIG | cut -d , -f 1)
+            imageRODevice=$(echo $UNIONFS_CONFIG | cut -d , -f 2)
+            unionFST=$(echo $UNIONFS_CONFIG | cut -d , -f 3)
         fi
         rm -f /etc/ireal.md5
         rm -f /etc/image.md5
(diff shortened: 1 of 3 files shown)


davidcassany:
@schaefi I have just been able to test it in x86_64, but since that part of the code for s390 and ppc is equivalent I assumed that those architectures are having the same issue. Maybe do you have some hints on how to test it for s390 or ppc?

schaefi:
Testing netboot on s390 requires a "virtual pxe" setup which actually passes a parm file to the system which ends up as proc/cmdline parameters. Tim Hardeck tested that a while back and I can't remember the details, sorry. I only test disk images (vmx) and disk image deployments (oem) for s390 because I got some test disks (dasd, zfcp) and a machine assigned to be able to login and build.

For ppc I let the ppc team test because there I don't have a machine nor further information how they do pxe.

The netboot stuff is most interesting for x86 so it's perfectly fine if it works there in the first place

Issue #323, opened 2017-05-10
Doc Redesign: Improve Task-Based Articles


tomschr:
## Problem description
Marcus did already a great job in restructuring the doc (see the `doc_redesign` branch). 👍 To go even further, we discussed some topics were we see improvements:

* Inconsistent structure of some articles
* Does not distinguish between concept-based and task-based

This issue is a summary of what we've discussed.

## Solution
As the structure is almost done, we came up with how we can improve "Building Images" part.

We distinguish between:

* **High level topics**
   these are topics like "Building a Docker Container Image". Usually, such articles describe how to build an image in format XYZ.
* **Low level topics**
  like "Using a Live Image on a Stick". These are the "building blocks" and are more or less independent from a specific image.

Each of these topics contains elements which should be consistent with a similar topics. Such a topic contains the following structural elements:

* **Title**
  The title starts with a verb ("Building/Creating...") to signal a topic which is task oriented.
* **Abstract**
  Leads in the topic. An itemized list contains the issues which are discussed.
* **The task**
  Contains the procedure with a step-by-step instruction.
* **Testing ...**
  Lists actions how the user can check the resulting image.
* **Additional Information**
  Optional links to related information that the user should be aware of.

**Update** The "Task" and "Testing" parts can be either steps or sections, depending on the amount of information.

PR #326, opened 2017-05-12
Revised edition of Live ISO building chapter

tomschr:

* Add abstract (use `.. sidebar`, not sure if we should use it)
* Collect all concept oriented text in the beginning
* Introduce procedure and create steps
* Remove lonely subsection and combine it with the procedure
* Revised example for issue #323 

TODO:

* Review 😄 
* Decide, if `.. sidebar` is ok (or should we better use `.. topic`?)
* Add additional, related information?

I wasn't sure if I should add them into the (still existing) `doc_review` branch, so I separated it into a different branch.

David, this is an example to show you how it looks like according to the above issue number.

Let me know what you think. :wink:

### DIFF
--- a/doc/source/building/build_live_iso.rst
+++ b/doc/source/building/build_live_iso.rst
@@ -9,15 +9,11 @@ Build an ISO Hybrid Live Image
 
    * how to build an ISO image
    * how to run it with QEMU
-   * how to dump the resulting ISO image on an USB stick
 
-In KIWI all generated ISO images are created to be hybrid. This means,
-the image can be used as a CD/DVD or as a disk. This works because
-the ISO image also has a partition table embedded. With more and more
-computers delivered without a CD/DVD drive this becomes important.
-
-The very same ISO image can be copied onto a USB stick and used as a
-bootable disk.
+A Live ISO image is a system on a removable media, e.g CD/DVD or USB stick.
+Once built and deployed it boots off from this media without interferring
+with other system storage components. A useful pocket system for testing
+and demo and debugging purposes.
 
 The following example shows how to build a live ISO image based on
 openSUSE Leap:
@@ -42,4 +38,4 @@ openSUSE Leap:
       $ qemu -cdrom LimeJeOS-Leap-42.1.x86_64-1.42.1.iso -m 4096
 
 After the test was successful, the image is complete and ready to use.
-See :ref:`dump_usb_stick` for further information.
+See :ref:`iso_to_usb_stick` for further information.
--- a/doc/source/building/build_live_iso.rst
+++ b/doc/source/building/build_live_iso.rst
@@ -1,44 +1,45 @@
 Build an ISO Hybrid Live Image
 ==============================
 
-.. hint::
+.. sidebar:: Abstract
 
-   Make sure you have checked out the example image descriptions
-   For details see :ref:`example-descriptions`
+   This page explains how to build a live image. It contains:
 
-The following example shows how to build a live system image
-based on SUSE Leap
+   * how to build an ISO image
+   * how to start it with QEMU
+   * how to dump the resulting ISO image on an USB stick
 
-.. code:: bash
-
-    $ sudo kiwi-ng --type iso system build \
-        --description kiwi-descriptions/suse/x86_64/suse-leap-42.1-JeOS \
-        --target-dir /tmp/myimage
-
-Find the image with the suffix :file:`.iso` below :file:`/tmp/myimage`.
-For testing the live system a virtual machine with the iso image attached
-as CD/DVD device can be used. The following example shows how to use the
-Qemu system to run the image:
-
-.. code:: bash
-
-    $ qemu -cdrom LimeJeOS-Leap-42.1.x86_64-1.42.1.iso -m 4096
-
-.. _hybrid_iso:
-
-Use Live System from USB Stick
-------------------------------
-
-In KIWI all generated iso images are created to be hybrid. This means
+In KIWI all generated ISO images are created to be hybrid. This means,
 the image can be used as a CD/DVD or as a disk. This works because
-the iso image also has a partition table embedded. With more and more
+the ISO image also has a partition table embedded. With more and more
 computers delivered without a CD/DVD drive this becomes important.
 
-As a used the very same iso image can be copied onto a USB stick
-and used as a bootable disk. Thus in order to use the above live
-system from stick do the following:
+The very same ISO image can be copied onto a USB stick and used as a
+bootable disk.
 
-1. Plug in a USB stick
+The following example shows how to build a live ISO image based on
+openSUSE Leap:
+
+1. Make sure you have checked out the example image descriptions,
+   see :ref:`example-descriptions`.
+
+2. Build the image with KIWI:
+
+   .. code:: bash
+
+      $ sudo kiwi-ng --type iso system build \
+           --description kiwi-descriptions/suse/x86_64/suse-leap-42.1-JeOS \
+           --target-dir /tmp/myimage
+
+   Find the image with the suffix :file:`.iso` below :file:`/tmp/myimage`.
+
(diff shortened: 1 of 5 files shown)


davidcassany:
Looks all good to me 😄 I am not convinced about the sidebar also, but, at least with the current styling, is the best option in comparison to `hint` or `topic`. I'll think up how to apply that in the docker.

I like the idea of making a new PR (in new branch) for each building page.

tomschr:
> I am not convinced about the sidebar also, but, at least with the current styling, is the best option in comparison to hint or topic

If I'm not mistaken, you could use a `.. container:: abstract` line and add some CSS magic to it. Not sure if we want to go that far, but that would be one solution.

doc/source/building/build_live_iso.rst
```diff
@@ -1,44 +1,45 @@
...
+.. sidebar:: Abstract
 
-   Make sure you have checked out the example image descriptions
-   For details see :ref:`example-descriptions`
+   This page explains how to build a live image. It contains:
 
-The following example shows how to build a live system image
-based on SUSE Leap
+   * how to build an ISO image
+   * how to start it with QEMU
```
schaefi: I found this misleading, start with qemu sounds like we give an introduction to qemu, which we don't , we just provide the command line how to start the image using qemu

doc/source/building/build_live_iso.rst
```diff
@@ -1,44 +1,45 @@
...
+
+   .. code:: bash
 
-As a used the very same iso image can be copied onto a USB stick
-and used as a bootable disk. Thus in order to use the above live
-system from stick do the following:
+      $ qemu -cdrom LimeJeOS-Leap-42.1.x86_64-1.42.1.iso -m 4096
 
-1. Plug in a USB stick
+4. Plug in a USB stick
```
schaefi: I don't think we should put the stick deployment as part of this sequence. Actually if you have built the iso and tested it in qemu you are done. Once done you have several other choices what to do with the image

1. burn it on a CD
2. dump it on a USB stick

both actions are performed to be able to store the image on a removable media. These days I would say nobody burns a CD anymore, almost all people will put it on a stick. Thus the stick deployment is worth some words but imho outside of the build and test list. Thoughs ?

doc/source/building/build_live_iso.rst
```diff
@@ -60,9 +61,9 @@ system from stick do the following:
 
       $ dd if=LimeJeOS-Leap-42.1.x86_64-1.42.1.iso of=/dev/<stickdevice>
 
-3. Run system from stick
+6. Start system from USB stick
 
-   Like in the example above use Qemu to run the system, but attach
+   Like in the example above use QEMU to start the system, but attach
```
schaefi: I think I would change this and focus more on the real use case once the image has been dumped on the stick.
The qemu test is basically the same as done before, what the user really will do is plug in the stick into a computer and boot from the stick which requires a change in the boot sequence of the machine to boot from the USB stick. The last information is very specific to the used computer and I think all we can say here is that the setup to boot from the USB stick must be done in the BIOS or Firmware setup tool of the used computer system.

doc/source/building/build_live_iso.rst
```diff
@@ -1,44 +1,45 @@
 Build an ISO Hybrid Live Image
 ==============================
 
-.. hint::
+.. sidebar:: Abstract
```
schaefi: 👍

schaefi:
forgot to mention, when building the docs I got

```
WARNING: undefined label: hybrid_iso (if the link has no caption the label must precede a section header)
```

So the changes deleted the link `-.. _hybrid_iso:`

doc/source/building/build_live_iso.rst
tomschr: Me too. 😉 What about the `container`?

doc/source/building/build_live_iso.rst
tomschr: Ok, what about replacing "start" -> "run"?

doc/source/building/build_live_iso.rst
tomschr: I agree, CDs aren't really important nowadays. We could separate the "create and test part" from the "dump part". However, that would introduce a single, lonely section which should be avoided IMHO.

Is the dump part something other topics could use? If yes, we could remove it alltogether and create a different "topic-like" structure and link to it (as we called it "low-level-topic").

doc/source/building/build_live_iso.rst
tomschr: Ok, if I got you right, you think this topic should be extended with information about booting from a computer system. Shouldn't be a problem to add that.

However, this can only be a bit "vague" as the menu items and the procedure is different on every computer (or BIOS).

tomschr:
> WARNING: undefined label: hybrid_iso 

Thanks for spotting this error. Fixed in commit 2d294e8

doc/source/building/build_live_iso.rst
schaefi: yes would be fine for me

doc/source/building/build_live_iso.rst
schaefi: yes I think it should be a low level topic, our first one :)

doc/source/building/build_live_iso.rst
schaefi: right, there is unfortunately no standard procedure, the only information which is true for everybody is that the boot from usb stick must be explicitly activated from the BIOS aka Firmware
which is before the operating System starts. May Firmware implementation offers a boot menu which can be activated at boot time via the F12 key, but as I said it's different in many ways

doc/source/building/build_live_iso.rst
schaefi: I rebuild the docs with '.. container:: Abstract' and to be honest, sidebar looks much better :)
The sidebar also better separates the abstract information from the text. So let's go with the sidebar

doc/source/building/build_live_iso.rst
tomschr: I've provided another commit (see 03cafdc) were you can see high- and low-level in place.

I've introduced a "Working with Images" section where we can put all of these "low-level topics". Currently, this section is placed under "Building Images"; I'm not sure if we should do that or move it to an upper level.

Let me know what you think about. 😉

doc/source/building/build_live_iso.rst
tomschr: Done, see commit 03cafdc

tomschr:
I've tried to address your concerns and separated the different levels (see commit 03cafdc).

Currently, with the above commit you get this:

```
Building Images
  +--- Building A
  +--- Building B
  +--- other "Building... " topics
  +--- Working with Images # <--- This is the new divison
         +--- Dumping an Image to an USB Stick  # <--- a low-level topic
```

I didn't find a better title so I used "Working with Images" for the time being (we can change that, of course, I wasn't very creative. Mea culpa).

I'm not sure if I like the current position, but I think, we can leave it as is now. Maybe, over time, we find out that this "Working with Images" part is a lot bigger than the "Building Images" part. In that case, we could move it to an upper level, as a sibling to "Building Images".

Thoughts?

schaefi:
Thanks Tom for starting the "low level" topics. I added another commit which adds information which base topic was referenced. I think this is important otherwise if someone only reads the working with images chapters it will be confusing from where e.g the referenced iso image came from.

Let me know what you think.

Other than that I'm fine with the current state which I think could be applied to other building chapters too :)

tomschr:
> I added another commit which adds information which base topic was referenced. I think this is important otherwise if someone only reads the working with images chapters it will be confusing from where e.g the referenced iso image came from.

Great idea! 👍 

> Other than that I'm fine with the current state which I think could be applied to other building chapters too :)

Perfect! 👍  Ok, if you satisfied with all, you can (squash) merge it to master. 😉

schaefi:
all right :)

PR #327, opened 2017-05-12
Added dracut output file format detection

### LINKED ISSUE #325: The name of the dracut initrd output file is different on each distro
## Problem description

The default output file of dracut is different on any distribution and several tools of the distribution
rely on that output name format. In order to stay compatible kiwi should be able to produce the
same output name schema

* Fedora uses initramfs-*.img,
* Mageia uses initrd-*.img,
* SUSE uses initrd-*,
* Debian uses initrd.img-*

## Actual behaviour

kiwi always produce the SUSE style names which inherits to the bootloader configuration
and causes trouble when the distro toolchain is used to e.g rebuild the initrd on a kernel
update

schaefi:
Unfortunately the dracut initrd output file format varies between
the different Linux distributions. Tools like lsinitrd, and also
grub2 rely on the initrd output file to be in that format.
Thus when kiwi uses dracut the same file format should be used
all over the place in order to stay compatible with what the
distribution does. Fixes #325

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -1009,7 +1009,7 @@ class DiskBuilder(object):
         all over the place in order to stay compatible with what the
         distribution does
         """
-        default_outfile_format = 'initrd-{kernel_version}'
+        default_outfile_format = 'initramfs-{kernel_version}.img'
         dracut_search_env = {
             'PATH': os.sep.join([self.root_dir, 'usr', 'bin'])
         }
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -16,6 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
+import re
 import platform
 import pickle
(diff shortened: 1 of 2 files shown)


kiwi/builder/disk.py
```diff
@@ -986,12 +987,44 @@ def _get_boot_names(self):
...
+    def _get_dracut_output_file_format(self):
+        """
+        Unfortunately the dracut initrd output file format varies between
+        the different Linux distributions. Tools like lsinitrd, and also
+        grub2 rely on the initrd output file to be in that format.
+        Thus when kiwi uses dracut the same file format should be used
+        all over the place in order to stay compatible with what the
+        distribution does
+        """
+        default_outfile_format = 'initrd-{kernel_version}'
```
Conan-Kudo: I would strongly suggest that you fallback to the default for `dracut`: `initramfs-{kernel_version}.img`, as that is more likely to work when you can't figure out what the format is.

schaefi: good point, I'll consider that

PR #331, opened 2017-05-15
Handle rpm-check-signatures flag for each package manager

davidcassany:
This commit adds support for the rpm-check-signatures flag, which
sets the package manager to verify or not each package signature.
By default KIWI assumes no gpg checks are done.


schaefi:
Looks great 👍

PR #332, opened 2017-05-16
Use 'yum-deprecated' if it exists when using Yum

### LINKED ISSUE #329: The "yum" binary can be named "yum-deprecated", which breaks KIWI for building with Yum
## Problem description
In Fedora and Mageia, the `yum` binary for Yum has been renamed to `yum-deprecated`, since `yum` is now a pointer to `dnf`.

KIWI needs to test for the existence of `/usr/bin/yum-deprecated` and use it instead of `/usr/bin/yum` when Yum is the selected package manager for an image build.

## Actual behaviour
Building an image with Yum will use DNF and send the wrong options in some situations (like for turning weakdeps off), which breaks the whole thing.

1. Install `dnf-yum`. (You'll need to remove yum first on openSUSE). 
2. Try to build an image with `yum` as the package manager.

Conan-Kudo:
On Fedora and Mageia systems, the Yum package manager binary is named `yum-deprecated`, and `yum` redirects to DNF. This is a problem for being able to build images of Linux systems that use Yum as its package manager, like RHEL/CentOS 7.

So, in order to ensure we use the right package manager, we check for the `yum-deprecated` binary and use it if it exists.

Fixes #329.

* Add `_get_yum_binary_name()` to `PackageManagerYum` class.
* Use `_get_yum_binary_name()` to get yum binary name for yum invocations.

### DIFF
--- a/kiwi/package_manager/yum.py
+++ b/kiwi/package_manager/yum.py
@@ -15,11 +15,13 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+import os
 import re
 
 # project
 from kiwi.command import Command
 from kiwi.package_manager.base import PackageManagerBase
+from kiwi.path import Path
 from kiwi.exceptions import (
     KiwiRequestError
 )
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
Note that this doesn't have any unit test changes for this, because I'm not sure how to adjust the unit tests to ensure the tests work reliably. As it is, I suspect the unit tests would fail on a Fedora or Mageia host because the strings wouldn't match for the asserts.

kiwi/package_manager/yum.py
```diff
@@ -88,21 +89,33 @@ def request_package_exclusion(self, name):
...
         self.exclude_requests.append(name)
 
+    def _get_yum_binary_name(self):
+        """
+        Identify whether yum is 'yum' or 'yum-deprecated'
+
+        :return: name of yum command
+        """
+        yum_binary = 'yum'
+        if os.path.exists('/usr/bin/yum-deprecated'):
```
schaefi: Could we change this to:

```python
from kiwi.path import Path

if Path.which(filename='yum-deprecated', access_mode=os.X_OK):
    ...
```

Conan-Kudo: Sure.

schaefi:
Thanks, the code looks good so far, see my comment on the binary file check. If you consider using Path.which you can design the unit test to be decorated with

```python
@patch('kiwi.path.Path.which')
```

with that mock you can test both conditions

Conan-Kudo:
@schaefi How does that work?

Conan-Kudo:
@schaefi @davidcassany:

I've written unit tests for the new function and made all the tests work on my local computer (running Fedora)! This is now good to merge!

schaefi:
👍 great work

PR #333, opened 2017-05-18
Fix oem partition install

davidcassany:
This PR includes a couple of commits related to bnc#1039469:

* One fixes the `dump` script of the oem boot code in order to use the
appropriate arguments in `loop_setup` method.
* The other one modifies the partitionID output when using GPT partitions,
with the change the partition ID codes are two hexa digits as in MBR.


schaefi:
looks good to me 👍

PR #336, opened 2017-05-19
Fixup boot-load-size for efi loader in iso

schaefi:
Pass the real boot-load-size of the used loader as number
of 512byte blocks to the iso creation call. Related to
(bsc#939456)


wfeldt:
looks ok

PR #338, opened 2017-05-19
Improve rpm-check-signatures support

davidcassany:
This commit ensures the signatures are checked for both: the
repository and the rpm package. It applies for zypper, dnf and
yum package managers.


schaefi:
Great 👍

PR #339, opened 2017-05-21
Refactoring quote_xml_aux, quote_attrib, quote_python

tomschr:
This fixes no special issues, but I just came across. It's a minor change in this pull request:

* Add test case for `quote_xml_aux`, `quote_attrib`, `quote_python`
* Basically, the original functions are refactored as this:
  * `quote_xml_aux()` is equally to `xml.sax.saxutils.escape`
  * `quote_attrib()` is almost the same as `xml.sax.saxutils.quoteattr`

**Question:**

* `quote_python()` is tricky; it is almost(!) equal to `repr()`. As `quote_python()` is never used, it is probably safer to remove it...?

### DIFF
--- a/kiwi/xml_parse.py
+++ b/kiwi/xml_parse.py
@@ -28,6 +28,7 @@ try:
     from lxml import etree as etree_
 except ImportError:
     from xml.etree import ElementTree as etree_
+from xml.sax.saxutils import escape, quoteattr
 
 
 Validate_simpletypes_ = True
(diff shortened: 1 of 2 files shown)


schaefi:
I'm sorry but kiwi/xml_parse.py is auto generated code as you can see in the header of the file

```
Generated Wed Apr 12 14:43:43 2017 by generateDS.py version 2.25a.
```

I'm afraid if you want to make your changes permanent this needs to be added in the upstream generateDS project. The test case should also go upstream as I don't think it makes much sense to maintain unit tests for code written by a machine

tomschr:
Argh, of course, you are right. Never mind, then we close this PR. Thanks for watching it! 👍

PR #340, opened 2017-05-21
Idea: Test Function Docstrings

tomschr:
This fixes no particular issue. It's up to you if you think it is useful or not. 😉 I don't mind if you close it without merging.

Yes, the testcases **fail** for the time being, but this has to be: not all functions contains a description of the function parameters. I would have documented them, but I wasn't sure on some, so it's "work in progress".

* Add two new testcases for docstrings (file `docstrings_test.py`):
  * `test_docstrings_nonempty()`
    just tests for an non-empty docstring for each functions in the KIWI code.
  * `test_docstrings_args()`
    tests , if a docstring contains descriptions of all parameters of a function
 
As you can see, the second one is more strict than the first. This was done on purpose as you can choose one or the other. 

Both tests skip private functions starting with "_" in the function name. Also classes. If you want them to be checked as well, additional code needs to be added (I wanted to start with something simple, I leave it as an exercise. 😉 )

An example output:
```
test/unit/docstrings_test.py::test_docstrings_nonempty[kiwi.xml_parse.quote_attrib] FAILED
[...]
==== 17 failed, 25 passed in 0.42 seconds ====
```

Some benefits:

* Improve code hygiene
* Improves API documentation
* New functions are not only checked for their functionality, but for their documentation (=docstring) too
* Adheres to a "definition of done" check (if you consider docstrings as valueable 😉 )
* Test can be as strict or as loose as you want it

Let me know what you think.

### DIFF
--- a/kiwi/xml_parse.py
+++ b/kiwi/xml_parse.py
@@ -488,12 +488,6 @@ def quote_python(inStr):
 
 
 def get_all_text_(node):
-    """Get all texts from a node and their child nodes
-
-    :param node: element node
-    :return: the combined string of the node and all their child nodes
-    :rtype: str
-    """
     if node.text is not None:
         text = node.text
     else:
@@ -505,12 +499,6 @@ def get_all_text_(node):
 
 
 def find_attr_value_(attr_name, node):
-    """Find attribute value from node
-
-    :param attr_name: name of the attribute
-    :param node: element node
-    :return: attribute value
-    """
     attrs = node.attrib
     attr_parts = attr_name.split(':')
     value = None
--- a/kiwi/xml_parse.py
+++ b/kiwi/xml_parse.py
@@ -38,6 +38,7 @@ else:
 
 
 def parsexml_(infile, parser=None, **kwargs):
+    """ TBD """
     if parser is None:
         # Use the lxml ElementTree compatible parser so that, e.g.,
         #   we ignore comments.
@@ -418,6 +419,13 @@ CurrentSubclassModule_ = None
 
 
 def showIndent(outfile, level, pretty_print=True):
+    """Indent code
+
+    :param outfile: file like object
+    :param level: integer level of spaces (multiplied with 4)
+    :param pretty_print: boolean to real pretty print (True)
+                         or not (False)
+    """
     if pretty_print:
         for idx in range(level):
             outfile.write('    ')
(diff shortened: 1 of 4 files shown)


kiwi/xml_parse.py
```diff
@@ -38,6 +38,7 @@
 
 
 def parsexml_(infile, parser=None, **kwargs):
+    """ TBD """
     if parser is None:
```
schaefi: This is auto generated code

tomschr: Oh yes! Thanks, will remove it and add a ignore list.

tomschr:
Commit baa392b contains now code which ignores all functions from `kiwi.xml_parse`.

However, after looking at the Travis build and my local run, it seems this is not enough. My code influences other tests too which makes them fail. 😢  Unfortunately, I don't know why. My guess is, that the loading of the kiwi modules circumvent the mocking somehow.

@schaefi Not sure how to fix that. Any idea?

test/unit/docstrings_test.py
```diff
@@ -0,0 +1,99 @@
...
+    :param module: the given module to search for members
+    :type module: :class:`module`
+    :param predicate: a function with expects an object and returns True or False
+                      see :func:`inspect.isfunction` as an example
+    :type prediate: :class:`function`
+    """
+    prefix = module.__name__ + "."
+    for importer, modname, ispkg in pkgutil.iter_modules(module.__path__, prefix):
+        module = importer.find_module(modname).load_module(modname)
+        yield from members(module, predicate)
```
schaefi: This is most probably conflicting with the loading pytest does

In addition there are doc strings evaluated by docopt. that covers all code below tasks/

Sorry I have no idea at the moment how to solve this

tomschr: Yes, I was suspicious about that part too. Maybe we can tell pytest to execute this test at last?

schaefi: I don't think the order of tests can be influenced, and to be honest the order should not serve as a "fix". doing so will strike back sooner or later and at that time it will be very hard to remember that the order of the tests was significant for success or failure. I'm sorry

tomschr:
I agree, I don't really consider changing the order as a good solution either. Was more of a desparate idea. 😉

As I need to "load" the KIWI submodules somehow to introspect them, I don't have a better idea to offer at the moment. Of course I could "manually" parse the files somehow, but I consider it also as a hack.

So maybe better we close it for the time being and if I get a better idea, I can reopen it. Sorry for the noise.

PR #344, opened 2017-05-22
Added --clear-cache option

### LINKED ISSUE #341: Provide option --clear-cache
## Problem description

As found for the zypper package manager, all data loaded into the cache is trusted, which means if the user changes the trust level on the image description level e.g via `rpm-check-signatures` it might still happen that a package with an unverified signature is taken from the cache

While we can't change the zypper behavior we could at least provide an option to kiwi which allows to
empty the cache to make sure all data is reloaded with the configured trust level

As this might apply to other package managers as well the option should apply to the cache of the selected package manager

schaefi:
The system prepare and build commands now provides the option --clear-cache which deletes all cache data associated with the repositories to build the image. This Fixes #341

### DIFF
--- a/kiwi/repository/zypper.py
+++ b/kiwi/repository/zypper.py
@@ -267,14 +267,20 @@ class RepositoryZypper(RepositoryBase):
         """
         Delete zypper repository cache
 
-        The cache data for each repository is stored in a directory
-        of the same name as the repository name. The method deletes
-        this directory to cleanup the cache information
+        The cache data for each repository is stored in a list of
+        directories of the same name as the repository name. The method
+        deletes these directories to cleanup the cache information
 
         :param string name: repository name
         """
         Path.wipe(
             os.sep.join([self.shared_zypper_dir['pkg-cache-dir'], name])
+        )
+        Path.wipe(
+            os.sep.join([self.shared_zypper_dir['solv-cache-dir'], name])
+        )
+        Path.wipe(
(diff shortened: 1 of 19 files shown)


kiwi/repository/zypper.py
```diff
@@ -263,6 +263,20 @@ def delete_all_repos(self):
...
+        The cache data for each repository is stored in a directory
+        of the same name as the repository name. The method deletes
+        this directory to cleanup the cache information
+
+        :param string name: repository name
+        """
+        Path.wipe(
+            os.sep.join([self.shared_zypper_dir['pkg-cache-dir'], name])
+        )
+
```
davidcassany: @schaefi this is not enough for zypper, since this only cleans cached packages, but not cached repository metadata (/var/cache/kiwi/zypper/raw/<repo> and there are other directories inside `zypper` folder which could be deleted, even they do not contain any signature). With this code the repository is still trusted if it was already on the cache. The build will fail, but due to package singature check, not due to repository signature check. 

As a side note not related with this PR, I just noticed that if the repository signature fails KIWI does not fail even the repository is automatically rejected, it just continues without the untrusted repository, which is likely to fail due to some missing packages, but could not be the case (imagine update repo with a different key).

schaefi: thanks will update the code

regarding the repo I think there is not much we can do. The way zypper implemented the non-interactive flag causes the behavior you see. in interactive mode you get a question and can cancel the entire process but with non-interactive as we need it for image building the repo is just skipped for this build.

PR #346, opened 2017-05-24
Add --signing-key option

### LINKED ISSUE #342: Allow import of rpm key for signature verification
## Problem description

To allow signature verification the rpm database of the image needs to have the signing keys imported.
The idea is to provide an option to allow the import of a key or a collection of keys

## Actual behaviour

The keyring of the rpm database created in the image root is always empty

Build an image with rpm-check-signatures turned on

davidcassany:
* Adds `--signing-key` option to import repository and package trusted keys at build time
* Disables signature check of the repository for Yum and Dnf

Fixes #342

### DIFF
--- a/kiwi/boot/image/__init__.py
+++ b/kiwi/boot/image/__init__.py
@@ -37,7 +37,7 @@ class BootImage(object):
                 xml_state, target_dir, signing_keys=signing_keys
             )
         elif initrd_system == 'dracut':
-            return BootImageDracut(xml_state, target_dir)
+            return BootImageDracut(xml_state, target_dir, root_dir)
         else:
             raise KiwiBootImageSetupError(
                 'Support for %s initrd system not implemented' % initrd_system
--- a/kiwi/boot/image/__init__.py
+++ b/kiwi/boot/image/__init__.py
@@ -28,14 +28,16 @@ class BootImage(object):
     """
         BootImge factory
     """
-    def __new__(self, xml_state, target_dir, root_dir=None):
+    def __new__(self, xml_state, target_dir, root_dir=None, signing_keys=None):
         initrd_system = xml_state.build_type.get_initrd_system()
         if not initrd_system:
             initrd_system = 'kiwi'
         if initrd_system == 'kiwi':
-            return BootImageKiwi(xml_state, target_dir)
+            return BootImageKiwi(
+                xml_state, target_dir, signing_keys=signing_keys
+            )
         elif initrd_system == 'dracut':
-            return BootImageDracut(xml_state, target_dir, root_dir)
+            return BootImageDracut(xml_state, target_dir)
         else:
             raise KiwiBootImageSetupError(
                 'Support for %s initrd system not implemented' % initrd_system
--- a/kiwi/boot/image/builtin_kiwi.py
+++ b/kiwi/boot/image/builtin_kiwi.py
@@ -52,7 +52,7 @@ class BootImageKiwi(BootImageBase):
             root_dir=self.boot_root_directory,
             allow_existing=True
         )
-        manager = system.setup_repositories()
+        manager = system.setup_repositories(signing_keys=self.signing_keys)
         system.install_bootstrap(
             manager
         )
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -208,13 +208,13 @@ class DiskBuilder(object):
         )
         self.custom_args = custom_args
 
-        self.boot_signing_keys = None
+        self.signing_keys = None
         if custom_args and 'signing_keys' in custom_args:
-            self.boot_signing_keys = custom_args['signing_keys']
+            self.signing_keys = custom_args['signing_keys']
 
         self.boot_image = BootImage(
             xml_state, target_dir, root_dir,
-            signing_keys=self.boot_signing_keys
+            signing_keys=self.signing_keys
         )
         self.firmware = FirmWare(
             xml_state
@@ -488,7 +488,7 @@ class DiskBuilder(object):
 
                 self.boot_image = BootImageKiwi(
                     self.xml_state, self.target_dir,
-                    signing_keys=self.boot_signing_keys
+                    signing_keys=self.signing_keys
                 )
 
                 self.boot_image.prepare()
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -171,7 +171,7 @@ class DiskBuilder(object):
     * :attr:`result`
         Instance of Result
     """
-    def __init__(self, xml_state, target_dir, root_dir):
+    def __init__(self, xml_state, target_dir, root_dir, custom_args=None):
         self.arch = platform.machine()
         if self.arch == 'i686' or self.arch == 'i586':
             self.arch = 'ix86'
@@ -206,8 +206,15 @@ class DiskBuilder(object):
         self.disk_setup = DiskSetup(
             xml_state, root_dir
         )
+        self.custom_args = custom_args
+
+        self.boot_signing_keys = None
+        if custom_args and 'signing_keys' in custom_args:
+            self.boot_signing_keys = custom_args['signing_keys']
+
         self.boot_image = BootImage(
-            xml_state, target_dir, root_dir
+            xml_state, target_dir, root_dir,
+            signing_keys=self.boot_signing_keys
         )
         self.firmware = FirmWare(
             xml_state
@@ -256,7 +263,7 @@ class DiskBuilder(object):
         * image="vmx"
         """
         disk = DiskBuilder(
-            self.xml_state, self.target_dir, self.root_dir
+            self.xml_state, self.target_dir, self.root_dir, self.custom_args
         )
         result = disk.create_disk()
 
@@ -480,7 +487,8 @@ class DiskBuilder(object):
                     self.xml_state.build_type.get_initrd_system()
 
                 self.boot_image = BootImageKiwi(
-                    self.xml_state, self.target_dir
+                    self.xml_state, self.target_dir,
+                    signing_keys=self.boot_signing_keys
                 )
 
                 self.boot_image.prepare()
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -100,7 +100,7 @@ class LiveImageBuilder(object):
     * :attr:`result`
         Instance of Result
     """
-    def __init__(self, xml_state, target_dir, root_dir):
+    def __init__(self, xml_state, target_dir, root_dir, custom_args=None):
         self.media_dir = None
         self.arch = platform.machine()
         if self.arch == 'i686' or self.arch == 'i586':
@@ -122,8 +122,12 @@ class LiveImageBuilder(object):
         if not self.live_type:
             self.live_type = Defaults.get_default_live_iso_type()
 
+        boot_signing_keys = None
+        if custom_args and 'signing_keys' in custom_args:
+            boot_signing_keys = custom_args['signing_keys']
+
         self.boot_image_task = BootImage(
-            xml_state, target_dir
+            xml_state, target_dir, signing_keys=boot_signing_keys
         )
         self.firmware = FirmWare(
             xml_state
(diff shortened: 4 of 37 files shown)


kiwi/repository/dnf.py
```diff
@@ -253,9 +263,9 @@ def _create_runtime_config_parser(self):
         self.runtime_dnf_config.set(
             'main', 'gpgcheck', self.gpg_check
         )
-        self.runtime_dnf_config.set(
-            'main', 'repo_gpgcheck', self.gpg_check
-        )
+        # We are not setting repo_gpgcheck, which forces repository
+        # signatures checks, because current Fedora 25 releases do not
+        # provide signed repositories, only signed packages
```
Conan-Kudo: It is possible for distributions to have signed repositories, or even individual repos that do. For example, in Fedora, the `fedora-cisco-openh264` repo is signed.

kiwi/repository/yum.py
```diff
@@ -248,9 +258,9 @@ def _create_runtime_config_parser(self):
         self.runtime_yum_config.set(
             'main', 'gpgcheck', self.gpg_check
         )
-        self.runtime_yum_config.set(
-            'main', 'repo_gpgcheck', self.gpg_check
-        )
+        # We are not setting repo_gpgcheck, which forces repository
+        # signature checks, because I could not manage to make it work.
+        # Nevertheless, package signature validation is functional.
```
Conan-Kudo: What did you try testing with? AFAIK, the CentOS 7 repositories are signed.

kiwi/repository/dnf.py
davidcassany: Yes you are right, I was in doubt about that, as it is not a problem of DNF but Fedora (the package manager could be used even in different distros a part form Fedora). But I could  not find a clear way to check it, so I preferred not to include it. Also if Fedora does not sign its main repos the whole singatures check would have to be disabled if the main repo is included (which is likely to happen). We are not setting the signature checks on repo basis, so in that case, I assume checking package signatures already better than nothing.

Conan-Kudo: It is, but perhaps you should consider adding a flag for repository signatures to the config.xml and just set it to false for the Fedora one. Other distributions may make different choices. For example, I'm hoping to get Mageia's repositories signed.

kiwi/repository/yum.py
davidcassany: Yes CentOS repos are signed, this is true. But I just could not make it run on my openSUSE host (tested with leap 42.2 and tumbleweed clean images). I did a custom `yum_config` file, as simple as:
```
[main]
cachedir = /tmp/yum-cache
reposdir = /tmp/yum-repos
pluginconfpath = /tmp/yum-pluginconf
debuglevel = 10
repo_gpgcheck = 1
metadata_expire = 180
```
Added repo in `/tmp/yum-repos/OSS.repo`:
```
[OSS]
name = OSS
baseurl = http://mirror.centos.org/centos/7/os/x86_64/
```
and then executed:

```
$ rpm --import http://mirror.centos.org/centos/7/os/x86_64/RPM-GPG-KEY-CentOS-7
$ yum -c yum_config makecache
```
Also tried some other similar combinations, always with the same result:
```
Config time: 0.008
Yum Version: 3.4.3
COMMAND: yum -c yum_config makecache 
Installroot: /
OSS/signature                                                                                                                                                                                                               |  811 B     00:00     
OSS/signature                                                                                                                                                                                                               | 3.6 kB     00:00 ... 
http://mirror.centos.org/centos/7/os/x86_64/repodata/repomd.xml: [Errno -1] Gpg Keys not imported, cannot verify repomd.xml for repo OSS
Trying other mirror.
Error: Cannot retrieve repository metadata (repomd.xml) for repository: OSS. Please verify its path and try again
```

Conan-Kudo: Can you try checking on CentOS 7 and Fedora 25 systems? If the behavior is broken/fixed in yum, I'd like to know when it was. That might help in identifying when it got fixed (if it did at all) and possibly backport it to the SUSE yum package.

davidcassany: Just tested with CentOS 7 JeOS with the same result. Using:

`yum-3.4.3-150.el7.centos.noarch`

davidcassany:
@schaefi, just in case, DO NOT MERGE this PR. I just noticed I can't request changes to myself 😛 I noticed a big issue while retesting yum support, signatures are not imported to the boot image root! Thus the boot image process fails (usually I did not wait the build to complete, I was just satisfied viewing the installation and validation of image packages and my full openSUSE tests were with dracut option 😛). I missed completely missed the boot step.

schaefi:
all right, I'll wait :)

schaefi:
@davidcassany just for information, I've added the blocked label on this one, if you are finished just delete the label which indicates we can have a look again. Thanks

davidcassany:
@schaefi thanks for pointing out the block label, I did not think about labels to mark it as a work in progress/blocked. Anyway, I have already removed the blocked label, it should be ready for reviewing now.

schaefi:
great thanks looking at it now

kiwi/boot/image/__init__.py
```diff
@@ -28,14 +28,16 @@ class BootImage(object):
...
     """
-    def __new__(self, xml_state, target_dir, root_dir=None):
+    def __new__(self, xml_state, target_dir, root_dir=None, signing_keys=None):
         initrd_system = xml_state.build_type.get_initrd_system()
         if not initrd_system:
             initrd_system = 'kiwi'
         if initrd_system == 'kiwi':
-            return BootImageKiwi(xml_state, target_dir)
+            return BootImageKiwi(
+                xml_state, target_dir, signing_keys=signing_keys
```
schaefi: just nitpicking but I think we don't need to use the key=value style if the variable name clearly indicates what this is. This I think:

```python
return BootImageKiwi(xml_state, target_dir, signing_keys)
```

is good enough

kiwi/boot/image/__init__.py
```diff
@@ -28,14 +28,16 @@ class BootImage(object):
...
+            )
         elif initrd_system == 'dracut':
-            return BootImageDracut(xml_state, target_dir, root_dir)
+            return BootImageDracut(xml_state, target_dir)
```
schaefi: Hmm, this looks wrong to me. In case of a dracut initrd system we _have_ to use the root_dir to operate and don't let the initrd creation process create a new and empty tmpdir. That's because dracut has to have access to the root system data it should boot later on. I also wonder how that change depend on the signing key implementation ?

kiwi/boot/image/builtin_kiwi.py
```diff
@@ -52,7 +52,7 @@ def prepare(self):
             root_dir=self.boot_root_directory,
             allow_existing=True
         )
-        manager = system.setup_repositories()
+        manager = system.setup_repositories(signing_keys=self.signing_keys)
```
schaefi: same here key=value style seems to provide redundant information

```python
manager = system.setup_repositories(self.signing_keys)
```

kiwi/builder/disk.py
```diff
@@ -206,8 +206,15 @@ def __init__(self, xml_state, target_dir, root_dir):
...
+        self.custom_args = custom_args
+
+        self.boot_signing_keys = None
+        if custom_args and 'signing_keys' in custom_args:
+            self.boot_signing_keys = custom_args['signing_keys']
+
         self.boot_image = BootImage(
-            xml_state, target_dir, root_dir
+            xml_state, target_dir, root_dir,
+            signing_keys=self.boot_signing_keys
```
schaefi: same here key=value style seems to provide redundant information
I think the name boot_signing_keys is misleading, sounds like we provide a key
for signing the boot code

```python
self.boot_image = BootImage(
    xml_state, target_dir, root_dir, self.boot_image_signing_keys
)
```

kiwi/builder/disk.py
```diff
@@ -480,7 +487,8 @@ def create_disk(self):
                     self.xml_state.build_type.get_initrd_system()
 
                 self.boot_image = BootImageKiwi(
-                    self.xml_state, self.target_dir
+                    self.xml_state, self.target_dir,
+                    signing_keys=self.boot_signing_keys
```
schaefi: same here key=value style seems to provide redundant information

kiwi/builder/live.py
```diff
@@ -122,8 +122,12 @@ def __init__(self, xml_state, target_dir, root_dir):
         if not self.live_type:
             self.live_type = Defaults.get_default_live_iso_type()
 
+        boot_signing_keys = None
+        if custom_args and 'signing_keys' in custom_args:
+            boot_signing_keys = custom_args['signing_keys']
+
         self.boot_image_task = BootImage(
-            xml_state, target_dir
+            xml_state, target_dir, signing_keys=boot_signing_keys
```
schaefi: same here key=value style seems to provide redundant information

kiwi/builder/pxe.py
```diff
@@ -84,8 +84,13 @@ def __init__(self, xml_state, target_dir, root_dir):
...
             xml_state=xml_state, root_dir=root_dir
         )
+
+        boot_signing_keys = None
+        if custom_args and 'signing_keys' in custom_args:
+            boot_signing_keys = custom_args['signing_keys']
+
         self.boot_image_task = BootImage(
-            xml_state, target_dir
+            xml_state, target_dir, signing_keys=boot_signing_keys
```
schaefi: same here key=value style seems to provide redundant information

kiwi/repository/yum.py
```diff
@@ -248,9 +258,9 @@ def _create_runtime_config_parser(self):
         self.runtime_yum_config.set(
             'main', 'gpgcheck', self.gpg_check
         )
-        self.runtime_yum_config.set(
-            'main', 'repo_gpgcheck', self.gpg_check
-        )
+        # We are not setting repo_gpgcheck, which forces repository
+        # signature checks, because I could not manage to make it work.
```
schaefi: can we reword this, it doesn't sound promising

schaefi:
What is also missing is an update of the manual pages regarding the signing-key option

kiwi/boot/image/__init__.py
davidcassany: Yes this looks wrong to me also... I have to double check it, but it looks like a very ugly confusion of mine.

kiwi/boot/image/__init__.py
davidcassany: We need the key=value style here, since base init signatures is 
 ```python
def __init__(self, xml_state, target_dir, root_dir=None, signing_keys=None)
```
Then I need the key=value style to state the third parameter is `signing_keys` instead of `root_dir`

kiwi/boot/image/builtin_kiwi.py
davidcassany: The same as before, setup_repositories also includes `clear_cache` with a default values in its signature, we need to clearly state this parameter goes for `signing_keys` instead of `clear_cache`.

kiwi/builder/disk.py
davidcassany: Ok, agree with the variable name change. Note here I have not used the key=value style as the `root_dir` parameter is also provided.

kiwi/repository/yum.py
davidcassany: Yes, I will reword that.

davidcassany:
About he manual update do you refer to the `doc/source/commands/system_*` pages? Yes, I missed the docs `system_create` page, I am updating it.

schaefi:
yes the doc/source/commands part, imho system_prepare and system_build are the two affected man pages. They are in line with the code in tasks/

kiwi/boot/image/__init__.py
schaefi: oh you are right, I've overlooked that we have more params, nevermind let's leave it where needed. sorry

davidcassany:
yes, system_prepare and system_build are already there. System_create is affected also due to the boot image, as this is created during the create step, so performing a two step build the signing-key option should be used in both commands.

PR #350, opened 2017-05-26
Fix existing root check, fixes #349

### LINKED ISSUE #349: Kiwi system build command always fail without --allow-existing-root option
## Problem description

A command like `kiwi system build --description mydesc/ --target-dir mydest` always fail.

## Actual behaviour

Kiwi checks if the root directory is already present after creating it, thus the check is always positive, even if the root directory did not exist before calling kiwi.

See `SystemBuildTask` code inside `prepare` method:

```python
image_root = os.sep.join([abs_target_dir_path, 'build', 'image-root'])
Path.create(image_root)
```

This is executed before creating the `SystemPrepare` instance.

davidcassany:
This commit fixes the validation of an existing root directory
for the command  `system build`. System build used to create the root
directory before performing the root existance check, thus the
check was always failing in any case. The root directory is created
inside the RootInit class within the `create` method.

Fixes #349

### DIFF
--- a/kiwi/system/root_init.py
+++ b/kiwi/system/root_init.py
@@ -74,6 +74,7 @@ class RootInit(object):
         initial setup
         """
         root = mkdtemp(prefix='kiwi_root.')
+        Path.create(self.root_dir)
         try:
             self._create_base_directories(root)
             self._create_device_nodes(root)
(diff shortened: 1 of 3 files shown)


msrex:
Yay! I can change my alias :-)

schaefi:
ah one of the after effects of changing the build command behavior . Thanks much for catching this one

PR #356, opened 2017-05-30
Fix use of pre requires in spec file

schaefi:
PreReq seems to be only handled by zypper

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -228,7 +228,11 @@ outside of the scope of kiwi appliance building.
 %package -n kiwi-pxeboot
 Summary:        KIWI - PXE boot structure
 Requires:       syslinux
-PreReq:         shadow
+%if 0%{?fedora} || 0%{?rhel}
+Requires(pre):  shadow-utils
+%else
+Requires(pre):  shadow
+%endif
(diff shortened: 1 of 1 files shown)


package/python-kiwi-spec-template
```diff
@@ -228,7 +228,7 @@ outside of the scope of kiwi appliance building.
 %package -n kiwi-pxeboot
 Summary:        KIWI - PXE boot structure
 Requires:       syslinux
-PreReq:         shadow
+Requires(pre):  shadow
```
Conan-Kudo: For Red Hat/Fedora, it's `shadow-utils`, whereas SUSE calls it `shadow`.

package/python-kiwi-spec-template
```diff
@@ -228,7 +228,11 @@ outside of the scope of kiwi appliance building.
 %package -n kiwi-pxeboot
 Summary:        KIWI - PXE boot structure
 Requires:       syslinux
-PreReq:         shadow
+%if 0%{?rhel_version} || 0%{?fedora} || 0%{?centos}
```
Conan-Kudo: `0%{?fedora} || 0%{?rhel}` is sufficient here.

PR #359, opened 2017-06-01
Fix/workaround invalid xsd pattern translation

### LINKED ISSUE #347: xsd pattern translation invalid
## Problem description

The kiwi schema contains the following pattern:

```
arch-name = xsd:token {pattern = "(x86_64|i586|i686|ix86|aarch64|arm64|armv5el|armv5tel|armv6hl|armv6l|armv7hl|armv7l|ppc|ppc64|ppc64le|s390|s390x)(,(x86_64|i586|i686|ix86|aarch64|arm64|armv5el|armv5tel|armv6hl|armv6l|armv7hl|armv7l|ppc|ppc64|ppc64le|s390|s390x))*"}
```

This will be translated by generateDS.py into

```python
# Validate type arch-name, a restriction on xs:token.
        if value is not None and Validate_simpletypes_:
            if not self.gds_validate_simple_patterns(
                    self.validate_arch_name_patterns_, value):
                warnings_.warn('Value "%s" does not match xsd pattern restrictions: %s' % (value.encode('utf-8'), self.validate_arch_name_patterns_, ))
    validate_arch_name_patterns_ = [['^(x86_64$|^i586$|^i686$|^ix86$|^aarch64$|^arm64$|^armv5el$|^armv5tel$|^armv6hl$|^armv6l$|^armv7hl$|^armv7l$|^ppc$|^ppc64$|^ppc64le$|^s390$|^s390x)(,(x86_64$|^i586$|^i686$|^ix86$|^aarch64$|^arm64$|^armv5el$|^armv5tel$|^armv6hl$|^armv6l$|^armv7hl$|^armv7l$|^ppc$|^ppc64$|^ppc64le$|^s390$|^s390x))*$']]
```

Validating an XML of the form

```xml
<package name="shim" arch="x86_64,aarch64"/>
```
is valid against the schema but causes a UserWarning on the python side

imho this is a bug in the xsd pattern translation of the generateDS utility, the regexp anchors ^$ are not part of the origin xsd pattern

schaefi:
The data structures are auto generated by the generateDS
tool which works nicely except for the arch-name xsd pattern
used in the RelaxNG schema. For some reason the used regular
expression is translated by generateDS into a python
expression not matching the original expression from the
schema. The result is an invalid python warning message after
the schema has successfully validated the arch string.
The problem has been reported to the generateDS developer.

As long as their is no fix available in generateDS the
following workaround in kiwi applies: The original xs:token
pattern validation will be disabled on the generateDS
level and applies only to the schema. This Fixes #347

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,12 +53,23 @@ kiwi/schema/kiwi.rng: kiwi/schema/kiwi.rnc
 	# the short form of the RelaxNG schema to the format used
 	# in code and auto generates the python data structures
 	trang -I rnc -O rng kiwi/schema/kiwi.rnc kiwi/schema/kiwi.rng
-	trang -I rnc -O xsd kiwi/schema/kiwi.rnc kiwi/schema/kiwi.xsd
 	# XML parser code is auto generated from schema using generateDS
 	# http://pythonhosted.org/generateDS
-	generateDS.py -f --external-encoding='utf-8' \
-		-o kiwi/xml_parse.py kiwi/schema/kiwi.xsd
-	rm kiwi/schema/kiwi.xsd
+	# ---
+	# a) modify arch-name xsd:token pattern to be generic because
+	#    generateDS translates the regular expression into another
+	#    expression which is different and wrong compared to the
+	#    expression in the schema
+	cat kiwi/schema/kiwi.rnc | sed -e \
+		s'@arch-name = xsd:token.*@arch-name = xsd:token {pattern = ".*"}@' >\
+		kiwi/schema/kiwi_modified_for_generateDS.rnc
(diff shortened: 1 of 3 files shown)


schaefi:
no idea why travis complains here ??

davidcassany:
me neither, doesn't look like the changes should affect tox call in any way, did it fail before without ignoring the warnings? I saw the warning messages while running locally, but I do not remember if they were causing travis to fail.

schaefi:
that must be something on the travis side, it failed in .travis.yml at "pip install -r .travis.requirements.txt" which hasn't been touched at all... I'm confused

PR #363, opened 2017-06-06
Rearrange Docker docs

davidcassany:
This commit rearranges the Docker image builds documentation
in order to match the criteria exposed in #323.

### DIFF
--- a/doc/source/building/build_docker_container.rst
+++ b/doc/source/building/build_docker_container.rst
@@ -1,190 +1,91 @@
 Build a Docker Container Image
 ==============================
 
-.. hint::
+.. sidebar:: Abstract
 
-   Make sure you have checked out the example image descriptions
-   For details see :ref:`example-descriptions`
+   This page explains how to build a Docker base image. It containts
 
-The following example shows how to build and run a container image
-based on SUSE Leap for the docker container system.
+   * basic configuration explanation
+   * how to build a Docker image
+   * how to run it with the Docker daemon
 
-Installing the Requirements
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+KIWI is capable of building native Docker images, from scratch and derived
+ones. KIWI Docker images are considered to be native since the KIWI
+tarball image is ready to be loaded to a Docker daemon, including common
+container configurations.
 
-Building Docker images with KIWI requires a couple of additional tools.
-In practice, :command:`umoci` and :command:`skopeo` utilities are mandatory.
-These tools are not yet in all regular openSUSE repositories, however a packaged
-version can be found in the *Virtualization:containers* repository
-`here <http://download.opensuse.org/repositories/Virtualization:/containers/>`_.
-In order to install them first add the repository using the :command:`zypper ar`
-command:
+The Docker configuration metadata is provided to KIWI as part of the
+:ref:`XML description file <decription_components>` using the
+``<containerconfig>`` section. The following parameters
+can be specified (more detailed schema documentation
+:ref:`here <k.image.preferences.type.containerconfig>`):
 
-.. code:: bash
-
-   $ zypper ar -f http://download.opensuse.org/repositories/Virtualization:/containers/<DIST>
-
-where the placeholder `<DIST>` is the preferred distribution. Once the
-repository has been added install the required utilities:
-
-.. code:: bash
-
-   $ zypper in umoci skopeo
-
-Building the Image
-~~~~~~~~~~~~~~~~~~
-
-Run the following KIWI command in order to build the base image example.
-
-.. code:: bash
-
-    $ sudo kiwi-ng --type docker system build \
-        --description kiwi-descriptions/suse/x86_64/suse-leap-42.2-docker \
-        --target-dir /tmp/myimage
-
-Find the image with the suffix :file:`.docker.tar.xz` below
-:file:`/tmp/myimage`. For testing the docker image it's required
-to have the docker subsystem installed and running. The following
-command can be used to check if docker is installed and functional
-on your system:
-
-.. code:: bash
-
-    $ docker images
-
-The command above lists all available docker images. If the docker
-daemon is running and responsive a list output is expected which
-could also be empty if no images are available. In any other case
-an error message is displayed which needs to be resolved prior to
-continue with the next steps.
-
-Once docker works on your system the KIWI built image needs to be
-loaded into the docker system with the following command:
-
-.. code:: bash
-
-    $ docker load -i Docker-Leap-42.2.x86_64-1.0.0.docker.tar.xz
-
-On success the docker image is available as part of the images list as follows:
-
-.. code:: bash
-
-    $ docker images
-
-    REPOSITORY          TAG  ...
-    opensuse            42.2 ...
-
-In order to run an interactive shell session of the loaded container
-the following command needs to be called:
-
-.. code:: bash
-
-    $ docker run -it opensuse:42.2 /bin/bash
-
-Container Configuration
-~~~~~~~~~~~~~~~~~~~~~~~
-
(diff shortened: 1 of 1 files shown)


schaefi:
Thanks David

@tomschr could you have a short look ?

doc/source/building/build_docker_container.rst
```diff
@@ -1,190 +1,91 @@
...
-Container Configuration
-~~~~~~~~~~~~~~~~~~~~~~~
-
-A docker container includes a set of information also called metadata.
-The metadata information is provided within the KIWI XML description
-as part of the ``<containerconfig>`` section. The following parameters
-can be specified:
+The Docker configuration metadata is provided to KIWI as part of the
+:ref:`XML description file <decription_components>` using the
+``<containerconfig>`` section. The following parameters
```
tomschr: As we talk about XML, we should better use the term "tag" or "element" instead of "section".

tomschr: Same with "parameters". I guess, these are (XML) attributes, right? If yes, better use "attributes" here.

doc/source/building/build_docker_container.rst
```diff
@@ -1,190 +1,91 @@
...
-* ``cmd``: Provides the default execution parameters of the container.
-* ``label``: Adds custom metadata to an image, is key-value pair.
+* ``subcommand``: Provides the default execution parameters of the container.
+  Equivalent of the `CMD` directive of a Docker file.
+* ``labels``: Adds custom metadata to an image, is key-value pair. Equivalent
+  of the `LABEL` directive of a Docker file.
 * ``expose``: Informs at which ports is the container listening at runtime.
-* ``env``: Sets an environment value, is key-value pair.
+  Equivalent of the `EXPOSE` directive of a Docker file.
+* ``environment``: Sets an environment values using key-value pairs.
```
tomschr: Plural or singular? I assume plural here, in that case I would rephrase to:

    Set environment values using key-value pairs.

It would be helpful to add *how* a single key-value pair can be separated from another one. Use spaces, commas, or something different?

doc/source/building/build_docker_container.rst
```diff
@@ -1,190 +1,91 @@
...
+  Equivalent of the `env` directive of a Docker file.
 * ``entrypoint``: Sets the executable that the container will run, it can
-  include parameters.
-* ``volume``: Creates a mountpoint with the given name and marks it to hold
-  external volumes from the host or from other contianers.
-* ``user``: Sets the user name or UID to be used when running `cmd` and
-  `entrypoint`.
+  include parameters. Equivalent of the `ENTRYPOINT` directive of a Docker
+  file..
+* ``volumes``: Creates mountpoints with the given name and marks it to hold
```
tomschr: Create mountpoints...

doc/source/building/build_docker_container.rst
```diff
@@ -1,190 +1,91 @@
...
+  external volumes from the host or from other contianers. Equivalent of the
+  `VOLUME` directive of a Docker file.
+* ``user``: Sets the user name or UID to be used when running `entrypoint` and
```
tomschr: I would slightly rephrase it as:  "Sets the user name or user id (UID)..."

doc/source/building/build_docker_container.rst
```diff
@@ -1,190 +1,91 @@
...
-
-* `ADD`
+  `entrypoint`. Equivalent of the `WORKDIR` directive of a Docker file.
 
-  In order to include a tarball and auto extract it in KIWI it can be done by
-  including the files in the :ref:`overlay tree <decription_components>`,
-  or by including an `<archive>` item inside the `<packages type="image">`
-  section. **ADD** also handles remote content, in KIWI this should be done
-  within the `config.sh` script.
+Other Docker file directives such as ``RUN`` and ``COPY`` or ``ADD``, can be
```
tomschr: Correct with:

  ... ``RUN``, ``COPY`` or ``ADD``, ...

doc/source/building/build_docker_container.rst
```diff
@@ -1,190 +1,91 @@
...
 
-  In order to execute commands during build time in KIWI it can be done by
-  extending the `config.sh` script. Note that the `config.sh` is executed
-  after all the packages have been installed and after the overlay tree has
-  been applied.
+1. Make sure you have checked out the example image descriptions,
+   see :ref:`example-descriptions`.
 
-* `COPY`
+2. Install the required toolchain for Docker images building.
```
tomschr: IMHO, I would separate the step of including the repo from installing the packages like this:

```
1. Include the ``Virtualization`` repository to your list:

   .. code:: bash
      $ zypper ar -f http://download.opensuse.org/repositories/Virtualization:/containers/<DIST>

2. Install the :command:`umoci` and :command:`skopeo` tools:
  [...]
```

doc/source/building/build_docker_container.rst
```diff
@@ -1,190 +1,91 @@
...
-is that the base image must be specified in the description file, it can
-be done using the **derived_from** optional attribute of ``<type>`` tag.
-The value of the **derived_from** attribute is the URI of the image;
-currently KIWI only supports references to local files (it expects a xz
-compressed tarball image) and any other URI type that is supported by
-the skopeo tool (for example DockerHub URIs as
-``derived_from="docker://opensuse:leap"``). The following example
-type specification shows how to specify a base image setup:
+   where the placeholder `<DIST>` is the preferred distribution. Then install
+   :command:`umoci` and :command:`skopeo` tools
```
tomschr: See [previous comment](https://github.com/SUSE/kiwi/pull/363/files#r120322349).

davidcassany:
@tomschr I think I have addressed all your comments, I have divided the list of configurable parameters between attributes and tags, I think this way is more clear. I did not want to document how to explicitly configure the Docker image, I assume that the users of this feature are already familiar with the Dockerfile syntax, so I just wanted to give an overview at a glance of the configuration possibilities, rather than explain them in detail.

schaefi:
recompiled the docs and to me it looks great 👍

tomschr:
@davidcassany 
>  I think I have addressed all your comments,

Great, thank you for your efforts!

>  I have divided the list of configurable parameters between attributes and tags, I think this way is more clear.

Yes, much better! 👍

> I did not want to document how to explicitly configure the Docker image, I assume that the users of this feature are already familiar with the Dockerfile syntax, so I just wanted to give an overview at a glance of the configuration possibilities, rather than explain them in detail.

Yes, totally agree. If you like and know a good link (maybe upstream Docker docu?) you can reference it and delegate it to to remote documentation. No need to copy existing texts.

tomschr:
@davidcassany, @schaefi after I've built the doc myself, I have a final question. 😉

How should we display commands that require root privileges?

In most cases, we use `$` as a prompt. IMHO it could be misleading. Although the experienced user would probably know how to deal with it, nonetheless we should be clear about that. I see the following solutions to it:

1. Use `#` to indicate the root prompt.
2. Use `sudo`.
3. Leave it as is and assume all users are experienced

This is mostly important for the `zypper` and `docker` commands, depending on how the user configured its system. Regardless what we use, we should be consistent. 😉

What do you think?

schaefi:
I prefix any command with a '$' and for those which require sudo i write '$ sudo'. I'm not sure if this follows the common practice but I did that all over the documentation

tomschr:
> I prefix any command with a '$' and for those which require sudo i write '$ sudo'.
> I'm not sure if this follows the common practice but I did that all over the documentation.

The only "common practice" is to be consistent. 😉 It doesn't matter if it is with `sudo` or with `#` as long as the reader knows what to do and what to expect.

So no need to change anything here. I just stumbled upon it. Sorry for the noise.

PR #368, opened 2017-06-08
Update PXE building chapter

### LINKED ISSUE #323: Doc Redesign: Improve Task-Based Articles

schaefi:
Redesign PXE building chapter to follow style guide.
This Fixes #323

### DIFF
--- a/doc/source/building/build_pxe_root_filesystem.rst
+++ b/doc/source/building/build_pxe_root_filesystem.rst
@@ -1,79 +1,119 @@
-Build a PXE root File System Image
+.. _build_pxe:
+
+Build a PXE Root File System Image
 ==================================
 
-.. toctree::
-   :maxdepth: 1
+.. _PXE: https://en.wikipedia.org/wiki/Preboot_Execution_Environment
+.. _TFTP: https://en.wikipedia.org/wiki/Trivial_File_Transfer_Protocol
+.. _NBD: https://en.wikipedia.org/wiki/Network_block_device
+.. _AoE: https://en.wikipedia.org/wiki/ATA_over_Ethernet
 
-   build_pxe_root_filesystem/setup_pxe_bootserver
 
-.. hint::
+.. sidebar:: Abstract
 
-   Make sure you have checked out the example image descriptions
-   For details see :ref:`example-descriptions`
+   This page explains how to build a file system image for use with
+   KIWI's PXE boot infrastructure. It contains:
 
-PXE is a network boot protocol that is shipped with most BIOS
+   * how to build a PXE file system image
+   * how to setup the PXE file system image on the PXE server
+   * how to run it with QEMU
+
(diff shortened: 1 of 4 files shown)


schaefi:
@tomschr this one is not so easy. As you can see I described several different configuration flavors. Do you think we should put those into "working with images" chapters ? and if so how to deal with it in a good way ?

I used a section to describe all possible configuration parameters in order to provide an overview. next I created subsections describing use cases. In the end all use cases covered all possible configuration parameters. I'm not sure how a good split could look like ?

Feel free to commit to the branch if you'd like to demonstrate your idea

Thanks much

doc/source/building/build_pxe_root_filesystem.rst
```diff
@@ -6,74 +6,103 @@ Build a PXE root File System Image
 
    build_pxe_root_filesystem/setup_pxe_bootserver
 
-.. hint::
```
tomschr: Better use "Root" instead of "root" in the title "Build a PXE root File System Image"

tomschr:
@schaefi I took your recommendation of "feel free to commit" to demonstrate my idea. 😉

I'm not 100% sure if this is the best solution, but let's see how you like it.

I did:

* Moved the `setup_pxe_bootserver.rst` to `source/building/working_with_images/` and added a references (in that file and in the referenced file)
* Move the "PXE Client Setup Configuration" section into a separate file `building/working_with_images/pxe_client_configuration.rst` and add references.
* Include the two files into 'working_with_images.rst`
* Fix some markup issues

The previous commits changes these. I didn't get any errors, maybe we should adjust some texts here and there.

Let me know what you think about this change.

schaefi:
great thanks much, I like it. I squashed the commits and update the base commit message

PR #369, opened 2017-06-08
Improve package/repository signature checking setup

### LINKED ISSUE #358: Improve package/repository signature checking setup
## Problem description

KIWI currently has only global flags for rpm signature checks, but doesn't support global repo signature checks and per-repository overriding of rpm/repo signature check behavior.

When appliance makers use a mix of repositories from different sources, this granularity matters.

For example:

* In Fedora and Mageia, *all* distribution repositories and third party repositories hosted by COPR are package signed, but all but one repository is _**not**_ repo signed. That means that the global default would be rpm sig checks on, repo sig checks off, but for one repo, repo sig checks would be switched on.
* In CentOS, the primary distribution repositories are package _and_ repo signed, but most ancillary and third party repositories are _not_ repo signed. This means the global default would be rpm sig checks on, repo sig checks off, and distribution repos would have repo sig checks switched on.
* In openSUSE/SLE, _all_ repositories are generally package and repo signed, so the global default would be rpm and repo sig checks on. In the event something is not signed, they can be switched off individually.

Without this granularity, it's impossible to provide the maximum assurance that the content sources are, in fact, secure.

This would also probably necessitate adding a field for specifying each repository's gpg public key file.

davidcassany:
This commit adds repository_gpgcheck and package_gpgcheck attributes
to the repository tag in schema. With this change each repository
can be configured individually to enable repository signatures check
and/or enable signature package check. The configuration is done at
repository level, thus it is not conflicting with the wide package
manager configuration flag <rpm-signatures-check>.

Fixes #358


kiwi/schema/kiwi.rnc
```diff
@@ -1254,6 +1254,16 @@ div {
...
+    k.repository.repository_gpgcheck.attribute =
+        ## Specify whether or not this specific repository is configured to
+        ## to run repository signature validation. If not set, no value is
+        ## appended into the repository configuration file.
+        attribute repository_gpgcheck { xsd:boolean }
+    k.repository.package_gpgcheck.attribute =
+        ## Specify whether or not this specific repository is configured to
+        ## to run package signature validation. If not set, no value is
+        ## appended into the repository configuration file.
+        attribute package_gpgcheck { xsd:boolean }
```
schaefi: One addition to the schema docs. We should fix the following:

```
#==========================================
# common element <rpm-check-signatures>
#
div {
    k.rpm-check-signatures.content = xsd:boolean
    k.rpm-check-signatures.attlist = empty
    k.rpm-check-signatures =
        ## Setup a Package Signature
        [
        db:para [
            "Setup if the package manager should check the package\x{a}"~
            "signature or not. This option could be ignored according\x{a}"~
            "to the used package manager."
        ]
        ]
        element rpm-check-signatures {...
```

* delete the db:para, we don't use that anymore since the docs are not docbook related anymore
* update the very short comment, explaining that this is the system wide package manager setup and can be overwritten by repository specific settings. Also I assume repo specific settings are higher in priority and overwrites the system wide setup iirc

kiwi/repository/zypper.py
```diff
@@ -233,19 +233,28 @@ def add_repo(
...
+                    name, 'priority', format(prio)
+                )
+            with open(repo_file, 'w') as repo:
+                repo_config.write(repo)
         self._restore_package_cache()
 
     def import_trusted_keys(self, signing_keys):
         """
-        Imports trusted keys into the image
+        Imports trusted keys into the imagen
```
schaefi: looks like a mistake ?

schaefi:
looks great 👍 also realized generateDS comes with a new version

PR #376, opened 2017-06-12
Prevent unneded packages in kiwi boot image

schaefi:
Cleanup isoboot descriptions
    
There is no need to install bootloader packages to the initrd, all information regarding the bootloader setup is taken from the system root directory. This one also includes refactoring and the start to cleanup the suseGFXBoot shell function


msrex:
Would the same isoboot-cleanup also apply to the non-SUSE distros?

schaefi:
yes that was part of the intention to get rid of the suseGFXBoot stuff. With this commits I have tested Fedora, Centos and suse. Today I'll test ubuntu

schaefi:
Also worked for Ubuntu

PR #377, opened 2017-06-12
Include imageonly attribute for repositories

### LINKED ISSUE #362: Set repos just for 'imageinclude' not for build time
## Description

Add the possibility to define repositories to be included into the image which are not actually used during build time.

This issues is included to discuss and track the latest comments from here #178

davidcassany:
This commit adds imageonly attribute support for the repository
element. imageonly is a boolean attribute that if true indicates
that the repository is not used for the build but needs to be
configured for the resulting image.

Fixes #362

### DIFF
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -99,7 +99,8 @@ class SystemPrepare(object):
         :rtype: PackageManager
         """
         repository_options = []
-        repository_sections = self.xml_state.get_build_repository_sections()
+        repository_sections = \
+            self.xml_state.get_repository_sections_used_for_build()
         package_manager = self.xml_state.get_package_manager()
         if self.xml_state.get_rpm_check_signatures():
             repository_options.append('check_signatures')
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -99,7 +99,7 @@ class SystemPrepare(object):
         :rtype: PackageManager
         """
         repository_options = []
-        repository_sections = self.xml_state.get_repository_sections()
+        repository_sections = self.xml_state.get_build_repository_sections()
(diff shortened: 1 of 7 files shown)


kiwi/system/prepare.py
```diff
@@ -99,7 +99,7 @@ def setup_repositories(self, clear_cache=False, signing_keys=None):
         :rtype: PackageManager
         """
         repository_options = []
-        repository_sections = self.xml_state.get_repository_sections()
+        repository_sections = self.xml_state.get_build_repository_sections()
```
schaefi: Could we rename this to

```python
get_repository_sections_used_for_build()
```

kiwi/system/setup.py
```diff
@@ -115,7 +115,7 @@ def import_repositories_marked_as_imageinclude(self):
         imageinclude attribute should be permanently added to
         the image repository configuration
         """
-        repository_sections = self.xml_state.get_repository_sections()
+        repository_sections = self.xml_state.get_image_repository_sections()
```
schaefi: Could we rename this to

```python
get_repository_sections_used_in_image()
```

schaefi: Reason for the name change is that I was confused what an image_repository is. With the background of the existing imageinclude and imageonly attributes it's clear but I think we could make it more explicit by a name change of the functions. What do you think ?

schaefi:
Looks great to me, nice python coding :) I should do more of these too. Just a minor issue with the function names. Let me know what you think

kiwi/system/setup.py
davidcassany: Yes, that looks great to me, I am not very good at naming stuff... I'll change that, I'm totally fine with more explicit names.

PR #378, opened 2017-06-14
Removing has_repositories_marked_as_imageinclude method

davidcassany:
With the current repository management this method is not required anymore, since the setup repositories method does not modify the image if no repositories are defined to be configured in the image.

It is related to #305 and #191. `has_repositories_marked_as_imageinclude` was included in #191 in order to avoid purging any repository if no repositories were configured to be included in the image. Since #305 including repositories in the image does not purge any already present repository.


schaefi:
:+1: less code makes sense :)

PR #379, opened 2017-06-14
Consider only repositories used for build in image info

davidcassany:
This commit makes sure that the repositories marked with `imageonly`
flag are not included in the packages solve procedure of the image info task.

This is related to #362


schaefi:
:+1:

PR #380, opened 2017-06-14
Fixup class docstrings

schaefi:
The attribute list should provide information about the
construction of an object of this class. Some fields
were missing or superfluous

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -72,6 +72,7 @@ class DiskBuilder(object):
 
     * :attr:`custom_args`
         Custom processing arguments defined as hash keys:
+        * signing_keys: list of package signing keys
     """
     def __init__(self, xml_state, target_dir, root_dir, custom_args=None):
         self.arch = platform.machine()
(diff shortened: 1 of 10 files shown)


schaefi:
@davidcassany would you mind to take a short look at this one. I stumbled over this when implementing the xz compression options and thought this should be cleaned up first.

Thanks

davidcassany:
@schaefi sure I am having a look

kiwi/builder/disk.py
```diff
@@ -61,115 +61,17 @@ class DiskBuilder(object):
...
 
-    * :attr:`generic_fstab_entries`
-        List of generic/persistent fstab entries
+    * :attr:`root_dir`
+        Root directory path name
 
-    * :attr:`result`
-        Instance of Result
+    * :attr:`custom_args`
+        Custom processing arguments defined as hash keys:
```
davidcassany: nit-pick, but I guess the colons should not be there if no key-values defined for the disk build.

schaefi: I just missed to add the key information... my bad, thanks for the review

Issue #373, opened 2017-06-09
Allow to pass in globally used compression options


schaefi:
## Problem description

In KIWI we use instances of the Compress class to implement compression, but there is no way to pass custom compression options. A customer reported that the default compression level for e.g xz was not sufficient and he wanted to change it which was possible in KIWI v7. Changing the level would significantly speed up the build (with an acceptable decrease in compression results) but there was no
way for the customer to tell KIWI

PR #384, opened 2017-06-19
Runtime config

schaefi:
Implement reading of runtime configuration file in base commandline class. 

This is implementing one required part of issue #373

### DIFF
--- a/kiwi/runtime_config.py
+++ b/kiwi/runtime_config.py
@@ -47,9 +47,12 @@ class RuntimeConfig(object):
                 self.config_data = yaml.load(config)
 
     def get_xz_options(self):
-        if self.config_data and 'xz' in self.config_data:
+        return self._get_attribute_list(element='xz', attribute='options')
+
+    def _get_attribute_list(self, element, attribute):
+        if self.config_data and element in self.config_data:
             try:
-                return self.config_data['xz'][0]['options'].split()
+                return self.config_data[element][0][attribute].split()
             except Exception as e:
                 raise KiwiRuntimeConfigFormatError(
                     '{error_type}: {error_text}'.format(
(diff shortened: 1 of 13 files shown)


kiwi/runtime_config.py
```diff
@@ -0,0 +1,61 @@
...
+
+        config_file = os.sep.join(
+            [self._home_path(), '.config', 'kiwi', 'config.yml']
+        )
+        if os.path.exists(config_file):
+            log.info('Reading runtime config file: {0}'.format(config_file))
+            with open(config_file, 'r') as config:
+                self.config_data = yaml.load(config)
+
+    def get_xz_options(self):
```
davidcassany: I am not sure if adding an specific method for that is required, couldn't we have a something like:

```python 
def get_cmd_options(self, cmd):
  if self.config_data and cmd in self.config_data:
    try:
      return self.config_data[cmd][0]['options'].split()
    except Exceptions as e:
      raise KiwiException(e)
```

I think it could be nice to have it generic, in a way that we can easily call it and use this method for other commands without having to update the RuntimeConfig class. Not very relevant commend since I think we won't be using it that much in the future, but just an idea.

schaefi: I like the idea and changed the code a bit, how about it now ?

davidcassany: 👍 yes, that looks goods to me

schaefi: For accessing runtime config information I'd like to keep the explicit methods reading the information instead of methods with parameters. The reason is that I'm not sure if we won't be
changing the format of the config file at same point in time, but the method name will be sticky

davidcassany: Yes, it makes sense

schaefi: cool thanks, merging :)

PR #386, opened 2017-06-19
Include patternType information to resolv packages in image info task

### LINKED ISSUE #381: Task image_info does not take into account the patternType to resolve packages
## Problem description

The command `image info --resolve-package-list` includes different packages than the ones actually used if running a `system build`.

## Actual behaviour

Apparently the `image info` includes recommended packages when it shouldn't.

davidcassany:
This commit includes ingore_recommended flag in the Sat.solve method.
This way if the description file states to include only required
packages (without recommendations) it is respected and taken into
account to resolv the packages list.

Fixes #381

### DIFF
--- a/kiwi/tasks/image_info.py
+++ b/kiwi/tasks/image_info.py
@@ -99,10 +99,12 @@ class ImageInfoTask(CliTask):
 
         if self.command_args['--resolve-package-list']:
             solver = self._setup_solver()
-            package_list = self.xml_state.get_bootstrap_packages() + \
+            boostrap_package_list = self.xml_state.get_bootstrap_packages() + \
+                [self.xml_state.get_package_manager()]
+            package_list = boostrap_package_list + \
                 self.xml_state.get_system_packages()
             bootstrap_packages = solver.solve(
-                self.xml_state.get_bootstrap_packages(), False,
+                boostrap_package_list, False,
                 True if self.xml_state.get_bootstrap_collection_type() is
                 'onlyRequired' else False
             )
--- a/kiwi/tasks/image_info.py
+++ b/kiwi/tasks/image_info.py
@@ -101,7 +101,17 @@ class ImageInfoTask(CliTask):
             solver = self._setup_solver()
             package_list = self.xml_state.get_bootstrap_packages() + \
                 self.xml_state.get_system_packages()
-            solved_packages = solver.solve(package_list)
+            bootstrap_packages = solver.solve(
+                self.xml_state.get_bootstrap_packages(), False,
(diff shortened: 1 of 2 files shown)


kiwi/tasks/image_info.py
```diff
@@ -101,7 +101,17 @@ def process(self):
...
+                self.xml_state.get_bootstrap_packages(), False,
+                True if self.xml_state.get_bootstrap_collection_type() is
+                'onlyRequired' else False
+            )
+            solved_packages = solver.solve(
+                self.xml_state.get_system_packages(), False,
+                True if self.xml_state.get_system_collection_type() is
+                'onlyRequired' else False
+            )
+            solved_packages.update(bootstrap_packages)
```
schaefi: I guess you moved from one solver operation to two solver operations because in theory we allow two different collection types between bootstrap and image packages processing, right ?

The only point I have here is that this will slow down the operation a bit. As there is no way out I just would like to ask if you have tested a real solver operation and have checked if the results stayed the same compared to the former code ?

Thanks

schaefi: If you can answer the question with a "yes" feel free to merge :)

davidcassany: Yes this is right, I made the solving in two operations to be able to apply different criteria for bootstrap and image packages.

I did some tests to validate the result was the same, but I'll make some more accurate tests, as I haven't done it with a complete image including the recommended packages everywhere.

davidcassany: @schaefi damm... your doubts pointed to the right direction... there are little differences within resolving bootstrap and image package all at once or in separate calls. Very little differences, but differences after all.

I used Leap 42.2 JeOS for the tests, and there were about 4 or 5 different libraries. In that case I think that or we find a way to make cumulative calls (at first sight I did not see how) or we assume a common configuration for bootstrap and image packages.

davidcassany:
@schaefi, the current implementation is not perfect, but I'd say better than before. With the docker image description I got the same result comparing the `*.packages` of the built image (with and without `patternType="plusRecommended"`) and the resolved packages from `image info`. Also I got the same result of `image info` with and without the current patch if `patternType="plusRecommended"`.

However when I did the same test using the `suse-leap-42.2-JeOS` live iso description I found out that `image info` is including `systemd-logger` which is not present in the resulting image and also that it misses `python3-gobject`.

I doubt we can achieve exactly same result or guarantee we provide the same result that will end up into the image. Basically because the solving is handled differently in those cases, zypper wrapper of libsolv will always be different and also I could not find a way in our Sat class to solve image packages assuming that all booststrap transaction is already there.

PR #390, opened 2017-06-20
Added creation of zsync control file in bundler

### LINKED ISSUE #364: Provide zsync serverside file from image
## Problem description

In order to keep a big data blob like an image in sync it would be great to have the opportunity to update only the changed blocks of the image file instead of transferring the entire blob with every change.
The zsync tool allows to do that. KIWI should be extended to provide a .zsync variant via:

```
$ zsyncmake image
```

as part of a bundler call

## Actual behaviour

The KIWI bundler provides its data not as a zsync variant
Also see: http://zsync.moria.org.uk/server for details

schaefi:
If a --zsync_source url is provided to the bundler call a
zsync control file is created for the compressed parts of
the bundle. This Fixes #364

### DIFF
--- a/doc/source/commands/result_bundle.rst
+++ b/doc/source/commands/result_bundle.rst
@@ -10,6 +10,7 @@ SYNOPSIS
 
    kiwi result bundle -h | --help
    kiwi result bundle --target-dir=<directory> --id=<bundle_id> --bundle-dir=<directory>
+       [--zsync_source=<download_location>]
    kiwi result bundle help
 
 DESCRIPTION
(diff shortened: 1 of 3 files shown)


Conan-Kudo:
This is optional right? zsyncmake doesn't require a URL, and can generate ones that work from any arbitrary source location...

schaefi:
If you don't give zsyncmake an url it prints a warning that it adds the basename of the file as url. As it therefore places a value in any case we pass -u with the value of the option which makes this explicit which I prefer over implicit actions happening in the background

if you don't specify a zsync source url it indicates you don't want it :) and yes when calling zsync the url can be overwritten

schaefi:
Also send a submit request to the buildservice to allow to check if kiwi has this feature
https://github.com/openSUSE/obs-build/pull/367

probonopd:
What is the progress here, who would need to do what so that a zsync file would start showing up in http://download.opensuse.org/repositories/home:/probono:/branches:/KDE:/Medias/images/iso/?

PR #391, opened 2017-06-21
Fixup repository setup with no type declaration

schaefi:
The type attribute in a repository section is optional, if not
specified the zypper add_repo code throws a KiwiRepoTypeUnknown
exception. However this is not needed because the repository
could be added without a type declaration in order to let zypper
auto detect the repository type

### DIFF
--- a/kiwi/repository/apt.py
+++ b/kiwi/repository/apt.py
@@ -125,7 +125,7 @@ class RepositoryApt(RepositoryBase):
         }
 
     def add_repo(
-        self, name, uri, repo_type='deb',
+        self, name, uri, repo_type=None,
         prio=None, dist=None, components=None,
         user=None, secret=None, credentials_file=None,
         repo_gpgcheck=None, pkg_gpgcheck=None
--- a/kiwi/repository/dnf.py
+++ b/kiwi/repository/dnf.py
@@ -124,7 +124,7 @@ class RepositoryDnf(RepositoryBase):
         }
 
     def add_repo(
-        self, name, uri, repo_type='rpm-md',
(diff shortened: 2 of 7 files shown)


Conan-Kudo:
Zypper's autodetection of repo types doesn't work. It always falls back to type "NONE", which makes Zypper behave oddly.

Conan-Kudo:
Yum and DNF only support `type=rpm-md`, anything else won't work.

schaefi:
well --type is only an option for zypper, if it can't cope correctly with it, it should be fixed. The changes here are basically to stay compatible with the schema which allowed to leave out the type attribute since a very long time. People actually use zypper addrepo without a type quite often, thus I don't want to force them into a type declaration even though we recommend to specify one

schaefi:
> Yum and DNF only support type=rpm-md, anything else won't work.

right and the repo_type parameter is therefore not used in the implementation and should not have a value

Conan-Kudo:
> People actually use zypper addrepo without a type quite often, thus I don't want to force them into a type declaration even though we recommend to specify one

I agree it should work, it's just that since openSUSE 13.2, it's been half broken. :(

schaefi:
man you are so right, I just did a short test in 13.2, Leap42.1, Leap42.2 and even Factory and it's all broken :( ok I think I'll revert this

Conan-Kudo:
@schaefi It bit me when I first started supporting openSUSE for the Datto Linux Backup Agent, I have no idea why this isn't fixed.

The fedora-repos package for Fedora 26+ also got adjusted to include `type=` because Zypper broke everyone else by putting `type=NONE`. :(

PR #394, opened 2017-06-25
Build schema documentation with Oxygen

schaefi:
Given there is a valid Oxygen license owned by the company
as described in https://www.oxygenxml.com/oxygen_scripting.html
the schema documentation can be created using Oxygen. This
commit changes the way the schema docs are generated in the
following way:

1. Install Oxygen and setup the license to activate the tool

2. Setup the KIWI development shell environment to export
   the schema_tool variable to point to the schemaDocumentation.sh
   script provided by the Oxygen installation

   export schema_tool=/path/to/Oxygen/schemaDocumentation.sh

3. In any case of a schema change manually call:
    
    tox -e schema
    
    This step was done before as part of the tox doc target
    but can't be done automatically because it would require
    a correctly licensed Oxygen installation in the travis
    environment. Thus the result data has to be part of the
    pull request

4. Build the documentation and review the result

    tox -e doc

5. Create the pull request

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ completion/kiwi-ng.sh
 kiwi/schema/kiwi.xsd
 kiwi/schema/xsi.xsd
 kiwi/schema/kiwi.revision
+doc/source/development/schema/xsdDocHtml.css
 
 # C tool binaries
 tools/dcounter
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -54,7 +54,7 @@ clean:
 
 .PHONY: html
 html:
-	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html && cp -a source/development/schema/images $(BUILDDIR)/html/development || true
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
--- a/doc/source/.static/css/custom.css
+++ b/doc/source/.static/css/custom.css
@@ -1,0 +1,447 @@
+/*----------------------------------------
+    Global 
+-----------------------------------------*/
+table {
+    font-family:arial, helvetica, sans-serif;
+    font-size:12px;
+}
+
+/*--------------------------------------------
+    Source code in the instance, source or 
+    annotations.
+--------------------------------------------*/
+span.tokenElement {
+    color: #000096;
+    background-color:inherit;
+}
+span.tokenAttrName {
+    color: #F5844C;
+    background-color:inherit;
+}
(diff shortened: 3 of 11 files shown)


schaefi:
Some words to the changes here:

* This is just a suggestion to come up with an alternative way to create the schema docs. It does not mean we have to go that path or that I'm not happy with the result of schema_parser.py. I received some feedback that people are confused why we did not stick to the doc style like we have used it with the legacy kiwi from here: https://doc.opensuse.org/projects/kiwi/schema-doc/

* The doc generator used here uses the SUSE owned Oxygen 10.3 license which was the same version which has generated the kiwi v7 document

* My idea with this pull request is:

  1. provide the same style people were used to with the next generation kiwi
  2. give us full freedom to work on an open solution based on schema_parser.py without being in a hurry to come up with a comparable solution

Let me know what you think

Thanks

schaefi:
For reference the version we have a license for can be found here:

* http://archives.oxygenxml.com/Oxygen/Editor/InstData10.3/Linux64/VM/oxygen.sh

Not required for testing though, that would only be needed if the schema changes

Conan-Kudo:
I'm not a fan of making kiwi documentation generation dependent on a proprietary tool. This makes it hard, if not impossible, for third parties to properly create the docs.

Is there *any* FOSS alternative that can come up with something similar and useful?

schaefi:
> Is there any FOSS alternative that can come up with something similar and useful?

All my researches in that direction did not came up with something similar. That's why we started with
helper/schema_parser.py which produces good results but not as complete and helpful oxygen can do.
I'm not a fan of proprietary tools too

> This makes it hard, if not impossible, for third parties to properly create the docs.

So this one applies to the online documentation only which is not going to be created by third parties.
Contributions can be done in the same way as before. I agree testing how a change would look like after the schema doc generator, would require the generator to be present on the host

davidcassany:
>* My idea with this pull request is:
>
>  1. provide the same style people were used to with the next generation kiwi
>  2. give us full freedom to work on an open solution based on schema_parser.py without being in a hurry to come up with a comparable solution

I agree with this and I like the idea of having a complete documentation as we used to have without having to invest big efforts of it. 

Then about working on an open based solution to provide that kind of documentation I really like the idea and I would definitely enjoy collaborating on it, but I am not sure it should be done in schema_parser.py as a KIWI specific helper. Improving schema_parser.py to collect most (not to say all) the data we need is a reasonable task, coding all the rendering to have a nice and organized representation of it looks like a tough task that I am not sure it should be part of the KIWI project.

schaefi:
yeah this surely has potential to become its own project, I fully agree

schaefi:
[Invest in a Schema Documentation Project](https://trello.com/c/JUHtK6pb/209-invest-in-a-schema-documentation-project)

PR #396, opened 2017-06-27
Flag --ignore-repos do not ignore imageonly repos

### LINKED ISSUE #395: --ignore-repos flag should not delete imageonly repos
## Problem description

`--ignore-repos` flag ignores all repos from the description file. That makes difficult to make use of the `imageonly` repository attribute in the buildservice.

## Actual behaviour

When setting the `--ignore-repos` flag all repositories listed in the XML are deleted, when only the ones used for build should be deleted. This way it would be still possible to explicitly configure repositories to into the resulting image.

This issue is related to #393

davidcassany:
This commit fixes #395, with it, using --ignore-repos, does not
delete imageonly repositories from the description file. This way
imageonly is prepared to be used in the buildservice even when using
the 'obsrepositories:/' reference style.

### DIFF
--- a/kiwi/tasks/image_info.py
+++ b/kiwi/tasks/image_info.py
@@ -76,7 +76,7 @@ class ImageInfoTask(CliTask):
         )
 
         if self.command_args['--ignore-repos']:
-            self.xml_state.delete_repository_sections()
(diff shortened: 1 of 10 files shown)


schaefi:
👍

PR #401, opened 2017-06-29
Allow imageinclude repositories inside the build service

### LINKED ISSUE #397: Allow imageinclude repository attribute inside the buildservice
## Problem description

Repositories marked with `imageinclude` are not permitted if KIWI is running inside the OBS, however some URL mapping updates in OBS are likely to permit `imageinclude` without leading to inconsistent repository configurations.

## Actual behaviour

The kiwi runtime checks verify if there is any repository with `imageinclude` and then if Defaults.is_obs_worker() returns true it throws an exception. Runtime check criteria needs to be updated.

Related with #393

davidcassany:
With this PR `imageinclude` attribute can be used in buildservice projects. It is a reponsability of the buildservice to translate those repositories to a correct URL using the `--ignore-repos` and `--add-repo` flags at runtime.

Fixes #397

### DIFF
--- a/kiwi/runtime_checker.py
+++ b/kiwi/runtime_checker.py
@@ -74,7 +74,7 @@ class RuntimeChecker(object):
                 repo_source = xml_repo.get_source().get_path()
                 repo_type = xml_repo.get_type()
                 uri = Uri(repo_source, repo_type)
-                if not uri.is_remote() or Defaults.is_obs_worker():
+                if not uri.is_remote():
                     raise KiwiRuntimeError(message % repo_source)
 
     def check_target_directory_not_in_shared_cache(self, target_dir):


schaefi:
nice one :) 👍

mlschroe:
What would I do with imageinclude? This can't be used in OBS at all, as you can't access such repos. imageonly would make sense. But, as said earlier, you'd need to have all of the other repo attributes like gpgcheck as well, so I don't know if this is feasible.

mlschroe:
(oops, wrong issue. The comment above was supposed to be in issue 400)

mlschroe:
(My point being: I can't translate those repositories because I can't specify imageonly on the command line)

PR #405, opened 2017-06-30
Do not translate obs scheme to suse scheme for imageonly repos

### LINKED ISSUE #404: obs to suse scheme uri translation should apply only to the repositories used for build
## Problem description

If `xml_state.translate_obs_to_suse_repositories()` method applies also to `imageonly` repositories then those are converted to a local URI which will not be useful for the resulting image. Moreover, those repositories, since are not used for build they do not need any kind of specific treatment when used inside of the buildservice.

## Actual behaviour

Using a repository section like 

```xml
<repository>
  <source path="obsrepositories:/"/>
</repository>
<repository type="rpm-md" imageonly="true">
  <source path="obs://openSUSE:Factory/snapshot"/>
</repository>
```
leads to include a repository with a local dir URI. KIWI should perform a regular `obs` type translation in that case, meaning the obs to suse translation should not happen.

davidcassany:
This PR enables to use obs style URI for imageonly repositories inside the build service.

Fixes #404

### DIFF
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -1216,7 +1216,7 @@ class XMLState(object):
         This will result in a local repo path suitable for a
         buildservice worker instance
         """
-        for repository in self.get_repository_sections():
+        for repository in self.get_repository_sections_used_for_build():
             source_path = repository.get_source()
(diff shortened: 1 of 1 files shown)


schaefi:
👍

PR #411, opened 2017-07-03
Accept an empty string as a repo priority

### LINKED ISSUE #408: Exception when a commandline repo has no priority
## Problem description
Using --add-repo to add a repo without a priority results in an exception because the None value cannot be converted into an integer:
```
  File "/usr/bin/kiwi", line 9, in <module>
    load_entry_point('kiwi==9.8.1', 'console_scripts', 'kiwi-ng-3')()
  File "/usr/lib/python3.4/site-packages/kiwi/kiwi.py", line 54, in main 
    App()
  File "/usr/lib/python3.4/site-packages/kiwi/app.py", line 35, in __init__
    task_class().process()
  File "/usr/lib/python3.4/site-packages/kiwi/tasks/system_build.py", line 157, in process
    *self.quintuple_token(add_repo)
  File "/usr/lib/python3.4/site-packages/kiwi/xml_state.py", line 1286, in add_repository
    imageinclude=repo_imageinclude
  File "/usr/lib/python3.4/site-packages/kiwi/xml_parse.py", line 2135, in __init__
    self.priority = _cast(int, priority)
  File "/usr/lib/python3.4/site-packages/kiwi/xml_parse.py", line 682, in _cast
    return typ(value)
ValueError: invalid literal for int() with base 10: ''
```

Note that kiwi-compat creates such --add-repo calls when there is no --add-repopriority used, like with osc.

davidcassany:
xml_parse code is autogenerated, thus no changes there should be
considered. When adding a repository from the command line it
can happen that the repository priority is mapped to an empty instead
to a None valuei, when the priority is not specifically set. xml_parse
will only support a parseable string to int or a None value as the
default option, so, if an empty string is used, it throws an
exception. With the current patch anything that cannot be casted
to an int, will be considered as None.

Fixes #408

### DIFF
--- a/test/unit/xml_state_test.py
+++ b/test/unit/xml_state_test.py
@@ -181,6 +181,15 @@ class TestXMLState(object):
         assert self.state.xml_data.get_repository()[3].get_priority() == 1
         assert self.state.xml_data.get_repository()[3].get_imageinclude() is True
 
+    def test_add_repository_with_empty_values(self):
+        self.state.add_repository('repo', 'type', '', '', True)
+        assert self.state.xml_data.get_repository()[3].get_source().get_path() \
+            == 'repo'
+        assert self.state.xml_data.get_repository()[3].get_type() == 'type'
+        assert self.state.xml_data.get_repository()[3].get_alias() == ''
+        assert self.state.xml_data.get_repository()[3].get_priority() == None 
+        assert self.state.xml_data.get_repository()[3].get_imageinclude() is True
+
     def test_get_to_become_deleted_packages(self):
         assert self.state.get_to_become_deleted_packages() == [
             'kernel-debug'
--- a/test/unit/xml_state_test.py
+++ b/test/unit/xml_state_test.py
@@ -187,7 +187,7 @@ class TestXMLState(object):
             == 'repo'
         assert self.state.xml_data.get_repository()[3].get_type() == 'type'
         assert self.state.xml_data.get_repository()[3].get_alias() == ''
-        assert self.state.xml_data.get_repository()[3].get_priority() == None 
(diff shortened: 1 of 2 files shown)


schaefi:
```
check runtests: commands[1] | flake8 --statistics -j auto --count /home/travis/build/SUSE/kiwi/test/unit
/home/travis/build/SUSE/kiwi/test/unit/xml_state_test.py:190:71: E711 comparison to None should be 'if cond is None:'
/home/travis/build/SUSE/kiwi/test/unit/xml_state_test.py:190:78: W291 trailing whitespace
1     E711 comparison to None should be 'if cond is None:'
1     W291 trailing whitespace
```

test/unit/xml_state_test.py
```diff
@@ -181,6 +181,15 @@ def test_add_repository(self):
         assert self.state.xml_data.get_repository()[3].get_priority() == 1
         assert self.state.xml_data.get_repository()[3].get_imageinclude() is True
 
+    def test_add_repository_with_empty_values(self):
+        self.state.add_repository('repo', 'type', '', '', True)
+        assert self.state.xml_data.get_repository()[3].get_source().get_path() \
+            == 'repo'
+        assert self.state.xml_data.get_repository()[3].get_type() == 'type'
+        assert self.state.xml_data.get_repository()[3].get_alias() == ''
+        assert self.state.xml_data.get_repository()[3].get_priority() == None
```
schaefi:
```python
assert self.state.xml_data.get_repository()[3].get_priority() is None
```

should fix the test

davidcassany: sure, it should be fixed now, I forgot to verify style with `tox -e check` before making the PR.

schaefi:
👍

PR #415, opened 2017-07-05
Make sure system image strip sections are applied in boot image

### LINKED ISSUE #414: strip section in the image description file is not applied to the boot image
## Problem description

Trying to prevent some file removal in the boot image through the system image `config.xml` does not work as exposed in [boot image documentation section](https://suse.github.io/kiwi/overview/workflow.html#faq-boot-image-customization).

## Actual behaviour

Including a section like the one below does not make any effect in the boot image:

```xml
<strip type="tools">
  <file name="lsblk"/>
</strip>
```

Build a leap 42.2 JeOS with `boot="oemboot/suse-leap42.2"` and try to keep **lsblk** tool with the strip block provided above. The lsblk tool will be deleted anyways.

davidcassany:
This commit calls copy_strip_sections from the system image xml to the
boot image xml. This is needed to make sure strip sections are
applied into the boot image.

Fixes #414

### DIFF
--- a/kiwi/boot/image/base.py
+++ b/kiwi/boot/image/base.py
@@ -211,6 +211,9 @@ class BootImageBase(object):
         strip_xml_state.copy_strip_sections(
             self.boot_xml_state
         )
+        self.xml_state.copy_strip_sections(
+            self.boot_xml_state
+        )
         preferences_subsection_names = [
             'bootloader_theme',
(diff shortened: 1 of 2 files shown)


schaefi:
Thanks, looks like this was forgotten at the time when I touched this code :) 👍

PR #416, opened 2017-07-05
Refactor handling of obs repositories

schaefi:
* Delete kiwi internal ibs: and suse: types
* Delete handling for --obs-repo-internal and provide a
  compatibility message to the user
* Buildservice download server url and scope can be configured
  via ~/.config/kiwi/config.yml
* Translate obs urls to http in import_repositories_marked_as_imageinclude
* Use new Uri.is_public method in renamed runtime check
  check_image_include_repos_publicly_resolvable

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -70,10 +70,18 @@ class Defaults(object):
         ]
 
     @classmethod
-    def is_obs_worker(self):
+    def is_buildservice_worker(self):
         # the presence of /.buildenv on the build host indicates
         # we are building inside of the open buildservice
         return os.path.exists('/.buildenv')
+
+    @classmethod
+    def get_obs_download_server_url(self):
+        """
+        The default download server url hosting the public open
+        buildservice repositories
+        """
+        return 'http://download.opensuse.org'
 
     @classmethod
     def get_s390_disk_block_size(self):
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -81,7 +81,7 @@ class Defaults(object):
         The default download server url hosting the public open
         buildservice repositories
         """
-        return 'http://download.opensuse.org'
+        return 'http://download.opensuse.org/repositories'
 
     @classmethod
     def get_s390_disk_block_size(self):
--- a/kiwi/system/uri.py
+++ b/kiwi/system/uri.py
@@ -18,6 +18,7 @@
 import os
 from tempfile import mkdtemp
 from six.moves.urllib.parse import urlparse
+import requests
 import hashlib
 
 # project
@@ -28,7 +29,8 @@ from kiwi.runtime_config import RuntimeConfig
 
 from kiwi.exceptions import (
     KiwiUriStyleUnknown,
-    KiwiUriTypeUnknown
+    KiwiUriTypeUnknown,
+    KiwiUriOpenError
 )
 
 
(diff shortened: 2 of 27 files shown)


kiwi/system/uri.py
```diff
@@ -91,38 +92,41 @@ def translate(self):
...
         elif uri.scheme == 'iso':
             return self._iso_mount_path(uri.path)
-        elif uri.scheme == 'suse':
-            return self._suse_buildservice_path(
-                name=''.join([uri.netloc, uri.path]),
-                fragment=uri.fragment,
-                urischeme=uri.scheme
-            )
-        elif uri.scheme == 'http' or uri.scheme == 'https' or uri.scheme == 'ftp':
+        elif uri.scheme.startswith('http') or uri.scheme == 'ftp':
```
davidcassany: What happened to https? Why is not taken into consideration? Now any https URL will throw an exception, isn't it? Maybe I miss something...

kiwi/defaults.py
```diff
@@ -70,12 +70,20 @@ def get_xz_compression_options(self):
...
         return os.path.exists('/.buildenv')
 
     @classmethod
+    def get_obs_download_server_url(self):
+        """
+        The default download server url hosting the public open
+        buildservice repositories
+        """
+        return 'http://download.opensuse.org'
+
```
davidcassany: I believe we could include the `repositories` part in the URL. Well, this is a matter of coordinating with he buildservice guys, but form Adrian example in email thread, I would assume they give the base URL including `/repositories`. I would like to stick as much as possible to the data provided by the buildservice, without having to parse it and just use it as is.

```
# osc api /configuration  | grep download_url
<download_url>http://download.opensuse.org/repositories</download_url>
```

kiwi/system/uri.py
```diff
@@ -202,24 +221,28 @@ def _local_path(self, path):
...
-        return ''.join(['http://download.opensuse.org/repositories/', name])
-
-    def _ibs_project(self, name):
-        ibs_project = 'http://download.suse.de/ibs/'
-        return ibs_project + name.replace(':', ':/')
+        return ''.join(
+            [
+                self.runtime_config.get_obs_download_server_url(),
+                '/repositories/', name
+            ]
```
davidcassany: Not sure we have to include repositories here, related to my comment above in defaults.py

kiwi/system/uri.py
```diff
@@ -91,38 +92,41 @@ def translate(self):
...
-                ''.join([uri.netloc, uri.path])
-            )
+            if check_build_environment and Defaults.is_buildservice_worker():
+                return self._buildservice_path(
+                    name=''.join([uri.netloc, uri.path]),
+                    fragment=uri.fragment,
+                    urischeme=uri.scheme
+                )
+            else:
+                if self.repo_type == 'yast2':
```
davidcassany: As discussed in IRC I doubt we should consider this case... specially inside the buildservice, inside the buildservice I would just use the provided <download_url> from the buildservice, any case that can't be handled with that parameter I would just not consider it.

schaefi: yes that will be dropped, still working on the code

kiwi/system/uri.py
schaefi: starts with http: covers http and https, all tests are still passing, well that would also cover httpx :)

kiwi/system/uri.py
schaefi: yes this part will also be replaced

kiwi/defaults.py
schaefi: ok I agree and take that into consideration, thanks

kiwi/system/uri.py
davidcassany: Oh damm, you are right, of course I missed the relevant part 😓

schaefi:
I'm happy to get rid of some of the weird stuff :)

kiwi/system/uri.py
```diff
@@ -66,18 +71,16 @@ def __init__(self, uri, repo_type=None):
...
             'dir': True,
             'file': True,
-            'suse': True,
             'obsrepositories': True
         }
 
-    def translate(self):
+    def translate(self, check_build_environment=True):
         """
         Translate repository location according to their URI type
```
davidcassany: nit-picking, but I would update the docs to include check_build_environment parameter explanation

kiwi/system/uri.py
```diff
@@ -201,25 +217,26 @@ def _iso_mount_path(self, path):
...
+        project = os.sep.join(name_parts)
+        try:
+            download_link = os.sep.join(
+                [
+                    self.runtime_config.get_obs_download_server_url(),
+                    project.replace(':', ':/'), repository
+                ]
+            )
+            request = requests.get(download_link)
+            request.raise_for_status()
```
davidcassany: 👍 nice, url gets verified right during the translation, I like it.

kiwi/system/uri.py
schaefi: oh yes, forgot that... stay tuned I'll update

schaefi: done now

kiwi/system/uri.py
schaefi: yes a nice side effect of the cleanup for the obs type

PR #417, opened 2017-07-05
Move man pages to correct section

Conan-Kudo:
The man pages were incorrectly written to section 2, which is for syscalls. Since KIWI is an administrator's tool, it has been moved to section 8.

* Move man pages from section 2 to section 8


schaefi:
Thanks looks good, I just found one place which needs to be adapted too. Could you do the change in kiwi/help.py as well ? Thanks

Conan-Kudo:
@schaefi Fixed.

PR #419, opened 2017-07-09
Fixed Uri is_remote method

schaefi:
If called inside of the buildservice the obs uri type is not
a remote uri because the translation ends in a local path

### DIFF
--- a/kiwi/system/uri.py
+++ b/kiwi/system/uri.py
@@ -196,13 +196,18 @@ class Uri(object):
 
         :rtype: bool
         """
-        if self.is_remote():
-            uri = urlparse(self.uri)
-            if uri.scheme == 'obs' and not self.runtime_config.is_obs_public():
-                return False
-            else:
-                return True
+        uri = urlparse(self.uri)
+        if not uri.scheme:
+            # unknown uri schema is considered not public
+            return False
+        elif uri.scheme == 'obs':
+            # obs is public but only if the configured download_server is public
+            return self.runtime_config.is_obs_public()
+        elif uri.scheme in self.remote_uri_types:
+            # listed in remote uri types, thus public
+            return True
         else:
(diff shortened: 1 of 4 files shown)


schaefi:
David I found the problem when building in the buildservice. The change for the obs type to be both a local and a remote uri type depending on the environment was not applied to the is_remote() method. This caused the obs type to be treated as a remote type when building in obs and that caused the local repos to be not bind mounted into the chroot, which caused the error message for zypper saying it can't find the repos when called chrooted

Please double check the changes and let us check this is asap, because all obs builds will currently fail

Thanks

kiwi/system/uri.py
```diff
@@ -179,15 +179,16 @@ def is_remote(self):
             raise KiwiUriStyleUnknown(
                 'URI scheme not detected %s' % self.uri
             )
-        if uri.scheme in self.remote_uri_types:
+        if uri.scheme == 'obs' and Defaults.is_buildservice_worker():
+            return False
```
davidcassany: This change is applied always, inside and outside of the buildservice, I am not sure it will behave as we want in is_public method, as any obs URI in the buildserive will not be considered as a plublic address. That's why I initially thought to include a `check_build_environment` parameter in that method. 

In any case this is a quite specific case we can verify and polish after applying this patch so, so current builds turn back to normality.

schaefi: right thanks for catching the is_public case. I will adapt those method too

schaefi: ok done, how about it now ?

davidcassany: Great! I like that solution, I think that now everything should work fine, I'll stay tuned, hopefully there won't be further surprises.

schaefi:
ok I'll merge and resubmit packages

keep fingers crossed that all obs builds will magically heal :)

PR #420, opened 2017-07-11
Fix url validation in buildservice

### LINKED ISSUE #418: Remote URL verification should not ever happen inside the buildservice
## Problem description

KIWI fails when trying to verify a translated obs:// URI inside the buildservice. By design, buildservice does not provide internet connection during the build, hence this test will always fail.

## Actual behaviour

KIWI does the appropriate URL translation but it should not try to validate it when building inside the buildservice.

A repository configuration like the following fails inside the public suse buildservice:

```xml
<repository type="rpm-md" imageonly="true">
  <source path="obs://openSUSE:Factory/snapshot"/>
</repository>
```

See the image [here](https://build.opensuse.org/package/show/home:dcassany:experimental_images:repositories_tests/obsrepositories_style)

davidcassany:
This patch changes the remote URL validation strategy when running
inside the build service. By design, inside the build service build
environment connections to the outside world are not allowed, thus
any validation attempting to do that will fail. With this patch, when
running inside the build service, KIWI will not try to test if any
download URL is reachable.

Fixes #418

### DIFF
--- a/test/unit/system_uri_test.py
+++ b/test/unit/system_uri_test.py
@@ -61,6 +61,27 @@ class TestUri(object):
         uri.runtime_config = self.runtime_config
         assert uri.translate()
 
+    @raises(KiwiUriOpenError)
+    @patch('kiwi.system.uri.Defaults.is_buildservice_worker')
+    def test_translate_obs_uri_in_private_buildservice(
+            mock_buildservice, self
+    ):
+        mock_buildservice.return_value = True
+        self.runtime_config.is_obs_public = mock.Mock(
+            return_value=False
+        )
+        uri = Uri('obs://openSUSE:Leap:42.2/standard', 'rpm-md')
+        uri.runtime_config = self.runtime_config
+        uri.translate(False)
+
+    @patch('kiwi.system.uri.Defaults.is_buildservice_worker')
+    def test_translate_obs_uri_inside_buildservice(self, mock_buildservice):
+        mock_buildservice.return_value = True
+        uri = Uri('obs://openSUSE:Leap:42.2/standard', 'rpm-md')
+        uri.runtime_config = self.runtime_config
+        assert uri.translate(False) == \
+            'obs_server/openSUSE:/Leap:/42.2/standard'
+
     def test_is_remote(self):
         uri = Uri('https://example.com', 'rpm-md')
         assert uri.is_remote() is True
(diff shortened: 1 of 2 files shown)


test/unit/system_uri_test.py
```diff
@@ -61,6 +61,30 @@ def test_translate_obs_uri_not_found(mock_request_get, self):
...
+    def test_translate_obs_uri_in_private_buildservice(
+            mock_buildservice, self
+    ):
+        mock_buildservice.return_value = True
+        self.runtime_config.is_obs_public = mock.Mock(
+            return_value=False
+        )
+        self.runtime_config.get_obs_download_server_url = mock.Mock(
+            return_value='obs_server'
+        )
```
davidcassany: @schaefi, I can't understand why I need to re-run the above lines, already lost half morning with it, those lines are redundant (to my understanding) but they do not hurt. If you have a clue on why are they needed I'd be very happy. Time to time MagicMock library just beats me...

schaefi: Hmm, I deleted the line and the test still works

davidcassany: The test works, but you'll see that the coverage is not complete, thus it does not perform as expected. An exception is thrown, but not the one that is supposed to happen.

schaefi: ok but coverage said the code was not reached, maybe because you are raising KiwiUriOpenError in a try block which raises again a KiwiUriOpenError... ah this is confusing

why do we check the public state ? I think it would be enough to log a warning that we can't verify the location. imho it would be possible somebody wants to add an obs repo as imageonly but build in the buildservice

davidcassany: yes this is confusing... you got a point, better to just log a warning, I like the idea. I'll code it and lets see if the test behaves as expected then.

schaefi:
```python
if not Defaults.is_buildservice_worker():
      ...
else:
     log.warning('Using {0} without location verification due to build in isolated environment'.format(download_link))
     return download_link
```

something like this would be imho enough, thoughts ?

davidcassany:
@schaefi  right, now looks way better, thanks for your feedback. If you approve the PR I'll merge and squash all commits in one.

schaefi:
yes man this is great work 👍

PR #422, opened 2017-07-13
Include .buildenv file inside the buildroot

### LINKED ISSUE #421: Include /.buildenv file inside the buildroot when building inside the buildservice
## Problem description

In current KIWI versions when building inside the buildservice the build root is not populated witht he /.buildenv file. This specific file is provided by the buildservice and some packages installation scripts behave differently depending on the presence or not of this file. This results in a different installation of some packages (in concrete detected with plymouth-branding-openSUSE) when comparing the builds done with former kiwi versions compared with the current kiwi-ng.

## Actual behaviour

One example of this differences is the initrd image that gets created during the installation of plymouth-branding-openSUSE package. This initrd is created only if we are not running inside of the buildservice, that produces a difference of around 20MB when comparing the image size of the some ISO builds using former kiwi and current kiwi-ng.

Create a live ISO and extract the resulting squashfs to verify that in includes a `/boot/initrd-xxxxx-default` file, which, in this concrete context, is just useless.

davidcassany:
With this patch if KIWI is running inside the buildservice the
/.buildenv file is copied from the build environment to the build
root of KIWI. At installation time some packages test the existence
of this file in order to know if they are being installed inside the
buildservice or not.

Fixes #421

### DIFF
--- a/kiwi/system/root_init.py
+++ b/kiwi/system/root_init.py
@@ -19,6 +19,7 @@ import stat
 from pwd import getpwnam
 from tempfile import mkdtemp
 from shutil import rmtree
+from shutil import copy
 import os
 
 # project
@@ -84,9 +85,8 @@ class RootInit(object):
             data.sync_data(
                 options=['-a', '--ignore-existing']
             )
-            Command.run([
-                'touch', os.sep.join([self.root_dir, '.buildenv'])
-            ])
+            if Defaults.is_buildservice_worker():
+                copy(Defaults.get_buildservice_env_name(), self.root_dir)
         except Exception as e:
             self.delete()
             raise KiwiRootInitCreationError(
--- a/kiwi/system/root_init.py
+++ b/kiwi/system/root_init.py
@@ -19,7 +19,6 @@ import stat
 from pwd import getpwnam
 from tempfile import mkdtemp
 from shutil import rmtree
-from shutil import copy
 import os
 
 # project
@@ -85,8 +84,9 @@ class RootInit(object):
             data.sync_data(
                 options=['-a', '--ignore-existing']
             )
-            if Defaults.is_buildservice_worker():
-                copy('/.buildenv', self.root_dir)
+            Command.run([
+                'touch', os.sep.join([self.root_dir, '.buildenv'])
+            ])
         except Exception as e:
             self.delete()
             raise KiwiRootInitCreationError(
--- a/kiwi/system/root_init.py
+++ b/kiwi/system/root_init.py
@@ -19,6 +19,7 @@ import stat
 from pwd import getpwnam
 from tempfile import mkdtemp
 from shutil import rmtree
+from shutil import copy
 import os
 
 # project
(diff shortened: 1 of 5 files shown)


davidcassany:
@DimStar77 I did a build of your x11 Rescue CD [here](https://build.opensuse.org/package/show/home:dcassany:branches:openSUSE:Factory:Live/kiwi-image-livecd-x11) including the current patch and the image builds and it is actually significantly smaller than the one done with the legacy kiwi version 😄 This was expected since the boot image of the current kiwi version is smaller. Hopefully the image is fine, I just built it and booted it, I have no idea how to further test it.

DimStar77:
@davidcassany smaller is GREAT news!  I could live with 'equally big' too, but will gladly buy you a bear over 'smaller' :P

kiwi/system/root_init.py
```diff
@@ -84,6 +85,8 @@ def create(self):
             data.sync_data(
                 options=['-a', '--ignore-existing']
             )
+            if Defaults.is_buildservice_worker():
+                copy('/.buildenv', self.root_dir)
```
schaefi: This assumes a buildservice worker has to provide /.buildenv which is most probably always true, but if not we will fail here which might be intentional and the correct reaction, however in the legacy kiwi we called

```
touch $root/.buildenv
```

which would never fail.

davidcassany: Yes I already thought about that, it is a matter of understanding what is .buildenv doing, I am not sure if .buildenv is something specific to the buildservice and its internal needs (this is what I initially thought and the main reason to limit the change only running inside the buildservice). On the other hand this file/flag is used by packages to figure out if they are being installed in some kind of chroot (or similar build/devel environment), in which case then KIWI should always make use of it.

davidcassany: Anyway, even if it is just to be consistent with builds inside and outside of the buildservice, you got a point in the sense KIWI should always make use of it, at least, this way all images will be the same, regardless having buildservice involved or not.

schaefi: Thinking twice about this, your former implementation was totally correct , sorry for the confusion. The detection if we are a bs worker is based on the presence of .buildenv, thus the copy of that file is safe and makes complete sense.

So hope you don't mind but I think we should revert the last commit

Once done I only have only issue. With the copy statement in place we would have two places in the code which deals with the .buildenv file name. I suggest we create a method in the defaults area e.g:

```
@classmethod
def get_buildservice_env_name(self):
    """
    The base name of the environment file in a buildservice worker
    """
    return '.buildenv'
```

and use that in the is_buildservice_worker() method and in the copy() call

Thoughts ?

schaefi: sorry again for making you performing an unneeded change

davidcassany: Sure @schaefi, no problem, I can revert it.

About the get_buildservice_env_name I like the idea, something similar I have on the way for other kiwi files in #423 (I did not make the PR yet because this one goes first).

schaefi: ok so I'll wait for the update on that request, thanks

schaefi:
👍 this is great now, thanks

bugfinder:
actually this breaks systems deployed from a kiwi image, since the next kernel update will not create an initrd if /.buildenv exists (see /usr/lib/module-init-tools/weak-modules2)

PR #425, opened 2017-07-14
Cleanup use of python interpreter invocation

### LINKED ISSUE #424: Kiwi - Makefile uses call to python3
## Problem description

The global Makefile uses explicit calls to python3. This prevents building a python2 only version for older distros.

## Actual behaviour

python2 setup.py build --cflags="${RPM_OPT_FLAGS}" 

in the spec file calls the Makefile which calls python3

Look into the git repo, at Makefile.

schaefi:
Prevent strict call of a specific version of the python
interpreter. All code has been written to work with py2
and py3 thus the venv environment setup should decide
what version a call of python is. Fixes #424

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,13 @@
 buildroot = /
 python_version = 3
+python_lookup_name = python
+python = $(shell which $(python_lookup_name))
 
 LC = LC_MESSAGES
 
 version := $(shell \
-    python -c 'from kiwi.version import __version__; print(__version__)'\
+    $(python) -c \
+    'from kiwi.version import __version__; print(__version__)'\
 )
 
 .PHONY: test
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -4,7 +4,7 @@ python_version = 3
 LC = LC_MESSAGES
 
 version := $(shell \
-    python3 -c 'from kiwi.version import __version__; print(__version__)'\
+    python -c 'from kiwi.version import __version__; print(__version__)'\
```
Conan-Kudo: Please use a variable for the python program so that it can be easily substituted at `make` invocation.

See https://github.com/livecd-tools/livecd-tools/blob/master/Makefile for an example of how I do it.

PR #428, opened 2017-07-16
fixed: SUSE / Fedora / CentOS / RHEL package dependencies and proper %if

mmohring:

* Fix Buildtime and Runtime dependencies for CentOS / RHEL / Fedora
* Allow to generate OBS packages for Fedora / CentOS / RHEL with correct runtime dependencies
* Generates python 2 only package on CentOS / RHEL which have no python3
* Kiwi package is then able to be run in OBS for Fedora / RHEL / CentOS targets

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -76,8 +76,6 @@ Name:           python-kiwi
 Version:        %%VERSION
 Release:        0
 Provides:       kiwi-schema = 6.5
-Provides:       kiwi
-Obsoletes:      kiwi <= 9.0
 Url:            https://github.com/SUSE/kiwi
 Summary:        KIWI - Appliance Builder Next Generation
 License:        GPL-3.0+
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -1,7 +1,7 @@
 #
-# spec file for package kiwi
+# spec file for package python-kiwi
 #
-# Copyright (c) 2016 SUSE LINUX GmbH, Nuernberg, Germany.
+# Copyright (c) 2017 SUSE LINUX GmbH, Nuernberg, Germany.
 #
 # All modifications and additions to the file contributed by third parties
 # remain the property of their copyright owners, unless otherwise agreed
@@ -11,11 +11,11 @@
 # case the license is the MIT License). An "Open Source License" is a
 # license that conforms to the Open Source Definition (Version 1.9)
 # published by the Open Source Initiative.
-#
 # Please submit bugfixes or comments via:
 #
 #       https://github.com/SUSE/kiwi/issues
 #
+
 %{!?python2_sitelib:%global python2_sitelib %{python_sitelib}}
 
 # translate version id to distribution name as it is used in kiwi
@@ -66,11 +66,18 @@
 %define distro rhel-07.0
 %endif
 
+# Fedora
+# use the rhel templates for CentOS, too
+%if 0%{?fedora_version}
+%define distro fedora-25.0
+%endif
 
 Name:           python-kiwi
 Version:        %%VERSION
-Provides:       kiwi-schema = 6.5
 Release:        0
+Provides:       kiwi-schema = 6.5
+Provides:       kiwi
+Obsoletes:      kiwi <= 9.0
 Url:            https://github.com/SUSE/kiwi
 Summary:        KIWI - Appliance Builder Next Generation
 License:        GPL-3.0+
@@ -79,13 +86,20 @@ Source:         %{name}.tar.gz
 Source1:        %{name}-boot-packages
 Source2:        %{name}-rpmlintrc
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
-BuildRequires:  python3-devel
-BuildRequires:  python3-setuptools
 BuildRequires:  python-devel
 BuildRequires:  python-setuptools
+%if 0%{?fedora_version} || 0%{?suse_version}
+BuildRequires:  python3-devel
+BuildRequires:  python3-setuptools
+%endif
+%if 0%{?suse_version}
 BuildRequires:  fdupes
-BuildRequires:  update-alternatives
 BuildRequires:  shadow
+BuildRequires:  update-alternatives
+%endif
+%if 0%{?fedora_version} || 0%{?rhel_version} || 0%{?centos_version}
+BuildRequires:  chkconfig
+%endif
 
 %description
 The KIWI Image System provides an operating system image builder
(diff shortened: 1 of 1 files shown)


mmohring:
This should fix the package .spec and together with #425 also the python2 / python3 Makefile issues.

package/python-kiwi-spec-template
```diff
@@ -231,14 +278,14 @@ outside of the scope of kiwi appliance building.
 %ifarch %ix86 x86_64
 %package -n kiwi-pxeboot
 Summary:        KIWI - PXE boot structure
+License:        GPL-3.0+
+Group:          System/Management
```
Conan-Kudo: These tags do not need to be repeated from the main header, as they are inherited during the package build.

package/python-kiwi-spec-template
```diff
@@ -66,11 +66,18 @@
 %define distro rhel-07.0
 %endif
 
+# Fedora
+# use the rhel templates for CentOS, too
+%if 0%{?fedora_version}
```
Conan-Kudo: Please use `%fedora` instead of `%fedora_version`.

package/python-kiwi-spec-template
```diff
@@ -66,11 +66,18 @@
...
+%define distro fedora-25.0
```
Conan-Kudo: Use `fedora-%{fedora}.0` to ensure the version is selected dynamically.

package/python-kiwi-spec-template
```diff
@@ -79,13 +86,20 @@ Source:         %{name}.tar.gz
...
+BuildRequires:  python3-devel
+BuildRequires:  python3-setuptools
+%endif
+%if 0%{?suse_version}
 BuildRequires:  fdupes
-BuildRequires:  update-alternatives
 BuildRequires:  shadow
+BuildRequires:  update-alternatives
+%endif
+%if 0%{?fedora_version} || 0%{?rhel_version} || 0%{?centos_version}
```
Conan-Kudo: Again, please use `%{fedora}`.

In addition, `%rhel_version` and `%centos_version` can be combined into a single `%{rhel}` check.

package/python-kiwi-spec-template
```diff
@@ -95,57 +109,74 @@ and cloud systems like Xen, KVM, VMware, EC2 and more.
 # python2-kiwi
 %package -n python2-kiwi
 Summary:        KIWI - Appliance Builder Next Generation
+License:        GPL-3.0+
 Group:          Development/Languages/Python
-Provides:       python-kiwi = %{version}-%{release}
-Provides:       kiwi-image:tbz
+Provides:       kiwi
```
Conan-Kudo: Please don't do this, as this makes it difficult to have only one version installed when someone wants to install `kiwi`.

Also, I'm fairly certain that SUSE doesn't want to do this yet, so you may need to undo this change entirely.

package/python-kiwi-spec-template
```diff
@@ -95,57 +109,74 @@ and cloud systems like Xen, KVM, VMware, EC2 and more.
...
+Requires(post): update-alternatives
+Requires(postun): update-alternatives
+%ifarch x86_64
+Requires:       grub2-x86_64-efi
+%endif
+Requires:       qemu-tools
+Requires:       multipath-tools
 Requires:       squashfs
+Requires:       gptfdisk
+Requires:       zypper
```
Conan-Kudo: Zypper is only available in Fedora 26+ and SUSE distributions.

package/python-kiwi-spec-template
```diff
@@ -155,59 +186,73 @@ Python 2 library of the KIWI Image System. Provides an operating system
...
+Requires:         chkconfig
+Requires(post):   chkconfig
+Requires(postun): chkconfig
+Requires:       qemu-img
 Requires:       squashfs-tools
+Requires:       device-mapper-multipath
+Requires:       gdisk
+Requires:       yum
 Provides:       kiwi-packagemanager:yum
+%if 0%{?fedora_version}
```
Conan-Kudo: This will also be in `%{?rhel} >= 8`.

package/python-kiwi-spec-template
mmohring: Currently, there are only tags for Fedora 25 in the python-kiwi-boot-packages so I use them also for Fedora 26.

package/python-kiwi-spec-template
mmohring: removed this.

package/python-kiwi-spec-template
mmohring: So I use it only for SUSE. Whats wrong ?

package/python-kiwi-spec-template
mmohring: Is this speculation or a fact that RHEL will switch to dnf for version 8 ?

package/python-kiwi-spec-template
mmohring: Removed all redundant License: tags.

package/python-kiwi-spec-template
Conan-Kudo: It's not wrapped in the appropriate conditional.

If needs `%if 0%{?fedora} >= 26 || 0%{?suse_version}`

package/python-kiwi-spec-template
```diff
@@ -96,56 +108,72 @@ and cloud systems like Xen, KVM, VMware, EC2 and more.
...
+Requires:         chkconfig
+Requires(post):   chkconfig
+Requires(postun): chkconfig
+Requires:       qemu-img
 Requires:       squashfs-tools
+Requires:       device-mapper-multipath
+Requires:       gdisk
+Requires:       yum
 Provides:       kiwi-packagemanager:yum
+%if 0%{?fedora} || 0%{?rhel} > 800
```
Conan-Kudo: It is only  `8`, not `800`.

package/python-kiwi-spec-template
```diff
@@ -155,59 +183,72 @@ Python 2 library of the KIWI Image System. Provides an operating system
...
+Requires:         chkconfig
+Requires(post):   chkconfig
+Requires(postun): chkconfig
+Requires:       qemu-img
 Requires:       squashfs-tools
+Requires:       device-mapper-multipath
+Requires:       gdisk
+Requires:       yum
 Provides:       kiwi-packagemanager:yum
+%if 0%{?fedora} || 0%{?rhel} >= 800
```
Conan-Kudo: It's only `8`, not `800`.

mmohring:
@Conan-Kudo I have now replaced 0{?fedora_version} with 0{?fedora} and 0{?rhel_version} || 0{?centos_version} with 0{?rhel} and put also some 0{?rhel} >= 800 in place for dnf, maybe they use dnf in RHEL 8.

package/python-kiwi-spec-template
```diff
@@ -66,11 +66,16 @@
 %define distro rhel-07.0
 %endif
 
+# Fedora
+# use the rhel templates for CentOS, too
+%if 0%{?fedora_version} >= 25
```
Conan-Kudo: Use `0%{?fedora} >= 25` instead.

package/python-kiwi-spec-template
```diff
@@ -79,13 +84,20 @@ Source:         %{name}.tar.gz
 Source1:        %{name}-boot-packages
 Source2:        %{name}-rpmlintrc
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
-BuildRequires:  python3-devel
-BuildRequires:  python3-setuptools
 BuildRequires:  python-devel
 BuildRequires:  python-setuptools
+%if 0%{?fedora_version} || 0%{?suse_version}
```
Conan-Kudo: `0%{?fedora}` here, please.

package/python-kiwi-spec-template
mmohring: Please take a look into the expanded file, the diff fools you. It is wrapped. But what you gonna tell me is that you can use zypper with Fedora 26 ?

Conan-Kudo: Yes, you can. https://admin.fedoraproject.org/pkgdb/package/rpms/zypper/

mmohring: I fixed it to support zypper also on Fedora >= 26 now.

package/python-kiwi-spec-template
```diff
@@ -96,56 +108,74 @@ and cloud systems like Xen, KVM, VMware, EC2 and more.
 %package -n python2-kiwi
 Summary:        KIWI - Appliance Builder Next Generation
 Group:          Development/Languages/Python
-Provides:       python-kiwi = %{version}-%{release}
-Provides:       kiwi-image:tbz
+Provides:       kiwi
+Obsoletes:      kiwi <= 9.0
```
Conan-Kudo: The `Obsoletes` and `Provides` for `kiwi` need to be removed.

package/python-kiwi-spec-template
```diff
@@ -155,59 +185,74 @@ Python 2 library of the KIWI Image System. Provides an operating system
...
 virtualization and cloud systems like Xen, KVM, VMware, EC2 and more.
 
+%if 0%{?fedora} || 0%{?suse_version}
 # python3-kiwi
 %package -n python3-kiwi
 Summary:        KIWI - Appliance Builder Next Generation
 Group:          Development/Languages/Python
-Provides:       kiwi-image:tbz
+Provides:       kiwi
+Obsoletes:      kiwi <= 9.0
```
Conan-Kudo: The `Obsoletes` and `Provides` for `kiwi` need to be removed.

package/python-kiwi-spec-template
mmohring: done all 700 / 800 => 7 / 8, and removed all centos_version now either.

package/python-kiwi-spec-template
```diff
@@ -231,14 +273,13 @@ outside of the scope of kiwi appliance building.
 %ifarch %ix86 x86_64
 %package -n kiwi-pxeboot
 Summary:        KIWI - PXE boot structure
+Group:          System/Management
```
Conan-Kudo: You don't need this, as you get it from the main source package. Same goes for `License` in all subpackages.

mmohring: I had copied this from the old kiwi package, which has different Group entries for the main package and some of the subpackages.

Conan-Kudo: If they have different Groups, then it's fine.

package/python-kiwi-spec-template
```diff
@@ -96,56 +108,72 @@ and cloud systems like Xen, KVM, VMware, EC2 and more.
...
 Provides:       kiwi-image:oem
+Provides:       kiwi-image:pxe
+Provides:       kiwi-image:tbz
+Provides:       kiwi-image:vmx
+Provides:       python-kiwi = %{version}-%{release}
+%if 0%{?fedora} || 0%{?suse_version}
 Recommends:     jing
-Requires:       update-alternatives
+%else
+Requires:       jing
```
schaefi: no requires of jing please, that would mean we require the entire java stack. This should stay as Recommends

mmohring: How do I make a Recommends with old rhel or centos 7 ?

mmohring:
@schaefi so you want me to redo the patch completely with 3 or 4 simple commits ?

package/python-kiwi-spec-template
schaefi: yes please skip it then. if we run into a validation problem and there is no jing on the system kiwi provides a message to the user that more detailed errors can be obtained if jing would have been installed. imho that's enough information to let a user know what to do. It's a nightmare to require a java based software when it is only an optional thing :) Thanks

schaefi:
I would prefer separate pull requests and handle one thing at a time yes, this also motivates more people to look at it and make it much easier for us to not overlook something.

I'm sorry if this is causing any inconvenience for you. All your changes makes sense and you can be sure we will merge them in, but please let us do it in a more controlled way. Thanks

package/python-kiwi-spec-template
mmohring: OK, I do it. I will now work on 2 or 3 separate pull requests for this patch.

mmohring:
OK, I will do a new version of this patch in separate pull requests.

schaefi:
Thanks much 👍

mmohring:
@schaefi I have now created a simplified patch with only adding the CentOS / RHEL / Fedora support. Its pull request #439 .

PR #430, opened 2017-07-17
This commit fixed root_init when running inside the buildservice

davidcassany:
This commit fixes regression bug introduced in #422


davidcassany:
@schaefi I did something stupid in #422 and the copy statement in root init fails inside the buildservice because I forgot to include the root slash (`/.buildenv`). With this ugly bug all builds in the buildservice fail.

schaefi:
*urks* ok thanks for the quick turnaround

PR #432, opened 2017-07-17
Exclude kiwi files from images

### LINKED ISSUE #423: KIWI specific internal data gets into the image (.profile, image folder, etc.)
## Problem description

In some types of images the KIWI specific files are not excluded to be part of the image. In concrete a live ISO includes the image folder together with some other KIWI specific files.

## Actual behaviour

The method `_get_exclude_list_for_root_data_sync` which is used to make sure some specific data does not get into the image is only applied for disks builds. There is no equivalent method of other builds
like live, archive, etc.

Build a live ISO image and check the squashfs, after mounting it is possible to see the image folder inside the root tree.

davidcassany:
This commit from one hand includes a
get_exclude_list_for_root_data_sync method in Defaults which returns
a list of the files used by KIWI that should not be part of the
resulting image. From the other hand makes use of the exclusion
default list in live, archive and container images, it fixes #423.

### DIFF
--- a/kiwi/builder/archive.py
+++ b/kiwi/builder/archive.py
@@ -85,7 +85,8 @@ class ArchiveBuilder(object):
                 self._target_file_for('tar')
             )
             archive.create_xz_compressed(
-                self.root_dir, xz_options=self.xz_options
+                self.root_dir, xz_options=self.xz_options,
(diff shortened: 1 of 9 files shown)


schaefi:
looks good to me 👍

PR #437, opened 2017-07-20
Include default container name

davidcassany:
This commit includes a default container name for KIWI container
images. This makes possible to create OCI and Docker containers
without forcing the user to include a `<contaierconfig>` section in
the description file.


schaefi:
That's great 👍

Issue #429, opened 2017-07-17
Refactor the way Xen images are build


schaefi:
## Problem description

With the move to the pvops kernel the distributions do no longer distinguish between a kernel for para virtualization and a kernel for hvm (full) virtualization. The used kernel however was a trigger in the legacy kiwi version to know about the target image and its configurations in terms of bootloader, modules etc... Therefore the XML schema to provide this information was aligned to this information and provided in the machine section like the following example shows:

```xml
<machine memory="512" domain="domU-PV"
    <vmdisk id="0" device="/dev/xvda" controller="ide"/>
    <vmnic interface=""/>
</machine>
```

This schema layout is still present in the next generation kiwi code base and needs to improve at
several places:

1. The domain setup is at the wrong place, this does not belong to the machine section and has to be removed
2. Allow to classify an image as xen_server=true|false in the <type> section. This will replace the old domain setup because all we should provide in the image description is if this is xen server or xen guest image
3. Allow to specify the target loader for a xen guest (hvmloader, pvgrub, pygrub, ...) in the machine section. This information would be relevant when kiwi should create the machine configuration file. It also impacts the bootloader configuration inside of the image because depending on the guest loader additional information needs to be present

An example Xen server setup could look like:

```xml
<type image="oem" .... xen_server="true"/>
```

An example Xen guest setup could look like:

```xml
<type image="vmx"/>

<machine memory="512" xen_loader="pvgrub">
    <vmdisk id="0" device="/dev/xvda" controller="ide"/>
</machine>
```

In addition to the schema changes there should also be at least one runtime check which which checks:

* image classified as xen_server plus machine section would be a mistake, you can be either a dom0 or a guest with a machine setup but not both at the same time

schaefi:
Still to be done:

* runtime checks

schaefi:
Still to be done:

* runtime checks

PR #438, opened 2017-07-20
Refactor xen build

schaefi:
First start to refactor the way Xen images are configured in the XML description
The changes here references issue #429 

* Update to schema 6.6 with auto conversion via XSLT stylesheet
* Deleted domain attribute from machine section
* Added xen_server attribute to specify a type to be a Xen dom0
* Added xen_loader attribute in machine section to specify the target guest loader this image is expected to become loaded with. There is currently no code which produces a different bootloader setup depending on the value of xen_loader. But there is now a way for us to actually handle it when needed
* Refactor code which deals with Xen to be more clear and straight forward

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -57,7 +57,7 @@ div {
         attribute xsi:schemaLocation { xsd:anyURI }
     k.image.schemaversion.attribute =
         ## The allowed Schema version (fixed value)
-        attribute schemaversion { "6.5" }
+        attribute schemaversion { "6.6" }
     k.image.kiwirevision.attribute =
         ## A kiwi git revision number which is known to build
         ## a working image from this description. If the kiwi git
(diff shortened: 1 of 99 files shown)


schaefi:
@davidcassany it looks like a ton of changes, but it's actually not. There are many auto changed files due to the schema adaptions. Would be great if you can give the human coded parts a short review :) Thanks

kiwi/schema/kiwi.rnc
```diff
@@ -1293,7 +1303,7 @@ div {
             sch:param [ name = "attr" value = "btrfs_root_is_snapshot" ]
             sch:param [ name = "types" value = "oem vmx docker" ]
         ]
-    k.type.btrfs_root_is_readonly_snapshot =
+    k.type.btrfs_root_is_readonly_snapshot.attribute =
```
davidcassany: Oh good catch 👍 
There were several without following our own criteria 😞

kiwi/xml_parse.py
```diff
@@ -15,7 +15,7 @@
 #   kiwi/schema/kiwi_for_generateDS.xsd
 #
 # Command line:
-#   /home/ms/Project/kiwi/.tox/2.7/bin/generateDS.py -f --external-encoding="utf-8" --no-dates --no-warnings -o "kiwi/xml_parse.py" kiwi/schema/kiwi_for_generateDS.xsd
+#   /home/ms/Project/kiwi/.env3/bin/generateDS.py -f --external-encoding="utf-8" --no-dates --no-warnings -o "kiwi/xml_parse.py" kiwi/schema/kiwi_for_generateDS.xsd
```
davidcassany: yey, finally works with python3

PR #439, opened 2017-07-20
KIWI - Specfile Template - support for Fedora > 25 and CentOS / RHEL 7

mmohring:
Added a working build target for Fedora > 25 and fixed CentOS / RHEL >= 7 package builds.

On CentOS, some additional python2 packages are needed. Due to missing python3 on CentOS / RHEL 7, for this target only the python2-kiwi version is build.

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -87,11 +87,11 @@ BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 %if 0%{?fedora} || 0%{?suse_version}
 BuildRequires:  python3-devel
 BuildRequires:  python3-setuptools
+BuildRequires:  fdupes
 %endif
 BuildRequires:  python-devel
 BuildRequires:  python-setuptools
 %if 0%{?suse_version}
-BuildRequires:  fdupes
 BuildRequires:  shadow
 BuildRequires:  update-alternatives
 %endif
@@ -383,7 +383,7 @@ mv %{buildroot}/%{_defaultdocdir}/packages/python-kiwi/* %{buildroot}/%{_default
 rm -rf %{buildroot}/%{_defaultdocdir}/packages
 %endif
 
-%if 0%{?suse_version}
+%if 0%{?fedora} || 0%{?suse_version}
 %fdupes %{buildroot}/srv/tftpboot
 %fdupes %{buildroot}/%{python3_sitelib}/kiwi/boot
 %fdupes %{buildroot}/%{python2_sitelib}/kiwi/boot
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -62,10 +62,15 @@
 
 # RHEL // CentOS
 # use the rhel templates for CentOS, too
-%if 0%{?rhel_version} == 700 || 0%{?centos_version} == 700
+%if 0%{?rhel} == 7
 %define distro rhel-07.0
 %endif
 
+# Fedora
+# use the rhel templates for CentOS, too
+%if 0%{?fedora} >= 25
+%define distro fedora-25.0
+%endif
 
 Name:           python-kiwi
 Version:        %%VERSION
@@ -79,13 +84,20 @@ Source:         %{name}.tar.gz
 Source1:        %{name}-boot-packages
 Source2:        %{name}-rpmlintrc
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
+%if 0%{?fedora} || 0%{?suse_version}
 BuildRequires:  python3-devel
 BuildRequires:  python3-setuptools
+%endif
 BuildRequires:  python-devel
 BuildRequires:  python-setuptools
+%if 0%{?suse_version}
 BuildRequires:  fdupes
-BuildRequires:  update-alternatives
 BuildRequires:  shadow
+BuildRequires:  update-alternatives
+%endif
+%if 0%{?fedora} || 0%{?rhel}
+BuildRequires:  chkconfig
+%endif
 
 %description
 The KIWI Image System provides an operating system image builder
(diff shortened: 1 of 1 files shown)


package/python-kiwi-spec-template
```diff
@@ -155,6 +181,7 @@ Python 2 library of the KIWI Image System. Provides an operating system
 image builder for Linux supported hardware platforms as well as for
 virtualization and cloud systems like Xen, KVM, VMware, EC2 and more.
 
+%if 0%{?fedora} || 0%{?suse_version}
```
Conan-Kudo: Add `0{?rhel} >= 8` too.

package/python-kiwi-spec-template
```diff
@@ -286,15 +338,19 @@ Provides manual pages to describe the kiwi commands
 # Build Python 2 version
 python2 setup.py build --cflags="${RPM_OPT_FLAGS}"
 
+%if 0%{?fedora} || 0%{?suse_version}
```
Conan-Kudo: Add `0%{?rhel} >= 8` too.

package/python-kiwi-spec-template
```diff
@@ -286,15 +338,19 @@ Provides manual pages to describe the kiwi commands
...
 # Build Python 3 version
 python3 setup.py build --cflags="${RPM_OPT_FLAGS}"
+%endif
 
 %install
 # Install Python 2 version
 python2 setup.py install --prefix=%{_prefix} --root=%{buildroot}
 
+%if 0%{?fedora} || 0%{?suse_version}
```
Conan-Kudo: Add `0%{?rhel} >= 8` too.

package/python-kiwi-spec-template
```diff
@@ -321,9 +377,17 @@ for i in KIWI pxelinux.cfg image upload boot; do \
 done
 %endif
 
+%if  0%{?fedora} || 0%{?rhel}
+install -m 755 -d %{buildroot}/%{_defaultdocdir}/python-kiwi
+mv %{buildroot}/%{_defaultdocdir}/packages/python-kiwi/* %{buildroot}/%{_defaultdocdir}/python-kiwi
+rm -rf %{buildroot}/%{_defaultdocdir}/packages
+%endif
+
+%if 0%{?suse_version}
```
Conan-Kudo: Add `0%{?fedora}` too.

package/python-kiwi-spec-template
```diff
@@ -333,13 +397,15 @@ done
 %{_sbindir}/update-alternatives \
     --install %_bindir/kiwicompat kiwicompat %_bindir/kiwicompat-2 10
 
+%if 0%{?fedora} || 0%{?suse_version}
```
Conan-Kudo: Add `0%{?rhel} >= 8` too.

package/python-kiwi-spec-template
```diff
@@ -349,13 +415,15 @@ done
 %{_sbindir}/update-alternatives \
     --remove kiwicompat %_bindir/kiwicompat
 
+%if 0%{?fedora} || 0%{?suse_version}
```
Conan-Kudo: Add `0%{?rhel} >= 8` too.

package/python-kiwi-spec-template
```diff
@@ -383,6 +451,7 @@ fi
 %{python2_sitelib}/*
 %config %_sysconfdir/bash_completion.d/kiwi-ng-2*.sh
 
+%if 0%{?fedora} || 0%{?suse_version}
```
Conan-Kudo: Add `0%{?rhel} >= 8` too.

package/python-kiwi-spec-template
```diff
@@ -79,13 +84,20 @@ Source:         %{name}.tar.gz
...
 Source2:        %{name}-rpmlintrc
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
+%if 0%{?fedora} || 0%{?suse_version}
 BuildRequires:  python3-devel
 BuildRequires:  python3-setuptools
+%endif
 BuildRequires:  python-devel
 BuildRequires:  python-setuptools
+%if 0%{?suse_version}
 BuildRequires:  fdupes
```
Conan-Kudo: `fdupes` is also available for Fedora.

mmohring:
As @schaefi reuqested, I would like to add CentOS 8 / RHEL 8 support and the %ifs you propose in a separate patch. The title of this patch is CentOS 7 / RHEL 7 / Fedora, and I have tested them also. Btw, where can I test with RHEL 8 ?

Conan-Kudo:
@mmohring You cannot, it doesn't exist yet. :)

However, RHEL 8 will be based on current Fedora, so the conditions are generally going to be the same.

mmohring:
Lets fix this patch. Then we can add RHEL 8 maybe, but what is the sense of it ? Is there some beta of RHEL 8 ?

package/python-kiwi-spec-template
mmohring: I have fixed this.

mmohring:
@schaefi can you take a look if this fits now ?

schaefi:
I vote for adding rhel8 when it actually exists :)

apart from that it looks good to me, @Conan-Kudo any objections to let that go in apart from rhel8 ?

Conan-Kudo:
The main thing missing is that for Fedora (and this may also work for SUSE), if you're going to use alternatives, you need the `Requires(*): %{_sbindir}/update-alternatives`

mmohring:
> The main thing missing is that for Fedora (and this may also work for SUSE), if you're going to use alternatives, you need the Requires(*): %{_sbindir}/update-alternatives

This is the file requires variant. I have used the package requires variant, which is:

Requires:         chkconfig
Requires(post):   chkconfig
Requires(postun): chkconfig

I have checked that.

Conan-Kudo:
Okay, then. 👍

mmohring:
My first KIWI NG patch got merged 👍 into master.

Conan-Kudo:
🎉

PR #446, opened 2017-07-24
Evaluate file strip before kernel strip

### LINKED ISSUE #442: Ability to delete drivers/firmwares
## Problem description
There is no way to define a list of drivers/firmwares to strip, there is `<strip type="delete">` but its more for generic file/binary use.

`<drivers>` should learn a new attribute `type="delete"` and ideally it should be included in the "Deleting unwanted module"-step

schaefi:
Information from the optional <strip type="delete"> section
was handled as part of the suseStripInitrd method which is
called after suseStripKernel. However if a request to delete
a driver file is part of the above mentioned strip section
the checks for the driver dependency and also for potential
superfluous firmware is not applied. Thus the evaluation of
the <strip type="delete"> section happens earlier in the
suseStripKernel method. Fixes #442

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -878,13 +878,6 @@ function baseStripInitrd {
         return
     fi
     #==========================================
-    # Remove unneeded files
-    #------------------------------------------
-    for delete in $kiwi_strip_delete;do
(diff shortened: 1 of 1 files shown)


schaefi:
Great, thanks :)

PR #447, opened 2017-07-25
Fixup lvm disk size setup and map cleanup

schaefi:
This change is two fold

1. Add LVM overhead for lvm based images
    
    LVM itself requires metadata stored in the metadata block
    kiwi did not take a size value for this data into account

2. Use dmsetup to cleanup device maps
    
    Instead of the broken kpartx -d we use dmsetup remove
    directly on the maps kiwi has created

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -434,6 +434,16 @@ class Defaults(object):
         return 30
 
     @classmethod
+    def get_lvm_overhead_mbytes(self):
+        """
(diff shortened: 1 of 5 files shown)


schaefi:
the kpartx issue did not cause the build to fail but it taints your system with open loops because kpartx -d did simply not delete the mapping :( I think they have a fix for it but this is present on many systems out there already. I also like the way that we only remove the maps kiwi has created and not just all of them which belongs to a master device

PR #448, opened 2017-07-25
Fix RHEL spec check

Jellyfrog:
Fixes #N/A.

* Check if OS is RHEL before checking OS-version, else yum will become a requirement on SLES etc.

Note; I haven't checked this on any other OS than SLES12-SP2, so please do before merging...

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -223,7 +223,7 @@ Requires:       squashfs-tools
 Requires:       device-mapper-multipath
 Requires:       gdisk
 %endif
-%if 0%{?rhel} < 8
+%if 0%{?rhel} && 0%{?rhel} < 8
 Requires:       yum
 Provides:       kiwi-packagemanager:yum
 %endif


Conan-Kudo:
@Jellyfrog Pulling in yum is fine, because KIWI is supposed to be able to build non-SUSE things from a SUSE OS.

schaefi:
yes but yum should not be a requirement for the suse package. I think only the package manager of the distribution the package is build for should be a requirement, because I assume the number one use case will be to build packages for that distribution, even though we support building in mixed environments too

Thus I think it would be ok to merge this one

I also found zypper is a requirement for "%if 0%{?fedora} >= 26", is that by intention ?

Conan-Kudo:
@schaefi Yes, though if you want, you could downgrade the `yum` and `zypper` dependencies to Recommends for Fedora, and downgrade the `yum` and `dnf` dependencies to Recommends for SUSE.

schaefi:
ok I'll look into it

PR #449, opened 2017-07-25
Fixup encoding handling for profile file

schaefi:
If an element like displayname or oem-boot-title contains characters outside of the ascii table this causes trouble when kiwi writes out the profile file and the code is called through python2. The reason here is that the default encoding on write() (and other methods) is set to ascii in python2 and when it receives unicode characters outside of the ascii spec a UnicodeEncodeError is thrown. Now all of kiwi is using Unicode which means this does not produce a problem when calling the code through python3 because the default encoding is utf-8 there. This patch introduces a method which allows to change python's default encoding and calls it at the code point where we write the profile because we got unicode and we want to write unicode in any case. This fixes at least one situation for python2-kiwi to allow the use of non ascii characters in the XML setup. If other places will be found the same approach should allow to fix it for python2

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -17,6 +17,8 @@
 #
 import os
 import glob
+import sys
+from six.moves import reload_module
 from collections import namedtuple
 import platform
 from pkg_resources import resource_filename
@@ -842,6 +844,22 @@ class Defaults(object):
         :rtype: string
         """
         return os.sep.join([root_dir, 'image', 'imported_root'])
+
+    @classmethod
+    def set_python_default_encoding_to_utf8(self):
+        """
+        Set python default encoding to utf-8 if not already done
+
+        This is not a safe operation since sys.setdefaultencoding()
+        was removed from sys on purpose when Python starts. Reenabling
+        it and changing the default encoding can break code that relies
+        on ascii being the default. Within the scope of kiwi the
+        operation is safe because all data is expected to be utf-8
+        everywhere and considered a bug if this is not the case
+        """
+        if sys.version_info.major < 3:
+            reload_module(sys)  # Reload required to get setdefaultencoding back
+            sys.setdefaultencoding('utf-8')
 
     def get(self, key):
         """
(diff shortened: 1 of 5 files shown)


schaefi:
@tomschr your feedback on this one is also much appreciated :)

schaefi:
* changing the default encoding at startup seemed too intrusive for me as I can't estimate the impact for other commands. Thus I limited it to the place where I know utf-8 encoding is required.
* any import of the sys module sets back the default encoding, thus it's not so easy to make sure one encoding setup is enough. I'm sure this is one of the reasons why they deleted the method on import ;)

I'm also not 100% happy with this code but I don't see a better way at the moment

tomschr:
@schaefi 
I'm not sure how to comment on that. I don't really feel qualified enough about encodings. 😉

> changing the default encoding at startup seemed too intrusive for me as I can't estimate the impact for other commands. Thus I limited it to the place where I know utf-8 encoding is required.

This is a good strategy IMHO.

> any import of the sys module sets back the default encoding, thus it's not so easy to make sure one encoding setup is enough. I'm sure this is one of the reasons why they deleted the method on import ;)

The official line in Python seems to be "don't use it!". However, it seems, in the wild, there are many cases where it is used.

I've found an interesting post on [StackOverflow](https://stackoverflow.com/a/29832646). His conclusions were:

1. Setting the defaultencoding to UTF-8 nowadays is safe, except for specialised applications, handling files from non Unicode ready systems.
2. The "official" rejection of the switch is based on reasons no longer relevant for a vast majority of end users (not library providers), so we should stop discouraging users to set it.
3. Working in a model that handles Unicode properly by default is far better suited for applications for inter-systems communications than manually working with Unicode APIs.

The post is longer than the short conclusion I've cited, so I would recommend to read it. I don't know if his claims are valid, but they looked accurate.

Hope this helps a bit. 😄

schaefi:
all right after a nice in person discussion with Tom I'm going to merge this as an acceptable solution :)

Thanks

PR #452, opened 2017-07-26
Adding mediacheck menu entry for liveCDs

### LINKED ISSUE #450: --isocheck is missing in KIWI-NG
## Problem description

Former KIWI version includes the flag `--isocheck` which includes a menu entry for live isos bootloader. such an entry is used to verify the integrity of the ISO image (commonly used to validate downloads)

## Actual behaviour

Former KIWI v7 version includes that option which is used to produce the openSUSE live CDs. While moving to KIWI-NG in Tumbleweed it appeared that the Mediacheck menu entry is missing.

This issue is related to bsc#1049304 reported by @DimStar77

davidcassany:
This PR adds the `checkiso` attribute in `type` tag, only available for
iso image types. Checkiso is boolean that determines the presence of0
a Mediacheck menu entry or not in the bootloader.

This patch is related to **bsc#1049304**

Fixes #450

### DIFF
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+import platform
 from collections import namedtuple
 
 # project
@@ -184,6 +185,17 @@ class BootLoaderConfigBase(object):
         if self.xml_state.build_type.get_installprovidefailsafe() is False:
             return False
         return True
+
+    def mediacheck_boot_entry_requested(self):
+        """
+        Check if a mediacheck boot entry is requested
+
+        :rtype: bool
+        """
+        if self.xml_state.build_type.get_checkiso() is True and \
+                platform.machine() in ['x86_64', 'i586', 'i686']:
+            return True
+        return False
 
     def get_boot_cmdline(self, uuid=None):
         """
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -15,7 +15,6 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
-import platform
 from collections import namedtuple
 
 # project
@@ -185,17 +184,6 @@ class BootLoaderConfigBase(object):
         if self.xml_state.build_type.get_installprovidefailsafe() is False:
             return False
         return True
-
-    def mediacheck_boot_entry_requested(self):
-        """
-        Check if a mediacheck boot entry is requested
-
-        :rtype: bool
-        """
-        if self.xml_state.build_type.get_checkiso() is True and \
-                platform.machine() in ['x86_64', 'i586', 'i686']:
-            return True
-        return False
 
     def get_boot_cmdline(self, uuid=None):
         """
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -72,7 +72,6 @@ class LiveImageBuilder(object):
         self.types = Defaults.get_live_iso_types()
         self.hybrid = xml_state.build_type.get_hybrid()
         self.volume_id = xml_state.build_type.get_volid()
-        self.checkiso = xml_state.build_type.get_checkiso()
         self.mbrid = SystemIdentifier()
         self.mbrid.calculate_id()
         self.filesystem_custom_parameters = {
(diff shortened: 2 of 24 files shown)


davidcassany:
@schaefi I have included a new attribute to port the `--isocheck` from kiwi v7, as I believe that the presence of the Mediacheck menu entry should not relay on the command line but on the description file, we could add a such a command line flag, but IMHO it would be strange to make flag for such a specific image detail.

@DimStar77, with this change the image description file is required to be updated in order to include the Mediacheck boot menu entry, would it be acceptable?

kiwi/schema/kiwi.rnc
```diff
@@ -1644,6 +1644,15 @@ div {
             sch:param [ name = "attr" value = "installpxe" ]
             sch:param [ name = "types" value = "oem" ]
         ]
+    k.type.checkiso.attribute =
+        ## Specifies if the bootloader menu should provide an"
+        ## mediacheck entry to verify ISO integrity or not.
+        ## Disabled by default and only available for x86 arch family."
+        attribute checkiso { xsd:boolean }
```
schaefi: Just thinking if mediacheck="true|false" would be a better name ? imho checkiso sounds like we would perform a check on the iso

schaefi:
Overall I like the implementation, especially to put on an attribute for the iso type only instead of the weird option I had done before. Just see my comments about the bootloader interface and the runtime checker

DimStar77:
> @DimStar77, with this change the image description file is required to be updated in order to include the Mediacheck boot menu entry, would it be acceptable?

I can easily live with that, yes

kiwi/builder/live.py
```diff
@@ -236,6 +237,15 @@ def create(self):
                 iso_header_offset, self.mbrid, self.isoname
             )
 
+        # include metadata for checkmedia tool
+        if self.checkiso is True and self.arch not in ['x86_64', 'ix86']:
+            log.warning(
+                'checkiso attribute is only supported for x86 architecture '
+                'family'
+            )
```
davidcassany: I am not entirely sure about that, I was thinking on having a single kiwi description that gets build for several architectures in OBS without having to include profiles. I am not even sure if actually exists a combination (firmware, bootloader, etc.) that can be applied for isos and mulitple architectures, but that was the motivation to ignore the attribute without failing.

kiwi/bootloader/config/base.py
```diff
@@ -185,6 +186,17 @@ def failsafe_boot_entry_requested(self):
...
+        """
+        Check if a mediacheck boot entry is requested
+
+        :rtype: bool
+        """
+        if self.xml_state.build_type.get_checkiso() is True and \
+                platform.machine() in ['x86_64', 'i586', 'i686']:
+            return True
+        return False
+
```
davidcassany: Yes you are right, I haven't done it this way because I wanted to have a slightly different check in builder/live.py. But if we include the runtime check to just fail in non x86 architectures then this is fine, the same xml_state method to get the isocheck could be used in both, builder/live.py and in this bootloader interface.

kiwi/builder/live.py
```diff
@@ -236,6 +237,15 @@ def create(self):
...
+        elif self.checkiso is True:
+            Iso.add_tagmedia_check(self.isoname)
```
davidcassany: your suggestion is fine for me

kiwi/schema/kiwi.rnc
davidcassany: Yes that's true, it sounds more like a buildtime check. I will update the attribute name.

kiwi/bootloader/config/base.py
schaefi: yes :+1:

kiwi/builder/live.py
schaefi: well if the mediacheck is globally requested but can't be done when building for another architecture my opinion is that the author of the image description either has to create an arch specific profile in the XML description or use different image configurations. I'm not a friend of the implicit handling of arch specific differences. We have that in kiwi v7 and in obs to no end and it always confuses me when things are different but not made explicit in the description

schaefi:
@davidcassany sorry the merge of the other schema related pull request caused a conflict of the auto generated code from this request. It should resolve easily of you rebase and re-create the xml_parse.py code via the Makefile

davidcassany:
Ah yes sure, I was already aware that it was going to happen to one of the two PRs, no problem, updating right now.

kiwi/xml_state.py
```diff
@@ -623,6 +623,17 @@ def get_build_type_spare_part_size(self):
...
+    def get_build_type_mediacheck(self):
+        """
+        Get the mediacheck attribute from the build type. If
+        no mediacheck is present returns False.
+
+        :rtype: bool
+        """
+        if self.build_type.get_mediacheck() is True:
+            return True
+        return False
```
schaefi: Just an idea and I leave it up to you if you want to change it:

The mediacheck attribute is specified as xsd:boolean, thus you can rely on the generated api to provide a True, False or None(if not specified) value. As the arch constraint is covered by the runtime test it would be possible to drop the method and directly read the information from xml_state.build_type.get_mediacheck(). As long as None(not specified) and False can be treated the same way which is the case here this could be done

schaefi: for the benefit of a few less lines of code

davidcassany: Yes I agree, I'll make sure there is no test to be updated.

PR #456, opened 2017-08-01
Extend scope of check_docker_tool_chain_installed

### LINKED ISSUE #454: Version check for umoci / skopeo tools used in kiwi
## Problem description

Since those tools seems to approach incompatible api changes in the future we should adapt the currently used runtime check to be more strict and check not only on the presence of the tools but also for their major version

schaefi:
The runtime check only checks for the presence of the docker
tools umoci and skopeo but as we are expecting incompatible
api changes in the next major version release the check has
been extended to look at the major versions of the installed
tools as well. Fixes #454

### DIFF
--- a/kiwi/runtime_checker.py
+++ b/kiwi/runtime_checker.py
@@ -151,7 +151,7 @@ class RuntimeChecker(object):
             Expected version: v{want_version}.x.x but got: v{got_version}.x.x
         ''')
 
-        expected_version = 0
+        expected_version = 1
 
         if self.xml_state.get_build_type_name() == 'docker':
             for tool in ['umoci', 'skopeo']:
--- a/kiwi/runtime_checker.py
+++ b/kiwi/runtime_checker.py
@@ -16,6 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
+import re
 import platform
 from textwrap import dedent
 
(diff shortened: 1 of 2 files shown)


kiwi/runtime_checker.py
```diff
@@ -134,19 +136,51 @@ def check_docker_tool_chain_installed(self):
...
 
             http://download.opensuse.org/repositories/Virtualization:/containers
         ''')
+        message_version_unsupported = dedent('''\n
+            {name} tool found in unsupported version
+
+            Expected version: v{want_version}.x.x but got: v{got_version}.x.x
+        ''')
+
+        expected_version = 0
```
davidcassany: I will coordinate with @cyphar about the version to require, but from what I understood from him, this should be fine for quite a while, as no breaking changes are expected.

schaefi: yeah maybe we can invite him to review, just could not find him in the list

schaefi: @cyphar hi interested to have a look if that makes sense ?

cyphar: Yeah, this looks good to me. However I would wait until `umoci` hits `1.0` before doing this (since SemVer only defines such version incompatibilities after `1.0`) -- or making the check just check for `1`. *If* we make the API incompatible it will be in a `2.0` (because we're going to maintain the `1.0` release branch for a reasonable period of time).

schaefi:
ok I have increased version expectation to 1.x.x, that means any version of umoci and skopeo <= 1.x.x
is expected to work the way we use it in kiwi. Does that makes sense ?

schaefi:
all right :)

PR #460, opened 2017-08-02
Export packages list file and package verification file also for dpkg

### LINKED ISSUE #457: Also dump .packages file for debian builds
You can use something like
```
dpkg-query -W -f '${Package}|${Version}|${Architecture}\n'
```
to get the list of installed packages. (You may want to split the version into epoch/version/release if you want to make the format more identical to rpm. But dpkg-query offers no interface for this, so you need to do that in some other way.)

davidcassany:
With this PR the methods `export_rpm_package_list` and `export_rpm_package_verification` are renamed to `export_package_list` and `export_package_verification` respectively. Also the new renamed methods are can list and verify deb or rpm packages, depending on the selected package manager on the description file. 

Fixes #457

### DIFF
--- a/kiwi/builder/archive.py
+++ b/kiwi/builder/archive.py
@@ -105,7 +105,7 @@ class ArchiveBuilder(object):
             )
             self.result.add(
                 key='image_packages',
-                filename=self.system_setup.export_rpm_package_list(
+                filename=self.system_setup.export_package_list(
                     self.target_dir
                 ),
                 use_for_bundle=True,
--- a/kiwi/builder/archive.py
+++ b/kiwi/builder/archive.py
@@ -114,7 +114,7 @@ class ArchiveBuilder(object):
             )
             self.result.add(
                 key='image_verified',
-                filename=self.system_setup.export_rpm_package_verification(
+                filename=self.system_setup.export_package_verification(
                     self.target_dir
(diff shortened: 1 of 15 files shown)


davidcassany:
@schaefi I am no very convinced of having such a test in kiwi/system/setup.py to behave in some specific way according to the package manager. However I don't think it makes sense to include those package listing and package verification methods on kiwi/package_manager or kiwi/repository, as at that point we don't want to create any new PackageManager or Repository instance and they serve different purposes. Also, I am not convinced those two export methods deserve new classes.

schaefi:
I like the implementation and the design 👍 I agree putting it into the package manager interface seems to be not really required and we are not blocked to do that later as a refactoring task when ever needed.
We might think of creating a new class e.g PackageMetadata which we structure like the package manager interface classes to allow us creating a Factory for different package meta data. As of today the list of packages and the verification information would be the only two interface methods of that class, thus I'm totally ok to take it as it is right now and think of a refactoring if there is more software meta data we need to manage

I would just merge it

Thoughts ?

davidcassany:
I agree, if some package-manager related metadata is included in the future we can think up about including a new interface with another package-manager Factory, but for the time being we are fine.

Let's merge

PR #464, opened 2017-08-11
Leap 15

lnussel:


schaefi:
@lnussel Thanks, just for my own understanding suse-leap15.0 is not the same as suse-SLES15 which already exists, meaning leap15.0 is a different distro than SLES15, correct ?

lnussel:
yes, just like leap 42 vs sle 12

lnussel:
so, ok to merge?

schaefi:
ok thanks

PR #469, opened 2017-08-21
Boot descriptions for s390 and ppc

schaefi:
Added boot descriptions for SLE15 on ppc and s390


schaefi:
[Create boot descriptions for SLES15](https://trello.com/c/ofVCsL1F/202-create-boot-descriptions-for-sles15)

PR #474, opened 2017-09-01
Include the source label in container images

### LINKED ISSUE #473: Include unique identifier for BS builds
## Problem description

Images build in the BS do not include any unique identifier to trace the source were is was build (like the DISTURL parameter in the RPMs).

## Actual behaviour

An image that is build in BS and published to a registry do not contain any reference to the former BS build, which is required by L3 in order to gather valuable information to provide support.

This is coming from bsc#1055542

davidcassany:
This commit fixes bsc#1055542, it includes the DISTURL parameter
as a label into the container metadata when the build happens inside
the buildservice. The new label included is named:

 * org.opencontainers.image.source

in line with the current OCI annotations schema.

Fixes #473

### DIFF
--- a/kiwi/container/oci.py
+++ b/kiwi/container/oci.py
@@ -107,6 +107,13 @@ class ContainerImageOCI(object):
             if 'xz_options' in custom_args:
                 self.xz_options = custom_args['xz_options']
 
+        # for builds inside the buildservice we include a reference to the
+        # specific build. Thus source label only exists inside the
+        # buildservice.
+        source_label = self._get_source_label()
+        if source_label:
+            self.labels.append(source_label)
+
         if not custom_args or 'container_name' not in custom_args:
             log.info(
                 'No container configuration provided, '
@@ -206,6 +213,23 @@ class ContainerImageOCI(object):
         oci_tarfile.create_xz_compressed(
             image_dir, xz_options=self.xz_options
         )
+
+    def _get_source(self):
+        if Defaults.is_buildservice_worker():
+            with open(os.sep + Defaults.get_buildservice_env_name()) as env:
+                for line in env:
+                    if line.startswith('BUILD_DISTURL') and '=' in line:
+                        return line.split('=')[1].strip()
+                log.warning('Could not find DISTURL inside .buildenv')
+
+        return None
+
+    def _get_source_label(self):
+        source = self._get_source()
+        if source:
+            return ''.join([
+                '--config.label=org.opencontainers.image.source=', source
+            ])
 
     def __del__(self):
         if self.oci_dir:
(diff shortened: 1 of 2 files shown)


kiwi/container/oci.py
```diff
@@ -107,6 +107,13 @@ def __init__(self, root_dir, custom_args=None):
             if 'xz_options' in custom_args:
                 self.xz_options = custom_args['xz_options']
 
+        # for builds inside the buildservice we include a reference to the
+        # specific build. Thus source label only exists inside the
+        # buildservice.
```
schaefi: imho it would be more obvious if the comment which says this is only for the buildservice and the actual code which checks that in Defaults.is_buildservice_worker() would be in the same block of instructions. When I read the comment I needed to search where we the buildservice check appeared. But I leave that up to you

kiwi/container/oci.py
```diff
@@ -207,6 +214,23 @@ def pack_image_to_file(self, filename):
...
 
+    def _get_source(self):
+        if Defaults.is_buildservice_worker():
+            with open(os.sep + Defaults.get_buildservice_env_name()) as env:
+                for line in env:
+                    if line.startswith('BUILD_DISTURL') and '=' in line:
+                        return line.split('=')[1].strip()
+                log.warning('Could not find DISTURL inside .buildenv')
+
+        return None
```
schaefi: Returning None is not needed, that's the default

kiwi/container/oci.py
```diff
@@ -207,6 +214,23 @@ def pack_image_to_file(self, filename):
...
+
+    def _get_source_label(self):
+        source = self._get_source()
+        if source:
+            return ''.join([
+                '--config.label=org.opencontainers.image.source=', source
+            ])
+
```
schaefi: not sure, but maybe the two private methods _get_source and _get_source_label could be combined into one _get_buildservice_source_label ? imho _get_source only exists to build the label right ? so I did not understand why an extra method is needed. I might be wrong on that though

davidcassany: I had splitted them because in my initial attempt the _get_source method was implemented inside kiwi/defaults.py, but later I thought that there was no need for a generic access to that value since I don't think we will need it for anything else. You are right they can be combined.

kiwi/container/oci.py
davidcassany: Yes I know, I included it for clarity, but I can certainly live without it.

kiwi/container/oci.py
davidcassany: yes, you are right, if the _get_source method is private there is no need to check Defaults.is_buildservice_worker inside that method, we can do it out of it.

davidcassany:
@scheafi I rearranged slightly different in order to get rid of a flake complaint of having a too complex __init__ method.

schaefi:
Looks good 👍

PR #477, opened 2017-09-04
Move to dracut for live ISO images

### LINKED ISSUE #470: Refactor live iso layout to be able to use dracut
## Problem description

kiwi implemented its own live iso layout which is nice and works. However with regards to the intention of kiwi to not re-invent the wheel and the fact that newer versions of dracut also implements a live boot mechanism it would be worth to investigate to use the dracut initrd instead of the kiwi isoboot code.
This would also generalize the live layout throughout the various distributions

schaefi:
Instead of the kiwi isoboot layout the live layout has been changed to follow the rules of the live system support offered by dracut. The initrd used to boot up the system is now a dracut generated initrd and can be customized via the 'flags' attribute in the kiwi XML description. In order to meet the kiwi live requirements a custom dracut module called kiwi-live is provided along with the changes in kiwi. This Fixes #470


davidcassany:
@schaefi, great job done here, looks very promising, congrats! I haven't gone through the code yet, but I'll try give it a try today/tomorrow morning and look closer to it, at least to satisfy my curiosity.

schaefi:
@davidcassany this one finally eliminates kiwi's custom isoboot code and uses dracut. The dracut live module will work with the iso layout kiwi provides but is not used. I wrote an additional dracut module to drive our live images which only uses overlayfs and does not intermingle with old code before overlayfs existed. In addition persistent writing is automatically setup if the target has the capability which is not supported that way by the upstream live module.

Therefore an additional package kiwi-dracut-iso is provided with kiwi and also as extra package for the other distributions we support. As the package will build when this one is merged we have a chicken and egg problem if you want to give it a real test. Thus I did a local osc build of the kiwi package to get the kiwi-dracut-iso package as it is used here: https://github.com/SUSE/kiwi-descriptions/pull/7
and put the package in a local repository with 'createrepo'

schaefi:
Here is the procedure to setup a local repo with the dracut module package

1. in the kiwi git call make build from within the venv
2. osc co Virtualization:Appliances:Staging/python-kiwi
3. cp dist/* Virtualization:Appliances:Staging/python-kiwi
4. osc build

this builds among others the kiwi-dracut-iso-9.10.6-0.x86_64.rpm package

mkdir myrepo
cp kiwi-dracut-iso-9.10.6-0.x86_64.rpm myrepo
cd myrepo
createrepo .

now you have a local repo with the package.

git clone https://github.com/SUSE/kiwi-descriptions.git
git checkout move_to_dracut_live_layout

vi suse/x86_64/suse-leap-42.3-JeOS/config.xml

and add

```xml
<repository type="rpm-md" priority="1">
    <source path="dir:///myrepo"/>
</repository>
```

all the rest of the description has already been adapted, you need that repo only for testing
after that you can directly build suse-leap-42.3-JeOS iso live image as an example from the
kiwi git branch: move_to_dracut_live_layout

schaefi:
I also would like to get rid of vmxboot but need to investigate if that can be done without loosing features. oemboot and netboot has to stay for a while because of their deployment code used in a wide range

package/python-kiwi-spec-template
```diff
@@ -308,6 +308,24 @@ This package contains the basic PXE directory structure which is
 needed to serve kiwi built images via PXE.
 %endif
 
+%package -n kiwi-dracut-iso
```
Conan-Kudo: Please rename this to `dracut-kiwi-live`, as that follows the conventions of dracut module packaging in Fedora, openSUSE, and Mageia.

package/python-kiwi-spec-template
```diff
@@ -501,6 +519,12 @@ fi
 %exclude %{_bindir}/kiwicompat-*
 %{_bindir}/*
 
+%files -n kiwi-dracut-iso
+%defattr(-, root, root)
+%dir %{_usr}/lib/dracut
+%dir %{_usr}/lib/dracut/modules.d
```
Conan-Kudo: Since you already require dracut, you should not attempt to co-own dracut directories.

package/python-kiwi-spec-template
schaefi: ok thanks for the hint, will do

package/python-kiwi-spec-template
schaefi: yes fixing

Conan-Kudo:
@schaefi I would be shocked if you couldn't implement netboot, vmxboot, and oemboot in a way that leverages dracut. Some of the functionality in those modes are already implemented in dracut for network boot, for example: [network](https://github.com/dracutdevs/dracut/tree/master/modules.d/40network), [livenet](https://github.com/dracutdevs/dracut/tree/master/modules.d/90livenet)

schaefi:
sure dracut could all do that but a little different than the netboot and oemboot code does it. I need to be careful to not break peoples deployment environment, that needs a closer look. I'm confident to get there

kiwi/runtime_checker.py
```diff
@@ -261,6 +261,34 @@ def check_consistent_kernel_in_boot_and_system_image(self):
...
+            Required dracut module package missing in package list
+
+            The package '{0}' is required for the selected
+            live iso image type. Please add the following in your
+            <packages type="image"> section to your system XML
+            description:
+
+            <package name="{0}"/>
+        ''')
+        required_dracut_package = 'kiwi-dracut-iso'
```
Conan-Kudo: This should be fixed to `dracut-kiwi-live`

kiwi/runtime_checker.py
```diff
@@ -261,6 +261,34 @@ def check_consistent_kernel_in_boot_and_system_image(self):
...
+        if self.xml_state.get_build_type_name() == 'iso':
+            package_names = \
+                self.xml_state.get_bootstrap_packages() + \
+                self.xml_state.get_system_packages()
+            if 'kiwi-dracut-iso' not in package_names:
```
Conan-Kudo: This should be fixed to `dracut-kiwi-live`.

PR #483, opened 2017-09-07
Fix oem_kiwi_initrd

### LINKED ISSUE #465: oem-kiwi-initrd fails on reboot
## Problem description

There is a report on the mailing list saying that the use of

```xml
<oem-kiwi-initrd>true</oem-kiwi-initrd>
```

will cause the image to boot once but failed on reboot in a configuration with lvm
enabled as follows:

```xml
<type image="oem" boot="oemboot/rhel-07.0" filesystem="ext3" installiso="true" bootloader="grub2" >
         <systemdisk name="volgroup01" preferlvm="true">
              <volume name="LVhome" size="200M" mountpoint="/home"/>
              <volume name="@root" freespace="5G" />
              <volume name="LVvar" mountpoint="/var" freespace="all"/>
              <volume name="LVlog" size="200M" mountpoint="var/log"/>
              <volume name="LVtmp" size="1800M" mountpoint="/tmp"/>
          </systemdisk>
          <oemconfig>
              <oem-kiwi-initrd>true</oem-kiwi-initrd>
          </oemconfig>
</type>
```

davidcassany:
This commit fixes the path of the bind mounted boot partition. Note that the boot partition does not include the boot folder, thus after `mount $imageBootDevice $prefix/$mpoint` the mountpoint holds the content of the `/boot/*` folder in a root tree.

This commit fixes #465

### DIFF
--- a/kiwi/boot/functions.sh
+++ b/kiwi/boot/functions.sh
@@ -4558,23 +4558,40 @@ function kernelList {
     local kernel=""
     local initrd=""
     local kpair=""
-    local krunning=`uname -r`
+    local krunning=$(uname -r)
+    local irdbase=initrd
+    local irdpostfix=""
     KERNEL_LIST=""
     KERNEL_NAME=""
     KERNEL_PAIR=""
     #======================================
+    # Enter boot path
+    #--------------------------------------
+    if [ -d ${prefix}/boot_bind ];then
+        pushd ${prefix}/boot_bind
+    else
+        pushd ${prefix}/boot
+    fi
+    #======================================
+    # check initrd naming style
+    #--------------------------------------
+    if [ -e initramfs-${krunning}.img ];then
+        irdbase=initramfs
+        irdpostfix=.img
+    fi
+    #======================================
     # search running kernel first
     #--------------------------------------
-    if [ -d $prefix/lib/modules/$krunning ];then
+    if [ -d ${prefix}/lib/modules/${krunning} ];then
         for name in vmlinux vmlinuz image uImage;do
-            if [ -f $prefix/boot/$name-$krunning ];then
-                kernel=$name-$krunning
-                initrd=initrd-$krunning
+            if [ -f ${name}-${krunning} ];then
+                kernel=${name}-${krunning}
+                initrd=${irdbase}-${krunning}${irdpostfix}
                 break
             fi
         done
         if [ ! -z "$kernel" ];then
-            KERNEL_PAIR=$kernel:$initrd
+            KERNEL_PAIR=${kernel}:${initrd}
             KERNEL_NAME[$kcount]=$krunning
             KERNEL_LIST=$KERNEL_PAIR
             kcount=$((kcount+1))
@@ -4583,7 +4600,7 @@ function kernelList {
     #======================================
     # search for other kernels
     #--------------------------------------
-    for i in $prefix/lib/modules/*;do
+    for i in ${prefix}/lib/modules/*;do
         if [ ! -d $i ];then
             continue
         fi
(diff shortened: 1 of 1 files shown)


schaefi:
Hmm, that change concerns me. Apart from that I don't like my own coding here, historical grown bash code, hard to read and to understand, I think this is not the right change.

So this entire boot partition thing must only happen if there is a real boot partition used by the image
The lines:

```
if [ ! "$label" = "BOOT" ] || [ ! -e "$imageBootDevice" ];then
        return
    fi
```

should asure this.

The boot data from the image boot directory has been preserved in /boot.tgz created by kiwi at another place of the boot code. The way this data is copied should make sure there is always the path structure available as used in the later bind mount:

```
mkdir -p /boot_bind
mount /dev/bootpartition /boot_bind
cp -a /some_prefix/boot /boot_bind
```

The cp command makes sure there is always the path:

* boot on /dev/bootpartition

except there is no directory /some_prefix/boot which is not checked by the code and could therefore be a part of the problem

Given the cp command worked as expected, the remount of the bootpartition to:

* mount /dev/bootpartition /some_prefix/boot_bind

should asure the path

*  /some_prefix/boot_bind/boot

to exist. Your patch changes this and I think we should not do this because subsequent code expects the path to be present in order to be able to resolve the bind mounted boot back to the standard way

I really think the problem starts with that 'cp -a' command

Thougths ?

davidcassany:
Yes I have to admit it wasn't easy to debug... I had to take paper and pencil to trace it :/ 

The change is hard to validate I am aware of it. If you don't mind,  before commenting on your thoughts, let me try to explain my own process to get to that change. This issue in the oem boot appears when there is a boot partition (it appeared in all tests I did with `bootpartition="true"`), the problem is that for some reason the path `/mnt/boot` is empty (no filer nor folders in it) thus the method `kernelList` in `kiwi/boot/functions.sh` does not find any kernel and the 'export KERNEL_LIST' exports and empty variable which causes the method `setupKernelLinks` to overwrite the actual kernel and initrd with a broken link, since the code:

```bash
rm -f $initrd && ln -s initrd.vmx $initrd
rm -f $kernel && ln -s linux.vmx  $kernel
```
translates to: 

```bash
rm -f && ln -s initrd.vmx
rm -f && ln -s linux.vmx
```

which overwrites the files **linux.vmx** and **initrd.vmx** with an inconsistent symlink pointing to themselves (I am surprised that those calls do not fail...). My debugging showed that there was no kernel detected because `/mnt/boot` was completely empty. Then I saw that the: 

```bash
mount --bind $prefix/$mpoint/boot $prefix/boot
```

translated to:

```bash
mount --bind /mnt/boot_bind/boot /mnt/boot
```
was mounting an empty folder and that the `/mnt/boot_bind` contained all the data expected in `/mnt/boot`. Then I thought that the `/boot` subfolder in `$prefix/$mpoint/boot` could be a typo. Also I would have expected a link like `boot -> .` to prevent it, but there was no.

> The boot data from the image boot directory has been preserved in /boot.tgz created by kiwi at another place of the boot code.

```bash
if [ -e /boot.tgz ];then
    tar -xf /boot.tgz -C /$mpoint
fi
```

This part of the code of `setupBootPartition` did not get executed, thus there was no /boot.tgz (no clues about it, I do not really understand this step).

>The cp command makes sure there is always the path:
>
>* boot on /dev/bootpartition
>
>except there is no directory /some_prefix/boot which is not checked by the code and could therefore be a part of the problem

In my tests there was no *boot* path in */dev/bootpartition*, if the codes relies on that, this is certainly a bug.

Then the question is form where the boot path should come from?

I did most of my tests in rhel (because of the download.opensuse.org downtime), thus not sure all my comments exactly apply for openSUSE, even I am quite confident about that because in openSUSE builds you can see the broken initrd.vmx and vmlinux.vmx links, as if no kernel was detected by `kernelList` method.

schaefi:
I agree this is hard to debug, thanks much for taking the time to dig so deep into it. I suggest I reproduce the situation as you described it and then provide my feedback. Together we will find a proper fix. I won't be able to run a test today (want to fix the mediacheck problem) thus I'll come back to this one on Monday. Is that ok for you ? I guess this one does not block other work correct ?

davidcassany:
@schaefi, sure this issue is not blocking anything that I am aware of. No problem, just let you know that Monday is bank holiday in Barcelona, thus I won't be here around.

schaefi:
@davidcassany hope you don't mind I think I've found the real cause of the problem and submitted on top of your branch a solution to the problem

davidcassany:
@schaefi, sure no problem, good you found the root issue

PR #484, opened 2017-09-08
Do not use fillup-templates/group.aaa_base

schaefi:
When kiwi initializes a new root system it copies an existing
group template file as a start. However it has turned out if
that group template is not compatible with the target image
it might lead to a broken etc/group file in the target image
This Fixes bsc#1056668


schaefi:
There is an ongoing discussion if kiwi should be changed like in this pull request. Therefore please do not merge until we come to an agreement here: https://bugzilla.suse.com/show_bug.cgi?id=1056668

schaefi:
This one won't be merged, see bug report for details

PR #485, opened 2017-09-08
Rename source label

### LINKED ISSUE #473: Include unique identifier for BS builds

davidcassany:
This commit renames the source label to a more explicit name. This
commit also relates to bsc#1055542 and fixes #473


schaefi:
:+1:

PR #486, opened 2017-09-09
Add mediacheck handler for live iso

schaefi:
This fixes bsc#1049304


schaefi:
@davidcassany this patch also includes a small refactoring, common code from the dracut module has been moved into a library to avoid code duplication

I raised another concern about the separate Mediacheck boot menu entry in the bug here

* https://bugzilla.suse.com/show_bug.cgi?id=1049304

Thus this one should not be merged until needinfo is clear from the bug. Of course I would appreciate your opinion about my questions from the bug report :)

Thanks

schaefi:
the mediacheck is a time consuming task. The bigger the iso the longer it takes. Thus it makes sense to keep the extra boot menu entry

PR #497, opened 2017-09-15
Limit the characters set for volid attribute

### LINKED ISSUE #493: Setting volume id with ISO images breakes dracut boot
## Problem description

Live iso images do not boot if `volid` attribute is used. Apparently giving a volume id breaks the boot process with dracut.

## Actual behaviour

iso images do not boot.

Use a `<type>` section like the following one:

```xml
<type primary="true" flags="overlay" hybrid="true" hybridpersistent="true" hybridpersistent_filesystem="ext4" image="iso" firmware="uefi" volid="GNOME Next"/>
```

The resulting image won't boot.

davidcassany:
This commit fixes #493


schaefi:
great 👍

PR #502, opened 2017-09-18
Fix trailing pipe character in .packages file

### LINKED ISSUE #501: Trailing '|' at the end of .packages
## Problem description

The generated .packages file contains a trailing '|' at the end of each line. This is different from former version of kiwi providing that information and should be fixed

In addition the format of the .packages and .verified files should follow the same layout no matter which package manager was used. This might be tricky and could be done in an extra effort if it turns out to be too complex

schaefi:
In addition make sure the field layout is consistent across
the .packages files no matter which package manager was used
to create the information. This Fixes #501

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -841,10 +841,10 @@ class SystemSetup(object):
                 'rpm', '--root', self.root_dir, '-qa', '--qf',
                 '|'.join(
                     [
-                        '%{NAME}', '%{EPOCH}', '%{VERSION}', '%{RELEASE}',
-                        '%{ARCH}', '%{DISTURL}', '\\n'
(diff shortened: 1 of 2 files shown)


kiwi/system/setup.py
```diff
@@ -856,7 +856,12 @@ def _export_deb_package_list(self, filename):
             [
                 'dpkg-query', '--admindir',
                 os.sep.join([self.root_dir, 'var/lib/dpkg']), '-W', '-f',
-                '${Package}|${Version}|${Architecture}\n'
+                '|'.join(
+                    [
+                        '${Package}', 'None', '${Version}',
+                        'None', '${Architecture}', 'None'
+                    ]
```
davidcassany: Great, makes a lot of sense

PR #504, opened 2017-09-19
Fixed package requires for dracut-kiwi-live

schaefi:
On Fedora/RHEL cdrkit is only a source package building other
packages. The package dracut-kiwi-live really needs is
genisoimage

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -316,7 +316,7 @@ Requires:       e2fsprogs
 Requires:       util-linux
 Requires:       dracut
 %if 0%{?fedora} || 0%{?rhel}
-Requires:       cdrkit
+Requires:       genisoimage
 %else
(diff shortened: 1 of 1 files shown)


schaefi:
I know a file requires like:

```
Requires: %{_bindir}/isoinfo
```

would be the preferred solution but this information is not used by the buildservice with reference to
https://en.opensuse.org/openSUSE:Build_Service_prjconf#FileProvides

As many images are build in the buildservice this would mean an addition FileProvides setup in every project building a live image. If not specified a potential error condition can not be covered by kiwi because it is expected that requires are taken into account which would not be the case

Thus I used a package requires instead of a file requires which I hope makes sense too

DimStar77:
OBS 'CAN' use file dependencies, but they need to be set up in prjconf

```osc meta prjconf openSUSE:Factory | grep -i fileprov```

PR #505, opened 2017-09-20
Checkmedia results and plymouth

schaefi:
Improve display of runMediaCheck results
    
The splash screen should be switched off in order to let the
user see the mediacheck results as well as a delay timeout
before the boot continues or stops is useful

### DIFF
--- a/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
@@ -120,20 +120,62 @@ function runMediaCheck {
     # as part of the pre-mount hook via the dracut-pre-mount.service
     # which only shows messages on stderr to the console and we want
     # to see the check results during boot
-    type plymouth &>/dev/null && \
-        plymouth --hide-splash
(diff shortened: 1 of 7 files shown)


DimStar77:
From reading the PR (no media tested with it yet) this seems to do exactly what I'd like it to do

schaefi:
ok, thanks. @davidcassany would you mind to double check my code changes ? I have done some tests but another user experience would be good. Thanks

schaefi:
I could not resist to make this nicer and use a dialog based session called by systemd now. plymouth got disabled for mediacheck now as ist stands in our way when we want to display information

second review is appreciated :)

Conan-Kudo:
@schaefi As an aside though, you could have actually had all of this rendered through Plymouth. It's possible to at least print the messages through Plymouth, rather than having to disable the splash entirely. As for the dialog box, I *think* there's a way, but I'm not sure.

schaefi:
I tried it but had some trouble, especially if there is no framebuffer which could happen. I also had trouble with interaction, all input is taken by plymouth and I could not navigate the dialog it offers. I did not want to dive deeper and just used a safe default with console tools

PR #511, opened 2017-09-27
Move depmod into baseUpdateModuleDependencies

### LINKED ISSUE #508: Run depmod after stripping kernel modules not before
## Problem description
When using `<drivers>` or  `<strip type="delete">` to remove kernel modules, `depmod` is not ran afterwards, which makes for example modprobe believe these modules still exists, which causes it to try to load them at boot.

`depmod` is executed in `baseCreateCommonKernelFile`:https://github.com/SUSE/kiwi/blob/336429e7ccee56fd019557a265e260a8fc919ea8/kiwi/config/functions.sh#L1188
However that happens *before* the stripping functions here:
https://github.com/SUSE/kiwi/blob/336429e7ccee56fd019557a265e260a8fc919ea8/kiwi/config/functions.sh#L1261-L1274

Current behavior:
kiwi-depmod1

## Intended behaviour
When `depmod` is ran after deleting the files;
kiwi-depmod2

1. Create kiwi image with `<drivers>` section, for example:
```xml
<drivers>
  <file name="drivers/acpi/processor.ko"/>
  <file name="drivers/edac/edac_core.ko"/>
  <file name="drivers/mfd/mfd-core.ko"/>
  <file name="kernel/arch/x86/kvm/kvm.ko"/>
</drivers>  
``` 
2. boot it

schaefi:
Kernel module dependencies should be resolved after kiwi has
called all the stripping functions and not as part of the
baseCreateCommonKernelFile which runs before.
This Fixes #508

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -1135,6 +1135,21 @@ function baseFixupKernelModuleDependencies {
 }
 
 #======================================
+# baseUpdateModuleDependencies
+#--------------------------------------
+function baseUpdateModuleDependencies {
+    local kernel_dir
+    for kernel_dir in /lib/modules/*;do
+        [ -d "${kernel_dir}" ] || continue
+        local kernel_version=$(/usr/bin/basename ${kernel_dir})
+        if [ -f /boot/System.map-${kernel_version} ];then
+            /sbin/depmod -F \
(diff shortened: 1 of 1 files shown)


Jellyfrog:
Haven't seen any issues with this patch yet

PR #513, opened 2017-09-28
Allow to ignore certain isohybrid warnings

schaefi:
kiwi treates warning from isohybrid as fatal errors becuase in
most cases they are fatal. However some of them are kind of
historical and should be ignored like the one described here:

* http://www.syslinux.org/archives/2015-March/023306.html

### DIFF
--- a/kiwi/iso.py
+++ b/kiwi/iso.py
@@ -99,6 +99,11 @@ class Iso(object):
         :param string mbrid: boot record id
         :param string isofile: path to the ISO file
         """
+        ignore_errors = [
+            # we ignore this error message, for details see:
+            # http://www.syslinux.org/archives/2015-March/023306.html
+            'Warning: more than 1024 cylinders'
+        ]
         isohybrid_parameters = [
             '--offset', format(offset),
             '--id', mbrid.get_id(),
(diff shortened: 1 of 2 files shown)


kiwi/iso.py
```diff
@@ -99,6 +99,11 @@ def create_hybrid(self, offset, mbrid, isofile, efi_mode=False):
         :param string mbrid: boot record id
         :param string isofile: path to the ISO file
         """
+        ignore_errors = [
+            # we ignore this error message, for details see:
+            # http://www.syslinux.org/archives/2015-March/023306.html
+            'Warning: more than 1024 cylinders'
+        ]
```
davidcassany: I think there is no good solution for this, however I am wondering if it is safe using an exact match on the warning message, not sure how many different versions of isohybrid are we using in supported distros.

schaefi: I fully agree and also don't like it. If the match fails the warning is still treated as a fatal error which is what I'd like to achieve. I have checked the isohybrid code and this particular warning hasn't changed since ~2years. I don't expect people to massage the error message, I rather expect it to go away which would not impact the code in here.

If we find out that a more fine grained or generic match is needed we could do this in a subsequent patch.

After all I would delete such a code rather yesterday than tomorrow but I don't see another way to distinguish between warnings I consider really errors and other warnings we should not care about.

Currently my load is high but if time permits the real fix should go to syslinux upstream and a correct error handling based on the exit code of isohybrid. Assuming a patch would be accepted we still had to deal with the problem for older versions. Thus this patch here seems kind of inevitable

Thougths ?

schaefi:
Hrm there is another problem with this patch. If isohybrid complains about several issues at once it could lead to a false positive result. This is all crap.... need to thing twice about this

kiwi/iso.py
davidcassany: > Currently my load is high but if time permits the real fix should go to syslinux upstream and a correct error handling based on the exit code of isohybrid. Assuming a patch would be accepted we still had to deal with the problem for older versions. Thus this patch here seems kind of inevitable

fully agree, the _real fix_ does not prevent us of having to code some ugly stuff. Thus let's start with it, then we'll see if we have some spare cycles to provide some feedback/patch to syslinux.

schaefi:
ok 👍

PR #514, opened 2017-09-29
Changed permissions of /srv/tftpboot to be readable (bsc#940608)

DimStar77:
Changed permissions of /srv/tftpboot to be readable (bsc#940608).


schaefi:
I agree

PR #518, opened 2017-10-18
Fix vmdk descriptor file

davidcassany:
This commit fixes the descriptor file of the vmdk images. Before this
commit the descriptior file was appended at the end of the image
instead of overwritting the current one at the very beginning.

Fixes bsc#1050665

### DIFF
--- a/kiwi/storage/subformat/vmdk.py
+++ b/kiwi/storage/subformat/vmdk.py
@@ -238,7 +238,7 @@ class DiskFormatVmdk(DiskFormatBase):
         vmdk_descriptor_lines.append(
             'ddb.toolsVersion = "%s"' % vmdk_tools_version
         )
-        with open(vmdk_image_name, 'ab') as vmdk:
+        with open(vmdk_image_name, 'r+b') as vmdk:
             vmdk.seek(512, 0)
             vmdk.write(bytes('\n'.join(vmdk_descriptor_lines), 'utf-8'))
             vmdk.seek(0, 2)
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi this is a bug that was discovered yesterday by @gmoro, the open call with 'ab' opens the file in append mode, thus the descriptor file instead of being updated it gets appended at the end. For some reasos this was not failing in ESXi 5 and pre vmdk v6 images, as the descriptor file was somehow still found by vmware, but this is not the case in ESXi 6. 

You can see that our vmdk images have two descriptor files, one at the beginning (the original one created by qemu) and the updated one at the end (appended by kiwi). You can see it with:

$ tail -c 512 appliance.vmdk // will show the Descriptor File at the end
$ head -c 1024 appliance.vmdk // will show the Descriptor File at the beginning

With this patch you'll find only one Descriptor File at the beginning and with the updated data.

I am trying to replicate @gmoro's project with a patch version of kiwi so to let him test the result with ESXi 5 and ESXi 6.

Now I will go through this strange Travis failures, for some reason Travis fails doing `apt-get update -qq`, provably it is a transitory issue on Travis side.

schaefi:
I just restarted the build, maybe it was just a hickup

schaefi:
An interesting patch, I was told vmdk's expects the descriptor at the end of the image. But that might be a very old information. Thanks for fixing this

schaefi:
typo in commit message though :)

davidcassany:
Sure, that's why I am creating a project in IBS with that purpose, so @gmoro can test and provide feedback. All the credits and thanks to @gmoro about this patch, he did the hard debugging job to see what was going on.

I'll include the project here when ready.

mnowaksuse:
Tested the image on ESXi 5.5 and it boots fine. `vmkfstools -x check` says `Disk is error free`.

PR #520, opened 2017-10-19
Include build tests

davidcassany:
This PR is including the code of the Open Build Service tests.


davidcassany:
@schaefi I am blocking this PR until I have updated all the packages in OBS and verified they build as expected.

I have started with x86 builds, now the `_service` is pulling this `include_build_tests` branch, once merged updating the `_service` to pull master branch will be needed. I this way we can verify nothing strange happens before merging.

schaefi:
sounds great :+1:

schaefi:
Add some changes to the branch

* I have merged master to this branch to make travis happy again due to the patch fixing the exception qualifier.

* I also moved the test-image-docker build test to use the source service

* I updated the readme to be in rest format

schaefi:
So at the time we merge this the integration build image description are no longer on obs but in the kiwi git. How would the process look like if we change the description in git

* does obs automatically fetch the changes ?
* if we change the description in a branch and use the standard pull-request workflow, obs will recognize the changes first after merge to master because the _service files will look to the master branch only. Thus we get notified that the build failed when we already merged it

This integration with obs is not quite clear to me

Another scenario is that the build fails for some reason in obs without us having changed anything.
In order to find out why the build fails I usually end up in a try and error game on the obs side. If we need to commit and push any change now in git to test and see when it starts building on obs again we probably would create a lot of "waste" commits in the master branch of the kiwi git.

Thoughts ?

schaefi:
All builds are passing (except s390 in blocked state, but that's not something which should hold us back)

```
make obs_test_status 
./.obs_test_status
Virtualization:Appliances:Images:Testing_x86/test-image-oem images x86_64:
--> success
Virtualization:Appliances:Images:Testing_x86/test-image-iso images x86_64:
--> success
Virtualization:Appliances:Images:Testing_x86/test-image-vmx images x86_64:
--> success
Virtualization:Appliances:Images:Testing_x86/test-image-pxe images x86_64:
--> success
Virtualization:Appliances:Images:Testing_x86/test-image-docker images x86_64:
--> success
Virtualization:Appliances:Images:Testing_x86/test-image-orthos-oem images x86_64:
--> success
Virtualization:Appliances:Images:Testing_x86/test-image-ec2 images x86_64:
--> success
Virtualization:Appliances:Images:Testing_x86/test-image-azure images x86_64:
--> success
Virtualization:Appliances:Images:Testing_x86/test-image-fedora-vmx images x86_64:
--> success
Virtualization:Appliances:Images:Testing_arm/test-image-panda-oem images armv7l:
--> success
Virtualization:Appliances:Images:Testing_s390/test-image-oem images s390x:
--> pending_build: want: 9.11.16 got: 
Virtualization:Appliances:Images/dice-image-docker images x86_64:
--> success
Virtualization:Appliances:Images/dice-image-vagrant-libvirt images x86_64:
--> success

```

schaefi:
So from my point of view we could merge this and change the source service to point to master

However we should be clear how our maintenance workflow will be, see my questions

davidcassany:
> * does obs automatically fetch the changes ?

No, it doesn't, provably we could add a github webhook that follows this [BS docs](http://openbuildservice.org/help/manuals/obs-reference-guide/cha.obs.authorization.token.html#idm140050581203296).

> * if we change the description in a branch and use the standard pull-request workflow, obs will recognize the changes first after merge to master because the _service files will look to the master branch only. Thus we get notified that the build failed when we already merged it.

Yes, to that regard we do not have any gain here, this is the same situation we had. I believe the only way to solve it would be having an staging branch in KIWI, and I am not entirely sure this is what we want, as the python-kiwi package used to build the build tests is not the one in master. I understand your concerns and see the confusion that this introduces.

> This integration with obs is not quite clear to me

To my eyes, the main issue is that the kiwi descriptions in build-tests used in Testing projects might not use the same commit that the one used in python-kiwi package that will be used to build the testing images. I am thinking if it would be possible to have some kind 'latest' tag that is always pointing to the last tagged version, so this is what would be used in _service instead of master HEAD. This, at least, would allow us to keep the sources used in testing projects consistent. Not sure if this is something doable without having to make a "git push --force" for every single version bump. If we go this way, then our gain would be keeping testing projects sources and kiwi sources in line in a single repository. But no gain or change in the current update workflow, because as you say we would not be capable to detect failures before merging, the way we aren't capable right now.

Another option would be to go for a more sophisticated jenkins job that does several things, like building a new package for every merge in master (or an hypothetical staging branch), trigger services in Testing_xxx and monitor the overall result. I don't think it would be very complex, but I am not really a fan of this kind of solutions as I believe this kind of stuff tends to grow and become hard to maintain; even though I would not entirely discard it.

schaefi:
I think I looked on this from a different angle. You are right that if we change the kiwi code in the git this is not directly the code which builds the testing image. But that's something I can live with, we do that already as you stated. So it's ok for me to create the release submit the package to Virt:App:Staging and wait for the test builds to finish. The test builds uses the python-kiwi package from Staging, if the builds are successful we submit to Virt:App:Builder and from there to anywhere else. So that's ok for me

I thought more of the fact that the very same test build fails due to changes which are not related to the kiwi code in git. For example the tests are based on tumbleweed, that distro changes and for whatever reason it might be needed that the test image description needs another <package> line. In the old setup I just did that in obs and checked in the changes. In many cases a series of check ins were needed before it builds again. Now with the new setup that change needs to be in git, then obs needs to get triggered to use the new sources (we definitly have to setup the webhook because I don't want to do that by hand all the time) and now assume the build still fails, next round starts and we are forced to push the change to master because otherwise obs doesn't see it. That's what I meant we would taint master with commits like "fix test build", "again fixed test build" "fixed test build again"... it could be like that.

So in case the test build fails and it is not kiwi's fault how should we do it ?

I think we could:

* create a branch
* move the source service temporarily to that branch
* do the changes in the branch until the build works again
* rebase branch to one useful commit and message
* merge the branch
* redirect the source service to master

and in that sequence obs should always use the latest version of the branch which I don't know how we could setup

Does this make sense or is my thinking wrong ? I'm of course open for any better solution

Thoughts ?

davidcassany:
Oh, you are right, now I understand your concern. Sure we don't want trial-error like commits in git master branch history. The sequence is fine for me, makes sense. The only annoying thing is having to modify `_services` file, but I hope it should not happen often and if multiple images are broken at the same time they are likely to be broken by the same issue.

> and in that sequence obs should always use the latest version of the branch which I don't know how we could setup.

I think this is really the next step, I'll try to verify the automation between github and OBS is possible. I just skimmed the docs and looks like it is, but need to verify.

schaefi:
> The only annoying thing is having to modify _services

yes but I don't see another way

ok so you are working on the webhook part ? where do we stand so far ?

davidcassany:
@schaefi I have been able to successfully automate the webhook, OBS is already a service supported or included in github webhook settings, so it is easy to enable the webhook, the issue I encountered so far is that is is designed to be applied to certain branches. Basically we can set to be applied on commits matching a comma separated list of references (i.e. `ref/heads/master,ref/tags/v*`), so I believe that it could be annoying to use a different branch every time we want to update the tests, since we would need to change `_service` file for every OBS package we want to test and also update the github service accordingly... I am not very satisfied with this.

I have set up a kiwi fork in [davidcassany/kiwi](https://github.com/davidcassany/kiwi) to test this service (I think only the repo administrator can do it) and verified that every commit to `ref/heads/tests` triggers the _service of `test-image-oem` package (which I have updated to pull from my fork). This is fine, but about branches management I do not see a clear solution, provably we can just assume we use always the same branch name to make test updates, but this is odd and we could forget. Also keeping a living branch just for that is quite annoying as then we should keep it in sync with master manually.

Provably another option would be to make all the tests locally disabling the service and once solved and after merging just renable the service, but I'd prefer to let the builds happen on remote env.

Any further idea?

schaefi:
Hmm, ok that sounds awkward. In this case I suggest another workflow when we have to update the test image descriptions for some reason. So if we see a test build fails and it is not because of a change in kiwi but for some other reason I suggest the following procedure:

1. Drop service

    osc service merge

2. Fix the build

    ... whatever is needed to let it build again

3. Build fixed ok, now manually apply the current obs patch to a branch in git
    and create a pull request

4. Merge pull request

5. Restore source service

     osc up -rX

     with X = release before osc service merge was called
   
     mv _service /tmp
     osc rm *
     mv /tmp/_service .
     osc add _service
     osc ci

     Comment like: "Fixed xy in kiwi git commit: id, restored source service"

It's still a not so nice process, especially the restore of the source service is awkward

Thougths ?

davidcassany:
I am fine with your proposal, provably we could disable the service instead of merging it in order to avoid having to restore the service from an old version.

1. Drop service
   ... include `mode="disabled"` in _service
   osc commit

2. Fix the build
   osc service disabledrun
   ... whatever is needed to let it build again
   
3. Build fixed ok, now manually apply the current obs patch to a branch in git
and create a pull request

4. Merge pull request

5. Restore source service
   ... remove `mode="disabled"` from _service
   osc commit

Anyway in general terms it is the same you suggested: unsync the code, develop the patch out of sync, commit and resync. The main idea is to not include into the repository the development process for a test fix, how the _service file is managed to disable it is minor thing.

I'd say we are fine to proceed, aren't we?

schaefi:
Thanks for the update, yes disabledrun is nice

> I'd say we are fine to proceed, aren't we?

yes let's merge this. I leave it up to you to merge since I guess you will adapt the _service files too

davidcassany:
Done, in case we want the images to build for each commit in master we need to add a webhook (Settings -> Integrations & services -> Add services -> Obs) for each package. 

However given the workflow we discussed I believe it doesn't have much sense to add a webhook as it would not have impact in the KIWI used for the build and the descriptions shouldn't be updated unless there are issues with the test descriptions.

schaefi:
agreed, thanks great job

PR #522, opened 2017-10-23
Add OCI reference in skopeo copy call and umoci config call

davidcassany:
Skopeo, since v1.24, does no longer assume 'latest' as the default
tag/reference and requires explicit tag or reference in skopeo
call. In KIWI the default was only used to import the base rootfs,
with this commit the imported container is tagged as 'base_layer'.
The current patch works for all skopeo versions.


schaefi:
looks great 👍

PR #525, opened 2017-10-24
Implement repository priorities for apt-get

michal42:
Translate the repository priorities in the XML into pin-priorities in
the apt preferences file. This is not perfect when working with OBS
repositories, because these are all hosted at the same site, but it's a
good start.

### DIFF
--- a/kiwi/repository/apt.py
+++ b/kiwi/repository/apt.py
@@ -148,6 +148,9 @@ class RepositoryApt(RepositoryBase):
         list_file = '/'.join(
             [self.shared_apt_get_dir['sources-dir'], name + '.list']
         )
+        pref_file = '/'.join(
+            [self.shared_apt_get_dir['preferences-dir'], name + '.pref']
+        )
         self.repo_names.append(name + '.list')
         if os.path.exists(uri):
             # apt-get requires local paths to take the file: type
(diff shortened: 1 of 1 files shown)


michal42:
Ugh, are you seriously asking people to write test coverage for every single line of code added?

schaefi:
Thanks much for working on the apt priorities, we didn't get to that point so far.

> Ugh, are you seriously asking people to write test coverage for every single line of code added?

I'm afraid this is the case. There is no need to really test every line of code especially for methods considered private but a test should have run the code and tested the expected results. This gives you full freedom to do changes in the future which is why we work that way

Hope that's not too bothersome ?

michal42:
> Hope that's not too bothersome ?

It actually is, because it requires me to learn and setup tox and there is little reward for writing the test.

schaefi:
you mean py.test right ? because tox is just like make

if I read your code correctly you are creating a "preferences-dir/repo.pref" file which contains a prio setup matching all packages according to the information provided in the uri. If you can give us an
example uri and the expected .pref file I can offer to write a test for it

However I think the investment in py.test would be worth it, it's used in many python projects

Thoughts ?

schaefi:
Setting up the environment to let you run the tests should be done quickly

```
tox -e 3.4
source .tox/3.4/bin/activate
cd test/unit
py.test <some-test>

deactivate
```

schaefi:
We documented that here:

* http://suse.github.io/kiwi/development.html#create-a-python-virtual-development-environment

The first paragraph lists the needed development packages, but you probably already have them installed on your machine. Just in case the tox setup failed it might be due to missing devel packages

michal42:
> if I read your code correctly you are creating a "preferences-dir/repo.pref" file which contains a prio
> setup matching all packages according to the information provided in the uri. If you can give us an
> example uri and the expected .pref file I can offer to write a test for it

Thanks. The preferences file should look like this for various URIs:

"file://srv/my-repo"
Package: *
Pin: origin ""
Pin-Priority: <priority from XML config>

"http://download.opensuse.org/repositories/Virtualization:/Appliances:/CommonBoot/Debian_9.0/"
Package: *
Pin: origin "download.opensuse.org"
Pin-Priority: <priority from XML config>

"http://example.org:8080/myrepo"
Package: *
Pin: origin "example.org"
Pin-Priority: <priority from XML config>

Basically it's the host name except for local repositories, in which case origin is an empty string.

schaefi:
Ok thanks for the information. Here is my version of the patch

* https://github.com/SUSE/kiwi/pull/533

Would be great if you can review the code, especially in terms of the expected results in the tests

Thanks

PR #531, opened 2017-10-31
Add mount hooks

Jellyfrog:
Fixes # .

* Adds mount hook
* Adds missing hooks to the other architectures


schaefi:
ok, let's merge this. Please be aware that when we move the oem boot to be driven by dracut those hooks will become obsolete because dracut implements their own workflow which is documented here:

* http://man7.org/linux/man-pages/man7/dracut.bootup.7.html

Just to make you aware I plan to support custom kiwi oemboot for some time after the move to dracut but it will go away from the maintained area of the kiwi git at some point in the not so distant future

PR #541, opened 2017-11-06
Remove hardcoded any /var/lib/rpm path

### LINKED ISSUE #537: Fail to build live ISO with rpm and new RPMDB location
RPM was changed the location for the rpmdb, out of  /var/lib to /usr/lib; this change is also in accordance with upstream.

Since this change, though, we are not able to produce live images, since the kiwi build process fails, as can be seen at:

https://build.opensuse.org/package/live_build_log/openSUSE:Factory:Live/livecd-gnome/images/x86_64

I am not (yet) sure if this is a pure kiwi issue or if there is also an rpm component to this bug (but would definitively appreciate the kiwi side of things to be validated)

davidcassany:
This commit changes any /var/lib/rpm reference to the call
'rpm -E %_dbpath' which returns the path of the rpmdb which
has been recently updated to a different location in recent rpm
versions. Now the rpmdb path is determined dynamically.

Fixes #537

### DIFF
--- a/kiwi/package_manager/zypper.py
+++ b/kiwi/package_manager/zypper.py
@@ -252,11 +252,15 @@ class PackageManagerZypper(PackageManagerBase):
             48: 'db48_load'
         }
 
-        cmd = Command.run(['chroot', self.root_dir, 'rpm', '-E', '%_dbpath'])
-        rpmdb = os.sep.join([self.root_dir, cmd.output.strip()])
-        if not os.path.exists(rpmdb):
+        rpmdb_path_request = Command.run(
+            ['chroot', self.root_dir, 'rpm', '-E', '%_dbpath']
+        )
+        rpmdb_path = os.path.normpath(os.sep.join(
+            [self.root_dir, rpmdb_path_request.output.strip()]
+        ))
+        if not os.path.exists(rpmdb_path):
             raise KiwiRpmDatabaseReloadError(
-                'Unable to get the rpmdb location {0}'.format(rpmdb)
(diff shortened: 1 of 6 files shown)


kiwi/package_manager/zypper.py
```diff
@@ -251,18 +251,26 @@ def dump_reload_package_database(self, version=45):
             45: 'db45_load',
             48: 'db48_load'
         }
+
+        cmd = Command.run(['chroot', self.root_dir, 'rpm', '-E', '%_dbpath'])
+        rpmdb = os.sep.join([self.root_dir, cmd.output.strip()])
```
schaefi: Can we make the variable names a bit more self explaining, e.g

```python
rpm_dbpath_request = Command.run(['chroot', self.root_dir, 'rpm', '-E', '%_dbpath'])
rpmdb_path = os.sep.join([self.root_dir, rpm_dbpath_request.output.strip()])
```

msrex:
@schaefi @davidcassany I saw the code in https://github.com/SUSE/kiwi/pull/541/commits/63faa2a2ff984afbe70f42d74ba6fbcd2ac0d71b change from querying `rpm` per @DimStar77, and instead just check whether the directory `/usr/lib/rpm` exists. On my Fedora 25/26 and Centos 6/7 test systems, this directory always exists with a bunch of rpm-related tools, while the actual database is located in the traditional place in `/var/lib/rpm`. The new code is going to always "detect" the rpmdb location in the wrong spot.

Was there a specific reason to move away from querying `rpm` and move to yet another hardcoded path?

PR #545, opened 2017-11-15
Dracut kiwi lib

schaefi:
Added dracut kiwi oem module and library
    
A new dracut module kiwi-repart used to be the successor of
the custom kiwi oemboot code to repartition the disk has
been added. Along with the module a dracut library kiwi-lib
will also be delivered.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_install:
 - sudo apt-get install -y xsltproc
 - sudo apt-get install -y genisoimage
 - sudo apt-get install -y enchant
+- sudo apt-get install -y shellcheck
 install:
 - pip install --upgrade pip
 - pip install -r .travis.requirements.txt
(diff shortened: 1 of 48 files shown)


schaefi:
This is basically the result of the hackweek and provides a good part of the oemboot code as dracut modules. In order to make use of it in an oem image one needs to specify

```xml
<type image="oem" .... initrd_system="dracut"/>
```

By default kiwi's own oemboot code is still active to prevent any breakage of existing descriptions.
All new shell code is covered by shellcheck which improves shell code quality a bit

Next steps after merge are:

* move oem deployment code used for installation images (installiso/installstick) to its own dracut module
* make use of dracut for oem images the default like it is already the case for iso and vmx
* move custom oemboot descriptions away from kiwi and into the kiwi-descriptions as a fallback method
* move on to let netboot play with dracut

schaefi:
As this is a pretty large patch I'm leaning towards to merge it, create a package in staging, check if all integration tests still build and let you play with the result.

The scope of oemboot is pretty large, thus we will have both boot systems in place for some time

Please let me know if there are any objections or things I might have overlooked

Thanks

PR #547, opened 2017-11-16
Removes the hardcoded path of the rpm database

### LINKED ISSUE #537: Fail to build live ISO with rpm and new RPMDB location

davidcassany:
With this commit the rpmdb path is evaluated from the %_dbpath
macro instead of being hardcoded.

Fixes #537

### DIFF
--- a/kiwi/boot/arch/arm/oemboot/linuxrc
+++ b/kiwi/boot/arch/arm/oemboot/linuxrc
@@ -418,7 +418,7 @@ if [ ! -z "$KIWI_RECOVERY" ];then
                 Echo "Failed to backup zypp database"
                 exit 1
             fi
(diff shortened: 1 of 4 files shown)


Conan-Kudo:
LGTM too

PR #548, opened 2017-11-17
Add max size post build command

### LINKED ISSUE #468: Add post build constraints
## Problem description

A discussion by mail brought up the idea to allow to setup general post build conditions in the kiwi config file like:

```yaml
build_constraints:
  - max_size: 2g
```

which causes kiwi to fail if one of the conditions is not fulfilled

## Actual behaviour

post build constraints does not exist

davidcassany:
This commit includes a post build constraint validation into the
Result class. With this commit any file included to the result
tuple is verified not to exceed the maximum size configured into
the runtime configuration file config.yml.

Fixes #468

### DIFF
--- a/kiwi/system/result.py
+++ b/kiwi/system/result.py
@@ -21,6 +21,9 @@ import os
 
 # project
 from kiwi.logger import log
+from kiwi.runtime_config import RuntimeConfig
+from kiwi.utils.size import StringToSize
+
 
 from kiwi.exceptions import (
     KiwiResultError
@@ -56,6 +59,7 @@ class Result(object):
         self.class_version = 1
 
         self.xml_state = xml_state
+        self.runtime_config = RuntimeConfig()
 
     def add(
         self, key, filename, use_for_bundle=True, compress=False, shasum=True
@@ -70,6 +74,7 @@ class Result(object):
         :param bool shasum: create shasum when bundling true|false
         """
         if key and filename:
+            self._verify_build_constraints(filename, key)
             self.result_files[key] = result_file_type(
                 filename=filename,
                 use_for_bundle=use_for_bundle,
@@ -124,3 +129,13 @@ class Result(object):
             raise KiwiResultError(
                 'Failed to pickle load results: %s' % type(e).__name__
             )
+
+    def _verify_build_constraints(self, filename, key):
+        max_size = self.runtime_config.get_max_size_constraint()
+        if max_size:
+            max_size_bytes = StringToSize.to_bytes(max_size)
+            if os.path.getsize(filename) > max_size_bytes:
+                raise KiwiResultError(
+                    'Build constraint failed: {0} {1} is bigger than {2}'
+                    .format(key, filename, max_size)
+                )
--- a/kiwi/system/result.py
+++ b/kiwi/system/result.py
@@ -21,9 +21,6 @@ import os
 
 # project
 from kiwi.logger import log
-from kiwi.runtime_config import RuntimeConfig
-from kiwi.utils.size import StringToSize
-
 
 from kiwi.exceptions import (
     KiwiResultError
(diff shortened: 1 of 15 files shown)


schaefi:
great coding and nice refactoring included 👍

schaefi:
you added the wip label, so you plan additional code ? imho it looks ready to be merged

davidcassany:
@schaefi, yes I was planning to refactor a bit more to include the `image_size` as a `kiwi/system/result` attribute (validating a unique value for each image type). More in line of what was discussed in the issue #468, this way each `kiwi/builder/*` would set the `image_size` and also add a `verify_constraints` method in `kiwi/system/result` assuming there could be further constraints defined in the future.

This is what I tried today morning but I doubt it really makes sense since I realized that for that I need to make a new version of `kiwi/system/result` (need to verify how to handle multiple versions with pickle) and the calculation of the size is the same for all (and could not manage to get a _docker like_ size for containers).

Also another question came to my mind, how to handle the max_size constraint with 'installiso=true'? it could happen that the you place a constraint for a CD (700Mb) but the raw image file is bigger then code of this PR would fail.

I think we could leave it as it is and do not expect any other requirement in that area at mid-term as the alternative is getting a bit too wide. Do you agree?

**off-topic:** I noticed since a while I do not get email notifications from github when someone writes in the issues I am assigned or I participate, does it also happens to you? I just read your comments now because I wanted to check your other PR and I saw there was a new comment, but I did not get any notification 😢

kiwi/system/result.py
```diff
@@ -70,6 +74,7 @@ def add(
         :param bool shasum: create shasum when bundling true|false
         """
         if key and filename:
+            self._verify_build_constraints(filename, key)
```
schaefi: Oh, thanks David for pointing this out. I was too fast in my review here. So you are right this implementation will add the size constraint check for any result.add() call. The problem here this would be out of context. Thus I suggest you move _verify_build_constraints() into its own public method of the Result class and give it a meaning, e.g

```python
def verify_image_size(filename, size_limit):
     # compare size_limit with the given filename, raise on size violation
```

The actual call of the method should be done with a context. Meaning you need to look at kiwi/builder and there to all calls of ```self.result.add(...)``` You will find e.g kiwi/builder/container.py

```
self.result.add(
            key='container',
            filename=self.filename,
            use_for_bundle=True,
            compress=False,
            shasum=True
        )
```

For this result file we want to run the size checker. Thus an additional line for every result where we want to run the check is needed. The reading of the correct size constraint from the config section also imho belongs to where the context is. That way we can validate different sizes at the place in the code where all information is present

```python
self.result.verify_image_size(self.filename, self.runtime_config.get_container_image_size_constraint())
```

if the size check is not enabled in the config file, this validation always passes. I think we would need
extra size constraint methods in the config for

* container
* disk
* filesystem
* archive
* livemedia
* installmedia

That should cover all we can build, and in a first shot this does not have to be complete

In order to make the size information part of the result type we need to add a field in the ```result_file_type```
tuple and allow to set that field in the add() method. This imho can be done in a subsequent pull request

Thoughts ?

schaefi:
I still get the e-mail notifications, you can check which e-mail is registered on github and send a test mail to this address to see if you receive your own posting

davidcassany:
OK agree, looks good to me, this is quite close to what I already have.

schaefi:
ok this looks good to me now

* for validation of install iso sizes we could provide another option in build_constraints  e.g max_install_image_size. But let us do that on request of our users

* adding the size into results could also be done in a new pull request. But I suggest to that also on request of our users

Thoughts ?

davidcassany:
Yes I agree we should wait users to start using the feature and see what actually they expect or if they need something more elaborated. Now fixed the import for the python2, so Travis is not failing anymore.

I'd say we are ready to go

schaefi:
yay :) I'll merge

PR #549, opened 2017-11-20
Update gfxboot.cfg for iso images

schaefi:
Don't move down one menu entry the first time a F-key is used
This Fixes bsc#1068790


schaefi:
@wfeldt I'd like to use the command to change the config file, it seems more robust to me. This gfxboot call is not a suse only extension correct ? meaning it will also work on other distros ?

Thanks

schaefi:
verified this to be working in a test build

PR #551, opened 2017-11-21
package: Prepare for Tumbleweed moving to suse_version 1550

DimStar77:
if suse_version is >= 1330, it can be either TW, Leap15 or SLE15 (and future products)

but both, Leap and SLE have the variable sle_version defined which does not exist in TW (TW is not based on SLE) - thus, suse_version >= 1330 and no sle_version implies TW


schaefi:
Thanks much for taking care on that, I'm always lost in this regard :)

PR #555, opened 2017-11-28
Adding --ingore-rpm-postscript-errors flag

### LINKED ISSUE #539: Capture rpm post script failures
## Problem description

Vladimir Nadvornik reported:

> JFYI, there is a recent change in zypper which now reports exit code 107
> if post script fails. Before it ignored such errors.
> Maybe Kiwi should ignore this exit code to get the old behavior.

I think it would make sense to have a possibility in kiwi by a commandline switch to
ignore rpm post script errors.

davidcassany:
This commit adds the --ignore-rpm-postscript-errors flag,
which enables to omit a failure when zypper returns 107 exit code.
In former zypper versions %post install script failures zypper
was not returning any error code, thus this flag allows to install
packages with flaky %post scripts as in previous zypper versions.

Fixes #539

### DIFF
--- a/kiwi/command.py
+++ b/kiwi/command.py
@@ -113,7 +113,7 @@ class Command(object):
         )
 
     @classmethod
-    def call(self, command, custom_env=None):
+    def call(self, command, custom_env=None, return_code_handler=None):
         """
         Execute a program and return an io file handle pair back.
         stdout and stderr are both on different channels. The caller
@@ -179,7 +179,7 @@ class Command(object):
             'command', [
                 'output', 'output_available',
                 'error', 'error_available',
-                'process'
+                'process', 'return_code_handler'
             ]
         )
         return command(
@@ -187,5 +187,6 @@ class Command(object):
             output_available=output_available(),
             error=process.stderr,
             error_available=error_available(),
-            process=process
+            process=process,
+            return_code_handler=return_code_handler
         )
--- a/kiwi/command_process.py
+++ b/kiwi/command_process.py
@@ -195,6 +195,10 @@ class CommandIterator(six.Iterator):
         :return: errorcode
         :rtype: int
         """
+        if self.command.return_code_handler:
+            return self.command.return_code_handler(
+                self.command.process.returncode
+            )
         return self.command.process.returncode
 
     def get_pid(self):
--- a/kiwi/package_manager/base.py
+++ b/kiwi/package_manager/base.py
@@ -203,6 +203,17 @@ class PackageManagerBase(object):
         """
         raise NotImplementedError
 
+    def ignore_postscript_errors(self):
+        """
+        Ignores the %post install script errors of rpm packages.
+
+        Implementation in specialized package manager class
+        """
+        raise NotImplementedError(
+            'Flag --ingore-rpm-postscript-error is currently only supported '
+            'for zypper package manager'
+        )
+
     def cleanup_requests(self):
         """
         Cleanup request queues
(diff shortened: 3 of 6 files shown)


davidcassany:
@schaefi, I have done this PR since I wanted to share the code before continuing the development. Up to now only `system prepare` task is updated with the `--ignore-rpm-postscript-errors` and the unit tests are not in place. 

I doubt to continue since I have the feeling that the implementation is a bit too hacky for a little gain. I do note really like to set some handler function in `kiwi/command.call` that is then evaluated in `kiwi/command_process.get_error_code` I feel this is quite cryptic and unnatural. Also, another reality that makes me doubt to adopt this new flag is that my tumbleweed test that used to fail it doesn't fail anymore, I guess some package got fixed, thus I am wondering if shouldn't we wait a little more to see if it really causes critic issues in order build images; after all, #539 only unveils flaky %post install scripts IMHO which is not really a kiwi issue.

Ideas?

schaefi:
> after all, #539 only unveils flaky %post install scripts IMHO which is not really a kiwi issue.

I agree completely

If dealing with this would be low hanging fruits for us I would say ok let's allow an option to ignore the error. But there is not only "the" error and that will end up in a guessing game

Another thing which came to my mind was, if the image builds in the buildservice and has this issue it would not be easy to just pass an option to the kiwi call made by obs. Those who have access to the build system and the builder could just temporary add

```
raise_on_error=False
```

in the kiwi code at kiwi/package_manager/zypper.py and hope that it leads to a working system

Last but not least I think we should not ignore errors from package installs, especially not if they are caused by broken packages. I think if we start to add support for that it will be abused

I'm on your side with the concerns and I think it would be ok to close the issue including the details we have found.

What do you think ?

davidcassany:
Yes I agree, we should not start supporting hacks to work around the issues caused by broken packages, should be other way round, it should put pressure to fix packages and help having better repositories.

About OBS, well, that's why initially in the comments I did to #539 I was also considering to instead of making a flag adding a runtime configuration value, this way, even not being immediate, a user could provide it in OBS without having to touch anything from KIWI or OBS. But I still believe it's not worth. 

About 

```
raise_on_error=False
```

I don't think it would solve the issue, unfortunately, because the `kiwi/packager_manager/zypper.py` does not make use of `kiwi/command.run`, it uses `kiwi/command.call` (in order to be able to show the progress bar). **Command.call** does not have the option of `raise_on_error` (normal we do not have any threading, thus there is no place in `kiwi/command.py` where the flag could be evaluated). This is the main reason of my complaints about not being a low hanging fruit, as the return code is not evaluated in `kiwi/command.py`. Provably this could be done if `kiwi/command.call` would use a thread to control the subprocess but I doubt it makes sense, I like the _keep it simple_ philosophy.

My overall opinion is that there should not be a flag to ignore package errors. About how to handle the potential issues derived from this zypper tool change I guess the right way to proceed would be by using the `--allow-existing-root`, during the second attempt to install the rootfs should not fail, as no package install should be needed. No clues on how to handle the issue in OBS.

schaefi:
> raise_on_error=False

right this was too short thinking, looking closer to my own code you are right :) for the call interface one would need to patch `command_process.py::get_error_code()`

> --allow-existing-root

that works for people who can access the building environment. I don't think it can be managed in obs without more ugly hacks

So yes all in all this will create only trouble on our side which I agree is not worth the effort when people should create good packages instead of counting on us to provide a "workaround api"

davidcassany:
Right, then I am closing the PR and the issue without patching anything.

PR #558, opened 2017-12-01
add missing deps for docker builds.

adrianschroeter:
(the other kiwi-image: provides should also get moved to -requires package? not tested...)

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -129,12 +129,6 @@ and cloud systems like Xen, KVM, VMware, EC2 and more.
 Summary:        KIWI - Appliance Builder Next Generation
 Group:          Development/Languages/Python
 Provides:       python-kiwi = %{version}-%{release}
-Provides:       kiwi-image:tbz
-Provides:       kiwi-image:docker
-Provides:       kiwi-image:iso
-Provides:       kiwi-image:vmx
-Provides:       kiwi-image:pxe
-Provides:       kiwi-image:oem
 %if 0%{?fedora} || 0%{?suse_version}
 Recommends:     jing
 %endif
@@ -206,12 +200,6 @@ virtualization and cloud systems like Xen, KVM, VMware, EC2 and more.
 %package -n python3-kiwi
 Summary:        KIWI - Appliance Builder Next Generation
 Group:          Development/Languages/Python
-Provides:       kiwi-image:tbz
-Provides:       kiwi-image:docker
-Provides:       kiwi-image:iso
-Provides:       kiwi-image:vmx
-Provides:       kiwi-image:pxe
-Provides:       kiwi-image:oem
 Recommends:     jing
 Requires:       python3-PyYAML
 Requires:       python3-docopt
@@ -390,6 +378,12 @@ overlay root filesystem
 Summary:        KIWI - buildservice package requirements for boot images
 Provides:       kiwi-boot:netboot
 Provides:       kiwi-boot:oemboot
+Provides:       kiwi-image:docker
+Provides:       kiwi-image:tbz
+Provides:       kiwi-image:iso
+Provides:       kiwi-image:vmx
+Provides:       kiwi-image:pxe
+Provides:       kiwi-image:oem
 Provides:       kiwi-filesystem:btrfs
 Provides:       kiwi-filesystem:xfs
 Provides:       kiwi-filesystem:ext3
@@ -397,6 +391,8 @@ Provides:       kiwi-filesystem:ext4
 Provides:       kiwi-filesystem:squashfs
 Requires:       e2fsprogs
 Requires:       xfsprogs
+Requires:       umoci
+Requires:       skopeo
 %if 0%{?fedora} || 0%{?rhel}
 Requires:       btrfs-progs
 %else


package/python-kiwi-spec-template
```diff
@@ -383,13 +381,16 @@ overlay root filesystem
 Summary:        KIWI - buildservice package requirements for boot images
 Provides:       kiwi-boot:netboot
 Provides:       kiwi-boot:oemboot
+Provides:       kiwi-image:docker
```
schaefi: Sorry that doesn't make sense to me. The kiwi-image:`image-type-name` provides indicates that this package provides the image type `image-type-name`. Support for the image types is provided by the python-kiwi base package aka python2-kiwi and python3-kiwi but not by the package kiwi-boot-requires where you would like to move it. Could you explain why this should be correct ?

schaefi: Even more confusing is that the docker type doesn't need a boot image at all which means for building docker images obs doesn't need kiwi-boot-requires at all. If the provides tag which indicates this kiwi can build docker images is now moved into a package which is not needed to build docker images we will completely loose the information. That can't be the intention, or did I misunderstood the concept ?

package/python-kiwi-spec-template
```diff
@@ -383,13 +381,16 @@ overlay root filesystem
...
 Provides:       kiwi-filesystem:btrfs
 Provides:       kiwi-filesystem:xfs
 Provides:       kiwi-filesystem:ext3
 Provides:       kiwi-filesystem:ext4
 Provides:       kiwi-filesystem:squashfs
 Requires:       e2fsprogs
 Requires:       xfsprogs
+Requires:       umoci
+Requires:       skopeo
```
schaefi: umoci and skopeo are only required when a docker image will be build. I'm ok adding this requires to the kiwi-boot-requires package but it's imho the wrong place with regards to my comments before. You should also keep in mind that kiwi-boot-requires will be deleted as soon as I have moved all custom kiwi boot code to dracut because at that point no kiwi boot descriptions will be existing anymore

Thus I'd prefer to only have provides, requires information for kiwi-boot-requires which are correct to the context of the package and will not cause trouble when this package goes away

schaefi:
any news to clarify this one ?

package/python-kiwi-spec-template
adrianschroeter: it indicates that it provides all tools and therefore all it dependencies to be able to build that image.

Currently docker builds are failing due to missing tools and despite the fact that this provider is installed.

package/python-kiwi-spec-template
adrianschroeter: And who shall require then these packages for docker builds?

schaefi: That's what I meant. kiwi-boot-requires will go away in the not too distant future (as soon as I have moved all kiwi boot code to dracut). I had planned that the python-kiwi package keeps the provides tags e.g "kiwi-image:docker". obs would be able to read that tag and map it to a list of tools (or a package which requires this tools) and use that information to setup its worker

I was just wondering when all that information is moved into kiwi-boot-requires, which will go away, we will loose the information. imho kiwi-boot-requires is abused here as tools-required-to-build-image-type-X and that's not what we should do

We can move it if you like but the problem you are facing with missing tools to build docker images in the obs worker will then be postponed to a later point in time.

If that's ok for you I'm fine. In this case I just like to ask if you could move all other provides tags to the kiwi-boot-requires package too, so that the change set is complete and I can merge it

Thanks

adrianschroeter: You want to maintain the kiwi dependencies outside of kiwi repository?
We had this nightmare before where builds constantly break, don't want to go back to that :/

But I will move the other provides as you requested ...

adrianschroeter: done

schaefi: No I don't want to maintain hard kiwi dependencies outside of kiwi.

I think I missed to make my point clear. The package kiwi-boot-requires was created to maintain package dependencies to the kiwi boot descriptions. I think you remember your first version of a script which extracts those information from the kiwi boot descriptions.

All of that will go away.

We already deleted isoboot and vmxboot and provided dracut module packages
called dracut-kiwi-live and dracut-kiwi-overlay for features dracut does not provide upstream.
Tools required for dracut and this modules are handled in that packages. The still to be moved
features from the boot descriptions: netboot and oemboot will follow and for a transition period the netboot and oemboot boot descriptions will be provided as extra boot descriptions in the kiwi-descriptions repository: (https://github.com/SUSE/kiwi-descriptions) repository because they are now optional.

At that time the kiwi-boot-requires package will become obsolete and that makes it also
much easier for the buildservice because there is no magic kiwi-boot-requires package anymore obs has to deal with currently

Now the problem:

This kiwi-boot-requires packages requires e.g "umoci" or "skopeo" but this tools has nothing to do with boot images. In this case they are needed to create a docker image which even has no initrd(boot) image. For building a docker image the build environment needs this tool because kiwi calls it. I don't consider this as a hard kiwi dependency. "umoci" and "skopeo" would be required if the kiwi image description has the build type "docker" configured and kiwi is requested to build that docker image. So this is a soft requirement depending on the image target type

One option is to require each and every tool kiwi would call to build one of its supported images.
But that's something I dislike because then kiwi would require all filesystems, all of lvm all of mdraid, all docker tools, all core tools and so on. Setting it as recommended limits it to suse packages only. And that's why we have runtime checks for needed tools and the provides tags in the kiwi base package which allows external parties to build an image build environment based on capabilities

With the provides tags now moved we need to talk about how to handle tools required to build
image type X when kiwi-boot-requires goes away. The proposed solution in this pull request is in my eyes only a workaround to the real issue and will come back ;)

merging next

PR #562, opened 2017-12-05
Do not match comments and PIs in XSLT templates

michal42:
I wanted to add a simple vim modeline to my XML description:

This made kiwi consume insane amounts of memory during the XSLT
transform step. While this may be a bug in my version of lxml, we do not
transform comments on processing instructions in the conversion
templates, so the easiest solution is not to match them.

Signed-off-by: Michal Marek <MichalMarek1@eaton.com>

### DIFF
--- a/kiwi/xsl/convert14to20.xsl
+++ b/kiwi/xsl/convert14to20.xsl
@@ -6,7 +6,7 @@
 <xsl:strip-space elements="type"/>
 
 <!-- default rule -->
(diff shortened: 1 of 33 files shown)


schaefi:
Looks ok to me, but I would appreciate if Tom has a look too

schaefi:
@tomschr if you have a minute, would be great, thanks

PR #563, opened 2017-12-05
Update failsafe kernel option list

### LINKED ISSUE #554: x11failsafe - deprecated?
I see 'x11failsafe' and 'nomodeset' in failsafe_kernel_options. By the way I can't boot my image with any of these options (X doesn't start at all). I think these options are deprecated and need to be removed. Thoughts?

schaefi:
Delete obsolete parameters and make sure a failsafe boot
does boot into runlevel 3. This Fixes #554

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -159,12 +159,8 @@ class Defaults(object):
                 'apm=off',
                 'noresume',
                 'edd=off',
-                'powersaved=off',
-                'nohz=off',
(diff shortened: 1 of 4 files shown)


schaefi:
One issue I see here is that passing ```3``` is not generally applicable. There are distributions which ignores that, but I don't have a better solution at the moment

Feedback welcome

sndirsch:
Thanks! Looks good to me.

PR #569, opened 2017-12-13
Allow to choose dracut live module

### LINKED ISSUE #568: Allow to choose dracut live module
## Problem description

dracut provides the dmsquash-live module and kiwi provides the kiwi-live module. The layout
of the iso image is compatible to both modules, however there is no way to let the user choose.
In order to support e.g the isofile feature one needs to use dmsquash-live 

* https://github.com/haraldh/dracut/blob/master/dracut.cmdline.7.asc#iso-scanfilename

schaefi:
There is the standard dracut dmsquash-live module based on
the device mapper technology and the kiwi-live module based
on the overlayfs technology. The setup of the live iso structure
in kiwi is compatible to both modules. Thus it makes sense
to allow to choose the technology via the flags attribute

    <type image="iso" ... flags="overlay|dmsquash"/>

Please note both modules supports a different set of live
features. This Fixes #568

### DIFF
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -297,8 +297,14 @@ class LiveImageBuilder(object):
 
     def _create_dracut_live_iso_config(self):
         live_config_file = self.root_dir + '/etc/dracut.conf.d/02-livecd.conf'
+        omit_modules = [
+            'kiwi-dump', 'kiwi-overlay', 'kiwi-repart', 'kiwi-lib'
+        ]
         live_config = [
-            'add_dracutmodules+=" kiwi-live pollcdrom "',
+            'add_dracutmodules+=" {0} pollcdrom "'.format(
+                Defaults.get_live_dracut_module_from_flag(self.live_type)
+            ),
+            'omit_dracutmodules+=" {0} "'.format(' '.join(omit_modules)),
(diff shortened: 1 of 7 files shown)


kiwi/builder/live.py
```diff
@@ -297,8 +297,14 @@ def create(self):
 
     def _create_dracut_live_iso_config(self):
         live_config_file = self.root_dir + '/etc/dracut.conf.d/02-livecd.conf'
+        omit_modules = [
+            'kiwi-dump', 'kiwi-overlay', 'kiwi-repart', 'kiwi-lib'
+        ]
         live_config = [
```
davidcassany: @schaefi, just curiosity, why we need now to omit those modules? I assume they were not created at the time of writing this method, is that the case?

schaefi: The reason is if e.g by accident someone installs another kiwi dracut module e.g dracut-kiwi-oem-repart this module will be also installed and hooks itself into the premount stage. As we create a generic dracut module (hostonly=false) all installed modules will be used. Therefore I omit the ones where I know would conflict with kiwi-live and/or dmsquash-live

davidcassany: Oh, I see, thanks, good to know

PR #572, opened 2017-12-19
Omit kiwi-repart dracut module in oemboot initrd

davidcassany:
KIWI's oemboot initrd with `initrd_system="dracut"` together with
`installiso="true"` requires to have dracut-kiwi-oem-repart package
installed in the system, thus it ends up also being included in the
recreated dracut initrd after booting the oemboot initrd from the
installation iso. This kiwi-repart module causes a boot failure in that
case since no `.profile` file is present, moreover, it has no sense to
run it at that stage, since the disk is already reparted by the
oemboot code.

This commit allows `installiso="true"` and `initrd_system="dracut"` to
play well together.

### DIFF
--- a/kiwi/boot/functions.sh
+++ b/kiwi/boot/functions.sh
@@ -1418,7 +1418,7 @@ function setupInitrd {
         # Call initrd creation tool
         #--------------------------------------
         if [ -x "$dracutExec" ]; then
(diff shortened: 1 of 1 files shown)


schaefi:
yes thanks good catch :+1:

PR #573, opened 2017-12-28
[boot] fix double quote in grub menu which makes kernel updates for C…

### LINKED ISSUE #496: kiwi grub config file is different from grub2-mkconfig result
## Problem description
If a software update is performed on a VMX Image created with KIWI NG, the grub config created by KIWI is either ignored or broken by the update (e.g. kernel package).

## Intended behaviour
The grub config should be updated by a kernel update. This works, but not with the KIWI NG created grub config.

1. create an image with KIWI NG for Fedora 25 / 26 or CentOS
2. perform a software update with dnf update or yum update
3. after the update, either the new kernel is not in the grub config or the system cannot boot at all

If you:
1. run grub2-mkconfig
2. run dnf update / yum update
3. reboot
then you get a modified grub boot menue where you can boot the new or the old kernels with different options.

mmohring:
Fixes #496 .

* It seems the double quote in grub.cfg from kiwi breaks on kernel update on CentOS / RHEL / Fedora
* so removed double quote, then system still boots after kernel update

### DIFF
--- a/kiwi/bootloader/template/grub2.py
+++ b/kiwi/bootloader/template/grub2.py
@@ -34,7 +34,7 @@ class BootLoaderTemplateGrub2(object):
             set timeout=${boot_timeout}
             if [ -n "$$extra_cmdline" ]; then
               submenu "Bootable snapshot $$snapshot_num" {
-                menuentry "If OK, run 'snapper rollback' and reboot." { true; }
+                menuentry "If OK, run snapper rollback and reboot." { true; }
               }
             fi
         ''').strip() + os.linesep


mmohring:
The grub.cfg is of course different from the one produced by grub-mkconfig . But the kernel update does not break the kiwi generated grub.cfg anymore.

mmohring:
@schaefi did you see this fix ?

schaefi:
Happy new year :)

I'm ok with the change

mmohring:
Happy new year back :)

Finally found the kernel update issue and fixed.

PR #579, opened 2018-01-12
Update locale settings

### LINKED ISSUE #577: Changes in locale and keyboard settings
## Problem description

Systemd and Yast2 changed the way locales, keyboard and some other system settings are handled. Now they are not read from `/etc/sysconfig/` anymore, but form /etc/{vconsole.conf, locale.conf} this change will be applied since SLE15.

The change is reported in fate#319454

## Expected behaviour

System wide configurations set in config.xml are applied correctly in SLE15 as they are in current SLE and openSUSE distros.

+ Operating system: SLE15

davidcassany:
* Updates the locale setting methods to use systemd-firstboot tools, which is already designed to be used before any systemd daemon is running or in installation chroot environments. Fixes #577 
* Makes use of `setup_hardware_clock()` method, which was never called before.

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -219,47 +219,60 @@ class SystemSetup(object):
 
     def setup_keyboard_map(self):
         """
-        Setup etc/sysconfig/keyboard console keyboard
+        Setup console keyboard
         """
         if 'keytable' in self.preferences:
-            keyboard_config = self.root_dir + '/etc/sysconfig/keyboard'
-            if os.path.exists(keyboard_config):
-                log.info(
-                    'Setting up keytable: %s', self.preferences['keytable']
-                )
+            log.info(
+                'Setting up keytable: %s', self.preferences['keytable']
+            )
+            if self._systemd_firstboot_has_option('--keymap'):
+                Path.wipe(self.root_dir + '/etc/vconsole.conf')
+                Command.run([
+                    'chroot', self.root_dir, 'systemd-firstboot',
+                    '--keymap=' + self.preferences['keytable']
+                ])
+            elif os.path.exists(self.root_dir + '/etc/sysconfig/keyboard'):
                 Shell.run_common_function(
                     'baseUpdateSysConfig', [
-                        keyboard_config, 'KEYTABLE',
+                        self.root_dir + '/etc/sysconfig/keyboard', 'KEYTABLE',
                         '"' + self.preferences['keytable'] + '"'
                     ]
                 )
             else:
                 log.warning(
-                    'keyboard setup skipped etc/sysconfig/keyboard not found'
+                    'keyboard setup skipped no capable '
+                    'systemd-firstboot or etc/sysconfig/keyboard found'
                 )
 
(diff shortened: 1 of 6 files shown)


davidcassany:
@schaefi while having a look at this issue I realized that `setup_hardware_clock` from SystemSetup was never called, I assumed it wasn't just because of a distraction and I have included it together with the other related calls (locale, keyboard and timezone); however I also noticed that the function suseConfig from `kiwi/config/functions.sh` should be deprecated, as all the stuff done there is already done within the python code, thus no need to do anything related in `config.sh`. 

Would you agree to subsitute suseConfig function content with a message claiming the method is deprecated?

davidcassany:
reverted last commit, since I noticed the method `setup_hardware_clock` tries to set (and in ubuntu even fails) the host hardware clock which is not the intention.

Anyway my comment in relation to the suseConfig is still valid (at least for the other parameters, timezone, keyboard and locale). And would probably modify `setup_hardware_clock` in a way that can be used, include a clear comment on why is not being called or just delete it.

kiwi/system/setup.py
```diff
@@ -890,3 +910,15 @@ def _export_deb_package_verification(self, filename):
         )
         with open(filename, 'w') as verified:
             verified.write(query_call.output)
+
+    def _systemd_firstboot_has_option(self, flag):
```
schaefi: I agree we need this option check to see the capability of a command. I also fear we might need this type of check more often in the future and as I know you and me are not a friend of it, I still think we should put this code into an extra class so that we can have it available in all parts of kiwi. What do you think. I suggest to put that into utils/command_capabilities.py and maintain it as class method so that you can write e.g

```python
CommandCapabilities.has_option_in_help(call='systemd-firstboot' root='...')
```

Help lookup is pretty common and should work the same way for almost all commands. Maybe more sophisticated checks are needed in the future and those could go there and are collective all at the same place

What do you think ?

A little more effort to start with but I think it's worth it

davidcassany: Sure you are I right,  I buy the idea, it makes sense to me.

schaefi:
Sorry overlooked the initial questions

>Would you agree to subsitute suseConfig function content with a message claiming the method is >deprecated?

yes, finally we can do that

I also think there is no way to mess with the hardware clock as there is only one when building the image and that's the wrong one :) So I'm ok with deleting this code

kiwi/system/setup.py
schaefi: ok great that means you plan to add commits to this one before we merge or want to do it in an extra pull request ?

davidcassany: I will add commits here, I was thinking on making the deprecated suseConfig an extra PR, here I will just add the CommandCapabilities class

schaefi: 👍

schaefi:
@davidcassany I like to merge it but not sure if you want the deletion of the hwclock stuff in an extra request ?

davidcassany:
Sure, merging then. The hwclock deserves another request.

Issue #576, opened 2018-01-09
Support oem install image via dracut


schaefi:
## Description

Creating an image of type oem with the installiso attribute activated creates in addition to the expandable disk image also an installation iso image which can be used to install (dump) the image on a target system and serves the purpose of a light weight installer

The implementation to drive this process is still part of the custom kiwi oemboot code and is the last part of the oemboot code which is not yet available as a dracut module package

## Plan

* Create a dracut-kiwi-dump package which provides the 90kiwi-dump dracut extension module to allow installation of kiwi images to a selected target disk

schaefi:
Work is in progress in branch: dracut_kiwi_dump

schaefi:
Once oem images with dracut supports all major components of oemboot the next question is how to handle the oemboot code ?

* As this is complex code which grew over years it should not be dropped but available in the kiwi-descriptions repo

* An extra package e.g kiwi-oemboot-descriptions could be created which provides the oemboot code and the descriptions for the distributions we want to support.

* obs could fetch that package if an image wants to build against oemboot via boot="/path/to/oemboot/description..."

Thoughts ?

davidcassany:
I like the idea of removing the code from the main kiwi package, this is clear. Not sure if the boot code should be moved to kiwi-descriptions or not, it feels odd to me to have the bash code there. However I like the idea to move it to somewhere else as it forces us to clean the kiwi repo and to clearly divide what is part of the boot code and what is not part of it. 

I am also fine with having a new package that includes the boot descriptions and the boot code. I think that obs depending on the image type and on the presence of boot=".." could require this new package, probably I would even move some provides from `kiwi-boot-require` to the `kiwi-oemboot-descriptions` (i.e. `Provides: kiwi-boot:oemboot`).

schaefi:
Thanks David for your feedback

> it feels odd to me to have the bash code there.

yes the legacy bash code should stick together with the legacy description(s) that's why I thought both should be moved to the kiwi-descriptions repo
 
>I am also fine with having a new package that includes the boot descriptions and the boot code. I >think that obs depending on the image type and on the presence of boot=".." could require this new >package, probably I would even move some provides from kiwi-boot-require to the kiwi-oemboot->descriptions (i.e. Provides: kiwi-boot:oemboot).

yes I agree and this will include some discussions with the obs people because they validate the provides from the kiwi-boot-requires package which I would like to get rid off too. Once there is no extra boot code this package will become obsolete

[...]

PR #580, opened 2018-01-16
Dracut kiwi dump

schaefi:
Implementation of dracut kiwi-dump module
    
Provide the capabilities of the oem install code as dracut module. The implementation covers the result   of the installiso="true" configuration. Installation from remote sources still needs to be done and will    follow in an extra pull request. This addresses Issue #576

### DIFF
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -3,6 +3,7 @@ type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
 type setup_debug >/dev/null 2>&1 || . /lib/kiwi-lib.sh
 type run_dialog >/dev/null 2>&1 || . /lib/kiwi-dialog-lib.sh
 type get_block_device_kbsize >/dev/null 2>&1 || . /lib/kiwi-partitions-lib.sh
+type fetch_file >/dev/null 2>&1 || . /lib/kiwi-net-lib.sh
 
 #======================================
 # Functions
@@ -153,11 +154,12 @@ function check_image_fits_target {
     fi
 }
 
-function dump_local_image {
+function dump_image {
     declare kiwi_oemsilentinstall=${kiwi_oemsilentinstall}
     declare kiwi_oemunattended=${kiwi_oemunattended}
     local image_source_files=$1
     local image_target=$2
+    local image_from_remote=$3
     local image_source
     local image_basename
     image_source="$(echo "${image_source_files}" | cut -f1 -d\|)"
@@ -165,8 +167,13 @@ function dump_local_image {
     local progress=/dev/install_progress
     local load_text="Loading ${image_basename}..."
     local title_text="Installation..."
+    local dump
 
-    read_local_image_metadata "${image_source}"
+    if [ ! -z "${image_from_remote}" ];then
+        dump=dump_remote_image
+    else
+        dump=dump_local_image
+    fi
 
     check_image_fits_target "${image_target}"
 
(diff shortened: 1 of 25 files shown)


schaefi:
In order to activate dracut for oem one explicitly has to state

```xml
<type image="oem" .... initrd_system="dracut"/>
```

schaefi:
Code for iso installer is provided with the sub package dracut-kiwi-oem-dump. Therefore the image description also must state:

```xml
```

An appropriate runtime check for this condition is still missing and will be provided next

schaefi:
Thanks for testing it. If you don't see severe issue I will continue to look into the network support and fill in the open TODO gaps. Thus there might be another round of review for this one :) Thanks

davidcassany:
Sure, so far so good, feel free to ask me tests/review/help/opinion at any time :) Later today if I have some time I'll do some tests with non openSUSE/SUSE distros.

schaefi:
@davidcassany ok I'm done with the remote install part :)

If you would like to test it here is what I did:

1. You need a proper dnsmasq setup like this one

```
# Don't function as a DNS server:
port=0

# Log lots of extra information about DHCP transactions.
log-dhcp

# Set the root directory for files available via FTP.
tftp-root=/srv/tftpboot

# The boot filename, Server name, Server Ip Address
dhcp-boot=pxelinux.0,,192.168.100.16

# Disable re-use of the DHCP servername and filename fields as extra
# option space. That's to avoid confusing some old or broken DHCP clients.
dhcp-no-override

# PXE menu.  The first part is the text displayed to the user.  The second is the timeout, in seconds.
pxe-prompt="Booting FOG Client", 1

# The known types are x86PC, PC98, IA64_EFI, Alpha, Arc_x86,
# Intel_Lean_Client, IA32_EFI, BC_EFI, Xscale_EFI and X86-64_EFI
# This option is first and will be the default if there is no input from the user.
pxe-service=X86PC, "Boot to FOG", pxelinux.0
pxe-service=X86-64_EFI, "Boot to FOG UEFI", ipxe
pxe-service=BC_EFI, "Boot to FOG UEFI PXE-BC", ipxe

dhcp-range=192.168.100.16,proxy
    
```

2. you need a working tftp server, I did not use dnsmasq's one, maybe that would work too
    So I have atftp installed and running

```
systemctl start atftpd.socket
systemctl start atftpd
```

3. Next build the image with ```installpxe="true"```

4. The result tarball contains:

LimeJeOS-Leap-42.3.initrd
LimeJeOS-Leap-42.3.kernel
LimeJeOS-Leap-42.3.md5
LimeJeOS-Leap-42.3.xz

which I put in /srv/tftpboot/image

It also contains:

pxeboot.kernel
pxeboot.initrd

which I put in /srv/tftpboot/boot as files ```initrd`` and ```linux```

5. you need a proper pxe boot config /srv/tftpboot/pxelinux.cfg/default

```
DEFAULT KIWI-Boot

LABEL KIWI-Boot
        kernel boot/linux
        append initrd=boot/initrd splash=silent vga=0x317 rd.kiwi.install.pxe rd.kiwi.install.image=tftp://192.168.100.16/image/LimeJeOS-Leap-42.3.xz rd.kiwi.debug rd.debug

LABEL Local-Boot
        localboot 0
```

6. and last but not least the pxelinux.0 loader from syslinux but I think you know about that

davidcassany:
@schaefi awsome, I will test it, hopefully today. Now I am reading `dmsquash-live` dracut module and `kiwi-live` module, pretty interesting reading :)

schaefi:
dmsquash-live is hard to read... as usual when code handles device mapper, my head is always spinning with devmapper, that thing is crazy cool and hard to do it right :)

schaefi:
Once we have moved all historical boot code all the rest is covered by shellcheck which is really nice. I'm happy that soon we will have all shell code under control like we do with the python code

schaefi:
@davidcassany I'll give this one a last test round today and will merge unless there are concerns

davidcassany:
sure, I'm currently giving it a try

schaefi:
I found one case in which systemd starts a rescue shell because a mount target is busy. I don't think it's because of the dump code but probably the repart code causes a udev event which mounts the target. I'll check that

schaefi:
Found the problem and will add one additional commit next

davidcassany:
All right, sorry it is taking me a little longer than expected to make some successful test, now I will test your PXE server receipt, since the VMX server I have fails in someway to deliver the image to download.

schaefi:
yeah for pxe testing I also had several iterations to come to a speedy setup. dnsmasq plus atftp works well for me and does not conflict with other services you might have in your network (if configured as shown above). qemu -boot n picks up the right server and you can test all on one machine

davidcassany:
Finally I managed to successfully test the pxe boot for oem installation. I works fine, your last commit was the key :+1 then I just had to install some missing packages into the image and it worked fine. I managed to boot using your dnsmasq configuration receipt and also using the libvirt network configuration I used to use before, which is great, I can pxe boot using the same config I used to use previously.

Ready to merge from my side :)

schaefi:
totally great, thanks for testing. I'll merge :)

PR #581, opened 2018-01-16
Deprecate suseConfig method and remove hardware config method

davidcassany:
This in this commit:

- the contents of the suseConfig method from kiwi/config/functions.sh
  are replaced by a deprecated message.
- the setup_hardware_clock method from SystemSetup is removed as it is
  not used anywere in the code.


schaefi:
I think we should delete it from the schema as well and provide an xslt which auto removes it if present. I have done the commits to do this, hope you don't mind it was low hanging fruits :)

Once merged schema docs needs to be rebuild, we don't need a pull request for this though as it's an automatic process

davidcassany:
Great thanks 👍

schaefi:
cool good to go then

PR #582, opened 2018-01-16
Extend CommandCapabilities to also check version

davidcassany:
* CommandCapabilities.check_version method has been included into CommandCapabilities and used to verify the tar command version.
* Also a raise_on_error option has been included into CommandCapabilities.has_option_in_help.


schaefi:
yay, the class already pays off :)

Thanks looks good to me

PR #584, opened 2018-01-17
Added runtime check for dracut-kiwi-oem-dump

schaefi:
The installation of the above mentioned dracut module package
is required for oem images which uses dracut as initrd system
and have the creation of an installation image requested.
This is related to Issue #576


schaefi:
Once #580 is in, this one needs to follow

schaefi:
will merge the runtime check and submit new packages to Staging

Issue #521, opened 2017-10-23
Kiwi image that runs from a VFAT USB-Stick


rakoenig:
## Problem description
Want to create a sort of swiss army knife that boots openSUSE from an USB stick and then performs maintenance operations on the booted system, like updating BIOS or DMI information with factory tools. 
Problem is that target users often just have a Windows PC and want to have a stick where they can put new BIOS versions. That means the stick should be FAT formatted so that Windows can write on it.

Currently I tried two approaches:

1. Create an type="oem" image with a spare partition. That seems to work, but plugging this into a Windows machine asks the user if all the linux partitions should be formatted. And some people also said they don't see the spare partion, even if its formatted with vfat. 

2. Create an type="iso" image and put this on the stick. But then the stick has a linux file system and a second partition for the "cow" function is created to get the changes in the live system.

## Expected behaviour
You can add files by copying them to the stick which is on a FAT partition.

davidcassany:
@rakoenig you can check #585 PR, it should mostly solve this issue.

schaefi:
Yes with iso-scan now available this also makes live-fat-stick to work with it

* https://software.opensuse.org/package/live-fat-stick

We should mention that in the documentation. Thus I'll keep this one open. The docs will receive a bigger change in the next days due to our effort to move all custom kiwi boot code to dracut

PR #585, opened 2018-01-20
Adding iso-scan feature to kiwi-live dracut module

### LINKED ISSUE #574: Support for iso-scan to live overlay module
## Problem description
Earlier openSUSE live were able to boot from iso images, after the change to dracut we are unable to do that, this is a regression.

## Expected behaviour
As dracut already supports iso-scan, adding support to overlay module will allow feature parity in this regards.

https://github.com/haraldh/dracut/blob/master/dracut.cmdline.7.asc#iso-scanfilename 
Appending iso-scan/filename=/$isoname root=LABEL=$isolabel should get it booting.

Having something equal to kiwi_cowdevice=/dev/disk/by-uuid/$stickuuid kiwi_cowsystem=/boot/cowfile would also be nice.

Also see: https://github.com/SUSE/kiwi/issues/521

davidcassany:
This commit adds the iso-scan scripts to kiwi-live module_setup
if the former files are found in the system inside the dmsquash-live
dracut module directory.

Fixes #574 and is related to #521

### DIFF
--- a/dracut/modules.d/90kiwi-live/module-setup.sh
+++ b/dracut/modules.d/90kiwi-live/module-setup.sh
@@ -23,10 +23,21 @@ installkernel() {
 install() {
     declare moddir=${moddir}
     declare systemdutildir=${systemdutildir}
+    declare dracutbasedir=${dracutbasedir}
+    local dmsquashdir=
     inst_multiple \
         umount dmsetup blockdev blkid lsblk dd losetup \
         isoinfo grep cut partprobe find wc fdisk tail mkfs.ext4 mkfs.xfs \
         checkmedia dialog
+
+    dmsquashdir=$(find "${dracutbasedir}/modules.d" -name "*dmsquash-live")
+    if [ -n "${dmsquashdir}" ] && \
+        [ -f "${dmsquashdir}/parse-iso-scan.sh" ] && \
+        [ -f "${dmsquashdir}/iso-scan.sh" ]; then
+        inst_hook cmdline 31 "${dmsquashdir}/parse-iso-scan.sh"
+        inst_script "${dmsquashdir}/iso-scan.sh" "/sbin/iso-scan"
+    fi
+
     inst_hook cmdline 30 "${moddir}/parse-kiwi-live.sh"
     inst_hook pre-udev 30 "${moddir}/kiwi-live-genrules.sh"
     inst_hook pre-mount 30 "${moddir}/kiwi-live-checkmedia.sh"


davidcassany:
With this change is it possible to boot from an iso (regardless of using `flags="overlay"` or `flags="dmsquash"`) with a grub2 menu entry like this:

```
menuentry 'Live openSUSE' --class suse --class gnu-linux --class gnu --class os {
        set isolabel=CDROM
        set isofile="/boot/iso/LimeJeOS-tumbleweed.x86_64-1.99.1.iso"
        loopback loop $isofile
        linux (loop)/boot/x86_64/loader/linux boot=isolinux iso-scan/filename=$isofile root=live:CDLABEL=$isolabel ro rd.live.image rhgb
        initrd (loop)/boot/x86_64/loader/initrd
}
```

Also managed to boot from a vfat usb using the `live-fat-stick` command and adjusting syslinux bootloader config afterwards:

```bash
host:~> sudo live-fat-stick LimeJeOS-tumbleweed.x86_64-1.99.1.iso /dev/sdc1
```

Where `/dev/sdc1` is the vfat partition of the USB stick. Later I updated the `<usbmountpoint>/boot/syslinux/syslinux.cfg` file to include the `iso-scan/filename=/LimeJeOS-tumbleweed.x86_64-1.99.1.iso root=live:CDLABEL=CDROM` parameters to the `append` line, something like:

```
LABEL LimeJeOS-tumbleweed.x86_64-1.99.1.iso
        kernel linux-LimeJeOS-tumbleweed.x86_64-1.99.1.iso
        append initrd=initrd-LimeJeOS-tumbleweed.x86_64-1.99.1.iso isofrom=/dev/disk/by-uuid/5E0C-AA8A:/LimeJeOS-tumbleweed.x86_64-1.99.1.iso isofrom_device=/dev/disk/by-uuid/5E0C-AA8A isofrom_system=/LimeJeOS-tumbleweed.x86_64-1.99.1.iso loader=syslinux splash=silent   quiet showopts iso-scan/filename=/LimeJeOS-tumbleweed.x86_64-1.99.1.iso root=live:CDLABEL=CDROM 
```

About #521 I am not sure that much more can be done from KIWI perspective, as moving the ISO to a vfat stick and set the bootloader and other parameters is out if the scope of KIWI, but there is the great `live-fat-stick` tool for that matter.

Persistent COW files are not part of this PR and I believe it should be handled in another PR in order to facilitate reviewing and tests.

davidcassany:
Yes, I was afraid that iso-scan scripts would be somehow coupled with dmsquash-live module, but it turned out that it is pretty simple and atomic.

Sure, live-fat-stick should be mentioned in the documentation, this is true.

cyberorg:
There is https://software.opensuse.org/package/live-grub-stick using grub2(works with all partitions supported by grub2) instead of syslinux(only vfat), and https://software.opensuse.org/package/live-usb-gui script that act as gui frontend to both those scripts.

cyberorg:
Tried with http://widehat.opensuse.org/tumbleweed/iso/openSUSE-Tumbleweed-Rescue-CD-x86_64-Snapshot20180129-Media.iso get just dracut shell, there is no /var/log/boot.kiwi, what should I be looking for to check if this patch is included in the initrd?

davidcassany:
@cyberorg In order to see if the patch is included you should check the dracut modules version, as I can see the patch should be in any `dracut-kiwi-live` package equal or higher than v9.12.0. You can check the installed version from the packages file produced by kiwi.

Also you can check `/usr/lib/dracut/modules.d/90kiwi-live/module_setup.sh` script from the dracut shell, there should be something like that:

```bash
dmsquashdir=$(find "${dracutbasedir}/modules.d" -name "*dmsquash-live")     
if [ -n "${dmsquashdir}" ] && \                                             
    [ -f "${dmsquashdir}/parse-iso-scan.sh" ] && \                          
    [ -f "${dmsquashdir}/iso-scan.sh" ]; then                               
    inst_hook cmdline 31 "${dmsquashdir}/parse-iso-scan.sh"                 
    inst_script "${dmsquashdir}/iso-scan.sh" "/sbin/iso-scan"               
fi
```

This is the patch I did to include iso-scan.sh script, which basically looks for the iso in devices and loop mounts it. If this part is there then the patch is included into the iso.

cyberorg:
Works, tested using images here: https://build.opensuse.org/project/show/openSUSE:Factory:Live

Thanks

PR #588, opened 2018-01-24
Adding package_gpgcheck flag as an option for set-repo and add-repo

### LINKED ISSUE #565: Expand --add-repo with package_gpgcheck
## Problem description
When adding a repo during Kiwi build (command line), I cannot use the option to enable/disable package_gpgcheck. This option IS available when adding a repo to the kiwi config.xml file. 

## Expected behaviour
All options that I can set for adding a repo in the kiwi config.xml file are also available when adding a repo on the command line.

Speaks for it self, running a Kiwi system build with --add-repo, i cannot set the gpgcheck. 

+ KIWI version: 9.11.24 (and earlier)
+ Operating system:  OpenSuse Leap 42.2 and 42.3

davidcassany:
Fixes #565

### DIFF
--- a/doc/source/commands/system_build.rst
+++ b/doc/source/commands/system_build.rst
@@ -14,8 +14,8 @@ SYNOPSIS
        [--clear-cache]
        [--ignore-repos]
        [--ignore-repos-used-for-build]
-       [--set-repo=<source,type,alias,priority,imageinclude>]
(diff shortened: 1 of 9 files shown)


schaefi:
Looks great, the only part missing is the man page update

* doc/source/commands/

davidcassany:
Oh you are right, going for it

schaefi:
nice :)

PR #590, opened 2018-01-24
- Image name generation for GCE images

rjschwei:
+ Fix the name generation for GCE images. The implementation assumed that
    GCE images are of a type that use a kiwi generated initrd, looking for
    image name components based on the boot attribute in the type element.
    This assumption is not correct, use the name provided in the XML to
    compose the image name.


rjschwei:
@schaefi please push to SLE 15 as quickly as possible testing in GCE is currently blocked.

schaefi:
Thanks good catch 👍

PR #592, opened 2018-01-25
Fixed kernel version check for zImage

### LINKED ISSUE #587: kiwi tool kversion does not work on zImage
## Problem description

kiwi provides a tool named kversion which extracts the kernel version information from a kernel image file. However if the file is a zImage like it is the case for kernels on arm the version extraction failed

```
mkimage -l
```

is able to list the version on zImage compressed kernel files

This problem currently causes arm builds to fail

```
[  249s] [ DEBUG   ]: 16:28:01 | EXEC: [kversion /usr/src/packages/KIWI-oem/build/image-root/boot/zImage]
[  249s] [ DEBUG   ]: 16:28:01 | EXEC: [chroot /usr/src/packages/KIWI-oem/build/image-root dracut --force --no-hostonly --no-hostonly-cmdline --xz --install /.profile --install /config.partids openSUSE-Tumbleweed-ARM-JeOS-panda.armv7l-2016.03.09.initrd.xz (no output on stdout)]
[  250s] [ DEBUG   ]: 16:28:01 | EXEC: Failed with stderr: dracut: Cannot find module directory /lib/modules/(no output on stdout)/
```

schaefi:
kernels build as zImage contains the decompressor code
as part of the kernel image and could be therefore
compressed by any possible compression algorithm.
In this case we assume/hope that there is also a
standard gz compressed vmlinux version of the kernel
available and check this one instead of the zImage
variant. Fixes #587

### DIFF
--- a/kiwi/system/kernel.py
+++ b/kiwi/system/kernel.py
@@ -53,8 +53,21 @@ class Kernel(object):
         for kernel_name in self.kernel_names:
             kernel_file = self.root_dir + '/boot/' + kernel_name
             if os.path.exists(kernel_file):
+                kernel_file_to_check = kernel_file
+                if 'zImage' in kernel_file_to_check:
+                    # kernels build as zImage contains the decompressor code
+                    # as part of the kernel image and could be therefore
+                    # compressed by any possible compression algorithm.
+                    # In this case we assume/hope that there is also a
+                    # standard gz compressed vmlinux version of the kernel
+                    # available and check this one instead of the zImage
+                    # variant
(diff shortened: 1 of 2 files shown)


schaefi:
Unfortunately I could not come up with a better solution as this workaround

It would be much effort to find which compression algorithm was used and where the actual compressed data starts or in other words the decompression code ends

Of course I'm open for more sophisticated solution

schaefi:
ok then let's give this a try and maybe we find a better solution in the future

PR #593, opened 2018-01-25
Make sure toplevel target dir keeps permissions

### LINKED ISSUE #557: obs_scm with root overlay causing permission problems
## Problem description
When using obs_scm to get a root overlay .obscpio like so; https://build.opensuse.org/package/view_file/Virtualization:Appliances:Images:Testing_x86/test-image-oem/_service?expand=1
tar_scm service will create the "root" folder with chmod 700 as seen below:
```
[   62s]   3609    4 drwxr-xr-x   4 root     root         4096 Dec  1 12:20 /usr/src/packages/SOURCES/
[   62s]   3614    4 -rw-r--r--   1 root     root          401 Dec  1 12:20 /usr/src/packages/SOURCES/_service
[   62s]   9701    4 drwx------   3 root     root         4096 Dec  1 12:20 /usr/src/packages/SOURCES/root
[   62s]   9702    4 drwxr-xr-x   6 root     root         4096 Dec  1 12:20 /usr/src/packages/SOURCES/root/etc
[   62s]   9705    4 drwxr-xr-x   2 root     root         4096 Dec  1 12:20 /usr/src/packages/SOURCES/root/etc/profile.d
....
``` 
When rsync copies the overlay to the image-root:
`[   63s] [ DEBUG   ]: 12:21:10 | EXEC: [rsync -v -r -p -t -D -H -X -A --one-file-system --links /usr/src/packages/SOURCES/root/ /usr/src/packages/KIWI-oem/build/image-root]`

 it will also include the parent folders (`/usr/src/packages/SOURCES/root`) attributes as stated in the rsync man pages;
>A trailing slash on the source changes this behavior to avoid creating an additional directory level at the destination. You can think of a trailing / on a source as meaning "copy the contents of this directory" as opposed to "copy the directory by name", but in both cases the attributes of the containing directory are transferred to the containing directory on the destination.

This will cause `/` on the new image to have `drwx------` as seen below:
```
[   63s] [ DEBUG   ]: 10:02:06 | 655373    4 drwx------  24 root     root         4096 Dec  1 10:02 /
...
```

1. Create a _service in OBS with:
``` 
    <service name="obs_scm">
        <param name="url">https://github.com/SUSE/kiwi.git</param>
        <param name="scm">git</param>
        <param name="subdir">build-tests/x86/test-image-oem/root</param>
        <param name="filename">root</param>
        <param name="version">_none_</param>
      	<param name="revision">master</param>
    </service>
```
2. add a `find / -ls` in `config.sh`
3. Observe permission of `/` in log output

+ KIWI version: 9.11.17
+ Operating system: SLES12-SP3
+ OBS version: 2.8

davidcassany:
When syncing data via rsync we make sure the toplevel target
directory the data gets synced to does not change it's origin
permissions. This Fixes #557

This is a backport to SLE12_SP3, fixes bsc#1077619

### DIFF
--- a/kiwi/utils/sync.py
+++ b/kiwi/utils/sync.py
@@ -15,6 +15,8 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+import os
+from stat import ST_MODE
 import xattr
 
 # project
@@ -69,11 +71,21 @@ class DataSync(object):
                 exclude_options.append(
                     '/' + item
                 )
+        target_entry_permissions = oct(os.stat(self.target_dir)[ST_MODE])
         Command.run(
             ['rsync'] + rsync_options + exclude_options + [
                 self.source_dir, self.target_dir
             ]
         )
+        # rsync applies the permissions of the source directory
+        # also to the target directory which is unwanted because
+        # only permissions of the files and directories from the
+        # source directory and its contents should be transfered
+        # but not from the source directory itself. Therefore
+        # the permission bits of the target directory before the
+        # sync are applied back after sync to ensure they have
+        # not changed
+        os.chmod(self.target_dir, target_entry_permissions)
 
     def target_supports_extended_attributes(self):
         """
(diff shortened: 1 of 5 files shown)


schaefi:
This was just a cherry pick from master correct ?

schaefi:
I'm just asking because then not much review would be needed ;)

davidcassany:
Sure, it was just a cherry pick, not sure if it can actually be tracked somehow if it comes from a cherry pick, I'd say there must be a way. But anyway note that the commit date is from 4th Dec and you and me (because I amended the comment to include the bug number)  are the authors.

schaefi:
yes, so this is great to go then. Thanks

PR #594, opened 2018-01-25
Added build of python-kiwi for Debian based distro

schaefi:
This is a first step to also build the kiwi project for
Debian based distros. However so far this only builds
the dracut module packages as those are required for
Debian builds on non Debian based build hosts. In the
long run this package setup should be extended to have
kiwi fully packaged for Debian


schaefi:
my deb skills are limited, so forgive me if there is anything stupid in here :)

Conan-Kudo:
It's possible to build for Debian targets using the spec file instead of maintaining two different sets of packaging. I do this for many of the projects I work. As an example: https://github.com/datto/dattobd/blob/master/dist/dattobd.spec

@schaefi If you're interested in this, I can prepare a PR that does this for you. :)

schaefi:
>  If you're interested in this, I can prepare a PR that does this for you. :)

@Conan-Kudo this would be super super cool if you could do that. Thanks much

schaefi:
@Conan-Kudo do you still plan to help us out here ? so I can close this pull request ?

would be really great, thanks

Conan-Kudo:
@schaefi I am working on it, so if you'd like to close this, go ahead.

schaefi:
closing, #618 is the way better approach to do this :)

PR #595, opened 2018-01-25
Fix arm test

davidcassany:
In this PR:

* Included openSUSE:Factory/ports repository


davidcassany:
@schaefi integration test in the OBS is pulling from that branch, since the build was blocked a couple of days ago I did not want to start making commits to master in a trial error fashion, thus I separated in this branch. 

The problem is that I can't test it locally as I did with the x86 tests before committing, thus a new branch looked like the way to go.

schaefi:
Hmm, Virtualization:Appliances:Images:Testing_arm/test-image-panda-oem is in state

* succeeded

Are you sure we need this change ?

schaefi:
Oh now I see it the _service is referencing to fix_arm_test branch
nevermind

schaefi:
With the switch to dracut the hook tarball is no longer needed. With this PR it would come back unused. So you can safely delete that commit

davidcassany:
Yes in fact you are right, I included it not to break the _service which was trying to extract it, but I did not realize it shouldn't be there as it is not even included in the packages section.

davidcassany:
just waiting for the test to be succeeded again and we'll be done here.

davidcassany:
now the test is green again, the _service file does not try to pull the uboot hooks and the repositories are well defined only within the appliance.kiwi file.

This is ready to merge to my eyes.

schaefi:
yes, thanks 👍

PR #596, opened 2018-01-26
Add support for building OVA images (v2)

michal42:
Implement support for format="ova" using VMware's ovftool. We use the
vmdk format as a basis and then just call ovftool to convert *.vmdk +
*.vmx to an OVA.

Changes since https://github.com/SUSE/kiwi/pull/583:
- Explicitly annotate functions that exceed the complexity threshold, instead of bumping the thershold
- Make use of CommandCapabilities.has_option_in_help


kiwi/storage/subformat/ova.py
```diff
@@ -0,0 +1,87 @@
...
+        try:
+            os.unlink(ova)
+        except OSError:
+            pass
+        ovftool_cmd = ['ovftool']
+        if CommandCapabilities.has_option_in_help(
+                'ovftool', '--shaAlgorithm', raise_on_error=False):
+            ovftool_cmd.append('--shaAlgorithm=SHA1')
+        ovftool_cmd.extend([vmx, ova])
+        Command.run(ovftool_cmd)
```
davidcassany: Just a minor comment: I think it makes sense to catch the CommandNotFound exception with a more explicit message warning the user ovftool is not an opensource tool installable from the repositories.

What do you think?

kiwi/storage/subformat/__init__.py
```diff
@@ -85,7 +86,7 @@ def __new__(self, name, xml_state, root_dir, target_dir):
             return DiskFormatGce(
                 xml_state, root_dir, target_dir, custom_args
             )
-        elif name == 'vmdk':
+        elif name == 'vmdk' or name == 'ova' or name == 'ovf':
```
davidcassany: I think it doesn't have much sense to keep 'ovf' format as a valid option. Allowing format="ova" and format="ovf" knowing the result will be identical is confusing IMHO.

In fact this issue was already here, in advance of this PR. The schema allows to set 'ova' or 'ovf' as format while there is or there was no code to handle it, leading to a late failure during the create step. I think we could consider removing the 'ovf' format as a valid one in the schema, but well, I could also live leaving it as is, this is just a mere opinion.

@schaefi what do you think?

kiwi/storage/subformat/ova.py
michal42: OK, I will change it, but I'm actually off for vacation in a few moments :)

kiwi/storage/subformat/__init__.py
schaefi: I agree it's confusing to keep 'ovf' to be handled in the code.

Let's delete this and keep only:

```python
elif name == 'vmdk' or name == 'ova':
```

Once merged let us create a follow up pull request which deletes the ovf value from the format attribute to get an early error condition

Thoughts ?

kiwi/storage/subformat/ova.py
schaefi: Enjoy your vacation and when you are back this one is only a minor change away from being merged 👍 Thanks for the good work

Conan-Kudo:
Is `ovftool` an open source tool packaged in distributions?

schaefi:
It's not, unfortunately. One can fetch it from VMware's download page. There was a project in the past called openOVF but it is no longer actively developed. afaik ovftool is also only capable to create an ovf for the vmware provider. There are other platforms e.g  zvm, powervm and also xen which can read an ovf representation but with different layout. To me ovf felt like a markup for metadata around a binary image. The different platforms supporting it implements there own schema which unfortunately leads to the situation that if we want to support it need to call the platform dependent tools which in case of VMware is not opensource :(

schaefi:
@michal42 It reminds me that the support for ova added here and connected to VMwares ovftool should also check for the ```ovftype``` attribute to make sure this ova is for VMware's platform

See

```python
machine_section = xml_state.get_build_type_machine_section()
if machine_section.get_ovftype() == 'vmware':
    ...
```

and if that's not specified or not matching we should error out. This imho should be done as a runtime check to get an early errror condition. In addition this check should also lookup if ovftool could be found on the system and if not tell the user where to fetch it from

schaefi:
We could create the runtime check in an extra PR though

Conan-Kudo:
@schaefi Should this really be part of kiwi then? There's no reason kiwi itself couldn't generate the ova, as it's just an XML metadata file with a raw image wrapped in a zip file. The OVF spec is well-known and freely available...

schaefi:
>Should this really be part of kiwi then? There's no reason kiwi itself couldn't generate the ova,
> as it's  just an XML metadata 

From a markup perspective I agree with you. But to my knowledge there is no freely available schema how the consumer of the metadata expects it. Thus it's easy to create an ova document but if it is valid for the target, in this case VMware player can't be easily validated and is also a moving target. We had an ova generator in the legacy kiwi stream. This thing was constantly broken because it did not generated the expected results, but was valid ova according to the spec.

I'm ok letting kiwi call ovftool if it exists on the build system. I think that does not invalidate the opensource direction of kiwi. If the tool is not available we error out and leave a good message. It should also be clear that we do not test nor give any warranty to the produced result of a non open source tool

...now the argument could be why calling it then. And I agree the only reason we do it is to be user friendly

I think this is ok, but of course I'm open to discuss if there are concerns

Conan-Kudo:
I think this leads to an uncomfortable slippery slope. Already the documentation is generated with a non-free tool, but making part of the kiwi functions dependent on a tool that not everyone can necessarily get is more of a problem.

Has anyone asked VMware if they would be willing to open up `ovftool`?

schaefi:
> Already the documentation is generated with a non-free tool

only the online part to illustrate the schema documentation better, and by default our own open source schema generator is used. Apart from a non-oss schema docs generator alternative, everything else is sphinx based. I think that's a fair compromise, as it's only an option nobody is forced to use :)

> but making part of the kiwi functions dependent on a tool that not everyone can necessarily 
> get is more of a problem.

Everybody can get it, the problem is that it's source code is not public. For me this results in no warranty for the result of such a tool, but I would not outlaw it as long as it can be used to serve VMware's platform. Calling ovftool only makes sense in the context to create an ova format to be used by a VMware product. The target user group here is therefore exclusively users of proprietary software. At the moment I don't see a conflict with open source targets.

The question is: Is it evil if an open source tool is willing to call a non-oss tool if it exists on the
system to serve a result for a non-oss platform ? The license for the ovf tool is published here:

* https://www.vmware.com/download/eula/vmware-ovf-tool.html

I have not much knowledge about legal topics, so correct me when I'm wrong. To me the license reads like it is expected that this tool could be used in the scope of open source software

> Has anyone asked VMware if they would be willing to open up ovftool?

We have tried that in my team (public cloud development) when we were discussing opportunities with VMware's cloud platform. The images we would need to build also had to be build as ova's and that would have required presence of the tool in obs, best as package. But we did not had much luck here and for other reasons vCloud air was no longer a target. So in short, they did not re-license this in the past two years

davidcassany:
> I have not much knowledge about legal topics, so correct me when I'm wrong. To me the license reads like it is expected that this tool could be used in the scope of open source software

I am also do not have specific knowledge about the licenses topic, however I agree with the assumption that calling the tool from kiwi does not violate the ovf-tool license. Mainly because we are not distributing the tool in any way, neither the distributing the results of using it. In this case using the tool for internal purposes falls under the end user responsibility, since he/she is the responsible of including the tool in the builder host with the agreement of this tool license.

Also, by reading the license, it looks like VMWare already considers the option to distribute open source software that can make use of this ovf-tool.

The #596 looks good to me, I believe we are ready to merge.

schaefi:
All right makes all sense to me too :) I'm gonna merge this now and for reference we have this good conversation. Should there be any obstacles in the future I'm confident we will find a good solution

Thanks everybody

PR #597, opened 2018-01-27
Fixed kiwi dracut config for the final system

schaefi:
Once the image has been deployed on the target and the
system is up and running some of the kiwi dracut modules
used for deployment are no longer needed and should not be
taken into account when another dracut call happens on the
system.


schaefi:
Fixes the issue reported by Roger on the mailing list ```Booting trouble```

PR #600, opened 2018-01-30
Fixed use of profile variables in kiwi-dump module

schaefi:
The profile environment was not imported in the
dracut kiwi-dump module. Thus custom settings like
oem-unattended did not had an effect


schaefi:
The first time I like a shell patch :)

PR #602, opened 2018-01-30
Added chapter for live-fat-stick deployment

### LINKED ISSUE #521: Kiwi image that runs from a VFAT USB-Stick

schaefi:
With the support for the iso-scan feature in KIWI live
ISO image, also the ability to deploy file based on
FAT32 usb sticks via the live-fat-stick tool exists.
This chapter describes how to do it and Fixes #521

### DIFF
--- a/doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
+++ b/doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
@@ -64,29 +64,29 @@ to setup the USB stick with `live-fat-stick`:
    By default, the SYSLINUX configuration does not use the "iso-scan"
    feature. Thus this information needs to be added as follows:
 
-   * Mount the FAT32 partition of the USB Stick. This must be the
-     same device as used in the `live-fat-stick` command call:
+   1. Mount the FAT32 partition of the USB Stick. This must be the
+      same device as used in the `live-fat-stick` command call:
 
-     .. code:: bash
+      .. code:: bash
 
-         $ sudo mount /dev/sdz1 /mnt
+          $ sudo mount /dev/sdz1 /mnt
 
-   * Edit the :file:`/mnt/boot/syslinux/syslinux.cfg` and include the
-     following parameters to the existing **append** line:
+   #. Edit the file :file:`/mnt/boot/syslinux/syslinux.cfg` and include the
+      following parameters to the existing **append** line:
 
-     .. code:: bash
+      .. code:: bash
 
-         iso-scan/filename=/LimeJeOS-Leap-42.3.x86_64-1.42.3.iso root=live:CDLABEL=CDROM
+          iso-scan/filename=/LimeJeOS-Leap-42.3.x86_64-1.42.3.iso root=live:CDLABEL=CDROM
 
-   * Unmount the FAT32 partition:
+   #. Unmount the FAT32 partition:
 
-     .. code:: bash
+      .. code:: bash
 
-         $ sudo umount /mnt
+          $ sudo umount /mnt
 
 5. Boot from your USB Stick
 
    Activate booting from USB in your BIOS/UEFI. As many firmware has different
    procedures on how to do it, look into your user manual.
    Many firmware offers a boot menu which can be activated at boot time.
-   Usually this can be reached by :kbd:`Esc` or :kbd:`F12`.
+   Usually this can be reached by pressing the :kbd:`Esc` or :kbd:`F12` keys.
--- a/doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
+++ b/doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
@@ -1,6 +1,6 @@
 .. _iso_as_file_to_usb_stick:
 
-Deploy ISO Image as file on a FAT32 formated USB Stick
+Deploy ISO Image as File on a FAT32 Formated USB Stick
 ======================================================
 
 .. sidebar:: Abstract
@@ -11,31 +11,31 @@ Deploy ISO Image as file on a FAT32 formated USB Stick
 
    * :ref:`hybrid_iso`
 
-In KIWI all generated ISO images are created to be hybrid. This means,
+In KIWI, all generated ISO images are created to be hybrid. This means,
 the image can be used as a CD/DVD or as a disk. The deployment of such
 an image onto a disk like an USB stick normally destroys all existing
 data on this device. Most USB sticks are pre-formatted with a FAT32
 Windows File System and to keep the existing data on the stick untouched
 a different deployment needs to be used.
 
-The following deployment process just copies the ISO image as an
+The following deployment process copies the ISO image as an
 additional file to the USB stick and makes the USB stick bootable.
-The ability to boot from the stick is configured through a grub
+The ability to boot from the stick is configured through a GRUB v2
 feature which allows to loopback mount an ISO file and boot the
 kernel and initrd directly from the ISO file.
 
 The initrd loaded in this process must also be able to loopback
 mount the ISO file to access the root filesystem and boot the
 live system. The dracut initrd system used by KIWI provides this
-feature upstream called as `iso-scan`. Therefore all KIWI generated
+feature upstream called as "iso-scan". Therefore all KIWI generated
 live ISO images supports this deployment mode.
 
 For copying the ISO file on the USB stick and the setup of the
-grub bootloader to make use of the `iso-scan` feature an extra tool
+GRUB v2 bootloader to make use of the "iso-scan" feature an extra tool
 named `live-fat-stick` exists. The following procedure shows how
-to setup the USB stick with `live-fat-stick`
+to setup the USB stick with `live-fat-stick`:
 
-1. Install the `live-fat-stick` package
+1. Install the `live-fat-stick` package:
 
    The package is available for openSUSE at:
    https://software.opensuse.org/package/live-fat-stick
@@ -43,7 +43,7 @@ to setup the USB stick with `live-fat-stick`
 2. Plug in a USB stick
 
    Once plugged in, check which Unix device name the stick was assigned
-   to. The following command provides an overview about all linux
+   to. The following command provides an overview about all Linux
    storage devices:
 
    .. code:: bash
@@ -61,3 +61,4 @@ to setup the USB stick with `live-fat-stick`
    Activate booting from USB in your BIOS/UEFI. As many firmware has different
    procedures on how to do it, look into your user manual.
    Many firmware offers a boot menu which can be activated at boot time.
+   Usually this can be reached by :kbd:`Esc` or :kbd:`F12`.
--- a/doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
+++ b/doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
@@ -20,7 +20,7 @@ a different deployment needs to be used.
 
 The following deployment process copies the ISO image as an
 additional file to the USB stick and makes the USB stick bootable.
-The ability to boot from the stick is configured through a GRUB v2
+The ability to boot from the stick is configured through a SYSLINUX
 feature which allows to loopback mount an ISO file and boot the
 kernel and initrd directly from the ISO file.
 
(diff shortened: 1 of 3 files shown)


schaefi:
@davidcassany I wrote the chapter just by the docs of live-fat-stick and did not test it in real live. Thus would be nice if you can have a short look if it makes sense what I wrote. Thanks

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,63 @@
+.. _iso_as_file_to_usb_stick:
+
+Deploy ISO Image as file on a FAT32 formated USB Stick
```
tomschr: Use title style capitalization: "Deploy ISO Image as File on a FAT32 Formated USB Stick"

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,63 @@
...
+kernel and initrd directly from the ISO file.
+
+The initrd loaded in this process must also be able to loopback
+mount the ISO file to access the root filesystem and boot the
+live system. The dracut initrd system used by KIWI provides this
+feature upstream called as `iso-scan`. Therefore all KIWI generated
+live ISO images supports this deployment mode.
+
+For copying the ISO file on the USB stick and the setup of the
+grub bootloader to make use of the `iso-scan` feature an extra tool
```
tomschr: * grub -> GRUB
* `iso-scan`? I think this is marked wrongly, because it turns up as like a command. Maybe use normal quotes?

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,63 @@
...
+named `live-fat-stick` exists. The following procedure shows how
+to setup the USB stick with `live-fat-stick`
+
+1. Install the `live-fat-stick` package
```
tomschr: Use colon (:) at the end.

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,63 @@
...
+The initrd loaded in this process must also be able to loopback
+mount the ISO file to access the root filesystem and boot the
+live system. The dracut initrd system used by KIWI provides this
+feature upstream called as `iso-scan`. Therefore all KIWI generated
+live ISO images supports this deployment mode.
+
+For copying the ISO file on the USB stick and the setup of the
+grub bootloader to make use of the `iso-scan` feature an extra tool
+named `live-fat-stick` exists. The following procedure shows how
+to setup the USB stick with `live-fat-stick`
```
tomschr: Use colon (:) at the end.

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,63 @@
...
+
+   .. code:: bash
+
+      $ live-fat-stick LimeJeOS-Leap-42.3.x86_64-1.42.3.iso /dev/<stickdevice>
+
+4. Boot from your USB Stick
+
+   Activate booting from USB in your BIOS/UEFI. As many firmware has different
+   procedures on how to do it, look into your user manual.
+   Many firmware offers a boot menu which can be activated at boot time.
```
tomschr: Maybe, could you add some example of the common keystrokes? Something like this:

```
Many firmware offers a boot menu which can be activated at boot time.
Usually this can be reached by :kbd:`Esc` or :kbd:`F12`.
```

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,63 @@
...
+
+.. sidebar:: Abstract
+
+   This page provides further information for handling
+   ISO images built with KIWI and references the following
+   articles:
+
+   * :ref:`hybrid_iso`
+
+In KIWI all generated ISO images are created to be hybrid. This means,
```
tomschr: Put a comma after "In KIWI, ..."

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,63 @@
...
+the image can be used as a CD/DVD or as a disk. The deployment of such
+an image onto a disk like an USB stick normally destroys all existing
+data on this device. Most USB sticks are pre-formatted with a FAT32
+Windows File System and to keep the existing data on the stick untouched
+a different deployment needs to be used.
+
+The following deployment process just copies the ISO image as an
```
tomschr: remove "just" (filler word)

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,63 @@
...
+additional file to the USB stick and makes the USB stick bootable.
+The ability to boot from the stick is configured through a grub
```
tomschr: grub -> GRUB

I don't know if we should explicitly mention the version (probably GRUB 2)? Just an idea...

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,63 @@
...
+feature which allows to loopback mount an ISO file and boot the
+kernel and initrd directly from the ISO file.
+
+The initrd loaded in this process must also be able to loopback
+mount the ISO file to access the root filesystem and boot the
+live system. The dracut initrd system used by KIWI provides this
+feature upstream called as `iso-scan`. Therefore all KIWI generated
```
tomschr: Is `iso-scan` a command? If not, I would use normal double quotes.

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,63 @@
...
+
+1. Install the `live-fat-stick` package
+
+   The package is available for openSUSE at:
+   https://software.opensuse.org/package/live-fat-stick
+
+2. Plug in a USB stick
+
+   Once plugged in, check which Unix device name the stick was assigned
+   to. The following command provides an overview about all linux
```
tomschr: linux -> Linux

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
schaefi: yes good idea

schaefi:
@tomschr Thanks for your review, how about it now ?

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,64 @@
...
+
+   .. code:: bash
+
+      $ lsblk
+
+3. Call the `live-fat-stick` command as follows:
+
+   .. code:: bash
+
+      $ live-fat-stick LimeJeOS-Leap-42.3.x86_64-1.42.3.iso /dev/<stickdevice>
```
davidcassany: I am afraind this is not enough, when I did the test with live-fat-stick package (in 42.2 and tumbleweed) I had to manually update the syslinux config to make it work, as it was not using the iso-scan variables and flags. However I am also not sure this is something we should document here... since it would be a hack around live-fat-stick tool as it would be very difficult to keep it updated.

I tried the live-grub2-stick tool with the same result, it does not make use of the iso-scan variables.

I wrote how to make it work in a comment [here](https://github.com/SUSE/kiwi/pull/585#issuecomment-359165071), do you think it makes sense to include it in the documentation? Or shall we just include some docs about how to boot the iso from grub (just document the menu entry, as in dracut man page) and then just advise live-fat-stick or live-grub-stick tool does similar things but could require some manual work to update the bootloader config according to the iso-scan docs?

davidcassany: @schaefi I can try to provide some text for that part if you prefer.

schaefi: Oh I see, I completely overlooked that information, probably by the pleasant of an easy solution :)
I'll update the text

davidcassany:
Yes, now this receipt is fine and simple enough ;)

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,92 @@
...
+   .. code:: bash
+
+      $ sudo live-fat-stick LimeJeOS-Leap-42.3.x86_64-1.42.3.iso /dev/sdz1
+
+4. Update the SYSLINUX configuration on the USB Stick
+
+   By default, the SYSLINUX configuration does not use the "iso-scan"
+   feature. Thus this information needs to be added as follows:
+
+   * Mount the FAT32 partition of the USB Stick. This must be the
```
tomschr: Better use an ordered list because each step depends on each other and can not be exchanged.

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,92 @@
...
+     same device as used in the `live-fat-stick` command call:
+
+     .. code:: bash
+
+         $ sudo mount /dev/sdz1 /mnt
+
+   * Edit the :file:`/mnt/boot/syslinux/syslinux.cfg` and include the
```
tomschr: Insert the word "file" before the markup `:file:`.

doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst
```diff
@@ -0,0 +1,92 @@
...
+     .. code:: bash
+
+         $ sudo umount /mnt
+
+5. Boot from your USB Stick
+
+   Activate booting from USB in your BIOS/UEFI. As many firmware has different
+   procedures on how to do it, look into your user manual.
+   Many firmware offers a boot menu which can be activated at boot time.
+   Usually this can be reached by :kbd:`Esc` or :kbd:`F12`.
```
tomschr: I forget one thing in my last suggestion, sorry. Maybe use the following phrase:

```
Usually this can be reached by pressing the :kbd:`Esc` or :kbd:`F12` keys.
```

schaefi: I really hope nobody who does not know that ESC or F12 is a key on the keyboard and needs to be pressed to use it, will read this document :) :)

tomschr: Hehe, yes. Maybe I was a bit too exaggerating. 😉 Yes, maybe in this context the first version is enough.

Just to give you a little background why I did that.

From a writing perspective, I think it can be useful to have "the thing" named correctly. For example, sometimes it could be misleading if you don't have markup or font changes. A text like "Use File" without any style changes could mean anything (a menu item, a link, or something else). When you use "Use the File menu to..." the reader gets a better perspective what to look for.

schaefi:
>    Hehe, yes. Maybe I was a bit too exaggerating. 😉 Yes, maybe in this
>    context the first version is enough.

All is good, I liked your version and used it. I also agree the thing
should have a name :)

schaefi:
>    Edit to syslinux.cfg or grub.cfg is not needed, live-fat-stick and
>    live-grub-stick is fixed to support new initrd
>    [1]cyberorg/live-fat-stick@e794e1b
> 
>    Please use live-grub-stick script(not live-fat-stick) as it supports
>    more partition formats including vfat.

It would be super cool if you could send a patch to the kiwi
docs here. As I wrote the article a few days ago it would be
nice if it is not invalid :)

https://github.com/SUSE/kiwi/blob/master/doc/source/building/working_with_images/iso_to_usb_stick_file_based_deployment.rst

Thanks much

schaefi:
> It would be super cool if you could send a patch to the kiwi
> docs here. As I wrote the article a few days ago it would be
> nice if it is not invalid :)

never mind I saw you already did it :)

Thanks much

PR #603, opened 2018-01-30
Fixed use of stat result in os.chmod

### LINKED ISSUE #564: KIWI 9.11.5 - KiwiRootInitCreationError: TypeError: an integer is required
## Problem description
The new kiwi version generates an error for all OS Targets and stops for all my config.kiwi files previosly working, 9.11.4 worked, 9.11.5 does not.

## Expected behaviour
Should work like version before, e.g. 9.11.4.

1. replace old kiwi with new
2. rebuild old image description

+ KIWI version: 
Package for 9.11.25 from b.o.o Virtualization:Appliances:Staging/python-kiwi
+ Operating system: 
All target OS supported
+ OBS version: 
OBS 2.8 and git master

davidcassany:
oct method returns a string representation which was mistakenly
used in a subsequent os.chmod call. This Fixes #564 

This completes the backport for bsc#1077619

### DIFF
--- a/kiwi/utils/sync.py
+++ b/kiwi/utils/sync.py
@@ -71,7 +71,7 @@ class DataSync(object):
                 exclude_options.append(
                     '/' + item
                 )
-        target_entry_permissions = oct(os.stat(self.target_dir)[ST_MODE])
+        target_entry_permissions = os.stat(self.target_dir)[ST_MODE]
         Command.run(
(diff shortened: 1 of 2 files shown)


schaefi:
Oh, yes I stumbled over this oct() problem in the first implementation and needed an extra commit to solve it. That's not easy to catch in a cherry-pick process.

Looks good to me 👍

PR #606, opened 2018-02-02
Fix packages file generation

### LINKED ISSUE #605: RPM backward compatibility
## Problem description

Building Leap 42.3 or older images from a Tubleweed host results into an empty `.packages` file and into an empty `.verified` file, however the build ends successfully.

## Expected behaviour

The build ends successfully with valid `.packages` and `.verified` files.

Build a pre-tumbleweed image using a tumbleweed host. The cause seams to be related to different RPM versions. At the very end KIWI does some rpm calls using the host tool to query and analyze the build image database, the RPM version differs form the one used by the image (which is older) and and, apparently, because of that fails to query the database.

The call
```bash
rpm --root <path to target dir>/build/image-root -qa
```
does not work if run after the image build. However
```bash
chroot <path to target dir>/build/image-root rpm -qa
```
does.

+ KIWI version: Irrelevant
+ Operating system: Tumbleweed
+ OBS version: none

davidcassany:
This commit checks the %_dbpath value of the image rpm tool. This is
needed since recent rpm versions switched the default database path,
thus running rpm queries to different roots might lead to errors.

Fixes #605

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -853,6 +853,11 @@ class SystemSetup(object):
 
     def _export_rpm_package_list(self, filename):
         log.info('Export rpm packages metadata')
+        dbpath = self._get_rpm_database_location()
+        if dbpath:
+            dbpath_option = ['--dbpath', dbpath]
+        else:
+            dbpath_option = []
         query_call = Command.run(
             [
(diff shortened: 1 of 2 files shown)


schaefi:
:+1:

PR #607, opened 2018-02-03
Use live-grub-stick

cyberorg:
live-grub-stick is preferred over live-fat-stick, it handles grub menu entry required to support old as well as new style of isos.


schaefi:
:+1:

PR #608, opened 2018-02-04
GCE image bundle

rjschwei:
- Move the final image to the target image results directory for the GCE image format

### DIFF
--- a/kiwi/storage/subformat/gce.py
+++ b/kiwi/storage/subformat/gce.py
@@ -112,3 +112,21 @@ class DiskFormatGce(DiskFormatBase):
                 '.' + format_name
             ]
         )
+
+    def store_to_result(self, result):
+        """
+        Store the GCE image tarball into the provided result instance.
+
+        :param object result: Instance of Result
+        """
+        result.add(
+            key='disk_format_image',
+            filename=(
+                self.target_dir +
+                '/' +
+                self.get_target_name_for_format(self.image_format)
+            ),
+            use_for_bundle=True,
+            compress=False,
+            shasum=True
+        )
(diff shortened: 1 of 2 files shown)


rjschwei:
Note the implementation of get_target_name_for_format() in gce.py is inconsistent from the base class implementation for the return value. In the base class get_target_name_for_format() returns the $TARGET_PATH/$FILENAME in gce the method only returns $FILENAME. This inconsistency should probably be fixed, but in a separate PR.

Also locally the tests did not pass due to lack of coverage, but:
...kiwi/storage/subformat/gce.py                           34      0   100%
thus the lack of coverage appears to be in some other module that I did not touch.

rjschwei:
For master branch on my machine:

TOTAL                                                                                15948   1876    88%

FAIL Required test coverage of 100% not reached. Total coverage: 88.24%

========================= 1222 passed in 93.44 seconds =========================
ERROR: InvocationError: '/work/rjschwei/devel/kiwi/kiwi8/.tox/2.7/bin/py.test --no-cov-on-fail --cov=kiwi --cov-report=term-missing --cov-fail-under=100 --cov-config .coveragerc'
ERROR:   unit_py2: commands failed

schaefi:
>Note the implementation of get_target_name_for_format() in gce.py is inconsistent from the base >class implementation for the return value. In the base class get_target_name_for_format() returns >the $TARGET_PATH/$FILENAME in gce the method only returns $FILENAME. This inconsistency >should probably be fixed, but in a separate PR.

I agree the bug is in  get_target_name_for_format of the gce format code. I just wondered why the fix you made did not fix the bug in this method. It would be imho the right place and less code to fix the problem. See the #609 for details

I think the name ```get_target_name_for_format``` I chose is misleading as it returns a path spec.
The base class implements the version which is sufficient for almost all format, those which are an exception need to overwrite the method like in the gce case.

I will create another PR which changes the name to ```get_target_path_name_for_format```

rjschwei:
superseded by #609

PR #609, opened 2018-02-04
Fixed target path name for gce format

schaefi:
get_target_name_for_format in case of the gce format does
not return a path spec which leads to an incomplete result
information for a later bundle command. This patch fixes
this. In a follow up pull request we will also update
the method name from the misleading get_target_name_for_format
function name to get_target_path_name_for_format


schaefi:
something on travis has changed again, leading to the exclude configuration for the autogenerated code in xml_parse.py to be ignored....

PR #612, opened 2018-02-05
Early exit for inconsistent config

### LINKED ISSUE #610: Building uefi system on host that does not have uefi results in error
## Problem description
[ ERROR   ]: 13:59:36 | KiwiBootLoaderGrubModulesError: Module synchronisation failed with: rsync: stderr: rsync: change_dir "/work/tmp/cb-leap-oem/build/image-root/usr/lib/grub2/x86_64-efi" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1165) [sender=3.1.0]

## Expected behaviour
Image creation succeeds or user is informed taht image cration is not possible on this host

Use:
	<type image="oem" filesystem="xfs" boot="oemboot/suse-leap42.3" installboot="install" bootloader="grub2" firmware="uefi">

build on a host that does not have uefi enabled

+ KIWI version: KIWI (next generation) version 9.12.7
+ Operating system:  host openSUSE 42.3
+ OBS version:

rjschwei:
Early exit if firmware is efi or uefi but the expected grub module is not part of the package list

Fixes #610

### DIFF
--- a/kiwi/runtime_checker.py
+++ b/kiwi/runtime_checker.py
@@ -417,19 +417,30 @@ class RuntimeChecker(object):
     def check_grub_efi_installed_for_efi_firmware(self):
         """
         If the image is being built with efi or uefi firmware setting
-        we need the grub2-x86_64-efi package installed
+        we need a grub(2)-*-efi package installed. The check is not 100%
+        as every distribution has different names and different requirement
+        but it is a reasonable approximation on the safe side meaning the
+        user may still get an error but should not receive a false positive
         """
         message = dedent('''\n
-            Firmware set to efi or uefi but the grub2-x86_64-efi is not
+            Firmware set to efi or uefi but not grub*efi* package is
             part of the image build.
         ''')
         firmware = self.xml_state.build_type.get_firmware()
         if firmware in ('efi', 'uefi'):
+            grub_efi_packages = (
+                'grub2-x86_64-efi',  # SUSE
+                'grub-efi',  # Debian based distros have grub-efi-*
+                'grub2-efi'  # RedHat
+            )
             package_names = \
                 self.xml_state.get_bootstrap_packages() + \
                 self.xml_state.get_system_packages()
-            if 'grub2-x86_64-efi' not in package_names:
-                raise KiwiRuntimeError(message)
+            for grub_package in grub_efi_packages:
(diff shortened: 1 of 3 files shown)


kiwi/runtime_checker.py
```diff
@@ -412,6 +414,23 @@ def check_efi_mode_for_disk_overlay_correctly_setup(self):
...
+        message = dedent('''\n
+            Firmware set to efi or uefi but the grub2-x86_64-efi is not
+            part of the image build.
+        ''')
+        firmware = self.xml_state.build_type.get_firmware()
+        if firmware in ('efi', 'uefi'):
+            package_names = \
+                self.xml_state.get_bootstrap_packages() + \
+                self.xml_state.get_system_packages()
+            if 'grub2-x86_64-efi' not in package_names:
```
schaefi: This looks all good and works now. I just tried to avoid checking for explicit package names in the runtime checks because there is no standard for this name across the distributions. Therefore this most probably will fail for e.g centos or other distros and there is no way to switch off the runtime check.

I suggest you run this check only if the configured package manager is set to zypper, which kind of means this is a suse build and you can match for the package safer

rjschwei:
I decided that it would be better to expand the consistency check to other known names rather than just making the assumption that zypper == SUSE and ignoring the problem everywhere else. This creates some holes, but it's better than nothing.

kiwi/runtime_checker.py
```diff
@@ -412,6 +414,34 @@ def check_efi_mode_for_disk_overlay_correctly_setup(self):
...
+            Firmware set to efi or uefi but not grub*efi* package is
+            part of the image build.
+        ''')
+        firmware = self.xml_state.build_type.get_firmware()
+        if firmware in ('efi', 'uefi'):
+            grub_efi_packages = (
+                'grub2-x86_64-efi',  # SUSE
+                'grub-efi',  # Debian based distros have grub-efi-*
+                'grub2-efi'  # RedHat
+            )
```
davidcassany: I am fine with this change, I also prefer to not skip the run time check according to the package manager type. I guess those package names won't change that much over time.

schaefi:
Yes that's nice, thanks Robert :+1:

PR #614, opened 2018-02-06
improved docs 'dropped features'

M0ses:

### DIFF
--- a/doc/source/overview/legacy_kiwi.rst
+++ b/doc/source/overview/legacy_kiwi.rst
@@ -121,7 +121,7 @@ consider to use the legacy KIWI version.
    Therefore the recovery feature offered in the legacy KIWI version
    will not be continued.
 
-* OEM Partition Install
+* OEM Partition Install (**not** :ref:`deploy_from_iso`)
 
   The feature allows to install an image partition based.
   Instead of selecting a target disk the user selects target partitions.


schaefi:
Hmm, I'm confused about the change

Frank was reading the docs and got confused about the drop list item "OEM Partition Install" which lead to the impression kiwi would no longer support the creation of an installation image for the oem type, which is of course not the case.

We support installation of oem images but the special mode to this by mapping image partitions instead of the entire image has been dropped. That's why the list item explicitly states "Partition".
Obviously this is not enough or misleading so we need a better solution.

However the "not <link-ref>" addition now confuses me because the referenced chapter describes
oem images and also the functionality of the installation image for such an oem.

We basically dropped an experimental feature from the oem installation image and that reduces
the list of deployment methods described in the "not <link-ref>" by one and not more. How about a different approach:

```
* Partition based install method in OEM install image

  The section :ref:`deploy_methods` describes the supported OEM installation procedures.
  The legacy KIWI version also provided a method to install an image partition based.
  Instead of selecting a target disk the user selects target partitions. Target partitions could be on 
  several disks. Each partition of an image had to be mapped on a selectable target partition.
  In the end it has turned out that a lot of user experience in a very sensitive area of
  the operating system is required. Combined with a high risk to do the wrong thing and more
  user interaction to install the system. This is contrary to the idea of images which should
  allow to give the user a functional entity at hand, dump it and be happy. Thus the partition
  based install method will not be continued.
```

Thoughts ?

schaefi:
@tomschr I think if you have a minute your feedback would be a great fit :)

tomschr:
I guess, you mean that I review the text in https://github.com/SUSE/kiwi/pull/614#issuecomment-363358226. Ok, here are my suggestions:

> [...]
>  The legacy KIWI version also provided a method to install an image partition based.

should that be `an image based partition`? The other way around sounds weird to me.

>  Instead of selecting a target disk the user selects target partitions.
> Target partitions could be on several disks. 

For me, there is verb missing after "could be on". Maybe "located"? Or rewrite it as `can be available on`?

> In the end it has turned out that a lot of user experience in a very sensitive area of
  the operating system is required. Combined with a high risk to do the wrong thing and more
  user interaction to install the system. 

IMHO this is a bit too verbose and can be simplified. What about replacing it with the following sentence:

```
It turned out, users needed a lot of experience in a very sensitive
area of the operating system.
```

> This is contrary to the idea of images which should allow to give the user a 
> functional entity at hand, dump it and be happy.

Although I know what you mean by "functional entity", 😉  I would try to simplify the sentence. What about this:

```
This is contrary to the idea of images to be dumped and be happy.
```

schaefi:
> should that be an image based partition? The other way around sounds weird to me.

no it really means an installation based on the partitions of the disk image. Assume an image consists out of 3 partitions then the partition based installation maps each of those to another disk partition on the target disk or disks

schaefi:
ok so here is an updated version:

```
* Partition based install method in OEM install image

  The section :ref:`deploy_methods` describes the supported OEM installation
  procedures. The legacy KIWI version also provided a method to install
  an image based on the partitions of the OEM disk image.
  
  Instead of selecting one target disk to dump the entire image file to,
  the user selects target partitions. Target partitions could be located
  on several disks. Each partition of the OEM disk image must be mapped
  on a selectable target partition.

  It turned out, users needed a lot of experience in a very sensitive
  area of the operating system. This is contrary to the idea of images
  to be dumped and be happy.

  Thus the partition based install method will not be continued.
```

@M0ses how about this one ?

M0ses:
LGTM - especially the link will guide to the remaining installation options, so I can have a look only one click away.

schaefi:
@M0ses ok great, so seems we have an agreement. Could you then take the text and update the doc sources ?

Please have in mind that the ref "deploy_methods" does not yet exist. Therefore also add the following in ```doc/source/building/build_oem_disk.rst``` right before the "Deployment Methods" section

```
.. _deployment_methods
```

schaefi:
As I'm currently working on the docs I updated the dropped feature list already. Hope that's ok ?

schaefi:
closing on behalf of #616

M0ses:
THX

PR #615, opened 2018-02-07
Delete obsolete ovf format value

schaefi:
Support for ova has been added which makes the ovf format
value obsolete and redundant


schaefi:
yes we can add a runtime check if someone gets annoyed by a late error ;)

michal42:
Thanks! 👍

PR #617, opened 2018-02-09
Fixed cleanup of iso mount paths

schaefi:
When an iso file is used as repo, this iso will be loop mounted on
the host and bind mounted into the image root as long as the image
builds. When the mount is released a recursive cleanup of the complete
path happens. In the original code the toplevel of the mount path
was in /tmp. If the /tmp directory is empty at the time of the
cleanup also the toplevel directory will be deleted which leaves
a Linux system without a /tmp directory, causing problems because
this is expected to exist always. Thus this patch creates an extra
toplevel directory to manage iso repo mounts which is not part of
the core FHS standard. This fixes bsc#1080301

### DIFF
--- a/kiwi/path.py
+++ b/kiwi/path.py
@@ -90,13 +90,27 @@ class Path(object):
     def remove_hierarchy(self, path):
         """
         Recursively remove an empty path and its sub directories
-        ignore non empty paths and leave them untouched
+        ignore non empty or protected paths and leave them untouched
 
         :param string path: path name
         """
         Command.run(
-            ['rmdir', '-p', '--ignore-fail-on-non-empty', path]
+            ['rmdir', '--ignore-fail-on-non-empty', path]
         )
+        path_elements = path.split(os.sep)
+        protected_elements = [
+            'boot', 'dev', 'proc', 'run', 'sys', 'tmp'
+        ]
+        for path_index in reversed(range(0, len(path_elements))):
+            sub_path = os.sep.join(path_elements[0:path_index])
+            if path_elements[path_index - 1] in protected_elements:
+                log.warning(
+                    'remove_hierarchy: path {0} is protected'.format(sub_path)
+                )
+                return
+            Command.run(
+                ['rmdir', '--ignore-fail-on-non-empty', sub_path]
+            )
 
     @classmethod
     def which(
--- a/test/unit/path_test.py
+++ b/test/unit/path_test.py
@@ -1,4 +1,4 @@
-from mock import patch
+from mock import patch, call
 
 import os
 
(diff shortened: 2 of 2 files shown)


schaefi:
references: https://bugzilla.suse.com/show_bug.cgi?id=1080301

davidcassany:
@schaefi, I am having difficulties to understand how it can be that this only effects the iso mount tempdir. Couldn't it also happen in other mounts in /tmp? in btrfs and lvm code for example I think we have mountpoints over temp directories created in /tmp.

Haven't read carefully the code of lvm and btrfs, but it looks to me that the mentioned issue could also happen in other places (maybe it doesn't happen because we are lucky enough to have other folders around, but it doesn't make me feel safe).

schaefi:
>    [1]@schaefi, I am having difficulties to understand how it can be that
>    this only effects the iso mount tempdir. Couldn't it also happen in
>    other mounts in /tmp? in btrfs and lvm code for example I think we have
>    mountpoints over temp directories created in /tmp.

So the problem exists if Path.remove_hierarchy() is called
and the toplevel directory is empty but should stay in the system.

Path.remove_hierarchy() is called on cleanup of root_bind.py and walks
through all entries of mount_stack.

mount_stack contains mounts below Defaults.get_shared_cache_location()
which is a directory we don't want to keep in the system, which is why
Path.remove_hierarchy is by intention correctly working when it removes
this part.

mount_stack is also changed on invocaton of root_bind's mount_shared_directory
method. This method is called for any host directory which needs to be
available for some reason inside of the system root during build. For
the loop mounted iso repo tmppath this is the case and leads to the
described problem.

I could not find any other mkdtemp created host directory which is bind
mounted to the system root except for loop iso files. Which is the reason
why I just fixed this part.

You are right if there are more of this in the code the fix would be
incomplete. But I see mount_shared_directory only called in the repository
classes which imho should also be the only code parts which needs it

I could imagine if someone has a repo definition like this

```xml
<repository type="rpm-md">
    <source path="/tmp/foo/bar"/>
</repository>
```

it would result in /tmp/foo/bar to be bind mounted to <sysroot>/tmp/foo/bar
and could cause removal of <sysroot>/tmp if empty at cleanup time

Thus we could consider to rewerite the patch in a way that we exclude
top level directories defined in FHS to be deleted by Path.remove_hierarchy()

Thoughts ?

davidcassany:
> I could not find any other mkdtemp created host directory which is bind mounted to the system root except for loop iso files. Which is the reason why I just fixed this part. 

I see, I did not realize when I checked the patch that this was happening only in the context of bind mounted folders. Thus I just did a grep of mkdtemp calls leading to a mounted directory and I saw some. This narrows a lot the options of having some bad behavior, but I still see it little frightening. 

IMHO, the best would be to have Path.remove_hierarchy() patched, also to make sure it is not misused in the future or that we do not encounter another specific corner case where it causes issues. Limiting the removal to FHS sounds good to me.

schaefi:
ok let me prepare that

schaefi:
@davidcassany ok I refactored the code

let me say that this will prevent deletion of any protected sub path members but this will also apply if those members are not on the toplevel of the system image root directory. I think this is a weakness we could live with which also does not cause damage to the system. Thoughts ?

PR #621, opened 2018-02-13
Update boot description search path

schaefi:
With the move of the boot descriptions in its own package
the kiwi search path needs to be adapted to look for boot
descriptions in /usr/share/kiwi/custom_boot. In addition
a runtime check to exit early if the boot description could
not be found or is not specified will be provided. Related
to Issue #576

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -791,12 +791,12 @@ class Defaults(object):
     @classmethod
     def get_boot_image_description_path(self):
         """
-        Implements bootloader path for ISO images
+        Implements path to find custom kiwi boot descriptions
 
         :return: directory path
         :rtype: string
         """
-        return Defaults.project_file('boot/arch/' + platform.machine())
+        return '/usr/share/kiwi/custom_boot'
 
     @classmethod
     def get_boot_image_strip_file(self):
--- a/kiwi/runtime_checker.py
+++ b/kiwi/runtime_checker.py
@@ -182,6 +182,55 @@ class RuntimeChecker(object):
                             )
                         )
 
+    def check_boot_description_exists(self):
+        """
+        If a kiwi initrd is used, a lookup to the specified boot
+        description is done and fails early if it does not exist
+        """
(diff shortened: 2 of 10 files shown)


kiwi/defaults.py
```diff
@@ -791,12 +791,12 @@ def get_network_image_types(self):
...
     def get_boot_image_description_path(self):
         """
-        Implements bootloader path for ISO images
+        Implements path to find custom kiwi boot descriptions
 
         :return: directory path
         :rtype: string
         """
-        return Defaults.project_file('boot/arch/' + platform.machine())
+        return '/usr/share/kiwi/custom_boot'
```
davidcassany: I am confused about the architecture subfolder, are we expecting to include it in the boot attribute, something like: `boot="x86_64/oemboot/suse-leap-42.2"`? Asking because I cannot see anywhere in this PR the dynamic assignment of the arch in the path, as it used to be. The kiwi-boot-descriptions I understand is a noarch package that has the descriptions for any supported arch in a specific subfolder.

schaefi: ok let me try to resolve the confusion :)

1. The package kiwi-boot-descriptions installs description to /usr/share/kiwi/custom_boot as we discussed it. That's why the path changed

2. The package kiwi-boot-descriptions is an arch specific package, it's not noarch. If you build on e.g s390 you would install the s390 version of kiwi-boot-description and that only provides the s390 available boot descriptions. The Makefile of the kiwi-boot-descriptions package does the arch distinction. That's why we don't install into an arch folder.

3. The context of the boot= attribute does not change. It stays at e.g boot=oemboot/suse-leap42.3 which I also have tested in a local build to work correctly

4. cross arch building is something we never supported with the custom kiwi boot description.
    If we do so in the future it will surely be done via dracut

Makes sense ?

davidcassany: Thanks for the clarification, now it all makes sense. I did not notice the Makefile of kiwi-boot-descriptions was doing such a distinction.

PR #622, opened 2018-02-13
Create compatible boot options

schaefi:
The boot option root=install:CDLABEL= is mandatory for install
images which uses the dracut initrd system. But for the custom
kiwi oemboot descriptions this is causing a problem when detecting
the install device. Thus the above boot option is only applied
for the initrd system which actually makes use of it


schaefi:
Found this while testing dracut and custom kiwi boot configured images.

PR #626, opened 2018-02-15
Deleted obsolete boot descriptions

schaefi:
The custom kiwi boot descriptions has been moved into
the kiwi-descriptions github repo and builds the compat
package kiwi-boot-descriptions from there. The build
of the boot image(initrd) is done by dracut and the
dracut module packages provided by kiwi. The classic
custom boot descriptions can still be used as alternative
method if the above package is installed. Related to
Issue #576


schaefi:
After a long process we finally make the first step to move out the old boot descriptions

schaefi:
yes no dependency. I will talk to the obs people on Monday/Tuesday next week when I'm in Nuremberg
about deps and provides and such

Issue #623, opened 2018-02-14
EFI Live images are missing the grub2 theme


Vogtinator:
https://openqa.opensuse.org/tests/609090#step/bootloader_uefi/4

The ISO's `/boot/grub2/themes` directory is empty, because of:

```
[  644s] [ DEBUG   ]: 05:03:05 | EXEC: [mkdir -p /usr/src/packages/KIWI-iso/live-media.8wyjcz82/boot/grub2/themes]
[  644s] [ WARNING ]: 05:03:05 | Theme /usr/src/packages/KIWI-iso/live-media.8wyjcz82/boot/grub2/themes/openSUSE not found
```
I looks like it uses the wrong source path for the copy.

schaefi:
Can not reproduce and works from a test build. Can you make sure your image description installs the package:

```xml
<package name="grub2-branding-openSUSE"/>
```

schaefi:
accessing openqa timed out

Vogtinator:
It's installed just fine and appears in the root filesystem.

`<package name="grub2-branding-openSUSE" bootinclude="true"/>`.

openQA might be up again now.

schaefi:
ah well ok that link doesn't help. For reproducing we need:

1. the complete image description
2. the complete build log

Thanks

Vogtinator:
That's on OBS: https://build.opensuse.org/package/show/openSUSE:Leap:15.0:Images/livecd-leap-kde

schaefi:
Are you sure this is a kiwi error. The log shows:

```
[  493s] [ DEBUG   ]: 15:51:23 | system: (1223/1579) Installing: grub2-branding-openSUSE-15.0-lp150.1.3.noarch [.........done]
[  493s] [ INFO    ]: Processing: [#########                               ] 24%[ DEBUG   ]: 15:51:23 | system: Additional rpm output:
[  493s] [ DEBUG   ]: 15:51:23 | system: warning: /var/cache/kiwi/packages/3fcba24e373c1066020c42d21b671ce7/grub2-branding-openSUSE.rpm: Header V3 RSA/SHA256 Signature, key ID 3dbdc284: NOKEY
[  493s] [ DEBUG   ]: 15:51:23 | system: Use of uninitialized value $gfx_mode in concatenation (.) or string at /usr/share/grub2/themes/openSUSE/activate-theme line 35.
```

schaefi:
[...]

PR #627, opened 2018-02-15
Fixup grub2 theme setup

schaefi:
The presence of a background file was mandatory for kiwi to
use the theme. But the background information is optional

This is related to Issue #623


schaefi:
Tested the fix with themes with and without background data and worked for me. I'll go and merge this because I'd like to get a new package to Staging and see if all still works

PR #629, opened 2018-02-16
Do not emit linuxefi/initrdefi on non-x86 platforms

agraf:
Grub wants to have linuxefi/initrdefi commands to execute Linux via the UEFI
calling convention on x86. However, all other platforms simply use the normal
linux/initrd commands, because there grub assumes that an EFI grub wants to
run an EFI payload.

Reflect this architecture difference in the generated grub configuration.

Signed-off-by: Alexander Graf <agraf@suse.de>


schaefi:
Thanks looks great 👍

PR #632, opened 2018-02-20
APT: Run debootstrap with --variant=minbase

michal42:
This selects only packages with Priority: required, which is more in
line with what the other package managers are doing. A patch for
kiwi-descriptions has been submitted to adapt to this new behavior.

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -54,6 +54,7 @@ class PackageManagerApt(PackageManagerBase):
         self.custom_args = custom_args
         if not custom_args:
             self.custom_args = []
+        self.deboostrap_minbase = True
 
         runtime_config = self.repository.runtime_config()
         self.apt_get_args = runtime_config['apt_get_args']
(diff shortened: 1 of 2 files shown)


michal42:
https://github.com/SUSE/kiwi-descriptions/pull/17 is the kiwi-descriptions fixup

schaefi:
I like the change to make "onlyRequired" the default behavior and the code change is good

What came to my mind is; In the packages section we support the attribute patternType="onlyRequired" | "plusRecommended", I think it would be great if this patch could be extended in a way which allows the selection of more than Priority: required packages if the patternType is set to another value so that we don't loose the possibility we had before

Thoughts ?

schaefi:
I was thinking on mapping patternType="plusRecommended" to e.g scratchbox variant or none variant. But I'm not sure if that would be ok. Alternatively patternType could be extended to allow the variant types and if not set no variant will be set

davidcassany:
I like the idea to mimic somehow the patternType="plusRecommended" to some or none variant and patternType="onlyRequired" to minbase variant.

For plusRecommended, I think I prefer to leave it without any variant and just use defaults. I guess that having the minbase variant as default is enough, since you are always capable to easily add packages to your description; but is hard to remove packages that have been already included without being nasty. 

Also not sure if fakechroot is smaller than minbase, does anyone have a clue? fakechroot sounds also like a good candidate for the default bootstrap variant.

In short, my opinion would be to leave as a default (onlyRequired) the most minimalist option and leave it without any variant for plusRecommended. We could also plan to support some way to set the variant from the description file and support all of them,; but I would not go that way, I don't think there is a need.

michal42:
Thanks for the review. I never used plusRecommended, but I can have a look.

schaefi:
@michal42 just to put me right, you still plan to update your code with the suggested changes right ?

Thanks

michal42:
Sorry for the delay. Yes, that has been the plan. I pushed the change now, let me know what you think and enjoy your weekend.

Michal

schaefi:
yeah this looks great to me :+1:

PR #636, opened 2018-02-22
Update build-test descriptions for Kanku

schaefi:
As part of the buildservice there is now also a CI system called Kanku
which allows for image boot/run tests. The only requirement on the image
level which is missing in our build test descriptions is the setup of
the serial console. This patch updates those image descriptions which
could be tested by kanku to provide a serial console at boot time.

The plan is that our build-test images automatically gets boot tested
by the Kanku CI. This should include boot, console-login, reboot,
console login, shutdown. The required job description to do this
needs to be worked out together with the Kanku team

### DIFF
--- a/build-tests/x86/test-image-fedora-vmx/appliance.kiwi
+++ b/build-tests/x86/test-image-fedora-vmx/appliance.kiwi
@@ -15,7 +15,7 @@
         <keytable>us</keytable>
         <timezone>UTC</timezone>
         <rpm-check-signatures>false</rpm-check-signatures>
-        <type image="vmx" filesystem="ext3" bootloader="grub2" kernelcmdline="splash"/>
+        <type image="vmx" filesystem="ext3" bootloader="grub2" kernelcmdline="console=ttyS0 splash"/>
     </preferences>
     <users>
         <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root" groups="root"/>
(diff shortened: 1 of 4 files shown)


schaefi:
@davidcassany I took part in a presentation about Kanku while I was in NUE and I think this could be used for our integration tests. If you like you can get an account on kanku.suse.de which would be needed to run the ```kanku``` tool to ask for the status of the job

So far they have not created an api which would allow us to create our own job template. The web interface allows to modify existing jobs but not creating new ones. But that would not be a show stopper for me at the moment. I have direct connection to the people implementing this, thus they will hook our image description in and we gain a way to check boot results from any new image we build

Thoughts ?

schaefi:
The whole thing is based on libvirt

Conan-Kudo:
Umm... What is Kanku?

schaefi:
The buildservice team implemented this as a service for image builds. So far it exists internally and should iirc be available through the obs api later. As part of the obs build data it should be possible to also provide a job description for the kanku service. On success obs will then take this job together with the build image and run it. Running for an image build means boot and do something with it

I consider this a lightweight CI system for obs build results, which I think is a nice idea. For us this is low hanging fruits because all we need is a serial console configured

I think there will also be  a talk on the opensuse conference about this

schaefi:
as we build our images in the build service it would be nice to extend the scope of our test builds to be actually also booted.

davidcassany:
@schaefi, All that sounds wonderful to me, you know I'd like to have some kind of functional testing. I will have a look to the kanku tool, but it sounds really promising, specially having it integrated with OBS, thats awsome. I am so curious to know more about that tool.

Thanks for the update

M0ses:
For more info about kanku:

https://m0ses.github.io/kanku/

We use it in the OBS backend team to test our OBS-Appliance images and make it work with our images easier.

But please be aware that the documentation is a bit outdated.

PR #641, opened 2018-02-23
Update tox.ini to use python3

AdamMajer:
We should be using python3 instead of specific minor version
like python3.4

Fixes tests.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -29,12 +29,12 @@ whitelist_externals =
     /usr/bin/shellcheck
     /bin/bash
 basepython =
-    {check,doc,doc_travis,doc_travis_deploy}: python3.4
-    unit_py3: python3.4
+    {check,doc,doc_travis,doc_travis_deploy}: python3
+    unit_py3: python3
     unit_py2: python2.7
 envdir =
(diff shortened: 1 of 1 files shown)


schaefi:
yeah sorry don't like this. There are differences between minor versions of python3 which could be relevant. I want our tests to not weaken this and I agree to what David said

schaefi:
Since the discussion stopped I tried to summarize the feedback in another request: #646 
Let me know what you think

schaefi:
Closing due to #646 merged

AdamMajer:
On 02/27/2018 11:25 AM, Marcus Schäfer wrote:
> Since the discussion stopped I tried to summarize the feedback in
> another request: #646 <https://github.com/SUSE/kiwi/pull/646>
> Let me know what you think

Sorry didn't see earlier comment (no email from Github?)

Anyway, this doesn't really matter to me. It's just that when I started
changing python with the "contributing" instructions, it failed because
python3.6 is current, not 3.4.

Maybe I didn't look closely enough but there doesn't seem to be
difference in tests for different python versions?

- Adam

PR #643, opened 2018-02-23
Remove genisoimage

### LINKED ISSUE #642: Remove dependency on wodim for openSUSE:Factory
## Problem description
`wodim` has no maintenance for many years. Since we are shipping `cdrtools` and `kiwi` supports both `cdrtools` *and* `wodim`, we can just change dependencies.

This is important to remove `wodim` from upcoming SLE-15

## Expected behaviour
No bahaviour changes expected but wodim will no longer be required.

+ KIWI version: 9.12.6
+ Operating system: openSUSE:Factory
+ OBS version:

AdamMajer:
Fixes #642 

* Use mkisofs instead of genisoimage for openSUSE:Factory and SLE-15

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -136,7 +136,11 @@ Requires:       gdisk
 %endif
 Requires:       dosfstools
 Requires:       e2fsprogs
+%if 0%{?suse_version} >= 1500
+Requires:       mkisofs
+%else
 Requires:       genisoimage
+%endif
 Requires:       grub2
 Requires:       kiwi-man-pages
 Requires:       kiwi-tools
@@ -210,7 +214,11 @@ Provides:       kiwi-packagemanager:zypper
 %endif
 Requires:       dosfstools
 Requires:       e2fsprogs
+%if 0%{?suse_version} >= 1500
+Requires:       mkisofs
+%else
 Requires:       genisoimage
+%endif
 Requires:       grub2
 Requires:       kiwi-man-pages
 Requires:       kiwi-tools
@@ -372,10 +380,14 @@ Requires:       dmsetup
 Requires:       device-mapper
 %endif
 Requires:       dracut
+%if 0%{?suse_version} >= 1500
+Requires:       mkisofs
+%else
 %if 0%{?fedora} || 0%{?rhel} || 0%{?debian} || 0%{?ubuntu}
 Requires:       genisoimage
 %else
 Requires:       cdrkit-cdrtools-compat
+%endif
 %endif
 License:        GPL-3.0+
 Group:          %{sysgroup}


Conan-Kudo:
This breaks KIWI on Fedora, as we have multiple `mkisofs` providers with slightly incompatible options supported.

schaefi:
Uhm, I have several concerns touching this iso creation hell of things.

First of all the motivation for the change is to get rid of the wodim requirement but that is not relevant from a functional perspective. Thus can you guarantee mkisofs on sle15 produces correct results compared to genisoimage ? I know our opensuse live images for Factory, which is still almost the same than sle15 are created using genisoimage and the result is ok

We already suffered so much from the iso creation tool crap (sorry) that I'm really not motivated to align for any change unless there are good reasons and no breakage

* if you can point me to an image build with the sle15 mkisofs that produces correct results I would be ok.

* Other than that, could you revert the changes in the tests, they are unrelated to the spec file changes and if I got it right this is the only thing we need to change

Thanks and sorry but with isofs stuff we suffered so much that we are really careful with changes

AdamMajer:
OK, I've removed changes unrelated to the spec file template.

I can't guarantee anything - the world could be swallowed by a black hole tomorrow for all we know :smile:  But I have changed SLE-15 image building stuff to use mkisofs instead of genisoimage on IBS. You can see that in my project at https://build.suse.de/project/show/home:adamm:branches:SUSE:SLE-15:GA



So far I've tested the x86_64 images and they boot and install correctly including the memtest86.

AdamMajer:
The reason is not to have a package like `wodim` without upstream in SLE15 for many years to come. And since we already have `cdrtools` there and kiwi works with it, then removing dependency on `wodim` and removing `wodim` from the release makes sense.

schaefi:
I see only product builds:

```
iosc list home:adamm:branches:SUSE:SLE-15:GA

000product
product-builder
python-kiwi
```

What has kiwi to do with product iso's. imho that's all covered by the new product builder project which as far as I know is a perl project with some forked sources from kiwi v7 ?

For product-builder.spec you can control the requirements there not in kiwi. That's confusing

With regards to mkisofs vs. genisoimage there is a thread on irc which states that genisoimage contains patches to correctly work with hybrid iso's (isohybrid) which is what we use for any bootable iso we create. mkisofs seems to miss that patches

It still looks like solving your issue and creating new ones for us ;)

AdamMajer:
"thread on irc" doesn't help me. A link would help? Is there a mailing list you can point me to?

http://www.syslinux.org/wiki/index.php?title=Isohybrid

syslinux wiki seems to indicate it works with both.

Yes, I would like to remove a problem of no-upstream supported packages. Because then the "expert" for the package becomes someone like me that is not remotely an expert on those things. I don't know, but I think that's a problem for us. ;)

I'll get back to you about images with python3-kiwi.

schaefi:
> "thread on irc" doesn't help me.

right which was the reason why I summarized the important bits, which are that genisoimage contains patches to work with hybrid iso's which are important for bootable images (also affecting your product iso's) Those patches are not part of mkisofs if I got it right

>Yes, I would like to remove a problem of no-upstream supported packages. Because then the >"expert" for the package becomes someone like me that is not remotely an expert on those things. I >don't know, but I think that's a problem for us. ;)

It wouldn't be the first time that someone gets an export for stuff he or she is not an expert in. I remember myself being a poor maintainer of nettools for years. I took over the package because the real expert left the company ;) Don't get me wrong I got your point and I would do the same to get rid of such maintenance tasks. However I hope you understand that we are also not happy to receive bugs by changes in the tool chain we then even can't revert because the tool which was used to work before is gone

> I'll get back to you about images with python3-kiwi.

That would be great. If you can setup a live iso build using the sle15 code stream with mkisofs in the obs worker instead of genisoimage we have a chance to run extensive testing with bios/efi hybrid/non-hybrid, persistent/non-persistent boot tests. As as base you can branch from: 

* Virtualization:Appliances:Images:Testing_x86/test-image-iso

Thanks much

AdamMajer:
I've played around with this and I've created a "Dummy Product" image based on SLE-15 repository that boots with both qemu BIOS and UEFI (no secure boot) using OVMF. Of course, there is no installer or things like that since it's just "Dummy Product"

https://build.suse.de/package/show/home:adamm:branches:SUSE:SLE-15:GA:test_images/test-image-iso

I had to add a parameters to kiwi for the UEFI alt-boot to be set correctly. By default, -eltorito-platform defaults to x86_64 PC platform, and you can set it to others like PPC, Mac or efi or some numeric one if you want.

schaefi:
Thanks for the setup. I prepared a patch for the platform value which seems we have just missed so far. I'll give your build a test, thanks

schaefi:
I have tested your build and it looks good to me. I think if the platform patch could be merged we can also merge this one

PR #644, opened 2018-02-26
Cleanup shell code

schaefi:
Cleanup config functions shell code and integrate it to the code quality test target in tox via shellcheck
Also delete obsolete/unused methods


schaefi:
> Having it all under shellcheck radar will be very handy.

yes finally

PR #649, opened 2018-02-28
Fixed get_free_disk_bytes in dracut kiwi-lib

### LINKED ISSUE #648: get_free_disk_bytes method from kiwi-partitions-lib dracut module does wrong calculation
## Problem description

Get_free_disk_bytes does a wrong calculation of the free bytes. In order to get the free available space it makes the difference between the disk size and the addition of all partitions size, this is correct bbut `lsblk` can list the devices twice (meaning space used by partitions is counted twice) due to some device mapper and/or multipath issue.

## Expected behaviour

`lsblk` output just list the partitions once or the output is filtered in some way.

- Build `suse/x86_64/suse-leap-42.3-JeOS/` oem image from kiwi-descriptions repository. 
- Resize the resulting image with `kiwi image resize` to 8GB.
- Boot the image with qemu, `qemu-kvm -m 2048 -hda LimeJeOS-Leap-42.3.x86_64-1.42.3.raw`
- At grub stage edit the kernel command line in order to add a pre-mount break point (add `rd.break=pre-mount rd.shell`) and boot
- At the pre-mount shell run `lsblk -r -p -o NAME,TYPE  /dev/sda`

Then you'll see how the output includes `/dev/sda[1-3]` partitions and it includes also some `/dev/mapper/QEMU_HARDDISK_QM00001-part[1-3]` partitions which are linked to some `/dev/dm-[0-3]` devices. No idea why are those devices there since we do not have any LVM configuration and also they are just present at this boot stage, they disappear after (at switch-root stage they are not there anymore).

+ KIWI version: 9.13.1
+ Operating system: Leap 42.2
+ OBS version: None

schaefi:
The method used any device from the lsblk output, but it can
happen that some device nodes with different name point to
the same physical device. The method would then calculate
the free space wrong. This patch takes the PARTUUID value
into account to make sure any partition device size is used
only once. This Fixes #648

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -267,10 +267,26 @@ function get_free_disk_bytes {
     local rest_bytes
     rest_bytes=${disk_bytes}
     local part_bytes=0
+    local part_count=0
+    local part_uuids
     for part in $(
         lsblk -p -r -o NAME,TYPE "${disk}" | grep part | cut -f1 -d ' '
     );do
-        part_bytes=$((part_bytes + $(blockdev --getsize64 "${part}")))
+        current_part_uuid=$(get_partition_uuid "${part}")
+        for part_uuid in ${part_uuids[*]};do
+            if [ "${current_part_uuid}" = "${part_uuid}" ];then
+                # this partition uuid was already handled. The device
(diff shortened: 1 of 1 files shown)


schaefi:
@davidcassany I have not extensively tested this, thus would be great if you can give it a try. From a testing perspective I suggest to build the image once and then just copy the dracut modules from source into the root tree and rebuild with --allow-existing-root

Thanks

PR #651, opened 2018-03-01
Set eltorito-platform for efi

schaefi:
The default eltorito platform is set to "x86_64 PC", however
in an alternative bootloader spec for efi this would be the
wrong platform spec. This patch adds the correct platform
to the alt-boot setup for efi. References #643

### DIFF
--- a/kiwi/filesystem/isofs.py
+++ b/kiwi/filesystem/isofs.py
@@ -19,9 +19,6 @@
 from kiwi.filesystem.base import FileSystemBase
 from kiwi.command import Command
 from kiwi.iso import Iso
-from kiwi.path import Path
-
-from kiwi.exceptions import KiwiIsoToolError
 
 
 class FileSystemIsoFs(FileSystemBase):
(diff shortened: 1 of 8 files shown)


AdamMajer:
The problem is that this will fail with genisoimage as it doesn't have this option. It's a mkisofs specific option.

AdamMajer:
There was a patch added to wodim for UFI but then it was removed because it caused unbootable USB sticks. boo#939456

schaefi:
yeah and again the shit with the iso creation tools starts. All this is just a big mess and we suffered from it with every simple change :( I would love to have one tool which just does it right, maybe xorriso can do that in the future and we can get rid of all other crap

I'll modify the patch to make it work for both

davidcassany:
Yes all this tool mess sucks, in not that long period we had to update it several times...

schaefi:
just saw that some commands print their help on stderr not on stdout, guess what mkisofs is one of those. Thus we need to make CommandCapabilities aware of it by also checking data on command.error and not only on command.output

I'll do that here in an extra commit

never ending story

schaefi:
@davidcassany could you have a look at this one again

while I added the ability for has_option_in_help to also check the error channel I stumbled over failed tests in the ova area and that pointed me to a code part in the ovftool options check which I think needs to be refactored too

Thanks

PR #652, opened 2018-03-01
Disable multipath support when kiwi-oem-repar is active

### LINKED ISSUE #645: kiwi-repart dracut module looks like it gets interrupted
## Problem description

Installing kiwi-repart and kiwi-dump dracut modules causes Tumbleweed images to not boot and some erratic behavior in Leap 42.3 images (repartition is not done properly)

## Expected behaviour

Tumbleweed images should boot and Leap 42.3 should boot with the appropriate repartition done. Up to now I could trace that installing `dracut-kiwi-oem-repart` and `dracut-kiwi-oem-dump` together causes this issue, intalling only `dracut-kiwi-oem-repart` produces the expected behavior. Looks like there is some sort of wrong interaction with these two modules.

Build the oem image from the kiwi-descriptions (Leap 42.3) or the Tumbleweed image from `build-tests/x86/test-image-oem/`

+ KIWI version: 9.13.1
+ Operating system: Leap 42.2
+ OBS version: None

davidcassany:
This commit fixes #645. With this commit there is no interction between
multipathd service and kiwi-oem-repart module, this way there are no
issues with devices appearing and disappearing. This this commit during
firstboot there will not be multipath support.

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -650,7 +650,7 @@ class DiskBuilder(object):
             'dracut_rescue_image="no"'
         ]
         dracut_modules = []
-        dracut_modules_omit = ['kiwi-live', 'kiwi-dump']
+        dracut_modules_omit = ['kiwi-live', 'kiwi-dump', 'multipath']
         if self.root_filesystem_is_overlay:
             dracut_modules.append('kiwi-overlay')
         else:
--- a/test/unit/builder_disk_test.py
+++ b/test/unit/builder_disk_test.py
@@ -467,7 +467,7 @@ class TestDiskBuilder(object):
             call('dracut_rescue_image="no"\n'),
             # before dracut is called, image dracut setup
             call('add_dracutmodules+=" kiwi-lib kiwi-repart "\n'),
-            call('omit_dracutmodules+=" kiwi-live kiwi-dump kiwi-overlay "\n'),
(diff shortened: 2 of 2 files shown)


davidcassany:
Too nice to be true, my resistance to accept that after that hard debugging it could be solved that easily turned to be a good ally. This apparently works, but just under some circumstances. I tested the patch in pxe and iso installs with success, also booting from a disk when no image dump is required.

Omitting multipath results in not having multipathd started from initrd even the multipath-tools package is installed, because the binary tools are simply not included into the initrd image. However nothing prevents to starting the service from the user space once booted with a simple `systemctl start multipathd`, then multipathd fails to setup a device map, the service keeps running, but reports an error and no device map created. And things get worse if someone tries to enable the service at boot. It gets started after the sysroot.mount and hangs awaiting a device that never appears, after the timeout it just fails and leads to a rescue shell.

This option is NOT good. I believe we should try to make multipath module and kiwi-repart module compatible, otherwise I have the feeling the issue it will just strike back again and again.

davidcassany:
We could get rid of this by recreating the initrd, so next boot would not have the multipath omited and kiwi-repart present, but I believe this is not what we want, that would make first boot to take quite long.

davidcassany:
Closing this PR, as the solution will come in another PR with a different approach.

PR #653, opened 2018-03-02
Reactivate device mappings after repart. Fixes #645

### LINKED ISSUE #645: kiwi-repart dracut module looks like it gets interrupted

davidcassany:
This commit recreates removed device maps and triggers again
systemd generators in order to adapt again device units to the
new context. Fixes #645

### DIFF
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -77,6 +77,13 @@ function deactivate_device_mappings {
     deactivate_all_device_maps
 }
 
+function activate_device_mappings {
+    if type multipath &> /dev/null; then
+        multipath -r
+        systemctl daemon-reload
+    fi
+}
+
 function finalize_disk_repart {
     declare kiwi_RootPart=${kiwi_RootPart}
     finalize_partition_table "${disk}"
@@ -113,14 +120,14 @@ function repart_standard_disk {
     if [ "${kiwi_oemswap}" = "true" ];then
         new_parts=$((new_parts + 1))
     fi
-    # deactivate all active device mappings
-    deactivate_device_mappings
     # check if we can repart this disk
     if ! check_repart_possible \
         ${disk_root_mbytes} ${disk_free_mbytes} ${min_additional_mbytes}
     then
         return 1
     fi
+    # deactivate all active device mappings
+    deactivate_device_mappings
     # repart root partition
     local command_query
     local root_part_size=+${disk_have_root_system_mbytes}M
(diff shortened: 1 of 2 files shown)


schaefi:
This looks brilliant :+1: 

you set the wip label, means you are still in testing phase ?

davidcassany:
Oh forgot to press the comment button for the message below... :-P

I am pretty confident this is a good solution, however I am finalizing some more tests. Currently I am having an issue with PXE deploy, but I think it is either by some issue in my env or another unrelated bug, because of this I have included the work in progress label.

@schaefi In fact I would appreciate if you could test this with a PXE deploy

schaefi:
I'll give it a test on Monday, currently software.opensuse.org is also down for maintenance... so no show at the moment :)

schaefi:
I'm now testing an oem pxe deployment with your changes

davidcassany:
Great, thanks much.
I don't think anymore it is an issue of my env, thus should experience the same issues (boot-efi mount fails). It is an odd failure, probably some race issue while trying to mount boot-efi, as it mounts perfectly well in the emergency shell and then continues to boot normally. What I can't figure out at the moment is why it only happens in PXE deployment and how the change I did could have caused this issue.

schaefi:
Hmm, I remember that I saw this in the past too but I can't safely reproduce it. The test right now did not show the problem, all works fine. Regarding the change I also don't see how the efi mount is directly related to the change

...hmm

davidcassany:
Damn, it works for you... hmmm I am rebuilding my tests now from scracth, for both, Leap and Tumbleweed, and see if I can recreate it.

schaefi:
I tested Leap42.3, will try tumbleweed too

schaefi:
ah with tumbleweed I can reproduce the error

schaefi:
oh no wait, forgot to apply your changes

schaefi:
Works great on TW too with your patches :) 

NAME="openSUSE Tumbleweed"
# VERSION="20180302"

PR #654, opened 2018-03-02
The order of the options for mkisofs matters

schaefi:
Setting -eltorito-platform after -b causes mkisofs to fail


schaefi:
Not really surprising isn't it

working with these iso tools is just a pain in the ass

PR #655, opened 2018-03-06
Resize partition table after image resize

### LINKED ISSUE #534: Allow resize of partition table in kiwi image resize command
## Problem description

kiwi image resize allows to resize a disk image but it does not allow to resize the partition table to the changed geometry. It would be a nice extension if the command would provide an option e.g

* --update-partition-table

So that after boot of the image the free space is really visible to the system

schaefi:
The command 'kiwi image resize' allows to resize the size
of a disk image. Depending on the partition table type it
is also required to resize the partition table inside of
the image to let the file size change become effective
This Fixes #534

### DIFF
--- a/kiwi/partitioner/base.py
+++ b/kiwi/partitioner/base.py
@@ -98,3 +98,11 @@ class PartitionerBase(object):
         Implementation in specialized partitioner class
         """
         raise NotImplementedError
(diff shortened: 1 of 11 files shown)


schaefi:
correct :)

PR #656, opened 2018-03-06
Fixed btrfs search path in earlyboot script

schaefi:
If kiwi generates its own efi image for the boot process
it does not setup the btrfs relative path setup in the
earlyboot script embedded into the generated efi image.
This has a bad impact on the file search because the
btrfs setup done in kiwi puts root below the @ volume
which we then need to specify of the relative lookup
is not activated. Fixes bsc#1082155


schaefi:
Also include a bit of refactoring, prefer format() over message modifiers

agraf:
Reviewed-by: Alexander Graf <agraf@suse.de>

PR #659, opened 2018-03-08
Do not require multipath in dracut dump module

schaefi:
The presence of multipath in the initrd seems to expect
a multipath based rootfs environment and causes conflicts
to the device setup if this is not the case. Thus a generic
module requirement can not be done in the kiwi dump
module. Customers need to explicitly add the module in
an overlay dracut.conf file if they plan to provide the
rootfs via multipath

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -105,13 +105,8 @@ class DiskBuilder(object):
         self.bootloader = xml_state.build_type.get_bootloader()
         self.initrd_system = xml_state.get_initrd_system()
         self.target_removable = xml_state.build_type.get_target_removable()
-
-        self.root_filesystem_is_multipath = False
-        self.oemconfig = self.xml_state.get_build_type_oemconfig_section()
-        if self.oemconfig and self.oemconfig.get_oem_multipath_scan():
-            self.root_filesystem_is_multipath = \
-                self.oemconfig.get_oem_multipath_scan()[0]
-
+        self.root_filesystem_is_multipath = \
+            xml_state.get_oemconfig_oem_multipath_scan()
         self.disk_setup = DiskSetup(
             xml_state, root_dir
         )
@@ -658,7 +653,7 @@ class DiskBuilder(object):
         ]
         dracut_modules = []
         dracut_modules_omit = ['kiwi-live', 'kiwi-dump']
-        if not self.root_filesystem_is_multipath:
+        if self.root_filesystem_is_multipath is False:
             dracut_modules_omit.append('multipath')
         if self.root_filesystem_is_overlay:
             dracut_modules.append('kiwi-overlay')
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -105,6 +105,13 @@ class DiskBuilder(object):
         self.bootloader = xml_state.build_type.get_bootloader()
         self.initrd_system = xml_state.get_initrd_system()
         self.target_removable = xml_state.build_type.get_target_removable()
+
+        self.root_filesystem_is_multipath = False
+        self.oemconfig = self.xml_state.get_build_type_oemconfig_section()
+        if self.oemconfig and self.oemconfig.get_oem_multipath_scan():
+            self.root_filesystem_is_multipath = \
+                self.oemconfig.get_oem_multipath_scan()[0]
+
         self.disk_setup = DiskSetup(
             xml_state, root_dir
         )
@@ -651,6 +658,8 @@ class DiskBuilder(object):
         ]
         dracut_modules = []
         dracut_modules_omit = ['kiwi-live', 'kiwi-dump']
+        if not self.root_filesystem_is_multipath:
+            dracut_modules_omit.append('multipath')
         if self.root_filesystem_is_overlay:
             dracut_modules.append('kiwi-overlay')
         else:
--- a/build-tests/x86/test-image-oem/appliance.kiwi
+++ b/build-tests/x86/test-image-oem/appliance.kiwi
@@ -20,6 +20,7 @@
             <oemconfig>
                 <oem-unattended>true</oem-unattended>
                 <oem-swapsize>1024</oem-swapsize>
+                <oem-multipath-scan>false</oem-multipath-scan>
             </oemconfig>
             <systemdisk>
                 <volume name="home"/>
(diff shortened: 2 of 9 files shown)


schaefi:
I think it was not a good idea of me to generically require the multipath module. With regards to the failed tests in Kanku which again seems to outline a timing issue between the mount target and what multipath does I suggest to delete the strict requirement

To me it feels like the multipath module needs some refactoring to better fit into dracut and needs some serialization with the other services

Thoughts ?

schaefi:
yes it does not solve the issue, it just provides an escape path

I also would like to understand why it sometimes fail for Frank. I was not able to reproduce the issue he had in any of my tests.

davidcassany:
I believe this is the same issue of boot-efi.mount I told you a couple of days ago, I had it with libvirt and rd.debug enabled. Using libvirt the emulated hardware differs from qemu defaults, at least with the defaults that my virt-manager has, I could see show different modules were loaded because of that, but so far I could not narrow any offending driver or module.

Conan-Kudo:
I think a smarter way would be to make it so that you install snippets that add modules based on features requested via the schema. Since you ultimately control the behavior of these things, you can do that.

schaefi:
Hmm, let me think about it, so kiwi calls dracut to create a generic initrd which means all modules installed gets incorporated. Given we know that the image has not specified the multipath feature
we could add in the dracut.conf.d/... part from kiwi

```
omit_dracutmodules+="multipath"
```

The initrd would then have no multipath support but we could still require it from a package perspective. We still have to delete multipath from the depends() method in the module-setup because I think you can't omit depending modules

Once the system has booted no multipath is active, if a kernel update or whatever else triggers the creation of a hostonly initrd (which is the default) multipath will not be added because not active

Could this work ?

I'll need to test it

schaefi:
ok I have changed and tested the way how we could disable the use of multipath if not requested by the image.

@davidcassany Let me know what you think

davidcassany:
Well in fact I guess this goes on the path of disabling multipath in the initrd. This caused me issues in the tests I did last week, but probably my tests are not valid anymore. What I tested it to enable multipathd with `systemctl enable multipathd` (not sure if it makes any sense though), but I think at that time we still had dmsetup deleting devices in all boots and probably something else.

Just for the record, I did some tests with an active multipath and `e2fsck` and it works if device mapper is used. The problem is that our partition detection and parsing scripts do not consider having multiple representations of the same device. The commands I tested do not complain of being busy if used with the appropriate device (/dev/dm-* and its label or uuid links, instead of /dev/sda*).

schaefi:
> Well in fact I guess this goes on the path of disabling multipath in the initrd. This caused me issues in
> the tests I did last week, but probably my tests are not valid anymore.

I think what you tried was disabling the module by systemctl but this code disables the module by not even incorporating into the initrd. In my tests that worked well. So you can have multipath installed but it is not taken into account

schaefi:
Regarding the device you are right. I intentionally lookup the unix device name in the lookup_disk_device_from_root() method. To be honest I don't know exactly why :) I think we could in a follow up request change this and use the dev link according to the id provided in root=

by default root is specified as its uuid but kiwi also allows to specify the device persistency e.g by-label and that would allow us to control which dev link will be used.

For a first shot I think it's good to not incorporate a dracut module when we know it is not going to be used. With regards to multipath we have a switch in the oemconfig to activate it, thus I like the idea from @Conan-Kudo to use this information in the module setup

What do you think ?

davidcassany:
> I think what you tried was disabling the module by systemctl but this code disables the module by not even incorporating into the initrd. In my tests that worked well. So you can have multipath installed but it is not taken into account

No, I tried exactly to omit multipath from initrd :) I'll redo what I tested, but I think it will just work fine. Yes I also like the idea of @Conan-Kudom, I think this PR is fine :+1:

schaefi:
Ok I'll wait for your test result before merge. Thanks

kiwi/builder/disk.py
```diff
@@ -105,6 +105,13 @@ def __init__(self, xml_state, target_dir, root_dir, custom_args=None):
         self.bootloader = xml_state.build_type.get_bootloader()
         self.initrd_system = xml_state.get_initrd_system()
         self.target_removable = xml_state.build_type.get_target_removable()
+
+        self.root_filesystem_is_multipath = False
```
davidcassany: Now I just thought that probably we should do it the other way round, I mean leaving multipath included/untouched in any case and omit it when oem-multipath-scan is set to false, being that the default value.

I am just wondering if we should omit multipath for vmx and isos. Probably it doesn't make much sense to disable it these contexts. Mostly because with this configuration there can't be a live iso or vmx with multipath activated, unless I am missing something.

davidcassany: Oh went too fast with the live iso, it should not apply the live iso since it does not use the disk builder, but it is the case for vmx, isn't it?

schaefi:
Now I'm confused :)

* So multipath in the initrd is only needed if the rootfs is presented as a multipath device. As we have a switch to enable this for oem images the package dracut-kiwi-oem-dump requires multipath to pull in the needed tools. The default setup however is without multipath and because we create a generic dracut initrd which pulls in all installed modules we actively need to omit the module by default. This is the reason why self.root_filesystem_is_multipath is set to False. From your review you want to set it to True which confuses me ?

* on simple vmx types I agree we would have currently no opportunity to activate the multipath module as the switch is oem only. I'll adapt the code in this regard

schaefi:
Hmm, maybe I see your point... will think abut it over lunch... my head is spinning :)

schaefi:
@davidcassany I think I got your point, do you mind to give it another review ? Thanks

davidcassany:
> Now I'm confused :)

So sorry wasn't my intention to introduce confusion :)

> The default setup however is without multipath and because we create a generic dracut initrd which pulls in all installed modules we actively need to omit the module by default. This is the reason why self.root_filesystem_is_multipath is set to False. From your review you want to set it to True which confuses me ?

I just thought about having something like:

```python
self.root_filesystem_is_multipath = True
if self.xml_state.build_type.get_type() is 'oem':
    self.oemconfig = self.xml_state.get_build_type_oemconfig_section()
    if not self.oemconfig or not self.oemconfig.get_oem_multipath_scan():
        self.root_filesystem_is_multipath = False
```

This way we would have multipath untouched (which is the preferred situation IMHO) for vmx images and disable it for oem images that have no multipath configuration. I just suggested so because from my perspective not omitting any module should be the default and just actively omit modules based on user config for concrete cases. Anyway I agree the code above this is so confusing code :P and I did not explain it correctly.

davidcassany:
@schaefi ups we overlapped. Sure doing a review

PR #660, opened 2018-03-12
Add profile dependencies

### LINKED ISSUE #658: Profiles should be able to get grouped
## Problem description
When building an image description, we can potentially have multiple profiles that together make up the actual expected build target. For example, you may have profiles for "VMware", "x11_common" and "gnome". The expected build target would be "VMware-gnome" which can then combine all the targets required.

## Expected behaviour
There are multiple ideas on how to accomplish this. One idea would be to add a special "profiles group" tag in parallel to the "profiles" one that defines groups:

```
  <profilegroups>
    <profilegroup name="VMware-gnome" includes="VMware,x11_common,gnome">
  </profilegroups>
```
Another idea (probably better IMHO) would be to just allow profiles to include other profiles automatically:

```
  <profiles>
    <profile name="VMware" />
    <profile name="x11_common" />
    <profile name="gnome" />
    <profile name="VMware-gnome" includes="VMware,x11_common,gnome" />
  </profiles>
```

That's probably more consistent and easier to integrate in OBS.

@adrianschroeter @Vogtinator both can provide more ideas if needed.

+ KIWI version: 
+ Operating system: 
+ OBS version:

davidcassany:
With this commit the profile can include other profiles as a dependency.
This way it is possible build an image defined by multiple profiles by just
calling a single composed profile from the command line.

Fixes #658

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -889,6 +889,25 @@ div {
         ## packages.
         element profile {
             k.profile.attlist,
+            k.requires*
+        }
+}
+
+#==========================================
+# common element <requires>
+#
+div {
+    k.requires.profile.attribute = 
+        ## The profile name required as part of the current profile
+        ## definition.
+        attribute profile {text}
+    k.requires.attlist =
+        k.requires.profile.attribute
+    k.requires =
+        ## Requires is used to set profiles dependencies, with it a profile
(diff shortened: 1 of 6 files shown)


schaefi:
yes full agree the above should be written as

```xml
<profile name="groupped">
    <requires profile="profile1"/>
    <requires profile="profile2"/>
</profile>
```

PR #662, opened 2018-03-13
Add disk_start_sector type attribute

### LINKED ISSUE #575: Support for custom disk-start-sector
Copied from openSUSE/kiwi#662

>Hello,
>
>There is kiwi command line command --disk-start-sector which is supposed to be used for switching first partition start sector value between old legacy 63-style and new default 2048.
>
>However, it would be extremely useful to be able to specify this parameter as a part of XML kiwi receipt. Use case is the following. Modern non-x86 systems (i.e. aarch64 ones) uses custom vendor-defined layouts to store firmware at microSD cards, the same medium where OS is located. For instance, RockChip RK3328 requires different parts of the boot firmware to be stored at 64, 16384 and 24576 sectors from the media begin. This obviously overlaps with standard 2048 rule. If we were able to specify that first partition had to start at, say, 32768 sector in this particular case, then It would simplify JeOS building at OBS.

davidcassany:
This commit adds the disk start sector attribute to configure the
first disk sector for the first partition of the disk.

Fixes #575

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -111,7 +111,6 @@ class DiskBuilder(object):
             xml_state, root_dir
         )
         self.custom_args = custom_args
-        self.start_sector = xml_state.get_disk_start_sector()
 
         self.signing_keys = None
         if custom_args and 'signing_keys' in custom_args:
@@ -221,7 +220,8 @@ class DiskBuilder(object):
         loop_provider.create()
 
         self.disk = Disk(
-            self.firmware.get_partition_table_type(), loop_provider
+            self.firmware.get_partition_table_type(), loop_provider,
+            self.xml_state.get_disk_start_sector()
         )
 
         # create the bootloader instance
@@ -571,36 +571,31 @@ class DiskBuilder(object):
         if self.firmware.legacy_bios_mode():
             log.info('--> creating EFI CSM(legacy bios) partition')
             self.disk.create_efi_csm_partition(
-                self.firmware.get_legacy_bios_partition_size(),
-                self.start_sector
+                self.firmware.get_legacy_bios_partition_size()
             )
 
         if self.firmware.efi_mode():
             log.info('--> creating EFI partition')
             self.disk.create_efi_partition(
-                self.firmware.get_efi_partition_size(),
-                self.start_sector
+                self.firmware.get_efi_partition_size()
             )
 
         if self.firmware.ofw_mode():
             log.info('--> creating PReP partition')
             self.disk.create_prep_partition(
-                self.firmware.get_prep_partition_size(),
-                self.start_sector
+                self.firmware.get_prep_partition_size()
             )
 
         if self.disk_setup.need_boot_partition():
             log.info('--> creating boot partition')
             self.disk.create_boot_partition(
-                self.disk_setup.boot_partition_size(),
-                self.start_sector
+                self.disk_setup.boot_partition_size()
             )
 
         if self.spare_part_mbsize:
             log.info('--> creating spare partition')
             self.disk.create_spare_partition(
-                self.spare_part_mbsize,
-                self.start_sector
+                self.spare_part_mbsize
             )
 
         if self.root_filesystem_is_overlay:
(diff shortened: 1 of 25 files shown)


kiwi/defaults.py
```diff
@@ -59,7 +57,6 @@ def __init__(self):
         }
         self.profile_key_list = [
             'kiwi_align',
-            'kiwi_startsector',
```
schaefi: Very concerned about this change. The start sector must be in the profile so that the initrd code can access it for its repartition tasks. The method to_profile() in the Defaults space makes use of this values and makes sure they all land in the .profile file. Deleting it from this list has very intrusive consequences. I saw you were aware of this problem and addressed in the profile code but the reason why I did not put those into extra methods was that the values belongs together

I understand that an extra method for the start sector is nice but I would not go that route for start_sector, alignment and sectorsize. They all belong together and should not be separated.

Please keep the list intact, it will also make the amount of changes smaller

kiwi/defaults.py
```diff
@@ -658,6 +655,17 @@ def get_default_boot_timeout_seconds(self):
...
+    @classmethod
+    def get_default_disk_start_sector(self):
+        """
+        Implements the default initial disk sector for the first disk
+        partition.
+
+        :return: sector
+        :rtype: int
+        """
+        return 2048
```
schaefi: Should be:

```python
return self.defaults['kiwi_startsector']
```

kiwi/partitioner/dasd.py
```diff
@@ -42,26 +42,31 @@ def post_init(self):
...
-            if mbsize == 'all_free':
-                partition.write('n\np\n\n\nw\nq\n')
-            else:
-                partition.write('n\np\n\n+%dM\nw\nq\n' % mbsize)
+            partition.write(
+                'n\np\n{0}\n{1}\nw\nq\n'.format(
+                    '' if not start_sector else start_sector,
+                    '' if mbsize == 'all_free' else '+{0}M'.format(mbsize)
+                )
+            )
```
schaefi: A DASD device does not allow a custom start sector. s390 will not use a dasd with any other than the default start sectors. If you haven't tested s390 booting from dasd with custom start sectors any different than what fdasd does and it worked, please revert that change and leave s390 untouched for the moment

kiwi/partitioner/gpt.py
```diff
@@ -44,25 +44,29 @@ def post_init(self):
...
         :param list flags: additional flags
+        :param int start_sector: first sector in case it is the first partition
         """
         self.partition_id += 1
         if mbsize == 'all_free':
             partition_end = '0'
         else:
             partition_end = '+' + format(mbsize) + 'M'
+        if self.partition_id > 1 or not start_sector:
+            start_sector = 0
```
schaefi: Could you add a comment:

```
# A start  sector value of 0 specifies the default value defined in sgdisk
```

kiwi/partitioner/msdos.py
```diff
@@ -47,31 +47,32 @@ def post_init(self):
...
         :param string name: partition name
         :param int mbsize: partition size
         :param string type_name: partition type
         :param list flags: additional flags
+        :param int start_sector: first sector in case it is the first partition
         """
         self.partition_id += 1
         fdisk_input = NamedTemporaryFile()
+        if self.partition_id > 1:
+            start_sector = None
```
schaefi: Could you add a comment

```
# Undefined start sector value skips this for fdisk and use its default value
```

kiwi/defaults.py
davidcassany: I understand you point that they all belong together these values. I'll work to re-add the kiwi_startsector into the list, but then I guess I will need to modify somehow the `Profile.add` method and/or keep the `Profile._type_to_profile()` as is in this PR, otherwise the profile file includes always `kiwi_startsector='2048'` as a fixed value even the disk_start_sector was set to something different. In fact I included this change after testing and noticing the profile file was not including the correct value. I'll try to address it and see if I can find a convincing solution.

kiwi/defaults.py
davidcassany: Well, it should self initate first, which is a bit odd, but it perfectly works, in fact, this was my first attempt.

kiwi/partitioner/dasd.py
davidcassany: Ok, yes I haven't tested DASD, just verified that the fdasd call was the same without any custom start sector. You are right, better leave it untouched.

kiwi/builder/disk.py
```diff
@@ -570,31 +571,36 @@ def _build_and_map_disk_partitions(self):               # noqa: C901
         if self.firmware.legacy_bios_mode():
             log.info('--> creating EFI CSM(legacy bios) partition')
             self.disk.create_efi_csm_partition(
-                self.firmware.get_legacy_bios_partition_size()
+                self.firmware.get_legacy_bios_partition_size(),
+                self.start_sector
```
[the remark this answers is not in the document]
davidcassany: Agree, you have a point, better to implement that logic of start_sector usage inside the Disk. I will address it.

davidcassany:
@schaefi, I think now is all addressed and way better

schaefi:
yes this looks really good now :+1:

schaefi:
just one thing, we might want to have a log warning that start sector is ignored on dasd, iirc it's silently ignored right now. Just a minor thing

PR #663, opened 2018-03-14
Make build type mandatory in XMLState

davidcassany:
Make build type mandatory in XMLState

### DIFF
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -1760,6 +1760,9 @@ class XMLState(object):
         # build type is first type section in XML sequence
         if image_type_sections:
             return image_type_sections[0]
+        raise KiwiTypeNotFound(
+            'No build type defined. At least one type section is mandatory'
+        )
 
     def _profiled(self, xml_abstract):
         """
--- a/kiwi/config/strip.xml
+++ b/kiwi/config/strip.xml
@@ -7,6 +7,7 @@
         <specification>default items to keep or delete from the kiwi initrd</specification>
     </description>
     <preferences>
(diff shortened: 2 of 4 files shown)


kiwi/xml_state.py
```diff
@@ -1760,6 +1760,9 @@ def _build_type_section(self, build_type=None):
         # build type is first type section in XML sequence
         if image_type_sections:
             return image_type_sections[0]
+        raise KiwiTypeNotFound(
+            'No build type defined. At least one type section is mandatory'
+        )
```
schaefi: Wouldn't it be better to handle that at the earliest entry point which is the schema:

```
element preferences {
    k.type+ &
```

schaefi: The error message you provide is nicer than what the schema validation would tell (assuming no jing around) but other than that we might exit as early as possible

davidcassany: >```
>element preferences {
>    k.type+ &
>```

This is not enough, it could happen that all types are within a profile and one selects a profile that does not include any type. Then type is empty, see the `test/data/example_no_default_type.xml` I used for testing, yes this is a quite extreme case, but this is what @gmoro reported in IRC.

I am pretty sure it can be achieved by a schematron rule, but I am afraid of having some complex schematron rule that might be wrong in some insane profiles combination. Also I am not a big fan of including very specific schematron rules into the schema as they are hard to read. Anyway, probably we can try it and it turns to be plain and simple.

Any thought?

schaefi: ah ok, yes fair point. Ok so let's go with your original implementation

PR #664, opened 2018-03-14
kiwi_inode_size value is no longer hardcoded in more than one place

davidcassany:
Just nit-picking, if found it while coding #662 and preferred to fix it in a separate PR.


schaefi:
yes that makes a lot sense :+1:

PR #665, opened 2018-03-15
Iso tool api

schaefi:
Refactor Iso class api

In kiwi we have an abstraction layer for calling the iso filesystem creation tools. However this layer assumes we are option compatible with all tools we use. This is not correct if we consider support for
xorriso. Thus a refactoring is required first before we can think about xorriso

This patch is two fold

1. Cleanup Iso class to only provide generic helper methods around the creation of an iso
2. Provide an own namespace for the cdrtools (mkisofs/genisoimage)
3. Update the filesystem/isofs code to make use of the changed api

It touches quite some code but it should change nothing in the end. I have not tested intensively but will do the next days. So this is open for review to get your thoughts and ideas

Thanks

### DIFF
--- a/kiwi/filesystem/isofs.py
+++ b/kiwi/filesystem/isofs.py
@@ -18,7 +18,7 @@
 # project
 from kiwi.filesystem.base import FileSystemBase
 from kiwi.command import Command
-from kiwi.iso import Iso
+from kiwi.iso_tools.iso import Iso
 
 
 class FileSystemIsoFs(FileSystemBase):
--- a/kiwi/filesystem/isofs.py
+++ b/kiwi/filesystem/isofs.py
@@ -17,7 +17,6 @@
 #
 # project
 from kiwi.filesystem.base import FileSystemBase
-from kiwi.command import Command
 from kiwi.iso_tools.iso import Iso
 from kiwi.iso_tools.cdrtools import IsoToolsCdrTools
 
@@ -38,7 +37,6 @@ class FileSystemIsoFs(FileSystemBase):
         :param string exclude: unused
         """
         iso_tool = IsoToolsCdrTools(self.root_dir)
-        iso_tool_name = iso_tool.get_tool_name()
 
         iso = Iso(self.root_dir)
         iso.setup_isolinux_boot_path()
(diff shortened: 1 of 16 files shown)


kiwi/filesystem/isofs.py
```diff
@@ -36,28 +37,39 @@ def create_on_file(self, filename, label=None, exclude=None):
...
-        iso.add_efi_loader_parameters()
+        iso_tool.add_efi_loader_parameters()
+
         Command.run(
             [
-                Iso.get_iso_creation_tool()
-            ] + iso.get_iso_creation_parameters() + [
+                iso_tool_name
+            ] + iso_tool.get_iso_creation_parameters() + [
                 '-o', filename, self.root_dir
```
schaefi: This option handling needs to be addressed as we can't assume this to be correct for any tool

davidcassany: I guess this comment is in line to my comment below, or you meant something different?

kiwi/filesystem/isofs.py
```diff
@@ -36,28 +36,26 @@ def create_on_file(self, filename, label=None, exclude=None):
...
-            [
-                Iso.get_iso_creation_tool(),
-                '-hide', iso.header_end_name,
-                '-hide-joliet', iso.header_end_name
-            ] + iso.get_iso_creation_parameters() + [
-                '-o', filename, self.root_dir
-            ]
+
+        iso_tool.create_iso(
+            filename, hidden_files=[iso.header_end_name]
```
davidcassany: :+1: this looks great

kiwi/iso_tools/iso.py
```diff
@@ -445,7 +330,7 @@ def isols(self, isofile):
             'listing_type', ['name', 'filetype', 'start']
         )
         listing = Command.run(
-            [self.get_isoinfo_tool(), '-R', '-l', '-i', isofile]
+            [self._get_isoinfo_tool(), '-R', '-l', '-i', isofile]
```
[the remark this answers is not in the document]
schaefi: yes you are right xorriso can do what isoinfo does but it provides a different format. For the moment I have left out this part because I just need one implementation which can list contents of an iso. But you are right this should also be moved to be tool specific so that in the end we can move completely. I'll take a look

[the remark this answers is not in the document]
davidcassany: What you think of having the isols method as a classmethod of the an isotool class? If we assume that genisoimage or mkisofs are required for that we could have the isoToolsBase class calling the cdrtools class method or something similar.

schaefi:
@davidcassany ok I think I catched up with the review issues you mentioned. Thanks much for the review. How about it now ?

schaefi:
ok I'll run one final test build and then merge this one, so that the real xorriso work can start :)

Issue #666, opened 2018-03-16
invalid shebang for `kiwi/xml_parse.py` (also, says py2 compatible when isn't!)


Conan-Kudo:
## Problem description

In [the package review for kiwi for Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=1483339), the following issue came up from @decathorpe:

>The file "kiwi/xml_parse.py" has "#!/usr/bin/env python" shebang, which is wrong and must be replaced by "#!/usr/bin/python3" (sic!) and the file marked as executable, or the shebang should be removed entirely (for both the python2 and python3 version). That file isn't even compatible with python2, so ... I guess that should be fixed upstream.

## OS and Software information
+ KIWI version: 9.13.7
+ Operating system: Fedora 27

davidcassany:
About the shebang presence I am wondering why is it offensive, IMHO it doesn't hurt and I'd prefer to leave auto-generated code untouched. GenerateDS creates a code that can be run and tested as a script and even we do not need that feature I don't really see why we should change that. If the auto-generated code includes

```python
if __name__ == '__main__':                                                      
    main()
```
I think is correct to have a shebang.

In the worst case we could always remove the first line of that file in the Makefile, but it looks like an ugly hack to me.

Anyway, @Conan-Kudo thanks much for reporting us back this feedback here and your efforts to get kiwi into Fedora.

decathorpe:
The file having a shebang isn't the issue per se. However, having a shebang implies that the file can be executed as a script - which it can't, because it doesn't even have the executable bit set. rpmlint complains about this discrepancy.

Either way, this shebang has to be patched (and the file set as executable) - or removed - for fedora to comply with the Packaging Guidelines for python, which ban the use of `#!/usr/bin/env python` and unversioned `#!/usr/bin/python`. To quote the guidelines [0]:

> (...) `/usr/bin/python` (as well as `/usr/bin/env python` and similar) MUST NOT be used in shebang lines (...)

I don't care whether the fix occurs upstream or in a downstream patch.

Side note: I might be wrong about the python2 incompatibility.

[0]: http://fedoraproject.org/wiki/Packaging:Python#Multiple_Python_Runtimes

davidcassany:
@decathorpe, thanks much for your feedback. I'll think about it there must be a solution.

Vogtinator:
Note that we have the same policy in openSUSE - so this apparently got ignored during the submission process.

It still shows:

```
[...]

PR #670, opened 2018-03-19
Remove unneeded shebangs from dracut modules

### LINKED ISSUE #668: dracut module shell script files for kiwi-lib have shebangs when they shouldn't
## Problem description
In [the package review for kiwi for Fedora](https://bugzilla.redhat.com/show_bug.cgi?id=1483339), the following issue came up from @decathorpe:

> 8) dracut modules (in "/usr/lib/drac7t/modules.d/99kiwi-lib/*" have bash shebangs, but aren't marked as executable. Is that correct?

These files are purely functions and should not have shebangs, since there's nothing useful to execute.

+ KIWI version: 9.13.7
+ Operating system: Fedora 27

davidcassany:
This PR removes the shebang from some bash files that contain only functions, thus they are not executable (they are only sourced).

Fixes #668


davidcassany:
@schaefi, what you think about this change? I am not very happy about having to force the shell type in the shellcheck call, but after thinking about it I thought that probably this is the best solution. Another option would be to make all these files executable but I think it would just be worse.

schaefi:
I think it does not hurt but I guess apart from the adaption to the code and the shellcheck call there will be also rpmlint complaining about the missing shebang. Have you tried building an rpm package and check the rpmlint output ?

Apart from all that I wonder about the motivation for the change ? Did the shebangs cause any trouble elsewhere ?

Of course deleting code is always nice, thus if it does not cause us trouble I'm fine with the change :)

davidcassany:
> I think it does not hurt but I guess apart from the adaption to the code and the shellcheck call there will be also rpmlint complaining about the missing shebang. Have you tried building an rpm package and check the rpmlint output ?

This is a good point, I'll check. I assumed it would not complain since the whole idea behind is to minimize rpmlint complaints.

> Apart from that I wonder about the motivation for the change ? Did the shebangs cause any trouble elsewhere ?

Apparently in Fedora tool chain this causes rpmlint issues. If a shebang is there in a file it expects the executable attribute. See #666 comments (also the link with the package review is nice).

schaefi:
ah ok, thanks for the update. So if the suse package build is good to go I'm fine

davidcassany:
All right, it has no impact on our packages. I saw that had a warning on `kiwi/config/functions.sh` for the same reason, so I applied the same criteria and now we got rid of a couple of warnings. Also removed the rpmlint filter to ignore non-executable-script warnings in dracut modules.

schaefi:
great :+1:

PR #672, opened 2018-03-21
Remove xml_parse.py shebang in spec

### LINKED ISSUE #666: invalid shebang for `kiwi/xml_parse.py` (also, says py2 compatible when isn't!)

davidcassany:
It removes the shebang from the autogenerated file and also removes
the exception in rpmlint to ignore non-executable-script warning.

Fixes #666

### DIFF
--- a/package/python-kiwi-rpmlintrc
+++ b/package/python-kiwi-rpmlintrc
@@ -9,9 +9,6 @@ addFilter("no-manual-page-for-binary kiwi-ng")
 addFilter("suse-filelist-empty .*")
 addFilter("explicit-lib-dependency .*")
 
-# don't blame on auto generated code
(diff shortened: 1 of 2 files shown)


schaefi:
:+1:  :)

PR #673, opened 2018-03-22
Switch to xorriso

### LINKED ISSUE #635: Move from genisoimage to xorriso
## Problem description

`genisoimage` (from `cdrkit`) is old and unmaintained. These days, the modern replacement available is from [`xorriso`](https://www.gnu.org/software/xorriso/), which looks like it supports all the features that exist in the RH/SUSE variant of `genisoimage`.

Please switch from `genisoimage` to `xorriso`.

+ KIWI version: 9.13.0
+ Operating system: Fedora 27, openSUSE Tumbleweed (20180218)
+ OBS version: N/A

schaefi:
* Added IsoToolsXorrIso class
  As an alternative to mkisofs/genisoimage there is now the
  IsoToolsXorrIso class which implements the IsoTools interface
  by using xorriso. This Fixes #635

* Allow to configure iso tool category
  Switch to xorriso by default but allow to setup cdrtools
  in the runtime configuration file

### DIFF
--- a/kiwi/builder/install.py
+++ b/kiwi/builder/install.py
@@ -225,11 +225,12 @@ class InstallImageBuilder(object):
         )
         iso_header_offset = iso_image.create_on_file(self.isoname)
 
-        # make it hybrid
-        Iso.create_hybrid(
-            iso_header_offset, self.mbrid, self.isoname,
-            self.firmware.efi_mode()
-        )
+        # make it hybrid if not already done by iso tool
+        if iso_header_offset:
+            Iso.create_hybrid(
+                iso_header_offset, self.mbrid, self.isoname,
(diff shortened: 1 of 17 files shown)


schaefi:
@davidcassany would be great if you can do a test live iso and oem install iso build and see if it works for you too. xorriso is really nice and I got great help from Thomas Schmitt who is the main developer

schaefi:
As you can see I also set the default to be xorriso which if we agree to should also result in a follow up change of the spec file to require xorriso and finally get rid of any cdrtools stuff

schaefi:
Thanks for testing, I'll merge the first round. A cleanup pull request will follow

PR #675, opened 2018-03-22
Update spec file to require xorriso

schaefi:
Hope it can be done that way for all distros

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -133,11 +133,7 @@ Requires:       gdisk
 %endif
 Requires:       dosfstools
 Requires:       e2fsprogs
-%if 0%{?suse_version} >= 1500
-Requires:       mkisofs
-%else
-Requires:       genisoimage
-%endif
+Requires:       xorriso
 Requires:       grub2
 Requires:       kiwi-man-pages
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@schaefi It is the `xorriso` package in Fedora, EPEL (for CentOS/RHEL), Mageia, and Debian/Ubutnu.

schaefi:
thank you guys for checking it :+1:

DimStar77:
Xorriso is actuall very expensive, as xorriso requires /usr/bin/wish, which is provided by tk, and that in turn requires tcl.

(I had to unblock tk/tcl for the live images for now)

DimStar77:
And fedora counts as error free reference since when?

Conan-Kudo:
@DimStar77 So either delete it, or subpackage it out.

DimStar77:
No kidding... What would I do without such an advise. I would be lost. Thanks!!! 
(And yes, splitting is the best course here)

Conan-Kudo:
Sorry, I'm just incredibly frustrated, annoyed, and stressed because of https://github.com/openSUSE/open-build-service/issues/4761 completely screwing up my week.

PR #680, opened 2018-03-28
Delete clean dependencies

### LINKED ISSUE #625: Allow deletion of packages which aren't needed in the root fs anymore
Currently, it's not possible to remove grub/plymouth/dracut/genisoimage/etc. packages as they are used by kiwi later on.

It should be possible to delete those (and also its dependencies) by generating the needed files before processing the `delete` sections.

davidcassany:
This PR adds the delete_clean_dependencies type of packages that lists packages to delete including its dependencies.

Fixes #625

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -2522,7 +2522,8 @@ div {
         ## packages are only installed if this build type is requested.
         attribute type {
             "bootstrap" | "delete" | "docker" | "image" |
-            "iso" | "oem" | "pxe" | "vmx" | "oci"
+            "iso" | "oem" | "pxe" | "vmx" | "oci" |
+            "delete_clean_dependencies"
         }
     k.packages.profiles.attribute = k.profiles.attribute
     k.packages.patternType.attribute =
--- a/kiwi/schema/kiwi.rng
+++ b/kiwi/schema/kiwi.rng
@@ -3967,6 +3967,7 @@ packages are only installed if this build type is requested.</a:documentation>
           <value>pxe</value>
           <value>vmx</value>
           <value>oci</value>
+          <value>delete_clean_dependencies</value>
         </choice>
       </attribute>
     </define>
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -287,6 +287,31 @@ class SystemPrepare(object):
                     '%s: %s' % (type(e).__name__, format(e))
                 )
 
+    def delete_packages_and_dependencies(self):
+        """
+        Delete packages marked for deletion in the XML description
+        """
+        to_become_deleted_packages = \
+            self.xml_state.get_to_become_deleted_packages(force=False)
+        if to_become_deleted_packages:
+            log.info('Deleting packages and its dependencies')
+            try:
+                # PackageManager instace without repositories is created
+                # here because packages removal does not require zypper
+                # to have repositories included and it allows to delete
+                # packages even after buildtime repositories are cleaned
+                package_manager = self.xml_state.get_package_manager()
+                self.delete_packages(
+                    PackageManager(
+                        Repository(self.root_bind, package_manager),
+                        package_manager
+                    ), to_become_deleted_packages
+                )
+            except Exception as e:
+                raise KiwiInstallPhaseFailed(
+                    '%s: %s' % (type(e).__name__, format(e))
+                )
+
     def install_packages(self, manager, packages):
         """
         Install one or more packages using the package manager inside
--- a/kiwi/tasks/system_build.py
+++ b/kiwi/tasks/system_build.py
@@ -173,12 +173,6 @@ class SystemBuildTask(CliTask):
         self.runtime_checker.check_repositories_configured()
         self.runtime_checker.check_image_include_repos_publicly_resolvable()
 
-        package_requests = False
-        if self.command_args['--add-package']:
-            package_requests = True
-        if self.command_args['--delete-package']:
-            package_requests = True
-
         log.info('Preparing new root system')
         system = SystemPrepare(
             self.xml_state,
@@ -193,15 +187,14 @@ class SystemBuildTask(CliTask):
         system.install_system(
             manager
         )
-        if package_requests:
-            if self.command_args['--add-package']:
-                system.install_packages(
-                    manager, self.command_args['--add-package']
-                )
-            if self.command_args['--delete-package']:
-                system.delete_packages(
-                    manager, self.command_args['--delete-package']
-                )
+        if self.command_args['--add-package']:
+            system.install_packages(
+                manager, self.command_args['--add-package']
+            )
+        if self.command_args['--delete-package']:
+            system.delete_packages(
+                manager, self.command_args['--delete-package']
+            )
 
         profile = Profile(self.xml_state)
 
--- a/kiwi/tasks/system_build.py
+++ b/kiwi/tasks/system_build.py
@@ -225,6 +225,7 @@ class SystemBuildTask(CliTask):
         # setup permanent image repositories after cleanup
         setup.import_repositories_marked_as_imageinclude()
         setup.call_config_script()
+        system.delete_packages_and_dependencies()
 
         # make sure system instance is cleaned up now
         del system
--- a/kiwi/tasks/system_prepare.py
+++ b/kiwi/tasks/system_prepare.py
@@ -160,12 +160,6 @@ class SystemPrepareTask(CliTask):
         self.runtime_checker.check_repositories_configured()
         self.runtime_checker.check_image_include_repos_publicly_resolvable()
 
-        package_requests = False
-        if self.command_args['--add-package']:
-            package_requests = True
-        if self.command_args['--delete-package']:
-            package_requests = True
-
         log.info('Preparing system')
         system = SystemPrepare(
             self.xml_state,
@@ -180,15 +174,15 @@ class SystemPrepareTask(CliTask):
         system.install_system(
             manager
         )
-        if package_requests:
-            if self.command_args['--add-package']:
-                system.install_packages(
-                    manager, self.command_args['--add-package']
-                )
-            if self.command_args['--delete-package']:
-                system.delete_packages(
-                    manager, self.command_args['--delete-package']
-                )
+
+        if self.command_args['--add-package']:
+            system.install_packages(
+                manager, self.command_args['--add-package']
+            )
+        if self.command_args['--delete-package']:
+            system.delete_packages(
+                manager, self.command_args['--delete-package']
+            )
 
         profile = Profile(self.xml_state)
 
--- a/kiwi/tasks/system_prepare.py
+++ b/kiwi/tasks/system_prepare.py
@@ -214,6 +214,7 @@ class SystemPrepareTask(CliTask):
         # setup permanent image repositories after cleanup
         setup.import_repositories_marked_as_imageinclude()
         setup.call_config_script()
+        system.delete_packages_and_dependencies()
 
         # make sure system instance is cleaned up now
         del system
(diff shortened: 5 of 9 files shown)


davidcassany:
Currently only zypper package manager handles the deletion of packages in two different behaviors, (i) using the package manager to delete and clean respecting the package dependencies, or (ii) deleting only listed items even if it breaks dependencies.

Since the API is already in place, this enhancement for YUM, DNF and APT can be done in an independent pull request.

Conan-Kudo:
@davidcassany This feels like a really bad idea.

davidcassany:
@Conan-Kudo, not sure what are you referring to: the new delete_clean_dependencies feature, the implementation of this PR, or the idea of doing two different PRs. Could you give me a hint regarding that?

Conan-Kudo:
After talking to @Vogtinator on IRC about this, I understand what this is now.

Okay, first, this approach is nonsense, because it's adding an extra layer of complexity for no reason.

And the idea that we're going preserve the current behavior of `delete` attribute is insane. If you're going to do anything, I'd suggest dropping the old behavior and making `delete` use the high level package manager to dep resolved removals.

There's no decent reason why we should be doing `rpm --nodeps -e`, ever. This is not a good behavior to preserve at all.

davidcassany:
@Conan-Kudo I understand your concerns and I agree that `rpm --nodeps -e` should be considered a bad practice, something to avoid. However it has been in use in the past and it is also being used in the custom initrds that kiwi provides in order to make them lighter. I don't think can consider removing it without creating undesired backward compatibility issues.

Conan-Kudo:
@davidcassany But it creates issues unto itself doing it that way, because then the in-image package manager is not likely to function properly because of inconsistent or broken states in rpmdb.

schaefi:
After reading through all this I think you are all right, we are starting to create a weird concept to keep the old behavior of the packages type="delete" section compatible.

It probably was a mistake from my side from the beginning on to let this deletion be a forced deletion outside of the package manager control. I think it would be good if we change it. So here is my suggestion how to do it:

1. Make sure we agree on #681 and merge it
2. Come up with a pull request which makes packages type="delete" to use the no-forced variant of the deletion methods by default
3. Come up with a pull request which allows a packages type="force_delete"
4. Come up with a pull request for the kiwi-descriptions repo changing all boot image descriptions packages type="delete" section to use packages type="force_delete"

Now the behavior of the packages type="delete" has changed; what side effects could be expected:

* People still not on dracut using their own custom initrd description could be in trouble because the build could fail with the package manager not deleting the requested package because of a dependency problem. For boot (initrd) descriptions this will most likely be the case because you have to "break" the system to get it that small, at least that applies to suse. Solution would be to tell them to change the only line from "delete" to "force_delete" and suggest moving to dracut. I would be ok and willing to take the blame

* People using delete and expecting that only this package gets deleted could be astonished that more than that could be deleted. With the expectation that the package manager does it right this should lead to a cleaner system than before. If this is unexpected we could still suggest the force_delete way and maybe question the result of the package manager which is outside our responsibility. I would also be ok with that

Overall we gain a cleaner concept which I think is worth it

Feedback on this is very much appreciated

Vogtinator:
I would suggest to make `type="delete"` invalid - there won't be any cases which just keep working.

Either you end up with a totally broken image or the image build fails.

Maybe rename `type="delete"` to `type="force_delete"` and add a new `type="cleanup"`?

schaefi:
that would be misleading. It's still a deletion of packages driven by the package manager not a cleanup

Vogtinator:
That's bikeshedding - the name can be discussed after the general approach.

davidcassany:
> Let's abandon the idea of delete_clean_dependencies for the moment and discuss an alternative. I'm sorry David if parts of your effort here based on my own suggestion lead to work with no purpose

Sure, no problem, I believe the implementation won't be lost, as it is not complex and will be reusable if we go for the `force_delete` and `delete` approach.

> People still not on dracut using their own custom initrd description could be in trouble because the build could fail with the package manager not deleting the requested package because of a dependency problem. For boot (initrd) descriptions this will most likely be the case because you have to "break" the system to get it that small, at least that applies to suse. Solution would be to tell them to change the only line from "delete" to "force_delete" and suggest moving to dracut. I would be ok and willing to take the blame

I would be fine with it also. Let's hope it won't affect a lot of people. We could also discuss if an XSLT conversion from `delete` to `force_delete` is worth, it could help to avoid breaks, but probably it is better to make the change obvious so the user is aware of the change, I don't have a strong opinion regarding that.

> People using delete and expecting that only this package gets deleted could be astonished that more than that could be deleted. With the expectation that the package manager does it right this should lead to a cleaner system than before. If this is unexpected we could still suggest the force_delete way and maybe question the result of the package manager which is outside our responsibility. I would also be ok with that.

Probably this can be _solved_ with some kind of generic warning message that exposes the danger of using `delete` option, as it could clean packages needed by the build process (this is actually the case of the ubuntu image from kiwi-descriptions).

> Maybe rename type="delete" to type="force_delete" and add a new type="cleanup"?

I think this could be a reasonable option, but we need to find a good name, it is hard to beat `delete` and `force_delete` couple. Probably something like `force_delete` and `autoremove` or `remove`? This option has the benefit that backward compatibility issues could be nicely included in a XSLT conversion. As there would not a behavior change, just an attribute rename.

Despite the potential issues I am in favor of using `delete_force` and `delete`, or any other pair that we consider clear enough and lets us drop `delete` in order to facilitate compatibility issues.

schaefi:
Well after a few discussions with you I must say it's not easy to talk to you and stay focused on the goal. You brought up the name change from "delete" to "cleanup" which is misleading because we are talking about deletion of packages and not about cleanup of packages. You responded with that this is bikeshedding which I translate as "wasting time on trivial details". So why did you bring it up at all ?

You also said "I would suggest to make type="delete" invalid - there won't be any cases which just keep working." But this is wrong. A high level package manager based dep resolved removal will of course work, is the change we are after, and is also what you requested with your "zypper remove -u" call from the issue all this references to. The name "delete" is a very good match to my eyes

I'm sorry you are confusing me more and more

We won't be changing the packages type="delete" name. I'm open for changing how packages gets deleted and an optional force of that deletion process as described above and imho also suggested by Conan-Kudo.

So please let us focus on this

Thanks

schaefi:
>Probably this can be solved with some kind of generic warning message that exposes the danger of 
>using delete option, as it could clean packages needed by the build process

right, that could happen :( A warning message would be good. Really helpful would be an information what will be really deleted which will be part of the log as I think the package manager reports that information. So a warning message could point users to the log as one source for a first review ?

> (this is actually the case of the ubuntu image from kiwi-descriptions).

oh ok, so this needs to be fixed, which package triggers that ?

davidcassany:
> oh ok, so this needs to be fixed, which package triggers that ?

I am not sure which is the root cause of it, but using autoremove causes APT to delete `mkinitrd` and other tools so the image misses some stuff to finalize. I did not have time to perform further tests last Thursday and to dig into the dependencies tree. I was about to test if including some specific tools in packages sections would solve it. I don't know how it happened but apparently there are some important system tools that are considered to be orphan in the APT data base.

Vogtinator:
> A high level package manager based dep resolved removal will of course work, is the change we are after, and is also what you requested with your "zypper remove -u" call from the issue all this references to. The name "delete" is a very good match to my eyes

Yes, it will definitely "work" in the way that it does not fail.
However, many images intentionally break dependencies, which would in some cases make the image completely unusuable. That's a semantic change which needs to be communicated by explicitly breaking it and leaving the choice to the description's editor.

Conan-Kudo:
APT's concept of `autoremove` is... odd compared Yum/DNF and Zypper. The APT autoremove is global, and will deinstall _everything_, that doesn't have explicit `Depends`. Things that are using boolean `Depends` aren't properly considered, and can be orphaned if something else potentially satisfies it, which is why APT's autoremove is not default.

If it's marked as user-installed (which can be done simply by doing `apt-get install <foo>`, then it will be considered user-installed and not eligible for autoremove).

Conan-Kudo:
@davidcassany @schaefi This PR can be closed, as it's obsoleted by the merging of #681.

davidcassany:
@Conan-Kudo we could consider closing this PR since the `delete_clean_dependencies` won't be applied, at least as it is in this PR. However note that #681 does not change the behavior of anything and doesn't make this PR obsolete. The code in #681 just improves feature parity with zypper and the other supported package managers, but has no impact on how packages are deleted as currently only force delete is called in prepare step.

I think we should leave this opened for discussion and definitely decide on if `type="delete"` and `type="force_delete"` make sense even when it could cause some compatibility issues.

The @schaefi plan sounds quite good to me:

> 1. Make sure we agree on #681 and merge it
> 2. Come up with a pull request which makes packages type="delete" to use the no-forced variant of the deletion methods by default
> 3. Come up with a pull request which allows a packages type="force_delete"
> 4. Come up with a pull request for the kiwi-descriptions repo changing all boot image descriptions packages type="delete" section to use packages type="force_delete"

Just thought about another detail to consider for this change, I don't think it is an issue, but something to consider. Currently the deletion of `type="delete"` is forced and applied before running config.sh, IMHO, if the deletion is not forced, it makes more sense to apply it after running config.sh (this way we can install tools needed in config.sh and then remove them in a clean way).

davidcassany:
@Vogtinator, hi is the approach suggested by @schaefi acceptable for you?

> 1. Make sure we agree on #681 and merge it
> 2. Come up with a pull request which makes packages type="delete" to use the no-forced variant of the deletion methods by default
> 3. Come up with a pull request which allows a packages type="force_delete"
> 4. Come up with a pull request for the kiwi-descriptions repo changing all boot image descriptions packages type="delete" section to use packages type="force_delete"

I guess this is the case, but we have no confirmation so far. I'd like to move on with this PR.

Vogtinator:
No, I'm still very much in favor of either keeping the old behaviour of `type="delete"` and adding a new type or removing `type="delete"` altogether.

The approach suggested by @schaefi silently breaks backwards compat.

davidcassany:
@schaefi I have been thinking about this PR and what you think if we add just a new `type` as @Vogtinator suggests. We could have a new `uninstall` type, IMHO this would be a reasonable match to the behavior we expect, being `delete` a hard removal and `uninstall` a removal according to the package manager criteria.

Another approach could be also switch to `remove` and `force_remove`, but I have slightly preference for uninstall.

schaefi:
> We could have a new uninstall type, IMHO this would be a reasonable match to the behavior
> we expect, being delete a hard removal and uninstall a removal according to the package
> manager criteria.

It's a good idea, I would be ok with it.

Along with the change I suggest to add an example in doc/source/building/working_with_images which explains in a straight forward way what happens if you specify
```<packages type="delete|uninstall">```and also include what happens with ```<packages type="image-type-name">``` meaning they get only installed according to the selected build type. I think then all is covered up in a nice way

schaefi:
@Vogtinator if we do it that way you need to adapt your image description to use type="uninstall" which I guess is ok for you ?

schaefi:
Also I think we will allow to have both sections active ? Meaning a user could have both: packages in a section with type="delete" and packages in a section with type="uninstall". In this case I would say uninstall is first, followed by delete. Thoughts ?

Vogtinator:
> @Vogtinator if we do it that way you need to adapt your image description to use type="uninstall" which I guess is ok for you ?

Yes, that's exactly what I had hoped: No change unless explicitly requested in the .kiwi file.

schaefi:
> I vote to run also delete after the config.sh or just do not allow both at the same time.

I agree let's do it that way. With the ability to have type="delete" and also type="uninstall" as part of the XML description. There is imho no need for users to perform this action as part of the config.sh script. Thus I'm ok to run any package deletion/uninstall sequence after config.sh

davidcassany:
Closing as it is superseded by #724

PR #681, opened 2018-03-29
Enable no forced deletion for dnf, yum and apt package managers

davidcassany:
This commit makes apt, yum and dnf support equivalent to zypper in terms of
being capable to delete packages using the package manger tools
(implies dependencies deletion is handled by the package manager)
or deleting explicitly only listed packages using packager tools
even if this implies breaking dependencies.

### DIFF
--- a/kiwi/package_manager/yum.py
+++ b/kiwi/package_manager/yum.py
@@ -173,9 +173,6 @@ class PackageManagerYum(PackageManagerBase):
         """
         Process package delete requests (chroot)
 
-        Note: yum erase does not delete unneeded dependencies it only
-        makes sure no dependencies are broken.
-
         :param bool force: force deletion: true|false
         """
         delete_items = []
@@ -205,7 +202,7 @@ class PackageManagerYum(PackageManagerBase):
                 [
                     'chroot', self.root_dir, self._get_yum_binary_name()
                 ] + chroot_yum_args + self.custom_args + [
-                    'erase'
+                    'autoremove'
                 ] + delete_items,
                 self.command_env
             )
--- a/kiwi/package_manager/yum.py
+++ b/kiwi/package_manager/yum.py
@@ -173,6 +173,9 @@ class PackageManagerYum(PackageManagerBase):
         """
         Process package delete requests (chroot)
 
+        Note: yum erase does not delete unneeded dependencies it only
+        makes sure no dependencies are broken.
+
         :param bool force: force deletion: true|false
         """
         delete_items = []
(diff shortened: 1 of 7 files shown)


kiwi/package_manager/yum.py
```diff
@@ -187,14 +190,25 @@ def process_delete_requests(self, force=False):
...
+                ] + delete_options + delete_items,
+                self.command_env
+            )
+        else:
+            chroot_yum_args = self.root_bind.move_to_root(self.yum_args)
+            return Command.call(
+                [
+                    'chroot', self.root_dir, self._get_yum_binary_name()
+                ] + chroot_yum_args + self.custom_args + [
+                    'erase'
```
Conan-Kudo: You can use `yum autoremove` to do the same thing that `dnf remove` does by default (remove and remove unneeded deps). `dnf autoremove` functions in the same way.

Cf. https://www.mankier.com/8/yum-deprecated#autoremove
Cf. https://www.mankier.com/8/dnf#Commands-Auto_Remove_Command

kiwi/package_manager/yum.py
```diff
@@ -173,6 +173,9 @@ def process_delete_requests(self, force=False):
         """
         Process package delete requests (chroot)
 
+        Note: yum erase does not delete unneeded dependencies it only
+        makes sure no dependencies are broken.
```
Conan-Kudo: This can actually be done with `yum autoremove`

Cf. https://www.mankier.com/8/yum-deprecated#autoremove

schaefi: The version of yum I have here does not provide an autoremove command, I only see erase which was the reason I used this command in the initial implementation. I we can asure yum (like on rhel) has it I think it would be good to move to autoremove. If only yum-deprecated provides the command I think we should stick to the current implementation

davidcassany: @schaefi, I also used erase because of this, I do not see autoremove in `yum --help` in the version I have and I see the same in Centos 7.0, however I have just tested the image we have in kiwi-descriptions using the `yum autoremove` and it worked fine as @Conan-Kudo described. I guess there is lack of documentation in `--help` option. I haven't checked man pages as it uses the `excludedocs` flag.

I am about to commit a change to make use of `autoremove`.

schaefi: ah ok, if a build works I think we are good to go

Conan-Kudo: @davidcassany It's in RHEL 7, it was introduced in https://github.com/rpm-software-management/yum/commit/4ad8960e12a19cbd2f6e3bb039c7b65015f59e19

Prior to this, `yum --setopt=clean_requirements_on_remove=True remove` also worked.

schaefi: ok, thanks, so if no further objections I would merge it

Conan-Kudo: Go for it!

PR #684, opened 2018-04-04
Update etc/default/grub setup

schaefi:
kiwi writes optional grub boot parameters to the GRUB_CMDLINE_LINUX
variable in default/grub. This information is then picked up by
grub2-mkconfig and written to grub.cfg However there is also another
variable named GRUB_CMDLINE_LINUX_DEFAULT which according to the
documentation should be used preferably. While it does not seem to
matter for grub it matters for yast. Thus this patch changes the
configuration variable and fixes bsc#1084117


schaefi:
Also see

* https://bugzilla.suse.com/show_bug.cgi?id=1084117

PR #686, opened 2018-04-05
Fixed detection of disk node in live iso images

schaefi:
If the live iso is booted as disk the initrd code needs to find the correct disk node pointing to the iso image. This was formerly done by checking if the populated disk devices contains an iso header with an application id. The information was obtained using the isoinfo tool. isoinfo is a tool provided by the obsolete and xorriso replaced cdrtools kit. In addition the lookup was unsafe because any iso with an application id would have been valid.

Thus this patch changes the detection mechanism to use the volume id as it is used in the root assignment on the cmdline. The volume id is populated as device label for the assigned block device and can therefore be used as a unique id. The volume id itself is a configuration option in the image
XML description. If not set the default is 'CDROM'. The information can be obtained via blkid and therefore also eliminates the isoinfo requirement

### DIFF
--- a/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
@@ -2,13 +2,16 @@
 type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
 
 function lookupIsoDiskDevice {
+    local root=$1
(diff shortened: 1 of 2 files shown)


schaefi:
great thanks

lnussel:
can we get a package submission to all code streams please? This is needed for the Leap live images AFAICT.

Issue #689, opened 2018-04-06
suseStripKernel is not being obeyed for non iso


cybernard:
## Problem description
I was clearly not clear enough the last question.
I don't want to alter the initrd fs.  I did a lsinitrd and nothing in there I want to change.

sudo kiwi-ng  --logfile="kiwi-ova.log" --profile=iso system build --description /kiwi/kiwi-descriptions/suse/x86_64/suse-tumbleweed-firewall     --target-dir /tmp/myimag2

Boot the ISO into your favorite VMWare,virtual box or etc
find /lib/modules/\<version\>\/kernel --iname "*.ko"|wc -l
result 360 modules  YAY

sudo kiwi-ng  --logfile="kiwi-ova.log" --profile=ovf system build --description /kiwi/kiwi-descriptions/suse/x86_64/suse-tumbleweed-firewall     --target-dir /tmp/myimag2

Deploy to my ESXi server and boot:
find /lib/modules/\<version\>\/kernel --iname "*.ko"|wc -l
result 4002 modules  NO

## Expected behaviour

find /lib/modules/\<version\>\/kernel --iname "*.ko"|wc -l
result 360 modules 

## Steps to reproduce the behaviour
1. Go into config.sh 
2. add suseStripKernel
3. edit config.xml
4. add \<drivers\>  (list of drivers)  \</drivers\>
(list of drivers) contains a carefully select group of 360 (or so) modules.

## OS and Software information
+ KIWI version: 9.14.1
+ Operating system: opensuse tumbleweed
+ OBS version:  not using it right now

davidcassany:
See my comments in https://github.com/SUSE/kiwi/issues/687#issuecomment-379244349, I believe this is a bug, but the intended behavior is the other way round, `suseStripKernel` should be also skipped in this iso context.

With the current dracut system I believe the documentation regarding this functions is confusing. Also, probably a review of these functions to state which of them make sense or are useful beyond the custom boot image scenario would be a good exercise to better state what is actually supported and expected of them.

cybernard:
Why should my image be forced to have 3000+ modules I will **NEVER** load.
Even the kernel developers saw the usefulness in this option as "make localmodconfig" when building the kernel from source is **officialy** supported.

davidcassany:
Sorry if my comment seamed harsh I never meant to say you are forced to include all the modules into the image.

[...]

PR #690, opened 2018-04-06
Fix default initrd_system values

davidcassany:
This commit fixes the default initrd_system value for some image
types. Since this value is included in profile and potentially
taken into account for some of the config script functions, it is
important to have consistent values even when the image type
has no initrd choice or it doesn't have initrd at all.

Related to #689


schaefi:
yes makes sense :+1:

PR #694, opened 2018-04-10
Support lookup for fstab.append on fstab creation

schaefi:
At the time kiwi creates the fstab with all required fields
to boot the system it now also looks for an optional fstab.append
file and appends its contents to the fstab file. This allows
to setup custom fstab entries for filesystem mounts which are
established outside of the kiwi image building process by
e.g a service at first boot


schaefi:
This references the caasp image which needs an extra data partition and our suggested solution on the image description level in: https://github.com/SUSE/kiwi-descriptions/pull/21

PR #704, opened 2018-04-16
Cleanup api doc strings

### LINKED ISSUE #697: api doc consistency cleanup [toplevel]
## Problem description

Cleanup doc strings on the root py files of the project

schaefi:
This cleans up and fixes the api documentation for the
public interface of the projects toplevel source code
files and Fixes #697


schaefi:
@davidcassany: first round of api cleanup. If ok by style I suggest to use the same for the other cleanup tasks. I have done a local rebuild of the docs to see the result

@tomschr: Thanks for your feedback. I have started the doc string cleanup with reference to your feedback mails. If you find something which you think needs to be done better/different don't hesitate to send feedback

Thanks

kiwi/command_process.py
```diff
@@ -33,10 +33,13 @@
...
 class CommandProcess(object):
     """
-    Implements processing of non blocking Command calls
+    **Implements processing of non blocking Command calls**
 
     Provides methods to iterate over non blocking instances of
     the Command class with and without progress information
+
+    :param subprocess command: instance of subprocess
+    :param string log_topic: topic string for logging
```
tomschr: Not sure if you want to have parameters at the class docstring. It's probably a matter of taste if you add them on the constructor or on the class. The only issue is that it should be consistent. :wink:

kiwi/command_process.py
```diff
@@ -145,7 +151,9 @@ def __del__(self):
 
 class CommandIterator(six.Iterator):
     """
-    Implements Iterator for Instances of Command
+    **Implements Iterator for Instances of Command**
```
tomschr: Implements **an** Iterator?

kiwi/firmware.py
```diff
@@ -28,22 +28,13 @@
...
-
-    * :attr:`arch`
-        machine architecture
-
-    * :attr:`zipl_target_type`
-        XML configured zipl target type
-
-    * :attr:`firmware`
-        XML configured firmware name
+    * :param object xml_state: instance of XMLState
```
tomschr: Hmn, I think, this is a contradiction. Shouldn't it be something like this? Also added `:class:` directive:

```
   :param XMLState xml_state: instance of :class:`XMLState`
```

kiwi/firmware.py
```diff
@@ -160,7 +164,8 @@ def get_legacy_bios_partition_size(self):
         Size of legacy bios_grub partition if legacy BIOS mode is
         required. Returns 0 if no such partition is needed
 
-        :return: mbsize
+        :return: mbsize value
```
tomschr: "... as integer"?

kiwi/firmware.py
```diff
@@ -173,7 +178,8 @@ def get_efi_partition_size(self):
         Size of EFI partition.
         Returns 0 if no such partition is needed
 
-        :return: mbsize
+        :return: mbsize value
```
tomschr: Same here

kiwi/firmware.py
```diff
@@ -189,7 +195,8 @@ def get_prep_partition_size(self):
         Size of Prep partition if OFW mode is requested.
         Returns 0 if no such partition is needed
 
-        :return: mbsize
+        :return: mbsize value
```
tomschr: Same here

kiwi/help.py
```diff
@@ -36,6 +36,8 @@ def show(self, command=None):
         All kiwi commands store their manual page in the section '8'
         of the man system. The calling process is replaced by the
         man process
+
+        :param string command: man page name
```
tomschr: string -> str (in Python, the string datatype is `str`)

kiwi/kiwi.py
```diff
@@ -29,7 +29,17 @@ def extras(help_, version, options, doc):
     """
     Overwritten method from docopt
 
-    We want to show our own usage for -h|--help
+    Shows our own usage message for -h|--help
+
+    :param bool help_: indicate to show help
+    :param string version: version string
```
tomschr: string -> str

kiwi/kiwi.py
```diff
@@ -29,7 +29,17 @@ def extras(help_, version, options, doc):
     """
     Overwritten method from docopt
 
-    We want to show our own usage for -h|--help
+    Shows our own usage message for -h|--help
+
+    :param bool help_: indicate to show help
```
tomschr: Why is there a "_" after `help`? Looks a bit strange to me. :wink:

kiwi/kiwi.py
```diff
@@ -29,7 +29,17 @@ def extras(help_, version, options, doc):
...
+    :param string version: version string
+    :param list options:
+        list of option tuples
+
+        .. code:: python
+
+            [option(name='name', value='value')]
+    :param string doc: docopt doc string
```
tomschr: string -> str

kiwi/kiwi.py
```diff
@@ -85,6 +95,8 @@ def usage(command_usage):
        short form by default, long form with -h | --help
 
     3. the global options
+
+    :param string command_usage: usage data
```
tomschr: string -> str

kiwi/logger.py
```diff
@@ -220,7 +234,9 @@ def filter(self, record):
 
 class Logger(logging.Logger):
     """
-    Extended logging facility based on python logging
+    **Extended logging facility based on python logging**
```
tomschr: python -> Python

kiwi/mount_manager.py
```diff
@@ -25,20 +25,15 @@
...
 
-    Attributes
-
-    * :attr:`device`
-        device node name
-
-    * :attr:`mountpoint`
-        mountpoint directory name
+    * :param string device: device node name
+    * :param string mountpoint: mountpoint directory name
```
tomschr: string -> str (for both)

kiwi/path.py
```diff
@@ -126,9 +133,13 @@ def which(
         :param string filename: file base name
```
tomschr: string -> str

kiwi/privileges.py
```diff
@@ -25,14 +25,16 @@
...
-    Implements check for root privileges
+    **Implements check for root privileges**
     """
     @classmethod
     def check_for_root_permissions(self):
         """
         Check if we are effectively root on the system. If not
         an exception is thrown
 
+        :return: True or raise an Exception
```
tomschr: Here it would make sense to use the `:raises:` exception:

```
  :raises: Exception
  :return: True or raise an exception
```

kiwi/runtime_checker.py
```diff
@@ -34,15 +34,14 @@
...
+    **Implements build consistency checks at runtime**
 
     The schema of an image description covers structure and syntax of
     the provided data. The RuntimeChecker provides methods to perform
     further semantic checks which allows to recognize potential build
     or boot problems early.
 
-    * :attr:`xml_state`
-        Instance of XMLState
+    * :param object xml_state: Instance of XMLState
```
tomschr: Same as before

kiwi/runtime_config.py
```diff
@@ -58,6 +58,8 @@ def get_obs_download_server_url(self):
         if no configuration exists the downloadserver from
         the Defaults class is returned
 
+        :return: URL type data
+
         :rtype: string
```
tomschr: string -> str

kiwi/xml_description.py
```diff
@@ -48,25 +48,17 @@
...
-        path to XML description file
-
-    * :attr:`derived_from`
-        path to base XML description file
-
-    * :attr:`xml_content`
-        XML description data as content string
+    :param string description: path to XML description file
+    :param string derived_from: path to base XML description file
+    :param string xml_content: XML description data as content string
```
tomschr: All of them: string -> str

davidcassany:
@schaefi, sure going to follow the style in this request.

kiwi/command_process.py
schaefi: yes it was a matter of taste and we did it in the class docstring all over the place and it is rendered fine in sphinx. Thus I kept that concept and think it's ok if reading the class intro also describes how to construct on object of that class. I think we don't have a consistency problem because we did that from the beginning on, so there should not be any __init__ constructor with a docstring

kiwi/firmware.py
schaefi: Hmm, I always used __:param object argument_name: instance of Class__  to say argument_name is a python object of the class Class

where is this a contradiction ? and if so there are many many cases where I used it oh oh

kiwi/firmware.py
schaefi: isn't that clear by the line below saying :rtype: int ?

kiwi/kiwi.py
schaefi: because the extras mehod was overwritten by kiwi from docopt (as I wrote in the docstring :))
the docopt original version of the extras methd declared the argument '_help' for whatever reason. yes it's strange but I wanted to keep the same definition

kiwi/kiwi.py
schaefi: Uhm yes ok but I used string everywhere in the code and I think it's not really misleading nor causing any trouble. So if this is not urgent let us keep it

kiwi/firmware.py
tomschr: Oh, right, sorry.

kiwi/kiwi.py
tomschr: Yes, I was splitting hair. :laughing: I agree, it's not something urgent or misleading.

kiwi/firmware.py
tomschr: > Hmm, I always used :param object argument_name: instance of Class to say argument_name is a python object of the class Class

According to the Sphinx documentation, the synopsis for a `:param:` directive is this:

```
:param [TYPE] VARNAME: DESCRIPTION
```

In some cases, you can omit the `type` and write this:

```
:param VARNAME: DESCRIPTION
:type VARNAME: Type1, Type2
```

This can be used if the argument can have different datatypes (for example, a `str` or `bytes`).

In the above example, you used `object` for the datatype. Although, technically it is correct, it is **very** broad. Everything is an object in Python. :wink: It doesn't give the reader additional information.  If a developer reads this information, he should know what type the function/method is expected and IMHO `object` is not correct.

In your case, it's not only one object, but a specific object: an instance of the class `XMLState`.

> where is this a contradiction ? and if so there are many many cases where I used it oh oh

Ok, maybe contradiction was the wrong word; let's agree on that it's ambiguous. :wink:

If it causes too much trouble on your side, lets take the current state for now.

schaefi:
@tomschr thanks for the feedback. I'm going to merge this even though it's not perfect. I guess docs can always improve :) Regarding the string vs str I think we could do that by one match_and_replace commit once David and myself have corrected the api documentation completely. If we do that now it could end up in a conflict because David and myself are working together on cleaning up the api docs

tomschr:
@schaefi Perfectly fine! :) As a doc writer, I'm always nit picking, so don't worry. :smile:

PR #715, opened 2018-04-20
Fixed creation of machine settings file

### LINKED ISSUE #710: VMWare fix breaks vmdk builds
## Problem description
The SUSE/fix_vmware_settings_for_nic_setup (merge request #696 ) patch breaks existing builds of virtual machines for vmdk format.

## Expected behaviour

The patch #696 should not break existing KIWI descriptions for vmdk disk format.

1. Use an existing kiwi description which worked with KIWI 9.14.2
2. Build it with a KIWI git master version after merging the pull request #696
3. Wait til KIWI stops with the error

+ KIWI version: Any git master later than git commit d5bca0399e8fbfc9c62249cae51128a966359041
+ Operating system: OBS for target OS: CentOS, Fedora, SLE, openSUSE
+ OBS version: Any, the error message is KIWI related, but tested with 2.9 and git master

To Crosscheck the Issue, I have build from git master of KIWI a version and reverse applied fix_vmware_settings_for_nic_setup (merge request #696 ). Then the current git master works again.

schaefi:
If no vmnic setup is present a request to iterator over
a NoneType object is attempted and failed. This Fixes #710

### DIFF
--- a/test/unit/xml_state_test.py
+++ b/test/unit/xml_state_test.py
@@ -199,6 +199,7 @@ class TestXMLState(object):
     def test_get_build_type_vmnic_entries(self):
         assert self.state.get_build_type_vmnic_entries()[0].get_interface() \
             == ''
+        assert self.boot_state.get_build_type_vmnic_entries() == []
 
     def test_get_build_type_vmdvd_section(self):
         assert self.state.get_build_type_vmdvd_section().get_id() == 0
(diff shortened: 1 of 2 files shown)


test/unit/xml_state_test.py
```diff
@@ -199,6 +199,7 @@ def test_get_build_type_vmdisk_section(self):
     def test_get_build_type_vmnic_entries(self):
         assert self.state.get_build_type_vmnic_entries()[0].get_interface() \
             == ''
+        assert self.boot_state.get_build_type_vmnic_entries() == []
```
davidcassany: Good, having 100% coverage is a very good measure but it certainly does not verify all the logic or execution *paths*. A while ago a looked if there was any tool to warn about not covered *elses* for unit tests, but did not find anything.

PR #716, opened 2018-04-20
Use travis stages feature for pypi deployment

schaefi:
Problem ist that the deploy section is called for every
item in the former matrix: setup. This means the pypi
upload was triggered twice for the same archive which
means one target always fails. In order to deploy only
once travis provides a stages feature which is used
in this commit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,31 @@
 language: python
+
 branches:
   except:
   - gh-pages
-matrix:
+
+stages:
+- test
+- deploy
+
+jobs:
   include:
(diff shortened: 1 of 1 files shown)


schaefi:
I have not tested this. Would be great if someone with experience for the travis staging feature could have a look. Thanks

schaefi:
ok I will merge this and test what happens on travis. This requires me to do a release tag, so just in case you wonder what's going on. I don't expect this to work in the first shot :)

davidcassany:
I'll cross my fingers ;)

PR #719, opened 2018-04-24
Add test for the utils class StringToSize

davidcassany:
This commit adds a couple of unit tests for the StringToSize class.

### DIFF
--- a/test/unit/utils_size_test.py
+++ b/test/unit/utils_size_test.py
@@ -1,0 +1,17 @@
+from .test_helper import raises
+
+from kiwi.utils.size import StringToSize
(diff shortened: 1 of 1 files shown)


davidcassany:
@schaefi, I need this in SLE12SP3 branch as while backporting the unpartitioned feature I noticed that there was no StringToSize class in SLE12, so I backported it also, but then realized there were no tests for it. Since it is a single simple method I guess the code was just covered by other tests making use of it.

In SLE12SP3 coverage was not complete after backporting StringToSize then I noticed there were no tests for it. So here they are.

schaefi:
yes you are right the result was part of other tests. Great that you have added it explicitly :+1:

PR #724, opened 2018-05-04
Add uninstall packages type

### LINKED ISSUE #625: Allow deletion of packages which aren't needed in the root fs anymore

davidcassany:
This PR adds a new `uninstall` type for packages. Packages listed
with this type will be removed by the package manager cleaning also any
unneeded dependency. The removal is executed after running `config.sh`.

Also in this commit `delete` type for packages is now executed after
`uninstall` packages, meaing it also happens after `config.sh`.

Fixes #625

### DIFF
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -274,26 +274,34 @@ class SystemPrepare(object):
                     'System archive installation failed: %s' % format(e)
                 )
 
-    def pinch_system(self):
+    def pinch_system(self, manager=None, force=False):
         """
-        Delete packages marked for deletion in the XML description. First
-        uninstalls packages marked with `type="uninstall"` if any, then
-        proceeds with packages marked with `type="delete"` if any.
+        Delete packages marked for deletion in the XML description. If force
+        param is set to False uninstalls packages marked with
+        `type="uninstall"` if any; if force is set to True deletes packages
+        marked with `type="delete"` if any.
 
-        :raises KiwiInstallPhaseFailed: if the deletion packages process fails
+        :param object manager: instance of :class:`PackageManager`
+        :param bool force: Forced deletion True|False
+
+        :raises KiwiPackagesDeletePhaseFailed: if the deletion packages
+            process fails
         """
         to_become_deleted_packages = \
-            self.xml_state.get_to_become_deleted_packages()
-        to_become_uninstalled_packages = \
-            self.xml_state.get_to_become_uninstalled_packages()
+            self.xml_state.get_to_become_deleted_packages(force)
         try:
-            if to_become_uninstalled_packages:
-                self.delete_packages_without_configured_repos(
-                    to_become_uninstalled_packages
-                )
             if to_become_deleted_packages:
-                self.delete_packages_without_configured_repos(
-                    to_become_deleted_packages, force=True
+                log.info('{0} system packages (chroot)'.format(
+                    'Force deleting' if force else 'Uninstalling')
+                )
+                if manager is None:
+                    package_manager = self.xml_state.get_package_manager()
+                    manager = PackageManager(
+                        Repository(self.root_bind, package_manager),
+                        package_manager
+                    )
+                self.delete_packages(
+                    manager, to_become_deleted_packages, force
                 )
         except Exception as e:
             raise KiwiPackagesDeletePhaseFailed(
@@ -365,27 +373,6 @@ class SystemPrepare(object):
                 raise KiwiSystemDeletePackagesFailed(
                     'Package deletion failed: %s' % format(e)
                 )
-
-    def delete_packages_without_configured_repos(self, packages, force=False):
-        """
-        This is a wrapper of delete_packages that does not require a package
-        manager instance. It performs exactly the same code using a dummy
-        package manager instance created on the fly.
-
-        Repositories configuration is unneeded for packages removal, thus
-        using a dummy package manager instance allows to run the method even
-        after the build time repositories have been cleaned.
-
-        :param list packages: package list
-        :param bool force: force deletion true|false
-        """
-        package_manager = self.xml_state.get_package_manager()
-        self.delete_packages(
-            PackageManager(
-                Repository(self.root_bind, package_manager),
-                package_manager
-            ), packages, force
-        )
 
     def update_system(self, manager):
         """
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -34,7 +34,8 @@ from kiwi.exceptions import (
     KiwiSystemUpdateFailed,
     KiwiSystemInstallPackagesFailed,
     KiwiSystemDeletePackagesFailed,
-    KiwiInstallPhaseFailed
+    KiwiInstallPhaseFailed,
+    KiwiPackagesDeletePhaseFailed
 )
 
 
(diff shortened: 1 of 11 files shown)


kiwi/system/prepare.py
```diff
@@ -355,6 +366,27 @@ def delete_packages(self, manager, packages, force=False):
                     'Package deletion failed: %s' % format(e)
                 )
 
+    def delete_packages_without_configured_repos(self, packages, force=False):
+        """
+        This is a wrapper of delete_packages that does not require a package
+        manager instance. It performs exactly the same code using a dummy
+        package manager instance created on the fly.
+
+        Repositories configuration is unneeded for packages removal, thus
```
schaefi: Don't understand this. The force deletion doesn't need a repo setup but the uninstall imho does. In the zypper case we call it with the '-u' flag which does a backward resolver task looking which packages are no longer needed when the package to become deleted gets removed. imho that requires access to the configured repo metadata. I can't tell how the other package manager does it but I would say a clean dep removal task needs the repo metadata

In the context where this method is called you also have access to an instance of PackageManager, I wonder why this method is needed. Unless I completely overlook something this feels weird

kiwi/system/prepare.py
```diff
@@ -273,25 +274,31 @@ def install_system(self, manager):
...
-            except Exception as e:
-                raise KiwiInstallPhaseFailed(
-                    '%s: %s' % (type(e).__name__, format(e))
+        to_become_uninstalled_packages = \
+            self.xml_state.get_to_become_uninstalled_packages()
+        try:
+            if to_become_uninstalled_packages:
+                self.delete_packages_without_configured_repos(
+                    to_become_uninstalled_packages
+                )
```
schaefi: I would have thought this would be:

```python
self.delete_packages(manager, to_become_uninstalled_packages, force=False)
```

kiwi/system/prepare.py
```diff
@@ -273,25 +274,31 @@ def install_system(self, manager):
...
+            if to_become_deleted_packages:
+                self.delete_packages_without_configured_repos(
+                    to_become_deleted_packages, force=True
```
schaefi: and here I would have thought this would be:

```python
self.delete_packages(manager, to_become_deleted_packages, force=True)
```

kiwi/system/prepare.py
davidcassany: Agree that feels weird, I did not find a better way to do it. I am totally opened to improve or modify it.

> In the zypper case we call it with the '-u' flag which does a backward resolver task looking which packages are no longer needed when the package to become deleted gets removed. imho that requires access to the configured repo metadata.

Well, I am not sure about that, why would it need it? I'd say rpmdb should be enough IMHO. In my tests it worked fine without any repo configuration.

Anyway, the motivation of it comes from https://github.com/SUSE/kiwi/blob/348c0193245fd637c3ab841ef1ec34bd6ae5b56c/kiwi/tasks/system_prepare.py#L218 note that the manager instance (with build time repositories configured) is deleted before running `config.sh`. Thus after `config.sh` we don't have anymore the package_manager instance we used for the install phase. We could probably recreate it (by reconfiguring buildtime repos again), I just haven't considered it since it also looked to me weird and not as safe as do not touch any repo configuration.

Thoughts?

kiwi/system/prepare.py
davidcassany: We could do it this way, but then we need a manager instance in `build` and `prepare` tasks. It would be also ok to handle the new manager instance in `prepare` and `build` tasks before running `pinch_system`. This way the system_prepare API would not include this weirdness.

kiwi/system/prepare.py
schaefi: > Well, I am not sure about that, why would it need it? I'd say rpmdb should be enough

you are right, looking closer to how zypper deletes packages this works for the rpmdb case. If we can guarantee that this applies to all other package managers too I'm ok. It's most likely that this is the case as they all maintain a database and the package manager sits on top.

kiwi/system/prepare.py
schaefi: The reason why this delete statement exists is to free the resources of package manager and repository instance bound to the manager instance such that the config.sh script code can not play tricks with it while we are building. So imho we have two options:

* move del manager after pinch_system, which would allow config.sh writers to influence the temp repo setup. Probably we don't want to open this gate

* update the pinch_system method to work if manager is not set

   ```
   def pinch_system(self, manager=None, force=False):
       if not manager:
            ...
   ```

Thoughts ?

kiwi/system/prepare.py
davidcassany: Well, the tests I did were all successful with dnf, yum and zypper. With APT the behavior was different, but as @Conan-Kudo noted [here](https://github.com/SUSE/kiwi/pull/680#issuecomment-378209931), this is a difference on the package manager design, APT can't constraint the removal of unneeded dependencies to a specific package, it performs a global clean up. At the time I tested #681 I did all the tests together with #680, where also no repositories were configured and I had no issues with any of the package managers (apart of the APT that behaves different).

schaefi: ok :+1:

kiwi/system/prepare.py
schaefi: and then just change the pinch_system call to

```
system.pinch_system(force=...)
```

and prevent the additional delete_packages_with... method

davidcassany: I do not consider the first option to be feasible, it opens a range of cases to consider, note that the `imageinclude` and `imageonly` repos are already in place for `config.sh` with the current setup, altering this could cause issues.

I am fine with altering pinch_system to work with an undefined manager. I have some doubts about `force` argument. I am fine with it, but then we would have two `pinch_system` calls together:

```python
# Remove type="uninstall" packages
system.pinch_system(manager=None, force=False)
# Remove type="delete" packages
system.pinch_system(manager=None, force=True)
```
I am fine with both options, (i) having a `pinch_system` method that does a single operation (with force argument) or (ii) having a `pinch_system` method that performs both (without force argument).

kiwi/tasks/system_build.py
```diff
@@ -223,15 +216,14 @@ def process(self):                                      # noqa: C901
...
-            manager=manager, force=True
-        )
         # make sure manager instance is cleaned up now
         del manager
 
         # setup permanent image repositories after cleanup
         setup.import_repositories_marked_as_imageinclude()
         setup.call_config_script()
+        system.pinch_system(manager=None, force=False)
+        system.pinch_system(manager=None, force=True)
```
schaefi: I think manager=None must not be set as this is the default value. A short comment would imho also make sense. Thus I think it could be simplified into

```python
# handle uninstall package requests, gracefully uninstall with dependency cleanup
system.pinch_system(force=False)

# handle delete package requests, force uninstall
system.pinch_system(force=True)
```

kiwi/tasks/system_prepare.py
```diff
@@ -210,16 +204,14 @@ def process(self):                                      # noqa: C901
...
-        )
-
         # make sure manager instance is cleaned up now
         del manager
 
         # setup permanent image repositories after cleanup
         setup.import_repositories_marked_as_imageinclude()
         setup.call_config_script()
+        system.pinch_system(manager=None, force=False)
+        system.pinch_system(manager=None, force=True)
```
schaefi: Same here, see above comment

davidcassany:
Right, makes sense, I will write the documentation in a follow up PR.

schaefi:
:+1:

schaefi:
probably unit test for pinch_system call failed...

davidcassany:
now `pip install --upgrade pip` fails in Travis... I hope it will be a transitory failure

PR #727, opened 2018-05-08
Update arm integration test

schaefi:
Existing panda build was outdated and non functional.
Move the test to a more popular target and write the
image description to use technology matching the
suse arm development effort. Target is now Rpi(64bit)


schaefi:
Currently this builds here:

* https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_arm

once merged I will add a _service file to use the git sources

build-tests/arm/test-image-rpi-oem/editbootinstall_rpi.sh
```diff
@@ -0,0 +1,45 @@
...
+echo "RPi EFI system, installing firmware on ESP"
+mkdir -p ./mnt-pi
+mount ${loopname}p1 ./mnt-pi
+( cd boot/vc; tar c . ) | ( cd ./mnt-pi/; tar x )
+umount ./mnt-pi
+rmdir ./mnt-pi
+
+#==========================================
+# Change partition label type to MBR
+#------------------------------------------
```
davidcassany: Probably we could use the `efiparttable="msdos"` to get MBR, but I guess the ESP partition type would still fail, according to the comment looks like the partition type needs to be 0x0C instead of 0x83.

schaefi: yeah that part I have taken from the arm team. I have an rpi here but not the 64bit variant. I plan to bye one and do further tests with the build. I have several boards on my desk but they are pretty outdated and I'd like our integration tests to build at least for some useful target. You can expect further changes to this test but as a start it's good to go

Thanks for the review

PR #731, opened 2018-05-15
Container compression configurable

### LINKED ISSUE #725: Don't compress docker tar balls anymore in bundler
## Problem description
Don't compress docker tar balls with xz anymore in bundler code.

The OBS is about to store them in a registry like storage, so our build script is uncompressing the
tar.xz (but keeps the tar ball). So this is just eating time and CPU cycles now, just skip xz compression
for files like

 my_container.x86_64-1.0.0-Build1.1.docker.tar

please

### LINKED ISSUE #730: KiwiCompressionFormatUnknown: could not detect compression format for /containers/SUSE:Templates:Images:SLE-12-SP3:Base/images/container:SLE-12-SP3-Container-Base.tar
## Problem description
OBS has now container templates (see SUSE:Templates:Images:SLE-12-SP3/SLE-12-Container-kiwi ) and one template is based on kiwi. Since yesterday, the kiwi container no longer builds:
```
[   94s] [ ERROR   ]: 16:34:37 | KiwiCompressionFormatUnknown: could not detect compression format for /containers/SUSE:Templates:Images:SLE-12-SP3:Base/images/container:SLE-12-SP3-Container-Base.tar
```

## Expected behaviour
Given that the default templates are linked on the front page, I would expect that theses templates always build without problems. So please test against theses templates before releasing a new kiwi version. My experience with kiwi/OBS so far is not very good. I look every ~2 month at building containers with kiwi and OBS and there is always some sort of problem (that get's fixed quickly, thanks for that! but it is annoying..)

schaefi:
__Make container compression a configuration option__
    
Change the ContainerBuilder class to evaluate on the configuration options to decide if the container archive should be compressed or not. By default the archive will be compressed, thus there is no change to the former behavior but can be setup in ~/.config/kiwi/config.yml as follows:
    
```yaml

container:
   - compress: none|xz
```

This Fixes #725

__Allow docker root import from uncompressed file__
    
Check the given file name for its compression format and only uncompress if a supported format could be detected. This Fixes #730

### DIFF
--- a/kiwi/archive/tar.py
+++ b/kiwi/archive/tar.py
@@ -70,6 +70,7 @@ class ArchiveTar(object):
                 '-c', '-f', self.filename
             ] + self._get_archive_items(source_dir, exclude)
         )
+        return self.filename
 
     def append_files(self, source_dir, files_to_append, options=None):
         """
@@ -88,6 +89,7 @@ class ArchiveTar(object):
             ] + options +
             self.xattrs_options + files_to_append
         )
+        return self.filename
 
     def create_xz_compressed(
         self, source_dir, exclude=None, options=None, xz_options=None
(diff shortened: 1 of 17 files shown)


schaefi:
@davidcassany there is a small refactoring included for Compress and ArchiveTar classes to support me better in the rest of the implementation. It's actually only providing return values from the archiving and compression methods

PR #733, opened 2018-05-15
Add correct GPL-3.0-or-later license

### LINKED ISSUE #732: Fix KIWI License
## Problem description

Fix KIWI license, currently it uses `GPL-3.0+` in spec, but I believe the appropriate value should be `GPL-3.0-or-later`. Also I believe KIWI should start using `%license` as stated in [openSUSE guidelines](https://en.opensuse.org/openSUSE:Specfile_guidelines#License_files)

## Expected behaviour

`License:`  field of KIWI spec is set to `GPL-3.0-or-later` and `%license` macro is being used to install the License file for Tumbleweed.

davidcassany:
Add the correct license reference in the spec License field

Fixes #732

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -43,7 +43,7 @@ Provides:       kiwi-schema = 6.6
 Release:        0
 Url:            https://github.com/SUSE/kiwi
 Summary:        KIWI - Appliance Builder Next Generation
-License:        GPL-3.0+
+License:        GPL-3.0-or-later
 %if %{_vendor} == "debbuild"
 # Needed to set Maintainer in output debs
 Packager:       Marcus Schaefer <ms@suse.de>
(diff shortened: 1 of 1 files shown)


davidcassany:
Well, one of the issues of not using the `%license` macro or installing manually the License in a doc folder (as we do now) is that when using the `excludedocs` flag, the license is also excluded and thus not properly installed. AFAIK this is one of the motivations of the `%license` macro. I believe this macro expands as something like:

```
mkdir -p %{_defaultlicensedir}/%{name}
cp <licensefile> %{_defaultlicensedir}/%{name}/<licensefile>
```

Where `%_defaultlicensedir` expands to `/usr/share/licenses` in opensuse. It does not work for us since %{name} expands to `kiwi-man-pages` and not `python-kiwi`. 

Probably we could define/use `%_defaultlicensedir` instead of `%_defaultdocdir`, but I saw I needed to makes several changes including the Makefile. So that I preferred to leave it untouched for now, I was afraid of introducing some unexpected side effect.

Regarding the License spec field value, according to the [openSUSE guide-lines](https://en.opensuse.org/openSUSE:Packaging_guidelines#Licensing) we should use the short form of [SPDX](https://spdx.org/licenses/), which is `GPL-3.0-or-later`.

schaefi:
> So that I preferred to leave it untouched for now, I was afraid of introducing some unexpected side effect.

right I also think it could strike back

PR #736, opened 2018-05-18
Symlink base container images if not compressed

### LINKED ISSUE #735: Paths in `derived_from` including colons break skopeo calls
## Problem description

Using uncompressed tarballs in `derived_from` in OBS fails. The failure I see in my env is caused by skopeo, that has a damn poor image reference strategy, the offensive call is something like:

```
skopeo copy docker-archive:/containers/devel:CaaSP:kubic-container/container/opensuse-tumbleweed-image.tar oci:/tmp/kiwi_layout_dir.e0nekdtv:base_layer
```

The problem is that the image path contains colons, and skopeo (like other tools in the containers universe) uses them as a field separator. Before we never faced this issue because KIWI was uncompressing the image to a temporary file (randomly auto-generated) and then using it to call skopeo, thus skopeo never saw a path including colons or other potentially offensive characters. 

IMHO, some kind of sanitation before calling skopeo makes sense. Probably we could hardlink the image to a `NamedTemporaryFile` when the source is uncompressed, this way inner skopeo calls would remain like they used to be.

Related to that I realized we have a bug, we wipe the original image if this is uncompressed, see https://github.com/SUSE/kiwi/blob/bb28768b39596e969a2b226145e33cebfb8160a0/kiwi/system/root_import/docker.py#L41 and https://github.com/SUSE/kiwi/blob/bb28768b39596e969a2b226145e33cebfb8160a0/kiwi/system/root_import/oci.py#L103-L104

With the above suggested solution it would also be solved.

+ KIWI version: 9.15.2
+ Operating system: 
+ OBS version:

davidcassany:
This commit creates a temporary symlink for base container images
when they are provided in an uncompressed tarball form. This ensures
the reference path is sane.

Fixes #735

### DIFF
--- a/kiwi/system/root_import/docker.py
+++ b/kiwi/system/root_import/docker.py
@@ -15,6 +15,9 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+from tempfile import NamedTemporaryFile
+import os
+
 # project
 from kiwi.system.root_import.oci import RootImportOCI
 from kiwi.logger import log
@@ -38,7 +41,9 @@ class RootImportDocker(RootImportOCI):
                 compressor.uncompress(True)
                 self.uncompressed_image = compressor.uncompressed_filename
             else:
-                self.uncompressed_image = self.image_file
+                with NamedTemporaryFile() as tempfile:
+                    self.uncompressed_image = tempfile.name
+                os.symlink(self.image_file, self.uncompressed_image)
             skopeo_uri = 'docker-archive:{0}'.format(self.uncompressed_image)
         else:
             log.warning('Bypassing base image URI to skopeo tool')
--- a/test/unit/system_root_import_docker_test.py
+++ b/test/unit/system_root_import_docker_test.py
@@ -42,6 +42,8 @@ class TestRootImportDocker(object):
             'oci:kiwi_layout_dir:base_layer'
         ])
 
+    @patch('kiwi.system.root_import.docker.NamedTemporaryFile')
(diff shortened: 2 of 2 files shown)


schaefi:
Oh man we do that to create another name for the same file because skopeo can't deal with it if there is a colon in it ?

I understand that the most simple solution to this problem is a workaround in kiwi but wouldn't it be much better to aim for:

* Fix skopeo
* Don't build base containers with names not suitable for skopeo

instead of messing up kiwi

...if we really have to go that route I at least want to see a comment in the code why we do that. Simple readers of the code without the issue background will just think we are brain dead

davidcassany:
>  wouldn't it be much better to aim for:
    * Fix skopeo
    * Don't build base containers with names not suitable for skopeo

I agree that fixing skopeo would be the nicest, but I am afraid this is a bit out of our hands, as it doesn't look like a simple fix... And about do not use paths which aren't suitable for skopeo sounds fair enough to me, just wondering if we could easily include some meaningful message.

Ok let's block it and see if we can come with a better solution.

schaefi:
ok that's fair enough, let's look if we can build skopeo friendly named base images. If that's for a good reason not possible I'm ok with the change here given we add a meaningful comment in the code.
I'll approve under this conditions.

Thanks much

Conan-Kudo:
Has anyone talked to skopeo upstream about fixing it? I don't see anything about it here: https://github.com/projectatomic/skopeo/issues

davidcassany:
See projectatomic/libpod#489. It looks like colons will remain as a reserved character for containers/image library. Lets leave it as is for now and assume it is just not supported.

PR #740, opened 2018-05-24
Do not delete uncompressed base docker images

### LINKED ISSUE #739: Original base image file is deleted if the original is uncompressed
## Problem description

I realized we have a bug, we wipe the original image if this is uncompressed, see https://github.com/SUSE/kiwi/blob/bb28768b39596e969a2b226145e33cebfb8160a0/kiwi/system/root_import/docker.py#L41 and https://github.com/SUSE/kiwi/blob/bb28768b39596e969a2b226145e33cebfb8160a0/kiwi/system/root_import/oci.py#L103-L104

## Expected behaviour

The original image remains untouched

Use an uncompressed docker image tarball in `drived_from` attribute

+ KIWI version: 9.15.2
+ Operating system: 
+ OBS version:

davidcassany:
Fixes #739

### DIFF
--- a/kiwi/system/root_import/oci.py
+++ b/kiwi/system/root_import/oci.py
@@ -100,5 +100,6 @@ class RootImportOCI(RootImportBase):
             Path.wipe(self.oci_layout_dir)
         if self.oci_unpack_dir:
             Path.wipe(self.oci_unpack_dir)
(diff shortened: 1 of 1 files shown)


schaefi:
Looks good :+1:

PR #742, opened 2018-05-28
Do not create static dev nodes in root init

schaefi:
For compatibility reasons kiwi created a set of static device
nodes when initializing a new image root system. With the
presence of devtmpfs this should no longer be needed. In addition
the static dev node setup now also causes problems on filesystems
like btrfs which was the reason to delete this code now.
This Fixes bsc#1087104

### DIFF
--- a/kiwi/system/root_init.py
+++ b/kiwi/system/root_init.py
@@ -15,7 +15,6 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
-import stat
 from pwd import getpwnam
 from tempfile import mkdtemp
 from shutil import rmtree
@@ -79,7 +78,6 @@ class RootInit(object):
         Path.create(self.root_dir)
         try:
             self._create_base_directories(root)
-            self._create_device_nodes(root)
             self._create_base_links(root)
             self._setup_config_templates(root)
             data = DataSync(root + '/', self.root_dir)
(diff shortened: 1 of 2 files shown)


schaefi:
References:

* https://bugzilla.suse.com/show_bug.cgi?id=1087104

schaefi:
@davidcassany from a container perspective, do you see a problem here ? I think (have not tested) that devtmpfs also populates a complete dev tree in a privileged container

Also we should be aware that with this change the build host system _has to_ provide devtmpfs and all nodes on that host for building images. e.g loop devices. I'd say this is true for all distros we support so far but the static node setup was a failsafe variant to make this sure which is now lost

However I don't see any other good solution which is not a hack to solve the btrfs problem and it's also nice to deal with less code in kiwi

Vogtinator:
Just deleting the nodes should be perfectly safe in all cases, but I'm also all for removing unnecessary functionality.

davidcassany:
@schaefi, from a containers perspective I believe this change is transparent, see https://github.com/SUSE/kiwi/blob/18191c8312966c23f63a2ff08d580cc2c6639ece/kiwi/container/oci.py#L150-L154 we are not including `/dev` in contaiers, in fact we include custom container stuff there https://github.com/SUSE/kiwi/blob/18191c8312966c23f63a2ff08d580cc2c6639ece/kiwi/container/setup/base.py#L158-L178

I think the change is just fine.

schaefi:
@davidcassany this is not what I meant. I mean if you build with kiwi in a container like e.g dice does it. The buildprocess will mount the devtmpfs filesystem to populate required device nodes temporarily into the image root environment.

If you are inside of a container does a devtmpfs mount contain a reasonable complete list of device nodes. I think it does but I'm not sure. As the container shares the kernel with the host I think this will be the same list as it is on the docker host, thus safe

schaefi:
The code you pointing to is when we build docker images. I'm looking at the host environment when we call kiwi to build a new root image. The same question applies to the buildservice worker environment. I guess it will also populate a complete enough dev tree but I don't know.

We will find out all this only by accident if for example no loop devices are present in devtmpfs this change here will break all builds

davidcassany:
@schaefi, I see, I overlooked the PR. I guess some tests are required for that, frankly speaking I am not sure how devices are handled within a privileged container. 

For OBS we should check builds in chroot env as in local builds.

Gonna run some tests.

davidcassany:
Agree, my test within a container worked fine without issues.

schaefi:
ok let's merge and keep fingers crossed :)

PR #743, opened 2018-05-28
Add service dependencies in generators

### LINKED ISSUE #741: CentOS ISO fails on Switch Root target
## Problem description
CentOS live ISO fails on initramfs Switch Root target while boot. There's generated sysroot.mount unit which should be called from somewhere but it doesn't. 
Run "kiwi-ng --color-output --type=iso system build --description ./ --target-dir /tmp/builds/"
Description was taken from https://github.com/SUSE/kiwi-descriptions/tree/master/centos/x86_64/centos-07.0-JeOS
Then feed ISO to Virtualbox
+ KIWI version: 9.15.2
+ Operating system: CentOS 7.5

davidcassany:
Correctly adding the initrd-root-fs.target service dependency to
auto generated sysroot.mount for kiwi-live and kiwi-overlay dracut
modules.

Fixes #741

### DIFF
--- a/dracut/modules.d/90kiwi-live/kiwi-generator.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-generator.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -x
 
 type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
 
@@ -36,9 +36,11 @@ ROOTFLAGS="$(getarg rootflags)"
     _dev=LiveOS_rootfs
 } > "$GENERATOR_DIR"/sysroot.mount
 
-mkdir -p "$GENERATOR_DIR"/initrd-root-fs.target.requires
-ln -s "$GENERATOR_DIR"/sysroot.mount \
-    "$GENERATOR_DIR"/initrd-root-fs.target.requires/sysroot.mount
+if [ ! -e "$GENERATOR_DIR/initrd-root-fs.target.requires/sysroot.mount" ]; then
+    mkdir -p "$GENERATOR_DIR"/initrd-root-fs.target.requires
+    ln -s "$GENERATOR_DIR"/sysroot.mount \
+        "$GENERATOR_DIR"/initrd-root-fs.target.requires/sysroot.mount
+fi
 
 mkdir -p "$GENERATOR_DIR/$_dev.device.d"
 {
--- a/dracut/modules.d/90kiwi-live/kiwi-generator.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-generator.sh
@@ -36,6 +36,10 @@ ROOTFLAGS="$(getarg rootflags)"
     _dev=LiveOS_rootfs
 } > "$GENERATOR_DIR"/sysroot.mount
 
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi, apparently generators are expected to create the dependencies symlinks according to the [generators docs](https://www.freedesktop.org/software/systemd/man/systemd.generator.html). Those dependency links are only created in `systemctl enabe/disable` procedure, which does not happen for generated services. Generators produce something like `system daemon-reload` which does not enable/disable any service.

schaefi:
Interesting. I wonder why it ever worked. Recently new enough versions of systemd seems to handle this correctly without us explicitly creating the link ??

dracut/modules.d/90kiwi-live/kiwi-generator.sh
```diff
@@ -36,6 +36,10 @@ ROOTFLAGS="$(getarg rootflags)"
     _dev=LiveOS_rootfs
 } > "$GENERATOR_DIR"/sysroot.mount
 
+mkdir -p "$GENERATOR_DIR"/initrd-root-fs.target.requires
+ln -s "$GENERATOR_DIR"/sysroot.mount \
+    "$GENERATOR_DIR"/initrd-root-fs.target.requires/sysroot.mount
```
schaefi: What if sysroot.mount already exists ?

davidcassany:
> Interesting. I wonder why it ever worked. Recently new enough versions of systemd seems to handle this correctly without us explicitly creating the link ??

Yes interesting, to sum up: any kernelcmdline that includes a `root=<something>` causes fstab generator to create a `sysroot.mount` unit and the related `initrd-root-fs.target.requires/sysroot.mount` link in recent systemd versions. So, the fstab-generator was creating for the dependency link. The problem in older systemd is that fstab-generator only creates a sysroot.mount unit and dependencies if `root=<absolutepath>` is used in the kernelcmdline, thus no one creates the  `initrd-root-fs.target.requires/sysroot.mount` link.

If you boot a live ISO with `rd.break=pre-mount` in opensuse you will see that there are two `sysroot.mount`, one in `/run/systemd/generator` created by fstab-generator and another one in `/run/systemd/generator.early`. The `generator.early` has higher priority and the one executed when `intird-root-fs.target` requires `sysroot.mount` to complete.

If you boot a centos ISO without the patch, you will see only a single `sysroot.mount` created by us, but no `initrd-root-fs.target` dependency (check `systemctl list-dependencies initrd-root-fs.target`). Thus no one requires `sysroot.mount` and the mount never happens. If you boot with `rd.break=pre-mount` and run `systemctl start sysroot.mount` then it continues the boot without issues.

You can check the systemd change in fstab-generator in this commit https://github.com/systemd/systemd/commit/093c2cfe3b, thanks to it the fstab-generator does not requires a `root=<absolutepath>`.

dracut/modules.d/90kiwi-live/kiwi-generator.sh
davidcassany: mmm it should not happen `fstab-generator` does not create it in `generator.early` as we do, but you are right, having a check would be safer. I'll include it

schaefi:
great summary all clear now :+1:

PR #744, opened 2018-05-28
Added support for config-cdroot archive

### LINKED ISSUE #737: Support files inclusion outside the compressed area in liveCDs
Comes from the `Kiwi-NG: cdroot` thread in the mailing list

## Problem description

Former KIWI versions used to support `config-cdroot.tgz` and `config-cdroot.sh` files in order to handle files that are not included into _squashfs_ image.

According to the users in mailing list, this is specially handy when some very big files (probably already compressed) need to be included in to LiveCD and the compression/decompression related steps are so demanding in terms of CPU cycles.

Being capable to include files into the ISO apart of the compressed FS was possible with former KIWI and is causing issues to users migrating to KIWI-NG.

## Expected behaviour

KIWI-NG can include files out of the squashfs image is ISO's. Ideally it should be possible to include files without having to use a compressed tarball (unlike in former KIWI with `config-cdroot.tgz`).

Kiwi-NG ignores `config-cdroot.tgz` file if present.

schaefi:
The image description now allows an optional file named:
config-cdroot.tar[.compression_postfix]. The file gets
unpacked as user data for live and install ISO images.
This allows users to add e.g license files or reference
documentation to the ISO image. This Fixes #737

### DIFF
--- a/doc/source/overview/workflow.rst
+++ b/doc/source/overview/workflow.rst
@@ -235,6 +235,15 @@ These are the optional components of an image description:
    after all the packages included in the :file:`config.xml` file
    have been installed. Any already present file is overwritten.
 
+#. CD root user data
+
+   For live ISO images and install ISO images an optional cdroot archive
+   is supported. This is a tar archive matching the name
+   :file:`config-cdroot.tar[.compression_postfix]`. If present it will
+   be unpacked as user data on the ISO image. This is mostly useful to
+   add e.g license files or user documentation on the CD/DVD which
+   can be read directly without booting from the media.
+
 #. Archives included in the :file:`config.xml` file.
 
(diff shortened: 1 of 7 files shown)


schaefi:
I did not strictly port the feature, meaning:

* as discussed we don't support the shell script extension config-cdroot.sh
* the name of the tar archive is config-cdroot.tar* Thus config-cdroot.tgz would not match. That's in line with all other tar file name matches in the next generation code base. Hope this is not a problem

schaefi:
Realized there is no documentation, I'll add that part

schaefi:
Thanks for the review. I'll merge it. Could you tell the user on the mailing list that 9.15.3 will have the support and probably point to the docs ? Thanks

PR #745, opened 2018-05-28
Fix setup of LOADER_TYPE in sysconfig/bootloader

schaefi:
LOADER_TYPE value for the grub2 bootloader depends on
the use of EFI. This Fixes bsc#1094883

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -187,8 +187,10 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         * FAILSAFE_APPEND
         """
         sysconfig_bootloader_entries = {
-            'LOADER_TYPE': self.boot_directory_name,
-            'LOADER_LOCATION': 'mbr'
+            'LOADER_TYPE':
+                'grub2-efi' if self.firmware.efi_mode() else 'grub2',
+            'LOADER_LOCATION':
+                'mbr'
         }
         if self.cmdline:
             sysconfig_bootloader_entries['DEFAULT_APPEND'] = '"{0}"'.format(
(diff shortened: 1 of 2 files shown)


kiwi/bootloader/config/grub2.py
```diff
@@ -187,8 +187,10 @@ def setup_sysconfig_bootloader(self):
         * FAILSAFE_APPEND
         """
         sysconfig_bootloader_entries = {
-            'LOADER_TYPE': self.boot_directory_name,
-            'LOADER_LOCATION': 'mbr'
+            'LOADER_TYPE':
+                'grub2-efi' if self.firmware.efi_mode() else 'grub2',
+            'LOADER_LOCATION':
+                'mbr'
```
Vogtinator: This looks wrong, on EFI there is no MBR.

I'm not sure where LOADER_LOCATION is used though, there doesn't seem to be sufficient documentation.

schaefi: I agree it's wrong for efi and it also could be wrong for bios if e.g installed into a partition. This value exists only because in the past perl-bootloader and iirc yast required a value. I would be more than happy to delete it entirely but wasn't sure if it cause any unwanted after effects. The reason why you see it in the diff is just because the indentation was changed by me. So it's present since a long time.

I would be ok deleting it and take the risk to see if something breaks. In this case we finally have a chance to know why it exists, add some comments and do the right thing. So far nobody really knows.

Thoughts ?

Vogtinator: Sounds like a good idea - but would fit better into a separate PR which is not part of an important last-minute bugfix. We shouldn't risk breaking that submission.

schaefi: agreed, I'll open an issue so we don't forget

PR #748, opened 2018-05-29
omit multipath module in live iso initrd

schaefi:
The multipath module creates device maps which puts the device
in a busy state and prevents the creation of a persistent write
partition. As multipath seems never useful for the root of a live
iso image we generally omit this module from being included


schaefi:
yes

ok so let's merge this. I'm sure multipath will follow us on other stages ;)

PR #749, opened 2018-05-30
Relax runtime check for ISO images using dmsquash

davidcassany:
This commit relaxes the dracut-kiwi-live package requirement if
dmsquash dracut module is selected in flags attribute.


schaefi:
Thanks, makes sense :+1:

PR #750, opened 2018-05-31
Updating shellcheck call from tox

davidcassany:
Recent shellcheck versions are more strict and complain about
backslashes `"\"` used in literals claiming it is preferred to use
double backslashes `"\\"`. As is just a styling advise and we use
backslashes in multiple commands (echo, sed, etc) I believe this
can be ignored.

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -213,9 +213,9 @@ function suseImportBuildKey {
         return
     fi
     if [ -d "/usr/lib/rpm/gnupg/keys" ];then
-        pushd "/usr/lib/rpm/gnupg/keys"
+        pushd "/usr/lib/rpm/gnupg/keys" || return
     else
-        pushd "${TDIR}"
+        pushd "${TDIR}" || return
         if [ -x "${dumpsigs}" ];then
             ${dumpsigs} /usr/lib/rpm/gnupg/suse-build-key.gpg
         fi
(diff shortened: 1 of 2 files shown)


davidcassany:
@scheafi, after updating to Leap 15 I realized the updated shellcheck is more strict that it used to be and I have a couple of new complaints:

* [SC1117](https://github.com/koalaman/shellcheck/wiki/SC1117) Complaints about backslashed in literals (as we do in sed calls or echo calls to sgdisk).
* [SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164) Complaints about using `pushd` and `popd` with checking a potetial failure. It demands for something like `pushd || exit`.

For the first I believe this should be ignored, I don't want to touch those literals, testing them all and risking that we miss any side effect.

For the second, I have doubts, because of that I added the WIP label. There are few (about 6 complaints), thus it would be simple to solve, but I wonder if this is really an enhancement, I find it quite redundant to add `pushd || exit 1`, it will eventually fail without the exit statement. Probably in some cases we could `pushd || return` or refactor some code using `pushd` return code to validate the directory existence and do not perform any previous sanity check. However I have the feeling we are adding harder to read code, for little or no gain at all. 

In short, I vote for ignoring [SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164) also, but it is true that it can be easily fixed. Also we could fix some and ignore the ones that are redundant by placing ignores within the code instead of ignoring them all. Any preference regarding that?

schaefi:
I agree on ignoring SC1117

Regarding SC2164 I can only find use of pushd/popd in kiwi/config/functions and I think it makes sense to check if that call was successful because if not all subsequent code will operate in the wrong directory. I vote for fixing those

schaefi:
:+1:

PR #757, opened 2018-06-04
Changed GUID format from signed to unsigned

schaefi:
With reference to Microsoft Guid constructors it seems unsigned
values are allowed which could exceed the value range of the
binary unpack used in kiwi. This Fixes bsc#1095267


schaefi:
* https://bugzilla.suse.com/show_bug.cgi?id=1095267

PR #765, opened 2018-06-06
Use the versioned Python interpreter path to run build helper scripts

Conan-Kudo:
The kiwi build process mostly does the right thing in terms of executing
with the correct Python interpreter throughout the build process, with
the exception of the Makefile not correctly locating the versioned Python
interpreter executable path, and the bash completion generator being executed
using `/usr/bin/python` without regard for what the target environment was.

This is a problem when a build environment complying with PEP 394 as it
stands today does not have Python 2 installed, such as when kiwi is being
built as a Python 3-only package in Fedora. Thus, the Makefile has been
adjusted to not only correctly locate the versioned Python binary, but
to also execute the completion_helper with the correct binary.

Additionally, a trivial change to the shebang to the completion_helper
was made to be consistent with other Python-based build helper scripts.

This is part of the overall effort to eliminate the dependencies on
Python 2 in Fedora.

Reference: https://fedoraproject.org/wiki/FinalizingFedoraSwitchtoPython3

* Use the versioned Python interpreter path to run build helper scripts


schaefi:
:+1:

PR #768, opened 2018-06-19
Prevent building custom efi image

schaefi:
If the distribution provides a prebuilt efi image kiwi
should use it instead of building its own image.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -417,11 +417,9 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             self._setup_EFI_path(lookup_path)
 
         if self.firmware.efi_mode() == 'efi':
-            log.info('--> Creating unsigned efi image')
-            self._create_efi_image(mbrid=mbrid, lookup_path=lookup_path)
+            self._setup_efi_image(mbrid=mbrid, lookup_path=lookup_path)
             self._copy_efi_modules_to_boot_directory(lookup_path)
         elif self.firmware.efi_mode() == 'uefi':
-            log.info('--> Setting up shim secure boot efi image')
             self._copy_efi_modules_to_boot_directory(lookup_path)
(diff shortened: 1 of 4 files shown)


kiwi/bootloader/config/grub2.py
```diff
@@ -610,6 +632,23 @@ def _create_efi_image(self, uuid=None, mbrid=None, lookup_path=None):
...
+            self._create_early_boot_script_for_uuid_search(
+                efi_boot_config, uuid
+            )
+        else:
+            self._create_early_boot_script_for_mbrid_search(
+                efi_boot_config, mbrid
+            )
+        with open(efi_boot_config, 'a') as config:
+            config.write(
+                'configfile ($root)/boot/{0}/grub.cfg{1}'.format(
```
agraf: If $prefix is set correctly, it should be enough to say "configfile grub.cfg". I can't shake off the feeling that calling "normal" is the better option though, but I'll have to double check with Raymund.

schaefi: ok, please clarify with Raymond and I'll give the shortest form a test :) Thanks

schaefi: just calling 'normal' worked for me, I'll update the pull request

agraf:
I had a quick chat with Raymund. The secure boot grub.cfg contains the following:

> set btrfs_relative_path="yes"
> search --fs-uuid --set=root 4f9ef382-6c1a-4d03-8aab-906e1dbbbc64
> set prefix=(${root})/boot/grub2
> source "${prefix}/grub.cfg"

He did say that "normal" should work just as well (and IMHO is cleaner). So let's leave it with normal.

Are you setting btrfs_relative_path?

schaefi:
> He did say that "normal" should work just as well (and IMHO is cleaner). So let's leave it with normal.

good

> Are you setting btrfs_relative_path?

yes

PR #769, opened 2018-06-29
avoid double quoting of disturl in label (...="'obs://...'").

adrianschroeter:
avoid always printed warning of missing disturl
make use of disturl variable

Fixes # .

*
*


Vogtinator:
Is `BUILD_DISTURL` guaranteed to not contain whitespace or any other unexpected sequences?

PR #773, opened 2018-07-04
Added system cleanup methods

schaefi:
Some files in the system gets created by services like
systemd or zypper and are meant to be created once on the
target system. However in the image they might be unwanted.
Thus this commit adds convenience methods to delete files
which gets automatically re-created by the services on
startup. Whether or not the methods are used is in the
responsibility of the author of the image descripion.
This Fixes bsc#1098535

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -1158,6 +1158,33 @@ function baseSetRunlevel {
 }
 
 #======================================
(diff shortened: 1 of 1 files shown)


schaefi:
yes, not touching machine-id currently as deleting it could have unexpected side effects as the article already points out.

PR #774, opened 2018-07-06
Add '--add-container-label' flag

### LINKED ISSUE #770: Add a way to specify the disturl
Taking the disturl from the /.buildenv file is a horrible idea. Please add a sane way the build script can use to pass in the disturl.

davidcassany:
This commits adds a command line flag to add a label for container image
types. The flag can be used multiple times.

Fixes #770

### DIFF
--- a/kiwi/tasks/system_build.py
+++ b/kiwi/tasks/system_build.py
@@ -28,6 +28,7 @@ usage: kiwi system build -h | --help
            [--delete-package=<name>...]
            [--set-container-derived-from=<uri>]
            [--set-container-tag=<name>]
+           [--add-container-label=<label>...]
            [--signing-key=<key-file>...]
        kiwi system build help
 
@@ -71,6 +72,10 @@ options:
         overwrite the container tag in the container configuration.
         The setting is only effective if the container configuraiton
         provides an initial tag value
+    --add-container-label=<name=value>
+        add a container label in the container configuration metadata. It
+        overwrites the label with the provided key-value pair in case it was
+        already defined in the XML description
     --set-repo=<source,type,alias,priority,imageinclude,package_gpgcheck>
         overwrite the first XML listed repository source, type, alias,
         priority, the imageinclude flag and the package_gpgcheck flag
@@ -164,6 +169,14 @@ class SystemBuildTask(CliTask):
             self.xml_state.set_container_config_tag(
                 self.command_args['--set-container-tag']
             )
+
+        if self.command_args['--add-container-label']:
+            for add_label in self.command_args['--add-container-label']:
+                if '=' in add_label:
+                    label_value_pair = add_label.split('=', maxsplit=1)
+                    self.xml_state.add_container_config_label(
+                        label_value_pair[0], label_value_pair[1]
+                    )
 
         if self.command_args['--set-container-derived-from']:
             self.xml_state.set_derived_from_image_uri(
--- a/kiwi/tasks/system_build.py
+++ b/kiwi/tasks/system_build.py
@@ -172,10 +172,13 @@ class SystemBuildTask(CliTask):
 
         if self.command_args['--add-container-label']:
             for add_label in self.command_args['--add-container-label']:
-                if '=' in add_label:
-                    label_value_pair = add_label.split('=', maxsplit=1)
-                    self.xml_state.add_container_config_label(
-                        label_value_pair[0], label_value_pair[1]
+                try:
+                    (name, value) = add_label.split('=', 1)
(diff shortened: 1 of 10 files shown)


davidcassany:
@mlschroe, in order to keep backward compatibility the disturl from `/.buildenv` is only parsed if there isn't any label using the `org.openbuildservice`, up to now the disturl label was `org.openbuildservice.disturl`. Is is a recommended practice for Docker and OCI to use reverse domains for labeling, thus I'd expect that any label included by the build script of OBS has this `org.openbuildservice` prefix in it, thus it make sense?

mlschroe:
Yes, fine with me. But maybe you should test for `org.openbuildservice.disturl` instead of `org.openbuildservice`?

kiwi/tasks/system_build.py
```diff
@@ -165,6 +170,14 @@ def process(self):                                      # noqa: C901
                 self.command_args['--set-container-tag']
             )
 
+        if self.command_args['--add-container-label']:
+            for add_label in self.command_args['--add-container-label']:
+                if '=' in add_label:
+                    label_value_pair = add_label.split('=', maxsplit=1)
+                    self.xml_state.add_container_config_label(
+                        label_value_pair[0], label_value_pair[1]
+                    )
```
schaefi: I think this could be done more explicit. imho it's a must have that the format is ```labelname=value``` Thus I would write the code like this:

```python
try:
    (name, value) = add_label.split('=', maxsplit=1)
    self.xml_state.add_container_label(name, value)
except Exception:
    log.warning(
        'Container label {0} ignored. Invalid format: expected labelname=value'.format(add_label)
    )
```

or even raise an error if in this situation we don't want to continue with the defaults

kiwi/tasks/system_prepare.py
```diff
@@ -152,6 +157,14 @@ def process(self):                                      # noqa: C901
                 self.command_args['--set-container-tag']
             )
 
+        if self.command_args['--add-container-label']:
+            for add_label in self.command_args['--add-container-label']:
+                if '=' in add_label:
+                    label_value_pair = add_label.split('=', maxsplit=1)
+                    self.xml_state.add_container_config_label(
+                        label_value_pair[0], label_value_pair[1]
+                    )
```
schaefi: see above

kiwi/tasks/system_prepare.py
```diff
@@ -207,6 +220,7 @@ def process(self):                                      # noqa: C901
         # make sure manager instance is cleaned up now
         del manager
 
+        log.info('Preparing repos')
```
schaefi: imho should be more verbose, preparing what repos ?

kiwi/xml_state.py
```diff
@@ -923,7 +966,7 @@ def get_volumes(self):
 
         * name: name of the volume
         * size: size of the volume
-        * realpath: system path to lookup volume data. If no mountpoint
+        * realpath: system h to lookup volume data. If no mountpoint
```
schaefi: sounds like a mistake ?

kiwi/tasks/system_build.py
davidcassany: Agree, this code is just better. I did not throw an error because we don't for `--set-container-tag` and I would expect both to behave similar. I'd be open to throw an error here also, I do not have strong opinion regarding that, I just tried to be consistent a similar flag.

kiwi/tasks/system_prepare.py
davidcassany: :sweat:  Damn, should not be here, this from days ago while debugging an image...

kiwi/xml_state.py
davidcassany: Yes  mistake... shame on me. Should stop doing any PR just right before lunch.

davidcassany:
> py 2.7 split doesn't like the split keyword. Actually I think we don't need the maxsplit as I would assume key=value and no other '=' in this information ?

Well it turns that py 2.7 accepts it, but not named. Thus removing the keyword would work for both. I don't know, if we should consider a possible value that includes `=`, I would not discard this option.

schaefi:
> if we should consider a possible value that includes =, I would not discard this option

ok

kiwi/tasks/system_build.py
schaefi: I think it's fine if we log it but we should avoid a silent ignore

kiwi/xml_state.py
schaefi: nevermind that just tests if I take the review serious ;)

PR #775, opened 2018-07-06
Activate btrfs quota groups

### LINKED ISSUE #772: Support setup of btrfs quota groups
## Problem description

On a btrfs filesystem one run into trouble when subvolumes/snapshots gets filled with data such that there is not enough space left to let btrfs operate its tree organization. In order to avoid that the concept of quota groups exists which can receive limits.

In the quota system, each subvolume is assigned a quota group in the form of

* $qgroupid/(subvolume id)

In kiwi the setup of a quota group per subvolume should be done at image creation time, such that
further limits can be assigned later easily with ```btrfs qgroup limit```

Details on the qgroup setup can be found here:

* https://btrfs.wiki.kernel.org/index.php/Quota_support

So far the basic steps to invoke after the filesystem and all its volumes/snapshots were created are:

```
qgroupid=0
btrfs quota enable /
btrfs subvolume list / | cut -d' ' -f2 | xargs -I{} -n1 btrfs qgroup create $qgroupid/{} /
btrfs quota rescan /
``` 

The value for the qgroupid should be obtained dynamically by checking used groups with level 1 and calculating a free number. The following can be done __after__ quota was enabled

```
btrfs qgroup show /
```

The selected qgroupid should be written to /etc/snapper/configs/root if the directory /etc/snapper exists in the system to allow a pre-setup of the snapper toolkit

schaefi:
Enable quota group system for btrfs based systems.
This Fixes #772

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -1254,7 +1254,7 @@ div {
         attribute btrfs_quota_groups { xsd:boolean }
         >> sch:pattern [ id = "btrfs_quota_groups" is-a = "image_type"
             sch:param [ name = "attr" value = "btrfs_quota_groups" ]
-            sch:param [ name = "types" value = "oem vmx docker" ]
+            sch:param [ name = "types" value = "oem vmx" ]
         ]
     k.type.btrfs_root_is_snapshot.attribute =
         ## Tell kiwi to install the system into a btrfs snapshot
@@ -1263,7 +1263,7 @@ div {
         attribute btrfs_root_is_snapshot { xsd:boolean }
         >> sch:pattern [ id = "btrfs_root_is_snapshot" is-a = "image_type"
             sch:param [ name = "attr" value = "btrfs_root_is_snapshot" ]
-            sch:param [ name = "types" value = "oem vmx docker" ]
+            sch:param [ name = "types" value = "oem vmx" ]
         ]
     k.type.btrfs_root_is_readonly_snapshot.attribute =
         ## Tell kiwi to set the btrfs root filesystem snapshot read-only
@@ -1275,7 +1275,7 @@ div {
         attribute btrfs_root_is_readonly_snapshot { xsd:boolean }
         >> sch:pattern [ id = "btrfs_root_is_readonly_snapshot" is-a = "image_type"
             sch:param [ name = "attr" value = "btrfs_root_is_readonly_snapshot" ]
-            sch:param [ name = "types" value = "oem vmx docker" ]
+            sch:param [ name = "types" value = "oem vmx" ]
         ]
     k.type.target_blocksize.attribute =
         ## Specifies the image blocksize in bytes which has to match
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -1248,6 +1248,14 @@ div {
             sch:param [ name = "attr" value = "xen_server" ]
             sch:param [ name = "types" value = "oem vmx" ]
         ]
+    k.type.btrfs_quota_groups.attribute =
+        ## activate the quota system if the filesystem is btrfs based.
+        ## By default the quota system is inactive
+        attribute btrfs_quota_groups { xsd:boolean }
+        >> sch:pattern [ id = "btrfs_quota_groups" is-a = "image_type"
+            sch:param [ name = "attr" value = "btrfs_quota_groups" ]
+            sch:param [ name = "types" value = "oem vmx docker" ]
+        ]
     k.type.btrfs_root_is_snapshot.attribute =
         ## Tell kiwi to install the system into a btrfs snapshot
         ## The snapshot layout is compatible with the snapper management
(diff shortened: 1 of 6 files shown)


schaefi:
This is a start and needs review by some btrfs experts. So the patch enables quota groups which automatically leads to group generation per subvolume. I have tested building an image with the following simple type setup:

```xml
<type image="vmx" filesystem="btrfs">
    <systemdisk>
        <volume name="home"/>
     </systemdisk>
</type>
```

Once booted I get the following subvolumes and groups information

```
btrfs subvolume list /
ID 256 gen 27 top level 5 path @
ID 257 gen 11 top level 256 path home

btrfs qgroup show /
qgroupid         rfer         excl 
--------         ----         ---- 
0/5          16.00KiB     16.00KiB 
0/256       799.75MiB    799.75MiB 
0/257        16.00KiB     16.00KiB 
255/257      16.00KiB     16.00KiB
```

If snapper is installed kiwi also wrote:

```
cat /etc/snapper/configs/root 
0
```

Is this the expected behavior ?

schaefi:
Hmm, good point. I think quota groups makes sense in general because of the thing that btrfs become completely unusable if a volume fills up with data and exceeds the limit from which one btrfs doesn't work anymore. But that's just my personal understanding and I'm far away from being a btrfs expert. Surely Arvin could answer the question

schaefi:
@davidcassany there was no response by Arvin, probably he's busy with other stuff. I agree we should have an attribute to control qgroups enablement. I added:

```xml
<type ... btrfs_quota_groups="true"/>
```

Would you mind to take a short look again ?

Thanks

kiwi/schema/kiwi.rnc
```diff
@@ -1248,6 +1248,14 @@ div {
             sch:param [ name = "attr" value = "xen_server" ]
             sch:param [ name = "types" value = "oem vmx" ]
         ]
+    k.type.btrfs_quota_groups.attribute =
+        ## activate the quota system if the filesystem is btrfs based.
+        ## By default the quota system is inactive
+        attribute btrfs_quota_groups { xsd:boolean }
+        >> sch:pattern [ id = "btrfs_quota_groups" is-a = "image_type"
+            sch:param [ name = "attr" value = "btrfs_quota_groups" ]
+            sch:param [ name = "types" value = "oem vmx docker" ]
```
davidcassany: I don't see why docker is there... I believe it should only be enabled for disk images, thus oem and vmx only.

davidcassany: > found that btrfs_root_is_readonly_snapshot as well as btrfs_root_is_snapshot is also targeted to be valid for docker. I also think that's wrong but wasn't 100% sure

mmm yes that looks like a mistake... Then I guess I am the one to blame :sweat: AFAIR there is no filesystem management at all in containers code.

kiwi/schema/kiwi.rnc
```diff
@@ -1248,14 +1248,22 @@ div {
             sch:param [ name = "attr" value = "xen_server" ]
             sch:param [ name = "types" value = "oem vmx" ]
         ]
+    k.type.btrfs_quota_groups.attribute =
+        ## activate the quota system if the filesystem is btrfs based.
+        ## By default the quota system is inactive
+        attribute btrfs_quota_groups { xsd:boolean }
+        >> sch:pattern [ id = "btrfs_quota_groups" is-a = "image_type"
+            sch:param [ name = "attr" value = "btrfs_quota_groups" ]
+            sch:param [ name = "types" value = "oem vmx" ]
```
davidcassany: Yes this makes more sense

kiwi/schema/kiwi.rnc
```diff
@@ -1267,7 +1275,7 @@ div {
         attribute btrfs_root_is_readonly_snapshot { xsd:boolean }
         >> sch:pattern [ id = "btrfs_root_is_readonly_snapshot" is-a = "image_type"
             sch:param [ name = "attr" value = "btrfs_root_is_readonly_snapshot" ]
-            sch:param [ name = "types" value = "oem vmx docker" ]
+            sch:param [ name = "types" value = "oem vmx" ]
```
davidcassany: Thanks

kiwi/schema/kiwi.rnc
```diff
@@ -1248,14 +1248,22 @@ div {
...
+        ]
     k.type.btrfs_root_is_snapshot.attribute =
         ## Tell kiwi to install the system into a btrfs snapshot
         ## The snapshot layout is compatible with the snapper management
         ## toolkit. By default no snapshots are used
         attribute btrfs_root_is_snapshot { xsd:boolean }
         >> sch:pattern [ id = "btrfs_root_is_snapshot" is-a = "image_type"
             sch:param [ name = "attr" value = "btrfs_root_is_snapshot" ]
-            sch:param [ name = "types" value = "oem vmx docker" ]
+            sch:param [ name = "types" value = "oem vmx" ]
```
davidcassany: Thanks for the fix :+1:

PR #778, opened 2018-07-10
Fixed some code smells

schaefi:
* Use of unused variables, and dangerous defaults
* Update of design patterns on pylint and landscape


schaefi:
maybe we should discuss alternatives to landscape.io. The site is pretty slow. I'm open for any better

kiwi/utils/command_capabilities.py
```diff
@@ -50,10 +50,11 @@ def has_option_in_help(
 
         :rtype: bool
         """
+        help_args = help_flags or ['--help']
```
davidcassany: Just out of curiosity, what is the smelling part of having default values different than `None`? I am perfectly fine with this modified code, just want to understand the landscape reasoning.

.pylintrc
```diff
@@ -0,0 +1,34 @@
+[DESIGN]
```
davidcassany: I guess this complete file is here to better match the analysis criteria with our own coding styling right? I assume most of the values are defaults or slightly variations of them.

kiwi/utils/command_capabilities.py
schaefi: See this: https://www.pythoncentral.io/fun-with-python-function-parameters/
[Potential Python Function Parameter Problems]

.pylintrc
schaefi: yes this is in case we run pylint on the code. I have adapted some default values to match our criteria. My hope is that landscape will import that file and apply it to the pylint call they run

schaefi: problem is that landscape.io is sooo slow that I still have no check results.

I'm not sure if that service will survive. Thus I would be ok if we switch to some other service, e.g codacy.

We also should check how we integrate with github services vs. apps. Travis as an example will move us over to travis.ci.com, we just have to wait. I don't know if landscape and friends already supports github apps as they should.

Would be great if you have some experience here to share your thoughts 

Thanks

kiwi/utils/command_capabilities.py
davidcassany: Wow, interesting reading... That's valuable knowledge, thanks!

.pylintrc
davidcassany: To be honest I don't have experience or background to help here. However I can spend some time tomorrow to get some ;)  and experiment with landscape, some alternatives and do some digging.

schaefi: Thanks for the offer. If you are interested in code analytic it's surely nice to see what they do. These are all external services which needs approval by the github org maintainers. Thus playing with them is only easily possible in your personal github account. For the moment I'll stick with landscape.io and will jump to something more productive. Thanks

PR #779, opened 2018-07-11
Support label attribute in volumes

### LINKED ISSUE #738: Add labels to partitions
Root, boot, efi, and swap partitions are created using labels. Is it possible to extend this to allow users to set their own custom labels for partitions?

For example:

```xml
<systemdisk name="System">
  <volume name="tmp" size="10G" label="tmp"/>
  <volume name="mysql" size="30G" mountpoint="/data/mysql" label="mysql"/>
  <volume name="home" size="5G" label="misleadinglabel"/>
</systemdisk>
```

schaefi:
The optional label attribute in a volume section allows
to specify a filesystem label for the selected volume.
The label setup will have no effect on filesystems
which implements their own volume management like it's
the case for btrfs. This Fixes #738

### DIFF
--- a/kiwi/runtime_checker.py
+++ b/kiwi/runtime_checker.py
@@ -102,6 +102,39 @@ class RuntimeChecker(object):
                 message % (target_dir, shared_cache_location)
             )
 
+    def check_volume_label_used_with_lvm(self):
+        """
+        The optional volume label in a systemdisk setup is only
+        effective if the LVM, logical volume manager system is
+        used. In any other case where the filesystem itself offers
+        volume management capabilities there are no extra filesystem
+        labels which can be applied per volume
+        """
+        message = dedent('''\n
+            Custom volume label setup used without LVM
+
+            The optional volume label in a systemdisk setup is only
+            effective if the LVM, logical volume manager system is
+            used. Your setup uses the {0} filesystem which itself
+            offers volume management capabilities. Extra filesystem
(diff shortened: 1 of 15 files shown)


schaefi:
@davidcassany Do we want a runtime check to let the build fail if labels are assigned when btrfs is in use ?

davidcassany:
> Do we want a runtime check to let the build fail if labels are assigned when btrfs is in use?

I think there should be some sort of validation, it could mean some warning messages to inform the user that labels will be ignored. Or even it could be a schematron rule to block labels usage if btrfs file system is selected. 

In general I like the idea to enforce as much as possible within the schema itself, but abusing of schematron rules turns schema hard to read, and schematron issues are not simple to debug and detect. Schematron code can easily grow to something unmaintainable. I remember I tried to find out a generic pattern in the past to check incompatible attributes, but I was not convinced with the results I got.

Thus I'd be fine with some runtime check, in which case I'd say failing is better than warning. We could also warn in volume_manager/btrfs.py instead of failing, but I don't really like this idea.

schaefi:
yep this all matches what I thought too. I'll come up with a runtime check next... Thanks for your feedback

davidcassany:
Looks good to me :+1: Also the codacy complains only about the assert usage in unit tests,  this is great.

schaefi:
all right as I don't see why an assert in a unit test is something bad I'll go and merge it :smile:

PR #781, opened 2018-07-13
Fixed GCE image file name

schaefi:
In former times Google requires the image name to follow
their naming conventions. However that seems to have changed
and it is no longer required to match a certain pattern.
Thus this patch changes the output name of the GCE tar file
to use the same naming schema as KIWI applies to its output
files


schaefi:
@rjschwei if you look at this take a closer look on the test which outlines the name change best. As I said I had in mind that google applies a name pattern but that was from the times I used the sdk. I did an upload today using the web console and it seems they don't care anymore

PR #783, opened 2018-07-16
Fixed building VMware images with pvscsi adapter

schaefi:
Qemu does not natively support the pvscsi adapter type.
However there is a VMware suggested procedure which allows
to change the lsilogic setup to pvscsi inside of the DDB
of a formerly created lsilogic configured image format.
This patch implementes that procedure and
Fixes bsc#1099569

### DIFF
--- a/kiwi/storage/subformat/vmdk.py
+++ b/kiwi/storage/subformat/vmdk.py
@@ -17,6 +17,11 @@
 #
 import os
 
+# In python2 bytes is string which is different from
+# the bytes type in python3. The bytes type from the
+# builtins generalizes this type to be bytes always
+from builtins import bytes
+
 # project
 from kiwi.storage.subformat.base import DiskFormatBase
 from kiwi.command import Command
@@ -24,7 +29,7 @@ from kiwi.storage.subformat.template.vmware_settings import (
     VmwareSettingsTemplate
 )
 
-from ...exceptions import (
+from kiwi.exceptions import (
     KiwiTemplateError
 )
 
@@ -44,6 +49,10 @@ class DiskFormatVmdk(DiskFormatBase):
         self.image_format = 'vmdk'
         self.options = self.get_qemu_option_list(custom_args)
 
+        self.patch_header_for_pvscsi = False
+        if custom_args and 'adapter_type=pvscsi' in custom_args:
+            self.patch_header_for_pvscsi = True
+
     def create_image_format(self):
         """
         Create vmdk disk format and machine settings file
@@ -56,6 +65,8 @@ class DiskFormatVmdk(DiskFormatBase):
                 self.get_target_file_path_for_format('vmdk')
             ]
         )
+        if self.patch_header_for_pvscsi:
+            self._inject_pvscsi_adapter_type()
         self._create_vmware_settings_file()
 
     def store_to_result(self, result):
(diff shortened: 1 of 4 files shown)


schaefi:
* https://bugzilla.suse.com/show_bug.cgi?id=1099569

kiwi/storage/subformat/vmdk.py
```diff
@@ -168,3 +179,24 @@ def _create_vmware_settings_file(self):                 # noqa: C901
...
+            vmdk_descriptor = bytes(vmdk.read(1024))
+
+        if vmdk_descriptor:
+            vmdk_descriptor = vmdk_descriptor.replace(
+                b'"lsilogic"', b'"pvscsi"'
+            )
+            with open(vmdk_image_name, 'r+b') as vmdk:
+                vmdk.seek(512, 0)
+                vmdk.write(vmdk_descriptor)
+                vmdk.seek(0, 2)
```
davidcassany: Is this line needed? I don't see why the fd should be moved up to the end after writing the descriptor

davidcassany:
@schaefi by the way I just saw you got a green test in codacy despite using asserts in tests, did you configure anything for that?

davidcassany:
> @schaefi by the way I just saw you got a green test in codacy despite using asserts in tests, did you configure anything for that?

Oh, answering myself. Just saw this verification is disabled at the UI :+1:

kiwi/storage/subformat/vmdk.py
schaefi: If you open with r+b meaning open for read and write and you close the file descriptor the file gets truncated at the position of the file pointer. Thus to my knowledge from the university it's required to position to the end before closing... might be old fashioned though and I'm open to change it if that's no longer needed.

schaefi:
> Oh, answering myself. Just saw this verification is disabled at the UI +1

yes it's possible to click on the issue and then remove the pattern for the entire project. This is a great setup possibility :)

schaefi:
Regarding your concerns on writing in the header I fully agree this is a hack and I really would like to avoid it. However if you take a look into the referenced bug report you see that there seems to be legal issues which doesn't allow qemu to support pvscsi adapter natively. I haven't looked closer but the qemu people won't accept the suggested patch.

From VMware directly there is a suggestion how to change an lsilogic target (which is also scsi) into a pvscsi target and that way seems to be ok because kiwi as the builder just creates that target but does not boot it. Booting seems to be only allowed by VMware's virtualization. I would expect you will not find any opensource player which can boot the pvscsi target... which of course limits our testing capabilities.

All this proprietary stacks just causing trouble. I remember well the loops we drive when adding support for vmware's ova target via ovftool. After all I think it's important to stay as user friendly as possible. Those who want to build an image for the pvscsi adapter should at least be able to do it with kiwi if that does not violate our code and policies.

Thoughts ?

liayan:
Actually the patch you submitted to QEMU is doing the same as commit here. QEMU will create the "pvscsi" image same as "lsilogic" with your patch. The reason QEMU does not support "pvscsi" is because QEMU is not responsible for build such an image, and "pvscsi" as a para-virtualization device needs a hypervisor level support, also vmdk of qemu is totally following VMDK spec file. So kiwi as a image build tool, could support "pvscsi" by 1) different back-end tool from vmware side, or 2) a quirk like here.

schaefi:
@liayan thanks for the details, very much appreciated

> Actually the patch you submitted to QEMU

That one was done by Jochen Roeder, not me :)

davidcassany:
@schaefi, I could not manage to boot the image I did with `controller="pvscsi"` using VMWare Player 14.1.2, it does not even reach grub, it complaints about not finding any OS in disk.

Still debugging, I doubt I have to configure anything on VMWare side, I understand it should be straight forward. Still debbuging.

schaefi:
> I doubt I have to configure anything on VMWare side

yes it should be enough to let the VMware player read the .vmx file written by kiwi. In that file the reference to the .vmdk is listed. Sometimes the guestOS attribute is required to be set.

If in doubt you can also create a new machine file using VMware tools and just assign our .vmdk as os disk

liayan:
@davidcassany , I doubt it could work on a simple vmware player, pvscsi is a para virtualization device, which means it also needs host driver support. You may need try a vmware workstation or ESXi.

davidcassany:
On the other hand, I could not manage to test it with VMPlayer... I could only verify the file is patched as expected and described in bugzilla by checking the result (`head -c 1024 <imagegile>`). I doubt it is possible to test it with VMPlayer, after googling looks like this controller is for remote instances.

kiwi/storage/subformat/vmdk.py
```diff
@@ -44,6 +49,10 @@ def post_init(self, custom_args):
         self.image_format = 'vmdk'
         self.options = self.get_qemu_option_list(custom_args)
 
+        self.patch_header_for_pvscsi = False
```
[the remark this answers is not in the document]
schaefi: Thanks for catching it. This is exactly the reason why I don't like super calls. The oo design pragma of loosely coupled code gets lost with it. I will fix the DiskFormatOva implementation which performs the super call causing the trouble.

schaefi:
Thanks much for testing this. It seems like testing pvscsi requires a running ESX server and I don't have one around. I also think we spent enough time on the proprietary part. From an image perspective two people verified the image content wise, thus let's go on with this once you @davidcassany are ok with my last commit

Thanks

PR #791, opened 2018-07-30
Add build tests for CentOS

davidcassany:
Adding a CentOS integration test


davidcassany:
See the staging builds [here](https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86/test-image-centos)

I'll merge when confirmed all tests are green

PR #794, opened 2018-07-31
Fixed make obs_test_status

schaefi:
The helper script .obs_test_status looks up the build
results from the integration tests. With the introduction
of multibuild integration tests the script has to apply
some modifications to get the correct results
This is related to Issue #791

### DIFF
--- a/.obs_test_status
+++ b/.obs_test_status
@@ -5,7 +5,7 @@ function kiwi_version {
 }
 
 function kiwi_build_version {
-    echo -n "$1" | grep python3-kiwi- | head -n 1 | cut -f 3 -d -
+    echo -n "$1" | grep python[32]-kiwi- | head -n 1 | cut -f 3 -d -
 }
 
 kiwi_local_version=$(kiwi_version)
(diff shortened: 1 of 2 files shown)


schaefi:
hrm, something is broken with travis and the dev env

schaefi:
It's kind a hidden but imho helpful to get a quick status of the integration test builds :)

Have you seen the commit changing the xattr requirement in the .virtualenv setup. It was the only quick fix I could come up with. Are you ok if we keep that until the latest pip version of xattr has been fixed. Otherwise I think none of the follow up pull request will have a successful travis test result

Thoughts ?

davidcassany:
> Have you seen the commit changing the xattr requirement in the .virtualenv setup. It was the only quick fix I could come up with. Are you ok if we keep that until the latest pip version of xattr has been fixed. Otherwise I think none of the follow up pull request will have a successful travis test result

Yes I saw it and I am fine with it. The only alternative I see is manually installing this concrete version on `.travis.yml` within the install phase, this way it wouldn't be needed to set a concrete version in `.virtualenv` file. However I find this option a bit obscure,  I prefer having the dependency in `.virtualenv` as this is the place I'd expect to find dependencies listed.

schaefi:
ok, yes that's what I thought too, let's go for it and once this is resolved on pip we can go back

PR #797, opened 2018-08-06
Delete dmraid aka: softraid soft/fakeraid support

schaefi:
In fate#323743 the decision was made to drop dmraid from the distribution. Along with the low business case for those controllers and the support for linux softraid via mdadm we also drop the support in kiwi for oem-ataraid-scan

* https://fate.suse.com/323743


schaefi:
@davidcassany the doc update came just to my mind when testing. I realized some links have permanent redirect locations and just fixed them in this pull request... even though not related to the dmraid drop

schaefi:
Thanks for the review, yes less code is better :)

PR #800, opened 2018-08-07
Fix baseStripUnusedLibs config method

### LINKED ISSUE #798: libnss_* is being stripped from netboot
## Problem description
I'm building a LEAP 15.0 PXE image from Tumbleweed that uses NBD root:

`    <type image="pxe" filesystem="squashfs" boot="netboot/suse-leap15.0"/>
`

After the initrd is created, it cannot mount root because DNS fails. The root cause is missing libnss_* libraries. There were stripped.

```
$ grep libnss_dns build/image-root.log 
DEBUG: --> kiwi_strip_libs='libdmraid-events-isw libfontenc libfreetype libgcc_s libnsl libnss_compat libnss_dns libnss_files libpng libresolv librt libsysfs libutempter libutil libkmod libselinux libsepol libdevmapper libply-boot-client libply-splash-graphics libsplashy libsplashycnf libjpeg libkms libaio libdrm'
DEBUG: Removing library: /lib64/libnss_dns-2.26.so
```

I copied the `<strip type="libs">` section from `/usr/lib/python3.6/site-packages/kiwi/config/strip.xml` into my kiwi configuration. However, it did not change the behavior.

## Expected behaviour
libnss_dns should not be removed, specially from a netboot initrd

* Build a LEAP 15.0 pxe image using netboot/suse-leap15.0
* Configure KIWI to mount nbd. Use FQDN for kiwiserver and nbd-server, not ip address
* Boot will fail because it cannot resolve FQDN into an IP (BTW, host/dig does work)

+ KIWI version: 9.16.2
+ Operating system: OpenSUSE Tumbleweed
+ OBS version: none (local build)

davidcassany:
This commit arguments handling of the baseStripUnusedLibs
that was not prepared to handle quoted variable containing a list.

Fixes #798

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -552,18 +552,17 @@ function baseStripUnusedLibs {
     # /.../
     # add exceptions
     # ----
-    while [ ! -z "$1" ];do
+    for j in $1; do
         for i in \
-            /lib*/$1* /usr/lib*/$1* \
-            /lib/x86_64-linux-gnu/$1* /usr/lib/x86_64-linux-gnu/$1* \
-            /usr/X11R6/lib*/$1*
+            /lib*/$j* /usr/lib*/$j* \
+            /lib/x86_64-linux-gnu/$j* /usr/lib/x86_64-linux-gnu/$j* \
+            /usr/X11R6/lib*/$j*
         do
             if [ -e "$i" ];then
                 needlibs[$count]=$i
                 count=$((count + 1))
             fi
         done
-        shift
     done
     # /.../
     # find unused libs and remove it, dl loaded libs
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -552,14 +552,14 @@ function baseStripUnusedLibs {
     # /.../
     # add exceptions
     # ----
-    for j in $1; do
-        for i in \
-            /lib*/$j* /usr/lib*/$j* \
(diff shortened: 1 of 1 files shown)


davidcassany:
@schaefi, when cleaning `functions.sh` to match shellcheck criteria many quotes were added to variables to avoid derived issues of spaces (and other special characters) expansion. In this case a variable containing a list was passed as a function argument. Without quotes the method was getting as many arguments as elements in the list, with quotes, all elements were passed a single argument, that caused the reported issue.

I looked again into the the commit (11b3a2bb3255e0c0d8385ea52ec6070740dee2b0) and I could not see any other list passed as an argument. I had a quick look to variables in which brackets and quotes were appended.

kiwi/config/functions.sh
```diff
@@ -552,18 +552,17 @@ function baseStripUnusedLibs {
     # /.../
     # add exceptions
     # ----
-    while [ ! -z "$1" ];do
+    for j in $1; do
```
schaefi: Thanks for the fix this is much better. While we are on it why not use a more meaningful variable e.g replace 'j' with 'libname' or similar. Thoughts ?

davidcassany: yes I am fine with using better variable names. I'll change both `$j` and `$i`. I used `$j` just because the other inner loop was using `$i`. Next one would have been `$k`, `$l` ... :grinning: 

Gonna change it now

schaefi:
so much better to read :+1: 

Thanks

PR #801, opened 2018-08-08
Added support for system wide config file

schaefi:
If there is no user specific config file we are also
looking for a system wide /etc/kiwi.yml file

### DIFF
--- a/kiwi/runtime_config.py
+++ b/kiwi/runtime_config.py
@@ -31,7 +31,8 @@ class RuntimeConfig(object):
     """
     **Implements reading of runtime configuration file:**
 
-    * ~/.config/kiwi/config.yml
+    1. ~/.config/kiwi/config.yml
+    2. /etc/kiwi.yml
 
     The KIWI runtime configuration file is a yaml formatted file
     containing information to control the behavior of the tools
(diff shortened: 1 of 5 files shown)


schaefi:
We need this to allow using a config file inside of the obs buildservice. The environment there does not populate a user thus there is no "home" and you won't be able to provide user specific data

schaefi:
I'd like to continue testing this in the pubcloud builds and need to go ahead merging it. If you see issues just let me know and we address them in a follow up request. Thanks

davidcassany:
@schaefi, sure no problem, having a look now. Also if you are in rush and I can help you in some way let me know.

schaefi:
Thanks for the after merge review :) I wanted to fix that while my pubcloud team was still at work, which was a night session at my place but that's ok we solved it

> how are you providing the configuration file in the build service. Using specific RPM that provides it

yes if you look at:

Devel:PubCloud/kiwi-sle12-config

you will find a package kiwi_sle12_config which provides an /etc/kiwi.yml

In the projconf of the project Devel:PubCloud:GoogleCompute:ImagesSLE124P
you see the following entries:

preinstall: kiwi_sle12_config
Ignore: python3-kiwi:xorriso

which allows me to place the kiwi.yml into the worker before kiwi starts and overwrites the xorriso requirement of the latest kiwi version we want to use in the Devel space for sles12. sles12 does
not provide any xorriso and the config file switches the iso tool_category back to cdrtools

schaefi:
The kiwi version we submit to sle12 SP4 officially is from our branch and there we have changed the default iso tool setup to match the distro but for building pubcloud images in Devel we want to use the latest version of kiwi and port required features to sles12 as we see fit

PR #802, opened 2018-08-08
Update pxe server setup documentation

schaefi:
Delete the suggested modifications to /etc/sysconfig/atftpd
and trust the defaults provided by the package


schaefi:
@davidcassany well it's not a big shot but I think the suggested changes to the config are misleading so let's delete them

schaefi:
I have to continue with a test for pubcloud. I guess this one doesn't need a deep review. If you see issues feel free to re-open. Thanks

PR #805, opened 2018-08-20
Fixed deploy target

schaefi:
As part of the deploy process in travis a bundle to pypi is
uploaded. The bundle is missing the compiled manual pages because
the doc_travis stage did not create them.


schaefi:
Since we moved to travis stages this problem got introduced and I didn't noticed so far

PR #809, opened 2018-08-25
Move dracut modules into its own source archive

schaefi:
The dracut modules like kiwi provides it should not be part
of the default install target. If kiwi gets installed from
source or via pip all dracut code gets installed on that
system which is unwanted and in the worst case leads to
boot trouble next time this system rebuilds its initrd
via dracut. The dracut modules are built into sub-packages
and serves their purpose for the specific build type of
the image. They are not needed at all on the machine
building the images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,14 +25,15 @@ tools:
 	# called
 	${MAKE} -C tools all
 
+install_dracut:
+	install -d -m 755 ${buildroot}usr/lib/dracut/modules.d
+	cp -a dracut/modules.d/* ${buildroot}usr/lib/dracut/modules.d
+
 install:
 	# apart from all python source we also need to install
 	# the C tools, the manual pages and the completion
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
Please include it in the source archive, but have it so that pip doesn't install them. Multiple source archives will be trickier for me to deal with since the Fedora package is sourced from PyPI right now.

schaefi:
ok let me check how that can be achieved. Do you know a way to let "python setup.py install" skip a certain part of the sdist source ?

schaefi:
@Conan-Kudo yes that sounds much better, thanks. How about it now ?

Conan-Kudo:
Looks great to me!

Conan-Kudo:
@schaefi Once you squash the commits and clean it up, it looks good to me. :)

PR #812, opened 2018-09-04
Fix quota groups management when snapper is present

davidcassany:
* Fixes the config file path if root is snapshot
* Uses the correct `QGROUP="<group>"` syntax
* Do not overwrite the config file if already present

Fixes bsc#1093518


kiwi/volume_manager/btrfs.py
```diff
@@ -354,19 +354,41 @@ def _xml_pretty(self, toplevel_element):
...
+                self._get_quota_group_id(), os.linesep
+            )
+            snapper_root_lines = []
+            found = False
+            if os.path.exists(snapper_config_file):
+                with open(snapper_configs + 'root', 'r') as snapper_root:
+                    snapper_root_lines = snapper_root.readlines()
+                found = False
+                for i, line in enumerate(snapper_root_lines):
+                    if line.startswith('QGROUP'):
```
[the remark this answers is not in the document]
davidcassany: Right, I'll make this change.

schaefi:
The code looks good, but I have a couple questions:

1. Thanks David for fixing the format. The initial implementation was based on iirc information from Arvin Schnell which included just a number for the qgroup into the config file. The change here now changes that into a format QGROUP=<value_from_before>. My question is; is that a compatible change or something new only working with a specific version of snapper ?

2. The changes here reads in an eventual existing snapper root file and adds our qgroup setup only if no other qgroup is configured. I think we should not do that because there is only one valid qgroup for the image and that's the one the image builder has created in the scope of this image. Allowing to overlay that file from somewhere with custom content is something I think we should not do, which is the reason why my implementation did not allow it. Thus I'd like to understand the use case for it, are there are other configuration options allowed in etc/snapper/configs/root which could be pre-defined by an overlay version of that ? if not and this is only about the qgroup I strongly recommend to leave that value exclusively to the image builder to avoid building inconsistent images

Thoughts ?

Vogtinator:
> The changes here reads in an eventual existing snapper root file and adds our qgroup setup only if no other qgroup is configured.

This is wrong. It adds `QGROUP='x/x'` if it's not present in the file and overrides the existing value otherwise.

schaefi:
yes you are right, I overlooked ```snapper_root_lines[i] = config_line``` in case QGROUP was found. Ok so we set "our" value in any case which is good. But still I'd like to understand what lead us to the relative complex code writing this file, what other configuration values can go to snapper/configs/root ?

Reading ```man snapper-configs``` confused me:

```
DESCRIPTION
       Each file /etc/snapper/configs/* describes a snapper config.

so there are more ?

VARIABLES
       The following is a list of variables that can be present in the
       configuration file. Boolean values must be "yes" or "no".

that indicates there is only one ?
```

If all of the VARIABLES can go to  snapper/configs/root I'm fine with the implementation
but if not we could implement our part with more simple coding

Vogtinator:
Read `/etc/snapper/config-templates/default` for an example config file.

There can be multiple snapper configurations, but the active one(s) are mentioned in /etc/sysconfig/snapper.

Maybe the choice of `root` in kiwi should be documented or it should read sysconfig to get the config file to change.

schaefi:
Thanks, that sounds like the better approach to me. So instead of us writing a config file we just change the QGROUP value in /etc/snapper/config-templates/default. The file is provided by libsnapper4 and under package manager control which is great.

Thoughts ?

Vogtinator:
No, you completely misunderstood.

Do NOT change that file. a) It's a system file and not configuration. b) Not actually read by anything.

schaefi:
ok, out of curiosity, /etc/sysconfig/snapper does not serve for that purpose ?

schaefi:
and in my /etc/sysconfig/snapper I have

```
## Path: System/Snapper
## Type:        string
## Default:     ""
# List of snapper configurations.
SNAPPER_CONFIGS=""
```

so it would not read it ?

Vogtinator:
> ok, out of curiosity, /etc/sysconfig/snapper does not serve for that purpose ?

No, sysconfig/snapper only tells snapper which configuration to use.

> and in my /etc/sysconfig/snapper I have
> ```SNAPPER_CONFIGS=""```
> so it would not read it ?

I think so, yes.

schaefi:
/etc/sysconfig/snapper is not under package manager control, what provides and manages that file in suse ?

If I got it right we could pick our own config file, only changing QGROUP and register that in /etc/sysconfig/snapper SNAPPER_CONFIGS. But I guess that opens another can of worms, in which order are the files processed, what happens with conflicting values ?

Vogtinator:
> /etc/sysconfig/snapper is not under package manager control, what provides and manages that file in suse ?

fillup templates.

> If I got it right we could pick our own config file, only changing QGROUP and register that in /etc/sysconfig/snapper SNAPPER_CONFIGS. But I guess that opens another can of worms, in which order are the files processed, what happens with conflicting values ?

It's easy to keep the can of worms closed. If SNAPPER_CONFIGS is empty, do the equivalent of `snapper create-config` and copy the template as new `root` config. If it contains one config, edit the file and set `QGROUP`. If it contains more than one, error out.

That keeps both the default (no changes in config.sh) and the advanced use (config.sh sets up the config) working.

schaefi:
ok so that basically answers what I was concerned about. The snapper create-config/set-config commands should be used to manage the config files and the approach to manually create them seems possible but from a design point of view rather unwanted.

In this case we could also consider to call snapper if the tool is present in the image root:

```
# just create root config, no auto detection
snapper create-config -f btrfs /

# set the quota group ids
snapper set-config QGROUP=x/y
```

imho this would basically the same as implemented here but we don't enforce a config file name

Vogtinator:
If you only call `snapper --no-dbus create-config` if the config does not already exist, that should work as well. Otherwise it might overwrite config.sh changes.

davidcassany:
right, using snapper looks like a way cleaner solution, I am working on it. @schaefi, @Vogtinator thanks for your analysis

schaefi:
Great, thanks David for adapting the code

davidcassany:
@schaefi, @Vogtinator unfortunately the suggested approach does not work in case the `btrfs_root_is_snapshot` is enabled. `snapper create-config` assumes the command call is called in a system that lacks any snapshot configuration and related subvolumes, so `snapper create-config` attempts to create the `.snapshots` subvolume and fails because it is already there. It seams that `snapper create-config` is not capable to reuse/include already existing snapshots. If `btrfs_root_is_snapshot` is enabled it looks like there is the chicken and egg problem to run `snapper create-config`.

Vogtinator:
YaST uses `/usr/lib/snapper/installation-helper` to do that job. It has no documentation though.

davidcassany:
> If SNAPPER_CONFIGS is empty, do the equivalent of snapper create-config and copy the template as new root config. If it contains one config, edit the file and set QGROUP. If it contains more than one, error out.

I think we can do something like that and not call 'snapper create-config', but I am wondering what happens if `btrfs_root_is_snapshot` is disabled and a snapper configuration is added without creating the `.snapshots` subvolume. @Vogtinator do you have any clue about how I should that it?

Vogtinator:
If `btrfs_root_is_snapshot` is not true, it shouldn't create a snapper config.

schaefi:
yes I think if btrfs_root_is_snapshot is disabled snapper is not functional as it requires this core snapshot setup. So you can safely skip everything we do here with the snapper config in this case

kiwi/volume_manager/btrfs.py
```diff
@@ -354,19 +357,48 @@ def _xml_pretty(self, toplevel_element):
...
     def _create_quota_group_info(self):
-        snapper_configs = ''.join([self.mountpoint, '/@/etc/snapper/configs/'])
-        if os.path.exists(snapper_configs):
-            with open(snapper_configs + 'root', 'w') as snapper_root:
-                snapper_root.write(
-                    '{0}{1}'.format(self._get_quota_group_id(), os.linesep)
-                )
+        root_path = os.sep.join([self.mountpoint, '@/.snapshots/1/snapshot'])
+        snapper_default_conf = os.sep.join(
+            [root_path, 'etc/snapper/config-templates/default']
```
schaefi: I think that information should be move into the Defaults space

kiwi/volume_manager/btrfs.py
```diff
@@ -354,19 +357,48 @@ def _xml_pretty(self, toplevel_element):
...
+            config_file = self._set_snapper_sysconfig_file(root_path)
+            if not os.path.exists(config_file):
+                shutil.copyfile(snapper_default_conf, config_file)
+            Command.run([
+                'chroot', root_path, 'snapper', '--no-dbus', 'set-config',
+                'QGROUP={0}'.format(self._get_quota_group_id())
+            ])
+
+    @classmethod
+    def _set_snapper_sysconfig_file(self, root_path):
```
schaefi: For this type of coding kiwi already provides the SysConfig class. See kiwi/utils/sysconfig.py
I suggest to make use of it and extend the class if needed, but so far I think it already provides what you need

davidcassany:
@Vogtinator with the current code I use `snapper --set-config QGROUP=x/y` (using the quota group given by `btrfs qgroup show -f /` which is one of the auto created groups of level 0), however I am not sure this produces the expected result, as this is different from what `snapper setup-quota` does. The latter creates a new quota group with level 1, sets it on the `root` config and probably does other things that I haven't noticed.

Using `snapper setup-quota` was my first attempt but I had issues, it happens something similar to the failure with `snapper create-config`, snapper expects a different setup at the time of running it.

@schaefi what about this implementation? I am not very happy with it, as I had to include multiple specific file paths. Using the in image snapper binary causes issues and the `/usr/lib/snapper/installer-helper` tool doesn't look like a good option. I had a look the helper code and noted that it is a completely undocumented tool that has many options and does multiple different things, but looks like some kind of ad-hoc code to be used by some other application. I would not couple KIWI to that, it doesn't look like it is going to have a stable API.

kiwi/volume_manager/btrfs.py
davidcassany: right

kiwi/volume_manager/btrfs.py
davidcassany: Oh, I missed that, checking it now

schaefi:
This looks really good now :+1:  Thanks

davidcassany:
@Vogtinator, @aschnell do you have any further comment for this PR? or did you had the chance to give it a try? We haven't submitted it yet and we are holding some other minor backports to avoid creating multiple MRs. I'd like to move on regarding this quota issue and submit it to SLE-15 and SLE-12-SP4, but before that I'd also like some validation or confirmation from you. Thanks

Vogtinator:
@aschnell ping.

aschnell:
I do not understand _get_quota_group_id() or the general approach. Where is the higher level qgroup (something like "1/0") created?

aschnell:
A higher level qgroup is required as explained in bsc #1093518 commnt #8.

PR #816, opened 2018-09-12
Fixed overlay of intermediate config files

### LINKED ISSUE #807: Using etc/hosts as overlay file has no effect
## Problem description

Some config files ```etc/hosts, etc/resolv.conf``` and ```etc/sysconfig/proxy``` needs to be temporarly
copied from the buildsystem host to the image root system in order to allow e.g DNS resolution in the 
way as it is configured on the buildsystem host.

These files gets restored to their original content or replaced to the version provided by a package
at the end of the kiwi prepare step.

The problem is if a user has those files listed in its overlay file tree. The overlay files gets copied and once the restore code mentioned above kicks on they get overwritten again which is unwanted in this situation.

## Expected behaviour

Give overlay files the highest priority

schaefi:
Some config files e.g etc/hosts needs to be temporary copied from the buildsystem host to the image root system. This is done by a custom copy with the .kiwi extension and a symlink to that file.

During the installation process the package manager either overwrites the file or creates a .rpmnew variant. In case a .rpmnew variant exists there is code in kiwi which restores that .rpmnew variant to become the real file.

However that _restore_intermediate_config_rpmnew_variants() method runs after overlay files has been applied to the system because it's part of the final cleanup step. In order to preserve an eventual overlay version of the file the .rpmnew variant gets only restored if the real file does not exist.

This Fixes #807

### DIFF
--- a/kiwi/system/root_bind.py
+++ b/kiwi/system/root_bind.py
@@ -217,7 +217,10 @@ class RootBind(object):
         """
         for config in self.config_files:
             config_rpm_new = self.root_dir + config + '.rpmnew'
-            if os.path.exists(config_rpm_new):
+            if (
+                os.path.exists(config_rpm_new) and
+                not os.path.exists(self.root_dir + config)
+            ):
                 shutil.move(config_rpm_new, self.root_dir + config)
 
     def _cleanup_mount_stack(self):
--- a/test/unit/system_root_bind_test.py
+++ b/test/unit/system_root_bind_test.py
@@ -121,8 +121,13 @@ class TestRootBind(object):
         self, mock_move, mock_exists, mock_islink, mock_remove_hierarchy,
         mock_command, mock_is_mounted
     ):
+        os_exists_return_values = [False, True]
+
+        def exists_side_effect(*args):
+            return os_exists_return_values.pop()
+
         mock_is_mounted.return_value = False
-        mock_exists.return_value = True
+        mock_exists.side_effect = exists_side_effect
         mock_islink.return_value = True
         self.bind_root.cleanup()
         self.mount_manager.umount_lazy.assert_called_once_with()


rjschwei:
@schaefi This does not fix the problem we face with GCE images where we have

`echo '169.254.169.254 metadata.google.internal metadata.google.internal' >> /etc/hosts
`

in config.sh . If I understand the code and your explanation correctly the changes from config.sh really end up in the .kiwi temporary file due to the symlink and that gets cleaned up in any event.

I think the underlying question we face is more general than the fix applied with this commit. How do we detect changes to automatically generated and/or temporary files and then merge those changes properly. The other question I guess is, does kiwi need to solve this problem at all? It would certainly be nice if kiwi did, but there are other ways to skin that cat. If we were to decide that kiwi shold not solve this problem then it becomes a documentation item.

Further complicating the problem for kiwi is that we'd have to know all those special files, but then again, maybe there are only a few. Anyway, based on the current config for our GCE images we need merging. Worse we would want it in perl-kiwi as well.

Another complication is that file copied in from the build host could be different from the auto generated file. So in order to handle the change via config.sh, as we want it for our GCE image setup, kiwi would have to make a backup copy of the file that is copied into the image from the build host. Then before cleanup the backup copy and the to be deleted copy would have to be compared. Any differences found would need to be merged into the file that showed up as .rpmnew and becomes the "real" file in the image. It is not overly complicated but there is a bunch of overhead.

schaefi:
Yes this is correct if you modify /etc/hosts as part of a config.sh script the issue still persists. Actually this one here is resolving the problem if somebody provides etc/hosts as an overlay file. So if you would consider to provide a static version as overlay file it would fix the issue in all builders because the legacy kiwi has fixed that issue already. In the scope of etc/hosts I think we should be able to provide a static hosts file which serves the purpose in the cloud.

I think the merge scenario you describe here would be tough to solve and I'm also not sure if that's what we want to support.

A nasty but working workaround would be if you modify the config.sh script to adapt the .rpmnew variant of etc/hosts:

```
echo '169.254.169.254 metadata.google.internal metadata.google.internal' >> /etc/hosts.rpmnew
```

This would modify the hosts file as it comes from the package and the restore code in kiwi would then make this file the final version.

However that only works if /etc/hosts is package managed, which is the case for current distributions but not for older ones. etc/hosts is provided by netcfg

In any case if this is something we need to discuss more detailed we need a new issue

Thanks

PR #818, opened 2018-09-19
Fixed disk detection for live iso in loopback grub

schaefi:
If the live iso is used as a disk on e.g USB sticks and
is configured to setup a persistent write space via the
initrd option rd.live.overlay.persistent, it's required
to find the disk device to create a write partition on
it.

This detection worked if the live iso is binary dumped
on the disk and bootet due to the hybrid support each
iso built by kiwi provides.

However if the live ISO deployment should not destroy
existing data on the disk there is an alternative setup
which uses grub's loopback support and puts the iso as a
file on some partition of the disk. In this mode the
kiwi-live dracut module failed to find the disk device
and could not setup a persistent write partition.


schaefi:
@davidcassany there is a mail thread "Enable ISO image to be in any partition..." on the mailing list for further details.

PR #820, opened 2018-09-24
Include PDF build of documentation to the package

### LINKED ISSUE #819: Bundle documentation as PDF in the kiwi package
## Problem description

Documentation for kiwi is available online but not bundled with the package

schaefi:
Bundle a PDF version of the online documentation with the rpm package
build. Due to the complexity of getting a latex build environment into
the travis CI which does not take forever to install, the bundling of
a built PDF into the pypi archive has been skipped. Users installing
from pypi would need to install a latex env on their machine and
run make latexpdf from the installed bundle.

This Fixes #819

### DIFF
--- a/doc/source/development.rst
+++ b/doc/source/development.rst
@@ -45,7 +45,7 @@ for details). For Python 2.7 use :command:`virtualenv`, which is provided
 via pip or as an extra package in your favourite Linux distribution.
 
 However, for setting up a Python virtual development environment the
-following additional latex, include, header files and compilers are required
+following additional LaTeX, include, header files and compilers are required
 in order to allow for compiling the C parts of the runtime required
 Python modules:
 
@@ -63,8 +63,10 @@ Python modules:
 
 .. code:: bash
 
-    $ zypper in python3-devel libxml2-devel libxslt-devel libffi48-devel glibc-devel gcc
-    $ zypper in texlive-fncychap texlive-wrapfig texlive-capt-of trang
+    $ zypper install \
+        python3-devel libxml2-devel libxslt-devel libffi48-devel \
+        glibc-devel gcc texlive-fncychap texlive-wrapfig \
+        texlive-capt-of trang
 
 Once the basic python module requirements are installed on your system,
 the next step is to create the virtual development environment.
@@ -89,7 +91,7 @@ development environment:
 
     $ pip install -r .virtualenv.dev-requirements.txt
 
-4. Install KIWI in "development mode":
+4. Install KIWI in Development Mode:
 
    .. code:: bash
 
@@ -313,7 +315,7 @@ order to build the documentation just call:
 
     tox -e doc
 
-Whenever a change in the documentation is pushed to the git, it will be
+Whenever a change in the documentation is pushed to GitHub, it will be
 automatically updated via :command:`travis-sphinx` and is available at:
 
 https://opensource.suse.com/kiwi
--- a/doc/source/development.rst
+++ b/doc/source/development.rst
@@ -85,7 +85,7 @@ development environment:
 
    .. code:: bash
 
-    $ pip3.4 install -r .virtualenv.dev-requirements.txt
+    $ pip install -r .virtualenv.dev-requirements.txt
 
 4. Install KIWI in "development mode":
 
@@ -151,11 +151,11 @@ If you want to see the target, use the option `-l` to print a list:
     $ tox -l
 
 To only run a special target, use the `-e` option. The following
-example runs the test cases for the 3.4 interpreter only:
+example runs the test cases for the 3.6 interpreter only:
 
 .. code:: bash
 
-    $ tox -e 3.4
+    $ tox -e 3.6
 
 Create a branch for each feature or bugfix
 ------------------------------------------
--- a/doc/source/development.rst
+++ b/doc/source/development.rst
@@ -45,7 +45,7 @@ for details). For Python 2.7 use :command:`virtualenv`, which is provided
 via pip or as an extra package in your favourite Linux distribution.
 
 However, for setting up a Python virtual development environment the
-following additional include, header files and compilers are required
+following additional latex, include, header files and compilers are required
 in order to allow for compiling the C parts of the runtime required
 Python modules:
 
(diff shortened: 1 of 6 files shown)


schaefi:
Uhm, to solve this travis CI problem we would need to add

```
sudo apt-get install texlive-full
```

but this will slow down CI testing performance a lot, LaTeX is huge...

We need a better/different solution

schaefi:
@tomschr If you know about a better latex setup in the sphinx conf.py file it would be great if you can share your thoughts. Other than that the pdf document is ok but due to the fact that it's build from the structure of an online documentation it surely has some design issues. For the moment this is ok for be because we don't have a versioned documentation which we ship with the package and that should be changed. Making it nice would be the next step, but if you know things we can change now I'm happy to add them. Thanks

schaefi:
@Conan-Kudo would also be great if you can have a look, but I think it should not negatively influence your packaging. Thanks

Conan-Kudo:
Have you considered just changing the Travis config to use containers to give you a useful environment?

In rpmlint, I use a configuration that lets me build and test rpmlint on Fedora and openSUSE: https://github.com/rpm-software-management/rpmlint/blob/master/.travis.yml

That way you can also validate that it works on Fedora and openSUSE in CI too. :)

doc/source/development.rst
```diff
@@ -63,6 +64,7 @@ Python modules:
 .. code:: bash
 
     $ zypper in python3-devel libxml2-devel libxslt-devel libffi48-devel glibc-devel gcc
+    $ zypper in texlive-fncychap texlive-wrapfig texlive-capt-of trang
```
tomschr: No errors per se, but two ideas to think about:

1. To reduce installing useless doc packages, add the `--no-recommends` option?
1. Although I'm a big fan of aliases (like `in`), for clarity reasons in documentation I would prefer `install` (would also be consistent with `pip install`).

tox.ini
```diff
@@ -100,14 +101,23 @@ commands =
         --cov-report=term-missing --cov-fail-under=100 --cov-config .coveragerc
 
 
+# Documentation build run, PDF target
+[testenv:pdfdoc]
+skip_install = True
+usedevelop = True
+deps = {[testenv]deps}
```
tomschr: I have two comments, not about this line, but in general about building the PDF documentation:

1. I had to install the `sphinxcontrib-spelling` module, otherwise it fails. I guess this needs to be added into `.virtualenv.dev-requirements.txt`?
1. Another alternative idea
   Would a separate requirement file under `kiwi/doc` be useful? Use case could be if someone is only interested in the documentation. With a separate file would that be possible. You can still refer to the "doc requirements" from  `.virtualenv.dev-requirements.txt` for a complete build.

doc/source/development.rst
```diff
@@ -45,14 +45,15 @@ for details). For Python 2.7 use :command:`virtualenv`, which is provided
 via pip or as an extra package in your favourite Linux distribution.
 
 However, for setting up a Python virtual development environment the
-following additional include, header files and compilers are required
+following additional latex, include, header files and compilers are required
```
tomschr: You are probably referring to the project (LaTeX) instead of the command (latex), right? If yes, replace latex -> LaTeX. :wink:

doc/source/development.rst
```diff
@@ -85,7 +87,7 @@ development environment:
 
    .. code:: bash
 
-    $ pip3.4 install -r .virtualenv.dev-requirements.txt
+    $ pip install -r .virtualenv.dev-requirements.txt
 
 4. Install KIWI in "development mode":
```
tomschr: Capitalize: **D**evelopment **M**ode

doc/source/development.rst
```diff
@@ -314,4 +316,4 @@ order to build the documentation just call:
 Whenever a change in the documentation is pushed to the git, it will be
```
tomschr: * Remove ~the~
* you usually push to GitHub, Bitbucket, GitLab, or any other (remote) repository and not "git" itself. So better use "... to GitHub".

tox.ini
```diff
@@ -30,12 +31,12 @@ whitelist_externals =
     /usr/bin/shellcheck
     /bin/bash
 basepython =
-    {check,doc,doc_travis,doc_travis_deploy}: python3
+    {check,pdfdoc,doc,doc_travis,doc_travis_deploy}: python3
     unit_py3_6: python3.6
     unit_py3_4: python3.4
```
tomschr: Just curiosity: No Python 3.5?

schaefi:
>  Well, depends on how you define "better". "Better" in what sense? ￼

@tomschr: yes sorry I was not very clear on that. I just meant from your experience about standards for manuals in terms of e.g font-size, pagestyle, etc... I just used the defaults. But the links you provided were helpful. I have added a few preamble settings which I think produces better readable results. Thanks

tox.ini
schaefi: sphinxcontrib-spelling is listed in .virtualenv.dev-requirements.txt. That should be part of your tox env if you have set it up using this documentation ;)

Separation of the dev-requirements is currently not an option because I don't think we gain much value. The setup speed of a development env for kiwi is currently not really relevant. The components download and compiling is done in under ~2min. However if I separate them I need to change the tox env for building the docs or running the tool which I think is not handy

tox.ini
schaefi: yes it seems it never was part of the distro we packaged it for. There was 3.4 and next I saw 3.6
If this is an issue we should resolve please open an extra issue for it such that we don't mix things here. Thanks

davidcassany:
>  In our case the add-on stuff would be the LaTeX distribution which would end up in an installation of many additional packages. So the original problem would be the same, or am I mistaken ?

Well this is the point, I guess we would require an image with all latex stuff in it. I doubt we'll find one out of the box, thus I guess we'd need to create it and maintain it in some public registry.

> The idea sounds great, let's preserve the discussion in an issue. I'll open one and put the relevant parts to it.

Yes I think this could be useful in order to get a controlled environment. Currently travis env time to time causes issues and it is hard to handle them due to not having full control over the env.

schaefi:
I have opened #822 for further discussions on ideas which came up here. Will continue merging the outcome of this one now. Thanks everybody

PR #821, opened 2018-09-25
Make volume id customizable for installation ISOs

### LINKED ISSUE #811: Allow user defined installation media volume ID
## Problem description

With former KIWI `volid` attribute could also be used for `oem` images having the `installiso` attribute. Current KIWI does not permit the use of `volid` attribute for `oem` images, moreover, the volume ID is fixed to `INSTALL` and the `kiwi-dump` dracut module is dependent to that value.

## Expected behaviour
The section as: 
`<type image="oem" installiso="true" volid="_LABEL_" ...>`
Results in a installation media labeled with the defined label

Build an OEM image with the installation media. The installation media will be always labeled as `INSTALL`.

+ KIWI version: 9.16.12

davidcassany:
This commit makes volid attribute also available for OEM images.
The installation media makes use of the volid value. Only posix
safe names are allowed, up to 32 characters.

Fixes #811


schaefi:
@davidcassany not reviewed yet but encountered a funny typo in the commit message: installaion

davidcassany:
@schaefi, oh right, gonna amend the commit message now

PR #823, opened 2018-09-26
Create parent qgroup when snapper is present

davidcassany:
This commit creates a new parent quota group (1/0) of level 1 when
btrfs_quota_groups is enabled and snapper present into the image
root tree.

Related to bsc#1093518 and #812

### DIFF
--- a/kiwi/volume_manager/btrfs.py
+++ b/kiwi/volume_manager/btrfs.py
@@ -312,7 +312,7 @@ class VolumeManagerBtrfs(VolumeManagerBase):
             )
             if self.custom_args['quota_groups'] and \
                     self.custom_args['root_is_snapshot']:
-                self._create_quota_group_info()
+                self._create_snapper_quota_configuration()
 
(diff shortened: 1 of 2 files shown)


schaefi:
That simplifies the code which is nice

kiwi/volume_manager/btrfs.py
```diff
@@ -358,18 +358,22 @@ def _xml_pretty(self, toplevel_element):
         xml_data_domtree = minidom.parseString(xml_data_unformatted)
         return xml_data_domtree.toprettyxml(indent="    ")
 
-    def _create_quota_group_info(self):
+    def _create_snapper_configuration(self):
```
Vogtinator: Misleading function name?

davidcassany: what about `_create_snapper_quota_configuration`? I want to add snapper into the name because all the changes done here are snapper specific.

PR #831, opened 2018-10-05
Fixed URI handling with token query option

### LINKED ISSUE #828: Crashes with repo URLs which have query keys without values
Adding a URI such as `http://foo.bar/baz?asdf` crashes kiwi:

```
[ ERROR   ]: 22:54:23 | Unexpected error:
  File "/usr/bin/kiwi", line 9, in <module>
    load_entry_point('kiwi==9.16.6', 'console_scripts', 'kiwi-ng-2')()
  File "/usr/lib/python2.7/site-packages/kiwi/kiwi.py", line 64, in main
    App()
  File "/usr/lib/python2.7/site-packages/kiwi/app.py", line 35, in __init__
    task_class().process()
  File "/usr/lib/python2.7/site-packages/kiwi/tasks/system_build.py", line 201, in process
    self.command_args['--signing-key']
  File "/usr/lib/python2.7/site-packages/kiwi/system/prepare.py", line 149, in setup_repositories
    repo_user, repo_secret, uri.credentials_file_name(),
  File "/usr/lib/python2.7/site-packages/kiwi/system/uri.py", line 144, in credentials_file_name
    query = dict(params.split('=') for params in uri.query.split('&'))
ValueError: dictionary update sequence element #0 has length 1; 2 is required```

### LINKED ISSUE #830: Kiwi strips the query of repo URLs
When invoking kiwi with `--add-repo https://foo.bar/baz?asdf=qwertz`, only `https://foo.bar/baz` is added to the repos used by zypper.

This means it's impossible to use KIWI to build SLE images as the query contains the credentials.

### PR DESCRIPTION
**Pull request by schaefi**
So far only the query format ?credentials=... was supported.
In case of ?random_token_data the returned uri was truncated
and also the format check on the query caused a python trace.
This Fixes #830 and Fixes #828

### DIFF
--- a/kiwi/system/uri.py
+++ b/kiwi/system/uri.py
@@ -120,7 +120,14 @@ class Uri(object):
         elif uri.scheme == 'iso':
             return self._iso_mount_path(uri.path)
         elif uri.scheme.startswith('http') or uri.scheme == 'ftp':
-            return ''.join([uri.scheme, '://', uri.netloc, uri.path])
+            if self._has_credentials_query() or not uri.query:
+                return ''.join(
+                    [uri.scheme, '://', uri.netloc, uri.path]
+                )
+            else:
+                return ''.join(
+                    [uri.scheme, '://', uri.netloc, uri.path, '?', uri.query]
+                )
         else:
             raise KiwiUriStyleUnknown(
                 'URI schema %s not supported' % self.uri
@@ -134,13 +141,13 @@ class Uri(object):
 
         :rtype: str
         """
-        uri = urlparse(self.uri)
+        uri = self._has_credentials_query()
         # initialize query with default credentials file name.
         # The information will be overwritten if the uri contains
         # a parameter query with a credentials parameter
         query = {'credentials': 'kiwiRepoCredentials'}
 
-        if uri.query:
+        if uri:
             query = dict(params.split('=') for params in uri.query.split('&'))
 
         return query['credentials']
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: kiwi/system/uri.py [hunk 1]
```diff
@@ -134,13 +141,13 @@ def credentials_file_name(self):
 
         :rtype: str
         """
-        uri = urlparse(self.uri)
+        uri = self._has_credentials_query()
```
davidcassany: nitpick, but I think this variable name is confusing

kiwi/system/uri.py
```diff
@@ -134,13 +141,13 @@ def credentials_file_name(self):
...
         # initialize query with default credentials file name.
         # The information will be overwritten if the uri contains
         # a parameter query with a credentials parameter
         query = {'credentials': 'kiwiRepoCredentials'}
 
-        if uri.query:
+        if uri:
```
davidcassany: In fact here is where I got confused while reading the code, had to back to reread uri's assignment.

kiwi/system/uri.py
davidcassany: What about `credentials` instead of `uri`?

schaefi: ok, yes this is a confusing name, I'll change it

kiwi/system/uri.py
schaefi: this should be clear now with the method name change ?

davidcassany: :+1:

PR #832, opened 2018-10-07
Include livenet module with dmsquash-live support

### LINKED ISSUE #827: SquashFS as rootfs over HTTP needs livenet support
## Problem description
I'm trying to build an openSUSE JeOS image for PXE boot. The environment requires layer-3 file transfer so AoE is not possible. The dracut manpage said it supports mounting squashfs via HTTP, but it seems impossible with a kiwi-built image.

I used the example openSUSE JeOS 15.0 config to generate an ISO. Then I extracted the ISO and got the kernel, initrd and the rootfs. After that I used the following script to boot via iPXE.

```shell
#!ipxe

kernel http://server/linux rd.live.debug=1 rd.live.image rd.shell root=live:http://server/squashfs.img
initrd http://server/initrd
boot
```

The kernel and the initrd works, but when it is mounting the rootfs, the dmesg says:
```
loop: module loaded
dracut: FATAL: Don't know how to handle 'root=live:http://server/squashfs.img'
dracut: Refusing to continue
```

Then the kernel shuts down.

I've tried other package types, like `squashfs` and `oem`, they seems to have this problem too. 

+ KIWI version: KIWI (next generation) version 9.15.3
+ Operating system: openSUSE Leap 15.0
+ OBS version: N/A

schaefi:
The upstream dracut dmsquash-live module supports network
mode with the livenet module. But that module must be
explicitly included and is not fetched automatically.
This Fixes #827

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -915,7 +915,7 @@ class Defaults(object):
         """
         live_modules = {
             'overlay': 'kiwi-live',
-            'dmsquash': 'dmsquash-live'
+            'dmsquash': 'dmsquash-live livenet'
         }
         if flag_name in live_modules:
             return live_modules[flag_name]
--- a/test/unit/defaults_test.py
+++ b/test/unit/defaults_test.py
@@ -78,7 +78,7 @@ class TestDefaults(object):
         assert Defaults.get_live_dracut_module_from_flag('overlay') == \
             'kiwi-live'
         assert Defaults.get_live_dracut_module_from_flag('dmsquash') == \
-            'dmsquash-live'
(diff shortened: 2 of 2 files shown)


schaefi:
Please note this makes the curl program a requirement of the image. If curl is not installed the dracut call will fail and with it the image build process, but the dracut error message is pretty clear. In this case add

```xml
<package name="curl"/>
```

to your image description

schaefi:
@Conan-Kudo man you are fast :)

Conan-Kudo:
@schaefi I saw the change coming in and looked over it already. :)

PR #833, opened 2018-10-08
rebuild auto generated code with stable generateDS

schaefi:
For some reason the xml_parse code generated by generateDS v2.29.24
caused warnings on simple type XSD patterns. Therefore I rebuild
the code with the stable build version v2.29.14 which fixed that
issue

### DIFF
--- a/kiwi/xml_parse.py
+++ b/kiwi/xml_parse.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 #
-# Generated  by generateDS.py version 2.29.24.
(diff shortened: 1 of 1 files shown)


schaefi:
@davidcassany there is nothing to review because only auto generated code was recreated. See the report on the mailing list ```XSD validation warning from KIWI version 9.16.19``` for details. We should be careful when using the latest generateDS code

schaefi:
Can you double check the change is ok ? Thanks

schaefi:
no problem, we are fast in fixing bugs :) Thanks for double checking

PR #835, opened 2018-10-09
Added gfxboot to required packages in spec

schaefi:
kiwi calls gfxboot as tool when building live iso images.
Thus this tool provided by the gfxboot package should be
a requirement for kiwi

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -106,6 +106,7 @@ Requires:       grub2-x86_64-efi
 Requires:       qemu-tools
 Requires:       squashfs
 Requires:       gptfdisk
+Requires:       gfxboot
 %endif
 %if 0%{?fedora} || 0%{?rhel}
 Requires:         chkconfig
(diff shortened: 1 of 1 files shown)


schaefi:
I found gfxboot on the distros we support in that name and so I added it without a distribution flag. Please double check if that is ok. Thanks

schaefi:
I could find them for all on https://pkgs.org/download/gfxboot, but this has no information if the distro itself has it. Maybe @Conan-Kudo can tell. I think we have to distinguish them, will adapt the pr

Conan-Kudo:
`gfxboot` is unavailable for Fedora/CentOS at all currently. I don't know if `gfxboot` works for syslinux 6.x or not, either, since openSUSE never upgraded to that version ([from what I can tell](https://build.opensuse.org/package/show/system:install:head/syslinux)), and openSUSE is the upstream for `gfxboot`. In addition, `gfxboot` doesn't work with GRUB 2, either.

schaefi:
ok, thanks I'll adapt

schaefi:
@Conan-Kudo are you ok with that change ? I think it will not change anything on the rhel/fedora side. Thanks

PR #840, opened 2018-10-11
Fixed package specs for dracut modules on debian

### LINKED ISSUE #837: wrong dependencies of kiwi dracut modules for Ubuntu
## Problem description

`btrfs-progs` package does not exist in Ubuntu, thus `dracut-kiwi-oem-repart` package should not require it for Ubuntu/Debian builds. Instead there is a `btrfs-tools` package available.

Tested it for Xenial and Bionic

In addition `dracut-kiwi-live` and `dracut-kiwi-lib` should require `dracut-network`. Ubuntu has an extra package for dracut network modules. The network module is needed by `dracut-kiwi-live` and `dracut-kiwi-lib`, if this package is not present the dracut call to create the intird fails due to missing dependencies.

## Expected behaviour

`dracut-kiwi-oem-repart` is installable in Ubuntu and `dracut-kiwi-live` pulls `dracut-network`

Include `dracut-kiwi-oem-repart` or `dracut-kiwi-live` in an Ubuntu build.

schaefi:
btrfs tools are provided by btrfs-tools package and not by
btrfs-progs. In addition dracut-network is a separate package
on debian/ubuntu based distributions. This Fixes #837

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -273,7 +273,7 @@ BuildRequires:  dracut
 %endif
 Requires:       bc
 Requires:       cryptsetup
-%if 0%{?fedora} || 0%{?rhel} || 0%{?debian} || 0%{?ubuntu}
+%if 0%{?fedora} || 0%{?rhel}
 Requires:       btrfs-progs
 Requires:       gdisk
 %else
@@ -294,6 +294,9 @@ Requires:       curl
 %if 0%{?debian} || 0%{?ubuntu}
 Requires:       xz-utils
 Requires:       dmsetup
+Requires:       btrfs-tools
+Requires:       gdisk
+Requires:       dracut-network
 %else
 Requires:       xz
 Requires:       device-mapper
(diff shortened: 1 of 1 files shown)


package/python-kiwi-spec-template
```diff
@@ -294,6 +294,9 @@ Requires:       curl
 %if 0%{?debian} || 0%{?ubuntu}
 Requires:       xz-utils
 Requires:       dmsetup
+Requires:       btrfs-tools
+Requires:       gdisk
+Requires:       dracut-network
```
Conan-Kudo: `dracut-network` is also separate in Fedora/CentOS, since the upstream dracut spec file recommends it that way.

schaefi: oh, ok yes we should have a Fedora iso test build as well, the existing virtual disk image build does not show this. I'll adapt the spec change

Conan-Kudo:
@davidcassany It does not look like Tumbleweed has the subpackage: https://build.opensuse.org/package/binaries/openSUSE:Factory/dracut/standard

PR #842, opened 2018-10-11
Add alternative search path for syslinux binaries

### LINKED ISSUE #839: Isolinux.bin binary not found
## Problem description

Building Ubuntu ISO's there is a `KiwiIsoLoaderError: No isolinux loader %s found` failure. This is caused because the `isolinux.bin` binary is not found in root tree and, in consequence, not copied to the expected location, see https://github.com/SUSE/kiwi/blob/e9f715d82b446f069ec3788b475ead45ddd29ba4/kiwi/bootloader/config/isolinux.py#L243-L271

Ubuntu has `isolinux.bin` located at `/usr/lib/ISOLINUX/isolinux.bin`, thus it does not get copied to the  appropriate `/boot` location.

## Expected behaviour

`isolinux.bin` is found in `/boot` in the image root tree, so the ISO image creation does not fail.

Try to build an Ubuntu ISO.

schaefi:
On debian/ubuntu systems /usr/lib/ISOLINUX is a valid
search path for syslinux binaries like isolinux.bin.
This Fixes #839

### DIFF
--- a/kiwi/bootloader/config/isolinux.py
+++ b/kiwi/bootloader/config/isolinux.py
@@ -260,7 +260,8 @@ class BootLoaderConfigIsoLinux(BootLoaderConfigBase):
         ]
         syslinux_dirs = [
             '/usr/share/syslinux/',
-            '/usr/lib/syslinux/modules/bios/'
+            '/usr/lib/syslinux/modules/bios/',
+            '/usr/lib/ISOLINUX/'
         ]
         for syslinux_file_name in syslinux_file_names:
             for syslinux_dir in syslinux_dirs:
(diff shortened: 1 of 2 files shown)


schaefi:
We should create an ubuntu integration test build once all those patches are available ;)

schaefi:
I have setup Virtualization:Appliances:Images:Testing_x86/test-image-ubuntu-vmx and try to make it work. I did not stumble over another python-kiwi but probably I'm not yet at that point. It looks like you already played a bit with it, feel free to fix/change the build as you see fit :)

Conan-Kudo:
@davidcassany This is why I left our package as called `python2-kiwi`. :)

schaefi:
*argh* now I see it

PR #844, opened 2018-10-12
Fix use of uri type in debootstrap call

schaefi:
Referencing a file in a debootstrap call is done using
the file:/ source type. However when using file:// debootstrap
does something different and failed to find the file. The
additional / had a bad impact to the call.

### DIFF
--- a/kiwi/repository/apt.py
+++ b/kiwi/repository/apt.py
@@ -146,6 +146,7 @@ class RepositoryApt(RepositoryBase):
         if os.path.exists(uri):
             # apt-get requires local paths to take the file: type
             uri = 'file:/' + uri
+            uri = uri.replace('file://', 'file:/')
         if not components:
             components = 'main'
         with open(list_file, 'w') as repo:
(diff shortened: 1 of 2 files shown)


schaefi:
@davidcassany I have talked with the buildservice people about the build failure in:

* https://build.opensuse.org/package/live_build_log/Virtualization:Appliances:Images:Testing_x86/test-image-ubuntu-vmx/images/x86_64

```
[  271s] [ DEBUG   ]: 09:00:42 | EXEC: [debootstrap --no-check-gpg --variant=minbase xenial /usr/src/packages/KIWI-vmx/build/image-root.debootstrap file://usr/src/packages/SOURCES/repos/Ubuntu:16.04/standard]
[  271s] [ DEBUG   ]: 09:00:42 | EXEC: Failed with stderr: (no output on stderr), stdout: I: Retrieving InRelease 
[  271s] I: Failed to retrieve InRelease
[  271s] I: Retrieving Release 
[  271s] E: Failed getting release file 
```

Michael Schröder was so kind to fix the build script to add the presence of the release file.
But even after that fix debootstrap failed because of the ```file://``` source locator.

This part should be addressed with this patch here

schaefi:
I think with this fix we will be able to build ubuntu images in the buildservice too :)

PR #845, opened 2018-10-15
Added machine id setup in dracut preparation

### LINKED ISSUE #843: /etc/machine-id preset in image
## Problem description  
The file /etc/machine-id is preset in readily built images. This results in all deployments of images, the machine-id is identical although this should be a unique setting.

## Expected behaviour
Just add an empty file /etc/machine-id to the images. This makes systemd generating a new id during first startup.

Build an image and deploy it several times. The content of /etc/machine-id should be different.

+ KIWI version: 9.16.?
+ Operating system: all with systemd
+ OBS version: -

schaefi:
In case of a dracut booted image we empty out the systemd
machine-id configuration file to trigger the rebuild of that
information by the dracut boot code at boot time. This allows
for unique systemd identifiers if the same image gets deployed
on different machines. This also makes the script implementations
people put in in config.sh or images.sh to solve this problem
obsolete.

This Fixes #843

### DIFF
--- a/kiwi/boot/image/dracut.py
+++ b/kiwi/boot/image/dracut.py
@@ -55,7 +55,10 @@ class BootImageDracut(BootImageBase):
 
     def prepare(self):
         """
-        Prepare kiwi profile environment to be included in dracut initrd
+        Prepare dracut caller environment
+
+        * Create kiwi .profile environment to be included in dracut initrd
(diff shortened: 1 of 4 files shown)


schaefi:
Set on blocked as it might have unexpected after effects. See #843 for details

PR #850, opened 2018-10-17
Don't cache packages from local repos for zypper

### LINKED ISSUE #847: Kiwi-ng should not cache local Repositories
## Problem description
Unlike Kiwi 7.x, Kiwi-ng caches packages from local repos.

## Expected behaviour
Repositories begining with file: should not be cached. 

Build an appliance with at least one local Repository and check 
/var/cache/kiwi/packages/*/

+ KIWI version: 9.16.12
+ Operating system: openSuse 42.3

schaefi:
Access to packages from local repositories is as fast as reading
them from a cache location. The additional package copy and cache
update is superfluous and should be avoided. This Fixes #847

### DIFF
--- a/kiwi/repository/zypper.py
+++ b/kiwi/repository/zypper.py
@@ -22,6 +22,7 @@ from tempfile import NamedTemporaryFile
 # project
 from kiwi.command import Command
 from kiwi.repository.base import RepositoryBase
+from kiwi.system.uri import Uri
 from kiwi.path import Path
 
 from kiwi.exceptions import (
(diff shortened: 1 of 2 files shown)


schaefi:
I did a short research on dnf/yum and apt-get to see if there is a way to influence the cache setup for data from local repositories. So far I could not find a configuration option which should be set in the repo files for those package managers and I think they handle caching differently between local and remote repos. However I'm not sure about it, thus any feedback with regards to the other package managers is welcome.

Thanks

Conan-Kudo:
`--setopt=keepcache=0` for DNF and Yum.

Conan-Kudo:
@davidcassany I would suggest making [an RFE for DNF](https://github.com/rpm-software-management/dnf/wiki/Bug-Reporting#new-feature-request) to support a per-repo parameter for this.

PR #854, opened 2018-10-18
Allow simple path source in Uri class

schaefi:
This patch is needed as follow up fix for the setup of the
package cache in local repositories. The is_remote method
from the Uri class is used to identify if a repostory source
is remote or local. At that point the initial repository
source was already translated into its components. In case
of a local repository the Uri instance now receives a simple
path and the is_remote method raised with a style error.
This patch allows the Uri class to be more friendly and
initializes a local path as file:/ typed source.
Related to Issue #847

### DIFF
--- a/kiwi/system/uri.py
+++ b/kiwi/system/uri.py
@@ -51,7 +51,9 @@ class Uri(object):
     def __init__(self, uri, repo_type=None):
         self.runtime_config = RuntimeConfig()
         self.repo_type = repo_type
-        self.uri = uri if not uri.startswith(os.sep) else 'file:/' + uri
+        self.uri = uri if not uri.startswith(os.sep) else ''.join(
+            [Defaults.get_default_uri_type(), uri]
+        )
         self.mount_stack = []
 
         self.remote_uri_types = {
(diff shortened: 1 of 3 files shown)


schaefi:
@davidcassany I overlooked the impact of the change on obs builds. See all test builds failing due to the local vs. remote repo cache setup :/ Glad we have the Staging area

davidcassany:
Yes +1 for the tests and staging area.

kiwi/system/uri.py
```diff
@@ -51,7 +51,7 @@ class Uri(object):
     def __init__(self, uri, repo_type=None):
         self.runtime_config = RuntimeConfig()
         self.repo_type = repo_type
-        self.uri = uri
+        self.uri = uri if not uri.startswith(os.sep) else 'file:/' + uri
```
davidcassany: we have 'dir:/' and 'iso:/' also as a valid local schemas. As far as I remember we introduced 'file:/' schema usage while coding the `derived_from` attribute for containers, where using `dir:/<path>` for tarball was too just too confusing.

Does it means we can now add uris without schema in the description file as a consequence of this change? I am fine with the change if it is clear that using plain paths in the description file is prohibited or without unexpected side effects.

schaefi: >Does it means we can now add uris without schema in the description file as a consequence of this change?

yes it would be the consequence. But you have a point here, the default type for plain paths should be dir:/ not file:/ I'll change that

I could not see a negative side effect, dir:/ and file:/ translates both into _local_path() so that should be safe

PR #856, opened 2018-10-18
Add a hardcoded rpm database path to import trusted keys

### LINKED ISSUE #855: gpg keys of the repositories are not trusted even using '--signing-key' flag
## Problem description

Since openSUSE Leap 15 `--signing-key <keyfile>` flag does not prevent `zypper` from failing when gpg checks are enabled, even the provided keyfile is verified to be valid.

This is an issue also reported to libzypp as a bugzilla issue [bsc#1112357](https://bugzilla.opensuse.org/show_bug.cgi?id=1112357)

## Expected behaviour

Adding `--siging-key <keyfile>` with correct keys should be enough to ensure KIWI is capable to build images when 'repository_gpgcheck' and 'package_gpgcheck' attributes are set to `true`.

Enable `repository_gpgcheck` and/or `package_gpgcheck` for some repository and the installation of any package from that repo will cause a zypper install failure.

davidcassany:
This commits adds a hardcoded rpm database location to make
sure the imported keys are in the expected location for zypper.

Fixes #855

### DIFF
--- a/kiwi/repository/zypper.py
+++ b/kiwi/repository/zypper.py
@@ -24,6 +24,7 @@ from kiwi.command import Command
 from kiwi.repository.base import RepositoryBase
 from kiwi.system.uri import Uri
 from kiwi.path import Path
+from kiwi.defaults import Defaults
 
 from kiwi.exceptions import (
     KiwiRepoTypeUnknown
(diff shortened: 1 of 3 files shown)


kiwi/repository/zypper.py
```diff
@@ -253,7 +253,11 @@ def import_trusted_keys(self, signing_keys):
         :param list signing_keys: list of the key files to import
         """
         for key in signing_keys:
-            Command.run(['rpm', '--root', self.root_dir, '--import', key])
+            # Including --dbpath flag is a workaround for bsc#1112357
+            Command.run([
+                'rpm', '--root', self.root_dir, '--import',
+                key, '--dbpath', '/var/lib/rpm'
```
schaefi: Could we put the static path into a classmethod in the Defaults space ? I think we should maintain those type of static information all in one place.

PR #861, opened 2018-10-25
Cleanup code to make flake8 happy

schaefi:
The new version of flake8 is more strict on code checking and
complained at several places. This commit fixes the code smells
such that flake8 is happy again

### DIFF
--- a/kiwi/container/oci.py
+++ b/kiwi/container/oci.py
@@ -201,20 +201,20 @@ class ContainerImageOCI(object):
             Command.run(
                 ['umoci', 'config', '--image', container_name, '--tag', tag]
             )
+        umoci_config = \
+            self.maintainer + \
+            self.user + \
+            self.workingdir + \
+            self.entry_command + \
+            self.entry_subcommand + \
+            self.expose_ports + \
+            self.volumes + \
+            self.environment + \
+            self.labels
         Command.run(
             [
                 'umoci', 'config'
-            ] +
-            self.maintainer +
-            self.user +
-            self.workingdir +
-            self.entry_command +
-            self.entry_subcommand +
-            self.expose_ports +
-            self.volumes +
-            self.environment +
-            self.labels +
-            [
+            ] + umoci_config + [
                 '--image', container_name,
                 '--created', datetime.utcnow().strftime(
                     '%Y-%m-%dT%H:%M:%S+00:00'
--- a/kiwi/iso_tools/cdrtools.py
+++ b/kiwi/iso_tools/cdrtools.py
@@ -163,11 +163,14 @@ class IsoToolsCdrTools(IsoToolsBase):
                 hidden_files_parameters.append(hidden_file)
                 hidden_files_parameters.append('-hide-joliet')
                 hidden_files_parameters.append(hidden_file)
+        iso_parameters = \
+            hidden_files_parameters + \
+            self.iso_parameters + \
+            self.iso_loaders
         Command.run(
             [
                 self.get_tool_name()
-            ] + hidden_files_parameters +
-            self.iso_parameters + self.iso_loaders + [
+            ] + iso_parameters + [
                 '-o', filename, self.source_dir
             ]
         )
--- a/kiwi/iso_tools/iso.py
+++ b/kiwi/iso_tools/iso.py
@@ -154,15 +154,15 @@ class Iso(object):
                     new_volume_id = Iso._sub_string(
                         data=new_volume_descriptor, length=7
                     )
-                    if (bytes(b'TEA01') in new_volume_id or
-                            sector + 1 == ref_sector):
+                    if bytes(b'TEA01') in new_volume_id or \
+                       sector + 1 == ref_sector:
+
                         new_boot_catalog_sector = sector + 1
                         break
 
-            if (
-                new_boot_catalog_sector and
-                iso_metadata.boot_catalog_sector != new_boot_catalog_sector
-            ):
+            if new_boot_catalog_sector and \
+               iso_metadata.boot_catalog_sector != new_boot_catalog_sector:
+
                 new_boot_catalog = Iso._read_iso_sector(
                     new_boot_catalog_sector, iso
                 )
--- a/kiwi/package_manager/dnf.py
+++ b/kiwi/package_manager/dnf.py
@@ -184,8 +184,11 @@ class PackageManagerDnf(PackageManagerBase):
         else:
             chroot_dnf_args = self.root_bind.move_to_root(self.dnf_args)
             return Command.call(
-                ['chroot', self.root_dir, 'dnf'] + chroot_dnf_args +
-                self.custom_args + ['autoremove'] + delete_items,
+                [
+                    'chroot', self.root_dir, 'dnf'
+                ] + chroot_dnf_args + self.custom_args + [
+                    'autoremove'
+                ] + delete_items,
                 self.command_env
             )
 
--- a/kiwi/runtime_checker.py
+++ b/kiwi/runtime_checker.py
@@ -195,7 +195,6 @@ class RuntimeChecker(object):
             self._check_multitag_support()
 
     def _check_multitag_support(self):
-
         message = dedent('''\n
             Using additionaltags attribute requires skopeo tool to be
             capable to handle it, it must include the '--additional-tag'
@@ -203,15 +202,12 @@ class RuntimeChecker(object):
             --help').\n
             It is known to be present since v0.1.30
         ''')
-
-        if (
-            'additional_tags' in self.xml_state.get_container_config() and not
-            CommandCapabilities.has_option_in_help(
+        if 'additional_tags' in self.xml_state.get_container_config():
+            if not CommandCapabilities.has_option_in_help(
                 'skopeo', '--additional-tag', ['copy', '--help'],
                 raise_on_error=False
-            )
-        ):
-            raise KiwiRuntimeError(message)
+            ):
+                raise KiwiRuntimeError(message)
 
     def check_boot_description_exists(self):
         """
@@ -239,8 +235,8 @@ class RuntimeChecker(object):
         image_types_supporting_custom_boot_description = ['oem', 'pxe']
         build_type = self.xml_state.get_build_type_name()
         initrd_system = self.xml_state.get_initrd_system()
-        if (initrd_system == 'kiwi' and
-                build_type in image_types_supporting_custom_boot_description):
+        if initrd_system == 'kiwi' and \
+           build_type in image_types_supporting_custom_boot_description:
 
             boot_image_reference = self.xml_state.build_type.get_boot()
             if not boot_image_reference:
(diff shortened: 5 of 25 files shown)


kiwi/container/oci.py
```diff
@@ -204,17 +204,13 @@ def create(self, filename, base_image):
...
-            self.environment +
-            self.labels +
-            [
+            ] + list(
+                self.maintainer + self.user + self.workingdir
+            ) + list(
+                self.entry_command + self.entry_subcommand
+            ) + list(
+                self.expose_ports + self.volumes + self.environment
+            ) + self.labels + [
```
aplanas: is this OK?

kiwi/iso_tools/cdrtools.py
```diff
@@ -166,8 +166,9 @@ def create_iso(self, filename, hidden_files=None):
         Command.run(
             [
                 self.get_tool_name()
-            ] + hidden_files_parameters +
-            self.iso_parameters + self.iso_loaders + [
+            ] + list(
+                hidden_files_parameters + self.iso_parameters
+            ) + self.iso_loaders + [
```
aplanas: same

kiwi/iso_tools/iso.py
```diff
@@ -154,15 +154,15 @@ def relocate_boot_catalog(self, isofile):
                     new_volume_id = Iso._sub_string(
                         data=new_volume_descriptor, length=7
                     )
-                    if (bytes(b'TEA01') in new_volume_id or
-                            sector + 1 == ref_sector):
+                    if (
+                        bytes(
+                            b'TEA01'
+                        ) in new_volume_id or sector + 1 == ref_sector
+                    ):
```
aplanas: I am not sure about the original complain, but I guess that could be fixed moving a bit on the left the original line 158

kiwi/package_manager/dnf.py
```diff
@@ -184,8 +184,11 @@ def process_delete_requests(self, force=False):
         else:
             chroot_dnf_args = self.root_bind.move_to_root(self.dnf_args)
             return Command.call(
-                ['chroot', self.root_dir, 'dnf'] + chroot_dnf_args +
-                self.custom_args + ['autoremove'] + delete_items,
+                [
+                    'chroot', self.root_dir, 'dnf'
+                ] + chroot_dnf_args + self.custom_args + [
+                    'autoremove'
+                ] + delete_items,
```
aplanas: what kind of complain would fix this?

kiwi/runtime_checker.py
```diff
@@ -239,8 +239,10 @@ def check_boot_description_exists(self):
         image_types_supporting_custom_boot_description = ['oem', 'pxe']
         build_type = self.xml_state.get_build_type_name()
         initrd_system = self.xml_state.get_initrd_system()
-        if (initrd_system == 'kiwi' and
-                build_type in image_types_supporting_custom_boot_description):
+        if (
+            initrd_system == 'kiwi' and build_type in
+            image_types_supporting_custom_boot_description
+        ):
```
aplanas: Same question, in Python `if`s are not expected to be like this : )

kiwi/package_manager/dnf.py
schaefi: line break before/after operator

kiwi/system/root_import/oci.py
```diff
@@ -100,6 +100,6 @@ def __del__(self):
             Path.wipe(self.oci_layout_dir)
         if self.oci_unpack_dir:
             Path.wipe(self.oci_unpack_dir)
-        if (self.uncompressed_image is not None and
-                self.uncompressed_image != self.image_file):
-            Path.wipe(self.uncompressed_image)
+        if (self.uncompressed_image):
```
aplanas: no complain about the extra parens?

aplanas:
I see a lot of weird stuff here, may I suggest a more detailed analysis of each original complain?

kiwi/system/root_import/oci.py
schaefi: yeah that one was a leftover of the simplify of the condition, thanks

kiwi/runtime_checker.py
schaefi: I know but in the rare cases where if conditions are long the official pythonic way is to embed the expression into brackets. At least that's what people told me in the past. And yes I usually avoid the long conditions

kiwi/package_manager/dnf.py
aplanas: I am missing something, as I have the last version from git of each package for flake:

* flake8-3.6.0
* pyflakes-2.0.0
* pycodestyle-2.4.0
* mccabe-0.6.1

and no complains for this:

```
def xx():
    lx = (1 +
          2)
    print(lx)

    xx(['a'] +
       ['b'])
```

You can test it also using the last released version

```
python3 -m venv
source venv/bin/activate
pip install flake8
flake8 ./test.py
```

kiwi/runtime_checker.py
aplanas: The long line is a fair complain, but now there are usually two options to fix the complain (not counting the ones that will create new variables for subexpressions of the condition, as you did in some cases, that is OK).

* use parens and split, and indent one char to the left to avoid confusion (optional)

```python
def xx():
    initrd_system = ''
    build_type = ''
    image_types_supporting_custom_boot_description = []
    if True:
        if (initrd_system == 'kiwi' and
           build_type in image_types_supporting_custom_boot_description):
            pass
```

* no use parens but use backslash:

```python
def xx():
    initrd_system = ''
    build_type = ''
    image_types_supporting_custom_boot_description = []
    if True:
        if initrd_system == 'kiwi' and \
           build_type in image_types_supporting_custom_boot_description:
            pass
```

Both snippets have 0 complains in the same setup with the last version.

schaefi: if you use backslash you got the complain that using backslash is redundant. Sorry what's your point here ? Do you think I start touching this if flake8 would be happy ? This type of review to prove me wrong is weird. Do the following:

```
git checkout master
tox -e check
```

see all the complains. Now checkout the branch which fixes them

```
git checkout flake_cleanup
tox -e check
```

See all fixed and used for this PR.

Please stop constructing examples out of scope it's not helpful

Thanks

schaefi: btw, both of your suggestions caused flake complains. First version complains about line break after operator (and). Second version complains about backslash being redundant between brackets. As I said I touched this for a reason

aplanas: > if you use backslash you got the complain that using backslash is redundant. Sorry what's your point here ? 

My point is to help the project.

> Do you think I start touching this if flake8 would be happy ?

Actually yes. Mostly because I tested it before commenting.

I do not know how did you mess with your system. You must to have some random `.flake8.` in your `~` or `~/.config` that you are not providing into the project.

Please, before being so rude test this is a different machine:

```
git checkout master
python3 -m venv venv
source venv/bin/activate
pip install flake8
# This is a variant of what you have in tox.ini to check only one file
flake8 --statistics -j auto --count kiwi/runtime_checker.py
```

Output:

```
kiwi/runtime_checker.py:242:17: W504 line break after binary operator
kiwi/runtime_checker.py:431:17: W504 line break after binary operator
2     W504 line break after binary operator
2
```

Change the line 242 to exactly this, to avoid the first W504:

```python
        if initrd_system == 'kiwi' and \
           build_type in image_types_supporting_custom_boot_description:
```
Output:

```
flake8 --statistics -j auto --count kiwi/runtime_checker.py
kiwi/runtime_checker.py:431:17: W504 line break after binary operator
1     W504 line break after binary operator
1
```

The second case can be fixed in the same way:

```python
        if self.xml_state.get_build_type_name() == 'iso' and \
           self.xml_state.build_type.get_flags() != 'dmsquash':
```
```
flake8 --statistics -j auto --count kiwi/runtime_checker.py
0
```

> Please stop constructing examples out of scope it's not helpful

Really. I am testing your code.

> Thanks

You are welcome.

> btw, both of your suggestions caused flake complains.

You are doing something wrong for one of the cases then, or you are providing a bad .flake8 that I do not have.

kiwi/container/oci.py
aplanas: Is OK to resolve this, but IMHO is bad append list like that. You are using a weir effect of the list constructor, where `list([1] + [2]) == [1] + [2] == [1, 2]` only to avoid a flake8 warning.

```python
        _cmd = ['umoci', 'config'] + \
            self.maintainer + \
            self.user + \
            self.workingdir + \
            self.entry_command + \
            self.entry_subcommand + \
            self.expose_ports + \
            self.volumes + \
            self.environment + \
            self.labels + [
                '--image', container_name,
                '--created', datetime.utcnow().strftime(
                    '%Y-%m-%dT%H:%M:%S+00:00'
                )]
        Command.run(_cmd)
```

I am sorry, this is my last comment about the topic.

kiwi/runtime_checker.py
schaefi: Apologizes if I sounded rude, wasn't my intention. Changes like that in flake really brings no smile on my face. Ok I have applied your change but I'm sorry it complains:

```python
if (
    initrd_system == 'kiwi' and \
        build_type in image_types_supporting_custom_boot_description
    ):
```

```
flake8 --statistics -j auto --count kiwi/runtime_checker.py
kiwi/runtime_checker.py:243:41: E502 the backslash is redundant between brackets
1     E502 the backslash is redundant between brackets
```

I'm using tox to avoid any host side effects and the setup looks clean to me.
The files you mentioned .flake8 do not exist on my system, nowhere

I also wrote after testing, believe me :)

Frankly speaking I more trust the tox env build up from the .virtualenv.requirements.txt file because that includes all the python modules which will be present when the tool runs later on. The env as you setup it only has flake8 and maybe that's not enough. I also use the --upgrade switch and env contains:

pip list | grep flake

```
flake8                   3.6.0     
pyflakes                2.0.0  
```

So I still think the proposed changes here are ok

Thoughts ?

schaefi: So maybe you give my suggested tox command from one of the previous comments a try too ?

schaefi: I also tested the variant without brackets:

```
if initrd_system == 'kiwi' and \
    build_type in image_types_supporting_custom_boot_description:
```

which complained:

```
flake8 --statistics -j auto --count kiwi/runtime_checker.py
kiwi/runtime_checker.py:243:13: E125 continuation line with same indent as next logical line
1     E125 continuation line with same indent as next logical line
1
```

aplanas: I mean, sure, I understand the bad feeling. Please, be sure that I want to help and not complain gratuitously or bother you in any way.

For the specific question, the problem is in the parenthesis, you are providing them meanwhile I do not add them.

I put the line several times, but let me copy again:

This is OK

```python
        if initrd_system == 'kiwi' and \
           build_type in image_types_supporting_custom_boot_description:
```

This is not OK

```python
if (
    initrd_system == 'kiwi' and \
        build_type in image_types_supporting_custom_boot_description
):
```

schaefi: I know about E125 and always solved it by using brackets which imho is the recommendation by python pragma too

schaefi: I consider this a matter of taste and both is accepted by flake8

aplanas: > I also tested the variant without brackets:
> 
> ```
> if initrd_system == 'kiwi' and \
>     build_type in image_types_supporting_custom_boot_description:
> ```

As you can see, this is again not my line, as there is an extra space in the second line.

Before entering in a war of editors, please configure your editor to run flake8 as you type. If you are using emacs I can share my configuration file.

aplanas: > I consider this a matter of taste and both is accepted by flake8

Sure, but please, consider that in Python `if` do not use parenthesis. IMHO this put the valance in one place.

schaefi: ok, you have a point here. I'll adapt the code

aplanas:
This is beautiful, thank you!

schaefi:
Thank you and again sorry if I lost my countenance for a moment, hope you don't mind

PR #863, opened 2018-10-29
Container oci interface

### LINKED ISSUE #859: umoci vs. buildah two players one goal
## Issue description

umoci is the tool to create oci containers (tar archive) from the kiwi prepared system root.
It does its job very well but it seems the scope is limited to suse distributions and there seems
to be not much effort to provide it beyond that distribution. As consequence there is another
tool called buildah (https://rhelblog.redhat.com/tag/buildah) which duplicates the work done
in umoci.

With kiwi we provide an appliance builder which should be able to build any major distribution
and therefore exists for those targets (http://download.opensuse.org/repositories/Virtualization:/Appliances:/Builder) This however implies the required toolchain to be available for the supported distributions. In the end this results in a preference to those tools considered standard or core.

With the current implementation of umoci vs. buildah I see a competitive situation which I'd like to start a discussion if it would be better to move to buildah or provide at least provide support for it like I did it with cdrtools vs. xorriso

buildah is currently being provided in RHEL, CentOS, Fedora, and openSUSE ￼with an ongoing effort to add it to Debian too.

schaefi:
Added OCI tooling interface class
    
An initiative to formulate industry standards around container
formats and runtime is available at https://www.opencontainers.org
Different tools to implement the specifications had been
created. The purpose of this class and its sub-classes is
to provide a common interface in kiwi to allow using all
tools such that the container support in kiwi covers every
linux distribution no matter what tooling was preferred.

This Fixes #859

### DIFF
--- a/kiwi/container/docker.py
+++ b/kiwi/container/docker.py
@@ -15,8 +15,6 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
-import os
-
 # project
 from kiwi.container.oci import ContainerImageOCI
 from kiwi.path import Path
@@ -34,10 +32,6 @@ class ContainerImageDocker(ContainerImageOCI):
 
         :param string filename: file name of the resulting packed image
         """
-        oci_image = os.sep.join([
-            self.oci_dir, ':'.join(['umoci_layout', self.container_tag])
-        ])
-
         additional_tags = []
         for tag in self.additional_tags:
             additional_tags.extend([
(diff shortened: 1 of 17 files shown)


schaefi:
@davidcassany This is quite some refactoring and I hope I completely covered all parts. I did some build tests but it would be good if you can double check that there is no breakage. I haven't had time to look into buildah and did the interface class as I see fit from a class responsibility point of view. Thanks

schaefi:
codacy was very helpful to find an issue, I fixed it and rebased, since then the status did not change. Probably it takes some time to catch up or it does not like rebasing... I'll wait

davidcassany:
I tested the change in some several images (based and derived) with only green results.

> codacy was very helpful to find an issue, I fixed it and rebased, since then the status did not change. Probably it takes some time to catch up or it does not like rebasing... I'll wait

yes this is quite strange...

schaefi:
> yes this is quite strange...

yes I think this is a codacy issue. If I go on the issue and want to view the file it says that this file is no longer present in the pull request. I think the system got confused about me rebase and force push the fix.

Thanks for testing I'll merge it

PR #864, opened 2018-10-30
Distinguish install and image dracut config

### LINKED ISSUE #858: Install media dracut initrd is built using the same options as the initrd from the disk image
## Problem description

The dracut initrd build for the install media inherits build options form the disk initrd, which is not always correct. An specific example is when the disk makes use of a luks device. In that case the initrd installs `/etc/crypttab` also for the installation media. This causes the install media hang on boot until the time out for the disk UUID expires (1 or 2 minutes), after that it continues to boot.

## Expected behaviour

Install media initrd does not include files like `/etc/crypttab`

Build an OEM image with `installiso="true"` and `luks="passwd"` and boot from the install iso.

davidcassany:
This commit distinguishes the files that should be installed inside
the image dracut only than the ones installed in both, in install initrd
and image initrd.

Fixes #858

### DIFF
--- a/kiwi/boot/image/base.py
+++ b/kiwi/boot/image/base.py
@@ -78,7 +78,7 @@ class BootImageBase(object):
         """
         pass
 
-    def include_file(self, filename):
+    def include_file(self, filename, install_media=False):
         """
         Include file to boot image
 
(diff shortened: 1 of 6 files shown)


davidcassany:
@schaefi, I agree the concept is not the simplest one, it took me also a while to figure out the actual behavior. I tried to minimize the impact of the change with this PR, as those are sensitive parts. 

We can refactor the code in the future, I'd include an issue for that, so can clarify the scope and expectations with the refactor.

schaefi:
Sounds good

PR #865, opened 2018-10-31
Support ramdisk deployment in OEM images

schaefi:
Using the rd.kiwi.ramdisk boot option enables the deployment
into a ramdisk. If this option is enabled only ramdisk devices
as provided by the brd kernel driver will be available for
deployment

For testing I setup an image with the following type specification

```xml
<type image="oem" filesystem="ext4" installiso="true" bootloader="grub2" initrd_system="dracut" installboot="install" boottimeout="1" kernelcmdline="rd.kiwi.ramdisk ramdisk_size=2048000">
    <oemconfig>
        <oem-skip-verify>true</oem-skip-verify>
        <oem-unattended-id>/dev/ram1</oem-unattended-id>
        <oem-swap>false</oem-swap>
    </oemconfig>
</type>
```

This is quite handy for in RAM only deployments.

However the feature has its limitations. Only standard images which can be booted by a simple root mount and root switch can be booted that way. Usually we kexec after deployment such that the correct for the image created dracut initrd will boot the image. In case of a RAM only system you can't kexec because you would loose the ramdisk contents. Thus the dracut initrd driving the deployment is also the environment to boot the system. Under certain circumstances this environment is not suitable. Thinking of e.g luks encrypted images which requires unlock inside the initrd and probably we can construct more cases.

Thanks

### DIFF
--- a/doc/source/building/build_oem_disk.rst
+++ b/doc/source/building/build_oem_disk.rst
@@ -1,3 +1,5 @@
+.. _oem:
+
 Build an OEM Expandable Disk Image
 ==================================
 
--- a/doc/source/building/working_with_images.rst
+++ b/doc/source/building/working_with_images.rst
@@ -18,3 +18,4 @@ different image types.
    working_with_images/uninstall_package_requests 
    working_with_images/vagrant_setup
    working_with_images/pxe_live_iso_boot
+   working_with_images/oem_ramdisk_deployment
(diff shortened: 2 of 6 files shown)


davidcassany:
@schaefi I guess we should also document `rd.kiwi.ramdisk` usage in a follow up request.

schaefi:
@davidcassany yes will add a doc chapter for it

I'm running out of time, leaving in a couple minutes for the wedding event I told you about. I'll review your PR when I'm back, guess we have no time pressure here.

Thanks

davidcassany:
> I'll review your PR when I'm back, guess we have no time pressure here.

Sure no problem, we are not in hurry with it.

PR #866, opened 2018-11-05
Gh pages deploy setup

schaefi:
Use the travis pages provider for doc deployment
    
In the past we used the travis-sphinx script to deploy
the documentation to gh-pages via the travis CI. However
with the change to the travis github App we need to
change this deployment into a real deployment stage
of the travis setup

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,16 +17,24 @@ jobs:
     env: TOXENV=unit_py2_7,check,doc
   - stage: deploy
     python: 3.6
-    env: TOXENV=doc_travis,doc_travis_deploy
+    env: TOXENV=doc_travis
     deploy:
-      skip_cleanup: true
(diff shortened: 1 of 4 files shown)


schaefi:
I haven't tested the deployment since this would require the change to become applied in master. Kind of a chicken and egg problem. Thus another pair of eyes would be helpful to avoid obvious errors. Thanks much

schaefi:
good point, I'll change that to tags

schaefi:
ok, let's try... probably there will be followup changes required because I never got that right on the first shot :) I'll do those in master then, so stay tuned

davidcassany:
I am standing looking at Travis build :) 
 
> I'll do those in master then, so stay tuned

No problem, I'll make sure I update my local repo after changes on master

PR #871, opened 2018-11-06
Add history metadata for container builds

### LINKED ISSUE #852: Support arbitrary history entries
Currently, any image built with kiwi has this (or similar) as output of `docker history`:

```
IMAGE               CREATED             CREATED BY          SIZE                COMMENT
693d2f77e3c6        24 hours ago        umoci config        0B                  
<missing>           24 hours ago        umoci config        104MB
```

It should only have one of those entries if possible and the comment should be specified in the kiwi description. Example:

```
IMAGE              CREATED             CREATED BY        SIZE         COMMENT
693d2f77e3c6       24 hours ago        kiwi              104MB        A container I built
```

davidcassany:
This commit adds the history section in contianerconfig. With it
'author', 'created_by' and 'comment' can be customized. In addition
'created' is always included with the image creation date time.
'created_by' entry is set to 'KIWI __version__' by default if nothing
is provided.

Fixes #852

### DIFF
--- a/kiwi/container/oci.py
+++ b/kiwi/container/oci.py
@@ -128,7 +128,7 @@ class ContainerImageOCI(object):
             image_tar = ArchiveTar(base_image)
             image_tar.extract(self.oci.container_dir)
 
-        self.oci.init_layout(True if base_image else False)
+        self.oci.init_layout(bool(base_image))
 
         self.oci.unpack(self.oci_root_dir)
         oci_root = DataSync(
@@ -144,7 +144,7 @@ class ContainerImageOCI(object):
             for tag in self.oci_config['additional_tags']:
                 self.oci.add_tag(tag)
 
-        self.oci.set_config(self.oci_config, True if base_image else False)
+        self.oci.set_config(self.oci_config, bool(base_image))
 
         self.oci.garbage_collect()
 
--- a/kiwi/container/oci.py
+++ b/kiwi/container/oci.py
@@ -52,7 +52,12 @@ class ContainerImageOCI(object):
                 'expose_ports': ['80', '42'],
                 'volumes': ['/var/log', '/tmp'],
                 'environment': {'PATH': '/bin'},
-                'labels': {'name': 'value'}
+                'labels': {'name': 'value'},
+                'history': {
+                    'created_by': 'some explanation here',
+                    'comment': 'some comment here',
+                    'author': 'tux'
+                }
             }
     """
     def __init__(self, root_dir, custom_args=None):
@@ -95,6 +100,12 @@ class ContainerImageOCI(object):
             self.oci_config['entry_subcommand'] = \
                 Defaults.get_default_container_subcommand()
 
+        if 'history' not in self.oci_config:
+            self.oci_config['history'] = {}
+        if 'created_by' not in self.oci_config['history']:
+            self.oci_config['history']['created_by'] = \
+                Defaults.get_default_container_created_by()
+
         self.oci = OCI(self.oci_config['container_tag'])
 
     def create(self, filename, base_image):
@@ -117,7 +128,7 @@ class ContainerImageOCI(object):
             image_tar = ArchiveTar(base_image)
             image_tar.extract(self.oci.container_dir)
 
-        self.oci.init_layout(base_image)
+        self.oci.init_layout(True if base_image else False)
 
         self.oci.unpack(self.oci_root_dir)
         oci_root = DataSync(
@@ -133,7 +144,7 @@ class ContainerImageOCI(object):
             for tag in self.oci_config['additional_tags']:
                 self.oci.add_tag(tag)
 
-        self.oci.set_config(self.oci_config)
+        self.oci.set_config(self.oci_config, True if base_image else False)
 
         self.oci.garbage_collect()
 
--- a/kiwi/xml_parse.py
+++ b/kiwi/xml_parse.py
@@ -16,7 +16,7 @@
 #   kiwi/schema/kiwi_for_generateDS.xsd
 #
 # Command line:
-#   /home/ms/Project/kiwi/.tox/3.6/bin/generateDS.py -f --external-encoding="utf-8" --no-dates --no-warnings -o "kiwi/xml_parse.py" kiwi/schema/kiwi_for_generateDS.xsd
+#   /home/david/work/kiwi/.env3/bin/generateDS.py -f --external-encoding="utf-8" --no-dates --no-warnings -o "kiwi/xml_parse.py" kiwi/schema/kiwi_for_generateDS.xsd
 #
 # Current working directory (os.getcwd()):
 #   kiwi
@@ -4865,7 +4865,7 @@ class containerconfig(GeneratedsSuper):
     useful container information."""
     subclass = None
     superclass = None
-    def __init__(self, name=None, tag=None, additionaltags=None, maintainer=None, user=None, workingdir=None, entrypoint=None, subcommand=None, expose=None, volumes=None, environment=None, labels=None):
+    def __init__(self, name=None, tag=None, additionaltags=None, maintainer=None, user=None, workingdir=None, entrypoint=None, subcommand=None, expose=None, volumes=None, environment=None, labels=None, history=None):
         self.original_tagname_ = None
         self.name = _cast(None, name)
         self.tag = _cast(None, tag)
@@ -4897,6 +4897,10 @@ class containerconfig(GeneratedsSuper):
             self.labels = []
         else:
             self.labels = labels
+        if history is None:
+            self.history = []
+        else:
+            self.history = history
     def factory(*args_, **kwargs_):
         if CurrentSubclassModule_ is not None:
             subclass = getSubclassFromModule_(
@@ -4938,6 +4942,11 @@ class containerconfig(GeneratedsSuper):
     def add_labels(self, value): self.labels.append(value)
     def insert_labels_at(self, index, value): self.labels.insert(index, value)
     def replace_labels_at(self, index, value): self.labels[index] = value
+    def get_history(self): return self.history
+    def set_history(self, history): self.history = history
+    def add_history(self, value): self.history.append(value)
+    def insert_history_at(self, index, value): self.history.insert(index, value)
+    def replace_history_at(self, index, value): self.history[index] = value
     def get_name(self): return self.name
     def set_name(self, name): self.name = name
     def get_tag(self): return self.tag
@@ -4957,7 +4966,8 @@ class containerconfig(GeneratedsSuper):
             self.expose or
             self.volumes or
             self.environment or
-            self.labels
+            self.labels or
+            self.history
         ):
             return True
         else:
@@ -5019,6 +5029,8 @@ class containerconfig(GeneratedsSuper):
             environment_.export(outfile, level, namespace_, name_='environment', pretty_print=pretty_print)
         for labels_ in self.labels:
             labels_.export(outfile, level, namespace_, name_='labels', pretty_print=pretty_print)
+        for history_ in self.history:
+            history_.export(outfile, level, namespace_, name_='history', pretty_print=pretty_print)
     def build(self, node):
         already_processed = set()
         self.buildAttributes(node, node.attrib, already_processed)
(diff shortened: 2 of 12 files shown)


kiwi/container/oci.py
```diff
@@ -52,7 +52,12 @@ class ContainerImageOCI(object):
                 'expose_ports': ['80', '42'],
                 'volumes': ['/var/log', '/tmp'],
                 'environment': {'PATH': '/bin'},
-                'labels': {'name': 'value'}
+                'labels': {'name': 'value'},
+                'hisotry': {
```
Vogtinator: Typo.

davidcassany: Thanks solved

davidcassany:
@schaefi, @Vogtinator while implementing it I noticed we should agree on the behavior of the history records. With the current implementation a `type` section like the following one:

 ```xml
<type image="docker">
  <containerconfig
    name="opensuse"
    tag="tumbleweed"
    maintainer="David Cassany &lt;dcassany@suse.com&gt;">
    <history author="David" created_by="A kiwi build">This is some comment for history</history>
  </containerconfig>
</type>
```

Produces a `docker history opensuse:tumbleweed` output like the following one:

```
IMAGE               CREATED              CREATED BY          SIZE                COMMENT
9e0d44e62083        About a minute ago   A kiwi build        0B                  This is some comment for history
<missing>           About a minute ago   KIWI 9.16.36        113MB               umoci repack
```

I noticed that each `umoci config` and `umoci repack` creates a new history record into the image. Thus for any KIWI build we will have at least a couple of history entries. In both cases umoci can customize those entries with the flags:

```
--history.comment
--history.author
--history.created
--history.created_by
```

With the suggested implementation `--history.created` is always used for all commands with the build time date, this is **not** customizable in any case.

`umoci repack` is also **not** customizable and `comment` and `created_by` history values are set by default to `umoci repack` and to `KIWI __version__`.

The last `umoci config` call is customizable, in that case `created_by`, `author` and `comment` can be defined like in the above `<containerconfig>` example. In that case if no history is provided `created_by` is set to the default value: `KIWI __version__`. `comment` and `author` have no default value.

My doubts here are related to default values and what happens if multiple `umoci config` or `umoci repack` calls are executed (each additional tag represents at least an `umoci config` call).

Regarding default values:
* We could assume that default the `author` is the same author value provided inside `<description><author>`
* In the same direction we could assume that the default `comment` is the same value provided inside `<description><specification>`

I am not a fan of those options, but would make the description file less verbose and also fill history records with the user provided data.

Regarding multiple `umoci` calls causing new history entries:
* We could use the same history flags for every single call, in that case there would be duplicated entries in history.
* We could use the history flags only for the last `umoci config` call, in that case there would be at a least an entry with an empty comment and `umoci config` as the  `CREATED BY` value.
* We could use the history flags for the last `umoci config` call and use some defaults for the rest.

All those options have very little impact on the current suggested implementation, so I am opened to agree on a different behavior.

**Comment by Vogtinator** [comment]
Isn't it possible to just use a single umoci call for everything?

**Comment by davidcassany** [comment]
> Isn't it possible to just use a single umoci call for everything?

No, at least `umoci-repack` and `umoci-config` calls are required. After commenting it with @cyphar an option could be to add a `--no-history` flag in `umoci-config` or some other similar approach.

In any case we need to agree about the expectations for history records using current umoci.

**Comment by schaefi** [comment]
I think a good solution would be the ability to have a --no-history flag for umoci repack/config commands. I think that's what @cyphar suggested to add. However the current umoci doesn't provide this feature. Given it can be added I vote for that solution plus an updated version check in the check_docker_tool_chain_installed() runtime check.

Having a history entry for the umoci repack looks weird to me as we need that to process the initial creation of the container archive only. If we have to add a comment it imho should express something like ```kiwi: initial archive packing```. Having a history entry for the umoci config could express ```kiwi: initial container setup```.

That information wouldn't be that meaningful and I think if the container is used as a base container all that history entries gets inherited which creates a weird history.

So my first vote would be to have an option to skip the history entry such that we can create an image which only provides meaningful history entries

Thoughts ?

**Comment by Vogtinator** [comment]
> No, at least umoci-repack and umoci-config calls are required. After commenting it with @cyphar an option could be to add a --no-history flag in umoci-config or some other similar approach.

Yup.

> So my first vote would be to have an option to skip the history entry such that we can create an image which only provides meaningful history entries

As long as there's a fallback to having a mostly useless entry in case umoci doesn't support the option, I agree.

**Comment by cyphar** [comment]
The patch for `--no-history` is in openSUSE/umoci#/270.

> Having a history entry for the umoci repack looks weird to me as we need that to process the initial creation of the container archive only.

This is a peculiarity of OCI images. In order for other tools to understand your image history, you have to have an image history associated with each layer (it's really ugly how it works internally, it's just a dumb list with a boolean saying whether a particular entry is associated with a layer or not -- so layers without history entries break the history entries of later layers).

In theory you could order them in any order you like, but obviously we'd need to have a UX to do this and thus another `umoci` patch. I wouldn't particularly mind (one of the ideas I pitched to @davidcassany was this), but I think that `--no-history` is a simpler option.

**Comment by schaefi** [comment]
>Up to my understanding container configuration (umoci config) is not tight to any layer, is image wide, >thus we could even reorganize our calls order and call umoci-repack as the last step, then probably it >looks more natural to include history records there.

I agree, having the repack last would be a good call order change

**Comment by schaefi** [comment]
> The patch for --no-history is in openSUSE/umoci#270.

great you are fast :) So once merged this results in a package build somewhere ?

**Comment by davidcassany** [comment]
Given the discussion and @cyphar collaboration I suggest the following approach.

1- Make a new PR with a little refactor on how we call `umoci`, so we call `umoci-repack` as the last step. (no history flags used)
2- Update this PR (or new one) so history flags are used for the `umoci-repack` call only and the schema is updated to include the history data.
3- Finally another PR  to check weather `--no-history` flag is available and used it if possible for any `umoci-config` call.

I think this should be simple and safe. 

@Vogtinator, @schaefi  does it make sense to you?

**Comment by Vogtinator** [comment]
> @Vogtinator, @schaefi does it make sense to you?

Yup. I'm not sure why step 1 is necessary though.

**Comment by davidcassany** [comment]
> Yup. I'm not sure why step 1 is necessary though.

not necessary but desirable I'd say. Also for dervied images we have one extra `umoci-config` call that is not strictly needed that leads to an additional history record.

**Comment by schaefi** [comment]
workflow sounds good to me, I think we don't need a new PR and just update this one here

**Comment by davidcassany** [comment]
@schaefi, finally I have not reordered our `umoci-repack` call to be the last step, apparently `umoci-repack` restores the configuration that was in place at the time of `umoci-unpack`, thus any `umoci-config` in between gets lost. I just done a very little refactor to avoid the extra `umoci-config` call we had for derived images.

So now with a type section like:
```xml
<type image="docker">
  <containerconfig name="opensuse" tag="tumbleweed">
    <history>Include tumbleweed base image layer</history>
  </containerconfig>
</type>
```
we get a history record like:
```
IMAGE               CREATED             CREATED BY          SIZE                COMMENT
fa044e36fc1f        About an hour ago   KIWI 9.16.36        0B                  Include tumbleweed base image layer
<missing>           About an hour ago   umoci config        113MB
```
And for derived images, with type section like:
```xml
<type image="docker" derived_from="file:///mybaseimage.tar">
  <containerconfig name="opensuse/derived" tag="derived">
    <history author="David" created_by="A kiwi build">Adding a new layer</history>
  </containerconfig>
</type>
```
we get a history record like:
```
IMAGE               CREATED             CREATED BY          SIZE                COMMENT
bfc980576a05        About an hour ago   A kiwi build        0B                  Adding a new layer
<missing>           About an hour ago   umoci config        9.03MB
<missing>           About an hour ago   KIWI 9.16.36        0B                  Include tumbleweed base image layer
<missing>           About an hour ago   umoci config        113MB
```

This will be the result of any build using an `umoci` that does not have the `--no-history` flag. I'll make the implementation of `--no-history` flag in a follow up request.

cyphar:
@davidcassany 

> apparently umoci-repack restores the configuration that was in place at the time of umoci-unpack, thus any umoci-config in between gets lost.

Yes, this is the case. It's actually not that the configuration is being restored, it's that `umoci unpack` stores the exact descriptor that the extracted image referred to so that it know what image it is going to be based on when it creates a new layer. If you modify the image tag, it doesn't affect this descriptor (which cryptographically points at the manifest for the image and thus the whole Merkle tree).

```
<missing>           About an hour ago   umoci config        113MB
```

Are you sure that output is right? Why is a config change being listed as being `113MB` (looks like there's a mismatch in the history when it is parsed)? What are you using to get the history information (please don't tell me Docker broke something again)?

davidcassany:
> @davidcassany
> 
> > apparently umoci-repack restores the configuration that was in place at the time of umoci-unpack, thus any umoci-config in between gets lost.
> 
> There isn't really a robust way to work around this in umoci without making umoci potentially generate broken images

Yes I understand, anyway this is not an issue.

> ```
> <missing>           About an hour ago   umoci config        113MB
> ```
 
> Are you sure that output is right?

It is right, at least the copy & paste I did :)

> Why is a config change being listed as being `113MB` (looks like there's a mismatch in the history when it is parsed)? What are you using to get the history information (please don't tell me Docker broke something again)?

This is the command sequence:

```
in kiwi buid:
  umoci repack
  umoci config
  umoci gc
  skopeo copy

loading the image:
  docker load
  docker history
```

I have no checked the history records for the OCI image neither the history records before being loaded, thus I guess it could be also `skopeo` that is messing around with the history.

kiwi/container/oci.py
```diff
@@ -117,7 +128,7 @@ def create(self, filename, base_image):
             image_tar = ArchiveTar(base_image)
             image_tar.extract(self.oci.container_dir)
 
-        self.oci.init_layout(base_image)
+        self.oci.init_layout(True if base_image else False)
```
schaefi: base_image is None if not set so that evaluates the same as False in the way init_layout is coded. If set it's a string which evaluates as True in the way init_layout is coded. But I agree this change makes it more clear and bool correct

schaefi: just a side note

kiwi/xml_parse.py
```diff
@@ -16,7 +16,7 @@
 #   kiwi/schema/kiwi_for_generateDS.xsd
 #
 # Command line:
-#   /home/ms/Project/kiwi/.tox/3.6/bin/generateDS.py -f --external-encoding="utf-8" --no-dates --no-warnings -o "kiwi/xml_parse.py" kiwi/schema/kiwi_for_generateDS.xsd
+#   /home/david/work/kiwi/.env3/bin/generateDS.py -f --external-encoding="utf-8" --no-dates --no-warnings -o "kiwi/xml_parse.py" kiwi/schema/kiwi_for_generateDS.xsd
```
schaefi: One thing we should check is that we don't re-introduce the issue solved in (2ea7fbd13f13487)

@davidcassany can you make sure generateDS version is v2.29.14 ? Thanks

davidcassany: @scheafi yes it is, see 5th line :smile: Yes I remember this ugly issue...

kiwi/container/oci.py
cyphar: Couldn't you do `bool(base_image)`?

test/unit/container_image_oci_test.py
```diff
@@ -164,17 +167,16 @@ def test_create_derived(
...
         self.oci.oci.repack.assert_called_once_with('kiwi_oci_root_dir')
         self.oci.oci.set_config.assert_called_once_with({
             'container_name': 'foo/bar',
             'additional_tags': ['current', 'foobar'],
             'container_tag': 'latest',
-            'entry_subcommand': ['/bin/bash']
-        })
+            'entry_subcommand': ['/bin/bash'],
+            'history': {'created_by': 'KIWI {0}'.format(__version__)}
+        }, True)
```
davidcassany: I had added the `base_image` as real Boolean because this test was really confusing if not. I know the result is the same, but having a string pointing to a directory here felt really strange.

kiwi/xml_parse.py
schaefi: :+1:

kiwi/container/oci.py
davidcassany: @cyphar I haven't thought about that. Yes that's probably nicer, just need to make some tests on python2 and 3 to make sure it does not evaluate different.

test/unit/container_image_oci_test.py
schaefi: yes understood, actually I also liked what @cyphar suggested

```python
bool(base_image)
```

PR #876, opened 2018-11-08
Fix rsync call for filesystem images

### LINKED ISSUE #875: When using type image="[filesystem]", file system is inside a image-root directory
When trying to create a plain rootfs image with `<type image="[filesystem]">`, the root filesystem is all contained inside a image-root directory. I would expect the filesystem to be right at `/`.

```
$ sudo mount -o loop /tmp/kiwi-root/openSUSE-rootfs.x86_64-1.0.0.btrfs /mnt

$ tree -d /mnt -L 2                                                             
/mnt
└── image-root
    ├── bin
    ├── boot
    ├── dev
    ├── etc
    ├── home
  ...
```

Kiwi version: `KIWI (next generation) version 9.16.27`

/cc @davidcassany

davidcassany:
For filesystem images the rsync call was missing a finale slash for
the source path causing the sync also the containing directory. With
this change the filesystem image does not include the rootfs in any
subdirectory.

Fixes #875

### DIFF
--- a/kiwi/builder/filesystem.py
+++ b/kiwi/builder/filesystem.py
@@ -16,6 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import platform
+import os
 
 # project
 from kiwi.filesystem import FileSystem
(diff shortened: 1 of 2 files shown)


schaefi:
I'm not sure if this slash is causing after effects on non simple filesystem builds. I could imagine that vmx, oem and install media creation breaks with it. Could you double check that ?

The behavior of rsync with and without the '/' is pretty different, we should be very very careful

davidcassany:
Ok with this change now only filesystem builds should affected by this change. Anyway I validated that disk builds are still building and booting.

schaefi:
Yes this looks great now, only changing the filesystem builder :+1:

PR #880, opened 2018-11-09
kiwi-dump-image: improve dialog usage

seife:
Dialog's "--radiolist" feature requires to navigate to the item,
press "space" to select the item and then "enter" to execute.
With "--menu", it is enough to just navigate to the item and press
"enter" to execute, which is much more intuitive for most users.

Note that this will bring a change in behaviour and thus should probably be advertised.
Also, before pressing "space" on the disk did select it, now space acts already as "execute" (same as "enter")

### DIFF
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -34,7 +34,6 @@ function get_disk_list {
     local disk_device
     local disk_device_by_id
     local disk_meta
-    local item_status=on
     local list_items
     local blk_opts="-p -n -r -o NAME,SIZE,TYPE"
     if [ ! -z "${kiwi_devicepersistency}" ];then
@@ -86,8 +85,7 @@ function get_disk_list {
                 continue
             fi
         fi
-        list_items="${list_items} ${disk_device} ${disk_size} ${item_status}"
-        item_status=off
+        list_items="${list_items} ${disk_device} ${disk_size}"
     done
     if [ -z "${list_items}" ];then
         local no_device_text="No device(s) for installation found"
@@ -135,7 +133,7 @@ function get_selected_disk {
         else
             # manually select from storage list
             if ! run_dialog \
-                --radiolist "\"Select Installation Disk\"" 20 75 15 \
+                --menu "\"Select Installation Disk\"" 20 75 15 \
                 "$(get_disk_list)"
             then
                 die "System installation canceled"


schaefi:
:smile: I knew at some point this type of patch will come up.

Ok I had a reason for the radiolist and it was because you really have to think on your selection.
You can blame me paranoid but in this case I thought it would be good to make it not too easy to do probably the wrong thing.

Given that I also added the last message which lists the selected device and again asks if you want to continue I'm open to make the menu style more user friendly. There is one thing I dislike much with the --menu style and that's that space counts as Enter

I'd like to have at least the opinion from David and I also suggest to send a short feedback mail to the mailing list which explains the change and asks for objections. If we don't get any objections I'm ok and we can tell that we announced it.

I'd like to avoid negative user feedback with us changing the behavior and not announcing it beforehand

Makes sense ?

davidcassany:
> There is one thing I dislike much with the --menu style and that's that space counts as Enter

I completely agree with that. However the continue (yes/no) dialog after the selection should be already a good safeguard against any spurious space press.

I am fine with the change, but yes, this is a behavior change that needs to be announced.

> I'd like to avoid negative user feedback with us changing the behavior and not announcing it beforehand

Yes, probably we should agree on some time frame also, what about a couple of weeks? or a month? Just some time to gather any potential objection after announcing the change in the mailing list. I'd say a couple of weeks would be enough. We could also forward again the announcement after the first week.

seife:
I totally understand that this behaviour might be intended, but for me it was just too often counterintuitive --  I scrolled down 30 JBOD disks to the SSDs I wanted to install on, only to forget to hit "space" first and then install on the wrong disk :smile:
I did this more than once, and finally came up with this patch (originally for the kiwi-7.x series).
But as written, I can live with this being a local patch in my environment.
Maybe (if this is possible with dialog, not checked) an "instruction"-line under the buttons "select your disk with [space], then press [enter] to continue installation" would be good.

BTW: the confirmation message after selecting the disk has also room for improvement, as 
1) there is no way back. If you press "no", then the machine is supposed to reboot (unhelpful with big servers that take considerable amount of time to boot)
2) actually "die" does not seem to work, the process just halts and does not reboot the machine at all.

But those are different issues and I'll try to tackle them with patches separately.

schaefi:
> there is no way back. If you press "no", then the machine is supposed to reboot (unhelpful with big
> servers that take considerable amount of time to boot)

agreed, so it should either reboot or go back to the menu. I'll open an issue for that one

schaefi:
> actually "die" does not seem to work, the process just halts and does not reboot the machine at all

die just does what the name says ;) it was not intended to reboot when issue a die() call. I agree with you that we should change the behavior on abort of the install but the behavior of die is as expected

schaefi:
I'll write up an e-mail now to the list to gather some feedback. Two weeks is enough. After that we'll make the move on positive or none feedback

schaefi:
3 days to go and no objections so far. I'm convinced this will make it's way to master

PR #881, opened 2018-11-09
Adding OEM and ISO image types for Fedora tests

davidcassany:
This PR updates the Fedora test to include also OEM and ISO image builds


davidcassany:
@schaefi I have some testing build [here](https://build.opensuse.org/package/show/home:dcassany:branches:Virtualization:Appliances:Images:Testing_x86/test-image-fedora)

My idea, after merging, is to remove the `test-image-fedora-vmx` package from `Virtualization:Appliances:Images:Testing_x86` and add the `test-image-fedora` test I did including the multibuild.

schaefi:
Like we have it for centos :+1:

PR #883, opened 2018-11-13
dracut/kiwi-lib: add a custom "die" function

seife:
The message passed to "die" is never seen by most, because of
"loglevel=0" usually passed to e.g. oem iso installs.
Create a custom wrapper function, that increases kernel loglevel before
actually calling dracut's die, so users can see what's going on.

How to test: abort an interactive OEM ISO installation at the disk selection without this patch, and with it.
Without, you only see "Started dracut interactive", before the machine being silently halted (not powered off), with this version you can actually guess what is going on by the on-screen messages.

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-lib.sh
@@ -10,6 +10,11 @@ function setup_debug {
     fi
 }
 
+function kiwi_die {
+    echo 7 > /proc/sys/kernel/printk
+    die $@
+}
+
 function set_root_map {
     root_map=$1
     export root_map
@@ -32,7 +37,7 @@ function lookup_disk_device_from_root {
     declare root=${root}
     local root_device=${root#block:}
     if [ -z "${root_device}" ];then
-        die "No root device found"
+        kiwi_die "No root device found"
     fi
     if [ -L "${root_device}" ];then
         root_device=/dev/$(basename "$(readlink "${root_device}")")
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -12,7 +12,7 @@ function initialize {
     local profile=/.profile
 
     test -f ${profile} || \
-        die "No profile setup found"
+        kiwi_die "No profile setup found"
 
     import_file ${profile}
 }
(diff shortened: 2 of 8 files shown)


seife:
This might also be the base to make "die" actually behave like in old kiwi-7, which means: wait for 120 seconds to investigate, maybe spawn a debug shell and then reboot to retry (which all will then be handled inside kiwi_die())

davidcassany:
@seife apparently there is a small styling issue with this patch, see the Travis CI failure. We use `shellcheck` utility to prevent inconsistent styling and some common bash pitfalls.

dracut/modules.d/99kiwi-lib/kiwi-lib.sh
```diff
@@ -10,6 +10,11 @@ function setup_debug {
     fi
 }
 
+function kiwi_die {
+    echo 7 > /proc/sys/kernel/printk
+    die $@
```
davidcassany: `shellcheck` fails for this line with:

`SC2068: Double quote array expansions to avoid re-splitting elements.`

schaefi:
I don't think we need that. Use the following rd options for debugging:

```
rd.debug rd.kiwi.debug
```

The first makes sure all messages also the one die produces will be printed on the console. In addition the die leads to the rescue shell as dracut implements it. From that shell you can inspect and debug the kiwi related debug log in /run/initramfs/log/boot.kiwi which is written if rd.kiwi.debug is set

If you need a certain break point use rd.break

There is really no need to again start adding more extra sauce to the initrd stuff than needed. I rather would work upstream if the standard die method is not what you expect.

I'm sorry no kiwi_die please

I apologize for an eventual coding effort which will not make it upstream, maybe we discuss stuff like that in an issue beforehand

seife:
"rd.debug rd.kiwi.debug" on real hardware with slow VGA terminal leads to udev-initqueue timeouts and prevents anything at all, so using this as default is totally out of question. It also messes up the terminal so the dialog stuff is looking really bad.

Just try an interactive install with two disks, then cancel the disk selection and you will see what I mean. You get no message at all, but the system will just halt silently (but it looks like it is waiting for something and then continue).
This is not user friendly at all ;-)

dracut/modules.d/99kiwi-lib/kiwi-lib.sh
seife: pushed an amended version

seife:
changing loglevel=0 to loglevel=1 in kiwi/bootloader/config* might be a simple, but helpful solution:

image

This also makes the pre-dialog boot slightly more verbose, but I don't think this is a huge problem.

PR #886, opened 2018-11-15
Include the isolinux ubuntu paths

davidcassany:
This commit adds the ubuntu installation paths the places to look
for isolinux binaries.


schaefi:
I was wondering why no test update was needed with this change, will take a short look

schaefi:
@davidcassany mind if I add a commit which changes the xorriso test on that paths ?

davidcassany:
@schaefi sure no problem, great we add a test for that.

PR #887, opened 2018-11-19
Add parted dependency for dracut-kiwi-live package

davidcassany:
dracut-kiwi-live requires `partprobe` tool and this is provided by
parted package. Persistent overlay setup fails if parted is not
installed in the image.

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -406,6 +406,7 @@ Requires:       device-mapper
 %endif
 Requires:       dracut
 Requires:       xorriso
+Requires:       parted
 License:        GPL-3.0-or-later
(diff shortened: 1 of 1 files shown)


davidcassany:
@schaefi I noticed this issue while testing the live images for Debian and Ubuntu, where I had to include parted into the image to get the overlay feature working fine, see it [here](https://build.opensuse.org/package/view_file/home:dcassany:branches:Virtualization:Appliances:Images:Testing_x86/test-image-debian/appliance.kiwi?expand=1).

I have checked that `partprobe` is provided by `parted` package in openSUSE, Fedora, Centos, Debian and Ubuntu. Thus the change should be fine for all those.

PR #889, opened 2018-11-23
Create a custom `setup.py clean` command

davidcassany:
The default `setup.py clean` command does not clean `dist` and
`*.egg-info` folders. This commit makes sure those are cleaned and
also ands a `setup.py clean` command in the `clean` Makefile target


schaefi:
I'm going to merge this and submit new packages with the changes applied

Thanks much

PR #890, opened 2018-11-26
Use cow file on persistent grub live loop boot

### LINKED ISSUE #860: Persistent usb does not work with opensuse
I'm trying to use the command "sudo live-grub-stick --suse-persistent ./openSUSE-Tumbleweed-KDE-Live-x86_64-Snapshot20181018-Media.iso /dev/sdd1" to generate a persistent pendrive, but persistence is not working. Once the command in the grub "isofrom=" has been changed to "iso-scan/filename=" the persistence has not worked any more, this happens after the second half of 2017.

schaefi:
When using tools like live-grub-stick the live iso as generated
by kiwi will be copied as file on the target device and a grub
loopback setup is created there to boot the live system from file.
In such a case the persistent write setup which tries to create
an extra write partition on the target fails in almost all cases
because the target has no free and unpartitioned space available.
Therefore in case of such a loopback mounted system we create a
cow file(live_system.cow) instead of a partition to setup
persistent writing. The cow file will be created in the same
directory the live iso image file was read from grub
This Fixes #860

### DIFF
--- a/doc/source/overview/workflow.rst
+++ b/doc/source/overview/workflow.rst
@@ -873,6 +873,19 @@ the available kernel boot parameters for this modules:
   This variable tells a live iso image which filesystem should be
   used to store data on the persistent write partition.
 
+``rd.live.cowfile.mbsize``
+  This variable tells a live iso image the size of the cowfile in MB.
+  When using tools like `live-grub-stick` the live iso will be copied
+  as a file on the target device and a grub loopback setup is created
+  there to boot the live system from file. In such a case the
+  persistent write setup, which usually creates an extra write
+  partition on the target, will fail in almost all cases because
+  the target has no free and unpartitioned space available.
+  Because of that a cow file(live_system.cow) instead of a partition
+  is created. The cow file will be created in the same directory
+  the live iso image file was read from by grub and takes the
+  configured size or the default size of 500MB.
+
 ``rd.live.dir``
   This variable tells a live iso image the directory which contains
   the live OS root directory. Defaults to `LiveOS`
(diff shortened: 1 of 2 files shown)


schaefi:
@davidcassany the change of the variable mountpoint to mount_point was just because there is also a tool called ```mountpoint``` and that confused me. This changed wouldn't be required but it was just to put me right

schaefi:
Thanks for the review, yes this was an issue I wanted to solve for a long time, now finally done.
Also added a doc entry which I missed before

wandersfalsin:
> When using tools like live-grub-stick the live iso as generated
> by kiwi will be copied as file on the target device and a grub
> loopback setup is created there to boot the live system from file.
> In such a case the persistent write setup which tries to create
> an extra write partition on the target fails in almost all cases
> because the target has no free and unpartitioned space available.
> Therefore in case of such a loopback mounted system we create a
> cow file(live_system.cow) instead of a partition to setup
> persistent writing. The cow file will be created in the same
> directory the live iso image file was read from grub
> This Fixes #860


82/5000
Is it possible to change the cow file name live_system.cow at boot time?

schaefi:
>    Is it possible to change the cow file name live_system.cow at boot
>    time?

The name of the file is hard coded in the code. See: ```modules.d/90kiwi-live/kiwi-live-lib.sh```. There you will find:

```bash
local cow_file_name="${isoscan_loop_mount}/live_system.cow"
```

So at the moment there is no customization option for that part.
However you can work with several cow files you can name as you like and
__activate__ the current one by moving the file to ```live_system.cow```

wandersfalsin:
Could have a rd.live.overlay.cowname to meet this possible demand.

Em seg, 24 de jun de 2019 às 04:46, Marcus Schäfer <notifications@github.com>
escreveu:

schaefi:
>    Could have a rd.live.overlay.cowname to meet this possible demand.

would be possible yes, but we need an extra issue to track it and it
will have low prio at the moment since other issues seems to have a
higher demand

Thanks for your understanding

Issue #764, opened 2018-06-06
Support creating container images using buildah + skopeo


Conan-Kudo:
## Enhancement request description

kiwi currently has the ability to produce two types of container images: base images and layered images. The way this is currently done is by using [`umoci`](https://github.com/openSUSE/umoci) and [`skopeo`](https://github.com/projectatomic/skopeo).

However, there are a couple of issues with this:
* On Fedora, I don't have `umoci`. kiwi's ability to produce container images is basically broken because we only have half of the necessary tools (we have `skopeo`, but not `umoci`).
* Even if I have the ability to use `umoci` for image builds, I'd rather not use it because I'd like to leverage the same underlying tooling for producing images that is used for producing images in OpenShift to simplify the complexity of understanding the image creation process. Fedora uses OpenShift for producing our official container images, and I'd like to be able to leverage similar tooling for producing images using kiwi + OBS.

So, I'd like to see kiwi support [`buildah`](https://github.com/projectatomic/buildah) as an alternative to `umoci` for producing container images.

Today, `buildah` (along with [`podman`](https://github.com/projectatomic/libpod), the container runtime tool), is available in CentOS/RHEL 7, Fedora 26+, openSUSE Tumbleweed, openSUSE Kubic, and (soon) SUSE CaaSP.

## OS and Software information
+ KIWI version: 9.15.1
+ Operating system: Fedora 28
+ OBS version: N/A

davidcassany:
You are right, buildah got into Tumbleweed 3 months ago at the time it did podman and other OCI tools. I understand your preference for buildah as it is already in the base distribution, however we need to keep umoci support and I am wondering what else we gain beyond facilitating container builds in other distros.

Conan-Kudo:
My understanding is that `buildah` is going to be a cornerstone for doing container builds for the wider variety of Kubernetes-centric tools, especially for folks using the CRI-O runtime for Kubernetes/OpenShift.

Also, technically `buildah` doesn't require `skopeo` at all, as it can do its own container fetching for producing layered images, whereas when using `umoci`, you need `skopeo` to fetch the image first.

davidcassany:
After digging a little more about `buildah` I think this out of scope for the time being. Several reasons for that:

* We are not using it in SUSE and there are not plans neither any specific technical motivation for using it as an alternative or substitution of `umoci`.
[...]

PR #894, opened 2018-11-27
Another round of OCI tools refactoring

davidcassany:
In order to provide buildah support some of the logic about
temporary directories for OCI images creation needs to be moved
to the dedicated OCI tool class.

While umoci can operate in any directory and this is passed as an
argument, this is not the case for buildah. In buildah workflow
the storage path of work-in-progress images/containers is not
customizable neither the mountpoint of the container rootfs.

Related with #764

### DIFF
--- a/kiwi/container/oci.py
+++ b/kiwi/container/oci.py
@@ -16,12 +16,10 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
-from tempfile import mkdtemp
 
 # project
 from kiwi.defaults import Defaults
 from kiwi.path import Path
-from kiwi.utils.sync import DataSync
 from kiwi.archive.tar import ArchiveTar
 from kiwi.logger import log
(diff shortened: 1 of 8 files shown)


davidcassany:
@schaefi, just to place some specific example to justify this refactor.

Accessing to the rootfs of an OCI image with `umoci` we need:

`umoci unpack --image oci_layout:base_layer oci_root_dir`

But with buildah something like that is required:

`mountpoint=$(buildah mount mycontainer)`

In umoci we provide the root folder (`oci_root_dir`) but in buildah this is just given to us in stdout. Because of that methods like `oci_tools.base.unpack()` should not include any path. The management of the creation/removal of the rootfs path should be part of the specific class.

schaefi:
Funny my first version of oci_tools class looked like that :)

davidcassany:
I have created and tested both, base and derived images.

schaefi:
:+1:

PR #896, opened 2018-11-27
Add Ubuntu Bionic build tests

davidcassany:
Adding ubuntu tests for iso and vmx image types. The test for oem is there, but it does not resolves due to the already known issue of `sg3-utils` conflicting with `dracut`, there is already a track for this issue in Debian universe [here](https://salsa.debian.org/linux-blocks-team/sg3-utils/merge_requests).


davidcassany:
A build for these test is [here](https://build.opensuse.org/package/show/home:dcassany:branches:Virtualization:Appliances:Images:Testing_x86/test-image-ubuntu)

PR #900, opened 2018-12-10
More clarity on kernel version lookup

### LINKED ISSUE #899: recent get_kernel change from #897
## Problem description

Sorry guys but why got #897 merged the way it was coded ? There are several issues mostly related to code quality.

1. I wrote a good reason in the code why in case of zImage we use vmlinux.gz to read the kernel version. In the change of #897 no such reason to make it clear was given and I also don't understand why it should be needed. I'm running Leap15 and a simple ```kversion /boot/vmlinuz-4.12.14-lp150.12.22-default``` on x86_64 worked without any problem. Thus give me a good reason such that one reading the code understands why we are doing it

2. If we need another exception to the standard way it can be done much simpler
    ```if 'zImage' in kernel_file_to_check or ...``` There is no need for an extra block unless we decide
   this check depends on the architecture. All this was done in the scope of s390 but impacts all other
   archs because a vmlinuz version of the kernel is presented in other archs too but having the safe
   vmlinux.gz version around is not guaranteed. Thus I'm rather concerned than happy with this change

3. The ```os.path.exists(new_file_to_check)``` is superfluous and misleading. We use the rewritten
    target to run kversion on for a reason and call the command with ```raise_on_error=False``` if our 
    expectations are proven wrong I want to see that in the log and not got confused by an extra
    condition based on an assumption.

## Expected behaviour

Don't let us start to again mess up the kernel lookup with code paths nobody understands

I'll come up with a fix as I see it fit and then we can discuss what is wrong or needs further changing

Really wonder why #897 got merged

schaefi:
Lookup of the kernel version is done by directly reading the
kernel image via a small tool named kversion. The scope of the
tool is limited and does not work for e.g kernel images which
contains its own decompressor code. For the special cases we
defined exceptions, one was zImage. The recently added exception
for vmlinuz seemed too intrusive to me and was also not well
documented. This patch tries to clarify and get us back to
explicit and easy to read coding. Fixes #899

### DIFF
--- a/kiwi/system/kernel.py
+++ b/kiwi/system/kernel.py
@@ -16,6 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
+import platform
 from collections import namedtuple
 
 # project
@@ -51,6 +52,7 @@ class Kernel(object):
 
         :rtype: namedtuple
         """
+        arch = platform.machine()
         for kernel_name in self.kernel_names:
             kernel_file = self.root_dir + '/boot/' + kernel_name
             if os.path.exists(kernel_file):
@@ -66,12 +68,17 @@ class Kernel(object):
                     kernel_file_to_check = kernel_file_to_check.replace(
                         'zImage', 'vmlinux'
                     ) + '.gz'
-                if 'vmlinuz' in kernel_file_to_check:
-                    new_file_to_check = kernel_file_to_check.replace(
+                if 'vmlinuz' in kernel_file_to_check and 's390' in arch:
+                    # On s390 it seems the vmlinuz kernel file cannot be used
+                    # to read the kernel version from via the kversion tool.
+                    # Therefore we try to apply the same fallback as in the
+                    # zImage case.
+                    #
+                    # FIXME: The explanation here should be much better
+                    #
+                    kernel_file_to_check = kernel_file_to_check.replace(
                         'vmlinuz', 'vmlinux'
                     ) + '.gz'
-                    if os.path.exists(new_file_to_check):
-                        kernel_file_to_check = new_file_to_check
                 version = Command.run(
                     command=['kversion', kernel_file_to_check],
                     raise_on_error=False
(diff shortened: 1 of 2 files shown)


schaefi:
From @bugfinder 

> so what I'm doing now is, if we look for vmlinuz and also actually have a "vmlinux.gz", then
> we prefer this for the version determination, as the compressed executable binary of the kernel
> will always be safer for the version extraction as any arbitrary boot image format.

I agree but at the time the code was written the presence of a gz compressed kernel binary was not guaranteed. However the presence of some kernel file must be guaranteed otherwise you can't boot. This is why the code used the binary we got and defined exceptions.

I would be more than happy if we can change that in a way that for the version detection we always try to use the compressed binary and only if that does not exist try the arbitrary boot image format

So what is more likely to exist ?

schaefi:
If we agree that vmlinux.gz should take over the highest prio for a version check-up and only if not present we try something else I would be fine and change the code to reflect this. But please don't let us add this as another condition like it is done now

schaefi:
good catch we should reference the real file name and don't trust the symlink.

schaefi:
@davidcassany ok after refactoring if one does a test build with that patch the kversion call should be targeted to a vmlinux-***.gz file.

Should be tested before merge I think ;)

schaefi:
ok tested the build on x86 and still works with the change to:

```
DEBUG: 15:58:18 | EXEC: [kversion /tmp/mytest/build/image-root/boot/vmlinux-4.12.14-lp150.11-default.gz
```

I haven't tested arm or other distro builds but that will uncover when we submit a package to Staging and all the test builds runs.

@davidcassany could you give it a final short look, after that and with no objections I would merge it

schaefi:
of course feedback by @bugfinder is appreciated too :wink:

davidcassany:
@schaefi yes this looks good to me. I just took some time to run also test. So far so good.

schaefi:
Thanks :+1:

bugfinder:
as mentioned before, the problem is that another function already moves the files around,
you might want to get rid of that one first:
kiwi/config/functions.sh: baseCreateCommonKernelFile
does an unconditional:
            if [ -f "vmlinux-${kernel_version}.gz" ];then
                mv "vmlinux-${kernel_version}.gz" vmlinux.gz
            fi
if this "canonicalization" is no longer needed, then this code can go and you 
can rely on the long filenames including version and flavour of the kernel

PR #902, opened 2018-12-12
Make sure manual pages are part of pypi archive

schaefi:
As consequence of the change in the travis setup to make
use of the native 'pages' and 'pypi' providers from travis the
environment for the sdist target has changed. Within the
doc_travis tox env no manual pages was build. In combination
with the 'pypi' provider setup in .travis.yml which uses
skip_cleanup: true, this caused the pypi archive to miss
any manual pages which lead to a runtime error on install
via pip.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -131,6 +131,7 @@ usedevelop = True
 deps = {[testenv:doc]deps}
 changedir=doc
 commands =
+    {[testenv:doc.man]commands}
     travis-sphinx --outdir build_travis build --nowarn --source ./source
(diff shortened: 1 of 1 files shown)


schaefi:
>Nice we divided docs into two different flavours doc and doc_travis so we can be certain not to alter the >RPM tarball with this change, cool.

yahh well, that was also the reason for the bug in the pypi archive :)) but in general yes this was the idea mostly because the rpm package also contains the web docs packaged as pdf which would cause the entire latex stack to be available in travis

schaefi:
ok, I'll try if that fixes it

PR #907, opened 2018-12-18
Fix disk size calculation for VMX

### LINKED ISSUE #904: LVM logical volume with custom mountpoint is broken
## Problem description
I'm using the following LVM setup inside my config.xml for a VMX image based on SLES 12 SP4:

            <systemdisk name="systemVG">
                <volume name="swap" size="8G"/>
                <volume mountpoint="var" name="LVvar" size="115G"/>
                <volume mountpoint="var/opt" name="LVvar_opt" size="10G"/>
            </systemdisk>
            <size unit="G" additive="false">150</size>:

I use this to create a temporary logical volume which I convert to a swap partition on first boot. With Kiwi 7.x this was working fine. I needed to upgrade to Kiwi 9.15.3 due to SLES SP4 being applied. With the same Kiwi config I now get an error doing the build:

[ INFO    ]: 17:03:04 | Creating system image
[ INFO    ]: 17:03:04 | Preparing boot system
[ INFO    ]: 17:03:04 | Creating .profile environment
[ INFO    ]: 17:03:05 | Precalculating required disk size
[ INFO    ]: 17:03:06 | --> system data with filesystem overhead needs 8154 MB
[ INFO    ]: 17:03:06 | --> LVM overhead adding 80 MB
[ ERROR   ]: 17:03:06 | Unexpected error:
  File "/usr/bin/kiwi-ng", line 9, in <module>
    load_entry_point('kiwi==9.17.1', 'console_scripts', 'kiwi-ng-2')()
  File "/usr/lib/python2.7/site-packages/kiwi/kiwi.py", line 64, in main
    App()
  File "/usr/lib/python2.7/site-packages/kiwi/app.py", line 35, in __init__
    task_class().process()
  File "/usr/lib/python2.7/site-packages/kiwi/tasks/system_build.py", line 269, in process
    result = image_builder.create()
  File "/usr/lib/python2.7/site-packages/kiwi/builder/disk.py", line 170, in create
    result = disk.create_disk()
  File "/usr/lib/python2.7/site-packages/kiwi/builder/disk.py", line 221, in create_disk
    disksize_mbytes = self.disk_setup.get_disksize_mbytes()
  File "/usr/lib/python2.7/site-packages/kiwi/storage/setup.py", line 88, in get_disksize_mbytes
    root_filesystem_mbytes
  File "/usr/lib/python2.7/site-packages/kiwi/storage/setup.py", line 294, in _accumulate_volume_size
    data_volume_mbytes.volume[volume.realpath]
KeyError: 'swap' 

## Expected behaviour
A logical volume LVswap should be created with /swap as mountpoint. Documentation about the name attribute on the volume element states:

A required attribute. The name of the volume. If mountpoint is not specified, a directory with the given name will be created by KIWI if it does not already exist inside the root tree. However, if the name contains the KIWI internal path field separator “_”, it is required to specify the path in an additional mount point attribute. The special value @root can be combined with the size or freespace attribute to control the size of the root volume.

This seems to be broken in Kiwi 9.x

Create a Logical volume with a custom mount point

+ KIWI version: 9.15.3 
+ Operating system: SLES 12 SP4
+ OBS version: none

davidcassany:
Disk size calculation must take into account the empty volumes that
are to be mounted in a directory that does not exist in the root tree
otherwise there is KeyError. The result of
storate/setup._calculate_volume_mbytes must be a dict including all
defined volumes.

Fixes #904

### DIFF
--- a/kiwi/storage/setup.py
+++ b/kiwi/storage/setup.py
@@ -340,7 +340,7 @@ class DiskSetup(object):
     def _calculate_volume_mbytes(self):
         """
         Calculate the number of mbytes each volume path currently
-        consumes and also provide a total number of these values
+        consumes and also provide a total number of these values.
         """
         volume_mbytes_type = namedtuple(
             'volume_mbytes_type', ['volume', 'total']
@@ -356,7 +356,9 @@ class DiskSetup(object):
                         volume_size.accumulate_mbyte_file_sizes(),
                         self.filesystem
                     )
-                    volume_total += volume_mbytes[volume.realpath]
+                else:
+                    volume_mbytes[volume.realpath] = 0
+                volume_total += volume_mbytes[volume.realpath]
 
         return volume_mbytes_type(
             volume=volume_mbytes,
--- a/test/data/example_disk_size_volume_config.xml
+++ b/test/data/example_disk_size_volume_config.xml
@@ -25,6 +25,7 @@
                 <volume name="@root" freespace="500M"/>
                 <volume name="etc_volume" mountpoint="etc"/>
                 <volume name="bin_volume" size="all" mountpoint="/usr/bin"/>
+                <volume name="new_volume" mountpoint="newfolder"/>
             </systemdisk>
         </type>
     </preferences>
--- a/test/unit/storage_setup_test.py
+++ b/test/unit/storage_setup_test.py
@@ -180,7 +180,7 @@ class TestDiskSetup(object):
     @patch('os.path.exists')
     @patch('kiwi.logger.log.warning')
     def test_get_disksize_mbytes_volumes(self, mock_log_warn, mock_exists):
-        mock_exists.return_value = True
+        mock_exists.side_effect = lambda path: path != 'root_dir/newfolder'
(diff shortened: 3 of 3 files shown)


davidcassany:
@schaefi, do you think we should create a bugzilla ticket and make maintenance updates for SP4 and SLE15? This is not a critic bug at all, but still a bug. Asking because probably we should add the bugzilla ticket before merging. Also we could just backport it and it would be included in any further update even if we do not include a bugzilla ticket as it will be already part of the code.

I vote for the second.

schaefi:
> I vote for the second

me too

schaefi:
Regarding codacy I can't see a problem ? maybe they were just late :)

davidcassany:
all right, the patch is backported to SUSE/kiwi_sle12@5d0e9d42d1ace34c2963ff3824337008b74ea5d6 and to SUSE/kiwi_sle15@5b07deb738f03978fa6c45e64328b4ba18027658

So now this will also be included in any following SLE12SP4 and SLE15 update.

schaefi:
Thanks, that's great

PR #908, opened 2018-12-20
Use cls in @classmethods instead of self

tomschr:
No related issue, just to improve consistency.

* whenever there is a `@classmethod` decorator, use `cls` instead of `self`. This change makes it more consistent with the Python convention; see https://docs.python.org/3/library/functions.html#classmethod
* No other code changes


schaefi:
Thanks for the cleanup :+1:

PR #911, opened 2019-01-05
Use chkstat to verify and fix file permissions

### LINKED ISSUE #895: Run chkstat to apply configured ownership and permissions
## Problem description

In a kiwi build the newly generated root tree is the result of the installation of a collection of packages and custom archives. The permissions and ownership of the files according to the setup in ```/etc/sysconfig/security``` is not guaranteed by this.

Therefore kiwi should call:

```
chkstat --system --set
```

At the end of the prepare stage to apply the configured security settings

schaefi:
Call chkstat in system mode which reads /etc/sysconfig/security
to determine the configured security level and applies the
appropriate permission definitions from the /etc/permissions*
files. It's possible to provide those files as overlay files
in the image description to apply a certain permission setup
when needed. Otherwise the default setup as provided on the
package level applies. It's required that the image root system
has chkstat installed. If not present KIWI skips this step
and continuous with a warning. This Fixes #895

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -239,6 +239,38 @@ class SystemSetup(object):
         with open(machine_id, 'w'):
             pass
 
+    def setup_permissions(self):
+        """
+        Check and Fix permissions using chkstat
+
+        Call chkstat in system mode which reads /etc/sysconfig/security
+        to determine the configured security level and applies the
+        appropriate permission definitions from the /etc/permissions*
+        files. It's possible to provide those files as overlay files
+        in the image description to apply a certain permission setup
+        when needed. Otherwise the default setup as provided on the
+        package level applies.
+
+        It's required that the image root system has chkstat installed.
+        If not present KIWI skips this step and continuous with a
+        warning.
+        """
+        chkstat_search_env = {
+            'PATH': os.sep.join([self.root_dir, 'usr', 'bin'])
+        }
+        chkstat = Path.which(
+            'chkstat', custom_env=chkstat_search_env, access_mode=os.X_OK
+        )
+        if chkstat:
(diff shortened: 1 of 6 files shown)


kiwi/system/setup.py
```diff
@@ -239,6 +239,38 @@ def setup_machine_id(self):
...
+        package level applies.
+
+        It's required that the image root system has chkstat installed.
+        If not present KIWI skips this step and continuous with a
+        warning.
+        """
+        chkstat_search_env = {
+            'PATH': os.sep.join([self.root_dir, 'usr', 'bin'])
+        }
+        chkstat = Path.which(
```
davidcassany: Just in case you missed it, this code could benefit from the KiwiCommandNotFound exception logic, it is raised exactly with this same path check. So this could be coded as:

```python
try:
  log.info('Check/Fix File Permissions')
  Command.run(
    ['chroot', self.root_dir, 'chkstat', '--system', '--set']
  )
except KiwiCommandNotFound:
  log.warning(
    'chkstat not found in image. File Permissions Check skipped'
  )
```

Anyway I am fine with the current implementation.

schaefi: This is not the same. KiwiCommandNotFound is thrown if the command which is expected to be the first element in the list, is not found. In this case this applies to the ```chroot``` command. But my code explicitly tests if the ```chkstat``` command exists in the image root dir.

chkstat supports the --root flag but that would call chkstat from the host and I explicitly wanted to call the program from the image root because I don't trust the host installed chkstat to do the right thing if it's not from the same distro as the image

davidcassany: You are absolutely right sorry, I answered without noticing it was a chroot command.

PR #912, opened 2019-01-05
Fix use of SysConfig objects

### LINKED ISSUE #910: The btrfs volume manager is not working correctly
## Problem description
kiwi throws an exception while getting the sysconfig file for snapper.
```
[  219s] [ ERROR   ]: 13:06:39 | Unexpected error:
[  219s] Traceback (most recent call last):
[  219s]   File "/usr/bin/kiwi", line 11, in <module>
[  219s]     load_entry_point('kiwi==9.17.1', 'console_scripts', 'kiwi-ng-3')()
[  219s]   File "/usr/lib/python3.6/site-packages/kiwi/kiwi.py", line 64, in main
[  219s]     App()
[  219s]   File "/usr/lib/python3.6/site-packages/kiwi/app.py", line 35, in __init__
[  219s]     task_class().process()
[  219s]   File "/usr/lib/python3.6/site-packages/kiwi/tasks/system_build.py", line 269, in process
[  219s]     result = image_builder.create()
[  219s]   File "/usr/lib/python3.6/site-packages/kiwi/builder/disk.py", line 170, in create
[  219s]     result = disk.create_disk()
[  219s]   File "/usr/lib/python3.6/site-packages/kiwi/builder/disk.py", line 398, in create_disk
[  219s]     self._get_exclude_list_for_root_data_sync(device_map)
[  219s]   File "/usr/lib/python3.6/site-packages/kiwi/volume_manager/btrfs.py", line 315, in sync_data
[  219s]     self._create_snapper_quota_configuration()
[  219s]   File "/usr/lib/python3.6/site-packages/kiwi/volume_manager/btrfs.py", line 371, in _create_snapper_quota_configuration
[  219s]     config_file = self._set_snapper_sysconfig_file(root_path)
[  219s]   File "/usr/lib/python3.6/site-packages/kiwi/volume_manager/btrfs.py", line 384, in _set_snapper_sysconfig_file
[  219s]     if not sysconf_file.get('SNAPPER_CONFIGS') or \
[  219s] AttributeError: 'SysConfig' object has no attribute 'get'
```

## Expected behaviour
The check if the config file is present returns true or false.

All the image builds in devel:kubic:images:experimental on OBS are failing.

+ KIWI version: python3-kiwi-9.17.1-1.1
+ OBS version: Head

schaefi:
objects of that class does not provide a get method but overload
the bracket [] operator which itself accesses elements of a
standard dictionary using the get() method. Thus it's safe to
use the bracket operator. This patch fixes the one place where
the get() method was used. This Fixes #910

### DIFF
--- a/kiwi/utils/sysconfig.py
+++ b/kiwi/utils/sysconfig.py
@@ -41,6 +41,9 @@ class SysConfig(object):
     def __contains__(self, key):
         return key in self.data_dict
 
+    def get(self, key):
+        return self.data_dict.get(key)
+
     def write(self):
         """
         Write back source file with changed content but in same order
--- a/test/unit/utils_sysconfig_test.py
+++ b/test/unit/utils_sysconfig_test.py
@@ -21,6 +21,7 @@ class TestSysConfig(object):
 
     def test_get_item(self):
         assert self.sysconfig['name'] == ' "Marcus"'
+        assert self.sysconfig.get('name') == ' "Marcus"'
 
     def test_set_item_existing(self):
         self.sysconfig['name'] = 'Bob'


kiwi/utils/sysconfig.py
```diff
@@ -36,7 +36,7 @@ def __setitem__(self, key, value):
         self.data_dict[key] = value
 
     def __getitem__(self, key):
-        return self.data_dict[key]
+        return self.data_dict.get(key)
```
davidcassany: I really doubt it can create any issue, however this looks misleading to me. If I run something like `sysconfig_file['nonexistingKey']` I'd expect it to raise a KeyError Exception, however with this code it returns `None`. I think it is good to reuse the dictionary `__getitiem__` here, this way we get an analog behavior.

kiwi/volume_manager/btrfs.py
```diff
@@ -381,7 +381,7 @@ def _set_snapper_sysconfig_file(cls, root_path):
         sysconf_file = SysConfig(
             os.sep.join([root_path, 'etc/sysconfig/snapper'])
         )
-        if not sysconf_file.get('SNAPPER_CONFIGS') or \
+        if not sysconf_file['SNAPPER_CONFIGS'] or \
```
davidcassany: As said above I think this should be coded either using the `get` method (for that needs to be implemented in Sysconfig) or just verify key existance with something like:

```python
if 'SNAPPER_CONFIGS' not in sysconf_fikiwi/volume_manager/btrfs.pyle or ...
```

kiwi/utils/sysconfig.py
schaefi: ok, I'm gonna change things to be consistent with standard use of dicts

schaefi:
> I am really surprised this issues is not raising in our unit tests. 

Because the unit tests did not use the get method because it does not exist and also was not intended to exist. The code in the btrfs volume manager implicitly assumed there is a get method which it should not have done in the first place. Now that there is a get method I also updated the unit test

schaefi:
The unit test for the btrfs volume manager mocked the entire SysConfig class which is ok for a unit test at this point. This is the reason why no error was thrown in the btrfs unit testing.

PR #913, opened 2019-01-10
Make result compression in the bundler optional

### LINKED ISSUE #901: Some way to disable xz compression for vmdk
## Problem description

Currently since #869 vmdk image formats, among other, are always compressed in OBS builds. This is causing issues in openQA where they used to grab directly and boot it without preprocessing it.

Apparently it is not simple to add some pre process steps on openQA. This is something they are already doing for raspberry pi images it is really slowing down and over killing the tests.

Now the compression started to happen on TW for x86_64 because of #869 and the issue is likely to spread to many images.

Probably we could think about a way to configure that with the runtime yaml config file. 

+ KIWI version: master or Tumbleweed version
+ Operating system: Tumbleweed

schaefi:
Calling kiwi result bundle will take the image build results
and bundle the relevant image files according to their image
type. Depending on the result configuration this could instruct
the bundler to compress one or more files from the result.
By default this compression is switched off in the bundler but
can be activated to save storage space and speedup download
of the image with the following runtime configuration:

```yaml
bundle:
  - compress: true|false
```

If compression is activated the result image has to be
uncompressed before it can be used.

This Fixes #901

### DIFF
--- a/kiwi/tasks/result_bundle.py
+++ b/kiwi/tasks/result_bundle.py
@@ -123,45 +123,45 @@ class ResultBundleTask(CliTask):
                 bundle_file = ''.join(
                     [bundle_directory, '/', bundle_file_basename]
                 )
-                checksum_file = ''.join(
-                    [bundle_directory, '/', bundle_file_basename, '.sha256']
-                )
                 Command.run(
                     [
                         'cp', result_file.filename, bundle_file
                     ]
                 )
-                if result_file.compress:
+                if self.runtime_config.is_bundle_compression_requested() and \
+                   result_file.compress:
                     log.info('--> XZ compressing')
                     compress = Compress(bundle_file)
                     compress.xz(self.runtime_config.get_xz_options())
                     bundle_file = compress.compressed_filename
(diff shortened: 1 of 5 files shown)


kiwi/tasks/result_bundle.py
```diff
@@ -123,45 +123,45 @@ def process(self):
...
-                    [bundle_directory, '/', bundle_file_basename, '.sha256']
-                )
                 Command.run(
                     [
                         'cp', result_file.filename, bundle_file
                     ]
                 )
-                if result_file.compress:
+                if self.runtime_config.is_bundle_compression_requested() and \
+                   result_file.compress:
```
davidcassany: Is it me or there is an strange indentation here? Anyway flake does not complain, so I guess this is _pythonic_ :smile:

schaefi: Yes it's strange but pythonic. The long line break in conditional statements always have the problem that the indented block has to belong to the condition. The indentation has to be in between of the condition and the block it belongs to if the condition line has a line break. Very long discussion about alternative style and pros and cons here: #861

PR #920, opened 2019-01-24
Fixed file name of completion file

schaefi:
The bash completion file must match one of the alternatives links.
Otherwise the bash completion mechanism will not match. kiwi-ng is
the unique alternative link name compared to the still present
legacy kiwi version and should be used preferably


schaefi:
Found this by installing a new system and realized the completion was non functional. I moved the file in /etc/bash_completion.d and after that all worked as expected

PR #921, opened 2019-01-24
Fix documentation howto delete a package

schaefi:
The documentation described an invalid XML grammar


schaefi:
LOL: we need more people like defolos reading our docs :-)

PR #922, opened 2019-01-24
Fixed location of grub unicode font file

schaefi:
grub2 is expecting the unicode font below the fonts
directory in the /boot/grub*/ depending on how the
distribution installs grub2.

This Fixes bsc#1119416


schaefi:
@davidcassany I haven't done a testbuild but will do asap. If you have a chance to test it too it would be great. Thanks

davidcassany:
@schaefi sure I setting a test for it now

PR #923, opened 2019-01-25
Fixed Xen guest detection

schaefi:
We only support Xen setup e.g in the Amazon Cloud for the
x86_64 architecture. On other archs e.g aarch64 Xen does not
exist. The method in kiwi which checks for the guest setup
should take the architecture into account

### DIFF
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -541,8 +541,10 @@ class XMLState(object):
     def is_xen_guest(self):
         """
         Check if build type setup specifies a Xen Guest (domX)
-        The check is based on the firmware and xen_loader configuration
-        values:
+        The check is based on the architecture, the firmware and
+        xen_loader configuration values:
+
+        * We only support Xen setup on the x86_64 architecture
 
         * Firmware pointing to ec2 means the image is targeted to run
           in Amazon EC2 which is a Xen guest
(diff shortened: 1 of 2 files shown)


schaefi:
Another one which touches sensitive code parts. I need this to build x86 and arm images from the same description and for Amazon EC2. In general the xen guest check should take the architecture into account. Xen imho does not really exist on any other arch

Thoughts ?

kiwi/xml_state.py
```diff
@@ -554,6 +556,9 @@ def is_xen_guest(self):
 
         :rtype: bool
         """
+        if self.host_architecture != 'x86_64':
+            # All of Xen only really exists on x86_64
+            return False
         firmware = self.build_type.get_firmware()
         machine_section = self.get_build_type_machine_section()
         if firmware and firmware in Defaults.get_ec2_capable_firmware_names():
```
davidcassany: Here is where I would have added the arch check. I but probably is just simpler to avoid any XEN support for non x86 archs if no one is using it or if it doesn't exist a really working tool chain for that yet.

schaefi: This is exactly what I want "No Xen for any other arch than x86_64" You are right there are Xen stuff for other archs but honestly I don't care. I see Xen at a dead end. The only cloud vendor which still uses it is Amazon in EC2 and they also start to move to kvm. They have done the move from Xen para virtualization to Xen host virtualization (full virtualization) in almost all regions.

But I should change the comment line here

Issue #771, opened 2018-07-02
Support setting %_install_langs in the config.xml


Conan-Kudo:
## Problem description
I'd like for RPM to be configured with specific `%_install_langs` settings through the entire image build to minimize stuff being installed in the image.

This is especially necessary on target systems that don't have split langpacks and stuff like that, such as CentOS/RHEL, Mageia, and older openSUSE.

## Expected behavior
Some tag, like `<rpm-install-langs>` (similar to the other `rpm-*` tags), would let me set the value of `%_install_langs` to be used, like `C:C.UTF-8:en:en_US:en_US.UTF-8`

This would effectively do something like this:
`echo "%_install_langs C:C.UTF-8:en:en_US:en_US.UTF-8" > $CHROOT/etc/rpm/macros.d/macros.kiwi-langs`

## OS and Software information
+ KIWI version: 9.16.0
+ Operating system: Fedora 28
+ OBS version: N/A

davidcassany:
@Conan-Kudo looks like a reasonable feature, just wondering if there is something similar for DEB packages or APT manager (I'd like to avoid further `rpm-*` tags in the schema), I did not find anything at a glance, are you aware of something similar? Also would be nice to handle that at package manager level, at zypper, dnf and yum configuration and avoid playing around with RPM macros.

Conan-Kudo:
Unfortunately, this is an RPM specific feature, and while YUM has a config option for it, none of the others do.

@dmach Any chance of seeing the YUM option ported to DNF?

davidcassany:
I have been thinking and looking at it again and I don't think we achieve it in any reasonable way. This is a similar problem we had with #133 for yum, the only way of doing it is using rpm macro files. The problem I see here is that for the bootstrap phase the macros used and the package manger used is the one installed within the host, thus in order to get that feature during the bootstrap phase (which I consider a must) we need to modify the host system macros. Of course we won't be hacking host system wide configuration files, too bad.

I could not find a way to define for zypper, yum or dnf a custom rpm macro file location. If we could create runtime macro files and make the package managers make use of it for the bootstrap phase that would be enough. 

So, currently, it seams this feature can only be implemented with YUM and I'd say this is not enough for us to implement it. Don't get me wrong, I understand that YUM will be replaced by DNF at some point, thus adding specific code for YUM does not have much sense IMHO.

@Conan-Kudo, @schaefi any further thoughts?

dmach:
[...]

PR #931, opened 2019-02-04
Handle location of the rpm DB on the macro level

schaefi:
The location of the rpm database is no longer a standard
path one can trust. Some distributions put it to /var/lib
others to /usr/lib. This introduces the problem of dealing
with different locations between the bootstrapping (host rpm)
phase and the image installation (image rpm) phase.

This commit implements a solution based on an intermediate
rpm database configuration. KIWI creates the macros.kiwi file
inside of the image root which is read by any call of rpm
in the inner and outer system. During bootstrap phase the
rpm dbpath from the host system is used and later in the
install phase the dbpath from the rpm package as it was
installed by the target image distribution is used. In case
of a dbpath difference the database is automatically moved
to the new location by setting the _dbpath_rebuild macro
to the correct location. At the end the custom KIWI macro
is deleted.

As this process allows custom macro defintions during the
KIWI run it also serves as the base for a solution to
Issue #771 which will be done in a follow up request to
this commit.

Also the workaround for bsc#1112357 which uses a static
dbpath to store an optionally given signing key will be
addressed with this commit. The macro setup happens before
the import_trusted_keys method which makes any specification
for a strict dbpath obsolete.

### DIFF
--- a/kiwi/repository/yum.py
+++ b/kiwi/repository/yum.py
@@ -23,6 +23,7 @@ from tempfile import NamedTemporaryFile
 from kiwi.logger import log
 from kiwi.command import Command
 from kiwi.repository.base import RepositoryBase
+from kiwi.utils.rpm_database import RpmDataBase
 from kiwi.path import Path
 
 
@@ -89,6 +90,14 @@ class RepositoryYum(RepositoryBase):
         self._create_runtime_config_parser()
         self._create_runtime_plugin_config_parser()
         self._write_runtime_config()
+
+    def setup_package_database_configuration(self):
+        """
+        Make sure for bootstrapping the rpm database location
+        matches the host rpm database setup
+        """
+        rpmdb = RpmDataBase(self.root_dir)
+        rpmdb.set_database_to_host_path()
 
     def use_default_location(self):
         """
--- a/kiwi/repository/zypper.py
+++ b/kiwi/repository/zypper.py
@@ -24,7 +24,7 @@ from kiwi.command import Command
 from kiwi.repository.base import RepositoryBase
 from kiwi.system.uri import Uri
 from kiwi.path import Path
-from kiwi.defaults import Defaults
+from kiwi.utils.rpm_database import RpmDataBase
 
 from kiwi.exceptions import (
     KiwiRepoTypeUnknown
@@ -144,6 +144,14 @@ class RepositoryZypper(RepositoryBase):
 
         self._write_runtime_config()
 
+    def setup_package_database_configuration(self):
+        """
+        Make sure for bootstrapping the rpm database location
+        matches the host rpm database setup
+        """
+        rpmdb = RpmDataBase(self.root_dir)
+        rpmdb.set_database_to_host_path()
+
     def use_default_location(self):
         """
         Setup zypper repository operations to store all data
@@ -254,10 +262,8 @@ class RepositoryZypper(RepositoryBase):
         :param list signing_keys: list of the key files to import
         """
         for key in signing_keys:
-            # Including --dbpath flag is a workaround for bsc#1112357
             Command.run([
-                'rpm', '--root', self.root_dir, '--import',
-                key, '--dbpath', Defaults.get_default_rpmdb_path()
+                'rpm', '--root', self.root_dir, '--import', key
             ])
 
     def delete_repo(self, name):
--- a/kiwi/utils/rpm.py
+++ b/kiwi/utils/rpm.py
@@ -1,0 +1,127 @@
+# Copyright (c) 2019 SUSE Linux GmbH.  All rights reserved.
+#
+# This file is part of kiwi.
+#
+# kiwi is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# kiwi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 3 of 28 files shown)


schaefi:
@davidcassany as one side effect this pull request also activates the dump reload support for dnf. The adaption of the dbpath is part of that function and so I used it for dnf too

schaefi:
@davidcassany please also note, it does not fix the dnf issue you will still have on the verification level while in bootstrap. That is a bug in dnf and fixed in the latest version which @Conan-Kudo will update for us

@davidcassany another thing is a bit of code duplication in package_manager/dnf|zypper. I kept it for the moment since I don't have a good solution at the moment. dnf and zypper are both rpm based and thus the actions on the macro level and the dump/reload are the same

schaefi:
Hmm, ```You are using Python 2.7.14, which will no longer be supported in pytest 5.0```

Time to delete support for 2.7 ?

Conan-Kudo:
I'm not building kiwi for Python 2 on Fedora 29+. Once Fedora 28 is EOL, I'd be okay with dropping Python 2 entirely.

Conan-Kudo:
@schaefi We should consider fixing Travis CI to not use the Travis provided stuff, and switch to Fedora and openSUSE environments.

schaefi:
> @schaefi We should consider fixing Travis CI to not use the Travis provided stuff, and switch to Fedora and openSUSE environments. That'll make the warning go away.

Would be a good idea, on https://docs.travis-ci.com/user/reference/overview/#virtualization-environments I don't see other environments. So the only way is imho the docker container stuff right ? Which adds another layer... but probably worth the effort

davidcassany:
> Would be a good idea, on https://docs.travis-ci.com/user/reference/overview/#virtualization-environments I don't see other environments. So the only way is imho the docker container stuff right ? Which adds another layer... but probably worth the effort

Yes I believe using a docker container is the way to go. Ideally we should be capable to use a container coming from registry.opensuse.org, so we could maintain a derived container based on official tumbleweed base container. This way we would have continuous rebuilds of it and keep it automatically aligned with Tumbleweed. I could explore the capabilities of Travis to pull from registry.opensuse.org and prepare an image with the required env for KIWI's Tox. Does it sound reasonable?

kiwi/repository/zypper.py
```diff
@@ -256,8 +268,7 @@ def import_trusted_keys(self, signing_keys):
         for key in signing_keys:
             # Including --dbpath flag is a workaround for bsc#1112357
```
davidcassany: I believe we should delete this comment, shouldn't we? There is no `--dbpath` anymore.

schaefi:
@davidcassany ok I'm finished now, could have another look ? Thanks

kiwi/repository/yum.py
```diff
@@ -90,6 +90,16 @@ def post_init(self, custom_args=None):
...
         self._write_runtime_config()
 
+    def setup_package_database_configuration(self):
+        """
+        Setup package database configuration
+
+        For the supported RHEL versions there is no rpm database
+        setup required. We support building RHEL images on RHEL
+        host systems only at the moment.
+        """
```
Conan-Kudo: We do? Why?

kiwi/utils/rpm_database.py
```diff
@@ -0,0 +1,139 @@
...
+        Supported target rpm database db versions are
+
+        * 45 (db45_load)
+        * 48 (db48_load)
+
+        :param str version: target rpm db version
+        """
+        db_load_for_version = {
+            45: 'db45_load',
+            48: 'db48_load'
```
Conan-Kudo: I know this is technically unchanged, but this does't work for non-SUSE distributions, because everyone else moved past db4. Both RHEL 7 and Fedora use db5 (as does Mageia). OpenMandriva is using db6, and so on.

Why do we do this anyway?

kiwi/repository/yum.py
schaefi: because I thought dnf will be the successor, not correct ?

schaefi: ok, I've added the macro setup there too

kiwi/utils/rpm_database.py
schaefi: Thanks for the discussion on the matrix chat. I deleted that reload code as it is not needed anymore

schaefi:
@Conan-Kudo good to go ?

kiwi/utils/rpm.py
```diff
@@ -0,0 +1,127 @@
...
+        self.root_dir = root_dir
+        self.config = []
+        self.custom_config = []
+        self.macro_path = os.sep.join(
+            [
+                self.root_dir or '', Defaults.get_custom_rpmdb_macros_path()
+            ]
+        )
+        self.macro_file = os.sep.join(
+            [self.macro_path, 'macros.kiwi']
```
Conan-Kudo: I'd probably make this filename more specific: `macros.kiwi-rpmdb-cfg` or something. Just so that it's clear what it's for.

schaefi: we will use this macro file also for fixing the lang issue. So I suggest to be a bit more generic. How about:

* macros.kiwi-bootstrap-config

Thoughts ?

davidcassany: I prefer generic also because this can be later used for further purposes.

Also thinking more about this macro feature I believe it could be an interesting feature to allow some customization over it, I mean let user append some stuff in it. The fact that this enables setting custom macros for bootstrap is valuable. I struggled with it while trying to slim down container images using tricks similar to the lang issue (I used some macros to no install in `/usr/share/doc` and similar artifacts). By then I'd love to have a way to customize macros during bootstrap.

Anyway the point is that I think this macro strategy might have further use cases in the future, so I would not limit it to the db path issue.

schaefi: @Conan-Kudo do you agree on the proposed name ?

Conan-Kudo: The install-langs macro may actually probably stick around post-build so that it stays in effect.

PR #933, opened 2019-02-06
Fix Failed to cache rpm database on zypper addrepo

schaefi:
Occasionally zypper fails when adding the repo with the
rpm error message 'Failed to cache rpm database'. I was
not able to find out why this happens and I also could
not find a way to reproduce it safely. However this
commit adds a workaround that seems to fix the issue
when it happens. If the first call of zypper addrepo
fails kiwi now issues the exact same call again and
only if that fails too an exception is thrown

### DIFF
--- a/kiwi/exceptions.py
+++ b/kiwi/exceptions.py
@@ -522,13 +522,6 @@ class KiwiRepositorySetupError(KiwiError):
     """
 
 
-class KiwiRepoTypeUnknown(KiwiError):
-    """
-    Exception raised if an unsupported repository type is specified
-    for the corresponding package manager.
-    """
-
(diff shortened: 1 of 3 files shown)


schaefi:
I had the same feeling and thought ok, this is adding three lines of rescue code which probably will stick there forever but that would not hurt too much. The problem that you don't know if the build starts is however a negative factor for the stability of the builder. I bet people facing this problem will solve it by a shell script recalling kiwi or passing the --clear-cache option which seems to have a similar effect. In the end none of those solutions is somehow wanted from my perspective and so I would be willing to maintain that code forever and a day ;)

PR #936, opened 2019-02-07
Added support for %_install_langs rpm macro

### LINKED ISSUE #771: Support setting %_install_langs in the config.xml

schaefi:
During the image build the custom rpm macro %_install_langs
is configured with the <locale> setup from the KIWI XML
description. This allows to filter language specific packages
on the rpm level and Fixes #771

### DIFF
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -99,10 +99,15 @@ class SystemPrepare(object):
         repository_sections = \
             self.xml_state.get_repository_sections_used_for_build()
         package_manager = self.xml_state.get_package_manager()
+        locale_list = self.xml_state.get_locale()
         if self.xml_state.get_rpm_check_signatures():
             repository_options.append('check_signatures')
         if self.xml_state.get_rpm_excludedocs():
             repository_options.append('exclude_docs')
+        if locale_list:
+            repository_options.append(
+                '_install_langs%{0}'.format(':'.join(locale_list))
+            )
         repo = Repository(
             self.root_bind, package_manager, repository_options
         )
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -99,14 +99,14 @@ class SystemPrepare(object):
         repository_sections = \
             self.xml_state.get_repository_sections_used_for_build()
         package_manager = self.xml_state.get_package_manager()
-        locale_list = self.xml_state.get_locale()
+        rpm_locale_list = self.xml_state.get_rpm_locale()
         if self.xml_state.get_rpm_check_signatures():
             repository_options.append('check_signatures')
         if self.xml_state.get_rpm_excludedocs():
             repository_options.append('exclude_docs')
-        if locale_list:
+        if rpm_locale_list:
             repository_options.append(
-                '_install_langs%{0}'.format(':'.join(locale_list))
+                '_install_langs%{0}'.format(':'.join(rpm_locale_list))
             )
         repo = Repository(
             self.root_bind, package_manager, repository_options
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -1253,6 +1253,18 @@ class Defaults(object):
         return 'macros.kiwi-bootstrap-config'
 
     @classmethod
+    def get_custom_rpm_image_macro_name(cls):
+        """
+        Returns the rpm image macro file name created
+        in the custom rpm macros path
+
+        :return: filename
+
+        :rtype: str
+        """
+        return 'macros.kiwi-image-config'
+
+    @classmethod
     def get_default_packager_tool(cls, package_manager):
         """
         Provides the packager tool according to the package manager
--- a/kiwi/repository/dnf.py
+++ b/kiwi/repository/dnf.py
@@ -21,6 +21,7 @@ from six.moves.configparser import ConfigParser
 from tempfile import NamedTemporaryFile
 
 # project
+from kiwi.defaults import Defaults
 from kiwi.repository.base import RepositoryBase
 from kiwi.path import Path
 from kiwi.command import Command
@@ -62,6 +63,12 @@ class RepositoryDnf(RepositoryBase):
         else:
             self.gpg_check = '0'
 
+        self.locale = list(
+            item for item in self.custom_args if '_install_langs' in item
+        )
+        if self.locale:
+            self.custom_args.remove(self.locale[0])
+
         self.repo_names = []
 
         # dnf support is based on creating repo files which contains
@@ -94,11 +101,21 @@ class RepositoryDnf(RepositoryBase):
 
     def setup_package_database_configuration(self):
         """
-        Make sure for bootstrapping the rpm database location
-        matches the host rpm database setup
+        Setup rpm macros for bootstrapping and image building
+
+        1. Create the rpm image macro which persists during the build
+        2. Create the rpm bootstrap macro to make sure for bootstrapping
+           the rpm database location matches the host rpm database setup.
+           This macro only persists during the bootstrap phase
         """
-        rpmdb = RpmDataBase(self.root_dir)
-        rpmdb.set_database_to_host_path()
+        rpmdb = RpmDataBase(
+            self.root_dir, Defaults.get_custom_rpm_image_macro_name()
+        )
+        if self.locale:
+            rpmdb.set_macro_from_string(self.locale[0])
+        rpmdb.write_config()
+
+        RpmDataBase(self.root_dir).set_database_to_host_path()
 
     def use_default_location(self):
         """
--- a/kiwi/repository/yum.py
+++ b/kiwi/repository/yum.py
@@ -20,6 +20,7 @@ from six.moves.configparser import ConfigParser
 from tempfile import NamedTemporaryFile
 
 # project
+from kiwi.defaults import Defaults
 from kiwi.logger import log
 from kiwi.command import Command
 from kiwi.repository.base import RepositoryBase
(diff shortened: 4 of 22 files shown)


schaefi:
> After thinking about it I see your point in the implementation, if the user already configures a locale it
> makes sense to stick to that.

Right this was my motivation to provide not yet another XML attribute to configure the locale. Users will not understand what this rpm macro has to do with it. In addition the locale names we require from the ```<locale>``` section are 100% compatible with what rpm expects. So I'm not in favor of an extra attribute

> How this macro behave if the configured locale is not installed into the system (e.g. glibc-locale is not
> installed) or just setting a non existing locale.

If you look at what we do on the system configuration level regarding the locale, the same concern can be raised there too. So we take the ```locale``` information and either call ```systemd-firstboot --locale``` which messes with /etc/locale.conf or we setup /etc/sysconfig/language. In both cases if the locale is not installed your system will respond with the configured value in the ```locale``` command but any program in the system really using that locale will be in trouble. This is unrelated to the rpm macro

I expect the additional rpm macro to be evaluated by rpm only. I think that language information is not used in the context of the locale but only by rpm to make a selection. The selection will be based on the filelists installed by the packages and leading to prevent the installation of files connected to a locale setting. Having that specified in one ```<locale>``` section seems consistent to me. However specifying that as two attributes opens the possibility for users to create inconsistent locales. E.g telling rpm to only install stuff that matches en_US and setting the locale to jp_JP will probably lead to the install of glibc-locale but not with a Japanese locale installed. Therefore I still vote to not separate those things

> Shouldn't the C or POSIX locales be also included as default as suggested by @Conan-Kudo, I
> remember we added POSIX as a valid value for locale as not setting this locale in environments
> without the glibc-locale (like containers) was causing issues.

This is a good point, we can force ```POSIX``` to be in the list in any case

schaefi:
What I also wanted to mention is that it's not kiwi's responsibility to care for the system integrity in terms of configuration and their required components. If you build an image for aarch64 and set grub2 as the bootloader you also have to make to sure to install grub2-arm64-efi otherwise you are missing grub modules. The same applies to the locale. If you specify ja_JP and don't have glibc-locale installed I don't consider it kiwi's fault :)

schaefi:
Because packaging has no standard across distributions we can't care for those type of integrity

schaefi:
We also provide the config,sh method baseStripLocales() in kiwi/config/functions.sh which I expect to result in the same compared to having the locale set on the macro level for rpm when it comes to the install of glibc-locale. So users can already rip those things out of their system. I agree to your point that the change in the install behavior of rpm is different according to the macro setup and that this is kind of hidden but this is the nature of that macro system in rpm. I don't see how we could make that more explicit

davidcassany:
>  However specifying that as two attributes opens the possibility for users to create inconsistent locales. E.g telling rpm to only install stuff that matches en_US and setting the locale to jp_JP will probably lead to the install of glibc-locale but not with a Japanese locale installed. Therefore I still vote to not separate those things

well I imagined the install_lang element/attribute as a boolean that creates or not the macro according to what is set in locale element. Anyway, I buy your argumentation that it is not our fault if the user sets inconsistent config.xml files, neither if there are packages out there messing around with the lang RPM macros in some weird and unexpected way.

schaefi:
> well I imagined the install_lang element/attribute as a boolean that creates or not the macro according
> to what is set in locale element

But that sounds inconsistent to me. If I don't specify ```<locale>``` no macro will be set. The default that now applies is ```all``` That's what you get by the rpm default macro installation. So this already assumes all locales exists. If I now explicitly set a ```<locale>``` list why do I need to explicitly enable it ? Or in other words what would mean a ```<locale>``` setup with install_lang="false" ?

The ```<locale>``` element was used from the beginning on to reduce the system to that locale and it's required components. That's the reason why we also have that strip methods in the config.sh space.

If I just want to have all locales and a specific one configured I wouldn't specify ```<locale>``` but depending on the system write some setup code in config.sh like it's used for other system configuration tasks

Maybe the section name 'locale' doesn't express this very well

davidcassany:
I think we have a different understanding of the use case of the `<locale>`. I'd use it to set the language of the packages that provide translated user feedback (regardless if they are split or not into different packages).  Thus I just do not care about size in that case. In my head the scenario I had was something like having set the system to some language (e.g. jp_JP) to get as much as possible messages translated to that language. I am not sure if this could happen, but probably there are out there applications relaying on C.UTF-8 or, even worse, on en_US.UTF-8. But yes, as you already said, this should not be worried about this sort of inconsistencies, as they are mostly issues from other tools.

I am fine not having any additional XML field to set that. I was just trying to expose my thinking :)

schaefi:
> I think we have a different understanding of the use case of the ```<locale>```

I think so too and maybe that confuses other people too. When I think about that setting it's targeted to create a system that only supports that locales, including package files, translations, the system setup and the low level locale data. It's not meant to setup the language of a system... which I think could be interpreted also very differently.

So how are we going to solve this. The easy approach is to create a new section for any macro we allow to configure. To be honest I don't like it :) Having that would require a runtime check to check if the locale setup correlates with the ```<rpm-install-langs>....</rpm-install-langs>```

Things would then look like this

```xml
<locale>en_US</locale>
<rpm-install-langs>POSIX:en_US:de_DE</rpm-install-langs>
```
This looks very weird to me but would be the most explicit way to configure custom macros

@Conan-Kudo @davidcassany 
Thougths ?

Conan-Kudo:
@schaefi The original reason I requested an extra property to set this is precisely because of this distinction. Unfortunately, this is the only reasonable way to do this to properly convey the semantics.

Moreover, just because the _system locale_ is set one way doesn't mean that locale filtering is needed or wanted. Keep in mind that Fedora and openSUSE media are created to have the locale set to `en_US`, but all the languages are there for the purpose of easily switching to the user's preferred locale.

As for mandating locales, I'm not sure if it makes sense to mandate at least `POSIX` or `C`/`C.UTF-8` are set if the tag is used. It's not a _bad_ idea, but I don't know if we want to reasonably enforce that...

davidcassany:
Having two different places to set locales is just worse IMHO I also don't like that. Probably it is just fine keep the current PR implementation and just make more explicit the locale meaning in the schema docs, at least exposing this sets the install_lang macro preventing unmatching lang files to be installed.

By the way, just for curiosity, is this macro preventing the installation of unneeded files of `glibc-locale`? have you tested that? I am curious because glibc-locale weights up to 122MB in Leap and that could potentially reduce it's size to 3MB as `kubic-local-archive` (only contains C.UTF-8 and en_US.UTF-8).

kiwi/system/prepare.py
```diff
@@ -99,10 +99,15 @@ def setup_repositories(self, clear_cache=False, signing_keys=None):
...
             self.xml_state.get_repository_sections_used_for_build()
         package_manager = self.xml_state.get_package_manager()
+        locale_list = self.xml_state.get_locale()
         if self.xml_state.get_rpm_check_signatures():
             repository_options.append('check_signatures')
         if self.xml_state.get_rpm_excludedocs():
             repository_options.append('exclude_docs')
+        if locale_list:
+            repository_options.append(
+                '_install_langs%{0}'.format(':'.join(locale_list))
```
Conan-Kudo: This is wrong. It's supposed to be `%_install_langs <lang>:<lang>...`

schaefi: nope, not wrong. I just pass the information in that format to the Repository class. If you look further you find this:  https://github.com/SUSE/kiwi/pull/936/files#diff-55cc4623f69c01acca0ea339e5b6d70aR58

schaefi: so it gets resolved properly in the end, also the unit test shows that :) and I have tested it in real live runs a hundred times the last days :)

Conan-Kudo:
@davidcassany My understanding is that the `glibc-locale` package is not affected by it, but `glibc-locale-base` may be. The locale archive is what screws this up, because you can't filter a locale bundle archive.

This was true in Fedora as well, which led to us [splitting the package up in Fedora 24](https://fedoraproject.org/wiki/Changes/Glibc_locale_subpackaging). That with rich dependencies allowed us to make it a little less opaque for handling some aspects of language packages for heavier applications like Firefox, LibreOffice, etc. that have langpacks associated with locales (and are _not_ filtered by `_install_langs`).

schaefi:
>Moreover, just because the system locale is set one way doesn't mean that locale filtering is needed or >wanted. 

Then I suggest we add a locale filtering property like David suggested with a bool element

```xml
<locale>en_US,de_DE</locale>
<locale-filtering>true|false</locale-filtering>
```

kiwi/system/prepare.py
Conan-Kudo: Ah, OK.

Conan-Kudo:
@schaefi If we do that, we should enforce that `POSIX`, `C`, and `C.UTF-8` are always present when `%_install_langs` is used.

schaefi:
yes

schaefi:
@davidcassany code changed as discussed. Could you have a look again. Thanks

PR #937, opened 2019-02-08
Extend the .packages file by the license field

schaefi:
For rpm based builds the License field from the rpm metadata
is extracted into the .packages file. For Debian based build
the license information is in an extra file and not taken
into account for the moment.


schaefi:
This additional information is expected to be read by the public cloud team to check if an image build can be uploaded into the cloud or if there are license issues in some package which prevents that

PR #938, opened 2019-02-09
Sort filesystem listings

bmwiedemann:
Change proposed in this pull request:

Sort filesystem listings
so that kiwi works in a reproducible way
in spite of indeterministic filesystem readdir order
and http://bugs.python.org/issue30461

See https://reproducible-builds.org/ for why this is good.

Notes:

- not all sorts might be needed
- not tested
- listifies some iterators from iglob, which would have a performance impact if there were thousands of matching entries

### DIFF
--- a/kiwi/archive/tar.py
+++ b/kiwi/archive/tar.py
@@ -155,7 +155,7 @@ class ArchiveTar(object):
         final_archive_contents = []
         archive_contents = file_list
         if not archive_contents:
-            archive_contents = os.listdir(source_dir)
+            archive_contents = sorted(os.listdir(source_dir))
 
         for item in archive_contents:
             if not exclude_list:
--- a/kiwi/system/kernel.py
+++ b/kiwi/system/kernel.py
@@ -163,7 +163,7 @@ class Kernel(object):
             # lookup for the symlink first
             'vmlinux', 'vmlinuz'
         ]
-        kernel_dirs = os.listdir(''.join([self.root_dir, '/lib/modules']))
+        kernel_dirs = sorted(os.listdir(''.join([self.root_dir, '/lib/modules'])))
         if kernel_dirs:
             # append lookup for the real kernel image names
             # depending on the arch and os they are different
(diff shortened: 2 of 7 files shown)


kiwi/iso_tools/cdrtools.py
```diff
@@ -250,7 +250,8 @@ def _create_sortfile(self):
             boot_files = list(os.walk(self.source_dir + '/' + self.boot_path))
             boot_files += list(os.walk(self.source_dir + '/EFI'))
             for basedir, dirnames, filenames in boot_files:
-                for filename in filenames:
+                #dirnames.sort() # TODO - seems to break mocking
```
bmwiedemann: left this for now. One workaround is to write it as `dirnames=sorted(dirnames)`
but IMHO fixing the tests would be better

kiwi/iso_tools/cdrtools.py
```diff
@@ -250,7 +250,8 @@ def _create_sortfile(self):
             boot_files = list(os.walk(self.source_dir + '/' + self.boot_path))
             boot_files += list(os.walk(self.source_dir + '/EFI'))
             for basedir, dirnames, filenames in boot_files:
-                for filename in filenames:
+                # dirnames.sort() # TODO - seems to break mocking
```
schaefi: Thanks for your effort, looks all good to me except for the TODO here. We don't leave todo's in the kiwi code base, because they usually stay for ever :-) So I looked at that code and I think the sorting problem is also caused by the two lists dirnames and filenames handled separately. So they can be sorted on its own but when the result is written it still can lead to unsorted output in the sort file. I therefore propose the patch in #939 

Feel free to incorporate that one here or let me know if we should merge (after review) #939 first

Thanks

schaefi:
I'm fine with the changes and will merge the PR once the conflict got resolved. Thanks

bmwiedemann:
resolved.

schaefi:
yes that's nice, thanks :+1:

PR #939, opened 2019-02-11
Sort file and dir entries in iso sortfile

schaefi:
Related to #938

### DIFF
--- a/kiwi/iso_tools/cdrtools.py
+++ b/kiwi/iso_tools/cdrtools.py
@@ -249,14 +249,12 @@ class IsoToolsCdrTools(IsoToolsBase):
 
             boot_files = list(os.walk(self.source_dir + '/' + self.boot_path))
             boot_files += list(os.walk(self.source_dir + '/EFI'))
-            for basedir, dirnames, filenames in boot_files:
-                for filename in filenames:
-                    if filename == 'efi':
-                        sortfile.write('%s/%s 1000001\n' % (basedir, filename))
+            for basedir, dirnames, filenames in sorted(boot_files):
+                for entry in sorted(dirnames + filenames):
+                    if entry in filenames and entry == 'efi':
+                        sortfile.write('%s/%s 1000001\n' % (basedir, entry))
                     else:
-                        sortfile.write('%s/%s 1\n' % (basedir, filename))
-                for dirname in dirnames:
-                    sortfile.write('%s/%s 1\n' % (basedir, dirname))
+                        sortfile.write('%s/%s 1\n' % (basedir, entry))
 
             sortfile.write(
                 '%s/%s 1000000\n' % (self.source_dir, 'header_end')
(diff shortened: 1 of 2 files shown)


kiwi/iso_tools/cdrtools.py
```diff
@@ -249,14 +249,12 @@ def _create_sortfile(self):
 
             boot_files = list(os.walk(self.source_dir + '/' + self.boot_path))
             boot_files += list(os.walk(self.source_dir + '/EFI'))
-            for basedir, dirnames, filenames in boot_files:
-                for filename in filenames:
-                    if filename == 'efi':
-                        sortfile.write('%s/%s 1000001\n' % (basedir, filename))
+            for basedir, dirnames, filenames in sorted(boot_files):
+                for entry in sorted(dirnames + filenames):
```
bmwiedemann: When testing this in other python projects, I found that I had to do dirnames.sort() because the os.walk would continue using that value for its recursion further down.
Another approach might be to collect lines in an array and sort that before writing it to `sortfile`

schaefi: This is correct but will not fix the sorting issue. os.walk separates for any directory two lists. one containing filenames(filenames) and one containing sub-dirs(dirnames) of the dir it is currently in. The code as I had it before was not suitable for a good sorting because if you sort dirnames and filenames on its own the combined list is not necessarily sorted.

The sortfile for isolinux as we create it here is one list of entries which takes different parameters depending if the entry is a directory or a file. The overall list however should be sorted. So the code should also change semantically which is done here, working on ```sorted(dirnames + filenames)```. That also allows to handle it properly in the unit test

schaefi: So if you are ok with this we can merge this one and you would need to rebase your open PR

Thoughts ?

bmwiedemann: I still think, this code is not as good as it could be. It is perfectly fine to sort dirnames and filenames individually, because one list contains dirs and the other non-dirs. And both are on the same dir level, but they will not randomly change positions, because dirs remain dirs.

bmwiedemann: and you did not yet address the unsorted recursion in os.walk

schaefi: why that ? ```sorted(boot_files)``` makes sure basedir from os.walk is always sorted, that handles the unsorted recursion. ```sorted(dirnames + filenames)``` handles all entries (dirs and files) from current recursion are sorted.

> because one list contains dirs and the other non-dirs.

correct and sort(a) + sort(b) != sort(a+b)

```python
a = ['z', 'y']

b = ['a', 'b', 'c']



x1 = sorted(a) + sorted(b)

x2 = sorted(a+b)

print(x1)

print(x2)
```

bmwiedemann: But both variants would be deterministic. So I dont care.

schaefi: yes this is true, but in the sortfile for isolinux which is flat and contains one entry for each no matter if it's a dir or file I think we should care.

So you would be ok with that change then ?

Thanks

schaefi: I'm asking because a merge will create a conflict on your open PR

schaefi:
ok :)

PR #947, opened 2019-02-18
Add support for virtualbox vagrant boxes

dcermak:
Fixes #532.

### Changes proposed in this pull request
* ~~split `DiskFormatVagrantLibVirt` into `DiskFormatVagrantBase` and `DiskFormatVagrantLibVirt` (and equally with the unit tests)~~ (done in #948)
* extend the xml schema: `vagrantconfig.provider` now accepts `virtualbox` as an option and add the attribute `vagrantconfig.virtualbox_os_type`
* add a new class `DiskFormatVagrantVirtualBox` which implements support for virtualbox vagrant boxes using @Conan-Kudo's ovf template
* ~~add the function `vagrantSetup` to `functions.sh`, it performs all the necessary setup of the vagrant user, ssh, sudoers, etc. (so the end user does not need to do that)~~

### current problems and short-commings
- it is currently impossible to build a libvirt and virtualbox vagrant box, but it would be nice to be able to do that, especially if we are to add vmware support (which should now become relatively easy).
- ~~I am not sold on my modification of the XML schema, especially `vagrantconfig.virtualbox_os_type` is pretty ugly~~
- ~~Virtualbox will by default mount the share as `vboxfs`, which only works if the virtualbox guest additions are installed in the box. We cannot assume that this is true, but it would be a mistake to turn this to rsync by default, as that is afaik slower. Maybe add an xml option via which the user tells kiwi that they will install the guest additions?~~
- ~~I am unhappy with the current way how one can customize the `Vagrantfile` that is put into the box (see `DiskFormatVagrantBase._create_box_vagrantconfig` in `kiwi/storage/subformat/vagrant_base.py`). It makes the tests look ugly as one can't use `dedent`.~~

### TODO
- [x] Rebase the fixup commits and fix the commit order
- [x] improve test coverage

PR #948, opened 2019-02-19
Split common vagrant tasks into base class DiskFormatVagrantBase

dcermak:
Fixes part of the issues addressed in #947.

* Split `DiskFormatVagrantLibVirt` into `DiskFormatVagrantLibVirt` and `DiskFormatVagrantBase`, as creating a vagrant box involves a few steps that have to be done independently of the provider. The base class performs these instead of the libvirt specific class.
* Split the unit test for `DiskFormatVagrantLibVirt` into tests for the base class (which I have tried to write so that they test the documented behavior of the base class) and test of the libvirt class
* I have modified the `Vagrantfile` that is bundled with the libvirt provider to match the one that packer produces.

### DIFF
--- a/kiwi/storage/subformat/vagrant_base.py
+++ b/kiwi/storage/subformat/vagrant_base.py
@@ -88,9 +88,11 @@ class DiskFormatVagrantBase(DiskFormatBase):
                 'no vagrantconfig provided'
             )
 
-        self.image_format = 'vagrant.' + self.provider + '.box'
-
         self.vagrantconfig = custom_args['vagrantconfig']
+        self.vagrant_post_init()
+
+    def vagrant_post_init(self):
+        pass
 
     def create_box_img(self, temp_image_dir):
         """
--- a/kiwi/storage/subformat/vagrant_base.py
+++ b/kiwi/storage/subformat/vagrant_base.py
@@ -1,0 +1,212 @@
+# Copyright (c) 2019 SUSE Linux GmbH.  All rights reserved.
+#
+# This file is part of kiwi.
+#
+# kiwi is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# kiwi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kiwi.  If not, see <http://www.gnu.org/licenses/>
+#
+import json
+import os
+import random
+from textwrap import dedent
+from tempfile import mkdtemp
+from string import Template
+
+from kiwi.storage.subformat.base import DiskFormatBase
+from kiwi.command import Command
+
+from kiwi.exceptions import (
+    KiwiFormatSetupError
+)
+
+
+class DiskFormatVagrantBase(DiskFormatBase):
+    """
+    Base class for creating vagrant boxes.
+
+    The documentation of the vagrant box format can be found here:
+    https://www.vagrantup.com/docs/boxes/format.html
+    In a nutshell, a vagrant box is a tar, tar.gz or zip archive of the
+    following:
+
+    1. ``metadata.json``: A json file that contains the name of the provider
+       and arbitrary additional data (that vagrant doesn't care about).
+    2. ``Vagrantfile``: A Vagrantfile which defines the boxes' MAC address. It
+       can be also used to define other settings of the box, e.g. the method
+       via which the ``/vagrant/`` directory is shared.
+    3. The actual virtual disk image: this is provider specific and vagrant
+       simply forwards it to your virtual machine provider.
+
+    Required methods/variables that child classes must implement:
+
+    - ``provider: str``:
+      A static variable or property, should contain the name of the provider.
+      This value is used to create the ``metadata.json`` file and the attribute
+      :attr:`image_format`.
+      (Note: you also must add the image format to
+      :func:`kiwi.defaults.Defaults.get_disk_format_types`)
+
+    - :meth:`create_box_img`
+
+    Optional methods:
+
+    - :meth:`get_additional_metadata`
+
+    - :meth:`get_additional_vagrant_config_settings`
+
+    """
+    def post_init(self, custom_args):
+        """
+        vagrant disk format post initialization method
+
+        store vagrantconfig information provided via custom_args
+
+        :param dict custom_args:
+            Contains instance of xml_parse::vagrantconfig
+
+            .. code:: python
+
+                {'vagrantconfig': object}
+        """
+        if not custom_args or 'vagrantconfig' not in custom_args:
+            raise KiwiFormatSetupError(
+                'object init requires custom_args hash with a vagrantconfig'
+            )
+
+        if not custom_args['vagrantconfig']:
+            raise KiwiFormatSetupError(
+                'no vagrantconfig provided'
+            )
+
+        self.image_format = 'vagrant.' + self.provider + '.box'
+
+        self.vagrantconfig = custom_args['vagrantconfig']
+
+    def create_box_img(self, temp_image_dir):
+        """
+        Provider specific image creation step: this function creates the actual
+        box image. It must be implemented by a child class.
+
+        :param str temp_image_dir: path to a temporary directory inside which the
+            image should be built
+        :return: A list of files that were create by this function and that
+            should be included in the vagrant box
+        :rtype: List(str)
+        """
+        raise NotImplementedError
+
+    def create_image_format(self):
+        """
+        Create a vagrant box for any provider. This includes:
+
+        * creation of box metadata.json
+        * creation of box Vagrantfile
+        * creation of result format tarball from the files created above
+        """
+        self.temp_image_dir = mkdtemp(prefix='kiwi_vagrant_box.')
+
+        box_img_files = self.create_box_img(self.temp_image_dir)
+
+        metadata_json = os.sep.join([self.temp_image_dir, 'metadata.json'])
+        with open(metadata_json, 'w') as meta:
+            meta.write(self._create_box_metadata())
+
+        vagrantfile = os.sep.join([self.temp_image_dir, 'Vagrantfile'])
+        with open(vagrantfile, 'w') as vagrant:
+            vagrant.write(self._create_box_vagrantconfig())
+
+        Command.run(
+            [
+                'tar', '-C', self.temp_image_dir,
+                '-czf', self.get_target_file_path_for_format(self.image_format),
+                os.path.basename(metadata_json),
+                os.path.basename(vagrantfile)
+            ] + [
+                os.path.basename(box_img_file)
+                for box_img_file in box_img_files
+            ]
+        )
+
+    def store_to_result(self, result):
+        """
+        Store result file of the vagrant format conversion into the
+        provided result instance. In this case compression is unwanted
+        because the box is already created as a compressed tarball
+
+        :param object result: Instance of Result
+        """
+        result.add(
+            key='disk_format_image',
+            filename=self.get_target_file_path_for_format(
+                self.image_format
+            ),
+            use_for_bundle=True,
+            compress=False,
+            shasum=True
+        )
+
+    @classmethod
+    def get_additional_metadata(cls):
+        """
+        Provide :meth:`create_image_format` with additional metadata that will
+        be included in ``metadata.json``.
+
+        The default implementation returns an empty dictionary.
+
+        :return: A dictionary that is serializable to JSON
+        :rtype: dict
+        """
+        return {}
+
+    @classmethod
+    def get_additional_vagrant_config_settings(cls):
+        """
+        Supply additional configuration settings for vagrant to be included in
+        the resulting box.
+
+        This function can be used by child classes to customize the behavior
+        for different providers: the supplied configuration settings get
+        forwarded to :func:`VagrantConfigTemplate.get_template` as the
+        parameter ``custom_settings`` and included in the ``Vagrantfile``.
+
+        The default implementation returns nothing.
+
+        :return: additional vagrant settings
+        :rtype: str
+        """
+        pass
+
+    def _create_box_metadata(self):
+        metadata = self.get_additional_metadata()
+        metadata['provider'] = self.provider
+        return json.dumps(
+            metadata, sort_keys=True, indent=2, separators=(',', ': ')
+        )
+
+    def _create_box_vagrantconfig(self):
+        template = VagrantConfigTemplate()
+        vagrant_config = template.get_template(
+            custom_settings=self.get_additional_vagrant_config_settings()
+        )
+        return vagrant_config.substitute(
+            {'mac_address': self._random_mac()}
+        )
+
+    @staticmethod
+    def _random_mac():
+        return '%02x%02x%02x%02x%02x%02x'.upper() % (
+            0x00, 0x16, 0x3e,
+            random.randrange(0, 0x7e),
+            random.randrange(0, 0xff),
+            random.randrange(0, 0xff)
+        )
--- a/kiwi/storage/subformat/vagrant_base.py
+++ b/kiwi/storage/subformat/vagrant_base.py
@@ -18,108 +18,16 @@
 import json
 import os
 import random
-from textwrap import dedent
+
 from tempfile import mkdtemp
-from string import Template
 
 from kiwi.storage.subformat.base import DiskFormatBase
+from kiwi.storage.subformat.template.vagrant_config import VagrantConfigTemplate
 from kiwi.command import Command
 
 from kiwi.exceptions import (
     KiwiFormatSetupError
 )
-
-
-class VagrantConfigTemplate(object):
-    """
-    **Generate a Vagrantfile configuration template**
-
-    This class creates a simple template for the Vagrantfile that is included
-    inside a vagrant box.
-
-    The included Vagrantfile carries additional information for vagrant: by
-    default that is just the boxes' MAC address, but depending on the provider
-    additional information need to be present. These can be passed via the
-    parameter ``custom_settings`` to the method :meth:`get_template`.
-
-    Example usage:
-
-    The default without any additional settings will result in this
-    Vagrantfile:
-
-    .. code:: python
-
-        >>> vagrant_config = VagrantConfigTemplate()
-        >>> print(
-        ...     vagrant_config.get_template()
-        ...     .substitute({'mac_address': 'deadbeef'})
-        ... )
-        Vagrant.configure("2") do |config|
-          config.vm.base_mac = "deadbeef"
-        end
-
-    If your provider/box requires additional settings, provide them as follows:
-
-    .. code:: python
-
-        >>> extra_settings = dedent('''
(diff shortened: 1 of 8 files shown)


kiwi/storage/subformat/vagrant_libvirt.py
```diff
@@ -83,71 +44,24 @@ def create_image_format(self):
...
-            random.randrange(0, 0xff)
-        )
+    def additional_vagrant_settings(self):
+        # () -> str:
+        """
+        Returns settings for the libvirt provider telling vagrant to use kvm.
+        """
+        return """config.vm.provider :libvirt do |libvirt|
+    libvirt.driver = "kvm"
+  end"""
```
schaefi: Could you use the dedent module for returning strings like this:

```python
from textwrap import dedent

libvirt_driver_setup = dedent('''
    config.vm.provider :libvirt do |libvirt|
        libvirt.driver = "kvm"
    end
''')

return libvirt_driver_setup
```

This is imho nicer to read and explicit in what it returns

dcermak: I'll give it a try, but I think I didn't use it because it broke the indentation of the resulting `Vagrantfile` that ends up in the box.

schaefi: The dedent module allows to provide the indentation as you need it. If you strip() and add a linesep at the end it should be identical to the version you had in """ quotes

dcermak: I have changed the way `_create_box_vagrantconfig` pastes the customized vagrant config and the result looks better now, albeit requiring another call to `indent`.

schaefi: from textwrap import dedent

that exists in py2 and py3, kiwi already uses that module for years :)

schaefi: Still like the version from above more. Sorry but

```
 "  " + "  ".join(
            dedent("""
```

looks strange to me. Also python uses single quotes for text that is taken one by one.

```python
dedent('''
    config.vm.provider :libvirt do |libvirt|
        libvirt.driver = "kvm"
    end
''').strip()
```

produces the same formatting result and avoids and leading spaces. Any reason why that does not work for you ?

schaefi: Also saw you are using splitlines() now which turns the string including espace sequences into a list. That invalidates the doc string because the method now returns a list and not a string anymore. Also why is this better than before ? It's starting to confuse me.

schaefi: The concept of additional_vagrant_settings() is ok for me as it just added raw data to the vagrantfile. Maybe the method should be better named additional_vagrant_raw_config() or similar. If you wish to add more structure I suggest to pass a dictionary with information, but in case of the Vagrantfile which is actually ruby code the idea to pass raw data is probably the more lightweight solution. So I'm ok with the string but the list I don't understand

dcermak: Yes, I actually need the two leading spaces, otherwise it'll mess up the `Vagrantfile`.

Since it is quite confusing having to do that in child classes, I have moved the indentation code to `DiskFormatVagrantBase._create_box_vagrantconfig`, so that this is done once for all children.

schaefi: ok, thanks for the changes. I was reading the code again and I think we should do the vagrantconfig handling a bit different. My idea is to use a simple template. I hope you don't mind but that spacing and such could bite us in the future and is not very flexible. So here is my idea

```python
import os
  
from textwrap import dedent
from string import Template

class VagrantConfigTemplate(object):
    """
    **vagrant config template**
    """
    def __init__(self):
        self.indent = '  ';

        self.header = dedent('''
            Vagrant.configure("2") do |config|
        ''').strip() + os.linesep

        self.mac = self.indent + dedent('''
            config.vm.base_mac = ${mac_address}
        ''').strip() + os.linesep

        self.end = dedent('''
            end
        ''').strip()

    def get_template(self, custom_settings=None):
        template_data = self.header
        template_data += self.mac
        if custom_settings:
            template_data += self.indent
            template_data += self.indent.join(
                custom_settings.splitlines(True)
            )
            template_data += os.linesep
        template_data += self.end
        return Template(template_data)

def get_additional_vagrant_config_settings():
    return dedent('''
        config.vm.provider :libvirt do |libvirt|
          libvirt.driver = "kvm"
        end
    ''').strip()

def _create_box_vagrantconfig():
    template = VagrantConfigTemplate()
    vagrant_config = template.get_template(
        custom_settings=get_additional_vagrant_config_settings()
    )
    return vagrant_config.substitute(
        {'mac_address': 'foo'}
    )

print(_create_box_vagrantconfig())
```

What do you think ? It will connect all indention and layouting into a template class and also allows for better testing.

Thanks for your patience

dcermak: I'll give your idea a try and code something up so that we find out its pain points. I agree that just pasting stuff into the `Vagrantfile` (as I have done it) is pretty ugly and could result in bad things happening.

schaefi: great, thanks :+1:

dcermak: Done. I have refactored the code a little bit and converted it to a single function. Since I added a doctest to this function I tried to enable them for pytest, unfortunately it does not seem to collect them properly.

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,266 @@
...
+    template_data = HEADER
+    template_data += MAC
+    if custom_settings:
+        template_data += INDENT
+        template_data += INDENT.join(
+            custom_settings.splitlines(True)
+        )
+        template_data += os.linesep
+    template_data += END
+    return Template(template_data)
```
schaefi: Could we please put the template into its own class. We use this concept in kiwi in other places too. See e.g ```kiwi/storage/subformat/template/vmware_settings.py``` or ```kiwi/bootloader/template/```

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,266 @@
...
+    end
+
+    If your provider/box requires additional settings, provide them as follows:
+    >>> extra_settings = dedent('''
+    ... config.vm.hostname = "no-dead-beef"
+    ... config.vm.provider :special do |special|
+    ...   special.secret_settings = "please_work"
+    ... end
+    ... ''').strip()
+    >>> print(get_vagrantfile_template(extra_settings).substitute({'mac_address': 'DEADBEEF'}))
```
schaefi: That information imho belongs to a template class specification. For coding examples we use

```python
"""
.. code:: python
     ...
"""
```

The :returns: as well as :Example: as well as :type: seems to be not evaulated by the sphinx docs. So for method doc strings we use this format all over the place

```
def foo(arg):
    """
    Does something

    :param type arg: arg for foo

    :return: whatever it returns

    :rtype: return_type
    """
```

Could you adapt that to be consistent with the rest of the code base ? I know there are many ways to do this but it wold be good to have a common schema

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,266 @@
...
+            ),
+            use_for_bundle=True,
+            compress=False,
+            shasum=True
+        )
+
+    @classmethod
+    def get_additional_metadata(cls):
+        # type: () -> Dict[str, str]
+        return {}
```
schaefi: Could we avoid that type of comments: In this particular method it's also wrong. it returns an empty dict
methods of the public interface should have a doc string as explained before. private methods (_foo) doesn't need it and could take some information if not clear by the code. I usually avoid any comment block that is of no interest because not in the public api and try to write the code to be as readable as possible. From my experience if there are comment blocks that looks like code either the comment or the code is wrong :) So let's focus on the minimal needed and be as explicit and clear in the code as possible. Hope that makes sense to you too

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,266 @@
...
+
+    @classmethod
+    def get_additional_vagrant_config_settings(cls):
+        # type: () -> str
+        return ""
```
schaefi: It should just pass and avoid that comment block. When using a template class there is no need for the method to return an empty string. But it should mention that this method can be overwritten by the specialized provider
class. e.g

```python
@classmethod
def get_additional_vagrant_config_settings(cls):
    """
    Handle additional configurations in the Vagrantfile

    Depending on the provider additional vagrant configuration settings are required.
    The implementation of those are done in the specialized provider class.
    """
    pass
```

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,266 @@
...
+        vagrant_config = get_vagrantfile_template(
+            custom_settings=self.get_additional_vagrant_config_settings()
+        )
+        return vagrant_config.substitute(
+            {'mac_address': self._random_mac()}
+        )
+
+    @staticmethod
+    def _random_mac():
+        # type: () -> str
```
schaefi: superfluous comment

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,266 @@
...
+        # type: () -> Dict[str, str]
+        return {}
+
+    @classmethod
+    def get_additional_vagrant_config_settings(cls):
+        # type: () -> str
+        return ""
+
+    def _create_box_metadata(self):
+        # type: () -> str
```
schaefi: superfluous comment

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,266 @@
...
+                os.path.basename(metadata_json),
+                os.path.basename(vagrantfile)
+            ] + [
+                os.path.basename(box_img_file)
+                for box_img_file in box_img_files
+            ]
+        )
+
+    def store_to_result(self, result):
+        # type: (Result) -> None
```
schaefi: superfluous comment

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,266 @@
...
+        if not custom_args['vagrantconfig']:
+            raise KiwiFormatSetupError(
+                'no vagrantconfig provided'
+            )
+
+        self.image_format = 'vagrant.' + self.provider + '.box'
+        self.vagrantconfig = custom_args['vagrantconfig']
+
+    def create_image_format(self):
+        # type: () -> None
```
schaefi: superfluous comment

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,266 @@
...
+from kiwi.command import Command
+from kiwi.system.result import Result  # pylint: disable=unused-import
+
+from kiwi.exceptions import (
+    KiwiFormatSetupError
+)
+
+
+def get_vagrantfile_template(custom_settings=None):
+    # type: (Optional[str]) -> Template
```
schaefi: superfluous comment

kiwi/storage/subformat/vagrant_libvirt.py
```diff
@@ -15,63 +15,26 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
-import json
 import os
-import random
+from typing import Dict  # pylint: disable=unused-import
```
schaefi: why, unused imports should be deleted

kiwi/storage/subformat/vagrant_libvirt.py
```diff
@@ -15,63 +15,26 @@
...
 from kiwi.command import Command
 
-from kiwi.exceptions import (
-    KiwiFormatSetupError
-)
 
-
-class DiskFormatVagrantLibVirt(DiskFormatBase):
+class DiskFormatVagrantLibVirt(DiskFormatVagrantBase):
+    provider = 'libvirt'
```
schaefi: I think I overlooked something, but where is this variable used ?

kiwi/storage/subformat/vagrant_libvirt.py
```diff
@@ -83,71 +46,26 @@ def create_image_format(self):
...
 
-    def _random_mac(self):
-        return '%02x%02x%02x%02x%02x%02x'.upper() % (
-            0x00, 0x16, 0x3e,
-            random.randrange(0, 0x7e),
-            random.randrange(0, 0xff),
-            random.randrange(0, 0xff)
-        )
+    def get_additional_vagrant_config_settings(self):
+        # type: () -> str
```
schaefi: superfluous comment

kiwi/storage/subformat/vagrant_libvirt.py
```diff
@@ -83,71 +46,26 @@ def create_image_format(self):
...
+        """
+        Returns settings for the libvirt provider telling vagrant to use kvm.
+        """
+        return dedent('''
+        config.vm.provider :libvirt do |libvirt|
+          libvirt.driver = "kvm"
+        end
+            ''').strip()
```
schaefi: With templating in place and the fact that the string is stripped anyway this can be imho nicer indented:

```python
return dedent('''
    config.vm.provider :libvirt do |libvirt|
      libvirt.driver = "kvm"
￼    end
''').strip()
```

tox.ini
```diff
@@ -60,8 +60,9 @@ deps = {[testenv]deps}
 changedir=test/unit
 commands =
     bash -c 'cd ../../ && ./setup.py develop'
-    pytest --no-cov-on-fail --cov=kiwi \
-        --cov-report=term-missing --cov-fail-under=100 --cov-config .coveragerc {posargs}
+    pytest --doctest-modules --no-cov-on-fail --cov=kiwi \
+        --cov-report=term-missing --cov-fail-under=100 \
+        --cov-config .coveragerc {posargs}
```
schaefi: Changes in tox.ini do not belong to this pull request and while I agree with the ones done here they should be done in an extra pull request to avoid confusion

kiwi/storage/subformat/vagrant_libvirt.py
dcermak: That was necessary for the typing comment: `# type: () -> Dict[str, str]`.

This is an unfortunate consequence of using type hints in a python 2 compatible code base: you have to import all the types so that a type checker can evaluate these, but since you only use them in comment blocks, pylint & co. complain about unused imports since they don't evaluate type hints.

kiwi/storage/subformat/vagrant_libvirt.py
dcermak: This is used in `DiskFormatVagrantBase.post_init` where the base class constructs the image format from the provider name.

tox.ini
dcermak: Will drop them then.

kiwi/storage/subformat/vagrant_base.py
dcermak: I'll drop them.

kiwi/storage/subformat/vagrant_base.py
dcermak: Will do that!

Out of curiosity: do you know if sphinx can actually run these python code examples? I have added this as a doctest because they can be tested relatively easily via the doctest module.

kiwi/storage/subformat/vagrant_libvirt.py
schaefi: Ah ok now I saw it

ok I was looking for where self.provider gets initialized. I think DiskFormatVagrantBase should initialize a default provider. The setting for ```self.image_format = 'vagrant.' + self.provider + '.box'``` is provider specific and imho doesn't belong to the base class. I think DiskFormatVagrantBase should call another post_init() such that the specialized provider class can overload that method and just set ```self.provider = 'libvirt'```. The way you did it does not loosely couple the classes and ties them hard together. The methods in the base class that references ```self.image_format``` should only reference self.provider with the value that it has at call time or its default value.

Sorry if that sounds nitpicking but it took me a while to find out how provider is used as it's class global. We should avoid that imho

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,304 @@
...
+
+from kiwi.storage.subformat.base import DiskFormatBase
+from kiwi.command import Command
+
+from kiwi.exceptions import (
+    KiwiFormatSetupError
+)
+
+
+class VagrantConfigTemplate(object):
```
schaefi: ok, nice but could you put it into an extra file below the existing template directory: ```kiwi/storage/subformat/template```

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,304 @@
...
+        Vagrant.configure("2") do |config|
+    ''').strip() + os.linesep
+
+    _MAC = _INDENT + dedent('''
+        config.vm.base_mac = "${mac_address}"
+    ''').strip() + os.linesep
+
+    _END = dedent('''
+        end
+    ''').strip()
```
schaefi: I usually avoid class global values especially when we create an instance object of the class. The template is created in code later as ```template = VagrantConfigTemplate()``` and we operate on template. Thus all this values from above _INDENT, _HEADER, etc... deserves a constructor and should imho be set in the __init__ method.

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -0,0 +1,304 @@
...
+
+    @classmethod
+    def get_template(cls, custom_settings=None):
```
schaefi: I think this should be a normal method not a classmethod because we create an object from VagrantConfigTemplate

schaefi:
Also please check on the codacy report. I already marked the random issue as to be ignore because we don't really use it for security reasons as they complain :) The other issue however is valid and should be fixed. Thanks

kiwi/storage/subformat/vagrant_base.py
dcermak: I'll change that.

I have used class global values since they will be constructed once the module is imported and not every time a template is created. But since it's unlikely that we will create two templates, it won't make a difference.

kiwi/storage/subformat/vagrant_base.py
dcermak: Yes, I'll change that back to an ordinary method. I used a classmethod since codacy would have complained about `get_template` not using self.

dcermak:
I can't reply to this comment directly, so I'll do it here:

> Ah ok now I saw it
>
> ok I was looking for where self.provider gets initialized. I think DiskFormatVagrantBase should initialize a default provider. The setting for self.image_format = 'vagrant.' + self.provider + '.box' is provider specific and imho doesn't belong to the base class. I think DiskFormatVagrantBase should call another post_init() such that the specialized provider class can overload that method and just set self.provider = 'libvirt'. The way you did it does not loosely couple the classes and ties them hard together. The methods in the base class that references self.image_format should only reference self.provider with the value that it has at call time or its default value.
>
> Sorry if that sounds nitpicking but it took me a while to find out how provider is used as it's class global. We should avoid that imho

I have just implemented this and it has a disadvantage imho: I need a second `post_init` style function (which I have called `vagrant_post_init`). Using `post_init` and then calling the base classes' `post_init` via `super()` does not work in Python2.7, where this always results in `DiskFormatBase.post_init` being called (which does nothing).

I have just pushed this change, so that you can judge for yourself whether you like it or not.

schaefi:
Looks good to me now. Thanks. I'll merge and see if Virt:App staging builds still succeeds

PR #952, opened 2019-02-20
Fixed losetup call due to move in option names

schaefi:
util-linux dropped losetup-Add-support-for-setting-logical-blocksize.patch
because different implementations exists in the new kernel, and it has
a conflicting implementation in util-linux. This caused a change in the
option to specify the logical sector size. The option --logical-blocksize
was replaced by --sector-size. We adapt this now in kiwi too


schaefi:
From Rudi

> wie's aussieht heisst es auf sle12: --logical-blocksize
> und auf sle15 und factory dann: --sector-size (bzw. "-b")

We have to differentiate by capability

PR #957, opened 2019-02-21
Added some style fixes to the vagrant classes

schaefi:
Adapt doc strings to match style on lists. Change variable
names not class global to be lowercase. Use 80 chars per
line. Don't reach code that potentially uses undefined
variables

### DIFF
--- a/kiwi/storage/subformat/vagrant_base.py
+++ b/kiwi/storage/subformat/vagrant_base.py
@@ -39,23 +39,25 @@ class DiskFormatVagrantBase(DiskFormatBase):
     In a nutshell, a vagrant box is a tar, tar.gz or zip archive of the
     following:
 
-    1. ``metadata.json``: A json file that contains the name of the provider
+    1. ``metadata.json``:
+       A json file that contains the name of the provider
        and arbitrary additional data (that vagrant doesn't care about).
-    2. ``Vagrantfile``: A Vagrantfile which defines the boxes' MAC address. It
-       can be also used to define other settings of the box, e.g. the method
-       via which the ``/vagrant/`` directory is shared.
-    3. The actual virtual disk image: this is provider specific and vagrant
-       simply forwards it to your virtual machine provider.
+    2. ``Vagrantfile``:
+       A Vagrantfile which defines the boxes' MAC address. It
+       can be also used to define other settings of the box, e.g.
+       the method via which the ``/vagrant/`` directory is shared.
+    3. The actual virtual disk image: this is provider specific and
+       vagrant simply forwards it to your virtual machine provider.
 
     Required methods/variables that child classes must implement:
 
-    * ``provider: str``:
-      A static variable or property, should contain the name of the provider.
-      This value is used to create the ``metadata.json`` file and the
-      attribute :attr:`image_format`.
+    * :meth:`vagrant_post_init`
 
-      Note: you also must add the image format
-      to :func:`kiwi.defaults.Defaults.get_disk_format_types`
+      post initializing method that has to specify the vagrant
+      provider name in :attr:`provider` and the box name in
+      :attr:`image_format`. Note: new providers also needs to
+      be specified in the schema and the box name needs to be
+      registered to :func:`kiwi.defaults.Defaults.get_disk_format_types`
 
     * :meth:`create_box_img`
 
@@ -110,7 +112,7 @@ class DiskFormatVagrantBase(DiskFormatBase):
             path to a temporary directory inside which the image
             should be built
         :return:
-            A list of files that were create by this function
+            A list of files that were created by this function
             and that should be included in the vagrant box
         :rtype: list
         """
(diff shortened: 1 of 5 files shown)


kiwi/storage/subformat/vagrant_base.py
```diff
@@ -49,20 +49,20 @@ class DiskFormatVagrantBase(DiskFormatBase):
...
 
-    - ``provider: str``:
+    * ``provider: str``:
       A static variable or property, should contain the name of the provider.
-      This value is used to create the ``metadata.json`` file and the attribute
-      :attr:`image_format`.
-      (Note: you also must add the image format to
-      :func:`kiwi.defaults.Defaults.get_disk_format_types`)
+      This value is used to create the ``metadata.json`` file and the
+      attribute :attr:`image_format`.
```
dcermak: Ups, this is left over from my first draft, I though that I fixed the documentation. `image_format` must be set explicitly by the child class and is not influenced by `self.provider`.

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -87,23 +87,32 @@ def post_init(self, custom_args):
...
-        :param str temp_image_dir: path to a temporary directory inside which the
-            image should be built
-        :return: A list of files that were create by this function and that
-            should be included in the vagrant box
-        :rtype: List(str)
+        :param str temp_image_dir:
+            path to a temporary directory inside which the image
+            should be built
+        :return:
+            A list of files that were create by this function
```
dcermak: `create` -> `created` (typo by me)

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -115,29 +124,32 @@ def create_image_format(self):
...
-                'tar', '-C', self.temp_image_dir,
-                '-czf', self.get_target_file_path_for_format(self.image_format),
-                os.path.basename(metadata_json),
-                os.path.basename(vagrantfile)
-            ] + [
-                os.path.basename(box_img_file)
-                for box_img_file in box_img_files
-            ]
-        )
+        if self.image_format and self.provider:
```
dcermak: [suggestion] The function will blow up if called from a child class that did not properly set `image_format` or `provider`. With your change it will silently do nothing. I would advise not to do that.

kiwi/storage/subformat/vagrant_base.py
```diff
@@ -147,31 +159,30 @@ def store_to_result(self, result):
...
-        result.add(
-            key='disk_format_image',
-            filename=self.get_target_file_path_for_format(
-                self.image_format
-            ),
-            use_for_bundle=True,
-            compress=False,
-            shasum=True
-        )
+        if self.image_format:
```
dcermak: The same suggestion as for `create_image_format` applies here too.

kiwi/storage/subformat/vagrant_base.py
schaefi: yes you are fully correct, I'll adapt

kiwi/storage/subformat/vagrant_base.py
schaefi: Thanks, will adapt the doc texts and check on the sphinx output

schaefi: done

kiwi/storage/subformat/vagrant_base.py
schaefi: done, raising on improper setup

kiwi/storage/subformat/vagrant_base.py
schaefi: done, raising on improper setup

kiwi/storage/subformat/vagrant_base.py
schaefi: done

test/unit/storage_subformat_vagrant_libvirt_test.py
```diff
@@ -1,87 +1,61 @@
...
-        setattr(context_manager_mock, '__enter__', enter_mock)
-        setattr(context_manager_mock, '__exit__', exit_mock)
-        metadata_json = dedent('''
-            {
-              "format": "qcow2",
-              "provider": "libvirt",
-              "virtual_size": "42"
-            }
-        ''').strip()
-        vagrantfile = dedent('''
```
dcermak: Could we leave the check for the whole Vagrantfile in? When we only test `get_additional_vagrant_config_settings`, it would be possible that the template still creates a malformed Vagrantfile (I know, that is unlikely since we have tests for the template in place).

schaefi: yes let's check the merging. I'll extend the base class test

schaefi: done

schaefi:
How about it now ? Thanks for the review, very much appreciated

dcermak:
> How about it now ? Thanks for the review, very much appreciated

Kind of what I imagined with a small difference: I thought to keep testing the `Vagrantfile` that results from the build of the `DiskFormatVagrantLibVirt` class, not to patch the base class. I copied that Vagrantfile from a working box that packer produced and put it into the unit test as a regression test.

schaefi:
I agree it's a better test to take the information as it is used from the libvirt class. test code updated.
Good to go ?

dcermak:
> Good to go ?

Yes, thanks for adding this!

Issue #942, opened 2019-02-12
Use predictable interface names in dracut-kiwi-dump


schaefi:
## Problem description

The way kiwi's dump dracut module fetches from the network is a setup of a persistent interface name called ```lan0```. The fact that this interface name exists is created through a udev rule generator called ```kiwi-dump-net-genrules.sh```. However this approach is unstable if there are more network adapters available.

There is only one good solution which is to use predictable interface names. predictable interface names can be activated on the kernel commandline or the respective pxe client configuration with

```
net.ifnames=1
```

kiwi should not require a fixed interface name but allow them to be configurable with the dracut option:

```
rd.kiwi.install.pxe.interface=...
```

The default for that option could be ```eth0``` even though it's unstable and we should recommend by documenting it that the interface get properly setup

seife:
The standard way to do this has for a long time been to add `BOOTIF=01-${MAC_ADDRESS}` (as `IPAPPEND 2` in pxelinux config does, for example).
This is already handled by dracut without any further action needed by kiwi-dracut-foo.

But this overwrites the udev rule created by kiwi, so the interface is not named "lan0" anymore and thus "ifup lan0" fails.

I'll send a pull request to overcome at least that problem, then "BOOTIF=..." works like a charm :-)

seife:
https://github.com/SUSE/kiwi/pull/956

schaefi:
This has been resolved with #978

PR #959, opened 2019-02-21
fix network failure with "BOOTIF=" parameter

seife:
If a BOOTIF= parameter (pxelinux "IPAPPEND 2" option) is present, dracut
handles the network already *and* overwrites the 90-net.rules that
kiwi-*-net-genrules.sh created, thus the interface is not named "lan0"
and ifup is destined to fail.
Work around the issue by detecting if the interface is already handled
by generic dracut code and just skipping the ifup call.
Fixes / improves issue #942


schaefi:
Looks great :+1: Thanks

PR #964, opened 2019-02-25
add rd.kiwi.install.pass.bootparam boot parameter

seife:
if this boolean is set, most boot parameters are passed on
to the kexec kernel on OEM image deployments

### DIFF
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -362,7 +362,20 @@ function get_remote_image_source_files {
 
 function boot_installed_system {
     local boot_options
-    boot_options="$(cat /config.bootoptions)"
+    # if rd.kiwi.install.pass.bootparam is given, pass on most
+    # boot options to the kexec kernel
+    if getargbool 0 rd.kiwi.install.pass.bootparam; then
+        local cmdline
+        local option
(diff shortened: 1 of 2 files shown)


dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
```diff
@@ -362,7 +362,20 @@ function get_remote_image_source_files {
...
 function boot_installed_system {
     local boot_options
-    boot_options="$(cat /config.bootoptions)"
+    # if rd.kiwi.install.pass.bootparam is given, pass on most
+    # boot options to the kexec kernel
+    if getargbool 0 rd.kiwi.install.pass.bootparam; then
+        local cmdline
+        local i
+        read -r cmdline < /proc/cmdline
+        for i in $cmdline; do
```
schaefi: * Can we go with variable names that have a context, just ```i``` is very generic
* I also access variables in bash using ${...} all over the place. Even though shellcheck does not complain
  in any case about this it makes it absolutely clear and the scope doesn't allow for any other interpretation

 e.g

```bash
local option
for option in ${cmdline}; do
    ...
done
```

Thanks

seife: in a minute... :-)

schaefi: nice

PR #965, opened 2019-02-25
Fixed import of signing keys

schaefi:
Make sure the key import works with different locations of the
rpm database compared to the host and the image rpmdb location.
The patch imports the given keys depending on the install phase
to the host rpm database and to the image rpm database.

### DIFF
--- a/kiwi/utils/rpm_database.py
+++ b/kiwi/utils/rpm_database.py
@@ -81,6 +81,17 @@ class RpmDataBase(object):
     def write_config(self):
         self.rpmdb_image.write_config()
 
+    def import_signing_key_to_image(self, key):
+        """
+        Import the given signing key to the image rpm database
+        """
+        Command.run(
+            [
+                'rpm', '--root', self.root_dir, '--import', key,
+                '--dbpath', self.rpmdb_host.expand_query('%_dbpath')
+            ]
+        )
+
     def set_database_to_host_path(self):
         """
         Setup dbpath to point to the host rpm dbpath configuration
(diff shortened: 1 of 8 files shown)


schaefi:
Hrm, not correct... needs modifications in the bootstrap phase

kiwi/utils/rpm_database.py
```diff
@@ -81,6 +81,16 @@ def set_macro_from_string(self, setting):
     def write_config(self):
         self.rpmdb_image.write_config()
 
+    def import_signing_key_to_image(self, key):
+        """
+        Import the given signing key to the image rpm database
+        """
+        Command.run(
+            [
+                'rpm', '--root', self.root_dir, '--import', key
```
Conan-Kudo: You should probably make sure this knows the dbpath, otherwise it might not work properly in some circumstances...

schaefi: That import statement takes the host _dbpath setup into account which is what I would have expected and which works well together with the bootstrap phase and the move of the rpmdb to the final location according to the used distribution. I have tested this with a distro that has rpm on /var/lib/rpm(Leap42.3), a distro with rpm on /usr/lib/sysimage(Leap15) and a distro with rpm on /var/lib/rpm(Fedora28) in all kinds of vice versa setup... my head is spinning :sweat_smile: 

So if you have a clear indicator to pass --dbpath to the import call let me know but if not I think I'll stick to the default behavior which seems consistent across all distros

last famous words :smile:

Conan-Kudo: In mock, host rpm (TW/LP15) could not read the final chroot's installed package list because even with `--root`, it doesn't read the dbpath from the target environment macro. Don't know why, but it doesn't. So it failed until I passed `--dbpath` as well. The target environment had the rpmdb in `/var/lib/rpm`, but of course SUSE distributions use `/usr/lib/sysimage/rpm` now.

schaefi: ok, then let us fix that. Just a sec

schaefi: done

kiwi/utils/rpm_database.py
```diff
@@ -81,6 +81,17 @@ def set_macro_from_string(self, setting):
...
         self.rpmdb_image.write_config()
 
+    def import_signing_key_to_image(self, key):
+        """
+        Import the given signing key to the image rpm database
+        """
+        Command.run(
+            [
+                'rpm', '--root', self.root_dir, '--import', key,
+                '--dbpath', self.rpmdb_host.expand_query('%_dbpath')
```
Conan-Kudo: You need the _image_'s dbpath, not the host's.

Conan-Kudo: You need the _image_'s dbpath, not the host's.

schaefi: This information is not present at the time the signing key gets imported. The process is as follows:

1. there are signing keys, import them to the image root using the host rpm. At this point the image root is an almost empty directory. Thus this causes the init of a new rpm database which would be normally done at the first invocation of a rpm based package manager.

2. bootstrap install, rpm now has the signing keys available, the install of rpm from the distro could now introduce a new location of the rpm database. This is the first time I would know about that location, not before

3. switch to the new root, doing this involves the move/rebuild of the rpm database to be correct for the distribution we build

4. continue the rest of the install as chroot operation

I have tested this processing as mentioned with all major rpm based distros. So far this concept works nicely

Conan-Kudo: Okay. 👍

PR #969, opened 2019-02-26
Split of a base class from TestDiskFormatVagrantBase

dcermak:
Fixes huge number of changes from #947.

* move the initialization of the testing Mocks into `VagrantTestBase` 
* provide an initialization of the mocks for `test_create_image_format()` in that base class too

This removes code duplication from the unit tests.

### DIFF
--- a/test/unit/storage_subformat_vagrant_base_test.py
+++ b/test/unit/storage_subformat_vagrant_base_test.py
@@ -13,7 +13,8 @@ from kiwi.storage.subformat.vagrant_base import DiskFormatVagrantBase
 from textwrap import dedent
 
 
-class TestDiskFormatVagrantBase(object):
+class VagrantTestBase(object):
+
     def setup(self):
         xml_data = Mock()
         xml_data.get_name = Mock(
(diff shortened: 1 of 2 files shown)


test/unit/storage_subformat_vagrant_base_test.py
```diff
@@ -28,6 +29,29 @@ def setup(self):
...
+            values that were written to ``open``'d files
+        """
+        mock_rand.return_value = 0xa
+        mock_mkdtemp.return_value = 'tmpdir'
+
+        mock_open.return_value = MagicMock(spec=io.IOBase)
+        file_handle = mock_open.return_value.__enter__.return_value
+
+        return file_handle
+
```
schaefi: So sorry but I don't like this concept. I think it does not belong here. What you are trying to do is to reduce code duplication which is great but the code here is out of scope and has no relation to the vagrant base class testing. So I see two things:

1. you want to define init values for the rand and mkdtemp mocks. This is nice an generically useful. So there is ```test/unit/test_helper.py``` which should be the location for such helpers. However I would see that one in the scope of consistency and not so much helping to reduce code duplication

2. You want to create the file_handle mock as return to the open mock. I like this as it is used very often. You could consider another method to test_helper to serve you better here

config_create_image_format_mocks that reads like "do something" and what something is can only be understood by reading the code. And why this code exists can only be understood by reading and understanding the test. imho we shouldn't do that

test/unit/storage_subformat_vagrant_libvirt_test.py
```diff
@@ -1,29 +1,16 @@
...
-        self.xml_state = Mock()
-        self.xml_state.xml_data = xml_data
-        self.xml_state.get_image_version = Mock(
-            return_value='1.2.3'
-        )
-        self.vagrantconfig = Mock()
-        self.vagrantconfig.get_virtualsize = Mock(
-            return_value=42
-        )
+        super(TestDiskFormatVagrantLibVirt, self).setup()
```
schaefi: ok, calling the base class constructor via super(). I know many people like this and I'm ok with it for the sake of handling the unit testing. So ok let's keep it.

```
class TestDiskFormatVagrantLibVirt(TestDiskFormatVagrantBase):
```

From an oo design perspective and the loosely coupled code idea, which I really prefer, super is like hell. That simple line creates a hard coupling to the base class and often opens the question if it really makes sense. The back call to the base on constructing usually leads to big headaches if there is any change to the base. It usually leads people to not touching the code because they fear the consequences. That's why I really appreciate what Sandy wrote in https://www.poodr.com/

Tests are also code, yes but at that level I accept a certain amount of code duplication. I have no problem writing a test and using the same init code again which I have used before in another test if I can avoid a coupling of test code. This is because I want to have full freedom to do changes in the implementation but if the tests start to try to become clever they will stand in my way for implementation refactoring.

It might sound boring or wrong to you but I have learned my lessons with super() :)

test/unit/storage_subformat_vagrant_base_test.py
```diff
@@ -13,7 +13,8 @@
 from textwrap import dedent
 
 
-class TestDiskFormatVagrantBase(object):
+class VagrantTestBase(object):
```
schaefi: No renaming of the class please, see reasons below

dcermak:
I'll create a new pull request to remove some of the code duplication as its scope will be changed completely and neither the title nor the original description would be appropriate any more.

PR #971, opened 2019-02-27
Write sha256sum --check compatible shasum format

schaefi:
Change the output format of the bundler shasum file to be
compatible with a 'sha256sum --check' call.
This fixes bsc#1127173


schaefi:
Tested with

```
ms@d23:/tmp/mytest-bundle
> sha256sum --check LimeJeOS-Leap-15.0.x86_64-1.15.0-foo.iso.sha256
LimeJeOS-Leap-15.0.x86_64-1.15.0-foo.iso: OK
```

The file reference is added as basename, no file path reference exists. This means the check called must be issued inside of the directory the file exists. This is by intention

PR #973, opened 2019-02-28
Rework package manager selection to use DNF when YUM is requested

### LINKED ISSUE #972: YUM implementation needs to be removed
## Problem description
[YUM v3 is gone in Fedora 31](https://fedoraproject.org/wiki/Changes/Retire_YUM_3).

With DNF now available in RHEL/CentOS 7 and SLE 15 in some form, it's acceptable to just go ahead and rip it out. RHEL 8 is calling DNF "YUM v4", so the expectation seems to be that the YUM name will live on...

## Expected behavior
Using `yum` should silently use `dnf`.

N/A

+ KIWI version: 9.17.19
+ Operating system: Fedora 31
+ OBS version: N/A

Conan-Kudo:
YUM v3 is no longer developed, having been superseded by DNF for
several years. With DNF now available as a usable package manager
in Red Hat Enterprise Linux 7 through the Extras channel and
SUSE Linux Enterprise 15 through PackageHub, there is no reason for
keeping support for YUM v3 around.

We are keeping support for requesting YUM because in Red Hat Enterprise
Linux 7, DNF is referred to as "YUM v4", and it is simply referred to
as "YUM" in Red Hat Enterprise Linux 8. To avoid confusion from people,
we're just going to leave it in place as an alias to the DNF package
manager.

As for why this is happening now, Fedora is retiring YUM v3 in
Fedora 31, so we might as well get it over with and cut over now.

Reference: https://fedoraproject.org/wiki/Changes/Retire_YUM_3

Fixes #972.

* Deletes the implementation of the YUM class and reworks it to inherit from the DNF class


Conan-Kudo:
@schaefi The only question I have here is how do I handle the unit tests? If I just delete the files, it'll complain because there will be two classes without tests. If I don't get rid of the tests somehow, it's potentially going to fail due to yum-specific things being tested...

kiwi/repository/yum.py
```diff
@@ -15,343 +15,14 @@
...
-from kiwi.defaults import Defaults
-from kiwi.logger import log
-from kiwi.repository.base import RepositoryBase
-from kiwi.utils.rpm_database import RpmDataBase
-from kiwi.path import Path
+from kiwi.repository.dnf import RepositoryDnf
 
 
-class RepositoryYum(RepositoryBase):
+class RepositoryYum(RepositoryDnf):
```
aplanas: I wonder if instead of inheritance you want to use assignment here:

```python
RepositoryYum = RepositoryDnf
```

Conan-Kudo: I'm not sure if the semantics of that are exactly what we would expect, so I went with the conservative approach and changed it to be a subclass.

schaefi:
@Conan-Kudo Hmm, wouldn't it be much easier to go and edit the Factory:

```
kiwi/package_manager/__init__.py
```

and in there, change the class that handles yum

After that just delete all files that does something with yum, also delete the unit tests
and you will be fine

Conan-Kudo:
@schaefi That's a way better plan! I didn't think of that. 👍

Conan-Kudo:
@schaefi This should now be in good shape to merge!

PR #976, opened 2019-03-01
Ensure DNF is installed for kiwi itself and the test CentOS 7 appliance

### LINKED ISSUE #972: YUM implementation needs to be removed

Conan-Kudo:
In e33f53aa4513c38a42736c82db3ec5e0b9da41d4, we switched to DNF when requesting YUM.
This now means we need to ensure DNF is installed for images where
we previously used YUM for that.

Fixes #972.

* Adds `dnf` as a dependency on SUSE distros for kiwi
* Adds `nextgen-yum4` to the bootstrap for the test CentOS 7 appliance

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -130,6 +130,10 @@ Requires:       dnf
 Provides:       kiwi-packagemanager:dnf
 Provides:       kiwi-packagemanager:yum
 %endif
+%if 0%{?suse_version}
+# If it's available, let's pull it in
+Recommends:     dnf
+%endif
 %if 0%{?fedora} >= 26 || 0%{?suse_version}
 Requires:       zypper
 Provides:       kiwi-packagemanager:zypper
@@ -207,11 +211,13 @@ Requires(postun): chkconfig
 Requires:       qemu-img
 Requires:       squashfs-tools
 Requires:       gdisk
-%endif
-%if 0%{?fedora} || 0%{?rhel}
 Requires:       dnf
 Provides:       kiwi-packagemanager:dnf
 Provides:       kiwi-packagemanager:yum
+%endif
+%if 0%{?suse_version}
+# If it's available, let's pull it in
+Recommends:     dnf
 %endif
 %if 0%{?fedora} >= 26 || 0%{?suse_version}
 Requires:       zypper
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
@seife I didn't realize anyone was still using the master branch with a SLE 12 host, since it's available in some form on SUSE Linux 15 platform...

Conan-Kudo:
@seife I've adjusted it to trigger on `%suse_version >= 1500`

seife:
SLES15 also only has it in the obscure "backports" repo, which I would never enable on production systems :-)

In general, it's only needed if you build a RH/CentOS/Fedora image on a SUSE machine, right?
That's something many people never do (OBS installs the same OS that will be in the image to build the image).
So a "recommends" or a hint in the documentation, maybe a runtime check would be much better than blowing up RPM requires IMVHO.

package/python-kiwi-spec-template
```diff
@@ -208,7 +210,7 @@ Requires:       qemu-img
 Requires:       squashfs-tools
 Requires:       gdisk
 %endif
-%if 0%{?fedora} || 0%{?rhel}
+%if 0%{?fedora} || 0%{?rhel} || 0%{?suse_version} >= 1500
```
schaefi: I talked with the release people and they said we shouldn't hard require dnf. The reason is on the media and products there will be zypper only on suse products. suse products build with kiwi would then require dnf to be present too. This won't be accepted on next submission.

So we can set:

```
Recommends: dnf
```

for the suse package only

Sorry I didn't thought about this when talked on rio

Conan-Kudo:
@seife It's an unfortunately common case from what I've been told (and in my experience), due to OBS stripping extended rpm-md (comps, patterns, products, modules, etc.) which makes building images difficult when that data is needed.

seife:
BTW: I'm using recent release (not master, but the tagged ones) in my own OBS instance, so "it is available in build.opensuse.org" is not a valid argument for my case :-)

package/python-kiwi-spec-template
Conan-Kudo: Fixed.

schaefi:
The recommend setup looks good to go for me

@seife last chance to object :-) but I think that should also work for you

seife:
Recommends will work for me :-)

PR #978, opened 2019-03-04
pxe: use dracut's network code, drop own setup

seife:
* use dracut's networking code for PXE installation
* drop our own setup of lan0, which has problems with multiple interfaces etc.


schaefi:
@seife Looks great to me.

Now that we don't put on a fixed interface name, users have to make the decision for IPAPPEND
or the ip= setup such that the client gets a network. I think at least a short summary of the conversation by mail should end up here in the docs. Could you add that as well ? Would be great

Thanks

schaefi:
I think a good place for the doc update would be in:

* doc/source/building/build_oem_disk.rst

__4. Add/Update the kernel command line parameters__

seife:
No, that's not necessary, see the part in parse-kiwi-instal.sh

It works without BOOTIF and any special ip= parameters, just rd.kiwi.install.pxe

It works better with BOOTIF= (faster, if the first interfaces do not have link), but it also works without.
Tested here with a VMware VM that has 2 interfaces, first without LAN connected, second interface that boots the VM.

schaefi:
ok I saw the code adds

```
ip=dhcp
```

if there is no interface setup available. But this is an implicit setup and I think worth a short info in the docs that we try to run dhcp on the first interface found... and also that this is not deterministic because which interface is first is not guaranteed to be the same on every boot.

Wouldn't that be useful ?

seife:
Well, it does not matter. Dracut, with "neednet=1 ip=dhcp", which is implicit here if no custom options are given, tries all interfaces until it finds one it can use.

The old code did "try the first interface, but 'first' is more or less random, name it 'lan0' and try to set it up, but it's pure luck if this interface will work or not, if it does not work just error out"

So for the user, the situation with multiple interfaces should always improve, at least in all cases I can imagine, without any further need to add additional options.

I can of course add a snippet to the documentation explicitly stating that, without further configuration, all interfaces will be tried and that BOOTIF=... can be used to speed things up.

seife:
I took a look at the documentation and actually I'd rather provide an extra chapter about how the network boot process for pxe oem deployment works and how to debug specifics. It does not really fit into the existing chapter which is in the "build oem images" part of the documentation, and it does deserve a little bit more thorough explanation and documentation (there are also plain errors, e.g. that the kernel is always loaded via tftp ;-), https://github.com/seife/kiwispray is a project that explicitly avoids having to use tftp for anything with kiwi, but that's a minor and different issue)

schaefi:
>  I'd rather provide an extra chapter about how the network boot process for pxe oem deployment works

ok that works for me too. There is #981 which can be used as a reference and improvements on the doc side are pretty much appreciated.

From a code perspective what you did is all great, I'll merge this one now as it's a general improvement to the situation we had before

Thanks much

PR #980, opened 2019-03-05
Extension of the Gitlab CI pipeline

dcermak:
Fixes: Migration to GitLab CI.

* Add a multistage gitlab CI pipeline:
1. Test stage: this stage contains jobs that run the unit tests. I have added two jobs that use the recommended development environment via virtualenv & run the tests via tox and two additional jobs (one for Tumbleweed and the other for Fedora 29) that use the packages provided by the system's package manager.
2. Build doc stage: this stage builds the documentation (man pages, dirhtml, pdf) and runs the linkcheck & spell checks. Note though that the spell check reports 1001 spelling mistakes but doesn't cause a build failure, so I am not sure whether it's worth keeping. I have to test whether the linkchecker actually causes a failure when a link is dead. This stage is currently run in a Fedora container, as Tumbleweed lacks the sphinxcontrib-spelling package (and python-xdist is too outdated).
   The doc stage passes the built documentation further to the package stage.
3. Package stage: Here I added a Fedora container, that builds a source RPM and builds it locally via mock. This gives us immediate feedback, whether an accidental change broke the rpm build. This is run inside a Fedora container because osc cannot work without a repository and an account, which I don't think is worth adding via secret variables. A possible extension would be to try to install the resulting packages in the next stage.
4. Deploy stage: This stage copies the html documentation to GitLab pages (but only on the master branch, so I couldn't test this yet). This can be used as a "nightly" documentation (this would need to be enabled on GitLab) or we can just drop it.

The result looks like this: https://gitlab.com/D4N/kiwi-ci/pipelines/50328107

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,14 +1,78 @@
-test:
-  image: "python:3.6-stretch"
+stages:
+  - test
+  - package
+
+.install_deps: &install_system_deps
+  before_script:
+    - >
+      dnf install -y --refresh python3 python3-devel 'python3dist(pip)'
+      'python3dist(tox)' make gcc which xz xorriso libxml2-devel libxslt-devel
+      enchant genisoimage ShellCheck
+
+tox_check:
+  stage: test
+  image: fedora:latest
+  variables:
+    LANG: en_US.UTF-8
+  <<: *install_system_deps
   script:
-    - apt-get update -qq
-    - apt-get install -y git
-    - apt-get install -y xsltproc
-    - apt-get install -y genisoimage
-    - apt-get install -y enchant
-    - apt-get install -y shellcheck
-    - apt-get install -y python3-pip
-    - pip3 install --upgrade pip
-    - pip install tox
-    - tox -e check
+    - dnf install -y python36 python34 python2 python2-devel python2-virtualenv python2-pip
+    # Python 2.7
+    - export PYTHON=python2.7
+    - tox -e unit_py2_7 "-n $(nproc)"
+    # Python 3.4
+    - export PYTHON=python3.4
+    - tox -e unit_py3_4 "-n $(nproc)"
+    # Python 3.6
+    - export PYTHON=python3.6
     - tox -e unit_py3_6 "-n $(nproc)"
+    # Python 3.7
+    - export PYTHON=python3.7
+    - tox -e 'unit_py3_7,check' "-n $(nproc)"
+  cache:
+    key: "$CI_JOB_NAME"
+    paths:
+      - .tox
+
+build_doc:
+  stage: test
+  image: fedora:latest
+  variables:
+    TOXENV: 'packagedoc'
+    PYTHON: 'python3.7'
+  <<: *install_system_deps
+  script:
+    - >
+      dnf -y install latexmk texlive-cmap texlive-metafont texlive-ec
+      texlive-babel-english texlive-fncychap texlive-fancyhdr texlive-titlesec
+      texlive-tabulary texlive-framed texlive-wrapfig texlive-parskip
+      texlive-upquote texlive-capt-of texlive-needspace texlive-makeindex
+      texlive-times texlive-helvetic texlive-courier texlive-gsftopk
+      texlive-updmap-map texlive-dvips
+    - tox
+  artifacts:
+    paths:
+      - doc/build/
+  cache:
+    key: "$CI_JOB_NAME"
+    paths:
+      - .tox/
+      - doc/build/
+
+RPM:
+  stage: package
+  image: fedora:latest
+  before_script:
+    - dnf --refresh -y install mock perl-TimeDate which make gzip tar perl git
+  script:
+    - 'sed -i "s|build: clean tox|build:|" Makefile'
+    - make build
+    - mv dist/python-kiwi.spec .
+    - rm dist/python-kiwi.changes
+    - mock --old-chroot -r /etc/mock/fedora-29-x86_64.cfg --buildsrpm --sources ./dist --spec ./python-kiwi.spec 2>&1 | tee srpm_build_out
+    - "export SRC_RPM=$(ls $(grep 'INFO: Results ' srpm_build_out | awk -F ':' '{print $3}')/*.src.rpm)"
+    - mv $SRC_RPM .
+    - mock --old-chroot -r /etc/mock/opensuse-tumbleweed-x86_64.cfg $(basename $SRC_RPM)
+    - mock --old-chroot -r /etc/mock/fedora-29-x86_64.cfg $(basename $SRC_RPM)
+  dependencies:
+    - build_doc
--- a/README.rst
+++ b/README.rst
@@ -3,6 +3,8 @@ KIWI - Next Generation
 
 .. |Build Status| image:: https://travis-ci.com/SUSE/kiwi.svg?branch=master
    :target: https://travis-ci.com/SUSE/kiwi
+.. |GitLab CI Pipeline| image:: https://gitlab.com/schaefi/kiwi-ci/badges/master/pipeline.svg
+   :target: https://gitlab.com/schaefi/kiwi-ci/pipelines
 .. |Health| image:: https://api.codacy.com/project/badge/Grade/8ebd8ce362294fabb0870f50358e564f
    :target: https://www.codacy.com/app/Appliances/kiwi?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=SUSE/kiwi&amp;utm_campaign=Badge_Grade
 .. |Doc| replace:: `Documentation <https://opensource.suse.com/kiwi>`__
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -60,6 +60,7 @@ Source:         %{name}.tar.gz
 Source1:        %{name}-rpmlintrc
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 %if 0%{?fedora} || 0%{?suse_version}
+BuildRequires:  gcc
 BuildRequires:  python3-devel
 %endif
 %if 0%{?fedora} || 0%{?suse_version} || 0%{?ubuntu} >= 1804 || 0%{?debian} >= 9


schaefi:
Hi Dan, I also started fiddling with this. I think you already did much more but could you double check the changes I have in this branch: ```gitlab_ci_integration``` such that we can decide which branch to continue. I think it will be yours as I had to stop my effort in the  middle of the work because of other issues here at work.

Thanks

dcermak:
I see that you have experimented a lot too (I have rebased my changes frequently, so they are not really visible in the git log, but you can see the list of builds that ran here: https://gitlab.com/D4N/kiwi-ci/pipelines). 

The final result of your `.gitlab-ci.yml` contains one job, that is very similar to the `tox_37_check` that I have added (although you are running it with Python 3.6 and I am using Python 3.7).

Nevertheless, we don't have to add my monstrous pipeline (the downside is that it takes a while to run), that's why I have marked this as a draft.

.gitlab-ci.yml
```diff
@@ -1,14 +1,108 @@
...
+  - package
+  - deploy
+
+.fedora_install_deps: &fedora_install_deps
+  before_script:
+    - dnf install -y --refresh python3 python3-devel python3-pip
+    # requirements
+    - dnf install -y python3-docopt python3-lxml python3-pyyaml python3-pyxattr python3-six python3-future python3-requests
+    # dev-requirements
+    - dnf install -y python3-pytest python3-pytest-xdist python3-pytest-cov python3-mock python3-flake8 python3-sphinx python3-sphinx_rtd_theme.noarch python3-sphinxcontrib-spelling python3-enchant python3-tox make gcc which genisoimage xz
```
Conan-Kudo: We're generally using `xorriso` instead of `genisoimage` for kiwi.

.gitlab-ci.yml
```diff
@@ -1,14 +1,108 @@
...
+  - test
+  - doc
+  - package
+  - deploy
+
+.fedora_install_deps: &fedora_install_deps
+  before_script:
+    - dnf install -y --refresh python3 python3-devel python3-pip
+    # requirements
+    - dnf install -y python3-docopt python3-lxml python3-pyyaml python3-pyxattr python3-six python3-future python3-requests
```
Conan-Kudo: Fedora has generic `python3dist(modname)` Provides for all Python modules, in case you might prefer to use those here: https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_automatic_provides_with_a_standardized_name

.gitlab-ci.yml
```diff
@@ -1,14 +1,108 @@
...
+    - python3 setup.py build
+    - export PYTHONPATH=$(pwd)
+    - cd test/unit/ && pytest-3.7 --no-cov-on-fail --cov=kiwi --cov-report=term-missing --cov-fail-under=100 --cov-config .coveragerc -n $(nproc)
+
+tox_37_check:
+  stage: test
+  image: opensuse/tumbleweed
+  before_script:
+    - zypper --non-interactive refresh
+    - zypper --non-interactive install libxml2-devel libxslt-devel gcc python3 python3-devel python3-pip enchant ShellCheck which mkisofs xz gzip
```
Conan-Kudo: Fedora ships tox environment for multiple Python versions, so you could have one job template that does it: https://developer.fedoraproject.org/tech/languages/python/multiple-pythons.html

.gitlab-ci.yml
dcermak: I just tried that. The tests fail when genisoimage is not installed: https://gitlab.com/D4N/kiwi-ci/-/jobs/171908894.

Is that expected?

Conan-Kudo: No. That seems broken... @schaefi?

schaefi: I fixed that part in #c4f0855d1f160070982b0609d2cea831751cdc47

dcermak: That's really odd, since that commit is included in this PR.

schaefi: Hmm, I see I didn't catch this case. So there are two unit tests that actually test the content list of an iso for the cdrtools case. This tests read the example iso file using isoinfo and expects the tool to be present. I think we need a @pytest.mark.skipif there is no genisoimage installed. So that needs fixing

schaefi: https://github.com/SUSE/kiwi/pull/982

that should make it work

dcermak: I'll rebase my branch on master once #982 get's merged.

.gitlab-ci.yml
dcermak: I have added tox environments for Python 3.4 and 3.6 using the Fedora containers, but kept the openSUSE containers for Python 2.7 and 3.7.

.gitlab-ci.yml
dcermak: I have changed these accordingly.

test/data/example_dot_profile_config.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<image schemaversion="6.9" name="LimeJeOS-openSUSE-13.2" displayname="schäfer">
+<image schemaversion="6.9" name="LimeJeOS-openSUSE-13.2" displayname="schaefer">
```
Conan-Kudo: Why is this changing everywhere?

.gitlab-ci.yml
```diff
@@ -1,14 +1,139 @@
...
+  before_script:
+    - dnf -y install fedpkg mock perl-TimeDate which
+  script:
+    - 'sed -i "s|build: clean tox|build:|" Makefile'
+    - make build
+    - cd dist
+    - fedpkg --release f29 srpm
+    - export SRC_RPM=$(ls *src.rpm)
+    - mock --old-chroot -r /etc/mock/opensuse-tumbleweed-x86_64.cfg $SRC_RPM
+    - mock --old-chroot -r /etc/mock/fedora-29-x86_64.cfg $SRC_RPM
```
Conan-Kudo: Mock can build using a spec file directly if you give it `--sourcedir` and `--spec` instead of the `$SRC_RPM` argument. I advise that the `--sourcedir` only contain the sources that the spec would reference to make it faster for copying into the build environment.

test/data/example_dot_profile_config.xml
dcermak: For some obscure reason, the builds with Python 3.4 that I added fail on GitLab, see for instance: https://gitlab.com/D4N/kiwi-ci/-/jobs/171932726.

Python complains about not being able to decode the character `\xe4` (that is "ä" in one of iso-something encodings). Thus I removed the "ä" from @schaefi's name and the tests now pass.

I can of course just remove the Python 3.4 environment and drop this commit.

Conan-Kudo: Just drop Python 3.4.

.gitlab-ci.yml
Conan-Kudo: It's been merged now, so this should be rebased and fixed :)

schaefi: *grmbl* I take a look

schaefi: https://github.com/SUSE/kiwi/pull/990

This should finally make the tests happy

.gitlab-ci.yml
dcermak: This option is no longer/not yet present? Mock can build SRPMs via very similar options though.

dcermak: I have used the SRPM build option, so now fedpkg is no longer required.

dcermak:
@schaefi Can you restart the Fedora build on Gitlab? dnf install timed out...

.gitlab-ci.yml
```diff
@@ -1,14 +1,131 @@
...
+    - zypper --non-interactive install python3-docopt python3-lxml python3-PyYAML python3-xattr python3-six python3-future python3-requests
+    # dev-requirements
+    - zypper --non-interactive install python3-pytest python3-pytest-cov python3-mock python3-flake8 python3-Sphinx python3-sphinx_rtd_theme python3-pyenchant python3-tox make gcc which mkisofs xz gzip
+    # pytest-xdist from the repos is broken, use the one from PyPI instead
+    - pip3 install sphinxcontrib-spelling pytest-xdist
+  <<: *distro_test
+
+Fedora:
+  stage: test
+  image: fedora:29
```
Conan-Kudo: Do we want to do `fedora:29` or `fedora:latest`? We are using `opensuse/tumbleweed` for the openSUSE jobs, so we should do the same on the Fedora side...

dcermak: Well, Tumbleweed is unversioned by its nature. Although I have to admit that I added the 29 because I though that `latest` would be rawhide (now that I checked I know that is not true).

Anyway, my thoughts behind this are that the CI should be somewhat reproducible and pulling in the latest image means that on every Fedora switch it could break (because package names change, dependencies won't resolve, etc). Using a fixed version means that we consciously upgrade the CI and it doesn't happen accidentally/automatically.

Conan-Kudo: Then we should probably be using `opensuse/leap:15.0` then?

dcermak: Yeah, according to my logic we should, but Leap is really old... I'll use `:latest` then.

.gitlab-ci.yml
```diff
@@ -1,14 +1,131 @@
...
-    - apt-get install -y python3-pip
-    - pip3 install --upgrade pip
-    - pip install tox
-    - tox -e check
-    - tox -e unit_py3_6 "-n $(nproc)"
+stages:
+  - test
+  - doc
+  - package
+  - deploy
```
schaefi: For the first iteration can we please leave out the deployment stage ? The reason is this is done by travis on tags in master and another deployment system will interfere with it. I think if we move the deployment of the doc pages and the package on pip from travis to gitlab-ci it should be done in an extra pull request that deletes the stage in .travis.yml and adds one in .gitlab-ci.yml. Thanks

.gitlab-ci.yml
```diff
@@ -1,14 +1,131 @@
...
+    - make man SPHINXBUILD=/usr/bin/sphinx-build-3
+    - make latexpdf SPHINXBUILD=/usr/bin/sphinx-build-3
+    # FIXME: there is a lot of warnings here, but they are not fatal
+    # => remove this maybe
+    - make spell SPHINXBUILD=/usr/bin/sphinx-build-3
+  artifacts:
+    paths:
+      - doc/build/
+
+RPM:
```
schaefi: I'm not sure if the RPM test makes sense. Since the packaging for the suse distro and also the packaging done by @Conan-Kudo is probably different compared to that test target ?

.gitlab-ci.yml
dcermak: Sure, the deployment stage doesn't really do a lot anyway. It just publishes the current docs from master to gitlab pages, there is no interference with the official docs or the pip package.

.gitlab-ci.yml
Conan-Kudo: It's not _that_ different. We _should_ make sure that it packages fine there anyway since you do build it in the OBS project for RH/Fedora distributions.

dcermak: My guess is that @Conan-Kudo uses very similar tools, but maybe patches the spec file a little.

Your call whether you want to have this: on one hand we get an extra safety that the spec file is functional before you publish on OBS, on the other hand it takes quite some time to build and it uses mock instead of obs-build (which I don't know how to include here).

.gitlab-ci.yml
```diff
@@ -1,14 +1,116 @@
...
+  script:
+    - dnf -y install latexmk texlive-cmap texlive-metafont texlive-ec texlive-babel-english texlive-fncychap texlive-fancyhdr texlive-titlesec texlive-tabulary texlive-framed texlive-wrapfig texlive-parskip texlive-upquote texlive-capt-of texlive-needspace texlive-makeindex texlive-times texlive-helvetic texlive-courier texlive-gsftopk texlive-updmap-map texlive-dvips
+    - cd doc
+    - make html SPHINXBUILD=/usr/bin/sphinx-build-3
+    - make linkcheck SPHINXBUILD=/usr/bin/sphinx-build-3
+    - make man SPHINXBUILD=/usr/bin/sphinx-build-3
+    - make latexpdf SPHINXBUILD=/usr/bin/sphinx-build-3
+    # FIXME: there is a lot of warnings here, but they are not fatal
+    # => remove this maybe
+    - make spell SPHINXBUILD=/usr/bin/sphinx-build-3
```
schaefi: The cd doc and all the make targets should be replaced by calling

```
tox -e doc
```

.gitlab-ci.yml
```diff
@@ -1,14 +1,116 @@
...
+    # requirements
+    - dnf install -y 'python3dist(docopt)' 'python3dist(lxml)' 'python3dist(pyyaml)' 'python3dist(pyxattr)' 'python3dist(six)' 'python3dist(future)' 'python3dist(requests)'
+    # dev-requirements
+    - dnf install -y 'python3dist(pytest)' 'python3dist(pytest-xdist)' 'python3dist(pytest-cov)' 'python3dist(mock)' 'python3dist(flake8)' 'python3dist(sphinx)' python3-sphinx_rtd_theme 'python3dist(sphinxcontrib-spelling)' 'python3dist(pyenchant)' 'python3dist(tox)' make gcc which xz xorriso
+
+.distro_test: &distro_test
+  script:
+    - python3 setup.py build
+    - export PYTHONPATH=$(pwd)
+    - cd test/unit/ && pytest-3.7 --no-cov-on-fail --cov=kiwi --cov-report=term-missing --cov-fail-under=100 --cov-config .coveragerc -n $(nproc)
```
schaefi: I think I don't understand why we have the code from tox.ini that calls pytest running copied and called in extra test here ? Can you explain the motivation ?

schaefi:
Overall what I'd like to avoid is code duplication with regards to tox.ini and the stuff in the gitlab-ci.yml

.gitlab-ci.yml
Conan-Kudo: Aside from a single patch that changes xattr for pyxattr (which you can do with a targeted sed call), my packaging is more or less the same. The only other major change is that I ship the unversioned binaries in a kiwi-cli subpackage that `Provides: kiwi = %{version}-%{release}` so that OBS will always use this version instead of some legacy version.

.gitlab-ci.yml
dcermak: It's left over from my initial experimentation with gitlab-ci. And IIRC I had some problems with pytest on Fedora where `pytest` doesn't exist, but `pytest-3.7` does.

But yes, this is something that I wanted to remove (and forgot about it).

Conan-Kudo: Fedora doesn't _generally_ use alternatives to manage the unversioned names of the binaries.

Conan-Kudo:
@dcermak Can you please squash this down, the change set is a bit unwieldy...

gitlab-tox.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -ex
+
+${PYTHON} -m venv ${VENV}
+source ${VENV}/bin/activate
+pip install -r .virtualenv.dev-requirements.txt
+${PYTHON} setup.py develop
+tox "-n $(nproc)"
+deactivate
```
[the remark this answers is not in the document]
dcermak: > You are basically calling tox in a venv this is not required nor recommended.

I copied those steps from the documentation, so then the docs should be fixed.

> Here is how I would have done it, maybe we can combine this. Also setting LANG to an utf-8 locale avoids the multibyte hacks
>
>    #1002
>
> Thoughts ?

That's essentially the same as mine, minus the `RPM` stage. @Conan-Kudo do you want that one to be present? If not, then let's use yours, as it's going to be what you want.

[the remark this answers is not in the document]
Conan-Kudo: I would _like_ the packaging to be tested, but @schaefi seems to be highly against it... We probably need a step where we merge the changelog back into the spec, though.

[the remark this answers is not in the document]
schaefi: No no I'm not against it. All here is fine. I just added the portion with the tox calls as it seems I couldn't get my point clear.

> I copied those steps from the documentation, so then the docs should be fixed.

Those steps are fine to create your own development environment. Nothing wrong with it. But calling the test is covered through tox. The chapter you were reading has the section

```
Development and Contributing
============================

.. hint:: **Abstract**

   This document describes the development process of KIWI
   and how you can be part of it.
```

and later in the document:

```
Running Test Cases
------------------

For running test cases, the preferred method is to use Tox. The Tox
execution environment can be used to run any kind of target, tests are
just one, documentation is another one. Refer to :file:`tox.ini` for more
details. Tox itself creates a python virtual environment for each tox
target below :file:`./.tox`.

Before you start to contribute code make sure all tests pass by calling
the following command:

.. code:: bash

    $ tox
```

I mean isn't that good enough ?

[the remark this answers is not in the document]
Conan-Kudo: It probably is, if we were using all the tox envs configured. If we are, then yes.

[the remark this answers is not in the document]
schaefi: So if you agree to take my way to run tests just with tox and add it here, I'm fine with the RPM test and the other changes (except for the multibyte hack that is not needed if you set LANG, also see my PR)

Does this make sense ?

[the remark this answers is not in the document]
dcermak: Concerning the docs: the way they are currently structured *suggests* (it is not explicitly written there) that one should create a virtualenv and then run tox. I admit that I haven't really paid attention to the sentence `Tox itself creates a python virtual environment for each tox
target below :file:./.tox` and just copied the shell commands from that page. I'll create a separate PR for this.

I'll update this PR to remove the multibyte hack.

.gitlab-ci.yml
```diff
@@ -1,14 +1,78 @@
...
+RPM:
+  stage: package
+  image: fedora:latest
+  before_script:
+    - dnf --refresh -y install mock perl-TimeDate which make gzip tar perl git
+  script:
+    - 'sed -i "s|build: clean tox|build:|" Makefile'
+    - make build
+    - mv dist/python-kiwi.spec .
+    - rm dist/python-kiwi.changes
```
Conan-Kudo: We should test the conversion of changes into a changelog... Perhaps we can do that later, but we should consider that...

dcermak: This isn't really urgent, so we can merge #997 first  and then I'll rebase this branch on master. However, I am not sure if the current RPM stage will catch any issues with the changelog, as it is using the tools from Fedora (since I don't know how to run osc without credentials).

I can run rpmlint on the result though.

Conan-Kudo: @dcermak You can install `obs-build` and do the following:

```bash
$ /usr/lib/build/changelog2spec dist/python-kiwi.changes >> dist/python-kiwi.spec
```

schaefi: @Conan-Kudo could you create an issue with your idea in this regard ? I would put that as an extra PR and continue getting this one merged. It looks really good now

Conan-Kudo: Will do.

PR #982, opened 2019-03-05
Run cdrtools based unit tests on demand

schaefi:
Only if the underlaying system provides legacy cdrtools,
run the unit tests which actually reads the test data iso
using those tools.

### DIFF
--- a/test/unit/iso_tools_iso_test.py
+++ b/test/unit/iso_tools_iso_test.py
@@ -1,18 +1,23 @@
-from mock import call, patch
+from builtins import bytes
+from mock import (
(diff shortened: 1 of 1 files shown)


schaefi:
References issue found in #980

dcermak:
@schaefi Will we reach the required test coverage when cdrtools are not installed?

schaefi:
yes pytest.skipif is a way to skip tests without influencing the coverage

dcermak:
> yes pytest.skipif is a way to skip tests without influencing the coverage

Oh, that's handy!

PR #985, opened 2019-03-06
Fix zypper rpmdb linkage

schaefi:
This patch is two fold

__Fixup zypper/suse link to rpm database__
    
The compat code generating the /var/lib/rpm link was only effective if a signing key was specified, however it should be effective in any case

__Care for update alternatives on rpmdb move__
    
In set_database_to_image_path we also have to care for the move of the alternatives path to the new rpmdb location

### DIFF
--- a/kiwi/repository/apt.py
+++ b/kiwi/repository/apt.py
@@ -205,8 +205,9 @@ class RepositoryApt(RepositoryBase):
 
         :param list signing_keys: list of the key files to import
         """
-        for key in signing_keys:
-            self.signing_keys.append(key)
+        if signing_keys:
+            for key in signing_keys:
+                self.signing_keys.append(key)
 
     def delete_repo(self, name):
         """
--- a/kiwi/utils/rpm_database.py
+++ b/kiwi/utils/rpm_database.py
@@ -51,9 +51,14 @@ class RpmDataBase(object):
         """
         Rebuild image rpm database taking current macro setup into account
         """
-        Command.run([
-            'chroot', self.root_dir, 'rpmdb', '--rebuilddb'
-        ])
+        Command.run(
+            ['chroot', self.root_dir, 'rpmdb', '--rebuilddb']
+        )
+
+    def init_database(self):
+        Command.run(
+            ['rpm', '--root', self.root_dir, '--initdb']
+        )
 
     def set_macro_from_string(self, setting):
         """
--- a/kiwi/utils/rpm_database.py
+++ b/kiwi/utils/rpm_database.py
@@ -133,3 +133,16 @@ class RpmDataBase(object):
             self.rpmdb_image.write_config()
             self.rebuild_database()
             self.rpmdb_image.wipe_config()
+
+            root_rpm_host_dbpath = os.path.normpath(
+                os.sep.join([self.root_dir, rpm_host_dbpath])
+            )
+            root_rpm_alternatives = os.sep.join(
+                [root_rpm_host_dbpath, 'alternatives']
+            )
+            if os.path.exists(root_rpm_alternatives):
+                Command.run(
+                    ['mv', root_rpm_alternatives, root_rpm_image_dbpath]
+                )
+
+            Path.wipe(root_rpm_host_dbpath)
--- a/kiwi/repository/dnf.py
+++ b/kiwi/repository/dnf.py
@@ -199,9 +199,10 @@ class RepositoryDnf(RepositoryBase):
 
         :param list signing_keys: list of the key files to import
         """
-        rpmdb = RpmDataBase(self.root_dir)
-        for key in signing_keys:
-            rpmdb.import_signing_key_to_image(key)
+        if signing_keys:
+            rpmdb = RpmDataBase(self.root_dir)
+            for key in signing_keys:
+                rpmdb.import_signing_key_to_image(key)
 
     def delete_repo(self, name):
         """
--- a/kiwi/repository/zypper.py
+++ b/kiwi/repository/zypper.py
@@ -286,8 +286,11 @@ class RepositoryZypper(RepositoryBase):
         :param list signing_keys: list of the key files to import
         """
         rpmdb = RpmDataBase(self.root_dir)
-        for key in signing_keys:
-            rpmdb.import_signing_key_to_image(key)
+        if signing_keys:
+            for key in signing_keys:
+                rpmdb.import_signing_key_to_image(key)
+        else:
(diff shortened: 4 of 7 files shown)


Conan-Kudo:
@schaefi I'm confused why there's a problem with alternatives. `rpm < 4.14` doesn't try to regenerate the rpmdb directory, so alternatives in `/var/lib/rpm/alternatives` shouldn't break. However, `rpm >= 4.14` (in Leap 15.0+) does do that, but in Leap 15.0, alternatives was moved to `/var/lib/alternatives` (as RH/Fedora has done long ago).

schaefi:
The problem here is not rpm but the update-alternatives package. Let me make an example. If you build on host=Leap15 an image=Leap42.3 you have the well known problem host rpm inits db in usr/lib/sysimage/rpm and the rpm in Leap42.3 still uses /var/lib/rpm. We have a workflow and a working solution to handle the move in place as you know

Ok so now there is this update-alternatives package in Leap42.3 which when it gets installed has post scripts that reads macro definitions of rpm to construct its path for the alternatives. That of course ends in the alternative path /usr/lib/sysimage/rpm/alternatives and was not yet handled when we move it to the correct path.

This patch fixes that. The problem was reported by Till on the kiwi google group

Makes more sense now ?

Conan-Kudo:
@schaefi But that's not what's happening. 

SUSE's `updates-alternatives` package in Leap 42.x is hardwired to `/var/lib/rpm/alternatives`: https://build.opensuse.org/package/view_file/openSUSE:Leap:42.1:Update/dpkg/update-alternatives.spec?expand=1

However, in 15.0, it moved to `/var/lib/alternatives` and has a scriptlet to migrate it automatically: https://build.opensuse.org/package/view_file/openSUSE:Leap:15.0/dpkg/update-alternatives.spec?expand=1

This is why I don't understand the logic here.

schaefi:
> This is why I don't understand the logic here.

ok build Leap42.3 on Leap15, during bootstrap you have:

```
ll /tmp/mytest/build/image-root/var/lib/

lrwxrwxrwx 1 root root 26 Mar  6 14:00 rpm -> ../../usr/lib/sysimage/rpm
```

This is because the host rpm has the db_path on /usr/lib/sysimage/rpm.
The symlink is created in the zypper case because zypper has hardcoded references to /var/lib/rpm
During the bootstrap phase the update-alternatives package gets pulled in (by deps)

That package manages among others:

```
rpm -ql update-alternatives
...
/var/lib/rpm/alternatives
```

This means during bootstrap phase the alternatives directory is created (through the link)
in ```/usr/lib/sysimage/rpm/alternatives```

__Now bootstrap has finished__

We now check the __real__ db_path as rpm wants it and find it's ```/var/lib/rpm``` on Leap42.3
We now rebuild the database such that it appears at the path where rpm of this distro wants it.

__That rebuilddb will not move the alternatives path__

This is why the code in this pull request moves the alternatives manually if this is required

Makes sense ?

Conan-Kudo:
Yeah, now I get it. 🤦‍♂️ 

Ugh, this is so dumb. 🤦‍♂️

schaefi:
This are all after effects of moving the rpm database around. I think many people believed that will never happen and trusted /var/lib/rpm to be a constant of the universe. We see that in zypper, we see that in update-alternatives and I'm sure we will stumble over that more often in the future. There is this interconnection between rpm on the host and rpm in the image and I really hope we found 99% of all the bits to deal with it. But I would not put my hand on fire

schaefi:
I wait for feedback from Till and if you don't have objections merge it by then

kiwi/repository/apt.py
```diff
@@ -205,8 +205,9 @@ def import_trusted_keys(self, signing_keys):
 
         :param list signing_keys: list of the key files to import
         """
-        for key in signing_keys:
-            self.signing_keys.append(key)
+        if signing_keys:
+            for key in signing_keys:
+                self.signing_keys.append(key)
```
aplanas: JFYI (as is irrelevant for the real fix) extracting elements from an iterator and adding it to a list can be done in Python with `.extend`. So this can be transformed to:

```python
self.signing_keys.extend(signing_keys or [])
```

or if you like this more (I do):

```python
self.signing_keys.extend(signing_keys if signing_keys else [])
```

kiwi/utils/rpm_database.py
```diff
@@ -133,3 +133,16 @@ def set_database_to_image_path(self):
             self.rpmdb_image.write_config()
             self.rebuild_database()
             self.rpmdb_image.wipe_config()
+
+            root_rpm_host_dbpath = os.path.normpath(
+                os.sep.join([self.root_dir, rpm_host_dbpath])
```
aplanas: Is the use of `os.sep` instead of `os.path.join` intended here? If no there is one detail that needs to be taken into consideration when `rpm_host_dbpath` is an absolute path:

```python
>>> import os
>>> os.path.normpath(os.sep.join(['/var', 'lib/rpm.db']))
'/var/lib/rpm.db'
>>> os.path.normpath(os.sep.join(['/var', '/lib/rpm.db']))
'/var/lib/rpm.db'
>>> os.path.join('/var', 'lib/rpm.db')
'/var/lib/rpm.db'
>>> os.path.join('/var', '/lib/rpm.db')
'/lib/rpm.db'
```

There are situations where if rpm_host_dbpath is absolute you are looking for os.join.path behavior.

kiwi/repository/apt.py
schaefi: yes I thought about it but the other implementations of the interface contain more code that should not run when no signing_keys exists. Thus I put the decision in the individual method implementation. I realized it is the wrong place because the interface suggests a different method. So you can expect a follow up pull request

kiwi/utils/rpm_database.py
schaefi: Nope the use of os.sep is intentional here. That's because I intentionally don't want the intelligence of os.path.join. Your last example demonstrates that. The way it's used in the method
is join(root_dir, some_path). I have to trust the join method to join those two in __any__ case.
Now the os.path.join method is super clever and does not join as I expect it if the argument list of the given join path elements is absolute

```python
>>> os.path.join('/var', '/lib/rpm.db')
'/lib/rpm.db'
```

That would be horrible because the code would then no longer point to the image root but your host system and people will kill me if I remove their host rpm database :)

aplanas: Great, thanks!

PR #987, opened 2019-03-06
Respect lvm root size

### LINKED ISSUE #983: size setting for LVM @root volume is ignored
## Problem description
OEM image,
```
      <systemdisk name='systemVG'>
        <volume name='@root' size='10240M'/>
        <volume name='var' mountpoint='/var' size='6144M'/>
      </systemdisk>
      <oemconfig>
        <oem-swap>true</oem-swap>
<!--
        <oem-systemsize>20480</oem-systemsize>
-->
      <oemconfig>
```
deployed onto 980GB disk
## Expected behaviour
After deployment, / is 10G, /var is 6G, rest of systemVG is empty
## Observed behaviour
/var is 6G,  / is ~970G, systemVG is completely used

one workaround is to set the oem-systeemsize, but then systemVG will be only 20G and is harder to expand later.
+ KIWI version:  current master or 9.17.23
+ Operating system: CentOS-7 (current) or SLES-15
+ OBS version: 2.9.5

seife:
Fixes #983 

* if "@root size=..." is configured, respect that configuration
* if no size is given, fall back to old 'size=all' implicit config
* add a check if more than one 'size=all' volume exists and error out in that case

### DIFF
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -1130,6 +1130,10 @@ class XMLState(object):
                     realpath = '/'
                     name = 'LVRoot'
                     have_root_volume_setup = True
+                    # compatibility: if no size= is given use the full VG
+                    if not size and not freespace:
+                        log.info('--> LVM: @root volume without size attribute, assuming size="all"')
+                        size = 'all'
                 elif not mountpoint:
                     # setup volume without mountpoint. In this case the name
                     # attribute is used as mountpoint path and a name for the
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -31,7 +31,8 @@ from .utils.size import StringToSize
 from .exceptions import (
     KiwiProfileNotFound,
     KiwiTypeNotFound,
-    KiwiDistributionNameError
+    KiwiDistributionNameError,
+    KiwiLvmSetupError
 )
 
 
@@ -1051,7 +1052,7 @@ class XMLState(object):
 
         container_config_section.set_labels(labels)
 
-    def get_volumes(self):
+    def get_volumes(self):  # noqa: C901
         """
         List of configured systemdisk volumes.
 
@@ -1158,6 +1159,10 @@ class XMLState(object):
                 if ':all' in size:
                     size = None
                     fullsize = True
+                    if have_full_size_volume:
+                        raise KiwiLvmSetupError(
+                            'More than one LVM volume with size="all" attribute'
+                        )
                     have_full_size_volume = True
 
                 volume_type_list.append(
--- a/dracut/modules.d/99kiwi-lib/kiwi-lvm-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-lvm-lib.sh
@@ -160,7 +160,6 @@ function read_volume_setup_all_free {
             return
         fi
     done < "${profile}"
-    echo LVRoot,
 }
 
 function get_all_free_volume {
--- a/kiwi/exceptions.py
+++ b/kiwi/exceptions.py
@@ -428,6 +428,12 @@ class KiwiLoopSetupError(KiwiError):
     """
 
 
+class KiwiLvmSetupError(KiwiError):
+    """
+    Exception raised if the LVM setup is not valid.
+    """
+
+
 class KiwiMappedDeviceError(KiwiError):
     """
     Exception raised if the device to become mapped does not exist.
--- a/test/data/example_lvm_implicit_root_all_config.xml
+++ b/test/data/example_lvm_implicit_root_all_config.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<image schemaversion="6.9" name="LimeJeOS-openSUSE-13.2">
+    <description type="system">
+        <author>Marcus Schäfer</author>
+        <contact>ms@suse.com</contact>
+        <specification>
+            openSUSE 13.2 JeOS, is a small text based image
+        </specification>
+    </description>
+    <preferences>
+        <version>1.13.2</version>
+        <packagemanager>zypper</packagemanager>
+        <locale>en_US</locale>
+        <keytable>us.map.gz</keytable>
+        <timezone>Europe/Berlin</timezone>
+        <rpm-excludedocs>true</rpm-excludedocs>
+        <rpm-check-signatures>false</rpm-check-signatures>
+        <bootsplash-theme>openSUSE</bootsplash-theme>
+        <bootloader-theme>openSUSE</bootloader-theme>
+        <type image="oem" filesystem="ext3" boot="oemboot/suse-13.2" installiso="true" bootloader="grub2" kernelcmdline="splash" firmware="efi">
+            <systemdisk>
+                <volume name="@root"/>
+            </systemdisk>
+            <oemconfig>
+                <oem-systemsize>2048</oem-systemsize>
+                <oem-swap>true</oem-swap>
+            </oemconfig>
+            <machine memory="512" guestOS="suse" HWversion="4">
+                <vmdisk id="0" controller="ide"/>
+                <vmnic driver="e1000" interface="0" mode="bridged"/>
+            </machine>
+        </type>
+    </preferences>
+    <users>
+        <user groups="root" password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root"/>
+    </users>
+    <repository type="yast2">
+        <source path="obs://13.2/repo/oss"/>
+    </repository>
+    <packages type="image">
+        <package name="patterns-openSUSE-base"/>
+    </packages>
+    <packages type="bootstrap">
+        <package name="udev"/>
+        <package name="filesystem"/>
+        <package name="glibc-locale"/>
+    </packages>
+</image>
--- a/test/data/example_lvm_implicit_root_size_all_config.xml
+++ b/test/data/example_lvm_implicit_root_size_all_config.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<image schemaversion="6.9" name="LimeJeOS-openSUSE-13.2">
+    <description type="system">
+        <author>Marcus Schäfer</author>
+        <contact>ms@suse.com</contact>
+        <specification>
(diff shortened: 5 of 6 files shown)


seife:
I'll cater for the failing tests...

kiwi/xml_state.py
```diff
@@ -1154,6 +1159,10 @@ def get_volumes(self):
                 if ':all' in size:
                     size = None
                     fullsize = True
+                    if have_full_size_volume:
+                        raise KiwiLvmSetupError(
+                            'More than one LVM volume with size="all" attribute'
+                        )
```
schaefi: So this should be covered in a runtime check because we want to fail early. As I was working on the tests anyway I covered this one here:

* https://github.com/SUSE/kiwi/pull/992

You could cherry-pick or delete the if condition here

kiwi/xml_state.py
```diff
@@ -1051,7 +1052,7 @@ def add_container_config_label(self, label_name, value):
 
         container_config_section.set_labels(labels)
 
-    def get_volumes(self):
+    def get_volumes(self):  # noqa: C901
```
schaefi: I think we can get rid of noqa, more below

kiwi/xml_state.py
```diff
@@ -1130,6 +1131,10 @@ def get_volumes(self):
                     realpath = '/'
                     name = 'LVRoot'
                     have_root_volume_setup = True
+                    # compatibility: if no size= is given use the full VG
+                    if not size and not freespace:
+                        log.info('--> LVM: @root volume without size attribute, assuming size="all"')
+                        size = 'all'
```
schaefi: I wonder if this is needed. If you don't specify a ```@root``` volume at all the behavior should be exactly like that.
At the end of the get_volumes() method there is a handling for ```if not have_root_volume_setup```. If no other volume has taken the all size argument the root volume will have it.

If you want the root volume to be not resized to the full size you would need to add ```<volume name="@root"/>``` without any size setup.

For my understanding the above code is not needed

seife: I tried and without this, i got 'freespace=30' for @root without size argument.
As this is a change in behaviour, I added this code to reinstate the old behaviour.

seife: BTW: I'm totally fine with that change in behaviour,  but you requested in the ML that I keep the old behaviour :-)

kiwi/xml_state.py
seife: But if the implicit 'size=all' later comes into play, this will not be found by the runtime checker, or does it cover that case also? I guess it only checks the XML file.

schaefi: The implicit size=all should not be set if there is another volume with that configuration.
Actually it would be good if we don't need that, more below

kiwi/xml_state.py
schaefi: You have a point here. I'll give it a test, give me some time to check this. Sorry for the confusion

schaefi: before I test let me summarize what I would expect from a behavior point of view:

* <volume name="@root" size="2G"/>
  resize root volume to 2G

* <volume name="@root" freespace="2G"/>
   resize root volume to have 2G of free space

* No @root volume
  resize root volume to full size of no other volume has size="all" set

* <volume name="@root"/>
  resize root volume to have the min space defined by kiwi (30M) free

If I remember correctly that was the behavior I had in mind with the implementation of ```@root```
Does this match your expectations ? I'll give it a test now

seife: I did not try without `@root` (I thought this would be invalid, maybe it was in old perl-kiwi).
But before my patches, the last two cases (no `@root` and `@root` with or without size) were always handled the same, full size if no other volume had size=all.

schaefi: Hmm, ok I have done the tests with the current master branch (meaning without your patch). The result is as follows:

```
Image type setup
------------------

<type image="vmx" filesystem="ext4" firmware="efi">
    <systemdisk>
        <volume name="@root"/>
    </systemdisk>
    <size unit="G">10</size>
</type>

1. Boot with type setup

   vgdisplay:
     VG Size          9.69 GiB
     Free PE / Size   2155 / 8.42 GiB

   lvdisplay:
     LV Size          1.26 GiB

2. Boot with <volume name="@root" size="4G"/>

   vgdisplay:
     VG Size          9.68 GiB
     Free PE / Size   1453 / 5.68 GiB

   lvdisplay:
     LV Size          4.00 GiB

3. Boot with no @root volume

   vgdisplay:
     VG Size          9.68 GiB
     Free PE / Size   0 / 0

   lvdisplay:
     LV Size          9.68 GiB
```

This is what I would have expected from the code. How can I reproduce your case that the ```@root``` is not taken into account ?

seife:
```
    <type initrd_system="dracut" filesystem="xfs" bootloader="grub2" installiso="false" installpxe="true" installboot="install" installstick="false" kernelcmdline="net.ifnames=0 video=800x600" fsmountoptions="" firmware="efi" image="oem">
      <systemdisk name="systemVG">
        <volume name="@root" size="10240M"/>
        <volume name="var" mountpoint="/var" size="6144M"/>
      </systemdisk>
      <oemconfig>
        <oem-swap>true</oem-swap>
<!--
        <oem-systemsize>20480</oem-systemsize>
-->
      </oemconfig>
    </type>
```
(XFS because this is a CentOS image, but it fails the same with a LomeJeOS-Leap-15.0 based oem setup)

This is one example that failed (LVRoot always max size). The commented oem-systemsize was the workaround for us.
If you build an image like that, just booting the .raw after `truncate -s 100G foo.raw` will suffice for testing.

seife: tested right now with a LimeJeos-Leap15 based image with
```
        <type initrd_system="dracut" filesystem="ext4" bootloader="grub2" installiso="false" installpxe="true" installboot="install" installstick="false" kernelcmdline="net.ifnames=0 video=800x600" fsmountoptions="" firmware="efi" image="oem">
            <oemconfig>
                <oem-boot-title>Leap15 KIWI</oem-boot-title>
                <oem-swap>true</oem-swap>
            </oemconfig>
            <systemdisk name='systemVG'>
                <volume name='@root' size='10240M'/>
                <volume name='var' mountpoint='/var' size='6G'/>
            </systemdisk>
        </type>
```

Built, copied, resized, booted:
```
seife@vbox-seife:/tmp> cp /local/seife/kiwi/LimeJeOS-Leap-15.0.x86_64-1.15.0.raw foo.img --sparse=always
seife@vbox-seife:/tmp> truncate -s 100G foo.img 
seife@vbox-seife:/tmp> qemu-kvm -serial stdio -hda foo.img -m 768 -redir tcp:2222::22
[    0.000000] Linux version 4.12.14-lp150.12.48-default (geeko@buildhost) (gcc version 7.3.1 20180323 [gcc-7-branch revision 258812] (SUSE Linux) ) #1 SMP Tue Feb 12 14:01:48 UTC 2019 (268f014)
[    0.000000] Command line: BOOT_IMAGE=(hd0,gpt3)//vmlinuz-4.12.14-lp150.12.48-default net.ifnames=0 video=800x600 root=UUID=8f302958-e194-46f8-a5a1-0b7a2f8c87e6 rw console=ttyS0,115200 console=tty0
...
Welcome to openSUSE Leap 15.0 - Kernel 4.12.14-lp150.12.48-default (ttyS0).


localhost login: root
Password: 
Have a lot of fun...
localhost:~ # df -hl
Filesystem                               Size  Used Avail Use% Mounted on
devtmpfs                                 347M     0  347M   0% /dev
tmpfs                                    364M     0  364M   0% /dev/shm
tmpfs                                    364M  9.6M  355M   3% /run
tmpfs                                    364M     0  364M   0% /sys/fs/cgroup
/dev/mapper/systemVG-LVRoot               92G  850M   87G   1% /
/dev/mapper/QEMU_HARDDISK_QM00001-part3  283M   64M  201M  24% /boot
/dev/mapper/QEMU_HARDDISK_QM00001-part2   20M  1.1M   19M   6% /boot/efi
/dev/mapper/systemVG-var                 5.9G   29M  5.5G   1% /var
tmpfs                                     73M     0   73M   0% /run/user/0
localhost:~ # pvs
  PV                                      VG       Fmt  Attr PSize  PFree
  /dev/mapper/QEMU_HARDDISK_QM00001-part4 systemVG lvm2 a--  99.68g    0 
localhost:~ # lvs
  LV     VG       Attr       LSize   Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert
  LVRoot systemVG -wi-ao----  93.18g                                                    
  LVSwap systemVG -wi-ao---- 512.00m                                                    
  var    systemVG -wi-ao----   6.00g                                                    
localhost:~ # vgs
  VG       #PV #LV #SN Attr   VSize  VFree
  systemVG   1   3   0 wz--n- 99.68g    0 
localhost:~ # 
```

schaefi: I'm so stupid, my test is not comparable to the resize on deployment. Thanks @seife I'll now drive the right path

seife: well, maybe just documenting the change with no size= argument would also be fine? (I'd rather have explicit than implicit behaviour to be honest)

But of course it would change existing behaviour, I understand that this can be problematic.

seife: Just thinking... Let's make this hunk a loud `WARNING: no size attribute, this is now implicitly treated like frespace=..., add 'size=all' if you want old behaviour`, but not set size=all, and then we can drop the next one with the "duplicate have_full_size" check. And live happily ever after :-)

schaefi: ok finally tested and could reproduce the behavior you saw. I'm sorry but I think we are changing the wrong code. The real problem is in dracut/modules.d/99kiwi-lib/kiwi-lvm-lib.sh and there the method read_volume_setup_all_free(). The change you did there was almost complete but was missing the correct handling based on if there is a root volume setup available or not. So the only change we need is this:

* https://github.com/SUSE/kiwi/pull/994

I have tested the setup for all cases listed above and it worked for me. Do you mind to give your image a test with that patch ? Thanks

seife: looking at the #994 diff, it certainly looks correct, but I'll test it now anyway :-)

seife:
I think #994 is a sufficient (and easier) fix for this problem.

schaefi:
closing in favor of #994

PR #989, opened 2019-03-06
Follow up fix for zypper compat link

schaefi:
Move the code handling the compat setup of the rpm database
to the correct method of the repository API. Call the
import of the signing keys only if there are signing
keys


schaefi:
@aplanas that also covers the condition handling of the signing keys part you wrote about

PR #990, opened 2019-03-06
Complete test coverage on iso tools testing

schaefi:
If there is no mkisofs/genisoimage present on the system, the
tests based on real example iso files are skipped. The code
tested in this case is now covered in a mock style to run
through the logic even without the tools in place

### DIFF
--- a/test/unit/iso_tools_iso_test.py
+++ b/test/unit/iso_tools_iso_test.py
@@ -84,23 +84,47 @@ class TestIso(object):
     @pytest.mark.skipif(
         Path.which('isoinfo') is None, reason='requires cdrtools'
     )
-    def test_create_header_end_block(self):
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Does it run proper tests if xorriso is installed?

schaefi:
sure, on xorriso all of this header magic is not required. So the tests for the xorriso code are complete too but different. Out of scope for this one here I would say

dcermak:
@schaefi Let me cherry-pick your commit and test it in #980 so that we see whether gitlab works with that.

schaefi:
yep thanks it should be all green after the cherry-pick

dcermak:
> yep thanks it should be all green after the cherry-pick

Gitlab is happy: https://gitlab.com/D4N/kiwi-ci/pipelines/50713849, so feel free to merge this.

PR #992, opened 2019-03-07
Check for multiple allsize volumes setup

schaefi:
Added runtime check check_volume_setup_defines_multiple_fullsize_volumes
which reads the volume setup and counts the number of volumes that
define a 'all' volume. Only one is allowed


seife:
I'm wondering how the tests are passing, because `test/data/example_dot_profile_config.xml` contains 2x `size="all"`?

schaefi:
unit test != integration tests. The runtime checker uses the ../data/example_runtime_checker_config.xml
for testing.

I agree the other test example should be cleared up too. Will do that here, just a sec

schaefi:
I'd like to get a new Staging package out and think this one should be part of it. Hope you don't mind I'll go ahead and merge this now. It probably creates a conflict on your open PR #987. If you want me to help or fix the conflict let me know. Sorry but I'm in a hurry with testing on upstream as there are some changes that needs to be ported to SLE

seife:
I'll rebase PR#987 later, once we determined how to exactly go on with it :-)

seife:
May be a stupid question: what does the runtime test do?
I thought that would run across the to-be-built config.xml before starting the build? In that case, all the unit tests still should have failed. If not, the unit tests seem quite useless if they don't invoke the same code that actually building the image does...

schaefi:
> May be a stupid question: what does the runtime test do?
> I thought that would run across the to-be-built config.xml before starting the build? 

correct

> In that case, all the unit tests still should have failed.

No because you are mixing unit testing with integration testing. An integration test has to fail, a unit test is testing a code unit outside of the logic of the entire design or runtime. That's why unit tests allow for mocking and integration tests doesn't. In this case the unit tests which checks the individual runtime check methods are all using ../data/example_runtime_checker_config.xml as source for the test.

The invalid config data you pointed out is in the file example_dot_profile_config.xml which is used to check the .profile contents. That check did not validate on double all size setup because we simply didn't had that before. In the scope of testing that the .profile contents are correct for the shell the contents doesn't play a role. As I said unit testing != integration testing.

Now with the unit test in place any change on the respective code path will make the test to fail if the behavior of the code changes.

That's what unit testing is good for:

* full freedom to do changes

But unit testing never testing the logic or design of the complete implementation. This is reached with integration testing only. That's why both is needed in my eyes

I know some people try to write super clever unit testing which usually leads to an insane effort keeping the tests running which should all go into integration testing. I would agree to that concept if I don't see any line of Mock() in that tests, otherwise it's against the practice of agile design and testing

At least that's how I've learned it and which makes good sense to me

PR #994, opened 2019-03-08
Fixed evaluation of @root size setup

### LINKED ISSUE #983: size setting for LVM @root volume is ignored

schaefi:
The @root volume setup allows to specify a size for the root
volume in an LVM setup. The size setting was not respected and
the root volume was always resized to the rest space available.
This patch fixes the read_volume_setup_all_free method
on the dracut kiwi library to know if there is root volume
setup and respect its contents. This Fixes #983

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-lvm-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-lvm-lib.sh
@@ -134,33 +134,43 @@ function read_volume_setup {
 
 function read_volume_setup_all_free {
     # """
-    # read the volume setup for kiwi_allFreeVolume from the profile
-    # file and return a list with the following values:
+    # read the volume setup from the profile file and
+    # lookup the volume that should take the rest space
+    # available in the volume group.
+    #
+    # return a list with the following values:
     #
     # volume_name,mount_point
     #
-    # If no kiwi_allFreeVolume was configured only the default
-    # volume_name which takes the rest space is returned
+    # If no @root volume is configured in the kiwi XML
+    # description the default LVRoot volume will be used
+    # to take the rest space.
     # """
     local profile=$1
     local size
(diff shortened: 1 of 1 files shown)


seife:
This seems to work for me (still need to test one more variation, but I'm quite sure it will work).
But this is still a change in behaviour: if no size hint foir @root is given (neither size nor freespace), then it will now be handled as 'freespace:30', when  before it was 'size:all'.

I do not object to this change in behaviour, maybe it should be mentioned in release notes or similar as people might be surprised.

schaefi:
You are right with the behavior change but the current behavior is a bug. In case of an ```@root``` setup without a size specification the root volume is not expected to become resized. I broke this with a refactoring task in #47753768bbda. Shame on me. With this patch we basically restore the original behavior.

If you are ok with that I would merge this as a bugfix :)

Will keep this open until I get your feedback

Thanks

seife:
I'm fine with that, my configs all have a size set anyway, it just was not used until now :-)

PR #995, opened 2019-03-08
Use correct GRUB 2 and shim EFI packages for test CentOS 7 appliance

Conan-Kudo:
CentOS 7 has inherited the changes from Fedora in how its EFI binaries
are packaged, so we need to adapt to handle those and build the appliance
properly.


schaefi:
Thanks :+1:

PR #996, opened 2019-03-10
Adding a derived docker image build test

davidcassany:
This is a test for derived docker image. It uses the openSUSE Tumbleweed base image and appends a new layer with kiwi-ng and its runtime dependencies for vmx, iso, oem and docker image types.

### DIFF
--- a/build-tests/x86/test-image-docker-derived/appliance.kiwi
+++ b/build-tests/x86/test-image-docker-derived/appliance.kiwi
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="utf-8"?>
+<image schemaversion="6.9" name="docker-builder-image">
+  <description type="system">
+    <author>David Cassany</author>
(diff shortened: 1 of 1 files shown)


davidcassany:
Nice, I see a new CI Gitlab check :+1:  (I'll need to check **lots** of stuff to get back on track :sweat_smile: )

@schaefi I got some spare time and I worked on the buildah support and a second round of OCI tools refactor. We were missing this derived test and I think this is relevant if we perform some refactoring on this area. If tests are there verifying those builds by running them locally is simple, quick and easy.

schaefi:
Hi David, great to hear from you :smiley: 

yes there is quite some movement on the project which is really really great :+1: 

Thanks for the build test. I completely agree this is very useful :+1: 

Looking forward when you are back :smile:

davidcassany:
Thanks for merging, gonna add the test in OBS project :wink:

PR #997, opened 2019-03-10
Introduce better changelog handling

### LINKED ISSUE #979: Better changelog handling
## Problem description

The changelog for the package build is created via a script ```helper/changelog_generator```
The script takes the entire git history into account and writes the changelog file. This approach
however has some disadvantages:

* A change in the format of e.g the author of a contributor would cause a change of
  entries from the past too.

* The fix for a bug in the script would cause entries from the past to be fixed too

The problem with changing entries from the past is that at least for suse the validation
and checker tools consider this evil and will decline the package submission

## Solution Idea

The idea is now to keep a base changelog file inside the kiwi git and only update that
file with new changelog entries that appears from the git history. This concept would allow
us to keep an base file even with "errors" but don't violate checkers from the distribution.

Thoughts ?

schaefi:
The current way of creating the changelog file for the package
is based on reading the entire history of the git repository and
turns that information into a changelog. The downside of this
approach is that any change in the code that creates this changelog
information will impact older entries and could cause a conflict
on the changes file of the released package. This usually leads
to declined package submissions and blocks us in fixing bugs
in the changelog generator. This commit changes the process in
a way that only changes related to the oldest entry of a reference
changelog file compared to the current branch are taken into
account. This Fixes #979


schaefi:
@Conan-Kudo so with this first approach the changelog entry related to this commit would be

```
-------------------------------------------------------------------
Sun Mar 10 19:24:40 2019 +0100 - Marcus Schäfer <ms@suse.de>

- Introduce better changelog handling

  The current way of creating the changelog file for the package
  is based on reading the entire history of the git repository and
  turns that information into a changelog. The downside of this
  approach is that any change in the code that creates this changelog
  information will impact older entries and could cause a conflict
  on the changes file of the released package. This usually leads
  to declined package submissions and blocks us in fixing bugs
  in the changelog generator. This commit changes the process in
  a way that only changes related to the oldest entry of a reference
  changelog file compared to the current branch are taken into
  account. This Fixes #979
```

The question is; would this author and date setting be accepted by the suse packaging people, I have no idea so far

schaefi:
We would address the full author name issue as well as the time format. The other changes entries uses local timezone conversion which imho was a bad idea. So this is better now

Conan-Kudo:
@schaefi They've been taking it so far when I've converted all my changes files, even with historical entries. Honestly, I strongly suggest you regenerate it once with this new style and commit that. It's better for all of us if you do that, so that it remains consistent. It also makes it easier for us to find if there are any author errors that we should fix, since it's not getting regenerated every single time.

Conan-Kudo:
The date format you're using here is not valid for changes files: `Sun Mar 10 19:24:40 2019 +0100`.

You need it in the standard `date(1)` format, which is also the format accepted by rpm for full timestamp changelogs.

helper/changelog_generator.py
```diff
@@ -0,0 +1,67 @@
...
+            )
+            for line in commit_message:
+                log_data.append('  {0}{1}'.format(
+                    line.lstrip(), os.linesep)
+                )
+            commit_message = []
+    elif line.startswith('Author:'):
+        log_author = line.replace('Author:', '').strip()
+    elif line.startswith('Date:'):
+        log_date = line.replace('Date:', '').strip()
```
Conan-Kudo: This date needs to be converted to the following format: `%a %b %d %T %Z %Y`

schaefi:
>@schaefi They've been taking it so far when I've converted all my changes files, even with historical >entries being altered. Honestly, I strongly suggest you regenerate it once with this new style and >commit that. It's better for all of us if you do that, so that it remains consistent. It also makes it easier >for us to find if there are any author errors that we should fix, since it's not getting regenerated every >single time.

Hmm, I'm always open for changes and improvements. Regarding changes to the rpm changelog I have pretty bad experience. My packages got always declined for obscure reasons so I'm concerned changing old entries. This will be only for Factory but who knows if they will accept the submission

Conan-Kudo:
@schaefi They'll accept it. They've done it for me thus far. Just add a changelog entry about converting the format so that they know that it happened.

helper/changelog_generator.py
```diff
@@ -0,0 +1,76 @@
...
+
+# Author and Date
+log_author = None
+log_date = None
+
+# changelog header line
+log_start = '-' * 67 + os.linesep
+
+# timezone
+timezone = 'Europe/Berlin'
```
Conan-Kudo: Is this necessary? The timezone information has always been encoded in the commit dates, so this shouldn't be required. If you're going to force timezone conversions (which, you shouldn't!), use `UTC`.

schaefi: > Is this necessary?

I had hoped it wouldn't be. You are right the tzinfo is set in the date by git but the conversion via strftime using the %Z expansion does not work. It only worked for me on localizing and that required a timezone to be set. I'll investigate further but haven't found a solution so far

schaefi: Found a solution, will come up with new changes

schaefi:
> @schaefi They'll accept it.

Last famous words :) ok I have created the reference changelog from the beginning of time. Please have a look at ```package/python-kiwi.changes``` if this looks good for you too

Also note I use --topo-order for ordering the git history. That produces good results for me but a diff on the existing changelog of the currently released package shows some order differences here and there.
Usually suse is really picky on that, still concerned that they will take it. But I'm open to try it

helper/changelog_generator.py
schaefi: done using git log to do the job

Conan-Kudo:
@schaefi Uh oh, the changelog entries aren't in reverse chronological order. You may want to use the commit date rather than the author date here. The commit date shows up in `git log --format=fuller` for example.

schaefi:
@Conan-Kudo ok so this is more complex than I thought. I've updated the code to sort in reversed chronological order of the author date. I had done the implementation using the commit date but then I realized a commit date could be very old. Think of somebody who has a branch open for month and creates the pull request some time later. The merge commits are not taken into account as we don't want to see them in the changelog. Thus the commit dates could be possibly older than the latest entry of the reference changelog and would not be listed in the git log output.

Could you have a look at the generated reference changelog file if it's in correct chronological order ?

Thanks

dcermak:
@schaefi I have taken a quick look at the script and there are two things that got me puzzled:
1. dateutil is not in the virtualenv, so the script will fail when the virtualenv is active (at least it does for me)
2. This might be stupid: but aren't you tagging every release? Why not just create a changelog between the last tag and the one before that?

Additionally, although this might be a little out of scope, isn't the current `.changes` file a little too verbose? I mean it's the full git log, wouldn't it be enough to just include the first line of each commit (which should be the summary)?

schaefi:
> dateutil is not in the virtualenv, so the script will fail when the virtualenv is active (at least it does for me)

yes will fix that, thanks

Conan-Kudo:
@dcermak changes file is verbose, but that's because the style of including the reference info and such historically required it to be. We'll see if this will change in the future.

schaefi:
> This might be stupid: but aren't you tagging every release? Why not just create a changelog between the last tag and the one before that?

This would be unsafe, what if I forget it, or someone else does the submission. Many things can imho go wrong at this point.

schaefi:
@dcermak as @Conan-Kudo already said we can't change the style of the changes. That's a requirement of the distribution, partly also by rpm and has historical reasons

Conan-Kudo:
@schaefi Going forward, you're going to commit an updated changes file on each tag-release, right?

schaefi:
yes I use the python bumpversion tool which creates a tag and a commit. The changelog wil have that information. bumpversion is usually the only tool that pushes to master

dcermak:
> > This might be stupid: but aren't you tagging every release? Why not just create a changelog between the last tag and the one before that?
> 
> This would be unsafe, what if I forget it, or someone else does the submission. Many things can imho go wrong at this point.

Ah, ok, since that's how tito works: it tags every release and creates a changelog from that.

helper/update_changelog
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
+"""
+usage: update_changelog <reference_file>
+
+urguments:
```
dcermak: Is this a typo?

schaefi: fixed

schaefi:
@dcermak I've fixed the extension for all remaining helper scripts

.gitlab-ci.yml
```diff
@@ -59,20 +62,32 @@ build_doc:
...
 
-RPM:
+rpm:
   stage: package
   image: fedora:latest
-  before_script:
-    - dnf --refresh -y install mock perl-TimeDate which make gzip tar perl git
+  <<: *install_system_deps
   script:
+    - dnf install -y --refresh mock perl-TimeDate gzip tar perl git
```
Conan-Kudo: Do we need `perl-TimeDate` anymore?

Makefile
```diff
@@ -123,9 +123,10 @@ build: clean tox
 	tar -uf dist/python-kiwi.tar kiwi-${version}/doc/build/latex/kiwi.pdf
 	gzip dist/python-kiwi.tar
 	rm -rf kiwi-${version}
-	# provide rpm changelog from git changelog
-	git log | helper/changelog_generator |\
-		helper/changelog_descending > dist/python-kiwi.changes
+	# update rpm changelog using reference file
+	helper/update_changelog.py package/python-kiwi.changes > \
```
Conan-Kudo: Prefix this with `python3` so that it runs with Python 3 instead of requesting Python 2?

.gitlab-ci.yml
schaefi: don't think so, I deleted it, but added dateutil which is not in core

Makefile
schaefi: I've updated the install list to request only py3, doesn't make sense to add more, thanks

.gitlab-ci.yml
```diff
@@ -59,20 +62,34 @@ build_doc:
...
-RPM:
+rpm:
   stage: package
   image: fedora:latest
-  before_script:
-    - dnf --refresh -y install mock perl-TimeDate which make gzip tar perl git
   script:
+    - >
+      dnf install -y --refresh
+      mock gzip tar perl git python3 python3-dateutil python3-docopt
```
dcermak: Do we still need perl?

Makefile
dcermak: Not prefixing it with python3 makes the rpm stage fail on Fedora, as there is no `python` in env, there is only `python3`.

.gitlab-ci.yml
schaefi: I don't think so will delete it

schaefi: done

schaefi:
@dcermak ok I think I have covered the issues. How is it now ? Thanks

schaefi:
@Conan-Kudo @dcermak thanks guys for reviewing this. I will now merge it and keep my fingers crossed that the buildservice people will accept the submission. If not it will be not so easy to shuffle things back

dcermak:
@schaefi I guess `git revert` should be able do the trick in case they object.

PR #999, opened 2019-03-11
centos-appliance: reorder repos to fix build

seife:
EPEL has older package versions of e.g. librepo, which breaks dnf.
Re-prioritize to prefer the good CentOS packages instead of bad EPEL.

Fixes centos example appliance build in OBS.


Conan-Kudo:
🤦‍♂️

schaefi:
@seife you rock :+1: thanks much for looking into that.

Issue #822, opened 2018-09-26
Move travis process to be called in container


schaefi:
## Problem description

Recently the creation of the kiwi documentation as PDF came to our attention.
Adding the target in the build process was not a big deal but now the travis checks
failed because the travis runtime environment did not contain the LaTeX distribution
which was a requirement to build the PDF through sphinx. Adding LaTeX would
meant the following change in travis 

```
before_install:
  - sudo apt-get install -y texlive-full
```

This change would have caused a big performance problem because all of LaTeX
gets pulled in with every travis call and significantly drops testing performance.

If there would be a pre-built environment which provides, in this case a LaTeX distribution
that can be just used and gets cached in some way on travis it would be a better
solution.

A potential candidate to solve this technically is to use containers as suggested by @Conan-Kudo 

> Have you considered just changing the Travis config to use containers to give
> you a useful environment?
> In rpmlint, I use a configuration that lets me build and test rpmlint on Fedora and openSUSE:
> https://github.com/rpm-software-management/rpmlint/blob/master/.travis.yml
> That way you can also validate that it works on Fedora and openSUSE in CI too. :)

This would open a bunch of questions which we might want to discuss:

1. Do distributions provide e.g LaTeX flavored containers in some registry, or would that something we have to create and maintain first ?
2. Do we want to connect the travis checks to a specific distribution delivered by the container ? Currently we use an ubuntu environment because that comes with travis.
3. Do we agree that if the container has to install LaTeX first, it does not make any difference to install texlive-full as suggested above ?

I'll open that for a discussion, everyone welcome :)

Conan-Kudo:
> Do distributions provide e.g LaTeX flavored containers in some registry, or would that something we have to create and maintain first ?

At least in Fedora, we have a mechanism for offering container builds as part of the distribution. It's not a big thing yet, but it's certainly possible to have a TeXLive container offered by Fedora for this purpose. Just no one has done it yet!

(We could also have a kiwi container for building appliance images :P)

> Do we want to connect the travis checks to a specific distribution delivered by the container ? Currently we use an ubuntu environment because that comes with travis.

Well, obviously, I'd prefer to see Fedora being used in the CI. But reasonably, it makes sense to use at least openSUSE and Fedora, since checks failing in those distributions is actually a problem (openSUSE uses kiwi for its image creation tooling, and Fedora is what RHEL/CentOS will be, and that platform is important to support).

[...]

PR #1002, opened 2019-03-12
GitLab CI integration

schaefi:
The nice part of the CI system there is it's container
concept and the ease of use to run pipelines on different
containers. The changes here references the issue reported
in Issue #822

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,14 +1,62 @@
-test:
-  image: "python:3.6-stretch"
+stages:
+  - test
+
+.install_deps: &install_system_deps
+  before_script:
+    - >
+      dnf install -y --refresh
+      python3 python3-devel make gcc which xz libxml2-devel
+      libxslt-devel enchant python3-tox
+
+flake:
+  stage: test
+  image: fedora:latest
+  <<: *install_system_deps
   script:
-    - apt-get update -qq
-    - apt-get install -y git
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -1,14 +1,62 @@
...
+pytest:
+  stage: test
+  image: fedora:latest
+  <<: *install_system_deps
+  script:
+    - >
+      dnf install -y --refresh
+      xorriso genisoimage python36 python34 python2
+      python2-devel python2-pip
+    - export LANG=en_US.UTF-8
```
Conan-Kudo: This should not be necessary, as Fedora uses `C.UTF-8` by default.

schaefi: It is necessary. without it py2.7 fails

Conan-Kudo: Does it work with `C.UTF-8`?

schaefi:
> There's no package test job either?

yes I just created this as a reference for the tox calls. I'm fine with the package job.

.gitlab-ci.yml
schaefi: I haven't tested this, will change it and then we wait for the result

dcermak: Looks like it does.

Conan-Kudo:
Closing this in favor of #980 as @dcermak will incorporate these improvements there.

PR #1005, opened 2019-03-12
Drop duplicate EPEL7 reference in the test CentOS appliance

Conan-Kudo:
We already request the standard, update, and extras repos and
the EPEL7 repo separately. There's no reason to request the
"centos" EPEL7 repo which combines standard+update and EPEL7
repos already.


schaefi:
Let's hope it does not re-introduce the problem fixed by #999

PR #1006, opened 2019-03-13
kiwi-dump-image: fix kiwi_oemdevicefilter for rd.debug

seife:
Testcase:

* oemdevicefilter = /dev/sda
* oem_unattended = true
* two disks, sda and sdb
* rd.debug
* see that the oem image is deployed on sda

The reason is, that the get_disk_list normally returns lines:

```
/dev/sda 20G /dev/sdb 20G
```

with oemdevicefilter=/dev/sda it returns:

```
/dev/sda filtered out by: ... /dev/sdb 20G
```

And this is enough to make the code just accept /dev/sda as device to install onto ;-)

* redirect info message to stderr to not spoil get_disk_list


schaefi:
Huh, good catch :+1: Thanks much

PR #1007, opened 2019-03-13
add rd.kiwi.oem.maxdisk= boot parameter

seife:
This limits the disks considered for oem deployment to a given size.

As hardware inventory (and disk sizes...) might change after an image was built, it is useful to have this available as a runtime config option for pxe deployment (as opposed to having it in config.xml)

### DIFF
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -45,9 +45,16 @@ function get_disk_list {
     local disk_device_by_id
     local disk_meta
     local list_items
+    local max_disk
+    local kiwi_oem_maxdisk
     local blk_opts="-p -n -r -o NAME,SIZE,TYPE"
     if [ -n "${kiwi_devicepersistency}" ];then
         disk_id=${kiwi_devicepersistency}
+    fi
+    max_disk=0
+    kiwi_oem_maxdisk=$(getarg rd.kiwi.oem.maxdisk=)
+    if [ -n "${kiwi_oem_maxdisk}" ]; then
+        max_disk=$(binsize_to_bytesize "${kiwi_oem_maxdisk}") || max_disk=0
     fi
     if getargbool 0 rd.kiwi.ramdisk; then
         # target should be a ramdisk on request. Thus actively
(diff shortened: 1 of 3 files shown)


dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
```diff
@@ -34,6 +34,22 @@ function scan_multipath_devices {
     systemctl start multipathd
 }
 
+function binsize_to_bytesize {
```
schaefi: This is great but imho should be a method of the lib. Would you consider moving it to ```99kiwi-lib/kiwi-lib.sh```

dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
```diff
@@ -45,10 +61,16 @@ function get_disk_list {
     local disk_device_by_id
     local disk_meta
     local list_items
+    local max_disk md_string
```
schaefi: This is nitpick, can we specify one variable per line

```
local max_disk
local md_string
```

That having said I wouldn't know that md_string will contain the contents of the rd.kiwi.oem.maxdisk option. Would you consider a variable name that makes this more clear. e.g

```
kiwi_oem_maxdisk=(getarg rd.kiwi.oem.maxdisk=)
```

dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
```diff
@@ -77,6 +99,14 @@ function get_disk_list {
             continue
         fi
         disk_size=$(echo "${disk_meta}" | cut -f2 -d:)
+        if [ ${max_disk} -gt 0 ]; then
+            local disk_size_b
```
schaefi: I usually name variables that take byte sizes such that the unit is clear, disk_size_b was clear to me at assignment time but before I thought if there is disk_size_b where is disk_size_a :) Could we make this clear too. e.g

```
disk_size_bytes
```

This naming would then be consistent with other parts of code I wrote e.g in kiwi-partitions-lib.sh
Thanks

dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
seife: OK (to all comments ;) )

schaefi:
Nice :smiley:

PR #1009, opened 2019-03-14
Fixed reading of container config

schaefi:
The additional_tags attribute is optional. If not set
the container config hash does not contain this key.
Accessing the key without the get() method leads to
an unhandled python exception

### DIFF
--- a/kiwi/container/oci.py
+++ b/kiwi/container/oci.py
@@ -142,7 +142,7 @@ class ContainerImageOCI(object):
                 self.oci_config['container_tag']
             )
             additional_refs = []
-            if self.oci_config['additional_tags']:
+            if self.oci_config.get('additional_tags'):
                 additional_refs = []
                 for tag in self.oci_config['additional_tags']:
                     additional_refs.append('{0}:{1}'.format(


schaefi:
See https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_x86 for failed builds

schaefi:
@davidcassany if you have a second it would be great if you can look at this. If not it's not a problem but I will then merge it without your attention because the plan is to get a new kiwi released by tomorrow. I think this bug affects all docker builds that does not specify additional_tags. I guess that will affect most of the builds so this one must be fixed before I release. I apologize to ask you while you are on parental leave but I guess this was caused by the refactoring and you know best if there might be more places that could break.

The one here was an easy catch

Thanks

davidcassany:
OUCH! @schaefi I am really sorry about this one, just seen it right now. I thought I had solved this one in https://github.com/SUSE/kiwi/commit/d252c8cc8c6acf7b45927967486df4d7660dc71f but looks like I missed to commit it. I saw this issue while testing but looks like I forgot to include it into the commit.

You were that fast solving it that I did not see the PR :sweat_smile: and the github notification email has been buried into tons of emails until today...

PR #1011, opened 2019-03-14
Limit scope of virtualbox_guest_additions_present

### LINKED ISSUE #1003: Attribute virtualbox_guest_additions_present can be set when provider is libvirt
## Problem description

The following xml is accepted by KIWI:
``` XML
        <type image="vmx" filesystem="ext4" format="vagrant" boottimeout="0" bootloader="grub2">
            <vagrantconfig provider="libvirt" virtualbox_guest_additions_present="true" virtualsize="42"/>
            <size unit="G">42</size>
        </type>
```
but the presence of the attribute `virtualbox_guest_additions_present` indicates that this config is faulty. Either the provider is wrong or the author forgot to remove the attribute. In either case, this should result in an error to prevent nasty surprises.

## Expected behaviour

I'd suggest that KIWI reports an error when given the above XML.

+ KIWI version: git master

schaefi:
This attribute is only used with the virtualbox provider.
Added a schematron rule to limit the scope to the
virtualbox provider only. This Fixes #1003

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -70,6 +70,8 @@ rpm:
       dnf install -y --refresh
       mock gzip tar git python3 python3-dateutil python3-docopt
       make which
+    # set localtime to timezone of reference changes file
+    - ln -sf /usr/share/zoneinfo/Europe/Berlin /etc/localtime
     - 'sed -i "s|build: clean tox|build:|" Makefile'
     - make build
     - mv dist/python-kiwi.spec .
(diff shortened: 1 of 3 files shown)


dcermak:
The gitlab pipeline failure is unrelated to this change, it looks like the new changelog handling breaks down there as no timezone is defined.

schaefi:
yes currently trying to fix that timezone thing in the containers, so far no luck

schaefi:
finally I could add a workaround for the pipeline problem. A real fix in code will follow in the next minutes

PR #1012, opened 2019-03-15
Fix derived docker images build

davidcassany:
This commit fixes the derived docker images when the base image is a
compressed file. After the refactor in #998 the decompression of the
base image and the skopeo call to import the decompressed image happened
in absolutely independent scopes. NamedTemporaryFile python class by
default deletes the created temporary file when the class instance is
destroyed, this class is used to decompress the base image. Because of
isolated scopes at the time skopeo was called the NamedTemporaryFile
instance was not there anymore. This commit makes sure that the
temporary file instance is still in memory at the time skopeo is
called.

### DIFF
--- a/kiwi/system/root_import/oci.py
+++ b/kiwi/system/root_import/oci.py
@@ -39,27 +39,6 @@ class RootImportOCI(RootImportBase):
         Synchronize data from the given base image to the target root
         directory.
         """
-        oci = OCI()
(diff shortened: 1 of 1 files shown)


davidcassany:
@schaefi this fixes the test for derived images. Sorry for missing this in my oci refactor, locally I have configured docker builds to not use compression to build faster, thus this issue never appeared in my tests.

:+1: for the test, proven to be useful.

PR #1013, opened 2019-03-15
Add buildah tool support for OCI and Docker types

### LINKED ISSUE #764: Support creating container images using buildah + skopeo

davidcassany:
This commit adds buildah tool support for OCI and Docker types. It
requires buildah and skopeo to be installed in the host. The use of
umoci (kept as default choice) or buildah is configured using the
runtime configuration file; consider the following structure:

```
oci:
    - archive_tool: buildah | umoci
```

Fixes #764


schaefi:
Thanks David. Could you rebase your branch to the latest master- That will fix the gitlab-ci pipeline

davidcassany:
Finally I managed to submit the work I started on late 2018 to support buildah. The fact that buildah uses the host's containers-storage is quite annoying as it is hard to find some strategy to not collide with any already existing image or container in host. From a KIWI perspective I find umoci way more sane and easy to work with. Scripting around buildah without leaving any leftover after the build turned not to be that simple.

@Conan-Kudo I'd really appreciate if you could give it a try to build containers on Fedora with this implementation.

> Thanks David. Could you rebase your branch to the latest master- That will fix the gitlab-ci pipeline

Marcus thanks for the hint, however it turned to be my mistake, the implementation was already rebased on today's master. I did a wrong import in `runtime_checker.py` that was only failing on python 2.7.

kiwi/exceptions.py
```diff
@@ -778,3 +778,9 @@ class KiwiDecodingError(KiwiError):
     """
     Exception is raised on decoding literals failure
     """
+
+
+class KiwiBuildahError(KiwiError):
+    """
+    Exception raised on inconsisten buildah class calls
```
schaefi: typo ```inconsisten```

kiwi/oci_tools/buildah.py
```diff
@@ -0,0 +1,325 @@
...
+            )
+
+        # We are making use of skopeo instead of only calling 'buildah from'
+        # because we want to control the image name loaded into the containers
+        # storage. This way we are certain to not leave any left over after the
+        # build.
+        Command.run([
+            'skopeo', 'copy', container_image_ref,
+            'containers-storage:{0}'.format(self.imported_image)
+        ])
```
schaefi: style smell ```([```

kiwi/oci_tools/buildah.py
```diff
@@ -0,0 +1,325 @@
...
+            raise KiwiBuildahError(
+                "Container already initated, called: '{0}'".format(
+                    self.working_container
+                )
+            )
+
+        Command.run([
+            'buildah', 'from', '--name', self.working_container,
+            'containers-storage:{0}'.format(self.imported_image)
+        ])
```
schaefi: same here

kiwi/oci_tools/buildah.py
```diff
@@ -0,0 +1,325 @@
...
+
+    @classmethod                                                # noqa:C901
+    def _process_oci_config_to_arguments(self, oci_config):
+        """
+        Process the oci configuration dictionary into a list of arguments
+        for the 'umoci config' command
+
+        :param list oci_config: meta data list
+
+        :return: List of umoci config arguments
```
Conan-Kudo: This should be `buildah`, not `umoci`?

davidcassany:
@schaefi @Conan-Kudo thanks for your comments so far, already addressed those. I also removed some TODOs I had in the code to test some container configurations, already did some smoke tests.

I only left still there the TODOs regarding how to clean inherited entrypoint and cmd, this is still something I have to discover from buildah, as I have not seen any specific option for that I guess I'll have to ask to buildah developers the best way to do so (probably add a an empty string as cmd and empty list for entrypoint?).

TomSweeneyRedHat:
FYI @vrothberg @rhatdan @nalind @mtrmac

vrothberg:
Thanks for the ping, @TomSweeneyRedHat! 

> Scripting around buildah without leaving any leftover after the build turned not to be that simple.

@davidcassany, can you elaborate a bit on those issues?

I've skimmed the code and all data seems to be cleaned up properly. My Python is a bit rusty but the logic LGTM. Sharing a common storage allows for sharing layers etc. which is nice for concurrent builds but I understand that skopeo+umoci is more trivial for clean ups.

Maybe a `--clean-up` flag would help? If set, not only the build containers but also the pulled images will be removed.

schaefi:
@davidcassany interesting the unit testing passes in travis but fails on gitlab, probably a hash sorting issue, haven't looked too detailed on the error though

davidcassany:
> interesting the unit testing passes in travis but fails on gitlab, probably a hash sorting issue, haven't looked too detailed on the error though

Yes this is surprising. I'll try to investigate it.

davidcassany:
@vrothberg hi nice to hear from you :smile: I hope you are doing well :wink: 
 
> > Scripting around buildah without leaving any leftover after the build turned not to be that simple.
> 
> @davidcassany, can you elaborate a bit on those issues?

Well what I wanted to say with it is that from a builder POV and having the usual KIWI workflow in mind (build a root tree and then configure it and synchronize it to some disk, iso or whatever), is just more straight forward to use `umoci` plus `skopeo` tandem. With them I just create a temporary directory wherever I want or need for the plain OCI image and then can just extract the root tree to any other directory on the system, the operation is similar to what we do with other types of images.

With buildah I have to figure out a new unique name or identifier (in fact I am using random names without verifying if they exist or not...) for the loaded image, another one for the created container and then another one for the final image and keep them until the process ends to successfully clean up everything. In terms of a builder like KIWI it just feels more natural to operate over directories that I can easily create and delete to any desired location without interfering to other services than accessing the containers-storage that it is easily shared with other container related services. 

By the way, may I ask you if there is any way using buildah to clean up the  cmd and entrypoint inherited from the base image when configuring and committing a new image? I mean, imagine base image has '/bin/bash' as `CMD` and no `ENTRYPOINT` is set, so it is empty; than a derived image sets an `ENTRYPOINT` and wants an unset `CMD`. How should I do that using buildah? Do you think setting a `CMD` to empty string would be enough? And what about for the opposite situation, should I set the `ENTRYPOINT` to an empty list?

schaefi:
@davidcassany hope you don't mind I have rebased the branch to match master in order to fix the gitlab rpm target. I have done that work over the weekend and this should now finally be stable. Thanks and sorry for any inconvenience

kiwi/oci_tools/buildah.py
```diff
@@ -0,0 +1,328 @@
...
+        previous image or container present in the host.
+
+        :param int chars_num: Lenght of the generated random string
+        :param list allchars: List of possible characters
+
+        :return: generated random string
+
+        :rtype: string
+        """
+        return "".join(random.choice(allchars) for x in range(chars_num))
```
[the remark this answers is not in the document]
schaefi: In the scope of this use case the random data is random enough and does imho not serve the purpose of a security issue. I ignored that pattern for now

schaefi:
For me this looks good :+1:

davidcassany:
> hope you don't mind I have rebased the branch to match master in order to fix the gitlab rpm target. I have done that work over the weekend and this should now finally be stable. Thanks and sorry for any inconvenience

Sure no problem :wink: I think my last change got lost though :P Let me try if the gitlab is finally fixed adding that

schaefi:
> I think my last change got lost though :P 

urks, this was my fault, so sorry. Glad you had a local version of your changes... I stop breaking other people now... I promise ;)

davidcassany:
> For me this looks good

Good, I'll try to get rid of the two TODOs that are still in kiwi/oci_tools/buildah.py and add an ignore for the Bandit B311 next to the the _offending_ `random.choice` call.

vrothberg:
That's super valuable feedback, @davidcassany. Thanks a lot!

> With buildah I have to figure out a new unique name or identifier (in fact I am using random names without verifying if they exist or not...) for the loaded image, another one for the created container and then another one for the final image and keep them until the process ends to successfully clean up everything.

Can't Kiwi use the identifiers returned by the Skopeo/Buildah? They're guaranteed to be unique for the user.

> In terms of a builder like KIWI it just feels more natural to operate over directories that I can easily create and delete to any desired location without interfering to other services than accessing the containers-storage that it is easily shared with other container related services.

As long as no other service under the user forcefully removes an image or interferes in other ways, Kiwi doesn't need to worry. However, if you want to have a "one directory per build" way of doing it, that's possible. You can do that with Buildah's `--root` flag. Personally, I would use a shared storage to benefit from sharing data but I understand that it doesn't suit each use case.

> By the way, may I ask you if there is any way using buildah to clean up the cmd and entrypoint inherited from the base image when configuring and committing a new image?

I think that's a good use case for `buildah config`.

schaefi:
> Good, I'll try to get rid of the two TODOs that are still in kiwi/oci_tools/buildah.py

yes would be good.

schaefi:
sorry for the close, I hit the wrong button

davidcassany:
> That's super valuable feedback, @davidcassany. Thanks a lot!

Happy to help with my two cents :)
 
> Can't Kiwi use the identifiers returned by the Skopeo/Buildah? They're guaranteed to be unique for the user.

I am not sure I understand that, as far as I saw only when creating a new container a new identifier is generated (calling `buildah from ....`), in that case you are right and I could use the provided name. However I still have to figure out an strategy for naming the loaded image (I need a reference, not the shasum, so I can remove the reference after the build without having to check if the image was already loaded before or not) and the committed image (I need to commit to a unique name to not alter any existing reference and then push to the desired container name and tag).

> As long as no other service under the user forcefully removes an image or interferes in other ways, Kiwi doesn't need to worry. However, if you want to have a "one directory per build" way of doing it, that's possible. You can do that with Buildah's `--root` flag. Personally, I would use a shared storage to benefit from sharing data but I understand that it doesn't suit each use case.

That's an interesting idea I did not consider. You are right I could use a specific storage for each build so I would not have to find unique names for images. Anyway, as far as I understand, this just sets the containers-storage location, and I do not want KIWI to start playing with the containers-storage configuration (location, driver, etc.). In fact, to get buildah working in my Tumbleweed I had to change some little details in `/etc/containers/storage.conf` (I think there was some option not compatible with the btrfs driver IIRC) so I prefer KIWI to not play at all with storage configs and assume this is user's responsibility.
 
> I think that's a good use case for `buildah config`.

Yep just verified that a dervied docker image can clear entrypoint and cmd with `buildah config --cmd= --entrypoint=[]` which results to `"Cmd": null` and to `"Entrypoint": null`.

vrothberg:
> However I still have to figure out an strategy for naming the loaded image (I need a reference, not the shasum, so I can remove the reference after the build without having to check if the image was already loaded before or not) and the committed image (I need to commit to a unique name to not alter any existing reference and then push to the desired container name and tag).

To be honest, I wasn't aware that `buildah commit` doesn't support that already. I will open an issue and tag you there. Thanks again for the feedback!

davidcassany:
> To be honest, I wasn't aware that buildah commit doesn't support that already. I will open an issue and tag you there. Thanks again for the feedback!

Great! just let me add that if I could also load images renaming the reference (e.g `buildah from --image-name imported-image:kiwi docker-archive:/tmp/my_image_file.tar` and get the image loaded to imported-image:kiwi reference discarting any name and tag from the file) and push to archives adding multiple tags then I could do all the job without the need of skopeo. Probably this is out of buildah scope, just mentioning it as at first sight I was quite convinced I would not need skopeo at all but it turned to be necessary.

davidcassany:
@schaefi from my POV this PR is ready.

@Conan-Kudo did you have the chance to give it a try? Today's eve I may try to build some containers from a Fedora host.

Conan-Kudo:
@davidcassany I hadn't given it a shot yet due to the discussion with you and @vrothberg. But I'll give it a go now. :)

rhatdan:
We have just added features like this https://github.com/SUSE/kiwi/pull/1013#issuecomment-474791654 to podman load, except it just adds a tag rather then renaming on the fly.  It would probably be best if you added an RFE for buildah for this support.

davidcassany:
I did some successful tests on Fedora yesterday (built base and derived images from a Fedora host without issues), thus I believe we are ready to go here.

@Conan-Kudo is there any concern you may have with the current implementation? If not could you approve the changes?

Conan-Kudo:
@davidcassany I just need to do a test run myself (as I will also be patching kiwi to prefer buildah over umoci, since umoci isn't available outside of openSUSE).

schaefi:
@Conan-Kudo ok we are waiting for your test then. Thanks for giving this a test in your env, very much appreciated

schaefi:
I'm going to merge this now, since I'd like to run the integration tests on it to make sure no other stuff breaks. @Conan-Kudo would be great if you can give it a test and if there are issues we handle them in a follow up request, thanks much

Conan-Kudo:
@schaefi Seems that it works okay for me. Pushing kiwi 9.17.34 into Fedora now. :)

Bodhi updates:
* Fedora 30: https://bodhi.fedoraproject.org/updates/FEDORA-2019-d3291487eb
* Fedora 29: https://bodhi.fedoraproject.org/updates/FEDORA-2019-93a7015650
* Fedora 28: https://bodhi.fedoraproject.org/updates/FEDORA-2019-40fed64071

schaefi:
>    [1]@schaefi Seems that it works okay for me. Pushing kiwi 9.17.34 into
>    Fedora now. :)

Great, thanks much

Good luck for your presentation this week

PR #1014, opened 2019-03-15
Added timezone conversion to allow date comparison

schaefi:
The output of git log always provides timezone aware data/time
information converted to the local timezone such that pythons
datetime objects are complete. The reading of the reference
file data/time did not convert into the local timezone and that
leads to trouble if the localtime setup of the system is not
the same as the dates in the reference changelog. This commit
fixes that and deletes the workaround in the gitlab-ci rpm
stage


dcermak:
Gitlab still reports this warning:
```
/usr/lib/python3.7/site-packages/dateutil/parser/_parser.py:1204: UnknownTimezoneWarning: tzname CET identified but not understood.  Pass `tzinfos` argument in order to correctly return a timezone-aware datetime.  In a future version, this will raise an exception.
```

So I think to be safe from future breakage, we should be passing a `tzinfos` dict, as described here: https://dateutil.readthedocs.io/en/stable/parser.html

schaefi:
yes I will restructure that

schaefi:
@Conan-Kudo what should be wrong with it ?

From the Makefile:

```
helper/update_changelog.py --since package/python-kiwi.changes > \
    dist/python-kiwi.changes
helper/update_changelog.py --file package/python-kiwi.changes >> \
    dist/python-kiwi.changes
```

Produces a completely correct changes information. I hope you did not compare the
reference log that one is serving as the reference and I intentionally changed that
format (see the commit message where I tried to explain it)

As you know the format of the changelog must be '%a %b %d %T %Z %Y', the %Z here
is the problem. The timezone is used as tzname (e.g CET in my area). For correct calculations
we have to store the region (Europe/Berlin in my case) and as this was not present and we
don't want to statically set this information just to create correct date/time calculation/comparison.
So I changed the reference file to be the ```git log fuller``` data and the code works on that file
or on the actual git log. The produced result will be for the local timezone of the caller or UTC if requested by --utc.

This is the most flexible approach because the git log or the reference are the same format and use only numeric time/date/zone infos which basically can be converted to any timezone

Let me know if the produced changelog file from the above helper/update_changelog.py calls are invalid for you

Thanks

Conan-Kudo:
@schaefi Ah, you're right, I missed the purpose of that. Let me look and be sure now ;)

Conan-Kudo:
@schaefi Am I missing something? When I try to run it, I get this error...

```bash
[neal@yu-ohki ~/Projects/kiwi]$
helper/update_changelog.py --since package/python-kiwi.changes > \
    dist/python-kiwi.changes
helper/update_changelog.py --file package/python-kiwi.changes >> \
    dist/python-kiwi.changes

/usr/lib/python3.7/site-packages/dateutil/parser/_parser.py:1204: UnknownTimezon                                                                              eWarning: tzname CET identified but not understood.  Pass `tzinfos` argument in                                                                               order to correctly return a timezone-aware datetime.  In a future version, this                                                                               will raise an exception.
  category=UnknownTimezoneWarning)
  File "helper/update_changelog.py", line 101, in <module>
    if date_reference < author_date:
TypeError: can't compare offset-naive and offset-aware datetimes
usage: update_changelog --since=<reference_file>|--until=<reference_file>
```

schaefi:
@Conan-Kudo something seems wrong with your branch

see the usage output...

```
...
usage: update_changelog --since=<reference_file>|--until=<reference_file>
```

The PR here deletes the --until option. I guess you are not on the latest code from the add_timezone_conversion_in_changelog_helper branch ?

Conan-Kudo:
@schaefi Ah yes, you're right. This seems to work for me now.

PR #1016, opened 2019-03-18
relax the live iso root dir permissions

seife:
The root of the live ISO has permissions 0700, which is inconvenient if
you mount it to explore its contents. Relax to 0755 instead.

### DIFF
--- a/kiwi/iso_tools/xorriso.py
+++ b/kiwi/iso_tools/xorriso.py
@@ -150,6 +150,7 @@ class IsoToolsXorrIso(IsoToolsBase):
             [
                 self.get_tool_name()
             ] + self.iso_parameters + [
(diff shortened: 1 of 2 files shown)


schaefi:
From a coding perspective this makes sense to me, but you did not update the tests. os.chmod is really called and not covered by the unit test. I think you want to check the mocked call for 0755 in ```test/unit/builder_live_test.py``` and there in all tests that calls the create() method. You could alternatively think about changing the time when you "relax" the directory permissions. Maybe it's better to that on successful operations and not directly after the mkdtemp creation. Since for the standard processing it's not needed to immediately change the permissions bits, that will also relax the changes you need on the unit testing side of things.

schaefi:
That looks nice :+1:

PR #1017, opened 2019-03-18
Fixed race condition on live overlay mount

### LINKED ISSUE #1015: Live images fail to boot due to sysroot.mount race
Fails in openQA, reproducible locally as well. It fails approx. every 4th try.

https://openqa.opensuse.org/tests/881471

The error is
```
[    6.445371] overlayfs: failed to resolve '/run/overlayfs/rw': -2
```

AFAICT this is because `sysroot.mount` does not wait for udev to settle.

schaefi:
Make sure to wait for all targets of the overlay mount
This Fixes #1015

### DIFF
--- a/dracut/modules.d/90kiwi-live/parse-kiwi-live.sh
+++ b/dracut/modules.d/90kiwi-live/parse-kiwi-live.sh
@@ -34,5 +34,7 @@ info "root was ${liveroot}, is now ${root}"
 [ -z "${root}" ] && root="live"
 
 wait_for_dev -n /run/rootfsbase
+wait_for_dev -n /run/overlayfs/rw
+wait_for_dev -n /run/overlayfs/work
 
 return 0


schaefi:
@Vogtinator I did not run into the race condition but I could imagine this one fixes it

Vogtinator:
Hm, AFAICT those directories are not devices though, just plain folders. Does `wait_for_dev` apply to them as well?

If so, can you provide a python3-kiwi on OBS I can build a test .iso against?

schaefi:
That seems to run on directories as well. The method creates a /lib/dracut/hooks/initqueue/finished/devexists-run-overlayfs-rw.sh script which basically checks for the presence with '-e' that applies to directories and files as well.

Let's give it a try on staging. Once merged there will be an automated iso build here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86/test-image-iso

and you should be able to fetch a package from Virt:App:Staging

I'll do some tests beforehand. Still not sure if this is a real fix

Vogtinator:
It should definitely not make it worse, so I'd say: Go for it!

schaefi:
Will do. merging now and submit to Virt:App:Staging, from there you get a package and can run your test too

Vogtinator:
I wasn't able to reproduce the issue with the test iso from the staging project. That doesn't mean much though.

PR #1019, opened 2019-03-22
Support optional fstab.patch file

### LINKED ISSUE #945: KIWI does not allow creating images with current transactional-update mechanism
transactional-update 1.12 moved from using `/etc/fstab.sys` to `/etc/fstab` - so that needs special modifications.

Those are done by the `read-only-root-fs` package during package installation, but at that time `/etc/fstab` is not populated yet.

So a workaround would be to add the necessary entry in `config.sh` - that's how it was done until now, but kiwi overwrites the `/etc/fstab` completely in a later stage.

This makes it impossible to create images using transactional-update currently.

schaefi:
In addition to the support for fstab.append, users can now also
provide a patch file to change the contents of the fstab file
as it got written by kiwi. The feature is probably rarely used
but needed in the area of suse's transactional update mechanism.
This Fixes #945

### DIFF
--- a/doc/source/building/working_with_images/custom_fstab_extension.rst
+++ b/doc/source/building/working_with_images/custom_fstab_extension.rst
@@ -1,0 +1,66 @@
+.. _custom_fstab_extension:
+
+Add or Update the Fstab File
+============================
+
+.. sidebar:: Abstract
+
+   This page provides further information for customizing
+   the `/etc/fstab` file which controls the mounting of
+   filesystems at boot time.
+
+In KIWI, all major filesystems that were created at image
+build time are handled by KIWI itself and setup in `/etc/fstab`.
+Thus there is usually no need to add entries or change the
+ones added by KIWI. However depending on where the image is
+deployed later it might be required to pre-populate fstab
+entries that are unknown at the time the image is build.
+
+Possible use cases are for example:
+
+* Adding NFS locations that should be mounted at boot time.
+  Using autofs would be an alternative to avoid additional
+  entries to fstab. The information about the NFS location
+  will make this image specific to the target network. This
+  will be independent of the mount method, either fstab or
+  the automount map has to provide it.
+ 
+* Adding or changing entries in a read-only root system
+  which becomes effective on first boot but can't be added
+  at that time because of the read-only characteristics.
+
+.. note::
+
+   Modifications to the fstab file are a critical change. If
+   done wrongly the risk exists that the system will not boot.
+   In addition this type of modification makes the image
+   specific to its target and creates a dependency to the
+   target hardware, network, etc... Thus this feature should
+   be used with care.
+
+The optimal way to provide custom fstab information is through a
+package. If this can't be done the files can also be provided via
+the overlay file tree of the image description.
+
+KIWI supports two ways to modify the contents of the `/etc/fstab`
+file:
+
+1. Providing an `/etc/fstab.append` file
+
+   If that file exists in the image root tree, KIWI will take its
+   contents and append it to the existing `/etc/fstab` file. The
+   provided `/etc/fstab.append` file will be deleted after successful
+   modification.
+
+2. Providing an `/etc/fstab.patch` file
+
+   The `/etc/fstab.patch` represents a patch file that will be
+   applied to `/etc/fstab` using the `patch` program. This method
+   also allows to change the existing contents of `/etc/fstab`.
+   On success `/etc/fstab.patch` will be deleted.
+
+.. note::
+
+   `/etc/fstab.append` and `/etc/fstab.patch` can be used together.
+   If both files are provided, appending happens first and patching
+   afterwards.
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -612,14 +612,20 @@ class SystemSetup(object):
         """
         Create etc/fstab from given list of entries
 
-        Also lookup for an optional fstab.append file which allows
+        Also look for an optional fstab.append file which allows
         to append custom fstab entries to the final fstab. Once
         embedded the fstab.append file will be deleted
+
+        Also look for an optional fstab.patch file which allows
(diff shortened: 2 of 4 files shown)


kiwi/system/setup.py
```diff
@@ -629,6 +635,12 @@ def create_fstab(self, entries):
                     fstab.write(append.read())
                 Path.wipe(fstab_append_file)
 
+        if os.path.exists(fstab_patch_file):
+            Command.run(
+                ['patch', '-i', fstab_patch_file]
+            )
+            Path.wipe(fstab_patch_file)
```
aplanas: A patch can fail. Do makes sense to detect it and abort?

schaefi: yes and if it fails the Command.run raises with an exception including the failed text from patch and this would be the end result of that run. It would end with a KiwiCommandError(...fail info from path). I considered this enough and did not catch it here for an extra validation

Do you think we need more ?

aplanas: > Do you think we need more ?

No! Is perfect :+1:

schaefi:
> Could you also add a note about this feature to the documentation?

yep, done now

doc/source/building/working_with_images/custom_fstab_extension.rst
```diff
@@ -0,0 +1,63 @@
...
+============================
+
+.. sidebar:: Abstract
+
+   This page provides further information for customizing
+   the `/etc/fstab` file which controls the mounting of
+   filesystems at boot time.
+
+In KIWI, all major filesystems that were created at image
+build time are handled by KIWI itself and setup in `etc/fstab`.
```
dcermak: These should be two `

doc/source/building/working_with_images/custom_fstab_extension.rst
```diff
@@ -0,0 +1,63 @@
...
+  image specific to the target network it later runs in.
+
+* Adding or Changing entries in a read-only root system
+  which becomes effective on first boot but can't be added
+  at that time because of the read-only characteristics.
+
+.. note::
+
+   Modifications to the fstab file are a critical change. If
+   done wrong the risk that the system will not boot exists.
```
dcermak: wrong -> wrongly and the word order is incorrect in this sentence (exist should be at the 2nd position in this sentence)

doc/source/building/working_with_images/custom_fstab_extension.rst
```diff
@@ -0,0 +1,63 @@
...
+
+   If that file exists in the image root tree, KIWI will take its
+   contents and append it to the existing `etc/fstab` file. The
+   provided `etc/fstab.append` file will be deleted after successful
+   modification.
+
+2. Providing an `etc/fstab.patch` file
+
+   The `etc/fstab.patch` represents a patch file that will be applied
+   against the `etc/fstab` file using the `patch` program. This method
```
dcermak: against -> to

doc/source/building/working_with_images/custom_fstab_extension.rst
```diff
@@ -0,0 +1,63 @@
...
+   allows also for changing the existing contents of the current
```
dcermak: This method also allows to change the existing contents of `/etc/fstab`.

doc/source/building/working_with_images/custom_fstab_extension.rst
```diff
@@ -0,0 +1,63 @@
...
+   `etc/fstab` file. On success the `etc/fstab.patch` will be deleted.
```
dcermak: drop the `the`

doc/source/building/working_with_images/custom_fstab_extension.rst
```diff
@@ -0,0 +1,63 @@
...
+
+.. note::
+
+   `etc/fstab.append` and `etc/fstab.patch` can be used together.
+   If both files are provided, appending happens first and patch
```
dcermak: and patch will be last => and patching afterwards.

doc/source/building/working_with_images/custom_fstab_extension.rst
```diff
@@ -0,0 +1,63 @@
...
+   will be last.
+
+The optimal way to provide the fstab extension files is through a
```
dcermak: [suggestion] Move this information up, so that the reader will not start wondering where they should put `/etc/fstab.{patch,append}`

doc/source/building/working_with_images/custom_fstab_extension.rst
schaefi: it rendered fine, I have done a testbuild (tox -e doc)  before submission. We use the single backticks in several other places of the doc source.

Thanks for the review looking into the other issues asap

doc/source/building/working_with_images/custom_fstab_extension.rst
schaefi: will do, good idea

schaefi:
I've updated the docs, but kept the single backticks so far as it did not cause any harm.

doc/source/building/working_with_images/custom_fstab_extension.rst
dcermak: Oops, I always thought that one has to use two backticks.

doc/source/building/working_with_images/custom_fstab_extension.rst
```diff
@@ -0,0 +1,63 @@
...
+Thus there is usually no need to add entries or change the
+ones added by KIWI. However depending on where the image is
+deployed later it might be required to pre-populate fstab
+entries that are unknown at the time the image is build.
+
+Possible use cases are for example:
+
+* Adding NFS locations that should be mounted at boot time.
+  The information about the NFS location would make this
+  image specific to the target network it later runs in.
```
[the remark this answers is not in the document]
schaefi: I've rephrased that use case

PR #1020, opened 2019-03-22
Fix config.sh calling

### LINKED ISSUE #984: KIWI calls config.sh via bash thereby ignoring the shebang
## Problem description

KIWI calls the `config.sh` script during image creation explicitly via bash (see: [setup.py:869](https://github.com/SUSE/kiwi/blob/b20545897b0ef3422ffa197ce14fabb6646ab9f4/kiwi/system/setup.py#L869)):
``` Python
            config_script = Command.call(
                ['chroot', self.root_dir, 'bash', '/image/' + name]
            )
```

This has the downside, that scripts which specify a different shebang, like `#!/bin/bash -e` or `#!/bin/sh` will not work as expected.

## Expected behavior

Either, this should be explicitly stated in the docs with a big red warning or the script should be called directly. If this behavior should not change, then I'd suggest that we remove every shebang from all the examples, since that suggests that it actually does something.

Modify the shebang of `config.sh` to `#!/bin/bash -e` and add a command that returns a non-zero value to the script. It won't abort the build.

+ KIWI version: git master
+ Operating system: OpenSUSE Tumbleweed

dcermak:
Fixes #984.

* modify `SystemSetup._call_script` to call scripts only through bash if the executable bit is not set, otherwise let the OS call them
* adapt unit tests & docs

### DIFF
--- a/kiwi/path.py
+++ b/kiwi/path.py
@@ -21,6 +21,7 @@ import collections
 # project
 from .command import Command
 from .logger import log
+from .exceptions import KiwiFileAccessError
 
 
 class Path(object):
@@ -59,6 +60,40 @@ class Path(object):
             for path in ordered_paths_at_depth[path_depth]:
                 ordered_paths.append(path)
         return ordered_paths
+
+    @classmethod
+    def access(cls, path, mode, **kwargs):
+        """
+        Check whether path can be accessed with the given mode.
+
+        :param str path: The path that should be checked for
+            access.
+
+        :param int mode: Which access mode should be checked.
+            This value must be a bit-wise or of one or more of the following
+            constants: :py:const:`os.F_OK` (note that this one is zero),
+            :py:const:`os.X_OK`, :py:const:`os.R_OK` and :py:const:`os.W_OK`
+
+        :param kwargs: further keyword arguments are forwarded to
+            :func:`os.access`
+
+        :return: Boolean value whether this access mode is allowed
+        :rtype: bool
+
+        :raises ValueError: if the supplied mode is invalid
+        :raises kiwi.exceptions.KiwiFileNotFound: if the path does not exist or
+            is not accessible by the current user
+        """
+        if mode & ~(os.F_OK | os.X_OK | os.R_OK | os.W_OK) != 0:
+            raise ValueError("Invalid mode 0x{:X}".format(mode))
+        try:
+            os.stat(path)
+        except Exception as exc:
+            raise KiwiFileAccessError(
+                'Error accessing path {0} failed with: {1}'.format(path, exc)
+            )
+
+        return os.access(path, mode, **kwargs)
 
     @classmethod
     def create(cls, path):
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -877,10 +877,13 @@ class SystemSetup(object):
             )
 
     def _call_script(self, name):
-        if os.path.exists(self.root_dir + '/image/' + name):
-            config_script = Command.call(
-                ['chroot', self.root_dir, 'bash', '/image/' + name]
-            )
+        script_path = os.path.join(self.root_dir, 'image', name)
+        if os.path.exists(script_path):
+            command = ['chroot', self.root_dir]
+            if not Path.access(script_path, os.X_OK):
+                command += ['bash']
+            command += ['/image/' + name]
+            config_script = Command.call(command)
             process = CommandProcess(
                 command=config_script, log_topic='Calling ' + name + ' script'
             )
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -33,7 +33,8 @@ extensions = [
     'sphinx.ext.ifconfig',
     'sphinx.ext.viewcode',
     'sphinxcontrib.spelling',
-    'sphinx.ext.autodoc'
+    'sphinx.ext.autodoc',
+    'sphinx.ext.intersphinx',
 ]
 
 docopt_ignore = [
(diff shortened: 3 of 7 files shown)


schaefi:
The refactor bits sounds good to me, we should do that here. I'll set the work in progress flag.
The code looks good to me too, so let's just add the check for the x-bit to the Path class and merge afterwards

kiwi/system/setup.py
```diff
@@ -864,11 +865,20 @@ def _import_custom_scripts(self):
                 ]
             )
 
+    @staticmethod
+    def _is_file_executable(file_name):
```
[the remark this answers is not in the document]
dcermak: I didn't know that `os.access` exists, so why not use that one directly? `Path.access` would only call that function anyway.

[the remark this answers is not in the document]
davidcassany: That's a good point, just verified `os.access` does not throw an exception if the file does not exist, just returns `False`. Probably would make sense to make a dedicated method just to handle this sort of little cases.

When I suggested the method I imagined a method that checks existence first and then the access mode, this way we could easily have a method that reacts different than just providing `False` in case the file does not exist, we could throw an exception or not according to some optional parameter. 

Said that I would also be fine by just calling `os.access` method directly and just do not use any other method.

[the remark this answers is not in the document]
dcermak: > That's a good point, just verified os.access does not throw an exception if the file does not exist, just returns False.

While that's technically correct, it's unexpected behavior for a python function. I'll create a sane wrapper for that.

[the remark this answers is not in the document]
dcermak: Done.

dcermak:
@schaefi Could you take a look at 53268896197be66d6518beda89ddae1132492c25 again? I had to add some trickery to make this work for Python 2, as `os.stat` raises `OSError` there but Python 3 raises a `FileNotFoundError`. Not sure if you like the current state.

kiwi/path.py
```diff
@@ -15,12 +15,19 @@
...
 
 # project
 from .command import Command
 from .logger import log
+from .exceptions import KiwiFileNotFound
+
+
+# Work around os.stat() raising different exceptions for Python 2 & Python 3
+StatExceptionFileNotFound = OSError if sys.version_info.major == 2 \
+    else FileNotFoundError
```
schaefi: Can we do this differently, see below

kiwi/path.py
```diff
@@ -60,6 +67,38 @@ def sort_by_hierarchy(cls, path_list):
...
+        :raises ValueError: if the supplied mode is invalid
+        :raises kiwi.exceptions.KiwiFileNotFound: if the path does not exist or
+            is not accessible by the current user
+        """
+        if mode & ~(os.F_OK | os.X_OK | os.R_OK | os.W_OK) != 0:
+            raise ValueError("Invalid mode 0x{:X}".format(mode))
+        try:
+            os.stat(path)
+        except StatExceptionFileNotFound:
+            raise KiwiFileNotFound("Path {!s} not found".format(path))
```
schaefi:
```python
except (OSError, FileNotFoundError) as issue:
    raise KiwiFileNotFound(
        'Error accessing path {0}: {1}'.format(path, issue)
    )
```

dcermak: I opted not to do this, since that will catch `OSError` for Python 3 too, which we might not want.

schaefi: I'm just thinking why we don't want to catch OSError. If os.stat raised with an OSError I think we should catch that in any way. This would also be the reason why I tried to make the error message a bit more generic and use the exception information as part of the error message.

Actually if I think further down is there any reason why we have to specifically watch out for an exception type ? I mean if os.stat throws an exception it should be catched, or is there any raise condition which we want to handle later ? I think if we let pass one here it will end up in a raise at another code path that might be misleading or in worst case ends in a stack trace.

I found one more. You raise a ValueError but that is a python exception and outside of the scope of kiwi. This will surely lead to a stack trace and is treated as an "unknown error", but we want to treat it as a known error. So I suggest the following:

1. add exceptions to kiwi/exceptions.py

```python
class KiwiValueError(KiwiError):
    """
    Exception raised if an invalid value for an operation is used
    """


class KiwiFileAccessError(KiwiError):
    """
    Exception raised if accessing file metadata has failed
    """
```

2. Now in the code

```python
    if mode & ~(os.F_OK | os.X_OK | os.R_OK | os.W_OK) != 0:
￼        raise KiwiValueError('Invalid mode 0x{:X}'.format(mode))
    try:
        os.stat(path)
￼    except Exception as issue:
        raise KiwiFileAccessError(
            'Error accessing path {0}: {1}'.format(path, issue)
        )
```

Thoughts ?

dcermak: 1. I've added the raising of `ValueError` as a safety net for the programmer. This is something that a static type system could catch, but since we don't have that, I have opted for an exception instead. This part should only be reachable if a programmer messes up and passes a wrong bitmask to `Path.access`. Imo that should result in a backtrace, as that will make it obvious that that's a bug and not some error that the user inserted (which most `Kiwi*Error` exceptions indicate).

2. There's not really a reason to not catch `OSError`, I'll include the above code.

schaefi: >  Imo that should result in a backtrace, as that will make it obvious that that's a bug

ah ok

dcermak: I've incorporated the changes from 2. into the PR. 

If you don't like the `ValueError` throwing, I can also include that (it's your project, so it's your decision, I was merely explaining my thought process).

schaefi: That's all fine, thanks

PR #1021, opened 2019-03-22
Split code_style_plus_unit_test into two jobs for parallel runs

dcermak:
Fixes `code_style_plus_unit_test` takes quite long to run on GitLab CI.

* Split the job into two jobs that can run in parallel.


schaefi:
Looks nice. 2.7 and 3.4 will go away in the future anyway so this is a nice separation. Thanks

PR #1023, opened 2019-03-26
Replace @staticmethod with @classmethod when needed

aplanas:
@classmethod are used in Python to represent methods that can
query and update the class (cls parameter). Is expected to be
used for metaprograming, or advanced techniques that require the
access to the class itself, before we have an instance.

@staticmethod are used to associate a function to a class. It will
not be have access to the instance (self) not the class (cls). In
other programming languages are known as class methods.

This patch replace all the @classmethod with @staticmethod when
there is not need to access to the cls parameter, because the
intention is to be used as normal functions.

### DIFF
--- a/helper/schema_parser.py
+++ b/helper/schema_parser.py
@@ -23,8 +23,8 @@ class SchemaNode(object):
 
     Child = namedtuple('Child', ['node', 'properties'])
 
-    @classmethod
-    def is_type(self, node):
+    @staticmethod
+    def is_type(node):
         pass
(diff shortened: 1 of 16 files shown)


schaefi:
@aplanas sorry the merge of the path changes conflicts with your work. I think it should be easy to resolve. Just in case you have trouble let me know. Thanks

aplanas:
Rebased

schaefi:
Thanks

PR #1028, opened 2019-03-27
Problem: cannot customize options for apt-get (and others)

jimklimov:
Fixes: inability to customize packaging tools

Solution: allow to pass options via config.yml; avoid namespace clashes
Example: (note each CLI token must be a separate option)

````
package_manager_apt-get:
  - options:
    - -o
    - Acquire::Check-Valid-Until=false
````

Signed-off: Jim Klimov <jimklimov@gmail.com>
Author: Jim Klimov <EvgenyKlimov@eaton.com>

Practical use-case: our project is currently based on Debian Jessie, whose backports repo is no longer maintained since recently. Mirrors are available, but metadata is deemed expired:
````
[ ERROR   ]: 03:49:58 | KiwiCommandError: chroot: stderr: E: Release file for http://archive-klecker.debian.org/debian/dists/jessie-backports/InRelease is expired (invalid since 35d 7h 24min 15s). Updates for this repository will not be applied.
, stdout: Hit http://archive-klecker.debian.org jessie-backports InRelease
````

While our project migrates to a newer baseline, we must keep builds flowing as they were. This fix to our Kiwi part of the workflow (along with the config.yml snippet above) repairs the flow for us.

### CODE REVIEW

**Comment by jimklimov** [comment]
I am not sure I can address the Travis errors, they seem to be in code that was before me. And more so in the tester (coverage) tool...

**Comment by davidcassany** [comment]
Thanks much for the contribution, in fact this is an interesting feature. In general terms I like it, however I am afraid it could be easy to screw KIWI (or even the host system) if the user is free to enter any custom package manager option. I'd say we need some strategy to sanitize it, which does not look like a simple task.

Also, for your specific use case, we could address it by setting the option you included when signatures checks are disabled (`<rpm-check-signatures>false</rpm-check-signatures>`), probably we could assume that repo and package checks are generally disabled with that flag. I am not very convinced with this idea as `Acquire::Check-Valid-Until=false` is not related to signatures, but I would be fine reconsidering the scope of this option and, why not, a new name to better reflect its behavior.

Also as a side note, I believe we could consider renaming some of the `rpm-*` options that could be generic to all package managers, having `rpm` in there is misleading.

> I am not sure I can address the Travis errors, they seem to be in code that was before me. And more so in the tester (coverage) tool... UPDATE: Verified that our local Kiwi does not explode if config.yml is missing.

`config.yml` is not required, thus KIWI just uses all defaults if missing. Regarding the Travis errors, there are a couple of issues, the code you provided does not match flake standards so the `tox -e check` verification fails. In addition due to the code change there is at least a test failing which should also be fixed. And you are right that some errors for python 3.7 were already there before your contribution, looks like linkcheck for python 3.7 started to suddently fail in Travis for some unknown reason :(

But before spending time on fixing those little issues I'd prefer that we all agree on how this PackageManager custom options should look like.

Any thoughts?

**Comment by jimklimov** [comment]
Well, our original attempt was to make this toggle more constrained in scope, e.g. setting a per-repository setting (we do only need to skip the checks for that one repo, of several used in the build) in `config.xml`. However this impacted schema checking faults (kudos for having those) and we next looked at alternatives and came up with this bigger hammer. Also it is not clear with XML if we can guarantee that some tokens (e.g. `-o` and `Apt::Flag=value`) can be forced to go as separate tokens into the system call (probably needs to be elements like the `source` tag inside the repo tag), and at that they should be passed in deterministic order... maybe a CLI-argument element with one or two attributes, or attribute and optional content?

Also, looking back at this PR's proposal, I wonder if direct `runtime_config._get_attribute()` lookups into the YML contents abuse an intended-private interface? (The leading underscore got me wondering) Then maybe the logic should move into `runtime_config.py` as another public method to get a package manager global config. (This does not exclude proceeding with XML for repo-local configs)

I am not sure if either of these approaches impacts the host system; at least in our case of building Debian images on an OpenSUSE host, this all happens in buildroot. I tested first that creating `/etc/apt*` configs on the host does not leak into the build, that would have been the easiest bandaid for us ;)

**Comment by jimklimov** [comment]
Would documenting the feature as "you are now able and free to shoot yourself in the foot with this" suffice for addressing how users can screw their setups? ;)

**Comment by schaefi** [comment]
@jimklimov thanks for your effort, much appreciated

> I am concerned about letting the user set any custom packager manager option.

I was reading the code and thought about the idea and your concern. Actually I think it would be nice to do this. Yes specifying that options will break the build if done wrong. But what will happen; kiwi will stop at the package manager call and we will get an error message from the package manager that hopefully is descriptive enough to point out the problem. I could live with that :)

I did not see how the host system could be impacted negatively. As far as I see the implementation we will pass additional options 1:1 to the call. In the bootstrap phase that options will be passed to the package manager call done from the host and in the install phase it will be passed to the chrooted call. But we are not changing the host context as far as I can see. So this is a safe operation to my understanding.

From the implementation perspective I suggest that we change the yaml style a bit:

* just use the package manager name as section
* distinguish between repo setup and package processing

Like this:

```yaml
apt-get:
  repository_options:
    - a
    - b
  processing_options:
    - a
    - b
```

The reason why I suggest to differentiate here is because in kiwi we have a repository interface and a package manager interface. Package managers usually handles both areas, the repo setup as well as the processing part (install, update, ...) But different options might be used and I think that should be explicitly stated.

> code Style, Unit test needs to be updated accordingly

yes, if you need help or want us to jump in just let us know

@davidcassany @jimklimov what do you think ?

davidcassany:
> I was reading the code and thought about the idea and your concern. Actually I think it would be nice to do this. Yes specifying that options will break the build if done wrong. But what will happen; kiwi will stop at the package manager call and we will get an error message from the package manager that hopefully is descriptive enough to point out the problem. I could live with that :)

If that's the case I can certainly live with that too :) Also, as I already mentioned in my first comment, I like the idea behind this PR. I am just trying to be evil minded as we let user introduce strings that are part of a command which is run as root.

> I did not see how the host system could be impacted negatively. As far as I see the implementation we will pass additional options 1:1 to the call. In the bootstrap phase that options will be passed to the package manager call done from the host and in the install phase it will be passed to the chrooted call. But we are not changing the host context as far as I can see. So this is a safe operation to my understanding.

I was thinking in something like:

```yaml
zypper:
  repository_options:
    - --root
    - /my/crazy/path
```

Which up to my understanding will end to calls like:

before bootstrap, adding the repo:
```bash
zypper --root /my/crazy/path --root /kiwi_img_rootdir ar <some other kiwi defined options> <uri> <alias>
```

and similar for boostrap:
```bash
zypper --root /my/crazy/path --root /kiwi_img_rootdir install <some other kiwi defined options> <packages>
```

And... surprise surprise... just run similar commands in my tumbleweed VM and they do not fail :sob: They just ignore the second `--root`. We could _fix_ or workaround this specific insane mistake by appending custom options after the kiwi defined ones (so having `--root` calls in opposite order). But who know what other insane configs are possible. :sweat_smile: 

I believe one can badly harm the host by playing options in convenient way. Probably the question is if we should care or not. Probably there are already other _holes_ in kiwi, from a security POV.

> From the implementation perspective I suggest that we change the yaml style a bit:
> 
>     * just use the package manager name as section
> 
>     * distinguish between repo setup and package processing

Yes it makes sense :+1:

> > code Style, Unit test needs to be updated accordingly 
> yes, if you need help or want us to jump in just let us know

Sure we can help here. Just lets agree and specify the feature details first, otherwise we might code twice.
 
> @davidcassany @jimklimov what do you think ?

I am still kind a concerned about how much a user can mess around the build with it. But also I am not sure to what extend we should limit ourselves because of that, I don't have an strong opinion.

davidcassany:
@schaefi @Conan-Kudo @jimklimov Do you think it would make sense to limit this custom options to chroot only commands? Meaning no custom options for the bootsraphase. I doubt this would solve the former issue of this PR, as now I don't think we can easily define different repos depending we are on boostrap phase or not.

Conan-Kudo:
@davidcassany I don't think so. The problem is that we can't enforce isolation, so it's easy to do malicious things even with the "inner package manager".

davidcassany:
> @davidcassany I don't think so. The problem is that we can't enforce isolation, so it's easy to do malicious things even with the "inner package manager".

Well yes, but we are also not enforcing any other isolation in `config.sh` than calling it in a chroot, there you can call the package manager in any way you want. I believe this would be pretty much the same, isn't it?

jimklimov:
Thanks, I did not manage to get back to this during the week, but thanks for the discussion and many good ideas.

Regarding namespacing: simplifying YAML to "$packagemanager" could be done, though then carefully about dynamic manager name support (so as to not conflict with other valid strings, "xz" et al. I initially wanted a 3-level tree:
````
package_manager:
  - apt-get:
    - options:
       - -o
       - Some:Thing
````
but it is not something the existing routines could handle, per my cursory reading.

As for code style and unit tests... perhaps it would indeed move faster in the acceptable direction if you jumped in, especially if it is some small nuances as it sounds. I believe, on Github repo maintainers have access to change PR contents from forks like mine, so feel free :)
Making different sets of `- something_options` branches can be an option :) assuming we have more places in code where to add such overrides. So far I stumbled on one that blocked us, so just am not aware of others.

Safety-wise, maybe the global `config.yml` maintained by `root` under `/etc` could specify if these (and other eventually?) dangerous options may be interpreted from user configs (and maybe list what user accounts may do so). It is not quite a concern for us here because 1) we manage everything ourselves anyway (so modulo typos, there is little to protect from that we can't break in a less convoluted manner than calling Kiwi for an attack); and 2) our Kiwi's run in VMs dedicated to the task, so there is little to break and nothing irreparable, at least in the confines of a machine.

**Comment by davidcassany** [comment]
> Safety-wise, maybe the global `config.yml` maintained by `root` under `/etc` could specify if these (and other eventually?) dangerous options may be interpreted from user configs (and maybe list what user accounts may do so). It is not quite a concern for us here because 1) we manage everything ourselves anyway (so modulo typos, there is little to protect from that we can't break in a less convoluted manner than calling Kiwi for an attack); and 2) our Kiwi's run in VMs dedicated to the task, so there is little to break and nothing irreparable, at least in the confines of a machine.

I understand security-wise issues might not be a concern for you since you just isolate the build in a VM, but in any case we can't allow the user to shoot himself in a really bad way, like messing around with the host due to some inconsistent or badly understood options provided to the package manager. So the feature (the functionallity) as is in this PR I don't think we can let it go in.

@jimklimov do you think you could share a reproducible example of the failure (or some logs of it) that this PR fixes for you? I'd like to see where and how is it failing, which apt-get command is the offending one. I am wondering if letting the user to add custom options for the repository and/or package manager only for the image packages install phase would make sense and potentially provide something useful to solve your failure.

**Comment by jimklimov** [comment]
I am on trainings this week so mostly without my work computer, but I'll try to get the examples pasted in as soon as I can.

**Comment by jimklimov** [comment]
Here's one of the build logs with the issue which prompted me for this fix:

````
+ kiwi --type=vmx system build --description=. --target-dir=.
[ INFO    ]: 04:18:30 | Loading XML description
[ INFO    ]: 04:18:30 | --> loaded ./config.xml
[ INFO    ]: 04:18:30 | --> Selected build type: vmx
[ INFO    ]: 04:18:30 | Preparing new root system
[ INFO    ]: 04:18:30 | Setup root directory: /tmp/jenkins/workspace/build-OVA-image/ova/build/image-root
[ INFO    ]: 04:18:30 | Setting up repository http://security.debian.org/debian-security
[ INFO    ]: 04:18:30 | --> Type: apt-deb
[ INFO    ]: 04:18:30 | --> Translated: http://security.debian.org/debian-security
[ INFO    ]: 04:18:30 | --> Alias: 7ff84e9d1a022b23bb2550ad85ee8178
[ INFO    ]: 04:18:30 | Setting up repository http://archive-klecker.debian.org/debian/
[ INFO    ]: 04:18:30 | --> Type: apt-deb
[ INFO    ]: 04:18:30 | --> Translated: http://archive-klecker.debian.org/debian/
[ INFO    ]: 04:18:30 | --> Alias: 9072d87942b342d87571984140fe4869
[ INFO    ]: 04:18:30 | Setting up repository http://archive.debian.org/debian/
[ INFO    ]: 04:18:30 | --> Type: apt-deb
[ INFO    ]: 04:18:30 | --> Translated: http://archive.debian.org/debian/
[ INFO    ]: 04:18:30 | --> Alias: e2dc9367da1d6f8bac0902d69f216c32
[ INFO    ]: 04:18:30 | Setting up repository obs://Virtualization:Appliances:CommonBoot/Debian_8.0
[ INFO    ]: 04:18:30 | --> Type: apt-deb
[ INFO    ]: 04:18:31 | --> Translated: http://download.opensuse.org/repositories/Virtualization:/Appliances:/CommonBoot/Debian_8.0/
[ INFO    ]: 04:18:31 | --> Alias: eb12d33f2a821f9bbf8cbcd02331719f
[ INFO    ]: 04:18:31 | Using package manager backend: apt-get
[ INFO    ]: 04:18:31 | Installing bootstrap packages
[ INFO    ]: 04:18:31 | --> collection type: onlyRequired

### THIS ONE ###
[ ERROR   ]: 04:19:25 | KiwiCommandError: chroot: stderr: E: Release file for http://archive-klecker.debian.org/debian/dists/jessie-backports/InRelease is expired (invalid since 35d 7h 53min 42s). Updates for this repository will not be applied.
### Fixed by allowing to ignore expiration state ###

, stdout: Ign http://download.opensuse.org ./ InRelease
Hit http://archive-klecker.debian.org jessie-backports InRelease
Ign http://archive.debian.org jessie InRelease
Get:1 http://download.opensuse.org ./ Release.gpg [189 B]

[ INFO    ]: 04:19:25 | Cleaning up SystemPrepare instance
[ WARNING ]: 04:19:25 | Path /tmp/jenkins/workspace/build-OVA-image/ova/build/image-root/dev not a mountpoint
[ WARNING ]: 04:19:25 | remove_hierarchy: path /tmp is protected
````

**Comment by Conan-Kudo** [comment]
@davidcassany This actually looks more like we should just disable apt's repodata GPG checking in this case. I'm not sure whether we want to, but we do not do GPG checks for RPMs by default either...

**Comment by jimklimov** [comment]
I do not know the workings of `apt` that much, and whether "expiration" comes from GPG or just some timestamp field in Release file etc.

**Comment by Conan-Kudo** [comment]
@jimklimov `Valid-Until` is a property set for the `InRelease` file (with contains combined digest of the file and the GPG signature). The idea is that the `Valid-Until` string makes it so that apt would stop attempting to validate it after that date and force you to request fresher metadata. The hope is that Debian would be regenerating the metadata frequently enough you wouldn't notice... However, the consequence is that archived versions of Debian don't work anymore since `Valid-Until` blocks apt from working properly.

**Comment by jimklimov** [comment]
Okay, thanks for explanation.

The point about checking the expiration for currently active distros (and so noticing that e.g. our internal mirror is way out of sync) still stands though - so disabling the feature for everyone in core is not a good idea.

**Comment by Conan-Kudo** [comment]
@jimklimov This is also why the recommended way to mirror Debian for internal use as "stale"/"frozen" mirrors is to regenerate the metadata. At `$DAYJOB`, that's what we do for our Ubuntu mirrors to work around this.

**Comment by jimklimov** [comment]
I feel like running in circles around Codacy and have no idea why it complains, all descriptions of the error (`self` not used) do not apply - it is used... :(

**Comment by jimklimov** [comment]
Hopefully the code quality is now good enough, as far as standard requirements went. It also seems to still work on our build farm ;)

**Comment by jimklimov** [comment]
If I get the reviews correctly, the "formal" fixes requested (for CI to be green again) are achieved, and the harder question of whether to allow this behavioral change into kiwi remains, with the idea itself gaining praises from some and caution from others... so how do we go from here (and would someone like to take over and add the failsafes etc. to make it less abusable, as discussed above - e.g. via global-only config option to enable it at all)? :)

**Comment by davidcassany** [comment]
@jimklimov Sorry for my late reply. Got several things on my plate to handle last days and also Easter holidays did not help.

I have been having a deeper look to this issue I and I believe we have three options to follow up:

* Make `rpm-check-signatures=false` or `repository_gpgcheck=false` to also include the `Check-Valid-Until=false` option for APT package manager type. I dislike this option, since the `Check-Valid-Until` apt option is not related to GPG checks neither included in the disabled checks for `AllowInsecureRepositories` flag. This would be the simplest solution, but IMHO this is just a misleading workaround.

* Limit user defined customizations **only** for chrooted calls. This is only idea that comes to my mind in other to limit potential issues derived of a miss use of this feature. For KIWI I think this would be secure enough, after all there is `config.sh` which is a user defined script called as root in a chroot.

* Add a new configuration flag on repository level. So we could do something like:

```xml
<repository type="apt-deb" distribution="jessie-backports" components="main" check_valid_until="false">
     <source path="http://archive-klecker.debian.org/debian"/>
</repository>
```

And this would add the `[check-valid-until=no]` for the specific repo. With this we gain the granularity of controlling it on per repo basis, but forces us to add a new APT specific attribute to the `repository` element.

IMHO we could consider adding user defined flags for the chrooted calls to the package manager (probably only at package manager level) with a similar code of what is here suggested in this PR. I could also consider adding a repository attribute if you feel it is worth, but this is not my preferred solution, we can't add a new attribute for every single flag available in apt, dnf or zypper repository definition.

@schaefi, @jimklimov, @Conan-Kudo, Do you think it is reasonable to limit he scope of the user defined flags for the package manager to chrooted calls only?

Conan-Kudo:
@davidcassany `Check-Valid-Until` is related to repo gpgcheck, so I'd wire it into that.

davidcassany:
@jimklimov finally we agreed that enabling the user to add custom options to the package manager is just too risky. So we will not merge this feature, we could not find a sane way to keep this feature secure and not let user harm him/herself. Any of the workaround to sanitize or constrain the user input  makes this feature not useful and/or unmaintainable.

However since your issue is a real concern we should solve we agreed to follow @Conan-Kudo suggestion and just tie the `check-valid-until` apt option to the gpg checks. See #1096, using `repository_gpgcheck="false"` in your expired repository definition should be fine with this patch applied.

I hope this makes sense for your use case.

PR #1038, opened 2019-03-31
Fixed creation of image metadata files

### LINKED ISSUE #1037: KIWI creates an empty package list if rpm not installed in image
If rpm is not installed in the image (possible if only `bootstrap` packages are installed), the created `.packages` and `.verified` files are empty. This is due to:

```
[  122s] [ INFO    ]: 09:10:44 | Export rpm packages metadata
[  122s] [ DEBUG   ]: 09:10:44 | EXEC: [chroot /usr/src/packages/KIWI-docker/build/image-root rpm -E %_dbpath]
[  122s] [ DEBUG   ]: 09:10:44 | EXEC: Failed with stderr: chroot: failed to run command â€˜rpmâ€™: No such file or directory
```

As the bootstrap phase uses the database in the imageroot though, ignoring this failure should result in a working output.

schaefi:
At the end of a build process the metadata information files .packages
and .verified are created. On rpm based systems, rpm is invoked as
chrooted operation inside the new root tree. For images that gets
installed exclusively from the bootstrap phase there is no rpm inside
of the image and the call fails. The result are empty metadata files.
This patch prevents calling rpm inside of the image root tree if it's
not installed and also uses the RpmDataBase interface instead of
directly calling rpm. This Fixes #1037

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -38,6 +38,7 @@ from kiwi.path import Path
 from kiwi.archive.tar import ArchiveTar
 from kiwi.utils.compress import Compress
 from kiwi.utils.command_capabilities import CommandCapabilities
+from kiwi.utils.rpm_database import RpmDataBase
 
 from kiwi.exceptions import (
     KiwiImportDescriptionError,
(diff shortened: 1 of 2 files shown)


Vogtinator:
Copying comment from the issue here:

>Could you change it so that a failure of rpm -E %_dbpath is treated like rpm wasn't installed?
busybox installs a /usr/bin/rpm as well, but it only supports installing packages and not querying.

PR #1041, opened 2019-04-01
Add the possibility to disable runtime checks

### LINKED ISSUE #893: Allow to individually disable runtime checks
## Problem description

In kiwi we have a collection of runtime checks which strictly checks certain conditions to be fulfilled.
However there is no way to skip a check for exceptional reasons. The reason for this issue was
a customer who needs to replace the oem installer code with a custom version that works in his
environment. This installer code needs to be provided by a dracut module package. kiwi has a strict
check on the package provided by us and would fail that check if a customer provides another one.
In such a case it would be good to have an opportunity to skip that check.

Therefore we want to allow an additional section in the ```~/.config/kiwi/config.yml``` as follows

```yaml
runtime_checks:
    - disable: check_dracut_module_for_oem_install_in_package_list
```

The names matches a method name from ```kiwi/runtime_checker.py```

__Why not using kiwi caller options ?__

Caller options are not flexible enough to handle individual runtime checks. In addition they are hard to pass in subsystems like OBS

__Why not use the XML image description ?__

As we consider this feature to be used in rare cases and we don't want to add processing instructions to the image description, we consider this the wrong place

__How to use in OBS ?__

To use the feature in obs, the customer would need to create a package providing ```/root/.config/kiwi/config.yml``` and add that as a ```preinstall:``` in the project config

davidcassany:
This commit adds runtime configuration parameters to disable the runtime
checks. This is specially handy if someone does not want to use the kiwi
dracut modules and provide custom ones instead. In order disable some
runtime check consider a runtime config yaml like:

```yaml
runtime_checks:
    - disable:
        - check_dracut_module_for_oem_install_in_package_list
        - check_dracut_module_for_live_iso_in_package_list
```

This commit fixes #893

### DIFF
--- a/kiwi/runtime_checker.py
+++ b/kiwi/runtime_checker.py
@@ -190,24 +190,33 @@ class RuntimeChecker(object):
             if volume.mountpoint == '/':
                 raise KiwiRuntimeError(message)
 
-    def check_docker_tool_chain_installed(self):
+    def check_container_tool_chain_installed(self):
         """
-        When creating docker images the tools umoci and skopeo are used
-        in order to create docker compatible images. This check searches
-        for those tools to be installed in the build system and fails if
-        it can't find them
+        When creating container images the specific tools are used in order
+        to import and export OCI or Docker compatible images. This check
+        searches for those tools to be installed in the build system and
+        fails if it can't find them
         """
         message_tool_not_found = dedent('''\n
             Required tool {name} not found in caller environment
 
-            Creation of docker images requires the tools umoci and skopeo
-            to be installed on the build system. For SUSE based systems
+            Creation of OCI or Docker images requires the tools {name} and
+            skopeo to be installed on the build system. For SUSE based systems
             you can find the tools at:
 
             http://download.opensuse.org/repositories/Virtualization:/containers
         ''')
         message_version_unsupported = dedent('''\n
             {name} tool found with unknown version
+        ''')
+        message_unknown_tool = dedent('''\n
+            Unknown tool: {0}.
+
+            Please configure KIWI with an appropriate value (umoci or buildah).
+            Consider this runtime configuration file syntax (/etc/kiwi.yml):
+
+            oci:
+                - archive_tool: umoci | buildah
         ''')
 
         expected_version = (0, 1, 0)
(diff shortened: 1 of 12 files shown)


davidcassany:
@schaefi  in addition to the checks disable feature I have added the check 'check_grub_efi_installed_for_efi_firmware' for the `task/system_build`. I saw this test was part of the `task/system_prepare` and I believe this should also be part of `task/system_build`, I guess we just missed to add it to the list.

kiwi/tasks/base.py
```diff
@@ -177,6 +178,21 @@ def sextuple_token(self, option):
...
+        This method runs the given runtime checks excluding the ones disabled
+        in the runtime configuration file.
+
+        :param dict checks: A dictionary with the runtime method names as keys
+            and their arguments list as the values.
+        """
+        exclude_list = self.runtime_config.get_disabled_runtime_checks()
+        if self.runtime_checker is not None:
+            for method, args in {
+                k: v for k, v in checks.items() if k not in exclude_list
```
schaefi: nitpick, can we avoid variables names whose scope is only obvious after reading the code ? I mean k is key and v is value, which would be a nicer read imho

davidcassany: sure let me try with better names

kiwi/runtime_checker.py
```diff
@@ -190,25 +190,29 @@ def check_volume_setup_has_no_root_definition(self):
...
 
             http://download.opensuse.org/repositories/Virtualization:/containers
         ''')
         message_version_unsupported = dedent('''\n
             {name} tool found with unknown version
         ''')
+        message_unknown_tool = dedent('''\n
+            Unknown tool:\'{0}\'. Please set \'oci_archive_tools\' with
+            \'umoci\' or \'buildah\' value'
+        ''')
```
schaefi: Sorry overlooked that one on the first review. I was wondering if that message quoting is needed ? I know to printout single ticks you need to quote them but it looks weird and actually we are suggesting to export a shell variable oci_archive_tools if I read that correctly. In shell people mostly use ```"```. Couldn't we make the error message more explicit and say e.g

```
message_unknown_tool = dedent('''\n
    Unknown tool: {0}. Please setup your environment as follows

    for use with umoci:
    $ export oci_archive_tools=umoci

    for use with buildah:
    $ export oci_archive_tools=buildah 
"'"')
```

or similar

davidcassany: Ok, yes I agree to make it more explicit. What about now?

schaefi:
Looks great now, just a typo:

> W293 blank line contains whitespace

davidcassany:
> Looks great now, just a typo:
> 
> > W293 blank line contains whitespace

At least is not a random Travis failure :rofl:

schaefi:
yes our testing pipeline became useful again since I deleted the stupid linkcheck stuff :)

PR #1042, opened 2019-04-01
Update compression flag for qcow2 format

schaefi:
In case of a qcow2 format we store the result uncompressed
Since the format conversion only takes the real bytes into
account such that the sparseness of the raw disk will not
result in the output format and can be taken one by one
This Fixes bsc#1128146

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -420,7 +420,9 @@ class DiskBuilder(object):
             key='disk_image',
             filename=self.diskname,
             use_for_bundle=True if not self.image_format else False,
-            compress=True,
+            compress=self.runtime_config.get_bundle_compression(
+                default=True
+            ),
             shasum=True
         )
 
--- a/kiwi/builder/filesystem.py
+++ b/kiwi/builder/filesystem.py
@@ -126,7 +126,9 @@ class FileSystemBuilder(object):
             key='filesystem_image',
             filename=self.filename,
             use_for_bundle=True,
-            compress=True,
+            compress=self.runtime_config.get_bundle_compression(
+                default=True
+            ),
             shasum=True
         )
         self.result.add(
--- a/kiwi/builder/pxe.py
+++ b/kiwi/builder/pxe.py
@@ -196,7 +196,9 @@ class PxeBuilder(object):
             key='pxe_archive',
             filename=self.archive_name,
             use_for_bundle=True,
-            compress=False,
+            compress=self.runtime_config.get_bundle_compression(
+                default=False
+            ),
             shasum=True
         )
 
--- a/kiwi/runtime_config.py
+++ b/kiwi/runtime_config.py
@@ -90,7 +90,7 @@ class RuntimeConfig(object):
             obs_public = True
         return bool(obs_public)
 
-    def is_bundle_compression_requested(self):
+    def get_bundle_compression(self, default=True):
         """
         Return boolean value to express if the image bundle should
         contain XZ compressed image results or not.
@@ -102,7 +102,10 @@ class RuntimeConfig(object):
         of the bundle is smaller and the download speed increases.
         However the image must be uncompressed before use
 
-        By default the bundle will contain compressed results.
+        If no compression is explicitly configured, the provided
+        default value applies
+
+        :param bool default: Default value
 
         :return: True or False
 
@@ -112,9 +115,7 @@ class RuntimeConfig(object):
             element='bundle', attribute='compress'
         )
         if bundle_compress is None:
-            # if the bundle compression is not set,
-            # the default is compress image results
-            bundle_compress = True
+            bundle_compress = default
         return bool(bundle_compress)
 
     def get_xz_options(self):
--- a/kiwi/storage/subformat/base.py
+++ b/kiwi/storage/subformat/base.py
@@ -21,6 +21,7 @@ from collections import OrderedDict
 
 # project
 from kiwi.command import Command
+from kiwi.runtime_config import RuntimeConfig
 from kiwi.defaults import Defaults
 from kiwi.path import Path
 from kiwi.logger import log
@@ -50,6 +51,7 @@ class DiskFormatBase(object):
         self.temp_image_dir = None
         self.image_format = None
         self.diskname = self.get_target_file_path_for_format('raw')
+        self.runtime_config = RuntimeConfig()
 
         self.post_init(custom_args)
 
@@ -187,7 +189,9 @@ class DiskFormatBase(object):
                 self.image_format
             ),
             use_for_bundle=True,
-            compress=True,
+            compress=self.runtime_config.get_bundle_compression(
+                default=True
+            ),
             shasum=True
         )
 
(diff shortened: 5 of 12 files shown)


Vogtinator:
This way it's still possible to override the default using the config and thus generate `qcow2.xz` again, right? If so, huge +1 from my side!

schaefi:
> This way it's still possible to override the default using the config and thus generate qcow2.xz again, right?

yes

kiwi/builder/container.py
```diff
@@ -136,7 +136,9 @@ def create(self):
             key='container',
             filename=self.filename,
             use_for_bundle=True,
-            compress=False,
+            compress=self.runtime_config.get_bundle_compression(
```
Vogtinator: Doesn't this mean with force-enabled bundle compression it would result in `.tar.xz.xz` files?

davidcassany: I guess it would happen if one enables the compression of the bundle and also the compression of the image. KIWI build can produce compressed or uncompressed container images, thus if one applies compressed images plus compressed bundle, the result will be a double compression.

Vogtinator: That doesn't make sense, so should IMO not be possible.

schaefi: No because we create container tar's always uncompressed. But the container compression is an option as well. So if you set:

```yaml
container:
      - compress: xz

bundle:
    - compress: true
```

you will end up in the situation you described above. I think the above setup doesn't make much sense but it can be configured that way. Usually I don't prevent people from shooting themselves into their foot ;)

Vogtinator: Understandable, but apparently the container is compressed by default:

```
[  188s] [ DEBUG   ]: 12:00:01 | EXEC: [skopeo copy oci:/tmp/kiwi_oci_dir.xgvnk2l3/oci_layout:latest docker-archive:/usr/src/packages/KIWI-docker/opensuse-tumbleweed-image.x86_64-1.0.4.docker.tar:opensuse/tumbleweed:latest --additional-tag opensuse/tumbleweed:20190327 --additional-tag opensuse/tumbleweed:20190327.29.13]
[  192s] [ DEBUG   ]: 12:00:05 | EXEC: [xz -f --threads=0 /usr/src/packages/KIWI-docker/opensuse-tumbleweed-image.x86_64-1.0.4.docker.tar]
[  211s] [ INFO    ]: 12:00:24 | Export rpm packages metadata
[  211s] [ DEBUG   ]: 12:00:24 | EXEC: [chroot /usr/src/packages/KIWI-docker/build/image-root rpm -E %_dbpath]
[  211s] [ DEBUG   ]: 12:00:24 | EXEC: [rpm --root /usr/src/packages/KIWI-docker/build/image-root -qa --qf %{NAME}|%{EPOCH}|%{VERSION}|%{RELEASE}|%{ARCH}|%{DISTURL}\n --dbpath /usr/lib/sysimage/rpm]
[  211s] [ INFO    ]: 12:00:24 | Export rpm verification metadata
[  211s] [ DEBUG   ]: 12:00:24 | EXEC: [chroot /usr/src/packages/KIWI-docker/build/image-root rpm -E %_dbpath]
[  211s] [ DEBUG   ]: 12:00:24 | EXEC: [rpm --root /usr/src/packages/KIWI-docker/build/image-root -Va --dbpath /usr/lib/sysimage/rpm]
[  215s] [ DEBUG   ]: 12:00:28 | EXEC: [rm -r -f /tmp/kiwi_oci_root_dir._0cy3ji6]
[  215s] [ INFO    ]: 12:00:28 | Result files:
[  215s] [ INFO    ]: 12:00:28 | --> container: /usr/src/packages/KIWI-docker/opensuse-tumbleweed-image.x86_64-1.0.4.docker.tar.xz
[  215s] [ INFO    ]: 12:00:28 | --> image_packages: /usr/src/packages/KIWI-docker/opensuse-tumbleweed-image.x86_64-1.0.4.packages
[  215s] [ INFO    ]: 12:00:28 | --> image_verified: /usr/src/packages/KIWI-docker/opensuse-tumbleweed-image.x86_64-1.0.4.verified
[  215s] LANG=en_US.UTF-8 /usr/bin/kiwi result bundle --target-dir /usr/src/packages/KIWI-docker --id docker-Build29.13 --bundle-dir //usr/src/packages/KIWI.bundle/ 
[  216s] [ INFO    ]: 12:00:29 | Bundle build results from /usr/src/packages/KIWI-docker
```

No special config is set anywhere.

davidcassany: The default container compression is set to compressed, as @Vogtinator noted. Probably we could set container images to be uncompressed and make the bundle compressed by default (the result should be equivalent in terms of OBS) or even drop the compression option for containers images and just have the bundle option.

Vogtinator:
Honoring the bundle compression option for containers might lead to `tar.xz.xz` files, which is IMO a major regression - so I'm against merging it as-is.

davidcassany:
> Honoring the bundle compression option for containers might lead to `tar.xz.xz` files, which is IMO a major regression - so I'm against merging it as-is.

I do not consider it a regression, it does not change container results unless the user modifies the KIWI configuration as @schaefi suggested. Unless I am missing something, for current builds nothing changes.

schaefi:
>I do not consider it a regression, it does not change container results unless the user modifies the KIWI configuration as @schaefi suggested. Unless I am missing something, for current builds nothing >changes.

correct. You have to actively configure it that way and you could have done that before as well. There is no regression at all. If user explicitly states; compress my container and compress my results we just let him to that.

Vogtinator:
I'm not familiar with kiwi code, but for me it looks like this PR makes container bundle compression configurable with the change in `kiwi/builder/container.py`. If that's not the case (did the check get moved from `kiwi/tasks/result_bundle.py`?) then it's fine.

schaefi:
> but for me it looks like this PR makes container bundle compression configurable 

no, the container compression is configurable since this was requested in #725. Don't understand why you think this PR does anything with containers

Vogtinator:
> Don't understand why you think this PR does anything with containers

Because of the diff in the two files I mentioned.

schaefi:
any builder code creates a result object which stores metadata. Part of the metadata is the compress flag. That flag indicates if this result file should be compressed when bundled with the bundler. In the former implementation the global runtime configuration could be used to switch compression on/off for all build results. With your findings about the qcow2 format conversion we could handle that differently but this causes a refactoring to handle the compression setup individually per result file and that affects all other builders as well. This is why you see the one liner change for the compress flag in all builders, including the container builder and the removal of the global evaluation from the bundle tasks

schaefi:
> Understandable, but apparently the container is compressed by default:

you are right. I think it would be good to not evaluate the bundle compress flag for the container building. I'll change that

kiwi/builder/container.py
schaefi: The ability to setup the container compression was requested by iirc b1 systems and we can't just drop that. Having one way to control the container compression is enough and we can simply not evaluate the compress flag set on the bundler for the container builder.

Vogtinator:
Thanks for the explanation!

If the diff does what it looks like, LGTM.

PR #1035, opened 2019-03-29
kiwi-repart: avoid race between repart and fsck

seife:
See issue#1034, this fixes non-booting CentOS-7 OEM images that fail
during repartition / FS resize.
Summary: all parted actions trigger a rescan of partition tables and
this repeatedly starts / stops systemd-fsck-root.service until it
finally fails. The workaround is to mask systemd-fsck-root.service
during repartitioning / FS resize.

Fixes #1034

PR #1043, opened 2019-04-03
Fix regression on kiwi-repart dracut module

davidcassany:
There was a regression introduced in #1035 (048b5c3e). We mask
systemd-fsck-root.service during the repart, but we are not always
unmasking it. If the repart is not possible the service is masked but
not unmasked, which causes the boot to fail.

With this commit after masking the fsck service we trap EXIT and run the
unmask command on script exit. This way we ensure the unmask command is
always executed.

### DIFF
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -260,6 +260,16 @@ function check_repart_possible {
     return 0
 }
 
+function mask_fsck_root_service {
+    info "disable systemd-fsck-root.service"
+    systemctl mask systemd-fsck-root.service
+}
+
+function unmask_fsck_root_service {
+    info "enabling systemd-fsck-root.service"
+    systemctl unmask systemd-fsck-root.service
+}
+
 #======================================
 # Perform repart/resize operations
 #--------------------------------------
@@ -277,8 +287,11 @@ setup_debug
 # *and* the filesystem is brand new ;) by masking it.
 # "systemctl disable" does not work here, because it is event driven
 # More details: https://github.com/SUSE/kiwi/issues/1034
-info "disable systemd-fsck-root.service"
-systemctl mask systemd-fsck-root.service
+
+# make sure we unmask the fsck service
+trap unmask_fsck_root_service EXIT
+
+mask_fsck_root_service
 
 # initialize for disk repartition
 initialize
@@ -319,9 +332,6 @@ if lvm_system; then
 else
     resize_filesystem "$(get_root_map)"
 fi
-
-info "enabling systemd-fsck-root.service"
-systemctl unmask systemd-fsck-root.service
 
 # create swap space
 create_swap "$(get_swap_map)"


davidcassany:
@schaefi thanks to @dcermak I managed to see the regression. It has big impact on oem and other types of images also. The issue is easily reproducible if you add the `dracut-kiwi-oem-repart` package to a VMX image (which should be useless, but not harmful).

The is that the `systemd-fsck-root.service`, in that case, is masked but not unmasked, the script ends before reaching the unmask command because there is nothing to repart. This leads to an ugly boot failure on pre-mount stage. Something similar happens when rebooting an oem image after the first-boot, the service is masked but not unmasked.

@seife in #1035 we missed that the script might return before the umask command. In my tests this leads to errors even on oem reboots after the disk is partitioned, haven't you experimented this issue?

davidcassany:
@schaefi the updated dracut-kiwi-oem-repart is not yet into Factory, so I'll leave a time frame until tomorrow morning to merge and make a new release with the fix unless someone complaints.

dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
```diff
@@ -277,8 +287,10 @@ setup_debug
 # *and* the filesystem is brand new ;) by masking it.
 # "systemctl disable" does not work here, because it is event driven
 # More details: https://github.com/SUSE/kiwi/issues/1034
-info "disable systemd-fsck-root.service"
-systemctl mask systemd-fsck-root.service
+mask_fsck_root_service
+
+# make sure we unmask the fsck service
+trap unmask_fsck_root_service EXIT
```
seife: shouldn't this be the other way round? First trap, then mask?

I know it will not make a difference in practice, but still it looks wrong.

davidcassany: Yes I think you are right

seife:
I think I sort-of tested this case by booting the resized system (where no further resize happens), but of course have missed the case of "image that never even needed a resize".

The fix is obviously good and more robust than my version. Still I'm wondering where the "exit before unmask" could have happened? Because this would mean IMHO that one of the functions that is called between "...mask" and "...unmask" is badly designed and does exit the script, which should be fixed by returning an exit code or at least be annotated clearly.

seife:
found it: m)

```
# resize disk partition table
if lvm_system;then
    repart_lvm_disk || return
else
    repart_standard_disk || return
fi
```

My bad.

davidcassany:
@seife What I am wondering is how it could be that you can reboot without issues. In my tests now I had issues to reboot after the repartioning. Probably because you do not repart up to the whole disk? I think this would make a difference on how the script is executed on `repart_standard_disk`.

seife:
Probably. I was testing it with this image: https://build.opensuse.org/package/show/home:seife:kiwitest-centos/centos7-pxe

which has
```
    <type initrd_system="dracut" filesystem="ext3" bootloader="grub2" installiso="false" installpxe="true" installboot="install" installstick="false" kernelcmdline="net.ifnames=0 video=800x600" fsmountoptions="" firmware="efi" image="oem">
      <oemconfig>
        <oem-boot-title>CentOS-7 KIWI</oem-boot-title>
        <oem-swap>true</oem-swap>
        <oem-systemsize>16384</oem-systemsize>
      </oemconfig>
    </type>
```
Other images I tested with have
```
      <systemdisk name="systemVG">
        <volume name="@root" size="10240M"/>
        <volume name="var" mountpoint="/var" size="6144M"/>
      </systemdisk>
      <oemconfig>
        <oem-swap>true</oem-swap>
      </oemconfig>
```
and
```
      <systemdisk name="systemVG">
        <volume name="@root" size="10240M"/>
        <volume name="/var" size="8192M"/>
      </systemdisk>
      <oemconfig>
        <oem-boot-title>SLES 15 GA KIWI</oem-boot-title>
        <oem-swap>true</oem-swap>
        <oem-device-filter>/dev/(ram|nvme)</oem-device-filter>
      </oemconfig>
```
So none of them does "use the whole disk"...

In my defense: I clearly stated that more tests are probably needed ;-)

davidcassany:
> In my defense: I clearly stated that more tests are probably needed ;-)

No worries you did well, after all the your fix is still very nice. The initrd issues are always hard to detect on reviews and testing it is also not that simple.

schaefi:
@seife yes no worries, changes to the initrd are always sensitive and are mostly found by booting. Our integration test system does not cover that part. There are a few tests based on kanku that also boots, but this is far away from being complete.

PR #1047, opened 2019-04-08
Update bootstrap macro management

davidcassany:
This PR extends the bootstrap macro management to also consider
the case where image was already bootstrap. Note this is a common case
for building derived container images and also the situation when
`--allow-existing-root` flag is in use.

### DIFF
--- a/kiwi/repository/dnf.py
+++ b/kiwi/repository/dnf.py
@@ -105,7 +105,8 @@ class RepositoryDnf(RepositoryBase):
         1. Create the rpm image macro which persists during the build
         2. Create the rpm bootstrap macro to make sure for bootstrapping
            the rpm database location matches the host rpm database setup.
-           This macro only persists during the bootstrap phase
+           This macro only persists during the bootstrap phase. If the
+           image was already bootstrapped a compat link is created instead.
         """
         rpmdb = RpmDataBase(
             self.root_dir, Defaults.get_custom_rpm_image_macro_name()
@@ -114,7 +115,11 @@ class RepositoryDnf(RepositoryBase):
             rpmdb.set_macro_from_string(self.locale[0])
         rpmdb.write_config()
 
-        RpmDataBase(self.root_dir).set_database_to_host_path()
+        rpmdb = RpmDataBase(self.root_dir)
+        if rpmdb.has_rpm():
+            rpmdb.link_database_to_host_path()
+        else:
+            rpmdb.set_database_to_host_path()
 
     def use_default_location(self):
         """
--- a/kiwi/repository/zypper.py
+++ b/kiwi/repository/zypper.py
@@ -154,7 +154,8 @@ class RepositoryZypper(RepositoryBase):
         1. Create the rpm image macro which persists during the build
         2. Create the rpm bootstrap macro to make sure for bootstrapping
            the rpm database location matches the host rpm database setup.
-           This macro only persists during the bootstrap phase
+           This macro only persists during the bootstrap phase. If the
+           image was already bootstrapped a compat link is created instead.
         3. Create zypper compat link
         """
         rpmdb = RpmDataBase(
@@ -164,7 +165,11 @@ class RepositoryZypper(RepositoryBase):
             rpmdb.set_macro_from_string(self.locale[0])
         rpmdb.write_config()
 
-        RpmDataBase(self.root_dir).set_database_to_host_path()
+        rpmdb = RpmDataBase(self.root_dir)
+        if rpmdb.has_rpm():
+            rpmdb.link_database_to_host_path()
+        else:
+            rpmdb.set_database_to_host_path()
         # Zypper compat code:
         #
         # Manually adding the compat link /var/lib/rpm that points to the
--- a/kiwi/utils/rpm_database.py
+++ b/kiwi/utils/rpm_database.py
@@ -108,6 +108,33 @@ class RpmDataBase(object):
         )
         self.rpmdb_image.write_config()
 
+    def link_database_to_host_path(self):
+        """
+        Create a link from host database location to image database
+        location. The link is only created if both locations differ
+        and if the host database location does not exist.
+        """
+        rpm_host_dbpath = self.rpmdb_host.expand_query('%_dbpath')
+        rpm_image_dbpath = self.rpmdb_image.expand_query('%_dbpath')
+        if rpm_host_dbpath != rpm_image_dbpath:
+            root_rpm_host_dbpath = os.path.normpath(
+                os.sep.join([self.root_dir, rpm_host_dbpath])
+            )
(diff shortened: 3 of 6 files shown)


davidcassany:
@schaefi see that these fixes the builds that **do not** include any `<packages type="bootstrap">`. If there was nothing to bootstrap (as it usually happens with derived container images), the `setup_package_database_configuration` method executed in https://github.com/SUSE/kiwi/blob/18bb65b3b0314ecf68d88f9475386fb8ddae4c0a/kiwi/system/prepare.py#L114 is not reverted by calling `post_process_install_requests_bootstrap` in https://github.com/SUSE/kiwi/blob/18bb65b3b0314ecf68d88f9475386fb8ddae4c0a/kiwi/system/prepare.py#L215 because there was no bootstrap at all.

davidcassany:
I have included the Work in progress label because, IMHO, this reveals that `post_process_install_requests_bootstrap` and `setup_package_database_configuration` should be part of the same class. In fact, I'd say we could move `setup_package_database_configuration` to `PackageManager` classes and even probably rename it to something like `pre_process_install_requests_bootstrap`.

Does it make sense to you?

schaefi:
sorry but this change will break the signing keys setup. The methods are called intentionally that early to make sure the signing keys works in any case. There is a long thread on the ML with Till about the signing keys problematic with the rpmdb location. I doubt this one can be merged

Please explain why we really need this. So from my perspective the bootstrap only install is a special case. I would be ok if we delete the rpm database from the image completely because people that install the system from the host intentionally don't want rpm inside of the image. The cost is they loose information about the installed software which I think is a fair price. If we don't go that way it must be clear to users that the bootstrap only install brings in an rpm database as it exists on the host, even if the image distro would provide it elsewhere when it would be installed.

So there is a price to pay for a bootstrap only install and I don't think we should play tricks with the macros. Even if someone installs bootstrap only, there will be an rpm database and a macro setup during the installation phase initialized and used inside of the image. We intentionally implemented it that way to e.g allow the support for signed key package verification during bootstrap.

Let me know what you think

schaefi:
Also does the current implementation break anything in a bootstrap only setup ? For me the "missing cleanup" is a cosmetic issue, however the changes here breaks things. I'd like to understand the real motivation which is currently not clear to me. Thanks

schaefi:
>This commit moves the setup_package_database_configuration method
>execution as part of the boostrap phase. This way the rpmdb
>configuration via macros only happens if there is actually something to
>boostrap.

This would be the wrong motivation because at least one concept, the signing keys verification will break. This needs to be done before bootstrap. The entire concept of the rpm macros is based on that it exists as early as possible. This also applies to bootstrap only installations. Because even in this mode there is rpm existing during build time inside of the image

I would be ok if we delete all of the rpmdb and the macros in a bootstrap only install. I would also be ok to just keep the current state.

davidcassany:
> sorry but this change will break the signing keys setup. The methods are called intentionally that early to make sure the signing keys works in any case. There is a long thread on the ML with Till about the signing keys problematic with the rpmdb location. I doubt this one can be merged

Oh right, good point. I missed that...

> Please explain why we really need this. So from my perspective the bootstrap only install is a special case. 

This does nothing to do with the bootstrap only install. This is the other way round. Consider the test we have for derived docker images. There is no `<packages type="bootstrap">` at all, this leads to exit the `install_bootstrap` here https://github.com/SUSE/kiwi/blob/18bb65b3b0314ecf68d88f9475386fb8ddae4c0a/kiwi/system/prepare.py#L179

So the `post_process_install_requests_bootstrap` is never run before the system install. In that case if host and image have different rpmdb configurations the build breaks due to an inconsistent rpmdb configuration.

We could modify this suggested change and do it the other way round. Instead of moving the execution point of `setup_package_database_configuration`, we can include the `post_process_install_requests_bootstrap` before returning if no bootstrap is defined. That would cover at least the specific use case I saw.

Notice that this rpmdb management we thought does not take into consideration the use case when we use the root_import interface (currently only available for containers via `drived_from`). In those cases we should verify if there is already some rpmdb populated and available.

Thinking it twice probably the best option, to completely cover this cases, would be to make `setup_package_database_configuration` smarter so it does not modify the dbpath (or adds a compat link or any similar strategt) if there is already some populated rpmdb available. I'll make some tests.

I saw that testing the fedora example I did long ago in this [davidcassany/kiwi-derived-images](https://github.com/davidcassany/kiwi-derived-images) repo from an openSUSE host.

schaefi:
Thanks for the explanation. I was looking from a wrong perspective. Running from a derived image that was already bootstraped. I got it and yes that requires us to take care:

> Thinking it twice probably the best option, to completely cover this cases, would be to make
> setup_package_database_configuration smarter so it does not modify the dbpath (or adds a compat
> link or any similar strategt) if there is already some populated rpmdb available. I'll make some tests.

That sounds reasonable to me but I don't know how you would detect that. Maybe we omit ``` setup_package_database_configuration``` in case a derived image or ```--allow-existing-root``` is given and assume that existing data has been correctly bootstraped by kiwi at the time it was created.

schaefi:
@davidcassany any news on this one ?

davidcassany:
> @davidcassany any news on this one ?

I am testing the implementation I did. I expect to update the PR today. I am trying to make a good testing. I am building Fedora, CentOS and openSUSE vmx and derived containers from openSUSE and from Fedora hosts.

davidcassany:
Closed by mistake, pressed the wrong button

davidcassany:
@schaefi Currently I have tested CentOS, Fedora and openSUSE builds from [davidcassany/kiwi-derived-images](https://github.com/davidcassany/kiwi-derived-images) using a Tumbleweed host. I also tested CentOS and Fedora VMX from [SUSE/kiwi-descriptions](https://github.com/SUSE/kiwi-descriptions) using the same Tumbeweed host, but in this case for each build I did a first regular run and after a second one using the `--allow-existing-root` with some additional packages in image and bootstrap sections. Finally from a Fedora 28 host I did the same container builds and tested to build the Leap 15.0 VMX from [SUSE/kiwi-descriptions](https://github.com/SUSE/kiwi-descriptions), again twice, one without `--allow-existing-root` and a second one with an already populated tree and some bootstrap and image additions.

I believe this is quite a good test and also the changes I introduced are only changing the behavior when the dbpaths differ, thus they are transparent to most of cases. I still have pending to test some builds adding signatures validation with custom keys, gonna try to find some time to test it during theses days, that's why I haven't removed the work in progress label yet.

davidcassany:
All right I managed to also build openSUSE and Fedora VMX from both Tumbleweed and Fedora hosts passing the repository keys within the command line using the `--signing-key` flag and `<rpm-check-signatures>true</rpm-check-signatures>`. I don't have further tests in my mind to verify this PR.

PR #1048, opened 2019-04-09
Adding custom curl options for dracut-kiwi-lib module

### LINKED ISSUE #891: Allow custom curl options in fetch_file
## Problem description

KIWI's dracut net lib provides a fetch_file method which actually calls curl.
curl supports nice retry and delay options like explained here:

* https://unix.stackexchange.com/questions/428233/download-big-file-over-bad-connection

All of that is probably use case and environment specific. Thus it would be nice to allow
passing in curl options via cmdline e.g

```
rd.kiwi.install.pxe.curl_options
```

davidcassany:
This commit parses the `rd.kiwi.install.pxe.curl_options` argument from
the kernel command line to read and use curl options for the
`fetch_file` function of `kiwi-net-lib.sh` utility. Options are passed
as comma separated values.

Fixes #891

### DIFF
--- a/doc/source/building/build_oem_disk.rst
+++ b/doc/source/building/build_oem_disk.rst
@@ -207,6 +207,32 @@ target system:
    and initrd could be fetched from any server and doesn't have to be stored
    on the `PXE_SERVER`.
 
+   By default KIWI does not use specific `curl` options or flags. However it
+   is possible to add custom ones by adding the 
+   `rd.kiwi.install.pxe.curl_options` flag into the kernel command line.
+   `curl` options are passed as comma separated values. Consider the following
+   example:
+
+   .. code:: bash
+
+       rd.kiwi.install.pxe.curl_options=--retry,3,--retry-delay,3,--speed-limit,2048
+
+   The above tells KIWI to call `curl` with:
+
+   .. code:: bash
+
+       curl --retry 3 --retry-delay 3 --speed-limit 2048 -f <url>
+
+   This is specially handy when the deployment infraestructure requires
+   some fine tuned download behavior. For example, setting retries to be
+   more robust over flaky network connections.
+
+   .. note::
+
+      KIWI just replaces commas with spaces and appends it to the
+      `curl` call. This is relevant since command line options including
+      commas will always fail.
+
    .. note::
 
       The initrd and Linux Kernel for pxe boot are always loaded via tftp
--- a/doc/source/building/working_with_images/custom_volumes.rst
+++ b/doc/source/building/working_with_images/custom_volumes.rst
@@ -55,7 +55,7 @@ definition and some additional empty space for the root volume:
 
 .. code:: xml
 
-   <type ...>
+   <type>
      <systemdisk name="vgroup-name">
        <volume name="@root" freespace="5G"/>
        <volume name="home" size="40G"/>
--- a/dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh
@@ -10,7 +10,14 @@ function fetch_file {
     local source_uncompressed_bytes=$2
     local fetch_info=/tmp/fetch.info
     local fetch
-    fetch="curl -f ${source_url}"
+    local curl_options
+    curl_options=$(getarg rd.kiwi.install.pxe.curl_options=)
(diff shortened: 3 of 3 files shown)


davidcassany:
@schaefi I am not very happy about how this looks like once applied, I mean how it's usage is translated to the kernel command line. However the alternatives I thought they are not convincing me either.

Consider I want to call curl with:

```bash
curl --verbose --retry 3 --retry-delay 3 --speed-limit 2048 -f <url>
```

Then I should append to the kernel command somthing like:

```
rd.kiwi.install.pxe.curl_options=--verbose,--retry,3,--retry-delay,3,--speed-limit,2048
```

It looks ugly, but the implementation is so simple.

I also thought about something lilke
```
rd.kiwi.install.pxe.curl_options=verbose,retry=3,retry-delay=3,speed-limit=2048
```

Which looks nicer but I noticed that `--retry=3` is not valid for curl, it has to be `--retry 3`, so we would need to replace `=` for a space and also append `--` for each option, meaning only long explicit options would be possible.

I also wanted to test with
```
rd.kiwi.install.pxe.curl_options=" --verbose --retry 3 --retry-delay 3 --speed-limit 2048"
```

which is also not very nice, but easy to understand, but I could not manage to get options including spaces. `getarg` is not prepared for that and I don want to start parsing the command line with custom parsers.

Any thoughts?

schaefi:
Thanks for your effort on this one. The code looks good to me :+1: 

I think we should avoid trying to be clever on the commandline parsing. So I like the way you did it because the only parsing is a separation by ```,``` and that will not be part of any option. Thus I consider this a safe operation for any option curl might have.

From a usability aspect I would go and document how curl expects the options. There is a very good documentation about the cmdline parsing here:

* https://ec.haxx.se/cmdline-options.html

As you already found out with dracut's getarg implementation we can't pass in quoted text which
requires us to decide for an option separator. That is ```,``` and I think this is good enough

```
rd.kiwi.install.pxe.curl_options=--verbose,--retry,3,--retry-delay,3,--speed-limit,2048
```

It looks not really nice but imho is pretty clear.

I'm just missing a documentation entry that shortly explains rd.kiwi.install.pxe_curl_options
and after that I think this is good to go.

davidcassany:
> I think we should avoid trying to be clever on the commandline parsing. 

Fully agree, that's why I finally opted for the most straight forward option I saw.

> From a usability aspect I would go and document how curl expects the options. There is a very good documentation about the cmdline parsing here:
> 
>     * https://ec.haxx.se/cmdline-options.html

Interesting, that makes me realize that this implementation I did does not allow curl arguments with spaces (or at least in any sane way, there is always the option of replacing spaces by `,`). I could live with that, but this is something to document.
 
> I'm just missing a documentation entry that shortly explains rd.kiwi.install.pxe_curl_options
> and after that I think this is good to go.

Where you think we should document that? What about adding some additional information (or subsection) in [network deplyoment](https://opensource.suse.com/kiwi/building/build_oem_disk.html#network-deployment) ? or would you suggest a new section under [working with images](https://opensource.suse.com/kiwi/building/working_with_images.html)?

I'd probably go for increasing [network deplyoment](https://opensource.suse.com/kiwi/building/build_oem_disk.html#network-deployment) but I have no strong opinion.

schaefi:
> I'd probably go for increasing network deplyoment but I have no strong opinion.

I would it put there too

davidcassany:
> I'm just missing a documentation entry that shortly explains rd.kiwi.install.pxe_curl_options
> and after that I think this is good to go.

Ok agree, I added some text to the oem network deployment docs.

schaefi:
@davidcassany 

The doc build complains with a warning:

```
writing output... [100%] quickstart                                             
/home/ms/Project/kiwi/doc/source/building/working_with_images/custom_volumes.rst:56: WARNING: Could not lex literal_block as "xml". Highlighting skipped.
/home/ms/Project/kiwi/doc/source/building/working_with_images/vmx_setup_for_google.rst:53: WARNING: Could not lex literal_block as "xml". Highlighting skipped.
```

This is because of the xml tag you used in the ..code example.

You can reproduce this with a

```
$ cd doc
$ make html
```

in a venv

davidcassany:
Thanks, I missed that. Now I don't get it anymore in my env, I got confused since this was not introduced by my last commit.

schaefi:
Maybe you created the branch from a not up to date master

PR #1052, opened 2019-04-10
Fix rpm metadata creation

schaefi:
There are rpm versions that require access to /dev and fail
with the error message 'Failed to initialize NSS library'
Due to the rpm location checkup fix from Issue #1037 this
change impacts builds for distributions that uses an rpm
version with the above behavior (e.g CentOS) and fails as
consequence of the rpm call error. This commit fixes this
in a way that we make sure the host /dev is available at
the time of the call.

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -23,6 +23,7 @@ from collections import namedtuple
 from tempfile import NamedTemporaryFile
 
 # project
+from kiwi.mount_manager import MountManager
 from kiwi.system.uri import Uri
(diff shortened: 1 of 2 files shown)


schaefi:
I finally found the reason for the CentOS build problems. This was not easy to find. Also see:
https://bugs.centos.org/view.php?id=14767

Conan-Kudo:
@schaefi You might want to put a reference to the bug in the commit message so that it won't get lost in the future...

schaefi:
yep done, I've updated the commit message

PR #1056, opened 2019-04-14
Update suse integration tests per Factory changes

schaefi:
The way plymouth themes are provided has changed on suse.
The package plymouth-branding-openSUSE is no longer providing
the theme named openSUSE. In fact the plan is to switch to
the upstream bgrt theme which is provided in another package.
This commit adapts to the changes in the distribution

### DIFF
--- a/build-tests/x86/test-image-iso/appliance.kiwi
+++ b/build-tests/x86/test-image-iso/appliance.kiwi
@@ -14,7 +14,7 @@
         <timezone>Europe/Berlin</timezone>
         <rpm-excludedocs>true</rpm-excludedocs>
         <rpm-check-signatures>false</rpm-check-signatures>
-        <bootsplash-theme>openSUSE</bootsplash-theme>
+        <bootsplash-theme>bgrt</bootsplash-theme>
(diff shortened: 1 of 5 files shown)


build-tests/x86/test-image-iso/appliance.kiwi
```diff
@@ -31,7 +31,7 @@
         <package name="bind-utils"/>
         <package name="patterns-openSUSE-base"/>
         <package name="systemd"/>
-        <package name="plymouth-branding-openSUSE"/>
+        <package name="plymouth-branding-bgrt"/>
```
Conan-Kudo: Do we have a `plymouth-branding-bgrt` package? I thought we still have `plymouth-branding-openSUSE`? We also need `plymouth-theme-bgrt` in the package set.

schaefi: No we don't have one, I'm just stupid :) Updated the PR

I changed the iso build test to use this branch for building and it succeeded:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86/test-image-iso

PR #1064, opened 2019-04-18
Preserve license and other txt files in baseStripFirmware (bsc#1132455)

dirkmueller:
LICENSES are usually not large and should be kept alongside
of the binaries. Also some firmware files sideload additional
txt files (like for example brcmfmac43430 needs the sdio description
txt files). We should just always include them because they're
not listed as needed files.

Fixes # 1132455.

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -751,6 +751,15 @@ function baseStripFirmware {
             done
         done
     done
+    # Preserve licenses and txt files (which are needed for some firmware blobs)
+    find /lib/firmware \( -name 'LICENSE*' -o -name '*txt' \) -print | while read -r match; do
+        if [ -e "${match}" ];then
+            match=$(echo "${match}" | sed -e 's@\/lib\/firmware\/@@')
+            bmdir=$(dirname "${match}")
+            mkdir -p "/lib/firmware-required/${bmdir}"
+            mv "/lib/firmware/${match}" "/lib/firmware-required/${bmdir}"
+        fi
+    done
     rm -rf /lib/firmware
     mv /lib/firmware-required /lib/firmware
 }


dirkmueller:
(untested)

dirkmueller:
testing this in https://build.opensuse.org/package/show/home:dirkmueller:branches:openSUSE:Factory:ARM:Live/python-kiwi

kiwi/config/functions.sh
```diff
@@ -751,6 +751,15 @@ function baseStripFirmware {
             done
         done
     done
+    # Preserve licenses and txt
+    find /lib/firmware -name 'LICENSE*' -o -name '*txt' -print | while read match; do
```
dcermak:
```suggestion
    find /lib/firmware \( -name 'LICENSE*' -o -name '*txt' \) -print | while read match; do
```

ggardet:
Does it handle files with spaces, such as `/lib/firmware/brcm/brcmfmac43430a0-sdio.ONDA-V80 PLUS.txt`?

schaefi:
@dirkmueller Thanks for effort. In addition to the already raised concerns, could you also make sure the code passes shellcheck. As you can see from the tests it complains in here:

```
check run-test: commands[3] | bash -c 'shellcheck -e SC1091,SC1090,SC2001,SC2174,SC1117 /home/travis/build/SUSE/kiwi/kiwi/config/functions.sh -s bash'
In /home/travis/build/SUSE/kiwi/kiwi/config/functions.sh line 755:
    find /lib/firmware -name 'LICENSE*' -o -name '*txt' -print | while read match; do
                                                        ^-- SC2146: This action ignores everything before the -o. Use \( \) to group.
                                                                       ^-- SC2162: read without -r will mangle backslashes.
```

schaefi:
Another question that came to my mind. It seems you are on a legacy path still using custom kiwi boot descriptions instead of dracut. I didn't had  a look to the bug reference you added but if ever possible you should switch. This would also eliminate any use of baseStripFirmware since dracut's module handling would kick in

schaefi:
@dirkmueller JFI: https://github.com/SUSE/kiwi/tree/master/build-tests/arm/test-image-rpi-oem I moved the arm raspberry build to dracut and avoided and boot="oemboot/..." stuff

kiwi/config/functions.sh
```diff
@@ -751,6 +751,15 @@ function baseStripFirmware {
             done
         done
     done
+    # Preserve licenses and txt
+    find /lib/firmware \( -name 'LICENSE*' -o -name '*txt' \) -print | while read match; do
```
dcermak: I missed this one:
```suggestion
    find /lib/firmware \( -name 'LICENSE*' -o -name '*txt' \) -print | while read -r match; do
```

dirkmueller:
@schaefi thanks for all the comments. I have no means to mark this PR as "draft" anymore afaik after I accidentally clicked on create PR earlier. I'll test this first. We do use dracut based initrd for quite some time already, but we still run stripFirmware (afaik stripFirmware still does something useful with dracut initrd, will double check).

schaefi:
@davidcassany could you have a short look too. It looks good to me now

schaefi:
@dirkmueller before we merge can you please squash your commits into one and make sure the commit log message contains the reference to the bug id (bsc#1132455). We create the changelog from the git history so the commit log message should contain that information. Thanks

schaefi:
yes we just need the commit squash and the bnc id in the commit message change. Setting blocked label until then

dcermak:
@schaefi You can push to @dirkmueller's branch as he allowed edits from maintainers.

schaefi:
> You can push to @dirkmueller's branch as he allowed edits from maintainers.

well yes but not sure if this would be a nice move ;)

dirkmueller:
@schaefi why are you so  impatient? I still wanted to test the change ;) It seems to work, new images built with that look a lot better. also squashed and bsc# comment was added already in the first iteration. anything else?

schaefi:
> @schaefi why are you so impatient? 

yeah sorry, it looked all good and just wanted to get it in :) Thanks for the update

PR #1065, opened 2019-04-18
Simplify the creation of .profile variables

dcermak:
Most of the variables put into the dictionary `self.dot_profile` are created via a common scheme:
``` Python
dot_profile['kiwi_NAME'] = type_section.get_NAME()
```

This commit converts the duplicate settings into a loop.

### DIFF
--- a/kiwi/system/profile.py
+++ b/kiwi/system/profile.py
@@ -274,50 +274,25 @@ class Profile(object):
         # kiwi_btrfs_root_is_snapshot
         # kiwi_startsector
         type_section = self.xml_state.build_type
+        for var_name in (
+                'compressed', 'wwid_wait_timeout', 'hybridpersistent',
+                'hybridpersistent_filesystem', 'ramonly', 'target_blocksize',
+                'target_removable', 'firmware', 'bootloader',
+                'bootloader_console', 'btrfs_root_is_snapshot',
+                'gpt_hybrid_mbr', 'devicepersistency', 'installboot',
+                'bootkernel', 'fsmountoptions', 'bootprofile', 'vga'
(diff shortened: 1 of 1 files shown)


kiwi/system/profile.py
```diff
@@ -274,50 +274,25 @@ def _type_to_profile(self):
...
+                'compressed', 'wwid_wait_timeout', 'hybridpersistent',
+                'hybridpersistent_filesystem', 'ramonly', 'target_blocksize',
+                'target_removable', 'firmware', 'bootloader',
+                'bootloader_console', 'btrfs_root_is_snapshot',
+                'gpt_hybrid_mbr', 'devicepersistency', 'installboot',
+                'bootkernel', 'fsmountoptions', 'bootprofile', 'vga'
+        ):
+            self.dot_profile['kiwi_' + var_name] = \
+                getattr(type_section, 'get_' + var_name)()
+
```
schaefi: Hmm, sorry I don't like this hard coupling of a method name with the variable name of a shell environment. Please note the interface from the XML is auto generated code via generateDS. The method name e.g get_compressed() is created from the attribute named ```compressed``` in the Relax schema. The code above basically forces into a naming schema and now creates a direct hard coupling between schema attributes, auto generated reader interface and variable names in the shell env.

Variable names in shell apply different rules than attributes in XML. I'm sorry in this case I rather write more code than creating that dependency. In addition we avoid adding more to that shell env and most of the stuff is there only for compat reasons (custom kiwi boot code). So I expect this code to be not changed.

schaefi:
Let us not touch this for the moment. Sorry

PR #1070, opened 2019-04-23
Make the user.password attribute mandatory

### LINKED ISSUE #1061: Not providing a user element with a password results in a usermod or openssl error
## Problem description

The following completely legal (but obviously wrong) user configuration:
``` xml
    <users>
        <user home="/root" name="root" />
    </users>
```
is accepted by KIWI, but results in a `usermod` error:
```
[   54s] [ DEBUG   ]: 21:46:17 | EXEC: Failed with stderr: usermod: no options
[   54s] Usage: usermod [options] LOGIN
[   54s]
[   54s] Options:
```

Adding `pwdformat="plain"` into the `user` element results in an erroneous `openssl` invocation.

## Expected behavior

KIWI should reject the above XML snippet. I would expect the attribute `password` to be mandatory.

1. Take a working KIWI image description
2. Delete the `password="foo"` entry from a arbitrary `user` element
3. Build the image

+ KIWI version: 9.17.37

dcermak:
Not providing a user password results in an error when `usermod` or `openssl` is later called by kiwi (depending on the value of `pwdformat`).

Fixes #1061.

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -1772,7 +1772,7 @@ div {
         k.user.home.attribute &
         k.user.id.attribute? &
         k.user.name.attribute &
-        k.user.password.attribute? &
+        k.user.password.attribute &
         k.user.pwdformat? &
         k.user.realname.attribute? &
         k.user.shell.attribute?
--- a/kiwi/schema/kiwi.rng
+++ b/kiwi/schema/kiwi.rng
@@ -2728,9 +2728,7 @@ to the user according to he specifing toolchain behaviour.</a:documentation>
           <ref name="k.user.id.attribute"/>
         </optional>
         <ref name="k.user.name.attribute"/>
-        <optional>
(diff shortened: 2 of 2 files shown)


kiwi/xml_parse.py
```diff
@@ -3,7 +3,7 @@
 
 #
 # Generated  by generateDS.py version 2.29.14.
-# Python 3.6.5 (default, Mar 31 2018, 19:45:04) [GCC]
+# Python 3.7.2 (default, Dec 30 2018, 16:18:15) [GCC]
```
davidcassany: Were those changes done manually? I'd have expected them to be autogenerated by calling `make kiwi/schema/kiwi.rng` in which case I'd also expect it to modify the project path... Anyway, those are not relevant changes, so I would just omit any change here and keep the original file.

dcermak: I'll drop that, I just thought to bump the Python version string in that file so that we appear to be using the newest stuff :wink:

PR #1074, opened 2019-04-25
XML description

dcermak:
This is split of #1024 (unfortunately still huge).

TODO:
- [x] drop the first 3 commits once #1069 get's merged and rebase the PR afterwards

Changes proposed in this pull request:
*  Add a new doc folder: `working_with_kiwi` that shall replace `overview`
* Document the RNG schema as it existed for the old KIWI
* Remove `doc/source/building/working_with_images/uninstall_package_requests.rst` as I have incorporated that into `xml_description.rst`

PR #1082, opened 2019-05-02
Followup to #1074

dcermak:
Followup to #1074 addressing the comments from @davidcassany.

TODO: 
- [ ] add a rng conversion thingy (help me @schaefi, you're my only hope)

* Correct the wrong footnote in `xml-description.rst`
* remove the obsolete SolverRepositorySUSE (this one was made obsolete by #1067)
* remove the attributes/parameters `user`, `secret` & `credentials` from `Repository` & `Repository.add_repo()` (and in the child classes)
* remove `user` & `secret` from `SolverRepository*` classes
* remove `Uri.credentials_file_name()` (no longer needed)
* remove obsolete tests & cleanup existing ones


schaefi:
Sorry no way that I merge this. There are way too many changes on different levels of the code. I agree to some of them e.g SolverRepositorySUSE but other parts like the user & secret from repos we won't delete. First because it changes the public api, second because there are still repos supported with user and pass and zypper also supports this. third it does not cause any harm, and last but not least it's incomplete because the XSL transformation of the deletion of attributes is missing. With the last change in this direction (repo type) I could simply delete the type and put on the auto detection. Now consider a user has a repo setup with user & pass what should be the correct transformation ? there is none because we actually would delete a feature

I really appreciate your hard work but changes should be done in smaller steps. Also the topic of a PR and the initial comment should tell the reason for the change. As a followup to a documentation improvement I would not expect a change in the implementation.

Really sorry but we can't go that way

PR #1083, opened 2019-05-03
Cleanup TODO & FIXME from xml_description.rst

dcermak:
Followup to #1074


schaefi:
Thanks, much appreciated. I overlooked that part in the origin PR. The funny thing is that I read through the text and I remember the @schaefi but did not realize this is not part of the actual documentation :-)) blind reader

PR #1084, opened 2019-05-03
Fix locale setting

### LINKED ISSUE #1081: Locale setup is incomplete, LANG is not configured
## Problem description
The default locale for users created differs from building under openSUSE Leap 42.3 to building under openSUSE Leap 15.0.

The guest OS in both cases is 42.3.

This is the problematic case (build host Leap 15.0):
```
builton150:~ # su - user
user@builton150:~> locale
LANG=
LC_CTYPE="POSIX"
[...]
```

I'm guessing this issue is somewhat related to #985.

## Expected behaviour
This is how it previously was (build host Leap 42.3):
```
builton423~ # su - user
user@builton423~> locale
LANG=de_DE.UTF-8
LC_CTYPE="de_DE.UTF-8"
[...]
```

Generate an ISO for Leap 42.3 under a Leap 15.0 build host.

+ KIWI version: 9.17.31
+ Operating system: Build host: openSUSE Leap 15.0, guest host: openSUSE Leap 42.3
+ OBS version: n.a.

davidcassany:
For pre-Leap 15 openSUSE versions KIWI >= 9.12.0 was not completely
setting locale, as it was missingto set the RC_LANG variable from
`/etc/sysconfig/language` file. Current commit enforces to update
locale in `/etc/sysconfig/language` (if the file exists) at the
same time it applies systemd-firstboot configurations.

Fixes #1081

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -324,17 +324,12 @@ class SystemSetup(object):
                     'chroot', self.root_dir, 'systemd-firstboot',
                     '--locale=' + locale
                 ])
-            elif os.path.exists(self.root_dir + '/etc/sysconfig/language'):
+            if os.path.exists(self.root_dir + '/etc/sysconfig/language'):
                 Shell.run_common_function(
                     'baseUpdateSysConfig', [
                         self.root_dir + '/etc/sysconfig/language',
                         'RC_LANG', locale
                     ]
-                )
-            else:
-                log.warning(
-                    'locale setup skipped no capable '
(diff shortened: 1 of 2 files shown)


davidcassany:
With this changes I did a Leap 42.3 and Leap 15 builds and both had the same behavior, locale was set system-wide for non-root users. Before this commit builds for Leap 42.3 were not loading the locale configuration as expected. I could not see any difference in behavior for Leap 15 builds after this commit.

PR #1085, opened 2019-05-06
Discard default dependencies for sysroot.mount

### LINKED ISSUE #1015: Live images fail to boot due to sysroot.mount race

davidcassany:
This commit makes default dependencies from sysroot.mount to be
explicitly omitted. This fixes potential inconsistencies in
ordering pre-mount.service with local-fs.target. This change is
also applied to upstream sysroot.mount generator here:

https://github.com/systemd/systemd/pull/12281

Fixes #1015

### DIFF
--- a/dracut/modules.d/90kiwi-live/kiwi-generator.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-generator.sh
@@ -33,6 +33,7 @@ ROOTFLAGS="$(getarg rootflags)"
 {
     echo "[Unit]"
     echo "Before=initrd-root-fs.target"
+    echo "DefaultDependencies=no"
     echo "[Mount]"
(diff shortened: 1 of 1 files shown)


Vogtinator:
Can you submit this to Tumbleweed? Tests using live media are failing in about 20% of all cases because of this.

Vogtinator:
Ping!

Vogtinator:
Another ping. I'll submit it myself, but I you'll have to accept the review.

PR #1087, opened 2019-05-08
kiwi-live-lib: mount live ISO as read-only

aplanas:
During the boot process of a live image, dracut shows this WARNING:

dracut-initqueue: mount: /run/initramfs/live: WARNING: device write-protected, mounted read-only

This is not a problem, as the live ISO image is, indeed, read-only.

This patch fix this cosmetic issue being explicit in the mount
options in `mountIso` function.


schaefi:
Thanks looks good to me :+1:

PR #1088, opened 2019-05-09
Turn sphinx warnings into errors

dcermak:
* Modify the sphinx Makefile to treat warnings (e.g. undefined references) as errors

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -15,7 +15,7 @@ endif
 # Internal variables.
 PAPEROPT_a4     = -D latex_paper_size=a4
 PAPEROPT_letter = -D latex_paper_size=letter
-ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) source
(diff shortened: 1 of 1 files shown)


doc/Makefile
```diff
@@ -54,7 +54,7 @@ clean:
 
 .PHONY: html
 html:
-	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html && cp -a source/development/schema/images $(BUILDDIR)/html/development || true
+	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html && cp -a source/development/schema/images $(BUILDDIR)/html/development
```
schaefi: You need to revert this line. The copy of the static schema images is the result of the oxygen build which we apply only for the only documentation. It's basically what you can see here: https://opensource.suse.com/kiwi/development/schema.html. However this is only done in the environment where the suse oxygen license allows this. Thus on standard operation there are no images and we put on our opensource schema doc tool that does not create nice images like oxygen. This is probably also the reason why the gitlab pipeline failed. So for the moment please delete that part. Thanks

schaefi: I'd like to continue and get this merged. Thus I hope you don't mind I did the needed change on the branch.

PR #1089, opened 2019-05-09
Improve the documentation about building in the Build Service

dcermak:
* Incorporate @mlschroe's comments about building in OBS into the documentation

### DIFF
--- a/doc/source/building/build_in_buildservice.rst
+++ b/doc/source/building/build_in_buildservice.rst
@@ -85,13 +85,14 @@ The notable differences to running KIWI locally include:
      experience when using OBS.
 
 * OBS will by default only build a single image type. If your appliance
-  contains the multiple build types or uses profiles, use the `multibuild
-  <https://openbuildservice.org/help/manuals/obs-reference-guide/cha.obs.multibuild.html>`_ feature.
+  contains multiple build types or uses profiles, use the `multibuild
+  <https://openbuildservice.org/help/manuals/obs-reference-guide/cha.obs.multibuild.html>`_
+  feature.
 
-* Subfolders in OBS projects are ignored by default by :file:`osc` and must
-  be explicitly added via `osc add $FOLDER` [#f2]_. Bear that in mind when
-  adding the overlay files inside the :file:`root/` directory to your
-  project.
+* Subfolders in OBS projects are ignored by default by :command:`osc` and
+  must be explicitly added via :command:`osc add $FOLDER` [#f2]_. Bear that
+  in mind when adding the overlay files inside the :file:`root/` directory
+  to your project.
 
 
 .. _obs-recommended-settings:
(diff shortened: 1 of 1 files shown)


doc/source/building/build_in_buildservice.rst
```diff
@@ -167,31 +173,57 @@ your dependent packages. These repositories can be provided in two ways:
...
+
+   You will be unable to build images for different distributions from the
+   same OBS project when adding repositories to your project's
+   configuration. The problem is, that all your image builds share the same
+   repositories. This will result in dependency conflicts for different
+   distributions.
+
+   On the other hand this approach requires a lot less workarounds in the
+   project configuration as compared to adding the repositories via the
+   :file:`config.xml`.
```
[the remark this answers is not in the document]
dcermak: I've formulated it differently but tried to follow the ANSI norm. The "issue" with your formulation was imho that it sounds like you shouldn't add the repositories to the project configuration. But that is actually the opposite of what we want: you **should** put them there, as it makes working with OBS a ton easier. It is just something one needs to be aware of.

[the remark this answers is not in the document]
tomschr: Oh, I see! I just created that example to give you an idea about it. Great that you could apply it to this warning. :+1: Thanks Dan! :-)

doc/source/building/build_in_buildservice.rst
```diff
@@ -167,31 +173,60 @@ your dependent packages. These repositories can be provided in two ways:
...
+          <arch>x86_64</arch>
+        </repository>
+      </project>
+
+   Now `Virtualization:Appliances:Builder` is the last repository, which'
+   repositories are included into the search path. As
+   `openSUSE:Factory/snapshot` is among these, it can be omitted from the
+   repository list.
+
+#. Keep the repositories in your :file:`config.xml` configuration file. If
```
davidcassany: Probably i would add  note here with a link to some examples https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_x86 I found them explanatory and there you can also see the _painful_ prjconf configuration.

dcermak: That repository is already mentioned at the bottom of the same file in the section Project configuration. I could add a link to that section?

davidcassany: mmm I think this is not the same repository. Note that this is the repository for KIWI testing images, where we do not use `obsrepositories:/` trick so we are capable to build several distro in the same project at a cost of a hard to determine prjconf configuration.

dcermak: Ah, sorry, I was thinking you are referring to a different section. 

So the thing is, I am actually recommending here (as per @mlschroe's recommendation) to keep your normal repositories in `config.xml` and to put the kiwi repository into OBS' config. That will make it impossible to build multiple distros from the same project, too, but it should make working with OBS a lot easier (and it shouldn't require all these nasty `Prefer:`/`Substitute:`/etc. in the project config). Then referring to https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_x86 wouldn't be too helpful, as there these workarounds are unfortunately necessary, as the kiwi repo is in `config.xml` and not in the obs repository configuration.

PR #1096, opened 2019-05-22
Disable check-valid-until with repository_gpgcheck

davidcassany:
This commit is two fold:

* From one side fixes a wrong use of the `trusted` option for
  apt repositories. `trusted=no` does not force to run the gpg checks
  it just forces the repository to be considered untrusted regardless
  the result of the security checks.

* From the other side it disables the option `check-valid-until` in
  case gpg checks are disabled using the `repository_gpgcheck`. It
  works at repository level. This enables using unmaintained or
  expired repositories for the build.

Fixes #1028

### DIFF
--- a/kiwi/repository/apt.py
+++ b/kiwi/repository/apt.py
@@ -159,12 +159,10 @@ class RepositoryApt(RepositoryBase):
         if not components:
             components = 'main'
         with open(list_file, 'w') as repo:
-            if repo_gpgcheck is None:
-                repo_line = 'deb {0}'.format(uri)
+            if repo_gpgcheck is False:
+                repo_line = 'deb [trusted=yes check-valid-until=no] {0}'.format(uri)
             else:
-                repo_line = 'deb [trusted={0}] {1}'.format(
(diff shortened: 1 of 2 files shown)


kiwi/repository/apt.py
```diff
@@ -159,12 +159,10 @@ def add_repo(
         if not components:
             components = 'main'
         with open(list_file, 'w') as repo:
-            if repo_gpgcheck is None:
-                repo_line = 'deb {0}'.format(uri)
+            if repo_gpgcheck is False:
+                repo_line = 'deb [trusted=yes][check-valid-until=no] {0}'.format(uri)
```
Conan-Kudo: This is the wrong way to do this.

This is the way to do it:
```
deb [trusted=yes check-valid-until=no] {0}
```

Conan-Kudo: Reference: [`sources.list(5)`](https://manpages.debian.org/unstable/apt/sources.list.5.en.html#THE_DEB_AND_DEB-SRC_TYPES:_GENERAL_FORMAT)

PR #1097, opened 2019-05-27
Update documentation using prolog variables

schaefi:
Instead of hard coded names for example images and their
versions a prolog file including placeholders has been
added.


schaefi:
@tomschr I realized it is not possible to parse references inside of code:: or code-block:: expressions. Thus with this patch we are loosing syntax highlighting at the changed places. However I could live with that limitation if we create consistent names across the entire document.

Of course if you know a solution such that we can keep the highlighting as before it would be great.

Thanks much

tomschr:
@schaefi 
> I realized it is not possible to parse references inside of code:: or code-block:: expressions. 

Yes, it seems so. :cry: 

> Thus with this patch we are loosing syntax highlighting at the changed places. However I could live with that limitation if we create consistent names across the entire document.

I'm a big fan of consistency! :wink: So from that perspective, I like this patch. Although we can't use syntax highlighting anymore which is sad.

> Of course if you know a solution such that we can keep the highlighting as before it would be great.

Well, I have an idea, but I don't know if this is feasible. After some digging, I've found some interesting links:

Basically, you add a [source-read event handler](https://github.com/sphinx-doc/sphinx/issues/4054#issuecomment-329097229) into the `setup(app)` function of `config.py` (see the example on the referenced link). That looks quite promising as it replaces things before Sphinx does the rendering.

A more elaborate example can be found in the [sphinx-jinja](https://pypi.org/project/sphinx-jinja/) extension. I've got that when I've searched the [Sphinx user group](https://groups.google.com/forum/#!forum/sphinx-users).

The only problem I see is how to get the relationship right between the replace text and the placeholder inside the code.

Hope that helps.

tomschr:
After a second thought, I could offer you another alternative, without any coding effort. You may or may not like it. :wink: 

You could use a "virtual placeholder": It's a capitalized name embedded into the code, but not really replaced. You use the "real" text outside of the code. I have something like this in my mind:

```rst
.. code:: bash

    $ cd <git-clone-result-kiwi-descriptions>

    $ dice build suse/x86_64/DESCRIPTION
    $ dice status suse/x86_64/DESCRIPTION

Replace ``DESCRIPTION`` with your image description, for example |exc_description|.
```

I know, it's not perfect. However, you don't need to adapt or write any Sphinx extension. Plus you can use the syntax highlighting as before.

There is another, more subtle advantage: It forces the user to read and understand the text. Any thoughtless copy-and-paste actions would probably not be successful. 
Additionally, this kind of code "protects" the reader from doing stupid copy-and-paste actions: the reader are forced to read and understand the text (well, theoretically). If this is considered a good or a bad thing is up to you. :smile:

schaefi:
> About the highlighting issue I am not very concerned about it if it is still rendered in mono-spaced fonts

yes that's safe. I build the docs and looked at it before this PR. So in a code block the surrounding box is colored and the text inside is highlighted. In parsed-literal the surrounding box is grey and the text is like any other text but monospace font, cut&paste from it works. So it's ok but lost its beauty

schaefi:
> You could use a "virtual placeholder"

we use this concept when we have no clear reference. However with the examples we have a clear name references that matches the image description we provide along with the documentation. I agree we could make this less explicit by saying "use A and A can be "bob". To be honest it makes it more complex to run through the example. Usually I expect people who just build from our example descriptions to use cut&paste. I like docs that offers things in that way and we would loose this if we add virtual placeholders.

schaefi:
> Basically, you add a source-read event handler

This would be in my eyes a good solution to keep both, consistency and the highlighting as before

schaefi:
Ok I moved to the event handler model. That was a very good idea. Thanks @tomschr

schaefi:
found a lot more inconsistencies and fixed them

tomschr:
@schaefi Just a short, friendly question: is the `{placeholder}` syntax safe enough?

I'm a bit worried that it could be mistaken by some Shell expression (which would be valid too and looks similar). :wink: Would it be better to just use double curly braces (`{{placeholder}}`) in case you need to distinguish it?

schaefi:
I think it's pretty safe. The prologReplace method does

```python
result.replace(key, app.config.prolog_replacements[key]
```

So only an exact match of e.g '{exc_image_base_name}' will lead to a replacement and I can use any other characters as needed. The replacement happens before the sphinx validation. This means highligthers and syntax checkers e.g xml snippets are all happy with that.

If for whatever reason another key match is needed we can use whatever is needed in the prolog_replacements hash.

There is one disadvantage. If the replace did not apply the origin version stays and this is nowhere seen as an error. So only proofreading will outline this part. I can live with that because I think inconsistencies will be overlooked more often

Thoughts ?

tomschr:
Ahh, right! Sure, in that case we can really consider it as safe. That's brilliant! Forget my stupid comment. All fine, I'm happy. :+1:

schaefi:
It was your idea :+1: :-)

Let's merge it. Thanks much

PR #1101, opened 2019-06-03
Extend spare partition setup

schaefi:
The spare partition could be used to introduce one additional
partition table entry. With this patch the following new type
attributes will be added:

* spare_part_fs="fsname"
* spare_part_mountpoint="/location"
* spare_part_is_last="true|false"

Along with the setup of the partition size the filesystem and
its mountpoint can be specified. If set the contents of the
rootfs at the specified spare location will be synced to that
partition. The spare_part_is_last attribute will place the
spare partition at the end of the disk. Note this attribute
is only available for the simple vmx disk type. This is
related to bsc#1129566

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -73,6 +73,10 @@ class DiskBuilder(object):
         self.target_dir = target_dir
         self.xml_state = xml_state
         self.spare_part_mbsize = xml_state.get_build_type_spare_part_size()
+        self.spare_part_fs = xml_state.build_type.get_spare_part_fs()
+        self.spare_part_is_last = xml_state.build_type.get_spare_part_is_last()
+        self.spare_part_mountpoint = \
+            xml_state.build_type.get_spare_part_mountpoint()
         self.persistency_type = xml_state.build_type.get_devicepersistency()
         self.root_filesystem_is_overlay = xml_state.build_type.get_overlayroot()
         self.custom_root_mount_args = xml_state.get_fs_mount_option_list()
@@ -145,6 +149,10 @@ class DiskBuilder(object):
         # representing the boot/efi area of the disk
         self.system_efi = None
 
+        # an instance of a class with the sync_data capability
+        # representing the spare_part_mountpoint area of the disk
+        self.system_spare = None
+
         # result store
         self.result = Result(xml_state)
         self.runtime_config = RuntimeConfig()
@@ -186,7 +194,7 @@ class DiskBuilder(object):
 
         return result
 
-    def create_disk(self):
+    def create_disk(self):  # noqa: C901
         """
         Build a bootable raw disk image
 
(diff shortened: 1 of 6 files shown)


schaefi:
An example type setup would look like this:

```xml
<type image="vmx" ... spare_part="200M" spare_part_mountpoint="/var" spare_part_fs="ext4" spare_part_is_last="true"/>
```

kiwi/builder/disk.py
```diff
@@ -653,24 +698,37 @@ def _build_and_map_disk_partitions(self):               # noqa: C901
...
+            ) + Defaults.get_min_partition_mbytes()
             self.disk.create_root_readonly_partition(
-                int(squashed_rootfs_mbsize + 50)
+                squashed_rootfs_mbsize
             )
+            disksize_used_mbytes += squashed_rootfs_mbsize
+
+        if self.spare_part_mbsize and self.spare_part_is_last:
+            rootfs_mbsize = disksize_mbytes - disksize_used_mbytes - \
+                self.spare_part_mbsize - Defaults.get_min_partition_mbytes()
```
davidcassany: Removing the `Defaults.get_min_partition_mbytes()` in this count is required? Looks like this could cause the spare_part size mismatch, even the mismatch I see has a different value...

schaefi: > There is no validation that the size of the spare partition is big enough to hold data of the mount point.

That applies to the volume setup as well if you give it a fixed size. At the sync level people will recognize this with a "no space left on device" error at build time. It would be very hard to pre-check this and in addition in depends on the filesystem if such a pre-size check can be successful. Thus we allow people to set the size and see if it fits

> Removing the Defaults.get_min_partition_mbytes() in this count is required?

This is the case when the spare_part is the last one. This leads to a reduction of the now required size calculation of the root partition. I intentionally substract the min partition size constraint to reach a minimum size available for the rest and to have a way to calculate with the inaccuracy between data, fs-overhead, partition-boundaries and such.

You are right the number value for the spare_part is not 100% accurate if spare_part is last. I could live with that. Thoughts ?

davidcassany: > That applies to the volume setup as well if you give it a fixed size. At the sync level people will recognize this with a "no space left on device" error at build time. It would be very hard to pre-check this and in addition in depends on the filesystem

Well yes, you got a point here, probably we can't relay on making an accurate estimation of the minimal size required.

> You are right the number value for the spare_part is not 100% accurate if spare_part is last. I could live with that. Thoughts ?

I can certainly live with that, I was just wondering why the mismatch does not match the minimal partition default value.

All right then I think all is green here :+1:

schaefi: > Well yes, you got a point here, probably we can't relay on making an accurate estimation

I also did some tests with small and big spare_part's and I think you have a point here. I wanted to be at the very safe side with the size calculation. But I think we can be a bit less conservative. I've reduced the min size overhead which now leads to more accurate values.

spare_part="100M" ~ 102M
spare_part="5G" ~ 4984M

Let me know what your tests say

Thanks

davidcassany: Yes, my test are also giving me way more accurate results now. :+1:

spare_part="200M" ~ 199M
spare_part="3G" ~ 2969Mb

PR #1104, opened 2019-06-11
Make kiwi installable in a venv

### LINKED ISSUE #1022: Make KIWI python package installable for non root users
## Problem description

KIWI can't be installed from `pip` without being root due too the C tools installation and its makefile trying to install/clean in locations that are outside of the user space or the virtual environment (.e.g it fails trying to clean `/usr/bin/dcounter`). The installation fails when the python script is calling `make tools install`.

## Expected behaviour

KIWI module can be installed in a python virtual environment without requiring root permissions during the installation.

As a regular user run:

```bash
mkdir ~/tmp
cd ~/tmp
python3 -m venv virtualenv
pip install kiwi
```

davidcassany:
This commit sets the buildroot to the venv `sys.prefix`. This way
all the binaries are installed in a path under the venv.

Fixes #1022

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ from distutils.command import clean as distutils_clean
 import distutils
 import subprocess
 import os
+import sys
 
 import platform
 
@@ -18,6 +19,11 @@ from kiwi.version import __version__
 
 
 python_version = platform.python_version().split('.')[0]
+
+# sys.base_prefix points to the installation prefix set during python
+# compilation and sys.prefix points to the same path unless we are inside
+# a venv, in which case points to the $VIRTUAL_ENV value.
+is_venv = sys.base_prefix != sys.prefix if sys.version_info >= (3, 3) else False
 
 
 class sdist(setuptools_sdist.sdist):
@@ -151,6 +157,8 @@ class install(distutils_install.install):
         command = ['make']
         if self.root:
             command.append('buildroot={0}/'.format(self.root))
+        elif is_venv:
+            command.append('buildroot={0}/'.format(sys.prefix))
         command.append('python_version={0}'.format(python_version))
         command.append('tools')
         command.append('install')


setup.py
```diff
@@ -19,6 +20,9 @@
 
 python_version = platform.python_version().split('.')[0]
 
+is_venv = (hasattr(sys, 'real_prefix') or\
+            (hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix))
+
```
schaefi: I believe this works but it's a strange read. Actually entering the vend sets an environment variable:

```
$VIRTUAL_ENV
```

That is only present if you have activated the virtual environment. I saw reports that deactivating the venv still leaves that variable in the environment but that is not true from my tests. Thus I expect this to have been fixed in python. As we do care less on py2 I think checking for the environment variable might be an easier read, unless you have objections why that should be avoided ?

davidcassany: I have not coded it like that because I also read similar comments about not doing it as you did. I was also considering not including py2 into the equation. In fact the check on `real_prefix` is doen for `virtualenv` module from python2 and the other check `base_prefix` is done for `venv` module from python3. 

So if we do not care about python2 we could remove the `real_prefix` verification and then we would be using the appropriate check then. I verified this `base_prefix` check on the documentation [here](https://docs.python.org/3/library/sys.html#sys.base_prefix). Probably the  `hasattr(sys, 'base_prefix')` part can also be dropped as this is for python3 < 3.3 (the oldest python3 maintained upstream is 3.5).

So if we align with python3 upstream having

```python
is_venv = sys.base_prefix != sys.prefix
```
should be enough. I am fine reducing it to the above. I could also add a comment stating that `sys.prefix` points to the venv path, if present, and that `sys.base_prefix` points always to the prefix used during python compilation.

schaefi: I think this would be good :+1:

PR #1106, opened 2019-06-14
Provider better error message in prepare phase

schaefi:
In case of an error the real valuable information is often
somewhere in the package manager output that does not get
exposed to the users console by default. The error message
we provide should tell users how they can get further details
such that they know how to find the real cause of the problem.
This patch enhances the error message in that regard.


schaefi:
good point, updated now

PR #1113, opened 2019-06-26
Refactor ISO boot, use grub in EFI mode

### LINKED ISSUE #1092: Use grub instead of isolinux for non EFI ISO boot
## Problem description
Trying to build an ISO with `image="iso"` for aarch64 fails with the following error:
```
[ 1275s] [ INFO    ]: 13:46:58 | Setting up isolinux bootloader configuration
[ 1275s] [ ERROR   ]: 13:46:58 | KiwiBootLoaderIsoLinuxPlatformError: host architecture aarch64 not supported for isolinux setup
```

## Expected behaviour
We should be able to build ISO for non-x86.
`isolinux` is not available for non-x86, so we should be able to use `grub2` instead.

Build an image for non-x86 (e.g. aarch64)

+ KIWI version: 
+ Operating system: 
+ OBS version:

schaefi:
Before this commit isolinux was used to boot an ISO and thus forms a hard requirement. isolinux. However only exists for the x86 platform. This limitation did not  allow to create ISO images for other platforms. With this commit a refactoring of the ISO boot setup is introduced.
    
* isolinux is only used if the bios firmware is requested and the platform matches the x86 architecture. A runtime check will check for this condition and exits early if not applicable
    
* in case of the EFI firmware we already used grub in EFI mode but still had isolinux in place for the legacy/CSM boot. That part is now also replaced by a platform specific grub eltorito image and grub's boot_hybrid.img for hybrid boot. On platforms that do not provide those modules the support for it will be skipped
    
With this change in place it's possible to control the ISO boot layout through the firmware setup and all platform specific modules are handled as such. Therefore we also deleted the syslinux requirement.

This Fixes #1092

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -602,7 +602,8 @@ class Defaults(object):
         signed_grub_file_patterns = [
             '/usr/share/efi/*/grub.efi',
             '/usr/lib64/efi/grub.efi',
-            '/boot/efi/EFI/*/grub*.efi'
+            '/boot/efi/EFI/*/grub*.efi',
+            '/usr/share/grub*/arm64-efi/grub.efi'
         ]
         for signed_grub_pattern in signed_grub_file_patterns:
             for signed_grub in glob.iglob(root_path + signed_grub_pattern):
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -586,6 +586,57 @@ class Defaults(object):
                 return bios_grub_core
 
     @staticmethod
+    def get_syslinux_modules():
+        """
+        Returns list of syslinux modules to include on ISO
+        images that boots via isolinux
+
+        :return: base file names
+
+        :rtype: list
+        """
+        return [
+            'isolinux.bin',
+            'ldlinux.c32',
+            'libcom32.c32',
+            'libutil.c32',
+            'gfxboot.c32',
+            'gfxboot.com',
+            'menu.c32',
+            'chain.c32',
+            'mboot.c32'
+        ]
+
+    @staticmethod
+    def get_syslinux_search_paths():
+        """
+        syslinux is packaged differently between distributions.
+        This method returns a list of directories to search for
+        syslinux data
+
+        :return: directory names
+
+        :rtype: list
+        """
+        return [
+            '/usr/share/syslinux',
+            '/usr/lib/syslinux/modules/bios',
+            '/usr/lib/ISOLINUX'
+        ]
+
+    @staticmethod
+    def get_isolinux_bios_grub_loader():
+        """
+        Return name of eltorito grub image used as isolinux loader
+        in BIOS mode if isolinux.bin should not be used
+
+        :return: file base name
+
+        :rtype: str
+        """
+        return 'eltorito.img'
+
+    @staticmethod
     def get_signed_grub_loader(root_path):
         """
         Provides shim signed grub loader file path
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -82,6 +82,21 @@ class Defaults(object):
         ]
 
     @staticmethod
+    def is_x86_arch(arch):
+        """
+        Checks if machine architecture is x86 based
+
+        Any arch that matches 32bit and 64bit x86 architecture
+        causes the method to return True. Anything else will
+        cause the method to return False
+
+        :rtype: bool
+        """
+        if arch == 'x86_64' or arch == 'i686' or arch == 'i586':
+            return True
+        return False
+
+    @staticmethod
     def is_buildservice_worker():
         """
         Checks if build host is an open buildservice machine
@@ -543,6 +558,34 @@ class Defaults(object):
                 return unsigned_grub_file
 
     @staticmethod
+    def get_grub_bios_core_loader(root_path):
+        """
+        Provides grub bios image
+
+        Searches distribution specific locations to find the
+        core bios image below the given root path
+
+        :param string root_path: image root path
+
+        :return: file path or None
+
+        :rtype: str
+        """
+        bios_grub_core_patterns = [
+            '/usr/share/grub*/i386-pc/{0}'.format(
+                Defaults.get_bios_image_name()
+            ),
+            '/usr/lib/grub*/i386-pc/{0}'.format(
+                Defaults.get_bios_image_name()
+            )
+        ]
+        for bios_grub_core_pattern in bios_grub_core_patterns:
+            for bios_grub_core in glob.iglob(
+                root_path + bios_grub_core_pattern
+            ):
+                return bios_grub_core
+
+    @staticmethod
     def get_signed_grub_loader(root_path):
         """
         Provides shim signed grub loader file path
@@ -828,6 +871,17 @@ class Defaults(object):
             return default_module_directory_names[arch]
 
     @staticmethod
+    def get_bios_module_directory_name():
+        """
+        Provides x86 BIOS directory name which stores the pc binaries
+
+        :return: directory name
+
+        :rtype: str
+        """
+        return 'i386-pc'
+
+    @staticmethod
     def get_efi_image_name(arch):
         """
         Provides architecture specific EFI boot binary name
@@ -849,6 +903,17 @@ class Defaults(object):
         }
         if arch in default_efi_image_names:
             return default_efi_image_names[arch]
+
+    @staticmethod
+    def get_bios_image_name():
+        """
+        Provides bios core boot binary name
+
+        :return: name
+
+        :rtype: str
+        """
+        return 'core.img'
 
     @staticmethod
     def get_default_boot_timeout_seconds():
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -603,7 +603,7 @@ class Defaults(object):
             '/usr/share/efi/*/grub.efi',
             '/usr/lib64/efi/grub.efi',
             '/boot/efi/EFI/*/grub*.efi',
-            '/usr/share/grub*/arm64-efi/grub.efi'
+            '/usr/share/grub*/*-efi/grub.efi'
         ]
         for signed_grub_pattern in signed_grub_file_patterns:
             for signed_grub in glob.iglob(root_path + signed_grub_pattern):
--- a/kiwi/iso_tools/xorriso.py
+++ b/kiwi/iso_tools/xorriso.py
@@ -92,7 +92,12 @@ class IsoToolsXorrIso(IsoToolsBase):
 
         if Defaults.is_x86_arch(self.arch):
             if efi_mode:
-                loader_file = self.boot_path + '/loader/eltorito.img'
+                loader_file = os.sep.join(
+                    [
+                        self.boot_path, 'loader',
+                        Defaults.get_isolinux_bios_grub_loader()
+                    ]
+                )
                 mbr_file = os.sep.join(
                     [self.source_dir, self.boot_path, '/loader/boot_hybrid.img']
                 )
@@ -103,14 +108,16 @@ class IsoToolsXorrIso(IsoToolsBase):
                 ]
             else:
                 loader_file = self.boot_path + '/loader/isolinux.bin'
-                syslinux_lookup_paths = [
-                    '/usr/share/syslinux', '/usr/lib/syslinux/modules/bios',
-                    '/usr/lib/ISOLINUX'
-                ]
-                mbr_file = Path.which('isohdpfx.bin', syslinux_lookup_paths)
+                mbr_file = Path.which(
+                    'isohdpfx.bin', Defaults.get_syslinux_search_paths()
+                )
                 self.iso_loaders += [
-                    '-boot_image', 'isolinux', 'bin_path={0}'.format(loader_file),
-                    '-boot_image', 'isolinux', 'system_area={0}'.format(mbr_file),
+                    '-boot_image', 'isolinux', 'bin_path={0}'.format(
+                        loader_file
+                    ),
+                    '-boot_image', 'isolinux', 'system_area={0}'.format(
+                        mbr_file
+                    ),
                     '-boot_image', 'isolinux', 'partition_table=on',
                 ]
 
--- a/kiwi/iso_tools/xorriso.py
+++ b/kiwi/iso_tools/xorriso.py
@@ -22,6 +22,7 @@ from kiwi.iso_tools.base import IsoToolsBase
 from kiwi.path import Path
 from kiwi.command import Command
 from kiwi.exceptions import KiwiIsoToolError
+from kiwi.defaults import Defaults
 
 
 class IsoToolsXorrIso(IsoToolsBase):
@@ -64,7 +65,10 @@ class IsoToolsXorrIso(IsoToolsBase):
 
         :param list custom_args: custom ISO meta data
         """
+        efi_mode = False
         if custom_args:
+            if custom_args.get('efi_mode'):
+                efi_mode = True
             if 'mbr_id' in custom_args:
                 self.iso_parameters += [
                     '-application_id', custom_args['mbr_id']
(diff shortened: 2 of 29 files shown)


schaefi:
Along with this change I build a new package in the Staging are and created an arm integration test build for a live iso here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_arm/test-image-iso

Yay :smile:

ggardet:
> Along with this change I build a new package in the Staging area to make sure nothing breaks and also created an arm integration test build for a live iso here:
> 
> * https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_arm/test-image-iso
> 
> Yay 😄
> 
> The test build will be moved to the git as soon as we have merged this one

See my test result here: https://github.com/SUSE/kiwi/issues/1092#issuecomment-505796730

Conan-Kudo:
@schaefi Can you please add some Fedora AArch64 tests using the `uefi` mode? Unlike openSUSE, we're using `shim` for AArch64 Secure Boot, and I'd like to make sure that works before approving this PR. :)

schaefi:
@Conan-Kudo I haven't tested the secure boot path, will do that next

kiwi/defaults.py
```diff
@@ -602,7 +602,8 @@ def get_signed_grub_loader(root_path):
         signed_grub_file_patterns = [
             '/usr/share/efi/*/grub.efi',
             '/usr/lib64/efi/grub.efi',
-            '/boot/efi/EFI/*/grub*.efi'
+            '/boot/efi/EFI/*/grub*.efi',
+            '/usr/share/grub*/arm64-efi/grub.efi'
```
Conan-Kudo: This should probably be `/usr/share/grub*/*-efi/grub*.efi` There's probably going to be other architectures that will get this treatment...

ggardet:
@schaefi tell me if you build new images to test.

kiwi/defaults.py
schaefi: yes fully agree, done now. The integration test for secure boot on suse aarch64 has completed here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_arm/test-image-secure-boot-iso

Now we just have to make Fedora build. Just in case I couldn't get help from obs, do you have some aarch64 machine running with Fedora somewhere ? If not I'll check if I can run an a1 instance in AWS with Fedora. Not sure if there is an offering available

schaefi:
> @schaefi tell me if you build new images to test.

yep present now

ggardet:
> > @schaefi tell me if you build new images to test.
> 
> yep present now
@schaefi 
Hmm, something went wrong. The ISO is not bootable at all. The `grub.efi` file should be installed as `bootaa64.efi` for aarch64, not `gub.efi`.

schaefi:
Fixed it, a new image build will start soon

ggardet:
> Fixed it, a new image build will start soon

@schaefi 
It does boot properly in a secureboot environment! \o/ 

Regarding the "Boot from hard disk" entry, here is my `parted -l` output on a regular aarch64 system with openSUSE Tumbleweed installed:
```
Model: ATA ST1000DM003-1SB1 (scsi)
Disk /dev/sda: 1000GB
Sector size (logical/physical): 512B/4096B
Partition Table: gpt
Disk Flags: 

Number  Start   End     Size    File system     Name    Flags
 1      1049kB  211MB   210MB   fat16           UEFI    boot, esp
 2      212MB   484MB   273MB   ext3            lxboot  msftdata
 3      485MB   43.4GB  42.9GB  btrfs
 4      43.4GB  992GB   948GB   xfs
 5      992GB   1000GB  8347MB  linux-swap(v1)          swap
```
So, there are a `boot` flag and a `UEFI` label. (Ignore partition 2, it is a leftover)

schaefi:
> It does boot properly in a secureboot environment! \o/

yay :) thanks for testing. I see what I can do with the label thing

schaefi:
@ggardet I changed the way how we search for the "boot from harddisk" system. Still testing

schaefi:
Hmm, ok I had this idea to search for the installed OS

```patch
--- a/kiwi/bootloader/template/grub2.py
+++ b/kiwi/bootloader/template/grub2.py
@@ -300,7 +300,7 @@ class BootLoaderTemplateGrub2(object):
 
         self.menu_iso_harddisk_entry = dedent('''
             menuentry "Boot from Hard Disk" --class os --unrestricted {
-                search --set=root --label EFI
+                search --set=root --file /EFI/BOOT/${efi_image_name}
                 chainloader ($${root})/EFI/BOOT/${efi_image_name}
             }
         ''').strip() + os.linesep
```

Problem here is that this will also find the live iso itself if it would be first in the search list.
Probably not idea. But the label search is only safe if the deployment happened through a kiwi image.
So if you don't mind let's discuss this in a new issue. It's not low hanging fruit

ggardet:
> Hmm, ok I had this idea to search for the installed OS
> 
> ```diff
> --- a/kiwi/bootloader/template/grub2.py
> +++ b/kiwi/bootloader/template/grub2.py
> @@ -300,7 +300,7 @@ class BootLoaderTemplateGrub2(object):
>  
>          self.menu_iso_harddisk_entry = dedent('''
>              menuentry "Boot from Hard Disk" --class os --unrestricted {
> -                search --set=root --label EFI
> +                search --set=root --file /EFI/BOOT/${efi_image_name}
>                  chainloader ($${root})/EFI/BOOT/${efi_image_name}
>              }
>          ''').strip() + os.linesep
> ```
> 
> Problem here is that this will also find the live iso itself if it would be first in the search list.
> Probably not idea. But the label search is only safe if the deployment happened through a kiwi image.
> So if you don't mind let's discuss this in a new issue. It's not low hanging fruit

Maybe there is a way to `exit` and let UEFI firmware trying next entry?
Or just use `UEFI` label instead of `EFI`?
But, we can discuss that in a new separate issue.

schaefi:
@Conan-Kudo here is the test image you wanted for testing:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_arm/test-image-fedora-iso

schaefi:
The Fedora image is using shim because it exists there

```
[  322s] [ WARNING ]: 15:57:32 | --> Running fallback setup for shim secure boot efi image
[  322s] [ DEBUG   ]: 15:57:32 | EXEC: [cp /usr/src/packages/KIWI-iso/build/image-root/boot/efi/EFI/fedora/shim.efi /usr/src/packages/KIWI-iso/live-media.g5q0564l/EFI/BOOT/bootaa64.efi]
[  322s] [ DEBUG   ]: 15:57:32 | EXEC: [cp /usr/src/packages/KIWI-iso/build/image-root/boot/efi/EFI/fedora/grubaa64.efi /usr/src/packages/KIWI-iso/live-media.g5q0564l/EFI/BOOT]
```

This looks good to me, let's hope this is correct

schaefi:
Hey guys we are getting closer, quite some improvement this week here. I'm eager to get that merged :)

ggardet:
> Hey guys we are getting closer, quite some improvement this week here. I'm eager to get that merged :)

Yes, I am looking forward to get this in Factory:ARM! :)

I tested the Fedora image and `bootaa64.efi` and `BOOTAA64.EFI` (why 2?) are unsigned and cannot be used in a secure env. It looks like a problem with `shim.efi` in Fedora 28 for ARM which zould be unsigned?
`fbaa64.efi` and `grubaa64.efi` are signed.

schaefi:
> @schaefi Maybe you could build against Fedora 29 instead of 28?

I changed the build to run for Fedora29

ggardet:
> > @schaefi Maybe you could build against Fedora 29 instead of 28?
> 
> I changed the build to run for Fedora29

It fails due to missing installed packages, apparently.

schaefi:
That will self heal in a few minutes :) The package build of python-kiwi in Staging for Fedora 29 is just not yet done... give it a few minutes

schaefi:
@ggardet @Conan-Kudo Hmm, obs cannot build for Fedora29. This is odd.

```
<path project="Fedora:29" repository="update" url="http://download.opensuse.org/repositories/Fedora:/29/update/" />
  <path project="Fedora:29" repository="standard" url="http://download.opensuse.org/repositories/Fedora:/29/standard/" />
```

This is what obs tells me. But when I look at that url I there is no such distro.

ggardet:
> <path project="Fedora:29" repository="update" url="http://download.opensuse.org/repositories/Fedora:/29/update/" />
>   <path project="Fedora:29" repository="standard" url="http://download.opensuse.org/repositories/Fedora:/29/standard/" />

@adrianschroeter Any idea what is going on here? OBS can build for Fedora 28, but not for 29, apparently.

schaefi:
@ggardet I use the link you pasted to the fixed shim package. That should also install on Fedora28, so I just added a quick and dirty rpm update call in config.sh with some local packages from Fedora29. Not nice but I don't think we will get that resolved in obs quickly

ggardet:
> @ggardet I use the link you pasted to the fixed shim package. That should also install on Fedora28, so I just added a quick and dirty rpm update call in config.sh with some local packages from Fedora29. Not nice but I don't think we will get that resolved in obs quickly

`bootaa64.efi` is not signed, but `BOOTAA64.EFI` is signed. I think we should not have both, but only one.

Anyway, I tried to boot it with qemu with Microsoft key, and it does not boot as signature verification fails. I will let @Conan-Kudo to debug this further.

schaefi:
ok thanks much for testing. I think that's exactly the point where @Conan-Kudo should give an advise how we should continue here. I'll keep the PR open, maybe it's an easy solution we can fix here. I would also be open to merge the current state and have a followup fix for the secure boot setup on Fedora and kiwi's fallback code. So let's wait for feedback from Neal first

adrianschroeter:
@schaefi can you show me any example what did fail with Fedora 29?

Conan-Kudo:
@schaefi Is this also a problem if you go up to Fedora 30?

schaefi:
@adrianschroeter @Conan-Kudo 
I have switched the build to Fedora 30

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_arm/test-image-fedora-iso

Let's see what it says

Conan-Kudo:
@schaefi The inclusion of the `shim` package (which is formerly known as `shim-unsigned`) is confusing things. You're including two different versions of the files...

schaefi:
@Conan-Kudo ok I changed that to only add 

```xml
<package name="shim-signed-aa64"/>
```

That is correct right ?

But we have another problem here. I think obs is in some way broken for building Fedora30

```
[   87s] /.build/build-vm: line 1049: hostname: command not found
[   89s] -----------------------------------------------------------------
[   89s] ----- building appliance.kiwi
[   89s] -----------------------------------------------------------------
[   89s] -----------------------------------------------------------------
[   89s] creating repodata for Fedora:30/standard
[   89s] /bin/createrepo: /usr/share/createrepo/genpkgmetadata.py: /usr/bin/python: bad interpreter: No such file or directory
[   89s] /bin/createrepo: line 2: /usr/share/createrepo/genpkgmetadata.py: Success
[   89s] creating repodata for Fedora:30/update
[   89s] /bin/createrepo: /usr/share/createrepo/genpkgmetadata.py: /usr/bin/python: bad interpreter: No such file or directory
[   89s] /bin/createrepo: line 2: /usr/share/createrepo/genpkgmetadata.py: Success
[   89s] creating repodata for Virtualization:Appliances:Staging/Fedora_30
[   89s] /bin/createrepo: /usr/share/createrepo/genpkgmetadata.py: /usr/bin/python: bad interpreter: No such file or directory
[   89s] /bin/createrepo: line 2: /usr/share/createrepo/genpkgmetadata.py: Success
```

This in the end leads to dnf not being able to find any package which is not a surprise

@adrianschroeter you can check the buildlog yourself. Something is wrong when the build script setup the local repos

Thanks for any help we are blocked

Conan-Kudo:
@schaefi can you add `Substitute: createrepo createrepo_c`?

schaefi:
ok, just  a sec

schaefi:
done

schaefi:
@Conan-Kudo problem is still the same:

```
[  145s] /bin/createrepo: /usr/share/createrepo/genpkgmetadata.py: /usr/bin/python: bad interpreter: No such file or directory
```

strange

schaefi:
```
osc buildinfo images aarch64 | grep createrepo
  <bdep name="createrepo_c" notmeta="1" version="0.14.1" release="2.fc30" arch="aarch64" project="Fedora:30" repository="update" />
  <bdep name="createrepo_c-libs" notmeta="1" version="0.14.1" release="2.fc30" arch="aarch64" project="Fedora:30" repository="update" />
```

This looks good though

????

Conan-Kudo:
@schaefi I have no idea why OBS was forcibly pulling in `createrepo`, but I'm hoping that will fix that.

schaefi:
I need to call Adrian, I don't understand anything what's going on here...

kiwi/iso_tools/xorriso.py
```diff
@@ -85,23 +89,33 @@ def init_iso_creation_parameters(self, custom_args=None):
...
+                    [self.source_dir, self.boot_path, '/loader/boot_hybrid.img']
+                )
+                self.iso_loaders += [
+                    '-boot_image', 'grub', 'bin_path={0}'.format(loader_file),
+                    '-boot_image', 'grub', 'grub2_mbr={0}'.format(mbr_file),
+                    '-boot_image', 'grub', 'grub2_boot_info=on'
+                ]
+            else:
+                loader_file = self.boot_path + '/loader/isolinux.bin'
+                syslinux_lookup_paths = [
```
davidcassany: shouldn't we use here the list `syslinux_dirs` that is also defined in `IsoToolsBase`. Probably we should make that list an attribute of the base class, this way we can easily reuse it.

schaefi: yes good point. part of the refactoring was to move the setup of the media dir to the base class as it's independent of the used tooling. That provides the lookup paths and I agree we could now also nicely use in the xorriso implementation.

I'll update the code.

However I'm currently discussing an obs issue with Adrian and Michael regarding Fedora builds and there seems to be something very very weird on the obs side. I'd like to get that sorted out before I update the code.

schaefi:
@Conan-Kudo ok we found the problem for the Fedora builds and obs is fixed now :) Many thanks to @adrianschroeter. @ggardet  @Conan-Kudo Please find the Fedora30 build and let me know if that works for you

kiwi/iso_tools/xorriso.py
schaefi: @davidcassany I pushed a commit to address your issue. Should be good now

kiwi/iso_tools/xorriso.py
```diff
@@ -85,23 +89,40 @@ def init_iso_creation_parameters(self, custom_args=None):
...
+                self.iso_loaders += [
+                    '-boot_image', 'grub', 'bin_path={0}'.format(loader_file),
+                    '-boot_image', 'grub', 'grub2_mbr={0}'.format(mbr_file),
+                    '-boot_image', 'grub', 'grub2_boot_info=on'
+                ]
+            else:
+                loader_file = self.boot_path + '/loader/isolinux.bin'
+                mbr_file = Path.which(
+                    'isohdpfx.bin', Defaults.get_syslinux_search_paths()
+                )
```
davidcassany: This `Path.which` call returns `None` if not found, I guess this is always an issue. Probably we could add something like `raise_not_found` in `Path.which` add raise an exception of path not found if set to true. What you think? I discovered that playing around with Arch an it took me while to figure out which path was missing by looking at the xorriso command line. Probably providing an exception stating the missing path is more clear.

I would be fine if we handle that in a separate PR and probably find out all the places where it could make sense. Probably there are other `Path.which` in the code where this applies.

schaefi: When I implemented that it was done on purpose. So in the code you will always find the result of Path.which to be followed by a conditional check. I'm ok if we refactor that to be exception based like Command but for the moment I don't see a high demand and yes if this will be done then in a separate PR which will probably quite big because it's used pretty often in the code

davidcassany: Fine with me, we can leave it as is for the time being and reconsider it if at some point the topic raises again. I am also fine leaving `mbr_file` without a conditional check since it is just used straight after the `Path.which` call.

ggardet:
@schaefi for `Boot from Hard Disk` entry, we could just use `exit`. 
This is what is done on openSUSE installation ISO.

schaefi:
@ggardet can you give me an example in grub speech how the ```exit``` would look like ? and what happens in this case ? Thanks

schaefi:
@Conan-Kudo any news on the Fedora30 build ? are you ok with it I'd like to get this one merged if possible. Thanks

Conan-Kudo:
@schaefi the `exit` command makes the bootloader quit and the EFI system moves on to the next boot device, I believe.

schaefi:
> This changeset is huge, but it appears to work. I have Fedora 30 AArch64 ISO that boots from shim to grub properly.

Great, thanks for testing. Yes this is a big move but I think we implemented it well tested. Factory/TW will be the first distro that will have the change and I will submit after another round of testing the integration test builds. Thanks

Conan-Kudo:
@schaefi I'm pretty sure I can beat openSUSE to shipping a new release of kiwi. :)

schaefi:
:smile: yes you can

schaefi:
regarding the grub boot from hard disk

```
menuentry "Boot from Hard Disk" --class os --unrestricted {
    exit
}
```

you are saying that should do it ?

ggardet:
> regarding the grub boot from hard disk
> 
> ```
> menuentry "Boot from Hard Disk" --class os --unrestricted {
>     exit
> }
> ```
> 
> you are saying that should do it ?

Yes, that should do it.  As @Conan-Kudo wrote, grub will exit and next EFI entry will be tried as if you have no live image. I can try a last image before merge with this added, if you want.

schaefi:
I have tested this method in an x86 build and it worked nicely. Thus I added one last commit :)

Thanks

ggardet:
> I have tested this method in an x86 build and it worked nicely. Thus I added one last commit :)
> 
> Thanks

Perfect! I am looking forward to get this PR merged and pushed to Tumbleweed! :D

ggardet:
@schaefi Could you push version 9.17.41 (latest one) to Virtualization:Appliances:Builder/python-kiwi and to Factory, please?

schaefi:
>    [1]@schaefi Could you push version 9.17.41 (latest one) to
>    Virtualization:Appliances:Builder/python-kiwi and to Factory, please?

yes but I'm running the integration testing for x86 right now and
wanted to be sure nothing break. So give me some time. Probably by
tomorrow. I'm also traveling today so only limited resouces

ggardet:
Ok, no problem. Thanks.

PR #1117, opened 2019-07-01
Do not supress output for dracut call

### LINKED ISSUE #1114: dracut output should not be suppressed
## Problem description
I built an vmx image with LVM setup for root and /var and it did not boot, because "lvm2" was not in the packages list.

It would have helped the debugging, if dracut output during build would have been available, a message like:

```
dracut: dracut module 'lvm' will not be installed, because command 'lvm' could not be found!
```

would surely helped pin down the issue :-)

davidcassany:
This commit adds an option to CommandProcess class to redirect stderr
to stdout at python level. CommandProcess object still keeps the stdout
and stderr as separate attributes, however the output iterator, that used
to return stdout only lines, now is capable to return both. If
stderr_to_stdout flag is set to True, CommandProcess iterator will
return both, stdout and stderr output lines.

This is usefull to watch all output, stderr and stdout, on real time.

Fixes #1114

### DIFF
--- a/kiwi/boot/image/dracut.py
+++ b/kiwi/boot/image/dracut.py
@@ -98,7 +98,7 @@ class BootImageDracut(BootImageBase):
             else:
                 included_files = self.included_files
             dracut_initrd_basename += '.xz'
-            Command.run(
+            dracut_call = Command.run(
                 [
                     'chroot', self.boot_root_directory,
                     'dracut', '--force',
@@ -108,8 +108,10 @@ class BootImageDracut(BootImageBase):
                 ] + self.dracut_options + included_files + [
                     dracut_initrd_basename,
                     kernel_details.version
-                ]
+                ],
+                stderr_to_stdout=True
             )
+            log.debug(dracut_call.output)
             Command.run(
                 [
                     'mv',
--- a/kiwi/command.py
+++ b/kiwi/command.py
@@ -43,7 +43,9 @@ class Command(object):
     stdout and stderr is given to the caller
     """
     @staticmethod
-    def run(command, custom_env=None, raise_on_error=True):
+    def run(
+        command, custom_env=None, raise_on_error=True, stderr_to_stdout=False
+    ):
         """
         Execute a program and block the caller. The return value
         is a hash containing the stdout, stderr and return code
@@ -60,6 +62,7 @@ class Command(object):
         :param list command: command and arguments
         :param list custom_env: custom os.environ
         :param bool raise_on_error: control error behaviour
+        :param bool stderr_to_stdout: redirects stderr to stdout
 
         :return:
             Contains call results in command type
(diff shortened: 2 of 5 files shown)


davidcassany:
@schaefi for some reason I do not understand I also had to update the `path_test.py` which is absolutely unrelated of my PR. I guess this is cuased because I updated my host to Leap 15.1, however I can't understand why it has not appeared before (e.g. Factory). Now I am looking at Travis tests to see if all pass in all versions.

schaefi:
yes this was caused by an update of pytest on travis. I have fixed this in my open zipl PR which also has no relation. Let me merge my PR first then you can rebase

kiwi/boot/image/dracut.py
```diff
@@ -110,6 +111,11 @@ def create_initrd(self, mbrid=None, basename=None, install_initrd=False):
                     kernel_details.version
                 ]
             )
+            process = CommandProcess(
+                command=cmd_call, log_topic='initrd creation',
+                stderr_to_stdout=True
+            )
+            process.poll()
```
schaefi: Hmm, I don't understand why we poll on the process without any interactive status information. That should not be necessary and could be done simpler imho:

```python
dracut_call = Command.run(...)
log.debug(dracut_call.output)
# or maybe: log.debug(dracut_call.error) if dracut writes to stderr
```

Any reason why it can't be done that way ?

schaefi: I like the stderr to stdout extension though

davidcassany: Well yes I agree there is no need to get this output using poll. I just did it this way because it is more pleasant if you are actively looking at the logs during the build. Dracut output is reasonably huge and wanted to avoid dumping all output at once after the process is done. Dracut output is produced and shown to the user sequentially like a `zypper in -y pkg` call. So I thought about printing each line as soon as dracut produces it.

I know this is a weak argument :sweat_smile: I am opened to change that if you prefer.

schaefi: ok got it, makes sense to me. But can you add a short line of explanation. I will surely forgot it and I bed someone will think this needs to be refactored :) Thanks

schaefi:
@davidcassany Thinking more about the merge of stdout and stderr. I think you should handled that on the level of subprocess. Like this:

```python
subprocess.Popen(args,stdout=subprocess.PIPE,stderr=subprocess.STDOUT)
```

We should avoid that join and also special bytes handling. Thoughts ?

schaefi:
The code will also become a nicer read with the above concept

davidcassany:
@schaefi I agree with this approach you suggested, this is cleaner. Gonna update that

davidcassany:
At the end  opted for a simple implementation, the use of 
```
subprocess.Popen(args,stdout=subprocess.PIPE,stderr=subprocess.STDOUT)
```
was not straight forward in `Command.call` method as the error output then appears to be a different object (or `None`) and I did not want to perform a hard to test refactor on this area. I just added the `stderr_to_stdout` optional flag in `Command.run` and then just log all at once the dracut output using this new `Command.run` optional parameter.

PR #1118, opened 2019-07-02
Remove '^' which break BaseStripFirmware function (boo#1132455)

ggardet:
The `^` in `BaseStripFirmware` function make the firmware end in `/lib/firmware/lib/firmware` instead of `/lib/firmware`
It fixes https://github.com/SUSE/kiwi/commit/3a8e31b26dc8b317d35f55a5390d3a731f1aac65

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -749,7 +749,7 @@ function baseStripFirmware {
         for fname in $name ; do
             for match in /lib/firmware/"${fname}" /lib/firmware/*/"${fname}";do
                 if [ -e "${match}" ];then
-                    match="${match//^\/lib\/firmware\//}"
+                    match="${match//\/lib\/firmware\//}"
                     bmdir=$(dirname "${match}")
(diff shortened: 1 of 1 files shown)


ggardet:
@schaefi could you review, please?

schaefi:
Looks good to me. I assume you have tested this ?

ggardet:
> Looks good to me. I assume you have tested this ?

I just tested the loop with output of `match` and `bmdir`. You can check http://download.opensuse.org/ports/aarch64/tumbleweed/images/openSUSE-Tumbleweed-ARM-JeOS-raspberrypi3.aarch64-2019.05.17-Snapshot20190607.raw.xz where you have `/lib/firmware/lib/firmware` folder. But this image has no firmware inside the wrong folder, though.
`*.txt` and `LICENSE` files are added in the right folder and have no `^` in the pattern.

schaefi:
great :+1: thanks much for your effort

PR #1122, opened 2019-07-03
Do not crash on missing HOME

nadvornik:
* this fixes a crash when HOME is not set
* can be reproduced by calling kiwi via salt


schaefi:
Looks good, thanks. No HOME in env is rather rare, must be a special env ;)

Bischoff:
Still reproducing on Leap 15.0 with all updates :crying_cat_face: .

@schaefi and @davidcassany, are there plans to get it into the various distros?

@nadvornik we will be in trouble for retail if it remains in source code without being submitted.

PR #1126, opened 2019-07-04
Fix test breakage when /etc/kiwi.yml is present

dcermak:
The unit test for the `RuntimeConfig` class are also testing the default values in some functions. To work around a possibly existing `~/.config/kiwi/config.yml`, several functions change the value of `HOME`. This is unfortunately not sufficient if a `/etc/kiwi.yml` exists, as that one is taken as a fallback *if* `~/.config/kiwi/config.yml` isn't present.

This issue is fixed by creating a `default_runtime_config` with `os.path.exists` patched to return False. Thereby none of the runtime configuration files will be taken into account.


schaefi:
Thanks :+1:

PR #1129, opened 2019-07-08
Fixed get_persistent_device_from_unix_node

schaefi:
The method get_persistent_device_from_unix_node takes a device
path and looks up it's basename to be present as persistent
variant below the given schema, e.g by-id. However if the
method receives an already persistent device path e.g from
/dev/mapper the persistent schema representation will only
match a numbered dm-N device and not the map name. The method
then returns nothing but should return the original device
path. This caused bad after effects like reported in
Issue SUSE-Enceladus/azure-li-services#163


schaefi:
Thanks for the review, I'll merge and create a test build

PR #1130, opened 2019-07-09
Drop support for python2

### LINKED ISSUE #1036: Drop support for Python2
## Problem description

Python2 is announced to be unmaintained from Jan 2020. KIWI supports Python 2.7 and it should not support any python version that is not maintained upstream. Because of that KIWI should drop python2 support **before** the upstream dead line. 

Dropping python2 support 6 months in advance should be enough to enable an smooth transition, meaning that by June 2019 KIWI should drop python2.7 support.

schaefi:
Python2 is announced to be unmaintained from Jan 2020.
KIWI supports Python 2.7 and it should not support any python version that
is not maintained upstream. This Fixes #1036

### DIFF
--- a/helper/completion_generator.py
+++ b/helper/completion_generator.py
@@ -248,6 +248,5 @@ function __comp_reply {
 
 complete -F _kiwi -o default kiwi
 complete -F _kiwi -o default kiwi-ng
-complete -F _kiwi -o default kiwi-ng-2
 complete -F _kiwi -o default kiwi-ng-3
 ''').strip())
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -25,17 +25,21 @@
 %global python3_sitelib %(%{__python3} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
 %endif
 
+%if 0%{?el7}
+%global python3_pkgversion 36
+%else
+%{!?python3_pkgversion:%global python3_pkgversion 3}
+%endif
+
 %if 0%{?debian} || 0%{?ubuntu}
 %global is_deb 1
 %global pygroup python
 %global sysgroup admin
 %global develsuffix dev
-%global update_alternatives %{_bindir}/update-alternatives
 %else
 %global pygroup Development/Languages/Python
 %global sysgroup System/Management
 %global develsuffix devel
-%global update_alternatives %{_sbindir}/update-alternatives
 %endif
 
 Name:           python-kiwi
@@ -53,19 +57,12 @@ Group:          %{pygroup}
 Source:         %{name}.tar.gz
 Source1:        %{name}-rpmlintrc
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
-%if 0%{?fedora} || 0%{?suse_version}
 BuildRequires:  gcc
-BuildRequires:  python3-devel
-%endif
-%if 0%{?fedora} || 0%{?suse_version} || 0%{?ubuntu} >= 1804 || 0%{?debian} >= 9
-BuildRequires:  python3-setuptools
+BuildRequires:  python%{python3_pkgversion}-%{develsuffix}
+BuildRequires:  python%{python3_pkgversion}-setuptools
 BuildRequires:  fdupes
-%endif
-BuildRequires:  python-%{develsuffix}
-BuildRequires:  python-setuptools
 %if 0%{?suse_version}
 BuildRequires:  shadow
-BuildRequires:  update-alternatives
 %endif
 %if 0%{?debian} || 0%{?ubuntu}
 BuildRequires:  passwd
@@ -79,27 +76,23 @@ The KIWI Image System provides an operating system image builder
 for Linux supported hardware platforms as well as for virtualization
 and cloud systems like Xen, KVM, VMware, EC2 and more.
 
-%if 0%{?fedora} || 0%{?suse_version} || 0%{?ubuntu} >= 1804 || 0%{?debian} >= 9
 # python3-kiwi
-%package -n python3-kiwi
+%package -n python%{python3_pkgversion}-kiwi
 Summary:        KIWI - Appliance Builder Next Generation
 Group:          Development/Languages/Python
 Recommends:     jing
 %if 0%{?ubuntu} || 0%{?debian}
-Requires:       python3-yaml
+Requires:       python%{python3_pkgversion}-yaml
 %else
-Requires:       python3-PyYAML
+Requires:       python%{python3_pkgversion}-PyYAML
 %endif
-Requires:       python3-docopt
-Requires:       python3-lxml
-Requires:       python3-requests
-Requires:       python3-setuptools
-Requires:       python3-xattr
+Requires:       python%{python3_pkgversion}-docopt
+Requires:       python%{python3_pkgversion}-lxml
+Requires:       python%{python3_pkgversion}-requests
+Requires:       python%{python3_pkgversion}-setuptools
+Requires:       python%{python3_pkgversion}-xattr
 # tools used by kiwi
 %if 0%{?suse_version}
-Requires:       update-alternatives
-Requires(post): update-alternatives
-Requires(postun): update-alternatives
 %ifarch x86_64
 Requires:       grub2-x86_64-efi
 %endif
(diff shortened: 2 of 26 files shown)


package/python-kiwi-spec-template
```diff
@@ -531,14 +417,6 @@ done
     --install %_bindir/kiwicompat kiwicompat %_bindir/kiwicompat-3 10
 %endif
```
Conan-Kudo: Can we rip alternatives out entirely now? We don't need it if there's only one variant, right?

Conan-Kudo:
@schaefi Note that we're not currently building `python3-kiwi` for Ubuntu 16.04, so are we okay with OBS not being able to make those anymore with newer versions?

Conan-Kudo:
@schaefi We also lose support for building on OBS for CentOS 7, because we don't have the dependencies in place...

schaefi:
Hmm, yes let me check that. CentOS7, I think we don't want to loose it...

Conan-Kudo:
@schaefi We're only missing one dependency from EPEL, which we could package in OBS: `python3-xattr`

schaefi:
> Note that we're not currently building python3-kiwi for Ubuntu 16.04

That's imho ok. We are running integration testing against Ubuntu:18.04 (bionic)

schaefi:
> @schaefi We're only missing one dependency from EPEL, which we could package in OBS: `python3-xattr`

I do that already here: https://build.opensuse.org/package/binaries/Virtualization:Appliances:Staging/python-xattr/CentOS_7

schaefi:
But the integration test uses python2-kiwi... Virtualization:Appliances:Images:Testing_x86/test-image-
centos. Let me check why

schaefi:
The package build from here:

* https://build.opensuse.org/package/binaries/Virtualization:Appliances:Staging/python-kiwi/CentOS_7

does not build a python3-kiwi

setup.py
```diff
@@ -213,9 +211,9 @@ def run(self):
        'Intended Audience :: Developers',
        'License :: OSI Approved :: GNU General Public License v2 (GPLv2)',
        'Operating System :: POSIX :: Linux',
-       'Programming Language :: Python :: 2.7',
-       'Programming Language :: Python :: 3.4',
        'Programming Language :: Python :: 3.5',
```
dcermak: (although unrelated) Shouldn't we also test Python 3.5 if we officially support it?

package/python-kiwi-spec-template
schaefi: yes true, haven't thought about that part. How would we handle if different versions of py3 are installed ? py3.6, py3.7 ? suse doesn't support this, but on Fedora for example you can have multiple python3's installed

schaefi: One of the reasons why the gitlab pipeline uses a fedora container that comes with many py3 versions pre-installed

Conan-Kudo: Yes, but packages cannot depend on anything except the main python3 environment (which is the `python3` package). All the other ones are for dev/test only.

setup.py
schaefi: I think testing 3.6 and 3.7 is good enough for the upstream code base

package/python-kiwi-spec-template
schaefi: ok in this case we can rip out alternatives completely... that's nice

schaefi: you did it :)

helper/completion_generator.py
```diff
@@ -248,6 +248,4 @@ def traverse(self, tree=None, level=0, origin=None):
 
 complete -F _kiwi -o default kiwi
 complete -F _kiwi -o default kiwi-ng
-complete -F _kiwi -o default kiwi-ng-2
-complete -F _kiwi -o default kiwi-ng-3
```
Conan-Kudo: This file still exists, so you probably want to keep it, especially as other distros may not necessarily create the symlinks..

schaefi: Hmm, ok I revert it

Conan-Kudo: You can drop `kiwi-ng-2` entry, though.

schaefi: yes will do that in another commit, just a sec

package/python-kiwi-spec-template
```diff
@@ -19,34 +19,32 @@
...
 %global pygroup Development/Languages/Python
 %global sysgroup System/Management
 %global develsuffix devel
-%global update_alternatives %{_sbindir}/update-alternatives
 %endif
 
 Name:           python-kiwi
 Version:        %%VERSION
-Provides:       kiwi-schema = 6.6
+Provides:       kiwi-schema = 7.1
```
davidcassany: Wow, we were behind the published schema here, right? Unrelated with this PR, but probably we could parse this value from `k.image.schemaversion` in `kiwi/schema/kiwi.rnc`, so we have the schema version in a single place.

davidcassany:
@schaefi I am wondering how are we going to deal with it, meaning, it would be nice to merge that once we have a green (or green enough) result in integration tests, so the builds from our OBS test project are building. Are you doing some tests in some other area that is not `Virtualization:Appliances:Images:Testing_x86`?

package/python-kiwi-spec-template
schaefi: I realized the provides on accident and fixed it with this PR. You are right it was not related but as we touched the spec I thought this should be corrected. The provides tag here was exclusively used by the yast image builder module. This was a project that is dead since a long time. Actually I have thought to delete this provides but as I don't know who might use it, I just kept it.

The idea to put that setting into some automation is good, but let's do it in a separate PR

schaefi:
> @schaefi I am wondering how are we going to deal with it, meaning, it would be nice to merge that once we have a green (or green enough) result in integration tests, so the builds from our OBS test project are building. Are you doing some tests in some other area that is not `Virtualization:Appliances:Images:Testing_x86`?

There are also a couple of image builds that are not integration tests here: Virtualization:Appliances:Images

I check those as well but there is not more testing involved. I also trust the integration builds and those image builds to form a stable base. We introduce this change for Factory, TW... I'm not aware of any build that uses python2 here. Are you afraid of the change to break at other stages ?

schaefi:
@Conan-Kudo given the current state of the art, are you ok with it ?

Conan-Kudo:
@schaefi Yeah, I think now is as good of a time to pull the trigger as any.

davidcassany:
> I'm not aware of any build that uses python2 here. Are you afraid of the change to break at other stages ?

Well I was thinking more in non openSUSE images we have in x86 tests. CentOS, Fedora, etc. Even I think CentOS is the only one were we had a real requirement for python2 due to some py3 missing dependencies. My concern is not that the update might break in other stages or something terrible might happen. I am just _afraid_ of landing this on staging and starting to see Ubuntu, Fedora, CentOS not resolving due to something it can't be fixed in KIWI/OBS.

That's why I asked if it was tested in OBS in some other project as testing in our openSUSE host other distros might not be that easy and not reveal failures/misses in spec dependencies.

schaefi:
> Actually, the documentation needs to be fixed to remove the Python 2 references. For example: https://opensource.suse.com/kiwi/installation.html

This is done here in this request: https://github.com/SUSE/kiwi/pull/1130/files#diff-5ce866a82d21a3cdae0ca0aa1453453c

The docs will be automatically change once this is merged and a new tag is set

schaefi:
@davidcassany I'm sorry I did not understand what you mean ?

davidcassany:
> @davidcassany I'm sorry I did not understand what you mean ?

Nevermind no big issues, I am not very good with words today... it is just that am wondering of having CentOS builds, and probably others, in  `Virtualization:Appliances:Images:Testing_x86` unresolvable after merging the change. So that I was curious to know if this code was tested somehow against our tests in Testing_x86, because I presume that, if something like that happens, might not be that straight forward (or directly in our hands) to fix it.

Hopefully they all will build straight nothing to worry about :)

schaefi:
So CentOS was the only target that only builds with the py2 stack. So far I trust the Epel repo and the patch from here to build a correct py3 version of kiwi that should bring us on the py3 train for CentOS as well. If there are py3 stuff not available and we will find out, there is still the opportunity to do the same as we still have with the xattr package. So after all I'd like to merge this and see if we are on py3 for all distros and builds and if something fails we will find out. I'm sorry I could not come up with another idea. It's a bit of a chicken and egg problem

PR #1131, opened 2019-07-09
Don't add config.vm.base_mac to libvirt vagrant boxes

### LINKED ISSUE #1119: Don't set vm.base_mac in Vagrantfile created by kiwi
## Problem description

When kiwi builds a vagrant box, then it embeds a `Vagrantfile` into the box with roughly these contents (for Virtualbox boxes, for libvirt there's a little more):
``` ruby
Vagrant.configure("2") do |config|
  config.vm.base_mac = "00163E010101"
end
```

This can be a problem: depending on your exact setup, your boxes might not come online, because the DHCP server will assign all of them the same IP.

## Expected behavior

Either we shouldn't set the MAC address at all or create a random one for each launched VM, but this MAC address **must** persist over reboots.

## References

- I've asked upstream what is the official recommendation: https://github.com/hashicorp/vagrant/issues/10946

- @sigsteve ran into DHCP issues with KIWI boxes, but I was unable to reproduce these.

+ KIWI version: git master

dcermak:
- Refactor `VagrantConfigTemplate` to no longer require a `base_mac` by default
- Move `_random_mac()` to Virtualbox subclass
- Include config.vm.base_mac in Virtualbox boxes via `get_additional_vagrant_config_settings()`

This fixes #1119 (option 2)

### DIFF
--- a/kiwi/storage/subformat/template/vagrant_config.py
+++ b/kiwi/storage/subformat/template/vagrant_config.py
@@ -18,7 +18,6 @@
 import os
 
 from textwrap import dedent
-from string import Template
 
 
 class VagrantConfigTemplate(object):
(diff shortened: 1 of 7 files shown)


dcermak:
> This looks good to me +1 the bot complain about security is immaterial for creating a random mac in my eyes

Yes, it complains about that on master too, but I haven't found a way how to silence this warning (afaik it comes from bandit, which doesn't offer a way how to ignore certain issues).

PR #1133, opened 2019-07-11
Fixed spec for use with Recommends

schaefi:
The Recommends tag is not known by any distro

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -80,7 +80,9 @@ and cloud systems like Xen, KVM, VMware, EC2 and more.
 %package -n python%{python3_pkgversion}-kiwi
 Summary:        KIWI - Appliance Builder Next Generation
 Group:          Development/Languages/Python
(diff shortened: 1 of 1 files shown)


schaefi:
@Conan-Kudo I needed this spec fix to build for CentOS. Fedora and also Debian built with the Recommends: tag set but I think it only makes sense for suse ?

package/python-kiwi-spec-template
```diff
@@ -80,7 +80,9 @@ and cloud systems like Xen, KVM, VMware, EC2 and more.
 %package -n python%{python3_pkgversion}-kiwi
 Summary:        KIWI - Appliance Builder Next Generation
 Group:          Development/Languages/Python
+%if 0%{?suse_version}
```
Conan-Kudo: `%if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version} || 0%{?debian} || 0%{?ubuntu}`

schaefi: Thanks, will go with that one

PR #1135, opened 2019-07-12
Fix creation of /etc/sysconfig/bootloader for ISOs

### LINKED ISSUE #1112: BootLoaderConfigGrub2.write in EFI live images is called too late?
## Problem description

I am creating a JeOS based live image, that I expect that will work in UEFI and BIOS systems.

The image can be found here:

https://build.opensuse.org/package/show/home:aplanas:Images/openSUSE-Tumbleweed-JeOS

The logs can be found here:

https://build.opensuse.org/build/home:aplanas:Images/images/x86_64/openSUSE-Tumbleweed-JeOS:Live/_log

Investigating a jeos-firstboot issue (https://github.com/openSUSE/jeos-firstboot/pull/52), we were expecting `/etc/sysconfig/bootloader` to be present in every case, but the previous image shows that the file is missing.

Debugging kiwi I found that the `setup_sysconfig_bootloader()`, called by `BootLoaderConfigGrub2.write()` is called when the LiveOS/squashfs.img is already present, and in the `/tmp/.../live-media.XXXXXX/` directory only contains:

- EFI/
- LiveOS/
- boot/

This makes in `setup_sysconfig_bootloader` that `os.path.exists(sysconfig_bootloader_location)` do not exists anymore, and the bootloader file is not created.

I am not able to see if this is a bug or not, as the check itself suggest that is expected that the directory is not present, and the final ISO can be booted properly in EFI and BIOS systems.

## Expected behaviour

I would expect that the file `/etc/sysconfig/bootloader` is present in the media.

davidcassany:
This commit forces the creation of /etc/sysconfig/bootloader
before building the squashfs container. Since bootloader configuration
was generally prepared in a different area of the root-tree after
the creation of the squashfs container this sysconfig file was never
included into the image.

Fixes #1112

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -200,11 +200,11 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 self.cmdline_failsafe
             )
 
-        log.info('Writing sysconfig bootloader file')
         sysconfig_bootloader_location = ''.join(
             [self.root_dir, '/etc/sysconfig/']
         )
         if os.path.exists(sysconfig_bootloader_location):
+            log.info('Writing sysconfig bootloader file')
             sysconfig_bootloader_file = ''.join(
                 [sysconfig_bootloader_location, 'bootloader']
             )
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -139,6 +139,16 @@ class LiveImageBuilder(object):
             }
         }
 
+        # create /etc/sysconfig/bootloader if needed
+        if self.firmware.efi_mode():
+            bootloader_sysconfig = BootLoaderConfig(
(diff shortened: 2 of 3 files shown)


davidcassany:
@schaefi I added the question mark and work in progress labels to this PR because I have some doubts regarding the implementation. I came up to this solution because it is not that intrusive, but I dislike the fact of modifying the root-tree created during the prepare step. I was thinking of deleting again this file right after the real grub2 bootloader configuration, but probably this is even worse.

Do you think it is acceptable to add this file in the root tree during the `create` step? I don't like it but don't see how it could harm, as if re run the create step for a different image target this file be either ignored or rewritten.

schaefi:
Here is my idea:

https://github.com/OSInside/kiwi/pull/1138

Thoughts ?

davidcassany:
Closing #1138 is way better implementation for the same issue.

PR #1136, opened 2019-07-12
Make extended attribute check an on demand feature

schaefi:
Support for the xattr module is handled differently at the side
of the distribution vendors. There is xattr(GPL) and pyxattr(MIT)
and both might exist or not depending on the version of the desired
distribution. To deal better with this we made the strict requirement
in kiwi for xattr a distribution specific recommendation in the spec
file and changed the code to load the xattr module as an on demand
plugin. If no xattr module support exists the method
target_supports_extended_attributes() will return False which
effectively result in a warning message and rsync operations to run
without the -X and -A options.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ version := $(shell \
 
 .PHONY: test
 test:
-	tox -e unit_py3_6
+	tox -e unit_py3_6 "-n 5"
 
 flake:
 	tox -e check
--- a/kiwi/utils/sync.py
+++ b/kiwi/utils/sync.py
@@ -16,8 +16,8 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
+import importlib
 from stat import ST_MODE
-import xattr
 
 # project
 from kiwi.logger import log
@@ -94,6 +94,13 @@ class DataSync(object):
 
         :rtype: bool
         """
+        try:
+            xattr = importlib.import_module('xattr')
+        except Exception as issue:
+            log.warning(
+                'Could not load xattr module: {0}'.format(issue)
+            )
+            return False
         try:
             xattr.getxattr(self.target_dir, 'user.mime_type')
         except Exception as e:
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -83,6 +83,12 @@ Group:          Development/Languages/Python
 %if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version} || 0%{?debian} || 0%{?ubuntu}
 Recommends:     jing
 %endif
+%if 0%{?fedora} || 0%{?rhel} >= 8
+Recommends:     python%{python3_pkgversion}-pyxattr
+%endif
+%if 0%{?suse_version} || 0%{?debian} || 0%{?ubuntu}
+Recommends:     python%{python3_pkgversion}-xattr
+%endif
 %if 0%{?ubuntu} || 0%{?debian}
 Requires:       python%{python3_pkgversion}-yaml
 %else
@@ -92,7 +98,6 @@ Requires:       python%{python3_pkgversion}-docopt
 Requires:       python%{python3_pkgversion}-lxml
 Requires:       python%{python3_pkgversion}-requests
 Requires:       python%{python3_pkgversion}-setuptools
-Requires:       python%{python3_pkgversion}-xattr
 # tools used by kiwi
 %if 0%{?suse_version}
 %ifarch x86_64
--- a/test/unit/utils_sync_test.py
+++ b/test/unit/utils_sync_test.py
@@ -1,6 +1,8 @@
 import os
 from stat import ST_MODE
-from mock import patch
+from mock import (
+    patch, Mock
+)
 
 from kiwi.utils.sync import DataSync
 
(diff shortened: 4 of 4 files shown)


schaefi:
I'm not sure if it makes sense but we could go that way. Your feedback is pretty much appreciated

Thanks

schaefi:
> This PR is far too complex for dealing with that change. :)

from a code complexity point of view I think it's not a big deal. But I agree the solution for the problem could be simpler.

I thought this might be helpful just in case we could not come up with a consistent solution on the spec file level. I thought it could be helpful to let you deal with one spec patch less.

If we can provide a py3 variant of xattr in epel7 this would probably be the best solution. Because of the license issue I'm not sure how likely it is that this will become true and so I coded this patch.

It's just a plan B

schaefi:
JFI: there is an initiative to package pyxattr here:

```
Hi Dan,

> I don't use CentOS/RHEL on a daily basis, so I would prefer to do not
> have maintain it there. However, having a co-maintainer which would take
> care of that is perfectly fine for me (and in general I would be happy
> to have another person looking at the package :) ).

I have some experience in package maintenance, but not too much. I'm
currently maintaining a few python and C/C++ libraries/utilities in
Fedora and the Vagrant stack in openSUSE.

> Btw, I was planning to update pyxattr to 0.6.1 in Fedora/devel. Would
> you mind to make that upgrade and create a pull request in
> https://src.fedoraproject.org/rpms/pyxattr as a test run? :)

Done: https://src.fedoraproject.org/rpms/pyxattr/pull-request/2

I've taken the liberty and simplified the spec a bit along the line.
```

@dcermak thanks much for taking care, this is great

Makefile
```diff
@@ -13,7 +13,7 @@ version := $(shell \
 
 .PHONY: test
 test:
-	tox -e unit_py3_6
+	tox -e unit_py3_6 "-n 5"
```
dcermak: This shouldn't be hardcoded to 5.

dcermak:
> JFI: there is an initiative to package pyxattr here:
> 
> ```
> Hi Dan,
> 
> > I don't use CentOS/RHEL on a daily basis, so I would prefer to do not
> > have maintain it there. However, having a co-maintainer which would take
> > care of that is perfectly fine for me (and in general I would be happy
> > to have another person looking at the package :) ).
> 
> I have some experience in package maintenance, but not too much. I'm
> currently maintaining a few python and C/C++ libraries/utilities in
> Fedora and the Vagrant stack in openSUSE.
> 
> > Btw, I was planning to update pyxattr to 0.6.1 in Fedora/devel. Would
> > you mind to make that upgrade and create a pull request in
> > https://src.fedoraproject.org/rpms/pyxattr as a test run? :)
> 
> Done: https://src.fedoraproject.org/rpms/pyxattr/pull-request/2
> 
> I've taken the liberty and simplified the spec a bit along the line.
> ```
> 
> @dcermak thanks much for taking care, this is great

Please note that as @Conan-Kudo pointed out: this isn't really going to solve the situation on CentOS, as we need to submit python3-pyxattr as a new package.

Conan-Kudo:
> Please note that as @Conan-Kudo pointed out: this isn't really going to solve the situation on CentOS, as we need to submit python3-pyxattr as a new package.

This shouldn’t be hard for @dcermak to do, though.

schaefi:
> Please note that as @Conan-Kudo pointed out: this isn't really going to solve the situation on CentOS, > as we need to submit python3-pyxattr as a new package.

yes as long as CentOS doesn't have the package we could provide a custom build in the Virt:App
area as we have it right now for the xattr package. My point with all this is that suse has the xattr package and CentOS/Fedora puts on the pyxattr package. It would be imho good if the patch @Conan-Kudo carries wouldn't be needed such that we can handle this in a simple %if condition that applies to the distribution and we only carry the custom pyxattr build as long as it has not yet settled.

At the moment the xattr build is a hard requirement. Users of kiwi on CentOS has to fetch it from our custom build project and without trying to find a solution this will stay forever. So I like the move we make right now

Conan-Kudo:
@schaefi It’s easy enough to get `pyxattr` shipped in openSUSE too and switch things over. More things seem to request `pyxattr` over `xattr` anyway…

schaefi:
yes I agree that is even better. However this switch will suffer from the same problem you have with CentOS. In SLES there is xattr not pyxattr. To submit a new package in SLES we have to run through an ECO process. No matter how damn simple the package is it will take us probably months to get there. Thus from my experience it's more practical to leave the working state... meaning kiwi requires xattr on suse, it is always present and we handle other distributions as it is appropriate there. I think we will not be able to avoid an %if condition in the spec file here in some shape or form

Conan-Kudo:
For legacy SLES, the spec conditional and `sed -e 's/pyxattr/xattr/' -i setup.py` for older SUSE distributions in prep should be sufficient.

schaefi:
I have updated the spec file to hopefully better cope with the xattr requirement.

* for suse distros we continue to require python-xattr because it is present on all sles, factory, leap and has proven to work as expected
* for any other distro we now require python-pyxattr as it seems to be the standard for any other distros. I looked up the package on all targets we build: Debian, Ubuntu, Fedora, CentOS

CentOS is the exception in the matrix. It does not provide pyxattr but with the effort from Dan this will happen some time in the future. Thus I deleted the python-xattr custom build from Virt:App:Staging and added a python-pyxattr build as long as we need it.

My hope is that at some point in time EPEL provides python-pyxattr for CentOS and I can simply delete the custom build from Virt:App:Staging, which then completes what I'd like to achieve:

a) no extra patch needed by Neal for his package build
b) no custom python module needed to build in the KIWI space to support a distribution

Hope all that makes sense

I'm closing this now, but keep the branch open

Thanks

PR #1139, opened 2019-07-15
Fixed ISO creation in EFI mode with cdrtools

schaefi:
The changes introduced in #1113 will use the configured bootloader
e.g grub to boot the ISO image in BIOS and EFI mode. The creation
process works flawlessly if xorriso is used. However if cdrtools
are configured the options passed to e.g mkisofs were wrong. In
addition it's not possible to create a hybrid ISO based on
isohybrid if the loader is not isolinux. If cdrtools are in use
the process to make an ISO hybrid bootable is always based on
isohybrid and thus only works with isolinux. This patch also
covers this case with a warning message and the consequence that
we have to skip the hybrid setup in this case.

### DIFF
--- a/kiwi/filesystem/isofs.py
+++ b/kiwi/filesystem/isofs.py
@@ -15,7 +15,11 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+import os
+from textwrap import dedent
+
 # project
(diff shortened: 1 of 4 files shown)


schaefi:
I'm happy when we can get rid of all the cdrtools support in kiwi. As soon as there is xorriso in SLES12-SP5 I will drop the entire cdrtools code

PR #1143, opened 2019-07-18
Added support for fscreateoptions attribute

### LINKED ISSUE #1109: Looking for a way to set fscreateoptions for type=vmx images
## Problem description

I would like to customize the filesystem creation flags in my image. the usecases are for example (all ext4 as a filesystem, but this might exist in other filesystems as well, but those are less relevant to me): 

* set the default data journalling mode to writeback. While this appears to be settable via fsmountoptions="data=writeback", it actually isn't settable that way because it needs to be set in the initrd prior remounting the root fs on boot, and that isn't propagating there. Setting the default journalling mode at create achives that transparently.

* set the default journal size creation flags. Initially ext4 creates a very small journal, but it might make sense for an image that is going to be resized to specify the journal size already from the start. 

* set the flag to turn off journal on creation of ext4 filesystems. This is useful for throwaway ci jobs, where we don't want to suffer through the performance overhead of handling journalling. 

## Expected behaviour

I expect something like this:
```

    <type
        image="vmx"
        filesystem='ext4'
        bootloader='grub2'
        format='qcow2'
        vga="normal"
        boottimeout="1"
        fsmountoptions='acl,noatime,nobarrier'
        fscreateoptions='^has_journal'
        >
      <size unit='G' additive='false'>10</size>
    </type>
```

I would expect this to be passed down to **mkfs.ext4 -O ^has_journal** ... correspondingly. in order to handle journal size (which is -J) it might be a good idea to allow further options to be embedded there via

```
        fscreateoptions='-J 512MB -O ^has_journal'
```

schaefi:
Along with the fsmountoptions attribute there is now also the
fscreateoptions attribute which allows to control the options
used for creating the filesystem. Please note, it's not kiwi's
task to validate that the given option string is supported
by the selected filesystem. This means providing wrong values
here will cause the build to fail at the time the filesystem
gets created. This Fixes #1109

### DIFF
--- a/doc/source/working_with_kiwi/xml_description.rst
+++ b/doc/source/working_with_kiwi/xml_description.rst
@@ -307,6 +307,25 @@ build types and will be covered here:
   via the `-o` flag to :command:`mount` and are included in
   :file:`/etc/fstab`.
 
+- `fscreateoptions`: Specifies the filesystem options used to create the
+  filesystem. In KIWI the filesystem utility to create a filesystem is
+  called without any custom options. The default options are filesystem
+  specific and are provided along with the package that provides the
+  filesystem utility. For the Linux `ext[234]` filesystem, the default
+  options can be found in the :file:`/etc/mke2fs.conf` file. Other
+  filesystems provides this differently and documents information
+  about options and their defaults in the respective manual page, e.g
+  :command:`man mke2fs`. With the `fscreateoptions` attribute it's possible
+  to directly influence how the filesystem will be created. The options
+  provided as a string are passed to the command that creates the
+  filesystem without any further validation by KIWI. For example, to turn
+  off the journal on creation of an ext4 filesystem the following option
+  would be required:
+
+  .. code:: xml
+
+      <type fscreateoptions="-O ^has_journal"/>
+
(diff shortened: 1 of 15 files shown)


doc/source/working_with_kiwi/xml_description.rst
```diff
@@ -307,6 +307,9 @@ build types and will be covered here:
   via the `-o` flag to :command:`mount` and are included in
   :file:`/etc/fstab`.
 
+- `fscreateoptions`: Specifies the filesystem options used to create the
```
dcermak: Please extend the docs about this feature a bit. If I didn't know what this was about, I wouldn't be able to figure out how to use it. 

I can write something for that if that helps.

davidcassany: I would remark that the string provided here will just be part of the fs creation command call. This is not a comma separated value or anything like that. Probably adding here an straight forward example for ext4 as #1109 would make sense.

schaefi: fair enough, how about it now ?

dcermak: Very good, thank you!

Issue #1039, opened 2019-04-01
pxe image with compressed flag bundles both result files


schaefi:
## Problem description

Originally reported in https://bugzilla.suse.com/show_bug.cgi?id=1086257

> I have compressed PXE image

```
<type image="pxe" boot="netboot/suse-SLES12" compressed="true" .... </type>
```

> After building there are both compressed and uncompressed images.
> Then kiwi --bundle-build creates a tarball that contain both these files,
> so the size is at least doubled.

## Expected behaviour

Only keep the compressed variant if the compressed flag is set

schaefi:
This change needs approval by the OBS team, setting the feedback flag to not forget

schaefi:
@adrianschroeter any thoughts on this one ? Thanks

schaefi:
This is on the todo list and we are working on it.

davidcassany:
I don't see this issue happening in KIWI-NG, I just see it in former KIWI. I just did a PR for it there: OSInside/kiwi-legacy#678

PR #1145, opened 2019-07-19
Do not recompress PXE image if compressed flag is true

davidcassany:
This commit is two fold:

* Refactor PXE builder to make use of the ArchiveTar instead of
  directly calling tar command.
* Do not compress PXE tarball if the contained image is already
  compressed according the type's compressed attribute.

Related to #1039

### DIFF
--- a/kiwi/builder/pxe.py
+++ b/kiwi/builder/pxe.py
@@ -20,7 +20,6 @@ import platform
 
 # project
 from kiwi.defaults import Defaults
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi I did not find references in docs regarding the pxe_archive (the tarball containing everything) since our how to's on PXE images just refer to directly to the relevant files that are also extracted in the target-dir folder.

PR #1146, opened 2019-07-20
Fix the pyxattr dependency to allow kiwi to function

Conan-Kudo:
Most Linux distributions offer the `pyxattr` module, including openSUSE Tumbleweed. Going forward, we will use the `pyxattr` module by default as a dependency and only switch back to the
other `xattr` module when on older SUSE Linux distributions that lack the `pyxattr` module.

Note that because kiwi uses `setuptools` to create the CLI entry points, kiwi checks the Python dependencies before executing, so we change the dependency in the `setup.py` accordingly so that it will not fail to start.

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -92,7 +92,7 @@ Requires:       python%{python3_pkgversion}-docopt
 Requires:       python%{python3_pkgversion}-lxml
 Requires:       python%{python3_pkgversion}-requests
 Requires:       python%{python3_pkgversion}-setuptools
-%if 0%{?suse_version}
+%if (0%{?suse_version} && 0%{?suse_version} < 1550)
 Requires:       python%{python3_pkgversion}-xattr
 %else
 Requires:       python%{python3_pkgversion}-pyxattr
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
@schaefi Note, this is intended to fix the breakage we're seeing in testing on CentOS without breaking SUSE:

```
[   75s] cd /usr/src/packages/SOURCES && rm -rf /usr/src/packages/KIWI-oem && LANG=en_US.UTF-8 /usr/bin/kiwi --debug  --profile Disk system build --description /usr/src/packages/SOURCES --target-dir /usr/src/packages/KIWI-oem 
[   75s] Traceback (most recent call last):
[   75s]   File "/usr/bin/kiwi", line 6, in <module>
[   75s]     from pkg_resources import load_entry_point
[   75s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3095, in <module>
[   75s]     @_call_aside
[   75s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3079, in _call_aside
[   75s]     f(*args, **kwargs)
[   75s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3108, in _initialize_master_working_set
[   75s]     working_set = WorkingSet._build_master()
[   75s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 570, in _build_master
[   75s]     ws.require(__requires__)
[   75s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 888, in require
[   75s]     needed = self.resolve(parse_requirements(requirements))
[   75s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 774, in resolve
[   75s]     raise DistributionNotFound(req, requirers)
[   75s] pkg_resources.DistributionNotFound: The 'xattr' distribution was not found and is required by kiwi
```

schaefi:
Thanks much, I was just looking at the failed tests but you already fixed it :) very cool

PR #1149, opened 2019-07-22
Fixed use of Command calls in destructors

schaefi:
The Command class creates namedtuple to store the result
of the call. That tuple is created inside of the Command
namespace. If called inside of a __del__ destructor which
is often the case in kiwi classes I recently observed an
exception from python saying:

```
File "/usr/lib64/python3.6/collections/__init__.py", line 429, in namedtuple
File "<string>", line 1, in <module>
ModuleNotFoundError: import of builtins halted; None in sys.modules
```

I do not understand why this is a problem for python now
and only found the solution in making the command_type variable
a global namedtuple.

### DIFF
--- a/kiwi/command.py
+++ b/kiwi/command.py
@@ -28,6 +28,10 @@ from .exceptions import (
     KiwiCommandNotFound
 )
 
+command_type = namedtuple(
+    'command', ['output', 'error', 'returncode']
+)
+
 
 class Command(object):
     """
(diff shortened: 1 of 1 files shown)


schaefi:
The above can be reproduce with this simple api script

```python
from kiwi.storage.loop_device import LoopDevice
from kiwi.filesystem import FileSystem

loop_provider = LoopDevice(
    '/home/ms/mydata.ext4', 100
)
loop_provider.create()
```

schaefi:
If somebody can explain this or has a better way to solve it I'm more than open to ideas :)

Thanks

PR #1150, opened 2019-07-23
Allow multiple values for grub terminal setup

### LINKED ISSUE #1123: Support for multiple entries in bootloader_console
## Problem description

ATM you can only choose between 

```bootloader_console="serial"```

or

```bootloader_console="console"```

or 

```bootloader_console="gfxterm"```

as documented in 

https://opensource.suse.com/kiwi/development/schema.html#id471

## Expected behaviour

In some cases a combination like

```bootloader_console="serial console"```

would be useful and seems to be supported (at least by grub2):

https://www.gnu.org/software/grub/manual/grub/grub.html#terminal_005finput

```
terminal_input [--append|--remove]  [terminal1] [terminal2]
...
terminal_output [--append|--remove] [terminal1] [terminal2]
```

schaefi:
With regards to the grub documentation from
https://www.gnu.org/software/grub/manual/grub/grub.html#terminal_005finput
multiple terminal consoles can be active. This patch allows
to specify terminal collection between serial, console and gfxterm
This Fixes #1123

### DIFF
--- a/build-tests/x86/test-image-fedora/appliance.kiwi
+++ b/build-tests/x86/test-image-fedora/appliance.kiwi
@@ -28,7 +28,7 @@
         <type image="iso" flags="overlay" firmware="uefi" hybridpersistent_filesystem="ext4" hybridpersistent="true"/>
     </preferences>
     <preferences profiles="Virtual">
-        <type image="vmx" filesystem="ext4" bootloader="grub2" kernelcmdline="console=ttyS0 splash"/>
+        <type image="vmx" filesystem="ext4" bootloader="grub2" kernelcmdline="console=ttyS0 splash" bootloader_console="serial console"/>
     </preferences>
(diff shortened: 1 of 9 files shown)


Conan-Kudo:
@schaefi Could we add a Fedora test that makes sure we have multi-terminal for grub configured and working?

schaefi:
yes this is possible. I updated the integration test for the Fedora Disk build with a multi console setup. The kernel already uses the serial console, so this image should have serial console data from the bootloader to the login

If the mutliconsole setup in grub is actually working could only be tested by running that image build. We don't do that at the moment. I have setup a few boot tests in kanku.suse.de but the service is not fully featured at the moment. In this system I test if I can login to the machine but if a multiconsole works is not easy to auto test imho

schaefi:
going to merge and check on the integration test results

PR #1154, opened 2019-07-26
Fixed left over import of six module

schaefi:
The use of the six compat module was needed to support py2
With the drop of py2 support all of six was no longer needed.
However this one was overlooked.


Conan-Kudo:
Looks good to me, merging!

PR #1155, opened 2019-07-26
99-kiwi-lib requires rmdir, install it

danimo:
Fixes bsc#1143033


schaefi:
Thanks, good catch :+1:

PR #1160, opened 2019-08-02
Align setup.py with the correct license

davidcassany:
License metadata for the setuptools package was not in line with the real license provided within kiwi. This PR fixes it.


davidcassany:
@Conan-Kudo now it should fine

Conan-Kudo:
@davidcassany We're also missing the `'license' : 'GPLv3+'` statement in the `config {}` block.

Issue #874, opened 2018-11-08
Support custom /etc/crypttab configuration


davidcassany:
## Problem description

Currently it is not possible to configure an encrypted disk (`luks="password"`) image that can be booted automatically having the key location configured in `/etc/crypttab`. A common use case for that would be having an encrypted disk that boots without user interaction having the key accessible in a removable device (e.g. a USB stick).

## Expected behaviour

It is possible to configure `/etc/crypttab`.

sveldhuisen:
That woud be quite useful. If I may suggest https://github.com/dracut-crypt-ssh/dracut-crypt-ssh also. It would be pretty neat if we could control this in combination with dropbear and Kiwi.

Conan-Kudo:
I'd rather we only leverage functionality present in the baseline dracut package. It gets tricky otherwise.

davidcassany:
> I'd rather we only leverage functionality present in the baseline dracut package. It gets tricky otherwise.

Fully agree, however I could not manage to get `rd.luks.key` and related luks dracut parameters working. So I still believe that some strategy to add some customizations for `/etc/crypttab` makes sense. Probably it changed meanwhile I last tried, but at the time I was having a look a it I really got the impression there was no other way than ignoring `rd.luks*` family for the time being.

davidcassany:
All right, I did some more debugging to this feature. I am still uncertain how we should proceed. First relevant thing I discovered is that dracut (upstream) luks support is not executed if systemd is present in initrd. The `rd.luks.*` parameters documented on dracut docs are just not handled by dracut if systemd dracut module is there (which is the case for KIWI images), instead systemd itself manages the luks device using this [generator](https://www.freedesktop.org/software/systemd/man/systemd-cryptsetup-generator.html). Also this generator is updated to support key files in external devices since last August, see the commit [here](https://github.com/systemd/systemd/commit/70f5f48eb89) (late August 2018). Note that the parameters syntax is not 100% equivalent to what is documented with the dracut crypt module.

Had look to Fedora 29 (that includes close to upstream dracut and systemd versions) but F29 still does not include a systemd version supporting external devices (systemd >= 240). Regarding Tumbleweed I noticed the systemd is up to date, but this not the case for dracut, resulting in dracut-crypt module adding some hooks that I have the feeling that they should not be there.... Gonna continue debugging a bit more, but to my eyes it starts to look like more than few changes in Tumbleweed's dracut might be required to get it working using kernel parameters.

[...]

PR #1162, opened 2019-08-08
Document that grub tweaks may be required for LUKS encryption

aspiers:
At least on SLES (and presumably openSUSE), the grub configuration needs `GRUB_ENABLE_CRYPTODISK=y` adding, so document this caveat and give an example of how to do it.

### DIFF
--- a/doc/source/working_with_kiwi/xml_description.rst
+++ b/doc/source/working_with_kiwi/xml_description.rst
@@ -332,7 +332,12 @@ build types and will be covered here:
 - `luks`: Supplying a value will trigger the encryption of the partitions
   using the LUKS extension and using the provided string as the
   password. Note that the password must be entered when booting the
-  appliance!
+  appliance!  You may also need to tweak the grub configuration, e.g.
+  by adding something like the following to `config.sh`:
+
+  .. code:: bash
+
+      echo "GRUB_ENABLE_CRYPTODISK=y" >>/etc/default/grub
 
 - `primary`: Boolean option, KIWI will by default build the image which
   `primary` attribute is set to `true`.


schaefi:
Thanks much for the doc update. To my understanding the GRUB_ENABLE_CRYPTODISK setting is required if /boot is inside of the luks. At the time I've implemented luks support grub was not able to read data from an encrypted filesystem. Therefore if you build a luks encrypted image with kiwi you get a non encrypted extra boot partition such that grub can boot from it.

The creation of the extra boot partition can be switched off with the ```bootpartition="false"``` attribute. As far as I understand grub now has support to read from a disk that has /boot on the encrypted filesystem which would be great.

Thus I suggest to evaluate if this is correct and let kiwi set the GRUB_ENABLE_CRYPTODISK variable.

I wonder how grub will ask for the luks credentials though ? I also wonder if we might get prompted twice in the boot process for the luks credentials, first for letting grub read its config and second for letting the system boot

I haven't done any investigations in this regard, so any further information you can share here are very much appreciated.

I think if we leave the user with the extra grub variable it will probably not help him much to understand the reason behind it.

Thoughts ?

aspiers:
Thanks for the reply.  The image I built this way did indeed prompt for the same credential twice - once from grub and once (presumably) just before mounting the encrypted root filesystem.  Obviously this is not a good user experience, but I didn't learn enough about the whole system yet to be able to confidently suggest a better approach; I just found something which works to MVP level.

So all I know is that it doesn't work with kiwi out of the box by only setting the `luks` attribute - if you do that then the build fails with an error about `GRUB_ENABLE_CRYPTODISK` not being set.  Changing kiwi to set that automatically would be an improvement, but something which avoids the double prompt for encryption key would be even better.

If you want to experiment with the image I built, it's here:

https://build.opensuse.org/package/show/home:aspiers:branches:SUSE:Templates:Images:SLE-15/kiwi-templates-SLES15-JeOS-Apache

schaefi:
Thanks for the details and the link to the build project, that makes it easier for me. I will take a look

dcermak:
Having to enter your password twice is a known problem when having an encrypted boot partition: grub needs your password to decrypt `/boot/` and load the kernel, but that needs it again to unlock your disks.

There is a relatively simple-ish workaround, by creating a keyfile for the root partition and including it in the initial ramdisk:
``` bash
dd bs=512 count=4 if=/dev/urandom of=/.root.keyfile
chmod 0400 /.root.keyfile
cryptsetup luksAddKey /dev/$ROOT_DEVICE_HERE /.root.keyfile
echo 'install_items+=" /.root.keyfile "' > /etc/dracut.conf.d/99-root-keyfile.conf
dracut
```
**Important:** you need to add the keyfile to `/etc/crypttab`, change the line that reads as:
```
my_root		UUID=something
```
to:
```
my_root		UUID=something		/.root.keyfile
```

Unfortunately, the `cryptsetup luksAddKey` is interactive, so this cannot be included exactly like this in KIWI. Probably related to: #874

schaefi:
Thanks for all the details. I have a branch open that addresses this issue properly. Meaning we will support encrypted boot partitions and solve the error

```
grub2-install: error: attempt to install to encrypted disk without cryptodisk enabled. Set `GRUB_ENABLE_CRYPTODISK=y' in file `/etc/default/grub'.
```

> Unfortunately, the cryptsetup luksAddKey is interactive

I found a way to add the key

schaefi:
I'll open a new pull request for review today or on Monday

aspiers:
That's great, thanks a lot!

schaefi:
ok I finished the implementation. I think this is what customers really want :)

Thanks much for the details, it's really nice to be able to provide full system encryption even with /boot included

schaefi:
I'm closing this one on behalf of #1168 

Let's continue further discussions there. Thanks much

PR #1167, opened 2019-08-15
remove '-z' option from rsync local copy calls

seife:
rsync's "compress" option just does not make any sense when rsync is
used to copy files locally, it only increases CPU usage and slows down
the process ;-)


schaefi:
ok, if you say it :)

Issue #1175, opened 2019-08-22
BootImageBase.is_prepared returns wrong type


dcermak:
## Problem description

`BootImageBase.is_prepared` ([kiwi/boot/image/base.py:153](https://github.com/OSInside/kiwi/blob/master/kiwi/boot/image/base.py#L153)) is documented to return a boolean, but it actually returns a list. Which part is correct, the documentation or the code?

This issue didn't get caught by the tests, because they all either mock `is_prepared` to return a boolean or they mock `os.listdir` to return a boolean (the latter is incorrect). Also, `if List` is a valid expression in python and checks whether the list is not empty, so maybe that is the actual intention of the code?

schaefi:
> Which part is correct, the documentation or the code?

The docs, it should behave like that. I will fix this. Also the test are really poor. We even mock the wrong return for the listdir. Shame on me

Issue #1128, opened 2019-07-07
Refactor unit tests


schaefi:
## Problem description

Unit tests were written in compat mode to py2. With regards to #1036 some limitations will be gone and that allows us to refactor the unit tests

schaefi:
__mock of open__

```python
@patch_open
def test_...(self, mock_open):
    mock_open.return_value = self.context_manager_mock
```

should be changed to

```python
import io
def test_...(self):
    with patch('builtins.open', create=True) as mock_open:
        mock_open.return_value = MagicMock(spec=io.IOBase)
        ...
```

schaefi:
Overall test structure should be changed to use directory tree instead of file names

```
test/unit/utils_rpm_test.py
```

should be changed to

```
test/unit/utils/rpm_test.py
```

schaefi:
__use of raises___

```python
@raises(KiwiTemplateError)
def test_...(self):
    ...
```

should be changed to

```python
from pytest import raises
def test_...(self):
    with raises(KiwiTemplateError):
        ...
```

schaefi:
__Cleanup of test/unit/test_helper.py__

Delete raises() and mock_open() methods. Also delete patch_open stub

schaefi:
raises refactoring is done now

the handling of open calls is the last bit in this refactoring. It should be part of the next sprint imho and therefore gain some higher prio

PR #1176, opened 2019-08-22
[experiment] Add type hints to some files

dcermak:
I've started adding type hints to kiwi and chose an arbitrary starting point: `kiwi/boot/*`. To make type hinting a little bit easier, I've dropped the usage of `namedtuple` and added explicit classes for this (once we drop Python 3.6 compatibility, we could start using dataclasses). Then I've added type hints to some classes in `kiwi/boot/` and `kiwi/system/kernel.py`.

This already identified #1175. Furthermore mypy now reports quite a few issues already:
```
kiwi/system/kernel.py:55: error: Missing return statement
kiwi/system/kernel.py:109: error: Missing return statement
kiwi/system/kernel.py:196: error: Missing return statement
kiwi/boot/image/base.py:166: error: Incompatible return value type (got "List[str]", expected "bool")
kiwi/boot/image/base.py:297: error: Missing return statement
kiwi/boot/image/dracut.py:115: error: Unsupported operand types for + ("List[Optional[str]]" and "List[str]")
kiwi/boot/image/dracut.py:121: error: Unsupported operand types for + ("List[Optional[str]]" and "List[str]")
kiwi/boot/image/dracut.py:123: error: Item "None" of "Optional[KernelInfo]" has no attribute "version"
kiwi/boot/image/dracut.py:132: error: List item 0 has incompatible type "Optional[str]"; expected "str"
kiwi/boot/image/dracut.py:185: error: List item 0 has incompatible type "Optional[str]"; expected "str"
```

The "Missing return statement" errors are caused by the following pattern that I've found in kiwi:
``` Python
def example_func():
    if cond:
        return bar
```
which gets flagged by mypy, because it expects **every** branch to return something.

Question is now: do we want to disable such a check or ignore it? I would be actually in favor of enforcing explicit returns everywhere, because that means that we **must** have test coverage for the cases where `None` is returned. And we clearly don't have that, as the other errors show.

Any opinions how to proceed or whether to proceed at all, as this is going to be quite the grind?

### DIFF
--- a/kiwi/boot/image/dracut.py
+++ b/kiwi/boot/image/dracut.py
@@ -17,7 +17,7 @@
 #
 import os
 import re
-from typing import List, NamedTuple
+from typing import Dict, List, NamedTuple, Optional
 
 # project
 from kiwi.logger import log
@@ -42,7 +42,16 @@ class BootImageDracut(BootImageBase):
     """
     **Implements creation of dracut boot(initrd) images.**
     """
-    def post_init(self):
+
+    dracut_options: List[str]
+    included_files: List[str]
+    included_files_install: List[str]
+    modules: List[str]
+    install_modules: List[str]
+    omit_modules: List[str]
+    omit_install_modules: List[str]
+
+    def post_init(self) -> None:
         """
         Post initialization method
 
@@ -56,7 +65,8 @@ class BootImageDracut(BootImageBase):
         self.omit_modules = []
         self.omit_install_modules = []
 
-    def include_file(self, filename, install_media=False):
+    def include_file(self, filename: str,
+                     install_media: bool = False) -> None:
         """
         Include file to dracut boot image
 
@@ -68,7 +78,7 @@ class BootImageDracut(BootImageBase):
             self.included_files_install.append('--install')
             self.included_files_install.append(filename)
 
-    def include_module(self, module, install_media=False):
+    def include_module(self, module: str, install_media: bool = False) -> None:
         """
         Include module to dracut boot image
 
@@ -80,7 +90,7 @@ class BootImageDracut(BootImageBase):
         elif module not in self.modules:
             self.modules.append(module)
 
-    def omit_module(self, module, install_media=False):
+    def omit_module(self, module: str, install_media: bool = False) -> None:
         """
         Omit module to dracut boot image
 
@@ -92,7 +102,9 @@ class BootImageDracut(BootImageBase):
         elif module not in self.omit_modules:
             self.omit_modules.append(module)
 
-    def write_system_config_file(self, config, config_file=None):
+    def write_system_config_file(
+            self, config: Dict[str, str], config_file: Optional[str] = None
+    ) -> None:
         """
         Writes modules configuration into a dracut configuration file.
 
@@ -104,6 +116,7 @@ class BootImageDracut(BootImageBase):
             config_file = os.path.normpath(
                 self.boot_root_directory + Defaults.get_dracut_conf_name()
             )
+        assert config_file
         if config.get('modules'):
             dracut_config.append(
                 'add_dracutmodules+=" {0} "\n'.format(
@@ -117,10 +130,10 @@ class BootImageDracut(BootImageBase):
                 )
             )
         if dracut_config:
-            with open(config_file, 'w') as config:
-                config.writelines(dracut_config)
+            with open(config_file, 'w') as config_fd:
+                config_fd.writelines(dracut_config)
 
-    def prepare(self):
+    def prepare(self) -> None:
         """
         Prepare dracut caller environment
 
@@ -138,7 +151,8 @@ class BootImageDracut(BootImageBase):
         self.dracut_options.append('--install')
         self.dracut_options.append('/.profile')
 
-    def create_initrd(self, mbrid=None, basename=None, install_initrd=False):
+    def create_initrd(self, mbrid: None = None, basename: str = None,
+                      install_initrd: bool = False) -> None:
         """
         Call dracut as chroot operation to create the initrd and move
         the result into the image build target directory
@@ -203,7 +217,7 @@ class BootImageDracut(BootImageBase):
                 [self.target_dir, dracut_initrd_basename]
             )
 
-    def get_boot_names(self):
+    def get_boot_names(self) -> BootNames:
         """
         Provides kernel and initrd names for kiwi boot image
 
@@ -236,7 +250,7 @@ class BootImageDracut(BootImageBase):
             )
         )
 
-    def _get_dracut_output_file_format(self):
+    def _get_dracut_output_file_format(self) -> str:
         """
         Unfortunately the dracut initrd output file format varies between
         the different Linux distributions. Tools like lsinitrd, and also
--- a/kiwi/boot/image/dracut.py
+++ b/kiwi/boot/image/dracut.py
@@ -17,7 +17,7 @@
 #
 import os
 import re
-from collections import namedtuple
+from typing import List, NamedTuple
 
 # project
 from kiwi.logger import log
@@ -30,6 +30,12 @@ from kiwi.system.setup import SystemSetup
 from kiwi.path import Path
 
 from kiwi.exceptions import KiwiDiskBootImageError
+
+
+class BootNames(NamedTuple):
+    """Container for the kernel and initrd name."""
+    kernel_name: str
+    initrd_name: str
 
 
 class BootImageDracut(BootImageBase):
@@ -202,20 +208,17 @@ class BootImageDracut(BootImageBase):
         Provides kernel and initrd names for kiwi boot image
 
         :return:
-            Contains boot_names_type tuple
+            Contains BootNames
 
             .. code:: python
 
-                boot_names_type(
+                BootNames(
                     kernel_name='INSTALLED_KERNEL',
                     initrd_name='DRACUT_OUTPUT_NAME'
                 )
 
-        :rtype: tuple
+        :rtype: BootNames
         """
-        boot_names_type = namedtuple(
-            'boot_names_type', ['kernel_name', 'initrd_name']
-        )
         kernel = Kernel(
             self.boot_root_directory
         )
@@ -226,7 +229,7 @@ class BootImageDracut(BootImageBase):
                 self.boot_root_directory
             )
         dracut_output_format = self._get_dracut_output_file_format()
-        return boot_names_type(
+        return BootNames(
             kernel_name=kernel_info.name,
             initrd_name=dracut_output_format.format(
                 kernel_version=kernel_info.version
--- a/kiwi/system/kernel.py
+++ b/kiwi/system/kernel.py
@@ -47,11 +47,13 @@ class Kernel:
         file so that we do not have to search for many different
         names in this code
     """
-    def __init__(self, root_dir):
+    def __init__(self, root_dir: str) -> None:
         self.root_dir = root_dir
-        self.kernel_names = self._setup_kernel_names_for_lookup()
+        self.kernel_names: List[str] = self._setup_kernel_names_for_lookup()
 
-    def get_kernel(self, raise_on_not_found=False):
+    def get_kernel(
+            self, raise_on_not_found: bool = False
+    ) -> Optional[KernelInfo]:
         """
         Lookup kernel files and provide filename and version
 
@@ -103,7 +105,7 @@ class Kernel:
                 )
             )
 
-    def get_xen_hypervisor(self):
+    def get_xen_hypervisor(self) -> Optional[Xen]:
         """
         Lookup xen hypervisor and provide filename and hypervisor name
 
@@ -118,7 +120,8 @@ class Kernel:
                 name='xen.gz'
             )
 
-    def copy_kernel(self, target_dir, file_name=None):
+    def copy_kernel(self, target_dir: str,
+                    file_name: Optional[str] = None) -> None:
         """
         Copy kernel to specified target
 
@@ -137,7 +140,8 @@ class Kernel:
             )
             Command.run(['cp', kernel.filename, target_file])
 
-    def copy_xen_hypervisor(self, target_dir, file_name=None):
+    def copy_xen_hypervisor(self, target_dir: str,
+                            file_name: Optional[str] = None) -> None:
         """
         Copy xen hypervisor to specified target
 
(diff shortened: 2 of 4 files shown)


kiwi/boot/image/dracut.py
```diff
@@ -32,11 +32,26 @@
 from kiwi.exceptions import KiwiDiskBootImageError
 
 
+class BootNames(NamedTuple):
+    """Container for the kernel and initrd name."""
+    kernel_name: str
+    initrd_name: str
```
schaefi: I don't understand the advantage of a namedtuple return vs. a class that inherits from namedtuple and returning class variables that matches the tuple names. Is it not possible to write:

```python
def get_boot_names(self) -> namedtuple:
```

kiwi/boot/image/dracut.py
```diff
@@ -50,7 +65,8 @@ def post_init(self):
         self.omit_modules = []
         self.omit_install_modules = []
 
-    def include_file(self, filename, install_media=False):
+    def include_file(self, filename: str,
+                     install_media: bool = False) -> None:
```
schaefi: Can we keep the indentation style we use all over the place

```python
def include_file(
    self, filename: str, install_media: bool = False
) -> None:
```

kiwi/boot/image/dracut.py
```diff
@@ -74,7 +90,7 @@ def include_module(self, module, install_media=False):
         elif module not in self.modules:
             self.modules.append(module)
 
-    def omit_module(self, module, install_media=False):
+    def omit_module(self, module: str, install_media: bool = False) -> None:
```
schaefi: I wonder why we want to classify the return of a method that returns nothing with an explicit "None" type.
I see where you coming from and I was also into C and C++ for some time. Also in that language I found it cumbersome to explicitly state "void" when it's obvious that no return type definitions means we don't return
anything. Does this provide as any benefit ? meaning if an api method is used that is not expected to be called with an evaluated return value, does python complain before ? e.g

```python
a = omit_module(...)
```

would python complain before running the code if the method definition says "-> None" ? If not I don't see any value ?

kiwi/boot/image/dracut.py
```diff
@@ -98,6 +116,7 @@ def write_system_config_file(self, config, config_file=None):
             config_file = os.path.normpath(
                 self.boot_root_directory + Defaults.get_dracut_conf_name()
             )
+        assert config_file
```
schaefi: why the assert ?

Actually in any of the changes here the behavior of the code should stay untouched

kiwi/boot/image/dracut.py
```diff
@@ -111,10 +130,10 @@ def write_system_config_file(self, config, config_file=None):
                 )
             )
         if dracut_config:
-            with open(config_file, 'w') as config:
-                config.writelines(dracut_config)
+            with open(config_file, 'w') as config_fd:
+                config_fd.writelines(dracut_config)
```
schaefi: Same here, that change is imho not related to type hints ?

kiwi/boot/image/dracut.py
```diff
@@ -226,14 +243,14 @@ def get_boot_names(self):
...
-        return boot_names_type(
+        return BootNames(
             kernel_name=kernel_info.name,
             initrd_name=dracut_output_format.format(
                 kernel_version=kernel_info.version
             )
         )
 
-    def _get_dracut_output_file_format(self):
+    def _get_dracut_output_file_format(self) -> str:
```
schaefi: All type hinting we add I think should be done only for the public interface. That's the part where I think it would make sense. People using kiwi as an api and running our methods benefit from type checking. Private methods are not considered to be called and we are free to change them whenever we want. Type hinting there imho just makes it more complex

kiwi/system/kernel.py
```diff
@@ -16,14 +16,27 @@
...
+    filename: str
+    name: str
+
+
+class KernelInfo(NamedTuple):
+    """Storage for files related to a specific kernel version."""
+    name: str
+    filename: str
+    version: str
+
```
schaefi: Same here, I don't understand the motivation for the extra classes

kiwi/boot/image/dracut.py
dcermak: A `namedtuple` can be absolutely anything and returning one contradicts the point of type checks. The idea is that a static type checker (like mypy or pytype) is able to tell you that this is going to fail:
``` python
boot_names = image.get_boot_names()
boot_names.this_does_not_exist = 3  # whoops AttributeError is raised here
```
and it can only know that if you specify explicitly which elements your namedtuple contains.

Furthermore, specifying this explicitly is imho a good thing, because currently we are only specifying the elements of this namedtuple in the docstring, which doesn't cause any kind of failure when someone changes the namedtuple and not the docstring.

kiwi/boot/image/dracut.py
dcermak: Python itself won't complain, because python doesn't care about type hints. What cares about them is additional tooling (like mypy).

The idea of `func() -> None:` is to be the equivalent of `void func();` in C. I.e. this function doesn't return anything and saving it's return value is thus most likely a bug. If you only return `None` in some cases, then the return type should be `Option[T]` (and then mypy will also complain if you just assume that this value is always not `None`).

kiwi/boot/image/dracut.py
dcermak: This is a mypy issue: it isn't able to infer that `config_file` is never `None`. The `assert` is essentially only needed as a type hint, which should be ok, because that will be removed when byte compiling this file anyway.

kiwi/boot/image/dracut.py
dcermak: It is, `config` is declared as `Dict[str, str]` but here we are assigning a `FileIO` to it, which is a violation of static types. That's why I changed the variable name.

kiwi/boot/image/dracut.py
dcermak: > Private methods are not considered to be called and we are free to change them whenever we want.

Exactly. And if we change the signature, then we should get a warning/error when a caller of that function violates the function's contract (e.g. the return type), which type hints will provide us with. That's why I've added them to private functions too, not because I like to grind :wink:

schaefi: ok got it, looking from a developer perspective make sense

kiwi/boot/image/dracut.py
schaefi: I agree makes sense, also you have a point here with the documentation. We would need to change this at some places in the code because I used namedtuple a lot mostly because the values are static(read-only). So many return types makes use of it

kiwi/boot/image/dracut.py
schaefi: If we agree to base the type hints on py >= 3.6 as discussed a bit on the rio chat, would this still apply ?

kiwi/boot/image/dracut.py
schaefi: If we agree to base the type hints on py >= 3.6 as discussed a bit on the rio chat, would this still apply ? I think we should not add code to make mypy happy, thoughts ?

kiwi/boot/image/dracut.py
schaefi: Oh I see, yes makes sense.

kiwi/system/kernel.py
schaefi: Is clear now, thanks

schaefi:
@davidcassany your feedback would be much appreciated.

After collecting all the information I think it would be good to add the type hints if we could agree to the following constraints:

* done for python >= 3.6
* no requirement to be mypy compatible
* no extra code only done to please mypy
* implementation in a feature branch, we can use this one
* implementation considered complete before merge, no merge of intermediate state
* no type hints needed for unit tests

@Conan-Kudo @dcermak @davidcassany: Does this sound ok for everybody ? I know the patch will be very big but if we follow the above it should be mergable into any master state changes we do until done. From there we could enhance/extend the type hinting implementation when needed.

Thoughts ?

davidcassany:
@schaefi, sorry, I provided some thoughts on chat yesterday, should have done it here.

Yes I am fine with this approach. I agree that we should not code to facilitate mypy analysis, however we should make explicit how are we validating the annotations. Shouldn't we enforce mypy to provide a clean output and assume we might add `# type: ignore` on certain lines?

Just wondering how are we going to validate the completeness of the task, ideally this should be a validation done on Travis, this way we also enforce proper annotation in further code.

I am also in favor of not touching anything on unit tests, as this is currently our ground truth that the changes are not having undesired impact. Probably this a way to draw a line on what **no extra code only done to please mypy** means, if refactoring the unit test is needed you should validate the changes in code are really needed first.

dcermak:
> @davidcassany your feedback would be much appreciated.
> 
> After collecting all the information I think it would be good to add the type hints if we could agree to the following constraints:
> 
>     * done for python >= 3.6
> 
>     * no requirement to be mypy compatible
> 
>     * no extra code only done to please mypy

This is going to be a problem, as mypy currently complains **a lot** about the current state of the code base. For instance it does not like that certain functions only explicitly return in certain branches and that we sometimes simply assume that values that are `Optional[T]` are never `None` (the former is more of a stile thing, but the latter are actual bugs).

The issue with this is: we need mypy or pytype to verify that the type hints are correct, otherwise they will become outdated like the documentation. But if mypy/pytype errors out on KIWI, then we cannot use it, so we need to change the code to use these tools unfortunately.

Would it be ok if I introduce these code changes in separate pull requests?

> 
>     * implementation in a feature branch, we can use this one
> 
>     * implementation considered complete before merge, no merge of intermediate state
> 
>     * no type hints needed for unit tests
> 

What do you mean by that? The unit tests don't require type hints.

> 
> @Conan-Kudo @dcermak @davidcassany: Does this sound ok for everybody ? I know the patch will be very big but if we follow the above it should be mergable into any master state changes we do until done. From there we could enhance/extend the type hinting implementation when needed.
> 
> Thoughts ?

kiwi/boot/image/dracut.py
dcermak: Yes, a function that returns nothing (= equivalent to `void func();` in C) should be marked as `func() -> None:` for Python type hints.

schaefi:
> What do you mean by that? The unit tests don't require type hints.

I mean let us handle the implementation first and let us skip the unit testing code. In case type hints would make sense for the unit test code this should come later. Especially not before #1128 is done

schaefi:
> This is going to be a problem, as mypy currently complains a lot about the current state of the code base. 

referencing the changes in this PR there is one that add an assertion because of an issue in mypy. This is the type of code changes I don't want to have in kiwi. I agree to what you have said about other complains mypy might have but I don't want to add code in kiwi to workaround issues in mypy

schaefi:
> Would it be ok if I introduce these code changes in separate pull requests?

I'm thinking about it, but honestly adding code with no meaning just to workaround an issue in another tool was one of the things I fight harder against in the next generation code base of kiwi. We had tons of those workarounds in the legacy kiwi code base just because a patch at the right place was not possible for questionable reasons

dcermak:
> I mean let us handle the implementation first and let us skip the unit testing code. In case type hints would make sense for the unit test code this should come later. Especially not before #1128 is done

I would not add type hints to unit tests. What one can do is to run mypy over the unit tests to verify that mocks are doing the right thing (TM).

> referencing the changes in this PR there is one that add an assertion because of an issue in mypy. This is the type of code changes I don't want to have in kiwi. I agree to what you have said about other complains mypy might have but I don't want to add code in kiwi to workaround issues in mypy

Hm, to a certain extend these are workarounds due to bugs/missing features in mypy/pytype. In theory we can report all these in upstream to mypy & pytype and hope that they'll fix them, but I wouldn't get my hopes up, as some of these are pretty tough to solve.

> I'm thinking about it, but honestly adding code with no meaning just to workaround an issue in another tool was one of the things I fight harder against in the next generation code base of kiwi. We had tons of those workarounds in the legacy kiwi code base just because a patch at the right place was not possible for questionable reasons

Well, I hope that the actual code changes will be small, stuff like adding an explicit `return None`, explicit return types and maybe an assertion here and there (keep in mind that assertions are automatically removed from python bytecode, so there wouldn't even be a runtime difference).

Unfortunately, I cannot really estimate how many changes will be required to make mypy & pytype happy. I am especially worried about the automatically generated xml wrapper `kiwi/xml_parse.py`, as pytype currently completely chokes on it and mypy afaik didn't like it either. But maybe we can make `generateDS.py` to output type hints too...

davidcassany:
> I am especially worried about the automatically generated xml wrapper kiwi/xml_parse.py, as pytype currently completely chokes on it and mypy afaik didn't like it either. But maybe we can make generateDS.py to output type hints too...

I'd say `kiwi/xml_parse.py` is out of the scope we should not pretend to have any sort of type hints there IMHO. `kiwi/xml_state` is our interface to the generated code and this is the class we should add some type hints for the returning values of its methods and assume there won be validation between the interaction of `kiwi/parse_xml.py` and `kiwi/xml_state.py`. Obviously this might extend to other classes since xml_state is used in code to access to it's xml_data object.

In case ignoring type annotations for some code parts is creating a bunch of hard to handle issues in mypy I have no clue on how to proceed. I have limited experience on mypy, but I'd expect to be capable to set mypy to ignore certain modules.

schaefi:
> I'd say kiwi/xml_parse.py is out of the scope we should not pretend to have any sort of type hints here

Yes fully agree. The auto created code part is also out of scope for the unit testing. The api generated by generateDS takes the schema as source. The schema itself is explicit in types such that wrong input there is covered from the schema validation and the schematron rules. There is no need for us to work with the auto generated code at this point.

schaefi:
I hope mypy has a config file or something which we can place here that allows to skip certain source codings ?

tomschr:
> I hope mypy has a config file or something which we can place here that allows to skip certain source codings ?

Sorry to interrupt the party. Yes, it's possible to do that in the config.

You can even use a `# type: ignore` line. See the section [Spurious errors and locally silencing the checker](https://mypy.readthedocs.io/en/latest/common_issues.html#spurious-errors-and-locally-silencing-the-checker). :wink:

dcermak:
I must confess that I completely forgot about this PR, sorry folks!

Unfortunately, I **really** don't see myself driving this monstrous effort forward in the foreseeable future. So unless you don't mind having this open for a long, long time, please consider closing this PR.

schaefi:
@dcermak fair enough and thanks for letting us know. I agree it's a big change and should not be in the hands of one guy. At the moment I also don't see that we can put this on David's and my plate since there are other tasks that needs to be addressed first.

I'm leaning towards closing this and reconsider it on demand of the community or with help from e.g the doc team. Their plate is naturally also full so it might stay as an idea for some time.
I'll put the description of what we want to achieve on the idea project here in the projects tab

Thanks for your time on this even if we don't get it done at this point

PR #1177, opened 2019-08-22
Stop inheriting from object

dcermak:
Mass change of `class Foo(object):` to `class Foo:` as you no longer have to inherit from `object` in Python3.


davidcassany:
Looks good to me too :+1: In those cases is were we can really take advantage of a full coverage in tests :)

Issue #873, opened 2018-11-08
Refactor the use of BootImage for installation initrds


davidcassany:
## Problem description

In OEM images if some installation media is required at least a couple of initrds are created. One the the OEM image itself and another one for the installation image. Those two initrds are created using the same instance of BootImage while the scope is not the same. The code related to that is a candidate for a refactor, it could be done in a simpler or clearer way. Currently this part is hard to follow and, in consequence, hard or potentially hard to maintain in the future.

Related with #858

davidcassany:
I am starting to have look at this one. Unless I see a really straight forward implementation I will only document my thoughts and suggestions here.

davidcassany:
@schaefi I am going to provide a PR to discuss it I really thing there is no longer the oemboot problem since we already dropped oemboot descriptions in OSInside/kiwi-descriptions@64ddd90d. So I am wondering if we should keep the door opened for any custom oemboot code out there since we are not maintaining it and it might silently break.

schaefi:
sounds great, looking forward to it :+1:

davidcassany:
Some trials on the test builds revealed my suspicions, with or without the patch custom boot on encrypted luks disk does not boot. Without luks oemboot code works fine. I am unsure what should we do from now on with this refactor, should we try to investigate whats going on for oemboot on luks? Or just assume the refactor is not breaking anything new?

On dracut all seams to work fine.

schaefi:
Since we switched to dracut some years ago I expected that the custom oemboot code will be broken because the distro changes and nobody is testing or using oemboot anymore. Actually the only reference to a custom oemboot description lives in our integration tests and I'd like to keep the simple oem disk case to work as kind of a reference implementation if somebody wants to use this feature. However I expect this to be irrelevant anyway because we have dracut and that also supports custom extensions. So from my perspective it's more important to refactor the kiwi code such that the opportunity to build custom (non dracut based) initrds is still possible and works per integration test. If sophisticated layouts like LUKS needs special adaptions to a non dracut created initrd this should not be a priority anymore in my opinion.

As you found out LUKS+oemboot seems to be broken anyway and nobody is complaining I would leave it that way and offer help to customers when they stumble over it... which I think will never happen because we just use dracut.

> Or just assume the refactor is not breaking anything new?

if we can make that sure it would be ideal. I think what still has to work is the ```test-image-disk-legacy``` integration test.
It builds the most simple disk layout and install iso and we use this as a reference.

What do you think ?

[...]

PR #1179, opened 2019-08-23
Avoid default installation of kiwi dracut modules

### LINKED ISSUE #1103: Manual install of kiwi dracut packages breaks the system
## Problem description

__From QA__

In reference to: https://bugzilla.suse.com/show_bug.cgi?id=1136444

During some of our (maintenance QA testing) it occured that people installed dracut-kiwi-* on testing system. This triggers regeneration of initrd, which is then not usable for booting system (and errors received are far from helpfull).

Situation when user, in good faith can install package, which without any warning can render system unusable seems rather suboptimal.

davidcassany:
This PR fixes #1103

### DIFF
--- a/dracut/modules.d/90kiwi-dump/kiwi-ramdisk-deployment-generator.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-ramdisk-deployment-generator.sh
@@ -11,6 +11,8 @@ root_uuid=$(
     grep root= | cut -f2- -d=
 )
 
+[ -z "${root_uuid}" ] && exit 1
+
 {
     echo "[Unit]"
     echo "Before=initrd-root-fs.target"
--- a/kiwi/boot/image/dracut.py
+++ b/kiwi/boot/image/dracut.py
@@ -45,6 +45,10 @@ class BootImageDracut(BootImageBase):
         self.dracut_options = []
         self.included_files = []
         self.included_files_install = []
+        self.modules = []
+        self.install_modules = []
+        self.omit_modules = []
+        self.omit_install_modules = []
 
     def include_file(self, filename, install_media=False):
         """
@@ -57,6 +61,58 @@ class BootImageDracut(BootImageBase):
         if install_media:
             self.included_files_install.append('--install')
             self.included_files_install.append(filename)
+
+    def include_module(self, module, install_media=False):
+        """
+        Include module to dracut boot image
+
+        :param string module: module to include
+        :param bool install_media: include the module for install initrds
+        """
+        if install_media and module not in self.install_modules:
+            self.install_modules.append(module)
+        elif module not in self.modules:
+            self.modules.append(module)
+
+    def omit_module(self, module, install_media=False):
+        """
+        Omit module to dracut boot image
+
+        :param string module: module to omit
+        :param bool install_media: omit the module for install initrds
+        """
+        if install_media and module not in self.omit_install_modules:
+            self.omit_install_modules.append(module)
+        elif module not in self.omit_modules:
+            self.omit_modules.append(module)
+
+    def write_system_config_file(self, config, config_file=None):
+        """
+        Writes modules configuration into a dracut configuration file.
+
+        :param dict config: a dictionary containing the modules to add and omit
+        :param string conf_file: configuration file to write
+        """
+        dracut_config = []
+        if not config_file:
+            config_file = os.path.normpath(
+                self.boot_root_directory + Defaults.get_dracut_conf_name()
+            )
+        if config.get('modules'):
+            dracut_config.append(
+                'add_dracutmodules+=" {0} "\n'.format(
+                    ' '.join(config['modules'])
+                )
+            )
+        if config.get('omit_modules'):
+            dracut_config.append(
+                'omit_dracutmodules+=" {0} "\n'.format(
+                    ' '.join(config['omit_modules'])
+                )
+            )
+        if dracut_config:
+            with open(config_file, 'w') as config:
+                config.writelines(dracut_config)
 
     def prepare(self):
         """
@@ -95,9 +151,25 @@ class BootImageDracut(BootImageBase):
                 dracut_initrd_basename = self.initrd_base_name
             if install_initrd:
                 included_files = self.included_files_install
+                modules_args = [
+                    '--add', ' {0} '.format(' '.join(self.install_modules))
+                ] if self.install_modules else []
+                omit_modules_args = [
+                    '--omit', ' {0} '.format(
+                        ' '.join(self.omit_install_modules)
+                    )
+                ] if self.omit_install_modules else []
             else:
                 included_files = self.included_files
+                modules_args = [
+                    '--add', ' {0} '.format(' '.join(self.modules))
+                ] if self.modules else []
+                omit_modules_args = [
+                    '--omit', ' {0} '.format(' '.join(self.omit_modules))
+                ] if self.omit_install_modules else []
             dracut_initrd_basename += '.xz'
+            options = self.dracut_options + modules_args +\
+                omit_modules_args + included_files
             dracut_call = Command.run(
                 [
                     'chroot', self.boot_root_directory,
@@ -105,7 +177,7 @@ class BootImageDracut(BootImageBase):
                     '--no-hostonly',
                     '--no-hostonly-cmdline',
                     '--xz'
-                ] + self.dracut_options + included_files + [
+                ] + options + [
                     dracut_initrd_basename,
                     kernel_details.version
                 ],
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -373,14 +373,18 @@ class DiskBuilder:
         self._write_generic_fstab_to_system_image(device_map)
 
         if self.initrd_system == 'dracut':
-            self._create_dracut_config()
+            if self.root_filesystem_is_multipath is False:
+                self.boot_image.omit_module('multipath')
+            if self.root_filesystem_is_overlay:
+                self.boot_image.include_module('kiwi-overlay')
+                self.boot_image.write_system_config_file(
+                    config={'modules': ['kiwi-overlay']}
+                )
+            if self.build_type_name == 'oem':
+                self.boot_image.include_module('kiwi-repart')
 
         # create initrd cpio archive
         self.boot_image.create_initrd(self.mbrid)
-
-        # create dracut config omitting one time kiwi dracut modules
-        if self.initrd_system == 'dracut':
-            self._create_system_dracut_config()
 
         # create second stage metadata to system image
         self._copy_first_boot_files_to_system_image()
@@ -768,55 +772,6 @@ class DiskBuilder:
         self.disk.map_partitions()
 
         return self.disk.get_device()
-
-    def _create_dracut_config(self):
-        dracut_config = [
-            'hostonly="no"',
-            'dracut_rescue_image="no"'
-        ]
-        dracut_modules = []
-        dracut_modules_omit = ['kiwi-live', 'kiwi-dump']
-        if self.root_filesystem_is_multipath is False:
-            dracut_modules_omit.append('multipath')
-        if self.root_filesystem_is_overlay:
-            dracut_modules.append('kiwi-overlay')
-        else:
-            dracut_modules_omit.append('kiwi-overlay')
-        if self.build_type_name == 'oem':
-            dracut_modules.append('kiwi-lib')
-            dracut_modules.append('kiwi-repart')
-        self._write_dracut_config(
-            config=dracut_config,
-            modules=dracut_modules,
-            omit_modules=dracut_modules_omit
-        )
-
-    def _create_system_dracut_config(self):
-        dracut_modules = []
-        dracut_modules_omit = ['kiwi-live', 'kiwi-dump', 'kiwi-repart']
-        if self.root_filesystem_is_overlay:
-            dracut_modules.append('kiwi-overlay')
-        else:
-            dracut_modules_omit.append('kiwi-overlay')
-        self._write_dracut_config(
-            config=[], modules=dracut_modules, omit_modules=dracut_modules_omit
-        )
-
-    def _write_dracut_config(self, config, modules, omit_modules):
-        dracut_config_file = ''.join(
-            [self.root_dir, Defaults.get_dracut_conf_name()]
-        )
-        if modules:
-            config.append(
-                'add_dracutmodules+=" {0} "'.format(' '.join(modules))
-            )
-        if omit_modules:
-            config.append(
-                'omit_dracutmodules+=" {0} "'.format(' '.join(omit_modules))
-            )
-        with open(dracut_config_file, 'w') as dracut_config:
-            for entry in config:
-                dracut_config.write(entry + os.linesep)
 
     def _write_partition_id_config_to_boot_image(self):
         log.info('Creating config.partids in boot system')
--- a/test/unit/builder_disk_test.py
+++ b/test/unit/builder_disk_test.py
@@ -462,8 +462,6 @@ class TestDiskBuilder:
         assert mock_open.call_args_list == [
             call('boot_dir/config.partids', 'w'),
             call('root_dir/boot/mbrid', 'w'),
-            call('root_dir/etc/dracut.conf.d/02-kiwi.conf', 'w'),
-            call('root_dir/etc/dracut.conf.d/02-kiwi.conf', 'w'),
             call('boot_dir/config.bootoptions', 'w'),
             call('/dev/some-loop', 'wb')
         ]
@@ -471,13 +469,6 @@ class TestDiskBuilder:
             call('kiwi_BootPart="1"\n'),
             call('kiwi_RootPart="1"\n'),
             call('0x0f0f0f0f\n'),
-            call('hostonly="no"\n'),
-            call('dracut_rescue_image="no"\n'),
-            # before dracut is called, image dracut setup
-            call('add_dracutmodules+=" kiwi-lib kiwi-repart "\n'),
-            call('omit_dracutmodules+=" kiwi-live kiwi-dump multipath kiwi-overlay "\n'),
-            # after dracut was called, system dracut setup
-            call('omit_dracutmodules+=" kiwi-live kiwi-dump kiwi-repart kiwi-overlay "\n'),
             call('boot_cmdline\n'),
             call(bytes(b'\x0f\x0f\x0f\x0f'))
         ]
@@ -495,6 +486,11 @@ class TestDiskBuilder:
             call('/config.partids'),
             call('/recovery.partition.size')
         ]
+        self.boot_image_task.include_module.assert_called_once_with(
+            'kiwi-repart'
+        )
+        self.boot_image_task.omit_module.assert_called_once_with('multipath')
+        assert self.boot_image_task.write_system_config_file.call_args_list == []
 
     @patch('kiwi.builder.disk.FileSystem')
     @patch('kiwi.builder.disk.FileSystemSquashFs')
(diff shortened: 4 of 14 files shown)


davidcassany:
Regarding the dracut modules update. All my tests I used quite a brute force and I added ALL modules as part of the rootfs-tree overlay. So the resulting images had ALL modules in `/usr/lib/dracut/modules` those are the tests I did:

* OEM disk deployment:
    1. deploy to disk and boot
    2. on booted system verify there is no kiwi config in `/etc/dracut.conf.d/`
    3. run `dracut -f` and reboot

* OEM PXE deployment on persistent disk:
    1. PXE deploy to disk and boot
    2. on booted system verify there is no kiwi config in `/etc/dracut.conf.d/`
    3. run `dracut -f` and reboot

* OEM PXE deployment on ramdisk:
    1. PXE deploy to disk and boot
    2. on booted system verify there is no kiwi config in `/etc/dracut.conf.d/`

* OEM ISO deployment on persistent disk:
    1. ISO deploy to disk and boot
    2. on booted system verify there is no kiwi config in `/etc/dracut.conf.d/`
    3. run `dracut -f` and reboot

* OEM ISO deployment on ramdisk:
    1. ISO deploy to disk and boot
    2. on booted system verify there is no kiwi config in `/etc/dracut.conf.d/`

* VMX having all modules as part of the root-tree
    1. Verify boot
    2. run `dracut -f` and reboot

* ISO deployed on a persistent device:
    1. dd to a device and boot
    2. verify the existence of `/etc/dracut.conf.d/02-livecd.conf`
    3. run `dracut -f` and reboot

* Boot ISO as CDROM on KVM
 
All were successful :smiley:

davidcassany:
@schaefi JFYI I found something really strange while refactoring the builder_disk_test.py tests. In fact I had to patch `sys.argv` (you can see it in the code) as for some unknown reason the `raises` helper we have in `test/uni/test_heper.py` was swallowing a DocoptExit exception raised on `Defaults.get_shared_cache_location()`. I am even wondering if this method is correct from an API POV, as it requires a full set of consistent arguments according to the docopt rules.

schaefi:
I saw we avoid setting the omit_dracutmodules because the change in the check() method now requires us to explicitly state which modules to include. However there are other places that omits modules. See the code in install.py and live.py. I think those needs the same adaptions you did in disk.py

test/unit/builder_disk_test.py
```diff
@@ -212,23 +220,23 @@ def test_setup_ix86(self, mock_machine):
         )
         assert disk_builder.arch == 'ix86'
 
-    @raises(KiwiInstallMediaError)
     @patch('kiwi.builder.disk.FileSystem')
-    @patch_open
+    @patch('builtins.open')
```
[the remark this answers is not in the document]
davidcassany: Well, I can certainly do that, I even started doing it this way but I was facing a huge tests with all the code wrapped and intended inside the `with` clause, causing a massive diff and many breaks of the 80 lines max length, making the diff uglier, so that I opted for an in place replacement.

Is there any specific motivation to not patch the `builtins.open` as a decorator? Just willing to properly understand the refactor

dracut/modules.d/90kiwi-dump/kiwi-ramdisk-deployment-generator.sh
```diff
@@ -11,6 +11,8 @@ root_uuid=$(
     grep root= | cut -f2- -d=
 )
 
+[ -z "${root_uuid}" ] && exit 1
+
```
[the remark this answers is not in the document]
davidcassany: > under which circumstances can it happen that we don't have a root variable in config.bootoptions ? or is this sanity check just added for convenience ?

There shouldn't be any circumstance were we see this situation. However the insane boot I was debugging that triggered all this code was getting to this situation because of the missing crypt dependency in `kiwi-lib` (we already merged a fix for that in #1174) was causing an unmanaged script failure before even validating the existence of `config.bootoptions`. Dracut was never complaining of a missing config.bootoptions and the generator produced a wrong unit-file. This wrong unit-file causes systemd to fail on switch-root complaining of a worng unit file format.

I don't see any way now to get to a missing `config.bootoptions`. And my tests playing the stupidity role always ended up with a proper dialog with a `config.bootoptions` file missing message and a reboot button.

I opted to have this sanity check since the generator is overwriting the unit file generated by systemd, thus in the worst case that would be used. In bsc#1142899 they attempted a similar patch that _solved_ (log were still showing many failures of kiwi dracut modules but system was booting) the issue.

test/unit/builder_disk_test.py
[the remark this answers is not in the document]
schaefi: yes there is. The decorator patches the open for the complete test. However the test should be written in a way that only the code that issues an open is covered in a with() block. This allows to write tests and even use real open for e.g reading mock data that should be a return value for data read. Also if code that opens different files or files multiple times can be covered much better in a with() block

It's clear to me that the test file will change massive but that's part of the refactoring we agreed on. So I'm willing to accept this. Regarding the 80c line length I'm sure there will be a way to reformat that such that it fits in and probably become more readable.

dracut/modules.d/90kiwi-dump/kiwi-ramdisk-deployment-generator.sh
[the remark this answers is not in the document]
schaefi: Ah ok, thanks that makes sense :+1:

davidcassany:
> I saw we avoid setting the omit_dracutmodules because the change in the check() method now requires us to explicitly state which modules to include. However there are other places that omits modules. See the code in install.py and live.py. I think those needs the same adaptions you did in disk.py

I already did and I am surprised that we had that part already coded in the way we needed. Let me show you:

https://github.com/OSInside/kiwi/blob/53903534d398ce211f8eb37ae16c4094e16c9e98/kiwi/builder/install.py#L455-L468

In install.py we are already not including any dracut configuration inside the root-tree (the same way I changed in disk.py).

And in live.py we are adding the `/etc/dracut.conf.d/02-livecd.conf`

https://github.com/OSInside/kiwi/blob/53903534d398ce211f8eb37ae16c4094e16c9e98/kiwi/builder/live.py#L296-L311

I am fine keeping this file here because this image could be deployed on a persistent device with the overlay partition and we want to still reuse the same former dracut configuration to ensure the dracut-kiwi-live is used in case `dracut -f` is called.

We could argue if the `omit_dracutmodules+=` part is still needed in all cases, since we are returning 255 on `check()` for each module this shouldn't be necessary, but the less changes the better and the more security omitting the right modules the better IMHO. So that I decided to not change a single character of the dracut configuration files we use to create the intird. This part remains untouched, I only had got rid of the `/etc/dracut.conf.d/02-kiwi.conf` that was included in the final image to overwrite the buildtime dracut configuration.

davidcassany:
@scheafi I am clearly aware that this is a very sensitive change. I really appreciate you being nit-picky here, thanks.

schaefi:
>In install.py we are already not including any dracut configuration inside the root-tree (the same way >I changed in disk.py).

yes because it's a one time environment and the config was only needed for creating the install initrd. You should always have in mind the dracut calls is issued from the system root tree. In contrast to what we had with the custom kiwi initrd there is no extra environment for each initrd. It's all the same. The initrd to drive the install is an add-on of the process for creating the oem system image. That's why the initrd for the install creates the config, runs dracut, deletes the config. For the system however we kept the config. This part will now change.

For the user this means he will get a new hostonly initrd with the first invocation of dracut in the running system. e.g on kernel update. In most cases this is probably wanted and so I'm fine with the way we do it right now

Regarding the omit statements: It's ok to keep them but I think it's not needed. We are on the hook to change the concept of the kiwi dracut modules in a way that the installation of them are noops. Meaning having them installed doesn't lead to any activation. This means _any_ dracut call in kiwi has to explicitly state which kiwi modules should be included. For me that means no omit statement is required anymore but an explicit list of include statements. I didn't see that change to be applied for the live and install builders. Thus I asked. I think if we change the concept it should be done completely.

I understand the risk of change here but we are already beyond that line of fire with the current changes. In any case a merge of this requires extensive manual testing of:

* live image
* oem install (dump and repart)

Thoughts ?

davidcassany:
@schaefi I removed the second commit of this PR that was refactoring the builder_disk test. Since this PR is changing sensitive parts of the dracut modules I like the idea of keeping this PR diff as small as possible. I'll do a follow UP PR including the builder_disk tests refactor according to your suggestions.

davidcassany:
> Regarding the omit statements: It's ok to keep them but I think it's not needed. We are on the hook to change the concept of the kiwi dracut modules in a way that the installation of them are noops. Meaning having them installed doesn't lead to any activation. This means _any_ dracut call in kiwi has to explicitly state which kiwi modules should be included. For me that means no omit statement is required anymore but an explicit list of include statements. I didn't see that change to be applied for the live and install builders. Thus I asked. I think if we change the concept it should be done completely.

Ok fine with removing the omit statements. As far as I understand you suggest including the required dracut modules as part of the dracut call, this was my initial idea, then realized we are already adding them via a temporary configuration file before each dracut call, so I kept this strategy. Gonna check if it can be done all from the command line and get rid of all dracut configuration files.

schaefi:
I'm not suggesting that we get rid of the dracut conf files. What I'm saying is that we are moving from implicit to explicit. Before this change the install of the module also activated it. Thus no reason to explicitly tell dracut to include it. Now we have to tell it to dracut which is nice. I don't have a strong opinion to do it in a file or via the commandline. But in the current code I don't see where we explicitly add the required dracut modules for the call for e.g the live builder or the install builder

schaefi:
From a code perspective I would expect something like this

* disk_builder

```
# no kiwi modules needed
dracut_vmx_modules = []

# for oem we need repart and optionally overlay
dracut_oem_modules = ['kiwi-repart']
    dracut_oem_modules.append('kiwi-overlay')

```

* live builder

```
we need Defaults.get_live_dracut_module_from_flag(self.live_type)
```

That looks already good

* install builder

```
# for oem install
dracut_oem_install_modules = ['kiwi-dump']
```

I think we also don't need to explicitly include kiwi-lib as it will be a requirement of the modules we use

At the end I would not expect any omit statement. If that then makes the call easy enough to go only with commandline params we can simplify the dracut call and have them explicitly with commandline parameters.

If we still need other options it might be good to keep the file approach

schaefi:
Sorry I'm really really nitpick about changes here. But this is the most critical part and I don't want to screw it up. Thanks for your understanding

schaefi:
Please also have in mind that I use the dracut conf files to include other files like the profile env and other oem metadata relevant for the resize. So probably we should keep the file approach

davidcassany:
> Please also have in mind that I use the dracut conf files to include other files like the profile env and other oem metadata relevant for the resize. So probably we should keep the file approach

Yes I got to the same conclusion at the time I was having a look to it. The dracut configuration files were already adding the required modules (even it was not really needed, that was my surprise). I am mostly keeping the build time configuration file around mostly because there is this configuration `dracut_rescue_image="no"` which can't be passed from the command line. So far the profile env and other kiwi specific files are included directly on the command line.

davidcassany:
rebased on master to fix the gitlab pipe

kiwi/builder/disk.py
```diff
@@ -793,14 +790,17 @@ def _create_dracut_config(self):
 
     def _create_system_dracut_config(self):
```
schaefi: From my perspective the entire method can be deleted.

The _create_system_dracut_config only existed to create a dracut conf that omits the kiwi modules. Now that the kiwi modules are installed as noops that function is no longer needed

kiwi/builder/disk.py
```diff
@@ -793,14 +790,17 @@ def _create_dracut_config(self):
...
+                omit_modules=dracut_modules_omit
+            )
         else:
-            dracut_modules_omit.append('kiwi-overlay')
-        self._write_dracut_config(
-            config=[], modules=dracut_modules, omit_modules=dracut_modules_omit
-        )
+            Path.wipe(''.join(
+                [self.root_dir, Defaults.get_dracut_conf_name()]
+            ))
```
schaefi: I think deleting the dracut.conf is wrong. Actually the file is only written if the overlay module is in use. In case of the overlay module it has to stay active. If it's not taken into account again on e.g kernel-update the system will be broken. Our motivation should be to write the dracut config when needed and then it should also be persistently needed for the live time of the image

davidcassany: I am not sure I follow this comment. With the code written here the dracut configuration is always deleted unless there is a root overlay, for which we need the kiwi-overlay module. I can improve the code above to make it more explicit. But the result in terms of the root-tree remains the same. After the build we need a persistent dracut configuration for the case we have a root overlay and for the case of live image (`builder/live.py` does not remove the dracut configuration). 

In that latter case I doubt it is strictly needed as I doubt the initrd can be updated, but I definitely prefer to keep it around since this would be the correct configuration in case initrd could be updated.

davidcassany:
@schaefi I believe this is matching your concerns in terms of behavior and I believe the code simple to read.

I'd like to ask you, again, for a review at that point. I believe now the dracut configuration files management is quite clear and readable. I did several tests and all were green, just had some issues with the kiwi-overlay module but those issues were not introduced by these changes. I reproduced them in current master.

I'd fix the overlay module issue in a separate follow up PR.

davidcassany:
@scheafi the Travis issues I believe/hope are just some network issues on Travis side.

schaefi:
Hmm, ok from a processing perspective it's all clear now. But I think I didn't made my point clear.
With this pull request we are fundamentally changing the way the kiwi dracut modules are used. I like the change but the implementation has still room for improvement. Given all dracut modules provided from the kiwi scope are noops we should only create dracut config files when they are required for the image type and setup. In any other case I vote for caller parameters. Thus my first step would be a change on the class interface:

* kiwi/boot/image/base.py

```python
def include_module(self, ...):
    ...
```
This would lead to dracut being called to include the given module. The method should be allowed to be called several times to include more modules. The same concept has we have with include_file. For the builtin_kiwi.py this method just passes

With that in place let's go through the builders: disk.py, live.py, install.py. I think we are good with live image and install image builder because those needs extra and permanent dracut config files. The most changes I expect in disk.py. There is already a variable that accumulates the modules. Let's use that to pass on include_module instead of the, from my perspective, weird clear_dracut_conf method. At this point I would expect no omit statements and no config file creation anymore. Then there is the overlay setup which requires a dracut config file. This case shold then just be a simple if condition that creates the dracut config.

In the end our log file will explicitly show the included dracut modules at call time. All dracut config files we create are there any time and not created/deleted/modified.

Let's keep it simple

And sorry again for me being so nitpick here. All this will strike back horrible if we mess it up

Let me know if this makes sense or if I should come up with a poc

Thanks

davidcassany:
> Given all dracut modules provided from the kiwi scope are noops we should only create dracut config files when they are required for the image type and setup. In any other case I vote for caller parameters. Thus my first step would be a change on the class interface:
> 
>     * kiwi/boot/image/base.py
> 
> 
> ```python
> def include_module(self, ...):
>     ...
> ```
> 
> This would lead to dracut being called to include the given module. The method should be allowed to be called several times to include more modules. The same concept has we have with include_file. For the builtin_kiwi.py this method just passes

I am fine with it, I just kept using the dracut configuration files because I did not find a way to set `dracut_rescue_image="no"` from the command line.

>  At this point I would expect no omit statements and no config file creation anymore. Then there is the overlay setup which requires a dracut config file. This case shold then just be a simple if condition that creates the dracut config.

Yes the only part that is causing issues here is the already mentioned parameter we use at build time that is not possible to set it from the command line (or at least I could not find a clear alternative). I'll proceed with the implementation as you suggest and move dracut modules configuration to some class method. Then we can discuss what to do the leftovers, as they will also be pretty obvious then.

Thanks for you comments

schaefi:
I suggest to make dracut_rescue_image=no one of the standard options as you can see them in boot/image/dracut.py

schaefi:
Actually I believe it has no effect anyway. Searching for dracut_rescue_image in the sources doesn't match anything. A generic interface to pass on options to the initrd creation tool would be generally useful. But I would suggest to do this in a next step

davidcassany:
@schaefi ok, now I did a second round of refactoring. Now modules are set from to dracut on the command line using boot/image/dracut.py class toolkit. I believe this matches your thoughts. I had to add clear_modules_config method I don't like because of #873. The boot image instance is being reused between disk and install builds.

Currently I am doing some tests to see if the current implementation works.

kiwi/boot/image/dracut.py
```diff
@@ -58,6 +62,75 @@ def include_file(self, filename, install_media=False):
...
+    def clear_modules_config(self):
+        """
+        Clears the modules configuration
+        """
+        self.modules = []
+        self.modules_args = []
+        self.omit_modules = []
+        self.omit_modiles_args = []
+
+    def persistent_config(self, config=None, config_file=None):
```
schaefi: The name of the method is imho not telling what action is behind the call. I suggest something like

* write_system_config
* create_system_config
* create_persistent_config
* write_system_config_file

Something in that direction. Personally I like write_system_config_file()

kiwi/boot/image/dracut.py
```diff
@@ -58,6 +62,75 @@ def include_file(self, filename, install_media=False):
...
+        """
+        Writes modules configuration into a dracut configuration file.
+
+        If modules are not provided it uses the self modules attributes.
```
schaefi: Please don't let us do that. One of the primers of object oriented code is: write code for a single responsibility. The task of the method is to write a config file from given parameter information. Trying to be clever and use different information if no module data is provided extends the responsibility of the method a lot and makes it complex. The number of if statements in a code block is always a good indicator

Additionally the method is never called with empty modules. Thus you wrote code that is currently only used in the test. This is also something that should be avoided. I suggest to make the config parameter mandatory:

```
def write_system_config_file(self, config_params, config_file=None):
    ...
```

kiwi/boot/image/dracut.py
```diff
@@ -58,6 +62,75 @@ def include_file(self, filename, install_media=False):
...
+        Contrary, if modules are provided as arguments it only uses the
+        provided ones.
+
+        :param dict config: a dictionary containing the modules to add and omit
+        :param string conf_file: configuration file to write
+        """
+        dracut_config = []
+        if config is None:
```
schaefi: python evaluates empty dicts as well as None type values as False. You can safely write

```
if not config:
```

schaefi:
> I believe this matches your thoughts.

yes that's great

> I had to add clear_modules_config method I don't like because of #873. The boot image
> instance is being reused between disk and install builds.

yes I saw it and also realized that must be the reason for the method. I think you don't need the
method. See the include_file() method. There is a parameter: install_media. So we already differentiate between files included in the system and the install initrd. You can take the same concept for include and omit modules. At the time of the refactoring it will then be easier for us since all data is handled the same way

kiwi/builder/disk.py
davidcassany: removed

kiwi/builder/disk.py
davidcassany: done

kiwi/boot/image/dracut.py
davidcassany: renamed to write_system_config_file

kiwi/boot/image/dracut.py
davidcassany: config dictionary is mandatory now

davidcassany:
Damn! Travis is failing again on network issues :sob: 

@schaefi I addressed your comments and included the changes. Now I am performing some tests, up to now (basic OEM and ISO) everything looks good. Gonna run some kiwi-overlay and pxe dump tests though.

kiwi/boot/image/dracut.py
```diff
@@ -58,6 +62,58 @@ def include_file(self, filename, install_media=False):
...
+
+        :param dict config: a dictionary containing the modules to add and omit
+        :param string conf_file: configuration file to write
+        """
+        dracut_config = []
+        if not config_file:
+            config_file = os.path.normpath(
+                self.boot_root_directory + Defaults.get_dracut_conf_name()
+            )
+        if 'modules' in config and config['modules']:
```
schaefi: can be done simpler:

```
if config.get('modules'):
```

kiwi/boot/image/dracut.py
```diff
@@ -58,6 +62,58 @@ def include_file(self, filename, install_media=False):
...
+            dracut_config.append(
+                'add_dracutmodules+=" {0} "\n'.format(
+                    ' '.join(config['modules'])
+                )
+            )
+        if 'omit_modules' in config and config['omit_modules']:
```
schaefi: same here

kiwi/boot/image/dracut.py
davidcassany: nice!

schaefi:
I leave it up to you to merge once you have finished your sanity checks and builds

davidcassany:
> Perfect, thanks a ton and also thanks for being patient with me being picky +1

Sure no problem, this is sort of changes are very sensitive and should be deeply discussed and reviewed :smile:

davidcassany:
All right finally green on Travis too :+1: Doing my final PXE test before pressing the merge button

PR #1180, opened 2019-08-23
Fix GitLab CI jobs for building on Fedora

Conan-Kudo:
Mock is incorrectly using systemd sysusers now without a systemd dependency or
working sysusers scriptlets. For now, manually create the mock group.

In addition, let's go ahead and upgrade to a Fedora 30 chroot for the mock builds.


schaefi:
Hmm, someone cuts the edge we sit on.

PR #1182, opened 2019-08-26
Add support for custom fstab script extension

schaefi:
In addition to fstab append and patch features we also allow
an fstab.script file that is called chrooted. The change is
needed to support overlay mounting of filesystems as part
of the initrd. If system filesystems needs to be changed in
a way that they can be used in an overlay mount, the standard
mount entry has to take the x-initrd.mount capability which
requires a modification of the fstab which is cumbersome to
handle as a patch file. This concept is currently used as
part of the MicroOS project in SUSE and is applied in the
integration test build maintained for this target. This
Fixes bsc#1129566

### DIFF
--- a/build-tests/x86/test-image-MicroOS/appliance.kiwi
+++ b/build-tests/x86/test-image-MicroOS/appliance.kiwi
@@ -84,6 +84,7 @@
         <package name="kernel-default"/>
         <package name="shim"/>
         <package name="timezone"/>
+        <package name="read-only-root-fs"/>
     </packages>
     <packages type="bootstrap">
         <package name="udev"/>
--- a/build-tests/x86/test-image-MicroOS/config.sh
+++ b/build-tests/x86/test-image-MicroOS/config.sh
@@ -32,9 +32,9 @@ echo "Configure image: [$kiwi_iname]..."
 suseSetupProduct
 
 #======================================
-# Create Host Keys
+# Install overlay systemd mount props
 #--------------------------------------
-/usr/sbin/sshd-gen-keys-start
+cp /usr/sbin/setup-fstab-for-overlayfs /etc/fstab.script
 
 #======================================
 # Activate services
--- a/doc/source/building/working_with_images/custom_fstab_extension.rst
+++ b/doc/source/building/working_with_images/custom_fstab_extension.rst
@@ -42,25 +42,31 @@ The optimal way to provide custom fstab information is through a
 package. If this can't be done the files can also be provided via
 the overlay file tree of the image description.
 
-KIWI supports two ways to modify the contents of the `/etc/fstab`
+KIWI supports three ways to modify the contents of the `/etc/fstab`
 file:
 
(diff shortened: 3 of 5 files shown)


schaefi:
Travis has network problems

schaefi:
make sense, I missed the docs. I will update that. Also wanted to wait for Feedback from Vogtinator

Vogtinator:
Should work for our use-case, but I wonder whether this is the best way.

Can't you just run a script like `config.sh` chrooted after the /etc/fstab is generated? Something like images.sh, just later.

schaefi:
> Can't you just run a script like config.sh chrooted after the /etc/fstab is generated? Something like
> images.sh, just later.

nope, sorry we don't wanted to introduce yet another custom script layer. We already struggled adding it for this purpose but couldn't come up with a better idea as all the other ways (adding, patching) were not enough for your use case. I know people can abuse this script layer for something else but as I said this is most preferable to support your use case and not to introduce another script layer.

Actually your image design works quite nicely since a few days on my test machine. Nice job on the MicroOS side. Hope you find some customers in the near future. Is there any investment to position MicroOS vs. CoreOS ? I'm just interested

From that perspective I think I hear you are ok with the change ?

Vogtinator:
> > Can't you just run a script like config.sh chrooted after the /etc/fstab is generated? Something like
> > images.sh, just later.
> 
> nope, sorry we don't wanted to introduce yet another custom script layer. We already struggled adding it for this purpose but couldn't come up with a better idea as all the other ways (adding, patching) were not enough for your use case. I know people can abuse this script layer for something else but as I said this is most preferable to support your use case and not to introduce another script layer.

I don't quite get your argumentation here, as I don't really see a difference between adding an `fstab.script` which is in the image's `/etc` directory and having a script which is part of the image description. The latter is cleaner as it doesn't need to be put into the image just to be deleted later.

> Actually your image design works quite nicely since a few days on my test machine. Nice job on the MicroOS side. 

I'm glad to hear that, what did you use it for?

> Hope you find some customers in the near future. Is there any investment to position MicroOS vs. CoreOS ? I'm just interested

That's basically what MicroOS is already 

> From that perspective I think I hear you are ok with the change ?

As it works for this use-case, yes.

schaefi:
> I'm glad to hear that, what did you use it for?

I kept it running for a while and let it create and delete junks of data on that part of the system that is overlayed using the overlayfs filesystem. I wanted to find out how disk space behave and if the overlay is stable. So far it still works :)

> I don't quite get your argumentation here, as I don't really see a difference between adding an
> fstab.script which is in the image's /etc directory and having a script which is part of the image
> description. 

From a technical perspective you are right. From the perspective of the one that creates the image description if you have to create a script in /etc and have to name it fstab.script and probably read the docs about what it should be used for it might hold you back from using it for something else than handling fstab. If we just allow yet another script on the image description level beside config.sh and images.sh it would increase the gate to do "things" at another level of the image creation process. We don't want to allow more of those "uncontrolled" code during the build process.

PR #1191, opened 2019-08-30
Fixed user permission fix

### LINKED ISSUE #1190: userpermissions are not fixed in config.sh
## Problem description
Calling config.sh in the preparation does not fix user permissions.

## Expected behaviour
User permissions are set correctly

Create a file, not belonging to the user and put it to root/home/$USERNAME/

+ KIWI version: 9.18.10
+ Operating system: SUSE Leap 15.0

jpwidera:
The comand was evaluated as chmod -c -R '...' . This is not the correct
syntax. Removing quotation marks solved problem.

Fixes #1190  .

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -263,7 +263,7 @@ function baseSetupUserPermissions {
         fi
         if [[ ! "${shell}" =~ nologin|true|false ]];then
             group=$(grep "${group}" /etc/group | cut -f1 -d:)
-            chown -c -R "${usern}:${group} ${dir}/*"
+            chown -c -R ${usern}:${group} ${dir}/*
         fi
     done < /etc/passwd
 }
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -263,7 +263,7 @@ function baseSetupUserPermissions {
         fi
         if [[ ! "${shell}" =~ nologin|true|false ]];then
             group=$(grep "${group}" /etc/group | cut -f1 -d:)
-            chown -c -R ${usern}:${group} ${dir}/*
+            chown -c -R "${usern}":"${group}" "${dir}"/*
         fi
     done < /etc/passwd
 }
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -263,7 +263,7 @@ function baseSetupUserPermissions {
         fi
         if [[ ! "${shell}" =~ nologin|true|false ]];then
             group=$(grep "${group}" /etc/group | cut -f1 -d:)
-            chown -c -R "${usern}":"${group}" "${dir}"/*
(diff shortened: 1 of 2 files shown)


kiwi/config/functions.sh
```diff
@@ -263,7 +263,7 @@ function baseSetupUserPermissions {
         fi
         if [[ ! "${shell}" =~ nologin|true|false ]];then
             group=$(grep "${group}" /etc/group | cut -f1 -d:)
-            chown -c -R "${usern}:${group} ${dir}/*"
+            chown -c -R ${usern}:${group} ${dir}/*
```
aplanas: Spaces can be present here. I would suggest `chown -c -R "{usern}:${group}" "${dir}"/*`

jpwidera: Good point

kiwi/config/functions.sh
```diff
@@ -263,7 +263,7 @@ function baseSetupUserPermissions {
         fi
         if [[ ! "${shell}" =~ nologin|true|false ]];then
             group=$(grep "${group}" /etc/group | cut -f1 -d:)
-            chown -c -R "${usern}":"${group}" "${dir}"/*
+            chown -c -R "${usern}":"${group}" "${dir}"
```
aplanas: For me makes sense, but if we decided so we need to update the docs:
https://github.com/OSInside/kiwi/blob/master/doc/source/working_with_kiwi/shell_scripts.rst

jpwidera: > Search all home directories of all users listed in :file:`/etc/passwd` and change the ownership of all files to belong to the correct user and group

I think this is more a philosophical question. Does /home/X belong to user X or /home/X/* ? My installation of SUSE Leap 15.0 makes ownership of /home/X to 755. In practice it belongs to user X and the script changes the ownership towards the user...

aplanas: When the user is created, useradd takes care of fixing the ownership and permission (755) of the directory, and copy the data from /etc/skel with the correct ownership + permissions.

The issue about the documentation is a detail, as is referring about changing the content and not the /home/xxx directory. For me ${dir} or ${dir}/* makes sense, but the doc is referring to the second instance, not the first one. So if you like to change the behavior, you need to update the documentation.

I personally do not see any advantage / disadvantage on any of both approach.

jpwidera: I hope I understood it correctly: It is referring only to the **content** of /home/xxx only and should also refer to the homedirectory itself? See change

schaefi: Seems like there is an ongoing discussion. Labeled the PR as such

@aplanas could you answer the question ? Thanks

schaefi: iirc the doc and code change are now well aligned

@davidcassany can you have a short look too. For me the change looks good

Issue #1181, opened 2019-08-26
(psuedo-)soft raid-1 is not recognized during OEM installation


jpwidera:
## Problem description
During the deployment of type.image="oem", no raid devices are recognized in the "Select Installation Disk" menu. There, only the physical /dev/disk/by-[path|uuid|label] are found. Albeit, booting live-system with identical setup shows raid-devices.
I do not want to create a new raid but using an existing one.

Maybe the dracut-image does not include the necessary kernel modules or udev is not loaded correctly?

It is possible to write data to any of the raid-members but this causes the raid to degrade.

## Expected behaviour
I would expect some raid entries like /dev/disk/by-id/md-uuid-*

## Steps to reproduce the behaviour
Use one of the openSUSE oem-build-tests with firmware="efi" and install them on a computer with (pseudo)-raid-device. In this case: Intel Rapid Storage Technology.

## OS and Software information
+ KIWI version: 9.18.10
+ Operating system: openSUSE Leap 15.0

schaefi:
If no raid devs are detected it might be because of missing packages. For softraid support you need to have the following packages installed in your kiwi XML description

```xml
<package name="dmraid"/>
<package name="mdadm"/>
```

if dracut is now called by kiwi it should integrate all needed components such that the
raid devs will be found. In addition you should also activate the raid detection via commandline
in your XML description

```
<type .... kernelcmdline="rd.auto"/>
```

The list of devices to show in the dialog in kiwi is build from the following command

```
lsblk -p -n -r -o NAME,SIZE,TYPE
```

Every entry from that list that is a disk will be taken into account unless filtered
out by some oem rule you might have set in the XML description

If you still don't see your device I suggest to boot with

```
rd.debug rd.break=pre-mount rd.kiwi.debug
```

This should drop you into a shell in dracut and you can check if your devices are present or not

Hope that helps

jpwidera:
Getting into debug mode helped a lot, but still no working solution.

Maybe the loading of kernel modules dm* and md* is causing the problem: 
When booting the working live-system I find the /dev/md126 as well as the modules md_mod + raid1.
Booting the OEM installer in debug mode does not show md_mod and raid1, instead dm_mirror is loaded but I can load md_mod and raid1 manually (But this does not create the /dev/md126-device). 
I'm not sure but I guess both kernel modules are trying to do the same and conflicting.

I could not prevent the loading of the dm-modules since blacklisting the dm via 
`rd.dm=0`
 or 
[...]

PR #1195, opened 2019-09-10
Detect raid compound in root disk detection

schaefi:
Make sure raid managed storage e.g fake raid is detected
using the correct md device. Related to Issue #1181

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-lib.sh
@@ -30,9 +30,9 @@ function get_swap_map {
 
 function lookup_disk_device_from_root {
     declare root=${root}
+    declare kiwi_RaidDev=${kiwi_RaidDev}
     local root_device=${root#block:}
     local disk_device
-    local disk_matches=0
     local wwn
     if [ -z "${root_device}" ];then
         die "No root device found"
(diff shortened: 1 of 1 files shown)


schaefi:
This is untested code and needs approval by someone with a fake raid controller at hand.

dracut/modules.d/99kiwi-lib/kiwi-lib.sh
```diff
@@ -42,22 +42,37 @@ function lookup_disk_device_from_root {
...
+                fi
+            done
+        fi
+        # Check if root_device is managed by mdadm. If this
+        # is the case prefer the md device over the storage
+        # disks because this would create conflicts otherwise
+        if type mdadm &> /dev/null; then
+            for raid_disk in $(
+                mdadm --detail -Y "${disk_device}" 2>/dev/null |\
+                grep -E "MD_DEVICE_.*_DEV" | cut -f2 -d=
```
jpwidera: Sincerly, on the Intel Rapid Storage device, the last entry in $disk_device is /dev/sda, which can't be handled by mdadm. Should be /dev/md*

schaefi: mdadm --detail is called on any disk found. the ones that does not provide md information are not taken into account. The loop only runs if there is md information available and takes this md device and returns. I think if you read not by diff but by the complete code you will realized how things work

schaefi: after all you should always have in mind this code must be generic enough for any disk situation including multipath, so we can't just mess it up or make it only for for raid-disks ;)

jpwidera: Oh, excuse my rushing - I didn't see the do-loop changed.
Applying the code block gives at:
```
 mdadm --detail -Y "${disk_device}" 2>/dev/null |\
                grep -E "MD_DEVICE_.*_DEV" | cut -f2 -d=
```
`/dev/sda` and `/dev/sdb` as results.

Maybe it is easier to check, if 
```
mdadm --detail -Y "${disk_device}" 2>/dev/null
```
returns 0 (disk_device=/dev/sd* returns 1). In this case it is a valid md-disk.

schaefi: yes you are right that's much easier. I've updated the code. Thanks

dracut/modules.d/99kiwi-lib/kiwi-lib.sh
```diff
@@ -42,22 +41,31 @@ function lookup_disk_device_from_root {
...
+                    echo "${disk_device}"
+                    return
+                fi
+            done
+        fi
+        # Check if root_device is managed by mdadm. If this
+        # is the case prefer the md device over the storage
+        # disks because this would create conflicts otherwise
+        if type mdadm &> /dev/null; then
+            if mdadm --detail -Y "${disk_device}" 2>/dev/null; then
```
jpwidera: This is good. Maybe a 
```
>/dev/null 2>&1
```
prevents the output from spoiling with unneeded information about the md-device (Additionally, the warning message, that any device isn't manageable by mdadm is stdout).

schaefi: yep I changed it to &>/dev/null

jpwidera:
@schaefi Thx, your fix does solve the problem (Tested on a fake-raid OEM-deployment).

schaefi:
> I'll come back to this next week.

Thank you, much appreciated. I will merge this part next and submit a new package build to our staging area. You will find new packages at:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Staging/python-kiwi

It would be great if you can build your image with the kiwi from that staging area and run the test with rd.kiwi.debug enabled

Again thanks for your help we wouldn't be there without you testing on real fakeraid devices

jpwidera:
@schaefi 
I used the latest build on an empty directory. I can select and install a complete system into a fake-raid-device. And the fake-raid partitions are beeing resized.

Sadly, other problems came up with this (Not booting into kernel with kexec after deployment, and missing fake-raid-support in the deployed system) but **this** implementation works :)

> Once the system is up can you please send me the file /run/initramfs/log/boot.kiwi
> 
> I'd like to verify the partition resize of the raid. Thanks
> 
The system freezes after deployment, so I'm not able to get a log-file. I'll check this and go for a new issue

PR #1200, opened 2019-09-16
Fix GPT to MBR convertion

### LINKED ISSUE #1169: Fail to convert GPT to MBR
## Problem description

When I use  `force_mbr="true"`, kiwi fails to build the image as the following:

```
[  679s] [ INFO    ]: 06:20:59 | --> converting partition table to MBR
[  679s] [ DEBUG   ]: 06:20:59 | EXEC: [sgdisk -i=1 /dev/loop0]
[  680s] [ DEBUG   ]: 06:20:59 | EXEC: [sgdisk -i=2 /dev/loop0]
[  680s] [ DEBUG   ]: 06:20:59 | EXEC: [sgdisk -m 1:2 /dev/loop0]
[  681s] [ DEBUG   ]: 06:21:00 | EXEC: [sgdisk -t 1:8300 /dev/loop0]
[  681s] [ DEBUG   ]: 06:21:00 | EXEC: Failed with stderr: (no output on stderr), stdout: 
[  681s] ***************************************************************
[  681s] Found invalid GPT and valid MBR; converting MBR to GPT format
[  681s] in memory. 
[  681s] ***************************************************************
```

## Expected behaviour

Image is created.

Use `<type force_mbr="true">` like the following:

https://build.opensuse.org/package/view_file/home:matwey:branches:openSUSE:Factory:ARM:Live/JeOS/JeOS-zturn.kiwi?expand=1

+ KIWI version: 9.18.10
+ Operating system: openSUSE:Factory:ARM
+ OBS version: 2.11~alpha.20190815T153919.f9a03b0a07

I suppose that the issue is in `sgdisk -t 1:8300 /dev/loop0` command. I've checked manually, and it refuses to modify partition type on MBR based disk images.

davidcassany:
This commit swaps the order of the command in gpt to mbr convertion
in partitioner.gpt.set_mbr method.

sgdisk by default converts the partition table from MBR to GPT in memory.
The change is never applied unless you provide the -g option forcing to
overwrite the partition table format. If sgdisk does the convertion in
memory and the user does not provide the -g flag it returns an error
code.

The solution applied in this commit is to just run the GPT to MBR
convertion as the last sgdisk command and change partition type before
running the convertion.

Fixes #1169

### DIFF
--- a/kiwi/partitioner/gpt.py
+++ b/kiwi/partitioner/gpt.py
@@ -138,12 +138,12 @@ class PartitionerGpt(PartitionerBase):
             if '(EFI System)' in partition_info.output:
                 efi_partition_number = number
             partition_ids.append(format(number))
-        Command.run(
-            ['sgdisk', '-m', ':'.join(partition_ids), self.disk_device]
-        )
         if efi_partition_number:
             # turn former EFI partition into standard linux partition
             self.set_flag(efi_partition_number, 't.linux')
+        Command.run(
+            ['sgdisk', '-m', ':'.join(partition_ids), self.disk_device]
+        )
 
     def resize_table(self, entries=128):
         """
--- a/test/unit/partitioner_gpt_test.py
+++ b/test/unit/partitioner_gpt_test.py
@@ -97,8 +97,8 @@ class TestPartitionerGpt:
             call(['sgdisk', '-i=2', '/dev/loop0']),
             call(['sgdisk', '-i=3', '/dev/loop0']),
             call(['sgdisk', '-i=4', '/dev/loop0']),
-            call(['sgdisk', '-m', '1:2:3:4', '/dev/loop0']),
-            call(['sgdisk', '-t', '4:8300', '/dev/loop0'])
+            call(['sgdisk', '-t', '4:8300', '/dev/loop0']),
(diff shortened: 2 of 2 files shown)


davidcassany:
@matwey with the current fix the resulting layout of an image including a boot partition and EFI firmware results into something like:

```
Disk /dev/loop0: 1.2 GiB, 1327497216 bytes, 2592768 sectors
Disklabel type: dos
Disk identifier: 0xe757afc1

Device       Boot Start     End Sectors  Size Id Type
/dev/loop0p1       2048    6143    4096    2M ef EFI (FAT-12/16/32)
/dev/loop0p2       6144   47103   40960   20M 83 Linux
/dev/loop0p3      47104 2592734 2545631  1.2G 83 Linux
```

Does this matches your expectations? I am mostly wondering about the `ef` partition for the hybrid MBR, in general I'd say this doesn't affect the result on x86, but on ARM... not sure.

schaefi:
I'm merging this. The disklabel is dos the partitions are correctly aligned and the partition type ids are also correct. 

>  I am mostly wondering about the ef partition for the hybrid MBR,

After the conversion this results in a more less superfluous primary partition with no further meaning. The conversion itself only makes sense under certain circumstances on e.g arm where a msdos table is required but should look like an efi layout.

So all you did is fine, I'm merging it

PR #1204, opened 2019-09-18
Make kiwi-overlay module compatible with kiwi-repart

### LINKED ISSUE #1184: kiwi-overlay issues (testing vmx and oem with overlayroot="true")
## Problem description

Found some issues with overlayroot.

1. Building a disk image with overlayroot causes a squashed file system creation of the root-tree twice. The first one is just used to estimate the size of the squashed partition, the second one is the one producing the squashed system that will actually be part of the disk. This procedure is waste of CPU cycles. **Nice to have, it is not actually causing any functional issue**

2. ~~If the disk is not configured with a boot partition the overall build is likely to fail.~~ fixed in #1188

3. ~~We are missing the use of `-wildcards` flag to properly exclude some files (e.g. `/boot/*`).~~ fixed in #1188

4. kiwi-overlay and kiwi-repart modules are incompatible, even it is clearly stated in kiwi.rnc that overlayroot is a valid option for OEM. The issue I see is that kiwi-overlay mounts the `/run/overlayfs` in dracut-initqueue.service, however the repart tries to repart the partition in dracut-pre-mount.service that happens after. Also `kiwi-repart` is not capable to understand `root=overlay:<dev>`, as it seams to not skip the `overlay` part and runs something like:

```bash
[ ! -e "overlay:/dev/disk/by-uuid/adfaads" ]
```

to verify if the device is present or not.

## Expected behaviour

`overlayroot="true"` works fine in most/all allowed combinations in schema. That leads to the questions of: 

1. _which are the supported use cases of `overalyroot="true"`?_
2. _Is there a better strategy to estimate the squashed partition size without having to squash twice?_

I am wondering how different is a vmx (or oem) without a bootpartition comparted to hybrid ISO, both using grub, both based over a squashed filesystem with an overlay rw partition. Probably narrowing the uses of overlayroot makes sense here.

Add `overlayroot="true"` in this [vmx type](https://github.com/OSInside/kiwi-descriptions/blob/2646e86fe3e5d1f917a28e7c572df99727403bfd/suse/x86_64/suse-leap-15.1-JeOS/config.xml#L24) and build it.

+ KIWI version: master
+ Operating system: Leap 15.1

davidcassany:
This commit makes the modules provided by dracut-kiwi-overlay and
dracut-kiwi-oem-repart compatible. The main idea of this commit
is forcing the preparation of the overlay rootfs write partition
after kiwi-repart gets into the play. Since kiwi-overlay mounts the
write partitiontition it has to be executed after the kiwi-repart.

For that instead of doing the job using an initqueue event added by a
udev rule we just add the very same logic in pre-mount hook that happens
after the kiwi-repart pre-mount hook thanks to lower priority.

Fixes #1184


davidcassany:
@schaefi beyond the fact some of the overlay logic needed to be executed **after** the kiwi-repart expands the partition I found out that kiwi-repart was not capable to handle a root variable definition like `overlay:/dev/disk/by-uuid/uuid`. Because of that and in order to not change the kiwi-repart code (because is sensitive and because I already like the way it is now) I am chaning the root value from `overlay:UUID=uuid` to `block:/dev/disk/by-uuid/uuid` which is likely to be more generic for the cooperation with other modules. I also noticed that `block:` syntax is already being used in dracut base code, but I could not find any reference of `overlay:` thus I assume this is just our own convention.

I am explaining you all that because I'd like to know your opinion regarding this change. Another option would be to just not use `overlay:` at all, but I like having this `overlay:` in the kernel command line, it is quite explicit.

schaefi:
> I also noticed that block: syntax is already being used in dracut base code, but I could not find any reference of overlay: thus I assume this is just our own convention.

yes this was my own naming idea. I like yours better if it does not cause conflicts with other modules. That was my main motivation. dracut modules often looks at the root definition and make their action depending on that "id" string. With "overlay:..." I wanted to make sure no other dracut code evaluates this

> but I like having this overlay: in the kernel command line, it is quite explicit.

yes I like the explicit information here too and would like to keep it.

Thanks much for your effort. I will take a closer look on the code once I find some time. Today are some meetings and I will travel home. So maybe I come back to this one tomorrow

davidcassany:
> I will take a closer look on the code once I find some time. Today are some meetings and I will travel home. So maybe I come back to this one tomorrow

Sure, there is no hurry for this one. I would also appreciate if you could give it a shot for OEM and VMX images. Some further testing with probably slightly different setup compared to my env would be really great.

schaefi:
I'm testing vmx and oem builds with overlayroot="true"

- [x] vmx test
- [ ] oem test

vmx worked fine for me

schaefi:
For testing I used the image description from here and just added the one attribute:

* https://github.com/OSInside/kiwi-descriptions/tree/master/suse/x86_64/suse-leap-15.1-JeOS

schaefi:
If you want me to test more specific setup just let me know

Thanks

schaefi:
Also tested oem. It worked fine. I saw a warning message from systemd that the systemd-fsck-root.service does not exist which is correct in case of overlayroot. The warning was just ignored and the system proceeded. I'm not sure if we need to mask another part but the repart operation worked for me in the test

davidcassany:
I am not sure about the systemd-fsck-root.service. I'd leave it as is and expect that it will not turn into an error for future systemd or OS versions. I doubt we should take further action here, looks like we might easily end up messing something else while now it is working. The less systemd configurations we touch the better, IMHO :)

Thanks for the tests. Yes I also saw good behavior in TW, this should be enough.

Unless you have any concern or code review pending I am done with this PR.

schaefi:
yes better not touch this. ok let's go for it. Thanks much

PR #1208, opened 2019-09-23
Update kiwi live ISO chapter

### LINKED ISSUE #1158: Add information about dmsquash vs. overlay live modules
## Problem description

There is no clarity about the differences and when to use the upstream dmsquash live module or the kiwi provided overlay live module

## Expected behaviour

It should be clear to users under which circumstances which module selection makes sense

schaefi:
Compare overlay vs.dmsquash live ISO modules by their
features to allow customers a proper selection of the
module. This Fixes #1158

### DIFF
--- a/doc/source/building/build_live_iso.rst
+++ b/doc/source/building/build_live_iso.rst
@@ -41,6 +41,7 @@ live ISO images:
   If set to `dmsquash`, the dracut standard dmsquash-live module will be
   used to support a live ISO system based on squashfs and the device
   mapper. Note, both modules support a different set of live features.
+  For details see :ref:`live_features`
 
 - `hybridpersistent`: Accepts `true` or `false`, if set to `true` then the
   resulting image will be created with a COW file to keep data persistent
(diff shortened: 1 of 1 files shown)


doc/source/building/build_live_iso.rst
```diff
@@ -69,3 +70,60 @@ be tested with QEMU:
 The image is now complete and ready to use. See :ref:`iso_to_usb_stick` and
 :ref:`iso_as_file_to_usb_stick` for further information concerning
 deployment.
+
+.. _live_features:
+
+Decision for a live ISO technology
+----------------------------------
+
+The decision for the `overlay` or `dmsquash` dracut module depends on
```
davidcassany: Probably here at the beginning we could state that `overlay` module is mostly conceived around an overlayfs deployment for write partitions and `dmsquash` is conceived around device-mapper snapshots.  I know you already noted that in the feature list, however this difference is what actually makes them a different creature.

schaefi: Actually I think the reason for dmsquash in the beginning was that there was only squashfs and the device mapper as the only possibility to map write actions to somewhere else. With the invention of the overlayfs kernel module things changed. But I agree from a concept perspective the dmsquash module was written as a generic device mapping technology whereas the kiwi overlay module had a smaller scope. I also added some information here and hope this does not confuse people :)

schaefi:
Thanks for the review. I added some information about the media verification differences. I also realized that a few small patches to our live iso module setup could be done to be more compatible with the dmsquash setup in this regard. I will do that in an extra PR

PR #1210, opened 2019-09-24
Fixed kiwi-live module setup

schaefi:
The installation of the checkmedia tool is optional and
and not mandatory. In addition activate the media check
verification also through the upstream used rd.live.check
kernel parameter. This is related to Issue #1158


schaefi:
yes good idea, I'll add it here and you can have another look

schaefi:
@davidcassany I was also thinking that the ```--force``` option might have an influence on the return code. Not sure if we want to change that

schaefi:
@davidcassany I think we should go without the force option and add the verbose option. I've done some test builds and that worked for me. Thoughts ?

schaefi:
ok, keeping fingers crossed that this will not have bad after effects. We will see it once the integration test builds are resolvable again. Thanks

PR #1214, opened 2019-09-29
Refactor use of raises in unit tests

schaefi:
Use raises as context manager instead of a method decorator.
This clearly identifies which code part is expected to raise
an exception. Related to Issue #1128

### DIFF
--- a/test/unit/archive_tar_test.py
+++ b/test/unit/archive_tar_test.py
@@ -1,8 +1,8 @@
-from mock import patch, call
-
+from mock import (
+    patch, call
+)
+from pytest import raises
 import mock
-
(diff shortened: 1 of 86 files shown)


schaefi:
@davidcassany not much to review here, if all tests pass this is good to go as it's a refactoring that it is not expected to change anything. However during refactoring I found many smaller issues present in the tests. I realized when using a decorator instead of the context manager any test code after the method that raised the exception is pointless. When switching those to the context manager that code became active and I found several mistakes which are now all corrected

davidcassany:
Wow! Good and huge work here :+1:

PR #1216, opened 2019-09-30
Added support for metalink repo defintions

schaefi:
Repository source paths that points to a metalink file which
needs to be downloaded and handled by the configured package
manager requires a different repository configuration. Thus
we now support a new attribute metalink which allows to flag
the repository definition as a metalink source. Currently
the dnf repository class is the only one that makes use of
the information. This change is required to support
Fedora >= 30 images. I also expect this concept to be used
by other distributions in the future

### DIFF
--- a/kiwi/repository/apt.py
+++ b/kiwi/repository/apt.py
@@ -129,7 +129,8 @@ class RepositoryApt(RepositoryBase):
         self, name, uri, repo_type='deb',
         prio=None, dist=None, components=None,
         user=None, secret=None, credentials_file=None,
-        repo_gpgcheck=None, pkg_gpgcheck=None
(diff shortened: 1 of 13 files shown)


Conan-Kudo:
@schaefi It'd be good if you also added `mirrorlist` since that's used by CentOS too...

Conan-Kudo:
Note that `baseurl`, `metalink`, or `mirrorlist` are more or less mutually exclusive.

schaefi:
> @schaefi It'd be good if you also added `mirrorlist` since that's used by CentOS too...

Ok, I've changed the code to provide the attribute ```sourcetype``` which allows one of the values ```baseurl```, ```metalink``` or ```mirrorlist```

Issue #1217, opened 2019-09-30
Invalid warning about exposed port using "12356/tcp"


rhafer:
## Problem description

When using `<port number="12345/tcp"/>` to explicitly expose a tcp port in a container I get this warning:

```
./kiwi/xml_parse.py:5076: UserWarning: Value "b'11211/tcp'" does not match xsd pattern restrictions: [['^(\\d+$|^\\d+/(udp$|^tcp))$']]
  warnings_.warn('Value "%s" does not match xsd pattern restrictions: %s' % (value.encode('utf-8'), self.validate_portnum_type_patterns_, ))
```
To me the pattern used for validation here looks somewhat fishy.

## Expected behaviour

Kiwi should accept  `<port number="12345/tcp"/>`

## OS and Software information
+ KIWI version:  9.18.13
+ Operating system: Tumbleweed
+ OBS version: -

davidcassany:
Have you verified if beyond the warning is it actually making applying the configuration?

toabctl:
The build seems to work. See https://build.opensuse.org/package/live_build_log/home:tbechtold:openstack-containers:j2gen:SLE_15_SP1/memcached-image/images/x86_64 for an example log

davidcassany:
Correct, this is what I expected. The values are still being passed to kiwi and we can see a correct `umoci` call in logs. Anyway I agree that we should find a way to get rid of this annoying warnings.

davidcassany:
Currently awaiting feedback on #1230 and the generateDS issue.

PR #1218, opened 2019-09-30
Fix validation pattern for portnum type

rhafer:
The pattern failed for all port that specified the tcp protocol.

Issue #1217

Fixes # .

*
*

### DIFF
--- a/kiwi/xml_parse.py
+++ b/kiwi/xml_parse.py
@@ -5074,7 +5074,7 @@ class port(GeneratedsSuper):
             if not self.gds_validate_simple_patterns(
                     self.validate_portnum_type_patterns_, value):
                 warnings_.warn('Value "%s" does not match xsd pattern restrictions: %s' % (value.encode('utf-8'), self.validate_portnum_type_patterns_, ))
-    validate_portnum_type_patterns_ = [['^(\\d+$|^\\d+/(udp$|^tcp))$']]
+    validate_portnum_type_patterns_ = [['^(\\d+$|\\d+/(udp|tcp))$']]
     def hasContent_(self):
         if (
 


kiwi/xml_parse.py
```diff
@@ -5074,7 +5074,7 @@ def validate_portnum_type(self, value):
             if not self.gds_validate_simple_patterns(
                     self.validate_portnum_type_patterns_, value):
                 warnings_.warn('Value "%s" does not match xsd pattern restrictions: %s' % (value.encode('utf-8'), self.validate_portnum_type_patterns_, ))
-    validate_portnum_type_patterns_ = [['^(\\d+$|^\\d+/(udp$|^tcp))$']]
+    validate_portnum_type_patterns_ = [['^(\\d+$|\\d+/(udp|tcp))$']]
```
aplanas: is not this file autogenerated?

davidcassany: yes the file is autogenerated. The regular expression is taken from https://github.com/OSInside/kiwi/blob/368f44262cb56c448b101a373003efca4a790a02/kiwi/schema/kiwi.rnc#L37

kiwi.rng is autogenerated too :)

schaefi: Yes the Relax schema is maintained in its condensed format (RNC) Thus schema changes must be performed on kiwi.rnc. Once a change there is made call:

```
make kiwi/schema/kiwi.rng
```

This builds the RNG schema as well as auto generates the data structures. The make call must be performed from within a tox development environment. For details read:

* http://osinside.github.io/kiwi/development.html#create-a-python-virtual-development-environment

Thanks

davidcassany:
Ok I have just validated that by using the latest generateDS the regular expression is translated as expected.

@schaefi I believe it might be time to test a newer version of generateDS and see if it works fine. This not obvious to test, but I'd say that some local tests and the integration tests should be sufficient (note that we should check no warnings are shown within the integration tests logs).

What do you think?

kiwi/xml_parse.py
rhafer: > is not this file autogenerated?

Ah, ok. Didn't know that.

PR #1219, opened 2019-09-30
Delete use of linuxefi/initrdefi grub commands

schaefi:
The above commands were used by kiwi on x86 in EFI mode but
seems to be deleted from upstream grub and therefore tend
to be no longer present on the distributions. In addition
my tests with older distributions and simple use of the
linux/initrd standard commands in EFI mode were successful
which lead to this patch.


schaefi:
Whether or not this can be merged still needs to be clarified. @Conan-Kudo I'm going to write Peter and Javier an e-mail to get more clarity. This coding will help to discuss this better

Conan-Kudo:
@schaefi Based on the information from Javier, we should probably not do this...

PR #1220, opened 2019-09-30
Update contributing chapter

schaefi:
Simplify the setup of a development environment and
update the documentation appropriately

### DIFF
--- a/doc/source/development.rst
+++ b/doc/source/development.rst
@@ -44,71 +44,57 @@ Install Required Operating System Packages
 KIWI requires the following additional packages which are not provided by
 :command:`pip`:
 
-* XML processing libraries: `libxml2` and `libxslt` (for :mod:`lxml`)
-* Python header files (for :mod:`xattr`), usually provided by a package
-  called `python-devel` or `python3-devel`
-* The `enchant` spell checking library
-* gcc compiler and glibc-devel header files
-* The `ShellCheck <https://github.com/koalaman/shellcheck>`_ script
-  linter.
-* A manipulation program for ISO images, either: ``xorriso`` (preferred) or
-  ``genisoimage``.
+XML processing libraries
+  `libxml2` and `libxslt` (for :mod:`lxml`)
 
-A full LaTeX installation is required to build the PDF documentation
-[#f1]_.
+Python header files, GCC compiler and glibc-devel header files
+  Required for python modules that hooks into shared library context
 
-On SUSE based systems run:
+Spell Checking library
+  Provided by the `enchant` library
 
-.. code:: shell-session
+ShellCheck
+  `ShellCheck <https://github.com/koalaman/shellcheck>`_ script linter.
 
-   $ zypper install --no-recommends \
-         python3-devel libxml2-devel libxslt-devel glibc-devel gcc \
-         trang xorriso \
-         texlive-fncychap texlive-wrapfig texlive-capt-of \
-         texlive-latexmk texlive-cmap texlive-babel-english \
-         texlive-times texlive-titlesec texlive-tabulary texlive-framed \
-         texlive-float texlive-upquote texlive-parskip texlive-needspace \
-         texlive-makeindex-bin texlive-collection-fontsrecommended \
-         texlive-psnfss
+ISO creation program
+  One of ``xorriso`` (preferred) or ``genisoimage``.
 
-On Fedora the following commands should install the required packages:
+LaTeX documentation build environment
+  A full LaTeX installation is required to build the PDF documentation
+  [#f1]_.
+
+The above mentioned system packages will be installed by calling the
+`install_devel_packages.sh` helper script from the checked out Git
+repository as follows:
 
 .. code:: shell-session
 
-   $ dnf install python3 python3-devel 'python3dist(pip)' \
-      'python3dist(tox)' make gcc which xz xorriso libxml2-devel \
-      libxslt-devel enchant genisoimage ShellCheck
-   $ # LaTeX packages
-   $ dnf install latexmk texlive-cmap texlive-metafont texlive-ec \
-      texlive-babel-english texlive-fncychap texlive-fancyhdr \
-      texlive-titlesec texlive-tabulary texlive-framed texlive-wrapfig \
-      texlive-parskip texlive-upquote texlive-capt-of texlive-needspace \
-      texlive-makeindex texlive-times texlive-helvetic texlive-courier \
-      texlive-gsftopk texlive-updmap-map texlive-dvips
+   $ sudo helper/install_devel_packages.sh
 
+.. note::
+
+   The helper script checks for the package managers `zypper` and
+   `dnf` and associates a distribution with it. If you use a
+   distribution that does not use one of those package managers
+   the script will not install any packages and exit with an
+   error message. In this case we recommend to take a look at
+   the package list encoded in the script and adapt to your
+   distribution and package manager as needed.
 
 Create a Python Virtual Development Environment
 -----------------------------------------------
 
-We recommend to setup a Python virtualenv for development. For Python 3
-:command:`python3 -m venv` is used, as it is bundled with Python itself
-(see https://docs.python.org/3/library/venv.html for details). For Python
-2.7, use :command:`virtualenv`, which is provided via :command:`pip` or as
-an extra package by your favorite Linux distribution.
-
-The following commands initializes a development environment for Python 3:
+The following commands initializes and activates a development
+environment for Python 3:
 
 .. code:: shell-session
 
-   $ python3 -m venv .env3
-   $ source .env3/bin/activate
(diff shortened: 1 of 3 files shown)


doc/source/development.rst
```diff
@@ -44,71 +44,57 @@ Install Required Operating System Packages
...
 -----------------------------------------------
 
-We recommend to setup a Python virtualenv for development. For Python 3
-:command:`python3 -m venv` is used, as it is bundled with Python itself
-(see https://docs.python.org/3/library/venv.html for details). For Python
-2.7, use :command:`virtualenv`, which is provided via :command:`pip` or as
-an extra package by your favorite Linux distribution.
-
-The following commands initializes a development environment for Python 3:
+The following commands initializes and activated a development
```
tomschr:
```suggestion
The following commands initializes and activates a development
```

doc/source/development.rst
```diff
@@ -44,71 +44,57 @@ Install Required Operating System Packages
...
+   $ tox -e devel
+   $ source .tox/3/bin/activate
 
-The :command:`develop` target of the :command:`setup.py` script
-automatically creates the application entry point called
-:command:`kiwi-ng-3`, which allows you to call the your modified version of
-KIWI from inside the virtual environment:
+The commands above automatically creates the application entry point
+called :command:`kiwi-ng-3`, which allows you to call KIWI from the
+python sources inside the virtual environment:
```
tomschr:
```suggestion
Python sources inside the virtual environment:
```

doc/source/development.rst
```diff
@@ -122,8 +108,7 @@ KIWI from inside the virtual environment:
 
    .. code:: shell-session
 
-      $ sudo $PWD/.env3/bin/kiwi-ng-3 system build ...
-
+      $ sudo $PWD/.tox/3/bin/kiwi-ng-3 system build ...
 
 In order to leave the development mode call:
```
tomschr:
```suggestion
To leave the development mode, run:
```

doc/source/development.rst
```diff
@@ -174,19 +159,13 @@ If you want to see the available targets, use the option `-l` to let
...
     $ tox -l
-    check
-    unit_py3_7
-    unit_py3_6
-    unit_py3_4
-    unit_py2_7
-    packagedoc
 
 To only run a special target, use the `-e` option. The following
 example runs the test cases for the Python 3.7 interpreter only:
```
tomschr: You mention "Python 3.7", but the tox call in the next line uses `unit_py3_6`. Seems like a contradiction, right? :wink:

helper/install_devel_packages.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
```
tomschr: I know, this comment is beyond the scope of this PR and probably a "matter of taste", but... you asked for comments so it's free. :smile: 

Have you considered to check for a specific operating system than for a specific command? I would have something in my mind like investigating `/etc/os-release`. The reason for my idea is to make it possible to "fine-tune" it for a specific OS. Checking for a specific command would probably not enough.

doc/source/development.rst
```diff
@@ -44,71 +44,57 @@ Install Required Operating System Packages
...
-         texlive-float texlive-upquote texlive-parskip texlive-needspace \
-         texlive-makeindex-bin texlive-collection-fontsrecommended \
-         texlive-psnfss
+LaTeX documentation build environment
+  A full LaTeX installation is required to build the PDF documentation
+  [#f1]_.
 
-On Fedora the following commands should install the required packages:
+The above mentioned system packages will be installed by calling the
+`install_devel_packages.sh` helper script from the checked out git
```
tomschr:
```suggestion
`install_devel_packages.sh` helper script from the checked out Git
```

doc/source/development.rst
```diff
@@ -44,71 +44,57 @@ Install Required Operating System Packages
...
+sources as follows:
```
tomschr:
```suggestion
repository as follows:
```

doc/source/development.rst
```diff
@@ -44,71 +44,57 @@ Install Required Operating System Packages
...
-   $ pip install -r .virtualenv.dev-requirements.txt
-   $ python3 setup.py develop
+   $ tox -e devel
+   $ source .tox/3/bin/activate
 
-The :command:`develop` target of the :command:`setup.py` script
-automatically creates the application entry point called
-:command:`kiwi-ng-3`, which allows you to call the your modified version of
-KIWI from inside the virtual environment:
+The commands above automatically creates the application entry point
```
tomschr: I'm not sure about the term "application entry point". Technically, it's correct, but it's Python lingo. I fear, this could confuse people if someone isn't a vivid Python developer or heard about this term.

As such, I'd suggest to simplify it:

```suggestion
The commands above automatically creates the application script
```

doc/source/development.rst
```diff
@@ -44,71 +44,57 @@ Install Required Operating System Packages
...
+called :command:`kiwi-ng-3`, which allows you to call KIWI from the
```
tomschr:
```suggestion
called :command:`kiwi-ng-3`, which allows you to run KIWI from the
```

helper/install_devel_packages.sh
schaefi: Well I thought about how complex the coding should be, but this is a helper and zypper is used on suse/SLES exclusively and dnf is found on Fedora/RHEL/CentOS. dnf can also be used on suse but that's a rare case and zypper will match first in this case. In the end I thought for a start and as a helper this is good enough. If we need to make this more complex and error proof it can be done on follow up requests

schaefi:
@tomschr thanks much for the review, this is really much appreciated. I've adapted the text

doc/source/development.rst
schaefi: yes that was a mistake, great catch

helper/install_devel_packages.sh
Conan-Kudo: `zypper` works on Fedora, and `dnf` works on openSUSE. So it might make sense to be a little discriminating here...

schaefi: :smile: ok so any nice way to detect the os from bash other than reading /etc/os-release ? I guess not. It sounds trivial but I think it's not: 

```
NAME
VERSION
ID
ID_LIKE
VERSION_ID
```

Looking for a simple solution...

schaefi: Take a look here:

* https://stackoverflow.com/questions/1977306/python-module-to-detect-linux-distro-version

and you know what I mean :) there is no safe way imho

Conan-Kudo: You can just source `/etc/os-release` and use `ID` and `ID_LIKE` to figure things out.

schaefi: ok let me collect this: I'm aware of:

* ID="sles"
* ID="opensuse-*"
* ID=fedora

@Conan-Kudo I'm lacking the information for CentOS and RHEL if dnf is used. Can you provide me that part ?

Thanks

Conan-Kudo: CentOS uses `centos`, RHEL uses `rhel`.

schaefi: @Conan-Kudo Thanks, I updated the script, can you take a look ?

helper/install_devel_packages.sh
```diff
@@ -0,0 +1,35 @@
...
+function os_id {
+    grep -E "^ID=" /etc/os-release | cut -f2 -d= | tr -d \"
```
Conan-Kudo: Why not just do `source /etc/os-release` and use the `$ID` variable?

schaefi: yes would be easy but is it expected to be sourced ? I see e.g things like ANSI_COLOR that contains escape sequences. If some funny guy just adds code it gets executed on source. Maybe that's a bit paranoid but a simple grep avoids any source magic especially on files outside my control. So it's a matter of trust :)

schaefi: Also keep in mind the documentation tells the user to call ```install_devel_packages``` with root permissions. This is required to install software but it would also mean we source /etc/os-release as root. I think that should be prevented. Thoughts ?

Conan-Kudo: The file is designed to be sourced, yes.

schaefi: ok, I changed it, let's hope this will never be abused

helper/install_devel_packages.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+source /etc/os-release
+
+case ${ID} in
+    fedora|centos|rhel)
+        dnf install \
```
Conan-Kudo: `dnf --setopt=install_weak_deps=False install` matches the `zypper` command, so you may want to do this here...

schaefi: ok done

PR #1222, opened 2019-10-02
Fixed rpmdb compat link setup

schaefi:
On older versions of zypper the path /var/lib/rpm was hardcoded
and not used from the rpm macro definition. For such systems and
to support them properly on hosts that have the rpm database
already moved a compat link was created. However if the host has
the rpm database at /var/lib/rpm the link doesn't make sense.
This patch fixes this and therefore bsc#1150190

### DIFF
--- a/kiwi/repository/zypper.py
+++ b/kiwi/repository/zypper.py
@@ -193,18 +193,18 @@ class RepositoryZypper(RepositoryBase):
         # For further details on the motivation in zypper please
         # refer to bsc#1112357
         rpmdb.init_database()
-        Path.create(
-            os.sep.join([self.root_dir, 'var', 'lib'])
-        )
-        Command.run(
-            [
-                'ln', '-s', ''.join(
-                    ['../..', rpmdb.rpmdb_host.expand_query('%_dbpath')]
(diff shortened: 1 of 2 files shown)


schaefi:
* https://bugzilla.suse.com/show_bug.cgi?id=1150190

davidcassany:
Looks like this dbpath change was definitely under estimated... There are still tools relaying on a fixed old dbpath and we are just adding hacks around it, hacks that we'll have to maintain and keep adapting to cover all corner cases :disappointed:

schaefi:
yes it's a nightmare. The problematic part is that we from the kiwi side have to care for the host rpmdb setup because it's relevant during bootstrap. This is a problem none of the people who deal with the rpmdb path on the system level have. That's probably the reason why there are still tools that read or hardcode dbpaths that are not read from the rpm macro system. As long as those exists we need to handle compat code. But it's done in a clean way in kiwi such that we can delete it when the time comes

schaefi:
The fix for the defect here is fortunately not a major problem, just a cosmetic issue. So no porting into sles is required

PR #1225, opened 2019-10-04
Flexible use of linuxefi module

schaefi:
Not all distributions provides the grub linuxefi module anymore.
This means a static list for building an efi grub image if needed
is no longer appropriate. This patch changes the module handling
at the following places

1. Use linuxefi for building custom efi modules only if present
   on the host system

2. Use linuxefi related grub2-mkconfig variables only if the
   host grub2-mkconfig implementation supports it

3. Prevent building custom efi image on Fedora by extending
   the search path for the distro provided efi image and also
   adapt the spec file accordingly

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -349,6 +349,20 @@ class Defaults:
             return 'grub'
 
     @staticmethod
+    def get_grub_config_tool():
+        """
+        Provides full qualified path name to grub mkconfig utility
+
+        :return: file path name
+
+        :rtype: str
+        """
+        for grub_mkconfig_tool in ['grub2-mkconfig', 'grub-mkconfig']:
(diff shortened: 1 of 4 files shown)


kiwi/defaults.py
```diff
@@ -560,7 +573,8 @@ def get_unsigned_grub_loader(root_path):
         """
         unsigned_grub_file_patterns = [
             '/usr/share/grub*/*-efi/grub.efi',
-            '/usr/lib/grub*/*-efi/grub.efi'
+            '/usr/lib/grub*/*-efi/grub.efi',
+            '/boot/efi/EFI/fedora/grubx64.efi'
```
Conan-Kudo: This should be `/boot/efi/EFI/*/grubx64.efi`. It's not just Fedora that ships that way. RHEL/CentOS and Mageia do too.

schaefi: yep, done now

PR #1227, opened 2019-10-07
Refactor patch_open use in builder's tests

davidcassany:
This commit removes the use of `@patch_open` decorator
in favor of directly patching `builtins.open` and use
mock.mock_open utility to mock the context manager.

Related to #1128

### DIFF
--- a/kiwi/boot/image/dracut.py
+++ b/kiwi/boot/image/dracut.py
@@ -258,9 +258,9 @@ class BootImageDracut(BootImageBase):
         if dracut_tool:
             outfile_expression = r'outfile="/boot/(init.*\$kernel.*)"'
             with open(dracut_tool) as dracut:
-                outfile = re.findall(outfile_expression, dracut.read())[0]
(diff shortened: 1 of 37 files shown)


davidcassany:
@schaefi I did the refactoring of mock_open in our unit tests. For that I am mostly using the `mock.mock_open` builder in many places this substitutes the custom defined context manager we got in many places.

Beyond that I found  a couple of issues doing this refactor (looks like some exceptions were hidden with our implementation of `@patch_open`):

1. See the  `kiwi/boot/image/dracut.py` changes, this is the only file from real kiwi sources that needed to be changed. Once I removed the `@patch_open` and IndexError exception started to pop up.

2. I noticed in some tests I had to mock kiwi command line arguments. I believe in most cases this is revealing wrong code from an API PoV. Some of the functionality can't be used without having certain command line parameters in the environment context, if someone uses KIWI as an API this really annoying IMHO.

PR #1236, opened 2019-10-11
Install all kiwi executables at setuptools level

### LINKED ISSUE #1226: kiwi-ng command not available after pip install kiwi
## Problem description

The `kiwi-ng` command is not available after `pip install kiwi`.

As per the instructions, I got the impression that it should work:

kiwi

## Expected behaviour

The `kiwi-ng` command is available after `pip install kiwi`.

```
root@0e4606e7890b:~# pip install kiwi
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
Requirement already satisfied: kiwi in /usr/local/lib/python2.7/dist-packages (9.18.18)
Requirement already satisfied: lxml in /usr/local/lib/python2.7/dist-packages (from kiwi) (4.2.6)
Requirement already satisfied: requests in /usr/local/lib/python2.7/dist-packages (from kiwi) (2.21.0)
Requirement already satisfied: PyYAML in /usr/local/lib/python2.7/dist-packages (from kiwi) (3.13)
Requirement already satisfied: pyxattr in /usr/local/lib/python2.7/dist-packages (from kiwi) (0.6.1)
Requirement already satisfied: docopt>=0.6.2 in /usr/local/lib/python2.7/dist-packages (from kiwi) (0.6.2)
Requirement already satisfied: urllib3<1.25,>=1.21.1 in /usr/local/lib/python2.7/dist-packages (from requests->kiwi) (1.24.3)
Requirement already satisfied: certifi>=2017.4.17 in /usr/local/lib/python2.7/dist-packages (from requests->kiwi) (2019.6.16)
Requirement already satisfied: chardet<3.1.0,>=3.0.2 in /usr/local/lib/python2.7/dist-packages (from requests->kiwi) (3.0.4)
Requirement already satisfied: idna<2.9,>=2.5 in /usr/local/lib/python2.7/dist-packages (from requests->kiwi) (2.8)

root@0e4606e7890b:~# which kiwi-ng
root@0e4606e7890b:~# 

root@0e4606e7890b:~# find /usr/local/ | grep kiwi-ng$
root@0e4606e7890b:~# 
```

+ KIWI version:  9.18.18
+ Operating system: Ubuntu 18.04.3 LTS
+ OBS version: n/a

davidcassany:
This PR changes the console script installed by pip to be `kiwi-ng` and `kiwicompat` instead of `kiwi-ng-3` or `kiwicompat-3`. The former ones are still accessible by symlinks provided by the `python3-kiwi` RPM.

Fixes #1226

### DIFF
--- a/doc/source/development.rst
+++ b/doc/source/development.rst
@@ -93,12 +93,12 @@ environment for Python 3:
    $ source .tox/3/bin/activate
 
 The commands above automatically creates the application script
-called :command:`kiwi-ng-3`, which allows you to run KIWI from the
+called :command:`kiwi-ng`, which allows you to run KIWI from the
 Python sources inside the virtual environment:
 
 .. code:: shell-session
 
-    $ kiwi-ng-3 --help
+    $ kiwi-ng --help
 
 .. warning::
 
@@ -108,7 +108,7 @@ Python sources inside the virtual environment:
 
    .. code:: shell-session
 
-      $ sudo $PWD/.tox/3/bin/kiwi-ng-3 system build ...
+      $ sudo $PWD/.tox/3/bin/kiwi-ng system build ...
 
 To leave the development mode, run:
 
--- a/setup.py
+++ b/setup.py
@@ -200,8 +200,8 @@ config = {
     },
     'entry_points': {
         'console_scripts': [
-            'kiwi-ng-{0}=kiwi.kiwi:main'.format(python_version),
-            'kiwicompat-{0}=kiwi.kiwi_compat:main'.format(python_version)
+            'kiwi-ng=kiwi.kiwi:main',
+            'kiwicompat=kiwi.kiwi_compat:main'
         ]
     },
     'include_package_data': True,
--- a/doc/source/working_with_kiwi.rst
+++ b/doc/source/working_with_kiwi.rst
@@ -199,9 +199,9 @@ The prepare step consists of the following substeps:
 
    .. code:: shell-session
 
-      $ kiwi system prepare $ARGS
+      $ kiwi-ng system prepare $ARGS
       $ # make your changes
-      $ kiwi system create $ARGS
+      $ kiwi-ng system create $ARGS
 
    .. warning:: Modifications of the unpacked root tree
 
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -373,8 +373,8 @@ make buildroot=%{buildroot}/ docdir=%{_defaultdocdir}/ install_package_docs
 
 # Create symlinks for correct binaries
 ln -sr %{buildroot}%{_bindir}/kiwi-ng %{buildroot}%{_bindir}/kiwi
-ln -sr %{buildroot}%{_bindir}/kiwi-ng-3 %{buildroot}%{_bindir}/kiwi-ng
-ln -sr %{buildroot}%{_bindir}/kiwicompat-3 %{buildroot}%{_bindir}/kiwicompat
+ln -sr %{buildroot}%{_bindir}/kiwi-ng %{buildroot}%{_bindir}/kiwi-ng-3
+ln -sr %{buildroot}%{_bindir}/kiwicompat %{buildroot}%{_bindir}/kiwicompat-3
 
 %if %{_vendor} != "debbuild"
 # kiwi pxeboot directory structure to be packed in kiwi-pxeboot


setup.py
```diff
@@ -200,7 +200,10 @@ def run(self):
     },
     'entry_points': {
         'console_scripts': [
+            'kiwi=kiwi:main',
+            'kiwi-ng=kiwi:main',
```
Conan-Kudo: I'd actually rather retire the `kiwi-ng` name... Do we really need it anymore?

davidcassany: I'd say we don't but there might be people around using it. That's why I am also keeping the `kiwi-ng-3`. Specially if there is people using kiwi from pip.

But I agree that at certain point we should drop everything else than `kiwi`. This includes removing the compat code.

I think that perl based kiwi support fully ends on 2022 (together with SLE12), but I'd say we probably should drop that compat support on upstream before than that and also reduce the console_scripts.

Conan-Kudo: This is probably a good time to drop it as any...

davidcassany: We are referring to kiwi-ng or kiwi-ng-3 in many places in the docs.
I'd say this needs to be addressed first. Also we need to ensure kiwi-ng is not used in OBS and probably there is other people scripting around kiwi out there.

Anyway I see your point. Probably we could announce it and then start with a first step having only `kiwi-ng`, `kiwi` and `kiwi-compat` and later in a later stage leave only `kiwi`.

In any case docs goes first/together and at least this should be addressed in this PR.

@schaefi do you have any thoughts

davidcassany:
I think the PR as is now is a good step before actually removing completely the `kiwi-ng` and `kiwi-compat` binaries. I believe we should leave them around for a while before completely remove them.

In any case I am not in favor of removing them without any announcement or warning before loosing any entry point. Probably we could add a warning message stating that `kiwi-ng` is deprecated and potentially removed in further versions. Probably something like that is better than adding some notes in a mailing list or documentation. @Conan-Kudo, what you think?

Conan-Kudo:
@davidcassany I think we shouldn't include them in the PyPI package. It makes no sense to do so. They're legacy links, and the way we've been making them in the spec is sufficient for now, so that we can drop them without adding new users to those CLI tools.

davidcassany:
Well, I have no strong opinion if we should better handle the console script compatibility links at spec level or at pip level.

In any case, I believe that pip, if it is only installing a single console script, this one should not be `kiwi-ng-3`, but `kiwi` or `kiwi-ng` (I have preference for `kiwi`).  In that case the symlinks could done in spec but the other way round, being `kiwi` the actual script and `kiwi-ng` and `kiwi-ng-3` symlinks. Command line examples in docs should only make use of whatever is installed by pip.

So my suggestion would be to turn `kiwi-ng` and `kiwi-ng-3` a symlink to `kiwi` at spec level and install `kiwi` within pip and only use `kiwi` in docs. I'd say this would be a good solution for #1226 and it also makes clear that the main script is `kiwi` and other are just compatibility links.

@Conan-Kudo @schaefi would you be fine with my suggestion?

Conan-Kudo:
@davidcassany Yes, sounds good to me.

setup.py
schaefi: I agree to the plan to get rid of it. But two important points were already mentioned:

- docs
- obs

I'm pretty sure if we deleted the kiwi-ng we will break all builds in obs. A patch in the build script is most likely needed. This however will be difficult because obs also supports building with the legacy kiwi version. afaik I coded the script in a way that it uses kiwi (for legacy builds) and kiwi-ng for next generation builds. If we now delete kiwi-ng we might open a can of worms that are hard to control.

So for now please let us keep both names and focus first how and if we can come up with a good solution on the obs side

Thoughts ?

schaefi:
@davidcassany yes that sounds good to me as well :+1:

setup.py
davidcassany: Just for the notes I believe that OBS is not calling `kiwi-ng` at all, at least it is not doing so in `obs-build/build-recipe-kiwi` and I doubt OBS calls KIWI anywhere else. Anyway I agree we can't do such a change without a proper coordination with the OBS folks and without proper announcements before actually removing any console script.

I lets update docs and align pip installation with the docs.

davidcassany:
Allright I believe now it is good enough to for all

setup.py
schaefi: I remember I wrote a function there that distinguishes by path, here is the magic from build-recipe-kiwi

```bash
    local kiwi_path=/usr/bin/kiwi kiwi_legacy=false
    if ! test -L "$BUILD_ROOT/usr/bin/kiwi" ; then
    kiwi_path=/usr/sbin/kiwi
        kiwi_legacy=true
    fi
```

I hope all this still works after the change, I think it should

schaefi: @davidcassany please double check before merge

davidcassany: Thanks for the hint, I'll check this portion in some home project of mine in OBS and verify the recipe does not face any regression. I'll do the merge once confirmed it works as expected.

davidcassany: You were right portion of the code in build-recipe-kiwi is causing issues against the suggestion I did. This part expects `/usr/bin/kiwi` to be a symlink, which was not the case with my suggestion.

I had to revert some of the changes, so pip will install `kiwi-ng` and keep `kiwi` and `kiwi-ng-3` as compat links at spec level.

davidcassany:
@schaefi this time I really think the changes are acceptable and without regressions, I had tested it in OBS.

pip will install `kiwi-ng` (which is in line with #1226) and `kiwi-ng-3` and `kiwi` will be created as symlinks in spec. So after all this is quite a minor change in relation of what we used to have. I also updated a couple of command line calls in docs to make use of `kiwi-ng` to be consistent with what pip is actually installing.

doc/source/development.rst
```diff
@@ -93,12 +93,12 @@ environment for Python 3:
    $ source .tox/3/bin/activate
 
 The commands above automatically creates the application script
-called :command:`kiwi-ng-3`, which allows you to run KIWI from the
+called :command:`kiwi`, which allows you to run KIWI from the
```
schaefi: shouldn't that be kiwi-ng  instead of just kiwi ?

davidcassany: you are right

Issue #1194, opened 2019-09-09
Refactor BootLoaderConfigGrub2 to use grub2-mkconfig


Vogtinator:
When building a .qcow2 image, the grub2.cfg provided by the OS is just fine and kiwi does not have to create its own based on a hardcoded template.

This means that settings are taken from /etc/default/grub, sysconfig and elsewhere as they should be and all packages and features (tuned, health-checker, etc.) work fine.

schaefi:
You have several opportunities to use the distro provided grub.cfg:

1. write a simple script that copies the file over the one kiwi wrote. Use the ```<type ... editbootconfig="script"``` attribute in your XML description to make your script active

2. write a systemd oneshot unit that is called on firstboot and runs grub2-mkconfig to apply the hostonly setup on first boot. In this case you would boot first with the kiwi created grub setup. If this is not wanted use 1. or both in combination

We are not changing the grub template approach because I don't trust the distro provided config file. Too often they were just broken and every distro handles it differently.

Vogtinator:
> You have several opportunities to use the distro provided grub.cfg:
>
> write a simple script that copies the file over the one kiwi wrote. Use the <type ... editbootconfig="script" attribute in your XML description to make your script active

That's definitely not simple. During `editbootconfig` and `editbootinstall` the target system is neither in a chrootable state nor are all OS files copied over yet. This means that running `grub2-mkconfig` is impossible without doing the rsync itself inside the script.

> write a systemd oneshot unit that is called on firstboot and runs grub2-mkconfig to apply the hostonly setup on first boot. In this case you would boot first with the kiwi created grub setup. If this is not wanted use 1. or both in combination

This would indeed not work for us.

> We are not changing the grub template approach because I don't trust the distro provided config file. Too often they were just broken and every distro handles it differently.

Right, and I don't trust the kiwi generated config file as it's often just broken and not every image is the same.

Vogtinator:
What might work to support a hack is moving one of the editboot* hooks after rsync, or calling `fstab.script` chrooted after rsyncing and writing the config, so that it can call grub2-mkconfig itself.

Both would be hacks though.

schaefi:
[...]

PR #1238, opened 2019-10-15
Refactor grub2 config file setup

schaefi:
This is a big change. On a high level it does the following:

1. For disk based images the grub config file is being created by calling ```grub2-mkconfig```

2. It adapts the building of disk images if the kiwi initrd system is used instead of dracut.
    The required changes are divided into two extra commits

I'm afraid this is not easy to review because I also had to adapt the unit tests to follow our refactoring investment, For review I suggest to build oem and vmx images plus live and install images from our existing image descriptions. I did that and things worked nicely so far

Thanks much for review


schaefi:
This references #1194

schaefi:
gitlab failed with ```There has been a runner system failure, please try again``` The funny part is that there is no way to restart a pipeline. At least I haven't found any button ??

davidcassany:
> gitlab failed with `There has been a runner system failure, please try again` The funny part is that there is no way to restart a pipeline. At least I haven't found any button ??

I happened to me too some time ago and I thought it was because I did not have the right permissions :P

schaefi:
Thanks good point, the template contains unused data now. I'll delete the unused parts

schaefi:
Thanks :) I'm so happy we can do refactoring

PR #1239, opened 2019-10-17
Fixed fallback copy of grub config file

schaefi:
For ISO images that are EFI bootable as well as for EFI
secure boot configurations that are not based on shim-install
kiwi provides a fallback code that copies the grub config file
to the efi/efi-vendor boot path. This fallback code was broken
because of the recent change to let grub2 mkconfig create the
config file. The call of grub2 mkconfig happens at a later
stage which required an adaption of the fallback mechanism.
This is related to Issue #1194

### DIFF
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -48,6 +48,7 @@ class BootLoaderConfigBase:
         self.volumes_mount = []
         self.root_mount = None
         self.boot_mount = None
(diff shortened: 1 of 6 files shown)


schaefi:
@davidcassany The integration test builds for non suse distros showed the problem

schaefi:
This one is still tricky, testing the shim fallback on Fedora.

schaefi:
ok works now

PR #1241, opened 2019-10-17
Fixed method header of write_meta_data

schaefi:
Interface class and implementation class has to use the
same header definitions.


schaefi:
This should fix isolinux based builds:

https://build.opensuse.org/package/live_build_log/Virtualization:Appliances:Images:Testing_x86/test-image-ubuntu:Live/images/x86_64

schaefi:
I hate making stupid mistakes, touching the grub stuff is always a nightmare

PR #1245, opened 2019-10-21
Fixed oem resize message and invokation

### LINKED ISSUE #1102: Improve Warning messages on resize operation
## Problem description

In reference to https://bugzilla.suse.com/show_bug.cgi?id=1136427 there seems to be
an issue with the displayed warning messages from kiwi. If the resize operation as it was
requested with the setup in the XML description cannot be done due to disk size limitations, kiwi
writes:

```
warn "Requested sizes exceeds free space on the disk:"
warn "Disk won't be re-partitioned !"

warn "Minimum required additional size: ${min_additional_mbytes} MB:"
if [ -n "${kiwi_oemrootMB}" ];then
    local share=$((kiwi_oemrootMB - disk_root_mbytes))
    warn "==> Root's share is: ${share} MB"
fi
if [ ${swapsize} -gt 0 ];then
    warn "==> Swap's share is: ${swapsize} MB"
fi
warn "Free Space on disk: ${disk_free_mbytes} MB"
```

The warning messages here seems to be not clear enough and an alternative suggestion was
made with

```
"Partition at end of disk, no resizing required"

instead of "Requested sizes exceeds free space on the disk:".
```

schaefi:
The oem resize should only start if there is unallocated
space on the disk available. If it starts the message in
case of a not applicable resize should be more meaningful
This commit addresses both issues and Fixes #1102

### DIFF
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -233,7 +233,7 @@ function check_repart_possible {
     if [ -n "${kiwi_oemrootMB}" ];then
         if [ "${kiwi_oemrootMB}" -lt "${disk_root_mbytes}" ];then
             # specified oem-systemsize is smaller than root partition
-            warn "Requested OEM systemsize is smaller than root partition"
+            warn "Requested OEM systemsize is smaller than root partition:"
             warn "Disk won't be re-partitioned !"
             echo
             warn "Current Root partition: ${disk_root_mbytes} MB"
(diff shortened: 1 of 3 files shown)


schaefi:
with this PR two things will be achieved

1. Do not run any re-partition if there is no unallocated space. This also prevents any re-run of the repart code if it already has happened

2. Change the warning messages if the XML requested size constraints can't be achieved

schaefi:
There are two possible message types:

If the XML description specifies a new root size which is smaller than the one in the image

```
Requested OEM systemsize is smaller than root partition
Disk won't be re-partitioned !
Current Root partition: 1042 MB
==> Requested size: 2042 MB
```

and/or if the requested or auto computed sizes for the partitions (root, swap) cannot be reached because the disk is too small

```
Requested systemsize exceeds free space on the disk:
Disk won't be re-partitioned !
Requested size(s): root:(keep), swap:(1024 MB)
==> Free Space on disk: 42 MB
```

schaefi:
In case there is no unallocated space on the disk no messages will be shown and the system just continues to boot

schaefi:
So repartition messages are only shown if the repartition code is invoked

schaefi:
> I see that the initialize method is now called before the mask_fsck_root_service. is it safe? Have you tested and check this particular change?

yes I was aware of that change and also have tested it. The initialize method looks up the disk devices and imports metadata as well as calculates disk sizes. All this is not related to any systemd service and does not conflict with devices because at this point nothing has been done with them that influences systemd. So I consider it safe

schaefi:
I'm aware the change is at a sensitive place so I'm happy if you also look at what the initialize method does. But I'm 99% sure it's safe... last famous words :)

davidcassany:
> > I see that the initialize method is now called before the mask_fsck_root_service. is it safe? Have you tested and check this particular change?
> 
> yes I was aware of that change and also have tested it. The initialize method looks up the disk devices and imports metadata as well as calculates disk sizes. All this is not related to any systemd service and does not conflict with devices because at this point nothing has been done with them that influences systemd. So I consider it safe

Right, if you took it into consideration I am fine with it. I really doubt it has any side effect and if you already took care to test it, this is already enough :+1:

davidcassany:
> I'm aware the change is at a sensitive place so I'm happy if you also look at what the initialize method does. But I'm 99% sure it's safe... last famous words :)

yes gonna look at it, however we will never be 100% sure of being bug-free :)

schaefi:
>Unless there is a reason for this change probably it is better to stick with the previous order and move lines 282~288 after the service is being masked.

The reason why I moved the initialize method is because it provides the disk device for me and if that device has no unallocated space we drop from the code. As I think the code from initialize does not interfere with udev and/or systemd there should be no reason to invoke any systemd masking. But I agree we should double check

schaefi:
The code in initialize has calls:

```
lsblk -p -n -r -s -o NAME,TYPE
multipath -l -v1
mdadm --detail -Y
blockdev --getsize64
blkid -s PARTUUID -o value
grep MemTotal: /proc/meminfo
```

If none of these trigger any systemd events we are fine. I don't see that this happens since I can call any of those on my running system too. Thoughts ?

schaefi:
I also want to achieve as less as possible fiddling with systemd. If there is no unallocated space we would have a superfluous mask/unmask call which I think is good when it only happens when it's needed

PR #1249, opened 2019-10-23
Refactor logging

### LINKED ISSUE #1244: Refactor logging validation in tests
## Problem description

Within the pytest module for python >= 3.4 there is the `caplog` fixture that can easily be used to capture log and validate the messages that were appended in logs.

So, instead of `@patch(kiwi.logger.log.warning)` and the associated `mock_warn` variable we could use directly the `caplog` fixture in tests and do something like:

```python
def test_setup_install_boot_images_efi_secure_boot(..., caplog):
  # The actual tests code
  # ...
  warn_msg = 'Running fallback setup for shim secure boot efi image'
  assert warn_msg in caplog.text
```

There should be no need to patch logging methods, just capturing the records should be enough.

schaefi:
Refactor use of logging facility
    
Use getLogger method instead of a global log object
Also use caplog fixture to capture log messages in
unit tests. This Fixes #1244

### DIFF
--- a/doc/source/development/kiwi_from_python.rst
+++ b/doc/source/development/kiwi_from_python.rst
@@ -20,17 +20,6 @@ modified description on stdout.
     from kiwi.xml_description import XMLDescription
     from kiwi.xml_state import XMLState
 
-    # Import of log handler only needed if default logging
-    # setup is not appropriate for the project
-    # from kiwi.logger import log
-
-    # By default the logging level is set to DEBUG, which
-    # can be changed by the following call
-    # log.setLogLevel(logging.INFO)
-
-    # Logging can also be disabled completely
-    # log.disabled = True
-
     description = XMLDescription('path/to/kiwi/XML/config.xml')
 
     xml_data = description.load()


test/unit/bootloader/config/base_test.py
```diff
@@ -10,6 +13,10 @@
 
 
 class TestBootLoaderConfigBase:
+    @fixture(autouse=True)
```
davidcassany: Nice! Learned something new today :)

test/unit/bootloader/config/base_test.py
```diff
@@ -124,13 +131,10 @@ def test_get_boot_cmdline_initrd_system_is_dracut_with_overlay(
...
     @patch('kiwi.xml_parse.type_.get_firmware')
-    @patch('kiwi.logger.log.warning')
-    def test_get_boot_cmdline_firmware_ec2_no_uuid(
-        self, mock_log_warn, mock_firmware
-    ):
+    def test_get_boot_cmdline_firmware_ec2_no_uuid(self, mock_firmware):
         mock_firmware.return_value = 'ec2'
-        self.bootloader.get_boot_cmdline()
-        assert mock_log_warn.called
+        with self._caplog.at_level(logging.WARNING):
```
davidcassany: I assume we are not validating the log with an assert on purpose because we already have the 100% code coverage requirement. The assert we used to have was kind of a redundant check. Am I right?

schaefi: yes, I have assertions only for those parts where we checked for a specific log message. All those parts where we just expect some logging I only embedded in the with blocks. I think if the code changes and for some reason no log info exists anymore it should not lead the test to fail unless we explicitly tested for an expected message. Do you think this is ok ?

test/unit/bootloader/config/base_test.py
schaefi: I also learned a lot while doing this :-)

test/unit/bootloader/config/base_test.py
davidcassany: Yes I also think this is OK. No need to track in tests all the log calls and its messages.

PR #1251, opened 2019-10-24
Fix btrfs_root_is_snapshot

davidcassany:
This commit makes sure the chrooted operations of over a loop device
including btrfs subvolumes and root as an snapshot include the
`/.snapshots` subvolume mount. This is necessary for certain operations
that depend on the layout, e.g. grub configuration using `grub2-mkconfig`.

### DIFF
--- a/kiwi/volume_manager/btrfs.py
+++ b/kiwi/volume_manager/btrfs.py
@@ -118,6 +118,11 @@ class VolumeManagerBtrfs(VolumeManagerBase):
             Command.run(
                 ['btrfs', 'subvolume', 'create', snapshot_volume]
             )
+            volume_mount = MountManager(
+                device=self.device,
+                mountpoint=self.mountpoint + '/.snapshots'
+            )
+            self.subvol_mount_list.append(volume_mount)
             Path.create(snapshot_volume + '/1')
             snapshot = self.mountpoint + '/@/.snapshots/1/snapshot'
             Command.run(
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi I did an oem test over x86_64 (including boot and verify fstab) and a successful build in the aarch63 (I have not booted the result).

I believe the change is not really intrusive and could not envision any other relevant test to verify it.

schaefi:
great catch, thanks much :+1: 

I have one question...

kiwi/volume_manager/btrfs.py
```diff
@@ -200,15 +205,6 @@ def get_fstab(self, persistency_type='by-label', filesystem_name=None):
...
         device_id = block_operation.get_blkid(blkid_type)
-        if self.custom_args['root_is_snapshot']:
-            mount_entry_options = mount_options + ['subvol=@/.snapshots']
-            fstab_entry = ' '.join(
-                [
-                    blkid_type + '=' + device_id, '/.snapshots',
-                    'btrfs', ','.join(mount_entry_options), '0 0'
-                ]
-            )
-            fstab_entries.append(fstab_entry)
```
schaefi: why is it needed to remove the inclusion of the snapshots volume to the fstab file ?

davidcassany: See this portion of the code:

https://github.com/OSInside/kiwi/blob/a66849ff8ea9a1c637c6dde088eebba2a8809864/kiwi/volume_manager/btrfs.py#L203-L222

For btrfs subvolumes we set fstab with the detaulf volume as `/` and the by looping to all entries in`self.subvol_mount_list`, which now also includes the `/.snapshots` subvolume if the  `btrfs_root_is_snapshot` was set. If you do not remove this portion you will end up with an fstab with two duplicated entries at the very beginning.

I think this is cleaner in the way that now `/.snapshots` is not an exception, it is just an additional subvolume automatically added when `btrfs_root_is_snapshot`.

schaefi: Ah yes makes sense it's already in the mount_list now, all clear great

Issue #1248, opened 2019-10-23
set GRUB_ENABLE_BLSCFG in /etc/default/grub


schaefi:
## Problem description

From a conversation with the grub maintainers on Fedora we know that RH/Fedora has switched over to using a variant of the Bootloader Spec for configuring GRUB since Fedora 30:
https://fedoraproject.org/wiki/Changes/BootLoaderSpecByDefault

kiwi should make use of it which means we have to set:

```
GRUB_ENABLE_BLSCFG
```

in ```etc/default/grub```. The setting is specific to distributions that implements this part of the bootloader spec. My expectation is that we can set the option and it's just silently ignored by distros that do not handle it

```
If GRUB_ENABLE_BLSCFG is not set to true in /etc/default/grub, then the
generated grub.cfg will define the menu entries and the grubby-deprecated
package must be installed so the old grubby tool can update the grub.cfg.
```

Conan-Kudo:
This also means that the way we generate grub configurations must change. We should not generate a `grub.cfg` in this case, and we need to configure things through `grubenv` and other aspects supported by GRUB's BLS implementation.

Conan-Kudo:
Also note that RHEL 8 uses BLS too.

schaefi:
This has been already done. See #1194 

Only for building live iso and install iso images the static template mechanism is still in place.
From my perspective it should be easy to adapt here now

martinezjavier:
@schaefi @Conan-Kudo if `grub2-mkconfig` is called after the `/etc/default/grub` has been written, then I think that the only needed change is to write a `GRUB_ENABLE_BLSCFG=true` in `/etc/default/grub` for Fedora. I'll try to come up with a patch.

PR #1252, opened 2019-10-25
Enable BLS configuration if is supported

martinezjavier:
Fedora now uses a BLS configuration by default, but this is not supported
by all distributions. So check if is supported by the grub2-mkconfig tool
and only enable the option if that's the case.

Fixes: #1248


schaefi:
@martinezjavier Thanks much for your contribution. The code looks great to me. You just did not adapt the respective unit test. This would be the code in ```test/unit/bootloader/config/grub2_test.py```. Let us know if you need assistance with that part or if you want us to jump in.

Thanks

martinezjavier:
@schaefi right, sorry for forgetting to update the test.

Just pushed a new version that fixes this.

schaefi:
perfect :+1: Thanks

PR #1258, opened 2019-11-06
Ensure grub.cfg is copied in EFI/BOOT folder

davidcassany:
This commit fixes the live images in efi mode. Grub configuration file
is copied to the correct location in <boot_dir>/EFI/BOOT.

Fixes bsc#1155815

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -64,7 +64,7 @@ build_doc:
       texlive-tabulary texlive-framed texlive-wrapfig texlive-parskip
       texlive-upquote texlive-capt-of texlive-needspace texlive-makeindex
       texlive-times texlive-helvetic texlive-courier texlive-gsftopk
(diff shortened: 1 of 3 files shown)


davidcassany:
@schaefi as I mentioned in the kiwi matrix channel I also noted that the `grub.cfg` file is not included inside the efi vfat partition, this doesn't look like a requirement according to my tests, however we used to include it before the whole `bootloader/config/grub2.py` refactor. I assume it was there as a leftover, but was not an actual requirement. Am I missing something? Looks like having efi images in there is enough.

schaefi:
> this doesn't look like a requirement

yep it's not. for iso images to boot via EFI the firmware has to find a fat filesystem. From there it reads the efi binary which is a grub image. grub now looks for its config file on the iso in the standard boot path. So the config doesn't need to be embedded but I will give this a test

PR #1264, opened 2019-11-14
Fixed zipl bootloader setup for s390 images

### LINKED ISSUE #1173: Failure creating qcow image on s390x: Support for zipl bootloader config not implemented
## Problem description

I am new to this space, please bear with me. I am trying to create a SLES12.5 qcow2 image for s390x architecture. I could not locate any s390 examples in [kiwi-descriptions](https://github.com/OSInside/kiwi-descriptions), so I am working off the [s390 template from the kiwi legacy repo](https://github.com/OSInside/kiwi-legacy/tree/master/template/s390/suse-SLE12-community-JeOS). I've tried a few variations of the `<type>` element with no success. 

Currently, I have the following in config.xml:
```xml
<type image="vmx" primary="true" format="qcow2" filesystem="btrfs" bootloader="grub2_s390x_emu" kernelcmdline="TERM=linux console=ttyS0">
  <machine memory="2048" ncpus="2"/>
```

Running `sudo kiwi-ng --type vmx system build --description sles12sp5/ --target-dir /tmp/jmuro/kiwi/` results in:

```
[ ERROR   ]: 01:14:50 | KiwiCommandError: mv: stderr: mv: cannot stat '/tmp/jmuro/kiwi/build/image-root/boot/initrd.vmx': No such file or directory
mv: cannot stat '/tmp/jmuro/kiwi/build/image-root/boot/linux.vmx': No such file or directory
```

Here is complete log: [image-root.log](https://github.com/OSInside/kiwi/files/3528377/image-root.log)

I also tried changing bootloader to `zipl` as follows:

```xml
<type image="vmx" primary="true" format="qcow2" filesystem="btrfs" bootloader="zipl" kernelcmdline="TERM=linux console=ttyS0">
 <machine memory="2048" ncpus="2"/>
```

Running `sudo kiwi-ng --type vmx system build --description sles12sp5/ --target-dir /tmp/jmuro/kiwi/` results in:

```
ERROR: 00:48:33 | KiwiBootLoaderConfigSetupError: Support for zipl bootloader config not implemented
```

I am willing to contribute my example template if I can get it working.

+ Architecture: s390x
+ KIWI version: KIWI (next generation) version 9.17.15
+ Operating system: SUSE Linux Enterprise Server 15 SP1
+ OBS version: n/a

schaefi:
The preparation to call zipl and the call itself were wrong.
For whatever reason the kernel image the initrd are moved
to another location prior to calling zipl. That move broke
the system because no kernel/initrd existed at the expected
place anymore. In addition the zipl call itself was issued
from a the wrong directory. Also no config file was written
as an after effect of the refactoring in Issue #1194. This
Fixes #1173 and bsc#1156694

### DIFF
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -89,7 +89,7 @@ class BootLoaderConfigBase:
         pass
 
     def setup_disk_image_config(
-        self, boot_uuid, root_uuid, hypervisor, kernel, initrd, boot_options
+        self, boot_uuid, root_uuid, hypervisor, kernel, initrd, boot_options={}
     ):
         """
         Create boot config file to boot from disk.
@@ -99,9 +99,15 @@ class BootLoaderConfigBase:
         :param string hypervisor: hypervisor name
         :param string kernel: kernel name
         :param string initrd: initrd name
-        :param string boot_options:
-            custom options required to setup the boot. The scope
-            of the options is specified in the implementation
+        :param dict boot_options:
+            custom options dictionary required to setup the bootloader.
+            The scope of the options covers all information needed
+            to setup and configure the bootloader and gets effective
+            in the individual implementation. boot_options should
+            not be mixed up with commandline options used at boot time.
+            This information is provided from the get_*_cmdline
+            methods. The contents of the dictionary can vary between
+            bootloaders or even not be needed
 
         Implementation in specialized bootloader class required
         """
--- a/kiwi/bootloader/config/zipl.py
+++ b/kiwi/bootloader/config/zipl.py
@@ -18,7 +18,6 @@
 import platform
 import logging
 import re
-import os
 
 # project
 from kiwi.bootloader.config.base import BootLoaderConfigBase
@@ -101,29 +100,9 @@ class BootLoaderConfigZipl(BootLoaderConfigBase):
             with open(config_file, 'w') as config:
                 config.write(self.config)
 
-            log.info('Moving initrd/kernel to zipl boot directory')
-            Command.run(
-                [
-                    'mv',
-                    os.sep.join(
-                        [
-                            self.boot_dir, 'boot',
-                            os.readlink(self.boot_dir + '/boot/initrd')
-                        ]
-                    ),
-                    os.sep.join(
-                        [
-                            self.boot_dir, 'boot',
-                            os.readlink(self.boot_dir + '/boot/image')
-                        ]
-                    ),
-                    self._get_zipl_boot_path()
-                ]
-            )
-
     def setup_disk_image_config(
         self, boot_uuid=None, root_uuid=None, hypervisor=None,
-        kernel=None, initrd=None, boot_options=''
+        kernel='image', initrd='initrd', boot_options={}
     ):
         """
         Create the zipl config in memory from a template suitable to
@@ -134,7 +113,7 @@ class BootLoaderConfigZipl(BootLoaderConfigBase):
         :param string hypervisor: unused
         :param string kernel: kernel name
         :param string initrd: initrd name
-        :param string boot_options: kernel options as string
+        :param dict boot_options: unused
         """
         log.info('Creating zipl config file from template')
         parameters = {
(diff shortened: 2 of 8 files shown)


kiwi/bootloader/config/zipl.py
```diff
@@ -149,10 +128,8 @@ def setup_disk_image_config(
             'title': self.quote_title(self.get_menu_entry_title()),
             'kernel_file': kernel,
             'initrd_file': initrd,
-            'boot_options': ' '.join([self.cmdline, boot_options]),
-            'failsafe_boot_options': ' '.join(
-                [self.cmdline_failsafe, boot_options]
-            )
+            'boot_options': self.cmdline,
```
davidcassany: Is it mandatory to remove the boot options from the command line? I am surprised by this change. `boot_options` parameter is not used within the whole method. I assume this is on purpose, but just want to make sure this is not a simple mistake.

kiwi/bootloader/config/zipl.py
```diff
@@ -101,29 +100,9 @@ def write(self):
...
-                        ]
-                    ),
-                    self._get_zipl_boot_path()
-                ]
-            )
-
     def setup_disk_image_config(
         self, boot_uuid=None, root_uuid=None, hypervisor=None,
-        kernel=None, initrd=None, boot_options=''
+        kernel='image', initrd='initrd', boot_options={}
```
davidcassany: `boot_options` is not used in the whole method. I assume this is on purpose and the change here is done just to align with the definition of the same method in `bootloader/config/grub2.py`. I am fine with it, but I don't see the need as the common background for these two classes is `bootloader/config/base.py` where this method does not define any default value. Thus the default value is specific to the specialized class. If unused I'd probably set here `None` change the param definition to `:param dict boot_options: unused`. Anyway I can live with the dict default value if prefer it.

Also I just realized that the sphinx API docs are wrong for this method, the definition of `boot_options` is wrong in all three: base class, zipl class and grub2 class. This should be fix.

kiwi/bootloader/config/zipl.py
schaefi: Yes this was an intentional change. boot_options in this method is a dictionary containing information to setup and configure the bootloader. It should not be mixed with commandline options. For those we have other methods in that class. I realized the parameter name is not the best but don't want to change the api interface incompatible. Thus I updated the doc string which I hope makes the scope clear.

In the zipl case no further booting options like the efi device or things like that are needed. The code just creates the zipl config file as a static information

kiwi/bootloader/config/zipl.py
schaefi: good catch, I updated the api docs to match the implementation. I also changed the baseclass definition to use the same default value than any other specialized implementations. I want this method to be initialized with an empty dict because implementation code will access it as such.

I hope now it makes more sense ?

kiwi/bootloader/config/base.py
```diff
@@ -100,8 +100,14 @@ def setup_disk_image_config(
         :param string kernel: kernel name
         :param string initrd: initrd name
         :param string boot_options:
```
davidcassany: nitpick sorry, but this is not a string. shouldn't it be `:param dict boot_options:`?

kiwi/bootloader/config/base.py
```diff
@@ -100,8 +100,14 @@ def setup_disk_image_config(
...
-            custom options required to setup the boot. The scope
-            of the options is specified in the implementation
+            custom options dictionary required to setup the bootloader.
+            The scope of the options covers all information needed
+            to setup and configure the bootloader and gets effective
+            in the individual implementation. boot_options should
+            not be mixed up with commandline options used at boot time.
+            This information is provided from the get_*_cmdline
+            methods. The contents of the dictionary can vary between
+            bootloaders or even not be needed
```
davidcassany: :+1:

kiwi/bootloader/config/zipl.py
```diff
@@ -134,7 +113,7 @@ def setup_disk_image_config(
         :param string hypervisor: unused
         :param string kernel: kernel name
         :param string initrd: initrd name
-        :param string boot_options: kernel options as string
+        :param string boot_options: unused
```
davidcassany: I'd say this needs to be `dict` or `dictionary` instead of string.

kiwi/bootloader/config/base.py
schaefi: you are so right :) I'll adapt

schaefi: done now

kiwi/bootloader/config/zipl.py
schaefi: yes, this was also wrong in the grub code, also fixed it there

schaefi:
@davidcassany I think it's clean code now, thanks for looking at it so detailed :+1:

PR #1265, opened 2019-11-14
Update s390 integration tests

schaefi:
Added vmx build test to run in kvm on s390. Updated the
existing oem build test to deploy on DASD. The concept
of the network setup invented by kiwi for s390 does no
longer apply and was also not used on the s390 distribution


schaefi:
Once #1264 is merged I plan to re-enable the s390 integration tests

schaefi:
> Just one question, where is it stated that oem image is targeting DASD? is it the `zipl_tragettype`?

Welcome to the wonderful world of mainframes :) This is expressed in kiwi speech by the zipl_target_type. and that heads into the following documentation:

* https://www.ibm.com/support/knowledgecenter/en/linuxonibm/com.ibm.linux.z.lkdd/lkdd_r_zipl_lboot_parms.html

So in short the setting zipl_targettype="CDL" makes this image a DASD target

PR #1266, opened 2019-11-14
Allow use of relative paths for `dir:<path>` URIs

### LINKED ISSUE #1261: Support relative paths in rpm-dir repository definition
## Problem description

A repository definition like the one below is not possible

```xml
<repository type="rpm-dir">
  <source path="dir:<relative_path>">
<reposiroty>
```

## Expected behaviour

To translate the relative path to absolute one and form there work as usual.

Include a plain directory of RPMs as a repository in config.xml using relative paths and then run the build.

+ KIWI version: 9.18.16
+ Operating system: SLE15
+ OBS version: None

davidcassany:
This commit allows the use of relative paths for local URIs using the
the following format:

`dir:<path>`

This is helpful to set in config.xml local URIs for repositories.

Fixes #1261

### DIFF
--- a/kiwi/system/uri.py
+++ b/kiwi/system/uri.py
@@ -246,7 +246,7 @@ class Uri:
         return iso_mount.mountpoint
 
     def _local_path(self, path):
-        return os.path.normpath(path)
+        return os.path.abspath(os.path.normpath(path))
 
     def _obs_project_download_link(self, name):
         name_parts = name.split(os.sep)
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi I added the "Don't merge" label because I am awaiting for a bugzilla ticket.

davidcassany:
@schaefi just removed the don't merge label. I think this fix should not wait anymore to get it merged. I was a awaiting for the user to fill a new bugzilla ticket for that (so I could include the reference on the commit message). But this is already few days ago, so I will manually include the bugfix trace in SLE history later on if the ticket arrives. I don't want to block anymore this fix because of that.

PR #1269, opened 2019-11-19
Fixed subsequent reboot of oem disk

schaefi:
On a second reboot of an oem disk we check with gdisk's
verification command if the disk needs to be resized.
That command however mounts the disk in the background
and therefore it's urgently required to mask the systemd
rootfs service before. Otherwise systemd thinks this is
evil and drops into a rescue shell

### DIFF
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -279,14 +279,6 @@ PATH=/usr/sbin:/usr/bin:/sbin:/bin
 
 setup_debug
 
-# initialize for disk repartition
-initialize
-
-if ! disk_has_unallocated_space "${disk}";then
-    # already resized or disk has not received any geometry change
-    return
-fi
-
 # when repartitioning disks, parted and friends might trigger re-reads of
 # the partition table, in turn triggering systemd-fsck-root.service
 # repeatedly via udev events, which finally can cause booting to fail with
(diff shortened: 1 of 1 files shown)


schaefi:
@davidcassany this patch makes sure we mask before the gdisk verify. I have tested it.
Sorry this was my fault introduced in the change that checks if the disk has free space available.
I wasn't aware that the gdisk checkup causes a rootfs mount. Sometimes I would wish there would be not sp much event driven actions happening on device nodes.

schaefi:
You are absolutely right, when I did the change I should have tested the reboot not only the first boot. I'm very sorry about this. I hope this change is not out on sle. I have to double check this. Anyway with the end of that sprint the sle packages gets updated

davidcassany:
@schaefi no, it did not reach SLE, just Factory, we are safe on that regard. You have been fast finding the issue and fixing it.

phillxnet:
@schaefi Thanks for finding and fixing this. I'm pretty sure I ran into this with my own fledgling experiments with the oem type but put it down to my unfamiliarity. Apologies for not reporting it myself. I think I ended up working around it by introducing a "dracut -f" oneshot via "ConditionPathExists=/.kiwi_dracut_hostonly.trigger" arrangement. Reporting here now in case this sheds any light on further ramifications. Apologies if my comments are noise or misplaced here. And thank for all your efforts on Kiwi, much appreciated.

schaefi:
@phillxnet yeah sorry for any inconvenience this created on your end. Your suggested workaround is good and the preferred solution as long as we have no new packages. You can expect updated packages including this fix by today.

Thanks for being patient ;)

PR #1272, opened 2019-11-21
Swap not last

### LINKED ISSUE #1231: oem resize, swap as last partition questionable
## Problem description

If the ```<oem-swap>true</oem-swap>``` setting is used in a kiwi oem build, the repartition operation creates the swap space as last partition after the root partition. Such a system cannot be easily resized if the storage disk changes its disk geometry. On SAN systems it's possible to reside disks and that geometry change gets populates to the OS. For the system to
make use of the new space the partition table needs to be adapted. Currently this is can be done as follows:

* delete swap
* resize root partition to new size
* create swap partition
* create swap space
* resize root filesystem

If we consider to allow the swap space to be created before the root partition the sequence changes to

* resize root partition to new size
* resize root filesystem

The reasons to have swap at the end of the partition table was due to performance issues on swap if it's not at the end. However that decision is years ago and maybe the original reason for setting up like that is no longer valid.

Let's discuss if it would make sense to change this.

Thanks

### LINKED ISSUE #1259: swap is mounted by label and not by UUID as other partitions
## Problem description
On an aarch64 JeOS built with kiwi on OBS, all partitions are mounted by UUID, except swap which is mounted by label.

## Expected behaviour
swap should be mounted by UUID.

+ KIWI version: 9.18.31
+ Operating system: Tumbleweed aarch64

schaefi:
**Prevent swap partition to be the last one**
    
In an OEM deployment that requested the creation of a swap
partition via <oem-swap> that swap partition was created
at first boot and was always the last partition on the disk.
This was required because it could not be placed before
any other partition without destroying those partition
contents. This process leaves the system in an inflexible
condition if the storage device can change its geometry
dynamicly as it's the case for SAN systems. The typical
deployment target for OEM images are SAN storage clusters
and it's cumbersome to resize the root partition if swap
is last.
    
This commit Fixes #1231 and changes the handling of swap if
requested via <oem-swap> as follows:
    
1. The swap space is created as part of the image build process
    and no longer on first boot of the image via dracut code.
    This increases the size of the non compressed .raw disk image
    by the configured swap space size or the default. The
    compressed versions are not affected since zero initialized
    swap space compresses to almost no space. Deployment of
    the image however also deploys the swap partition which
    increases deployment time. For big swap configurations
    it's advisable to switch off image verification via
    oem-skip-verify. For very big swap configurations it's
    also recommended to prevent kiwi from adding them as part
    of the image and let them be created on first boot via
    a systemd service that e.g places a swap file, or creates
    a swap volume when possible such that the fexibility to
    resize the rootfs is still available.
    
2. The setup of the swap space is now explicit. It's no longer
    calculated by twice times RAM size because on newer machines
    this could lead to huge numbers. Either the kiwi encoded
    default swap size applies or the user configured value.
    
3. LVM based oem disks creates the swap space as logical volume.
    The volume is created as part of the image build process
    and no longer on first boot. The swap volume at build time
    of the image is of a minimal size and gets resized on first
    boot.
    
4. The move of the swap creation into the builder code also
    handles swap per configured device persistency schema like
    any other devices. This means by default swap is mounted
    via by-uuid name and thus also Fixes #1259

**Fixed extending LVM volumes**
    
For extending the LVM volumes a resize prefix is set for each
volume. If no extra prefix is needed it can happen that the
former prefix setting is used which is wrong. The resize
prefix must be reset for each new volume iteration


schaefi:
@davidcassany have to move on with this and see if all tests are passing and working. I've done many tests on this. The only change that is inevitable if swap should not be last and the traditional partition table is in use is that the swap space is part of the image at it's configured size and will therefore impact the size of the image and also the deployment time if people use a big swap space setup. This can't be prevented if swap should not be last. Apart from that there are more advantages for the change than disadvantages:

* less code in the initrd
* swap handled the same way as other partitions in fstab
* consistently part of the image build process
* easier rootfs resize code in the initrd
* swap requested means swap will exist, no implicit request
* no implicit swap space calculation by ram space of the target

From this perspective I see the change to be good and worthwhile. The question is what is the strategy for big swapsize setup in the XML description. I can think of the following:

- Disable swapsize setup by kiwi and create a systemd one-shot service to setup a swapfile
- Disable swapsize setup by kiwi and setup an oem-systemsize that is smaller than the target disk.
  This leaves space available on the disk and the user can create a swap partition. However that
  will then be again a partition after root and makes the resize of the rootfs again more difficult
- Move to LVM, this will manage swap as volume and we don't have any problems

So there are opportunities to get to the same state as before but some work to get there will be needed. Please let me know your thoughts about it

Thanks

davidcassany:
> So there are opportunities to get to the same state as before but some work to get there will be needed. Please let me know your thoughts about it

Yes @schaefi I just started having close look today early in the morning. Sorry for the delay, last week had other issues on the plate and I wanted to have a close relax look to it without hurries. Form the code perspective I don't have any concern and also from the feature set I really like it.

I have just one corner case that came to my mind that I am curious about. What if someone sets a volume in systemdisk section with the `SWAP` label? I know this is most likely to be wrong, but I have seen it more than once, people trying to set the swap partition within the volumes section. I wanted to run a test to see if at least is not creating some sort of weird hard to understand side effect.

schaefi:
> What if someone sets a volume in systemdisk section with the SWAP label? 

I would expect this to end with a build error at the time the label is set but already exists. I think we should create a runtime check for the reserved labels: SWAP, ROOT, SPARE

davidcassany:
> > What if someone sets a volume in systemdisk section with the SWAP label?
> 
> I would expect this to end with a build error at the time the label is set but already exists. I think we should create a runtime check for the reserved labels: SWAP, ROOT, SPARE

Yes a runtime check probably makes sense here. These labels should be prohibited.

schaefi:
@davidcassany ok, added the runtime check

davidcassany:
mm I realized another issue :/ 

If we use not up to date dracut modules against this code we will endup with two different swap partitions, one done at build time, another one done at boot time... both included in fstab. On my tests the first one, the done at build time, is the one active, probably it depends on the fstab order. 

I believe we have a couple of alternatives:

1. We make sure to update all versions at the same time: SLE15 GA and SP1 (not sure if we already have kiwi in SP2). Not sure what we should do with SLE12...

2. Make sure that we disable the creation of the swap partition if it has been already created at build time. Probably revert kiwi_oemswap once the partition is done?

I'd be more in favor of 2nd if there is a clean way to achieve it. Assume we'll be always up to date will cause resistance on maintenance.

I think this is a relevant issue as we have faced an L3 related to incompatibilities across service packs.

schaefi:
very good point, yes this needs a handling. I'll check that

schaefi:
I pushed a commit that should handle it, can you take a look ?

schaefi:
One problem with backward compatibility however stays. The opportunity to resize a swap LVM volume has been added with this change too. People building with the latest builder but old dracut modules and using LVM will see their swap volume to stay at the original size. Nothing will break but the swap volume gets resized without the swap space to be resized. This can be done inside the system at any later point in time but is an inconsistency I'm afraid we have to live with. Thoughts ?

davidcassany:
> One problem with backward compatibility however stays. The opportunity to resize a swap LVM volume has been added with this change too. People building with the latest builder but old dracut modules and using LVM will see their swap volume to stay at the original size. Nothing will break but the swap volume gets resized without the swap space to be resized. This can be done inside the system at any later point in time but is an inconsistency I'm afraid we have to live with. Thoughts ?

Yes I don't see a good solution for that... The expansion is done at boot time and the code in that case doesn't know how to deal with it, thus better don't do anything. Another wild idea would be to provide the dracut modules within the python3-kiwi (obviously this is a major change out of scope now), but then we probably would end up with other backward incompatibilities because of upstream code not being compatible with certain old distro.

schaefi:
perfect thanks much for your input on this. It's always better to have more than one looking on the code :+1:

schaefi:
I'll pass this along to staging, let's see if all test builds succeeds and boots

PR #1275, opened 2019-11-25
Fixed VolumeManager device map handling

schaefi:
The former implementation builds a new device map which is
a subset of the low level device map. However due to the
additional swap partition the maps provided by the VolumeManager
classes are now incomplete. Instead this commit changes the
VolumeManager interface to receive the current low level
device_map and merged device changes on top of it when
needed.


schaefi:
@davidcassany of course found a problem with the help of the integration tests :-)

davidcassany:
> @davidcassany of course found a problem with the help of the integration tests :-)

Good :+1:

PR #1277, opened 2019-11-26
Fixed access to device map for fstab setup

schaefi:
Swap is an optional element in the device map.
Thus access to an element of that dictionary
should not fail

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -863,12 +863,12 @@ class DiskBuilder:
             device_map['root'].get_device(), '/',
             custom_root_mount_args, fs_check_interval
         )
-        if 'spare' in device_map and \
+        if device_map.get('spare') and \
            self.spare_part_fs and self.spare_part_mountpoint:
             self._add_generic_fstab_entry(
                 device_map['spare'].get_device(), self.spare_part_mountpoint
             )
-        if 'swap' in device_map:
+        if device_map.get('swap'):
             if self.volume_manager_name:
                 self._add_simple_fstab_entry(
                     device_map['swap'].get_device(), 'swap', 'swap'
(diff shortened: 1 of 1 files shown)


schaefi:
@davidcassany hopefully last after effect fix to swap being part of the device map

kiwi/builder/disk.py
```diff
@@ -863,12 +863,12 @@ def _write_generic_fstab(self, device_map, setup):
             device_map['root'].get_device(), '/',
             custom_root_mount_args, fs_check_interval
         )
-        if 'spare' in device_map and \
```
davidcassany: I am surprised this is failing... wasn't it a hashmap device_map? Anyway I am absolutely fine with the alternative

schaefi: Only the part that accessed the dictionary with ```if 'swap' in device_map``` was failing because swap doesn't have to exist in the dict. If you access the dict that way python fails if the key does not exist. So actually I would have only needed to change that one line but it felt odd to me to access only the swap key via the get() method and so I changed it completely for consistency. Does this makes sense ?

davidcassany: mmm I am not that sure, in my prompt the following

```python
map = {'swap': 'somevalue', 'root': 'someothervalue'}

if 'swap' in map:
  print('swap found')
else:
  print('swap not found')

if 'home' in map:
  print('home found')
else:
  print('home not found')
```

returns

```
swap found
home not found
```
 Anyway I like the way you coded it, I also prefer it :)

schaefi: Hmm, you are completely right, so what stupid things am I doing here ? There must be another reason, damn. At least that access method is considered the preferred solution, so I did not make it worse :)

PR #1279, opened 2019-11-27
Add new root option in Path.which method

### LINKED ISSUE #1276: Search for grub2-mkconfing tool in wrong paths
## Problem description

The method kiwi/bootloader/config/grub2._get_grub_mkconfig_tool` from should be looking for the tool at the appropriate chroot instead of the current host, as this tool is only usedo chrooted environments. This not causing problems as long as the host distro and the image distro use the same tool path, however, when this is not the case the code fails.

## Expected behaviour

The `grub2-mkconfig` should be verified within the image root, not host root.

Build an Ubuntu image from an openSUSE host and KIWI will fail trying to call `grub2-mkconfig` when it should be calling `grub-mkconfig`.

+ KIWI version: master
+ Operating system: openSUSE host / Ubuntu image
+ OBS version:

davidcassany:
This commit adds to Path utility a couple of methods to manipulate
paths. One to rebase given paths to a new root and another one to
trim the given root path. In addition a new option in Path.which
is added to allow searches into chroot evironments.

Fixes #1276

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -810,7 +810,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
 
     def _get_grub2_mkconfig_tool(self):
         for grub_mkconfig_tool in ['grub2-mkconfig', 'grub-mkconfig']:
-            if Path.which(grub_mkconfig_tool):
+            if Path.which(grub_mkconfig_tool, root_dir=self.root_dir):
                 return grub_mkconfig_tool
 
     def _get_grub2_boot_path(self):
--- a/kiwi/path.py
+++ b/kiwi/path.py
@@ -160,9 +160,48 @@ class Path:
                 )
 
     @staticmethod
+    def move_to_root(root, elements):
+        """
+        Change the given path elements to a new root directory
+
+        :param str root: the root path to trim
+        :param list elements: list of path names
+
+        :return: changed elements
+
+        :rtype: list
+        """
+        result = []
+        for element in elements:
+            normalized_element = os.path.normpath(element)
+            result.append(
+                normalized_element.replace(
+                    os.path.normpath(root), os.sep
+                ).replace('{0}{0}'.format(os.sep), os.sep)
+            )
+        return result
+
+    @staticmethod
+    def rebase_to_root(root, elements):
+        """
+        Include the root prefix for the given paths elements
+
+        :param str root: the new root path
+        :param list elements: list of path names
+
+        :return: changed elements
+
+        :rtype: list
+        """
+        result = []
+        for element in elements:
+            result.append(os.path.normpath(os.sep.join([root, element])))
+        return result
+
+    @staticmethod
     def which(
         filename, alternative_lookup_paths=None,
-        custom_env=None, access_mode=None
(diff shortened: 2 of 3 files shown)


kiwi/system/root_bind.py
```diff
@@ -173,15 +173,7 @@ def move_to_root(self, elements):
...
-        result = []
-        for element in elements:
-            normalized_element = os.path.normpath(element)
-            result.append(
-                normalized_element.replace(
-                    os.path.normpath(self.root_dir), os.sep
-                ).replace('{0}{0}'.format(os.sep), os.sep)
-            )
-        return result
+        return Path.move_to_root(self.root_dir, elements)
```
davidcassany: @schaefi I left it here to keep the API as it was. I have seen that this root_bind method is only used within the package manager classes, thus it's got just few in the code. It wouldn't be a big effort to completely remove it and use `Path.move_to_root` instead. Do you think it makes sense? It doesn't bother me leaving it as it is now.

davidcassany:
This PR would allow us some code cleanups like in `kiwi/bootloader/install/grub2.py`:

```python
def _get_tool_name(
        self, root_path, lookup_list, fallback_on_not_found=True
    ):
        chroot_env = {'PATH': os.sep.join([root_path, 'usr', 'sbin'])}
        for tool in lookup_list:
            if Path.which(filename=tool, custom_env=chroot_env):
                return tool
 ```

where it could be rewritten as:

```python
def _get_tool_name(
        self, root_path, lookup_list, fallback_on_not_found=True
    ):
        for tool in lookup_list:
            if Path.which(filename=tool, root_dir=root_path):
                return tool
```

However if we decide to go for that I believe it deserves a separate PR and dedicated issue as it that simple to find the code parts that could be simplified, there is no unique pattern.

kiwi/system/root_bind.py
schaefi: I'm almost convinced it will not cause any breakage and also looks like a nice change. Call me picky but can we leave that code part untouched for the moment ? The reason is that we in theory agreed on no bigger refactoring and I saw myself violating that a bit when I did the swap changes ;) As we plan to roll out everything we did to SLES at the end of this sprint I think it would be good to have not so many changes at code parts that not urgently needed it. Thoughts ?

davidcassany: Yep absolutely fine with me. I also like having this sort of changes in a separate PR.

schaefi:
I agree with the additional methods in the Path class we can simplify other parts in the code. It would be good if all of those will be handled in a separate PR though. Maybe you can open an issue and add links to code parts you know that could be rewritten ?

davidcassany:
> I agree with the additional methods in the Path class we can simplify other parts in the code. It would be good if all of those will be handled in a separate PR though. Maybe you can open an issue and add links to code parts you know that could be rewritten ?

Yes I'll create an issue.

kiwi/system/root_bind.py
schaefi: ok, so once the PR is updated we can merge it. I'll approve prior that change

davidcassany:
@schaefi just reverted the change on `kiwi/system/root_bind.py`. This is what you suggested right?

I am creating a card for the code clean up by using these new Paths methods.

PR #1278, opened 2019-11-26
Do not rsync kernel filesystems on debootstrap

davidcassany:
This commit ensures that /proc and /sys are not rsynched when
debootstrapping an apt based image.

Fixes #1270

PR #1280, opened 2019-11-27
Do not rsync /dev on debootstrap

davidcassany:
This commit does not rsync /dev on debootstrap and instead it uses
the bind mount for /dev the same way it is done for other non apt
based bootstrap processes.

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -21,7 +21,6 @@ import logging
 
 # project
 from kiwi.command import Command
-from kiwi.mount_manager import MountManager
 from kiwi.utils.sync import DataSync
 from kiwi.path import Path
 from kiwi.package_manager.base import PackageManagerBase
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi ok this is in the spirit of #1278. As commented in https://github.com/OSInside/kiwi/pull/1278#issuecomment-558977774 I don't see it feasible to avoid calling debootstrap in a separate dir and then rsync without major (and needless) refactor. I started to think about it due to umount on '/dev' I saw in the code (umount that was never remounted and kept in the root_bind mount stack). I think the solution is way simpler, do not umount `/dev` and do not rsync it, and use the `/dev` bind mount already provided by `root_bind` instance.

I tested it locally with an Ubuntu build on my Leap host.

@schaefi this is a refactor that it actually does not solve any issue, just a nice to have. I am find to block the merge until 2020. However it was that small that preferred to just commit it. I am fine adding a `Don't merge` to it.

schaefi:
Well it's app so won't break on SLE, thus I'm less concerned for the upcoming sle submissions. The code looks good and we will run the integration tests. Also you have tested it, that creates enough trust for me in here :+1:

PR #1282, opened 2019-11-27
Lookup distribution specific grub font dir

### LINKED ISSUE #1253: Unicode font missing on Fedora EFI system
## Problem description

When building a Fedora appliance and booting it in EFI mode, grub searches the unicode font in ```EFI/fedora/fonts``` on the EFI partition. However kiwi provides the font in ```/boot/grub2/fonts``` on the boot partition

schaefi:
In addition to the generic grub font directory also lookup
distribution specific font paths in the system and copy the
grub unicode font into it. This Fixes #1253

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -565,6 +565,24 @@ class Defaults:
                 return shim_file
 
     @staticmethod
+    def get_grub_efi_font_directory(root_path):
+        """
+        Provides distribution specific EFI font directory used with grub.
+
+        :param string root_path: image root path
+
+        :return: file path or None
+
+        :rtype: str
+        """
+        font_dir_patterns = [
+            '/boot/efi/EFI/*/fonts'
+        ]
+        for font_dir_pattern in font_dir_patterns:
+            for font_dir in glob.iglob(root_path + font_dir_pattern):
+                return font_dir
+
+    @staticmethod
     def get_unsigned_grub_loader(root_path):
         """
         Provides unsigned grub efi loader file path
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -865,6 +865,10 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
     def _copy_theme_data_to_boot_directory(self, lookup_path, target):
         if not lookup_path:
             lookup_path = self.boot_dir
+        font_name = 'unicode.pf2'
+        efi_font_dir = Defaults.get_grub_efi_font_directory(
+            lookup_path
+        )
         boot_fonts_dir = os.path.normpath(
             os.sep.join(
                 [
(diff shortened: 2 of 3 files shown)


schaefi:
at some point we have to refactor the entire unit testing for the bootloader code, it's just horrible

davidcassany:
> at some point we have to refactor the entire unit testing for the bootloader code, it's just horrible

I'd say we can add it on the ideas project or directly on the backlog and then conver it to an issue when we decide to handle that. I agree we should refactor those tests, they are tought

kiwi/defaults.py
```diff
@@ -564,6 +564,15 @@ def get_shim_loader(root_path):
             for shim_file in glob.iglob(root_path + shim_file_pattern):
                 return shim_file
 
+    @staticmethod
+    def get_distribution_grub_font_directory(root_path):
+        font_dir_patterns = [
+            '/boot/efi/EFI/*/fonts'
```
Conan-Kudo: In Fedora, the fonts are stored in `/usr/share/grub`. Why? No idea. @martinezjavier might know why it's `/usr/share/grub` instead of `/usr/share/grub2`...

schaefi: Yes we do search it there. Maybe the name of the method is misleading. This method looks up the distribution specific font dir for grub at boot time. Fedora efi binaries looks up the font in ```/boot/efi/EFI/fedora/fonts```. whereas suse efi binaries looks it up in ```/boot/grub2/fonts```

I treat ```/boot/grub2/fonts``` to be the standard path which should be used by efi binaries as well as by x86 boot code. For the suse systems I know this is true

On Fedora only the x86 boot code searches the fonts in ```/boot/grub2/fonts``` but the efi binaries looks in  ```/boot/efi/EFI/fedora/fonts```

```
rpm -ql grub2-efi-x64

/boot/efi/EFI/fedora/fonts
/boot/efi/EFI/fedora/grub.cfg
/boot/efi/EFI/fedora/grubenv
/boot/efi/EFI/fedora/grubx64.efi
/boot/grub2/grubenv
/boot/loader/entries
/etc/grub2-efi.cfg
```

So kiwi should have a way to provide the font in the "standard" path and additionally provide it to an eventually existing distribution specific path.

@Conan-Kudo if you can come up with a more clear name on the method to somehow make that clear I would love to take it, it was already quite hard to find some :)

schaefi: Especially for images that are multiboot capable (x86 bios code plus efi) the different font lookup is not nice. But we can handle it here

schaefi: I'm testing on a Fedora30 VM in Azure

Conan-Kudo: `get_distribution_grub_font_directory` -> `get_grub_efi_font_directory`

schaefi: ok let's give it the efi scope. I hope this will only be needed in that area and is not different for x86 boot code

schaefi: @Conan-Kudo hope it makes more sense now

schaefi:
@Conan-Kudo good to go ?

PR #1284, opened 2019-11-29
Fixed setup of default grub config

schaefi:
In /etc/default/grub GRUB_CMDLINE_LINUX_DEFAULT also contained
the root= information. If grub2-mkconfig runs with that information
it places the root device information twice because grub2-mkconfig
resolves this information itself. This commit prevents the root=
information to be placed in the default grub config and
Fixes bsc#1156908


schaefi:
Still testing... I might found an inconsistency with this change on the overlayroot feature

schaefi:
ok also covered the overlayroot case

PR #1297, opened 2019-12-08
Fixed swap setup if btrfs is used

schaefi:
In case of a volume manager the simplified variant of the
device name is used in the fstab file to reference the
swap device. However this is only correct for the lvm
volume management but not for btrfs. In case of btrfs
the swap space is not a subvolume but a real partition
and thus the simplified device spec in fstab puts in the
loop mapped device which is wrong. This patch fixes it

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -870,7 +870,7 @@ class DiskBuilder:
                 device_map['spare'].get_device(), self.spare_part_mountpoint
             )
         if device_map.get('swap'):
-            if self.volume_manager_name:
+            if self.volume_manager_name and self.volume_manager_name == 'lvm':
                 self._add_simple_fstab_entry(
                     device_map['swap'].get_device(), 'swap', 'swap'
                 )


schaefi:
@davidcassany This one has to be merged to the kiwi_sle15 submission. I found this during the weekend, sorry for the bad timing

schaefi:
@davidcassany I've updated GA and SP1 branches with a cherry-pick of this PR and updated QA:Maintenance projects with a package including the fix. I needed this to verify for the SAP BOne installer to be fixed

phillxnet:
@schaefi Thanks for yet another timely fix. Am I right in thinking pre this pr one would end up with an fstab entry akin to:
```
/dev/mapper/loop0p3 swap swap defaults 0 0
```
when Kiwi config has the following entries:
```
<type
    image="oem"
    initrd_system="dracut"
    filesystem="btrfs"
...
<oemconfig>
    <oem-swap>true</oem-swap>
```
Which in turn, as /dev/mapper/loop0p3 doesn't exist, results in a systemd 1.5 minute boot wait of:
```
A start job is running for dev-mapper-loop0p3.device (1min 12s / 1min 30s)
```
If we built post pr:
"Swap not last" #1272

Apologies if this is noise, just trying to follow along with development in the hope I can contribute more in time.

schaefi:
@phillxnet yes this is exactly the problem and the change in here fixes this. My bad I should have seen this earlier.

phillxnet:
@schaefi Thanks again for the fix and for the confirmation.

Again I was too slow/unsure to report this myself. Just not that familiar yet with Kiwi but will try to chip in where I can, especially once our config has settle down so I know more what to expect.

schaefi:
@phillxnet that's great and much appreciated. Thanks, don't hesitate to report whatever you find :)

schaefi:
@davidcassany I can confirm this change fixes the problem. I leave it open to let you know to have that in the sle15 submission. Please note I already updated the kiwi_sle15 branches but I did not submit anything to SLE since I don't wanted to interfere with your process. Let me know if there are concerns and then we can merge this for master. Thanks

PR #1305, opened 2019-12-20
fix grub config for Fedora > 29

M0ses:
As documented here:

https://fedoraproject.org/wiki/Changes/BootLoaderSpecByDefault

Fedora >= 30 uses BootLoaderSpec (BLS). This means that
GRUB_CMDLINE_LINUX is used in favour of GRUB_CMDLINE_LINUX_DEFAULT.
(SEE https://fedoraproject.org/wiki/Changes/BootLoaderSpecByDefault#Config_files_provided_by_kernel.2C_grub2-switch-to-blscfg.2C_and_zipl-switch-to-blscfg)

This patch ensures that GRUB_CMDLINE_LINUX is also set to the same values
of GRUB_CMDLINE_LINUX_DEFAULT in case that BLS was detected.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -590,6 +590,10 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         )
         if enable_blscfg_implemented.returncode == 0:
             grub_default_entries['GRUB_ENABLE_BLSCFG'] = 'true'
+            if self.cmdline:
+                grub_default_entries['GRUB_CMDLINE_LINUX'] = '"{0}"'.format(
+                    re.sub(r'root=.* |root=.*$', '', self.cmdline).strip()
+                )
 
         if grub_default_entries:
             log.info('Writing grub2 defaults file')


davidcassany:
@M0ses thanks for the PR, I see failing unit tests, are you working on getting them fixed? I believe it is just a matter of adapting some command calls in `bootloader/config/grub2_test`.

We can help you on adapting the related test if you want.

On the other hand I assume you saw the issue on some image from IBS, could you point me to the image that is not booting?

Also, this is my last day before Christmas does I don't expect to merge that before 7th of January unless there is a clear urgency.

M0ses:
> @M0ses thanks for the PR, I see failing unit tests, are you working on getting them fixed?

No not ATM - I only wanted that this issue doesn't get lost while chrismas vacation:-)

> I believe it is just a matter of adapting some command calls in `bootloader/config/grub2_test`.
> 
> We can help you on adapting the related test if you want.

Yeah - would be great

> 
> On the other hand I assume you saw the issue on some image from IBS, could you point me to the image that is not booting?

no it was on the kanku images:
https://build.opensuse.org/package/show/devel:kanku:images/Fedora-31
but please be aware that the images there are build with a patched kiwi version
https://build.opensuse.org/package/view_file/devel:kanku:images:supply/python-kiwi/BLS.patch?expand=1
so you will not see the errors in it.

Please be also aware that the images are booting - but in my case I get no output  on the serial console as my cmdline parameters "console=ttyS0" are not passed to the kernel

> 
> Also, this is my last day before Christmas does I don't expect to merge that before 7th of January unless there is a clear urgency.

Nope - not urgent - have a nice vacation

schaefi:
Happy new year :)

@M0ses thanks for bringing this up, would you mind to check the solution suggested in #1306 ? if ok and accepted I vote for closing this one. Thanks

M0ses:
@schaefi LGTM - closing in favour of #1306

PR #1306, opened 2019-12-29
Fixed grub config if bootloader spec is in use

schaefi:
Distros that implements the grub bootloader spec prefers
GRUB_CMDLINE_LINUX over GRUB_CMDLINE_LINUX_DEFAULT. Details:
https://fedoraproject.org/wiki/Changes/BootLoaderSpecByDefault

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -529,7 +529,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         * GRUB_USE_LINUXEFI
         * GRUB_USE_INITRDEFI
         * GRUB_SERIAL_COMMAND
-        * GRUB_CMDLINE_LINUX_DEFAULT
+        * GRUB_CMDLINE_LINUX_DEFAULT | GRUB_CMDLINE_LINUX
         * GRUB_GFXMODE
         * GRUB_TERMINAL
         * GRUB_DISTRIBUTOR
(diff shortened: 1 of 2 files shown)


schaefi:
In reply to #1305 I suggest a different solution. The bootloader spec describes that GRUB_CMDLINE_LINUX is preferred over GRUB_CMDLINE_LINUX_DEFAULT is the spec is respected. For me this means either one or the other depending on GRUB_ENABLE_BLSCFG.
Thus we should prevent to set both config options to make it clear

M0ses:
passed checks in OBS/kanku - LGTM

Conan-Kudo:
@martinezjavier, @vathpela: Could either of you quickly look over and see if this change makes sense?

martinezjavier:
@Conan-Kudo sorry for the late reply, I was on holidays. I’ll take a look to it tomorrow if you don’t mind.

Conan-Kudo:
@martinezjavier Sounds good to me!

martinezjavier:
@schaefi @Conan-Kudo this was actually a bug in the Fedora's GRUB blscfg module since both the `GRUB_CMDLINE_LINUX` and `GRUB_CMDLINE_LINUX_DEFAULT` can be used to set the kernel command line parameters.

I've fixed this for rawhide some time ago: https://src.fedoraproject.org/rpms/grub2/c/d3ceae4bfd5a but I think this fix should also make it to Fedora 30 and 31.

So the correct place to fix this is in Fedora and not force kiwi to use `GRUB_CMDLINE_LINUX` for that distro.

schaefi:
@martinezjavier thanks much for the feedback :+1:

schaefi:
Just in case the fix will not make it to Fedora 30/31 for some reason, I'll keep the branch with the code for a while. Would be great if we can get a short notification if the fix is available in Fedora 30/31. Thanks

martinezjavier:
@schaefi sorry for the delay, I just submitted grub2 updates for Fedora 30 and 31 fixing this issue:

https://bodhi.fedoraproject.org/updates/FEDORA-2020-e3adf88ba7
https://bodhi.fedoraproject.org/updates/FEDORA-2020-95abadb310

schaefi:
Perfect thanks much @martinezjavier :+1:

PR #1309, opened 2020-01-09
Refactor kernel version lookup

### LINKED ISSUE #1308: armv6/7 images broken on dracut initrd
## Problem description
Since update to kernel 5.4, armv6/7 images are broken on dracut initrd because `(no output on stdout)` is print instead of the kernel version:
```
[ 2324s] [ INFO    ]: 05:43:01 | Creating generic dracut initrd archive
[ 2324s] [ DEBUG   ]: 05:43:01 | EXEC: [kversion /usr/src/packages/KIWI-oem/build/image-root/boot/zImage-5.4.7-1-default]
[ 2324s] [ DEBUG   ]: 05:43:01 | EXEC: [chroot /usr/src/packages/KIWI-oem/build/image-root dracut --verbose --no-hostonly --no-hostonly-cmdline --xz --install /.profile --add  kiwi-repart  --install /config.partids openSUSE-Tumbleweed-ARM-JeOS-efi.armv7l-2020.01.08.initrd.xz (no output on stdout)]
```
Full build log: 
https://build.opensuse.org/package/live_build_log/openSUSE:Factory:ARM:Live/JeOS:JeOS-efi-pxe/images/armv7l

+ KIWI version: 9.19.8

schaefi:
For the lookup of the kernel version kiwi provided a small
C program that reads the first bytes of the kernel binary
and extracts the version information. Given the various
compression formats and also the problem of kernel files
that contains the decompressor code which could be anything
it has turned out to be difficult to provide a stable tool
to read the version from the binary. Therefore the code in
kiwi was refactored to read the version via regexp match
from the filename which also makes kversion an obsolete
tool. This Fixes #1308


ggardet:
LGTM (even if I am not a kiwi expert) and the tests pass, so the regular expression seems ok. ;)

kiwi/system/kernel.py
```diff
@@ -56,34 +57,19 @@ def get_kernel(self, raise_on_not_found=False):
...
-                version = version.rstrip('\n')
-                kernel = namedtuple(
-                    'kernel', ['name', 'filename', 'version']
-                )
-                return kernel(
-                    name=os.path.basename(os.path.realpath(kernel_file)),
-                    filename=kernel_file,
-                    version=version
+                version_match = re.match(
+                    '.*?-(.*-[^\.]+)', os.path.basename(kernel_file)
```
davidcassany: I am surprised by this regexp. According to that a valid kernel file could be `-foo-b` and match would be `foo-b`.

I guess your intention is to make as less restrictive as possible just forcing the version to be expressed in two parts separated by a dash (e.g. `some-version`).

I am wondering if it would make sense to request for a numeric version for the first part, this is how I would have done it.

So far I could only see issues with this regexp if dashes are being used before the version or if the second part of the version is numeric including dots.

`linux-1.3.1-2.3.gz` -> `1.3.1-2`  
`linux-kernel-1.2.3` -> `kernel-1`
`linux-kernel-1.3.1-default` -> `kernel-1.3.1-default`

I am fine with the first example, I don't see any good way to distinguish between file extension and version portion if the file name ends with a numeric dot separated version. However I am wondering about the other two. Probably we could force a numeric version for the first part or remove the `?` (actually I am not sure why is it in there, I'd assume the first part of the version can't include `-`).


What you think, would it make sense to remove the `?` or force the first part of the version to be numeric?

Anyway all this is just nit-picking, I like the current approach and I see the risks of trying to be more specific here. So it is fine with me to consider the examples above as acceptable results at some point we have to assume the kernel file follows a convention and it won't be anything.

schaefi: >I guess your intention is to make as less restrictive as possible just forcing the version to be >expressed in two parts separated by a dash (e.g. some-version).

This is correct, the expression is based on the assumption that the kernel is named by the prefix list which we also use in _setup_kernel_names_for_lookup. The prefix list is:
```
'uImage', 'Image', 'zImage', 'vmlinuz', 'vmlinux', 'image'
```
This means the case that the kernel is named ```linux-kernel-1.2.3``` is not expected. If this happens kiwi will not find this kernel as a valid kernel file before.

The ? in the expression makes this first match a lazy match up to a '-' I do this because version numbers for kernels are not only numbers: vmlinuz-4.12.14-lp151.28.32-default
Therefore the version match should be generic and allow every character. But to match that I needed an anchor and this was the only one I could come up with so far

I'm open for any more robust expression

davidcassany: > I do this because version numbers for kernels are not only numbers: vmlinuz-4.12.14-lp151.28.32-default

Oh I see, makes complete sense. Yes in that case the `?` makes sense. I am happy assuming that anything coming after the `<prefix>-` is considered to be the version, regardless of the nature of what is in there. Looks reasonable to me.

At most I would consider something like `.*?(\d+.*-[^\.]+)` (the anchor is the first digit) or `.*?(\d+(\.\d+){2}.*-[^\.]+)` forcing the version to include a `major.minor.patch` pattern and use it as the anchor. But I hardly see a real benefit using it.

Have you tried this approach on fedora? Just saw that my fedora vm includes a kernel file named `vmlinuz-5.3.11-300.fc31.x86_64`, according to the regexp the version would be `5.3.11-300.fc31` I guess this could be a problem. The evaluated version is used by dracut to find modules in `/lib/modules/` right?

schaefi: > would be 5.3.11-300.fc31 I guess this could be a problem.

grmbl yes this would be a problem. It's the reason why I liked the kversion C program, damn it everybody does it differently, let me check if the expression can be even more generic.

davidcassany: There is the unpleasant option of the removing the file extension based on list of possible extensions... I dislike it, as we are likely to miss some options and new ones may come into play.

Also there is the option to validate it/guess it with the directories in `/lib/modules`. As a sanity check validate that the given version exists as directory in `/lib/modules`, and if not check if the matched version is at part of only one directory name in `/lib/modules`. I know it sucks, just raising all ideas that come to me.

schaefi:
@davidcassany ok I pushed another commit that makes the match even more simplistic. The only reason for the suffix expression was because of vmlinux-version.gz. In former times this was used to bootup and the .gz suffix has to be ignored for consistency with /lib/modules content. However I think those times are gone for a long time.

Can you double check:

a) with the kernel name from fedora ?
b) if my assumption seems to be correct ?

Thanks

schaefi:
> Also there is the option to validate it/guess it with the directories in /lib/modules

This is already done in the method that builds up the kernel names list

schaefi:
>There is the unpleasant option of the removing the file extension based on list of possible >extensions... I dislike it, as we are likely to miss some options and new ones may come into play.

I did not understand that part

davidcassany:
> > There is the unpleasant option of the removing the file extension based on list of possible >extensions... I dislike it, as we are likely to miss some options and new ones may come into play.
> 
> I did not understand that part

I was thinking in simply removing the file extension (e.g. `.gz`) based on a list of extension candidates (mostly common compression extensions). It could be done with a sophisticated regex or just postprocessing the match with something like `version_str.replace('.gz', '')` .

But I am also prefer to keep it as simple as possible, specially if we are confident that the `.gz` variants are not in use anymore. I verified that a `.gz` variant, or similar, is not in use in Fedora, Arch, SLE12 and Leap 15. I think we are good here.

PR #1310, opened 2020-01-09
Start using tftp system user package

davidcassany:
With this commit we start requiring tftp system user package. This
user was created and managed by multiple packages before, with the
risk of having inconsistent criteria on its defaults. Now there there
a system user package so whatever package that requries this user should
just require this package and do not create or modify the tftp user.

Related with bsc#1143454

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -189,6 +189,9 @@ Requires(pre):  shadow-utils
 %else
 Requires(pre):  shadow
 %endif
+%if 0%{?suse_version} >= 1550
+Requires(pre):  user(tftp)
+%endif
 License:        GPL-3.0-or-later
 Group:          %{sysgroup}
 
@@ -391,7 +394,7 @@ done
 %fdupes %{buildroot}/srv/tftpboot
 %endif
 
-%if %{_vendor} != "debbuild"
+%if %{_vendor} != "debbuild" && 0%{?suse_version} < 1550
 %ifarch %{ix86} x86_64
 %pre -n kiwi-pxeboot
 #============================================================
@@ -447,7 +450,9 @@ fi
 %if %{_vendor} != "debbuild"
 %ifarch %{ix86} x86_64
 %files -n kiwi-pxeboot
+%if 0%{?suse_version} < 1550
 %dir %attr(0755,tftp,tftp) /srv/tftpboot
+%endif
 %dir /srv/tftpboot/KIWI
 %dir /srv/tftpboot/pxelinux.cfg
 %dir /srv/tftpboot/image


davidcassany:
> There is no tftp user package: https://build.opensuse.org/package/view_file/openSUSE:Factory:Staging:I/tftp/tftp.spec?expand=1

Yes there is https://build.opensuse.org/package/binary/openSUSE:Factory/system-users/standard/x86_64/system-user-tftp-20170617-8.2.noarch.rpm

It is a subpackage of `system-users` https://build.opensuse.org/package/show/openSUSE:Factory/system-users

davidcassany:
In fact tftp package should start using it instead of creating its own and avoid scriptlet's like

```spec
# This group/user is shared with atftp, so please
# keep this in sync with atftp.spec
# add group
%{_sbindir}/groupadd -r tftp 2>/dev/null || :
# add user
%{_sbindir}/useradd -c "TFTP account" -d /srv/tftpboot -G tftp -g tftp \
  -r -s /bin/false tftp 2>/dev/null || :
```

It turns that it was not only a matter of `atftp` and `tftp` packages, but also `dnsmasq` and `kiwi`, probably some more. Due to a conflict between kiwi and dnsmasq  we requested the creation of this user as part of `system-users` this way we could just require it instead of creating it in every single package using it.

package/python-kiwi-spec-template
```diff
@@ -391,7 +400,7 @@ done
 %fdupes %{buildroot}/srv/tftpboot
 %endif
 
-%if %{_vendor} != "debbuild"
+%if %{_vendor} != "debbuild" && %{without tftp_user_package}
```
Conan-Kudo: This actually needs to be `||`

davidcassany: I have the feeling I might miss something pretty obvious, but switching the logic operator to an OR (`||`) isn't it allowing to get into the if clause any non debian build regardless of the `tftp_user_package` variable value?

Conan-Kudo: hm, you're right.

Conan-Kudo: ~debian && ~tftp = debian || tftp

Is that what we want?

davidcassany:
> After reading all comments I understood the motivation and as we do it only for the suse version of the package I'm fine with it

We do it only for Tumbleweed or any suse version higher than SLE/LEAP 15.x.

package/python-kiwi-spec-template
davidcassany: I am not sure I understand your expression...

What we want in plain text is:

Execute the tftp user creation if, and only if:

1. The build is not for debian

AND

2. The build is not for Tumbleweed

The `%{without tftp_user_package}` is only active for **non** Tumbleweed (or anything higher than Leap/SLE 15.x) builds.

davidcassany:
@Conan-Kudo are you finally fine with this PR? Sorry but I am not sure if you are still requesting changes here or not.

Conan-Kudo:
@davidcassany I'm a little concerned that the conditional logic with the debbuild and suse conditions aren't right. Do we have a test somewhere to verify this behaves as intended?

davidcassany:
> @davidcassany I'm a little concerned that the conditional logic with the debbuild and suse conditions aren't right. Do we have a test somewhere to verify this behaves as intended?

I see, gonna prepare a test for that. In fact I did a test before Christmas (when I actually coded it), but gonna prepare another, it is not a big effort, I'll come back with it. We are not in hurry to merge it and I want to get it right.

In any case, would you write this logic different? We need to skip the user creation for Debian and Tumbleweed builds.

Conan-Kudo:
@davidcassany I think the conditional is actually correct, I just don't trust myself completely when I do AND like this. I can't think of a reason for this to be wrong.

davidcassany:
@Conan-Kudo @schaefi I simplified the implementation. In fact I did some tests and I realized I did not fully understand the `bcond` macros. Thus I opted to not use them and use simpler expressions on `%if` clauses.

I did a build of this code [here](https://build.opensuse.org/package/show/home:dcassany:KIWI-NG/python-kiwi)

Conan-Kudo:
@davidcassany What did you change? I see no updated push?

davidcassany:
> @davidcassany What did you change? I see no updated push?

Since https://github.com/OSInside/kiwi/commit/2ef9f6d04f7064c853477a6f9bea0eeaeb61543c two days ago there haven't been any changes. However your request for changes still stands, that's why we are asking for your feedback

Conan-Kudo:
@davidcassany The logical failure is that now the script isn't run for _any_ distribution where `%suse_version` is not defined. That includes *all* Fedora and CentOS/RHEL too.

davidcassany:
> @davidcassany The logical failure is that now the script isn't run for _any_ distribution where `%suse_version` is not defined. That includes _all_ Fedora and CentOS/RHEL too.

My understanding is that in no suse builds this evaluates to `0`, thus the condition effectively includes Fedora and others. In fact the common condition `0%{?suse_version} && 0%{?suse_version} < 1550` is to make sure this applies for SUSE but not for tumbleweed, as in any suse build `0%{?suse_version}` this will evaluate to something like `0X` being X the defined value, however it evalutes to just `0` when undefined (any non suse distro). So that by using `0%{?suse_version} < 1550` I am stating any non Tumbleweed build (including older SUSE ones and others like Fedora).

In fact you can easily verify Fedora packages from my testing build:

```bash
docker pull fedora:31
docker run --rm -ti fedora:31
cat /etc/passwd | grep tftp
dnf install 'dnf-command(config-manager)'
dnf config-manager --add-repo https://download.opensuse.org/repositories/home:/dcassany:/KIWI-NG/Fedora_31/home:dcassany:KIWI-NG.repo
dnf install kiwi-pxeboot
cat /etc/passwd | grep tftp
```

```bash
docker pull opensuse/tumbleweed
docker run --rm -ti opensuse/tumbleweed
zypper ar zypper ar https://download.opensuse.org/repositories/home:/dcassany:/KIWI-NG/openSUSE_Tumbleweed/ KIWI_TW
zypper in kiwi-pxeboot
```

Conan-Kudo:
@davidcassany If that is the intent, then we can go with it. Otherwise, it's a bug.

davidcassany:
> @davidcassany If that is the intent, then we can go with it. Otherwise, it's a bug.

Yes this is the intended behavior: the `tftp` user being created by all distros besides Debian based and Tumbleweed (in which case it requires the `system-user-tftp` package).

PR #1311, opened 2020-01-12
Follow up fix for kernel version lookup

schaefi:
When using custom kiwi boot images a shell method named
baseCreateCommonKernelFile was used to create a common
kernel file name in the boot image to allow a simpler
search code in the builder. The search code in the builder
however uses the standard kernel naming schema and with
the latest changes to the kernel version lookup the
common name did not match the matching pattern anymore.
The use of such a common kernel file is obsolete and
the shell method can finally be deleted.

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -946,126 +946,33 @@ function baseUpdateModuleDependencies {
 }
 
 #======================================
-# baseCreateCommonKernelFile
-#--------------------------------------
-function baseCreateCommonKernelFile {
-    # /.../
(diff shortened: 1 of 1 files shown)


schaefi:
@davidcassany the legacy integration tests failed with the latest change to the kernel name lookup. I realized with that change we finally can delete the shell code that moved the kernel to a common file name (vmlinuz) if building with kiwi-ng but using the custom kiwi boot images (initrd).

With this fix here I have build the failing integration tests locally and also booted them (the oem one, not the pxe stuff)

PR #1312, opened 2020-01-15
Allow builds without KIWI dracut modules

### LINKED ISSUE #1300: Custom dracut modules are required to be called as their KIWI counterpart
## Problem description

It is not possible to build an image not including the required dracut modules since 07ea23a4. Having the options of building images without including the kiwi specific dracut modules is no longer possible as they are explicitly included within the dracut call.

In case the user aims to provide alternate dracut modules for oem repart or oem dump she/he is forced to overwrite the kiwi provided module by a custom one. Thus it is no possible to tell kiwi to just use a different module. They have to be called as the original ones, e.g. `kiwi-dump`.

This is not big issue as it is still possible to provide custom modules by overwriting the ones provided by kiwi. However, it feels like a limitation having to do so as in that case it would not be possible to hypothetically install both modules in some host, even I can't imagine a situation where this could actually happen at the moment.

## Expected behaviour

It is possible to disable the dracut modules runtime checks and build an oem image without having to explicitly include the dracut modules provided by kiwi (`kiwi-dump` and `kiwi-repart`).  

Probably they should be requested in initrd only if they are available on the root tree. The runtime checks already ensure the user is not missing them by mistake.

Disable runtime checks for dracut modules and try to build an oem image with `installpxe` and without the package `dracut-kiwi-oem-dump`. When creating the PXE initrd the build process is going to fail.

Adding some `omit_dracutmodules+=" module "` does not work, because any configuration file options can be overwritten from the command line.

+ KIWI version: 9.19.7

davidcassany:
This commit avoids requesting non installed dracut modules to the dracut
creation. It bumps a warning message if the module requested by KIWI is
not installed in the root-tree and ignores the request.

This allows the creation of images without including KIWI dracut modules
when the related runtime checks are disabled.

Fixes #1300

### DIFF
--- a/kiwi/boot/image/dracut.py
+++ b/kiwi/boot/image/dracut.py
@@ -46,6 +46,7 @@ class BootImageDracut(BootImageBase):
         self.install_modules = []
         self.omit_modules = []
         self.omit_install_modules = []
+        self.available_modules = self._get_modules()
 
     def include_file(self, filename, install_media=False):
         """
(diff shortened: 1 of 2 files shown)


davidcassany:
Having the runtime checks for dracut modules disabled and not including the dracut-kiwi-* packages results in a *successful* build including the following warnings in the logs:

```
[ WARNING ]: 11:00:59 | dracut module "kiwi-repart" not found in the root tree
[ WARNING ]: 11:00:59 | module "kiwi-repart" not included in initrd
```

PR #1317, opened 2020-01-17
Fixed order of fstab entries

schaefi:
If a volume manager is used the volumes are added before the
root filesystem(/) entry in fstab. This does not hurt because
at boot time systemd manages the mounting of the rootfs prior
to any other information in the fstab file but it's conceptually
broken. Users justifiably can expect the fstab entries in the
correct order such that mounting from top to bottom leads
to a consistent root filesystem state.

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -335,9 +335,6 @@ class DiskBuilder:
                 self.requested_filesystem
             )
             volume_manager.mount_volumes()
-            self.generic_fstab_entries += volume_manager.get_fstab(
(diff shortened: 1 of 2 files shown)


schaefi:
@davidcassany I realized this problem when working on the distribution migration system. This system does an offline upgrade from a live migration system and mounts the OS rootfs by reading the fstab from the system to upgrade.

schaefi:
@davidcassany what is really annoying is that I wrote a test that intentionally tested a wrong order :( shame on me :)

PR #1318, opened 2020-01-20
Fixed custom args check

schaefi:
If a custom argument exists in the dictionary but has no value
it should be treated as not set and initialized empty as intended


schaefi:
@davidcassany I realized this one is causing the microOS test to fail

Issue #1281, opened 2019-11-27
Code cleanup by using Path.move_to_root and Path.rebase_to_root


davidcassany:
## Problem description

With the addition of `Path.move_to_root` and `Path.rebase_to_root` methods some parts of the code are likely to be simplified.

Gonna list here the candidates for refactor:

- [x] kiwi/system/root_bind.move_to_root (it can be removed) #1319
- [x] kiwi/bootloader/install/grub2._get_tool_name #1321
- [x] kiwi/system/setup.setup_plymouth_splash #1321 
- [x] kiwi/defaults.*grub* (some methods to fine the image grub2 location) #1321

PR #1319, opened 2020-01-20
Remove RootBind.move_to_root method

davidcassany:
This commit removes the RootBind.move_to_root method as this can all be
done by using the Path.move_to_root utility method. This allows
to drop the RootBind attribute in PackageManager classes and focus
path manipulation methods into a common utility.

Related to #1281


schaefi:
@davidcassany I'll merge and create a new staging submission, let's check if all integration builds works with it

PR #1323, opened 2020-01-27
Fixed sat solver flags validation

schaefi:
The sat library from the python3-solv plugin does not expose
the flags information as method. Instead the flags value is
a variable pointing to an integer that has a name mapping
in self.solv.Selection from the library.


schaefi:
found this when playing with the solver code

PR #1324, opened 2020-01-28
Fixed ppc kernel lookup

schaefi:
On power the kernel is named e.g vmlinux-4.12.14-197.29-default
kiwi was missing that name match. This commit fixes it

### DIFF
--- a/kiwi/system/kernel.py
+++ b/kiwi/system/kernel.py
@@ -154,7 +154,7 @@ class Kernel:
             # depending on the arch and os they are different
             # in their prefix
             kernel_prefixes = [
-                'uImage', 'Image', 'zImage', 'vmlinuz', 'image'
+                'uImage', 'Image', 'zImage', 'vmlinuz', 'image', 'vmlinux'
             ]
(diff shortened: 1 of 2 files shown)


schaefi:
probably one out of some patches needed to get ppc enabled

schaefi:
I found this one by a short smoke test on the machine Mike gave us. I have several questions and will send them by e-mail, I'll include you in any of them to keep you up to date. This one though is pretty self explaining :-)

kiwi/system/kernel.py
```diff
@@ -154,7 +154,7 @@ def _setup_kernel_names_for_lookup(self):
             # depending on the arch and os they are different
             # in their prefix
             kernel_prefixes = [
-                'uImage', 'Image', 'zImage', 'vmlinuz', 'image'
+                'uImage', 'Image', 'zImage', 'vmlinuz', 'image', 'vmlinux'
```
davidcassany: Good :+1: Arch uses `vmlinux` too ;)

PR #1326, opened 2020-01-29
Disable RPM module content validation and filtering when building in OBS

Conan-Kudo:
The Open Build Service builds images by identifying the requested dependencies,
downloading them into an isolated environment, regenerating the repository
metadata from scratch with *only* that content, and then passing those
new repositories to be used for building images. This enforces the
reproducibility of the image build process.

However, when building images for Linux distributions that have
AppStreams/modules (such as Red Hat Enterprise Linux/CentOS 8)
in an Open Build Service system, the repository metadata associated
with modules is not present as OBS does not generate it.

This causes the image build to fail because the normal module
content filtering rules make it so that modular RPMs are disabled
unless there is module metadata in the repository that identifies
them and that the module has been configured to be enabled.

As it is not possible for us to satisfy those conditions, instead
we disable modular filtering entirely when we detect that the image
build is occurring inside the build service, as we are reasonably
certain that OBS will not give us bad or broken package sets.

### DIFF
--- a/kiwi/repository/dnf.py
+++ b/kiwi/repository/dnf.py
@@ -198,6 +198,13 @@ class RepositoryDnf(RepositoryBase):
             repo_config.set(
                 name, 'gpgcheck', '1' if pkg_gpgcheck else '0'
             )
+        if Defaults.is_buildservice_worker():
+            # when building in the build service, modular metadata is inaccessible...
+            # in order to use modular content in the build service, we need to disable
+            # modular filtering, which is done with module_hotfixes option
+            repo_config.set(
+                name, 'module_hotfixes', '1'
+            )
         with open(repo_file, 'w') as repo:
             repo_config.write(repo)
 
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
@davidcassany I think I just fixed those, we'll see if the CI agrees. 😉

Conan-Kudo:
> I am wondering what happens to the image after being built with packages from different modules that are not active by default (I assume the resulting image would get default modules when there isn't any sort of specific module configuration applied). Would this result on annoying and hard to understand conflicts in the rpmdb? Sorry I haven't test this features myself that much, I am a noob about rpm filters. Do you have some test around?

Thankfully, it doesn't cause insanity to the rpmdb, but if non-default modules are used, DNF isn't going to work in the appliance until the required modules are enabled. That _is_ tricky and I don't yet know how to solve that. I figure that it's better to unblock part of this now, which would give us the opportunity to figure out how to fix the rest.

schaefi:
good catch :+1:

Issue #1325, opened 2020-01-28
Add support for powerPC disk images


schaefi:
## Problem description

IBM Cloud systems are ppc64le based and boots through KVM. At the moment kiwi can't build
images for this target for several reasons which needs individual investigations each. We got a testing system from Mike Friesenegger which allows to run tests. This issue should be used as reference for all problem we will find.

## Expected behaviour

- [x] ppc64le supported for vmx image types
- [x] optional boot from real hardware if available
- [x] at least one integration test exists for the platform

Conan-Kudo:
We should make sure ppc64le works for both Fedora and openSUSE...

schaefi:
>    We should make sure ppc64le works for both Fedora and openSUSE...

I agree completely. Nothing we do should be distro exclusive. From the current
patch, do you see anything that limits it to a certain distro ?

I would love to also test on Fedora(ppc64le) but so far couldn't
come up with an easy way to setup a machine for testing. At the moment
I got a test system for sle12 setup by one of my colleagues who has
access to hardware

Thanks

Conan-Kudo:
@schaefi If you have a Fedora account and become a packager, you can use the Fedora dev/test resources for non-x86 arches: https://fedoraproject.org/wiki/Test_Machine_Resources_For_Package_Maintainers

schaefi:
Thanks, actually something I wanted to do some time ago. Done now. I did not join a group since I couldn't find something matching (imaging, virtualization, pubcloud) or similar does not exist. Also I do not maintain a fedora package so far. Maybe we clarify this elsewhere ? JFI: username is "osinside" :) all other attempts were taken

Conan-Kudo:
@schaefi Even `schaefi`?

Conan-Kudo:
@schaefi there is a cloud group, it's part of the Fedora Server WG: https://fedoraproject.org/wiki/Server

schaefi:
> @schaefi Even `schaefi`?

yes

schaefi:
I'm just reading to how this all works. For getting access to machines I think you have to be part
of the packager team: https://fedoraproject.org/wiki/SIGs/Package_Review

I'm thinking to become kind of a co maintainer to the kiwi2 package as you maintain it for Fedora if I got that all right.. but need to read all the joining information first

Conan-Kudo:
@schaefi That's weird, because I looked up `schaefi` and it said it didn't exist...

schaefi:
[...]

PR #1328, opened 2020-02-05
PPC integration tests

schaefi:
**Added suse and fedora integration tests for Power**
   Simple disk image templates to run in KVM

**Update status helper script**
    Also show new ppc integration test builds

This is related to #1325


schaefi:
* https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_ppc:suse

* https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_ppc:fedora

build-tests/ppc/fedora/test-image-vmx/config.sh
```diff
@@ -0,0 +1,49 @@
...
+
+#======================================
+# Greeting...
+#--------------------------------------
+echo "Configure image: [$kiwi_iname]..."
+
+#======================================
+# Mount system filesystems
+#--------------------------------------
+baseMount
```
davidcassany: I know this does not hurt, but I am confident this is not needed anymore (if needed it probably should be treated as a bug), the python code already ensure the mount points are ready.

build-tests/ppc/fedora/test-image-vmx/config.sh
```diff
@@ -0,0 +1,49 @@
...
+
+#======================================
+# Setup default target, multi-user
+#--------------------------------------
+baseSetRunlevel 3
+
+#======================================
+# Umount kernel filesystems
+#--------------------------------------
+baseCleanMount
```
davidcassany: The same here, python code already umounts whatever is needed the `config.sh` is executed.

build-tests/ppc/fedora/test-image-vmx/root/etc/login.defs
```diff
@@ -0,0 +1,59 @@
+# *REQUIRED*
```
davidcassany: Just curious to know from where you got this file or its contents. The rest of fedora config files here I have the feeling that with some patience and trials I could have produced them, but I am not that sure about this one. It is already provided by some package and just slightly patched?

build-tests/ppc/suse/test-image-vmx/config.sh
```diff
@@ -0,0 +1,89 @@
...
+
+#======================================
+# Greeting...
+#--------------------------------------
+echo "Configure image: [$kiwi_iname]..."
+
+#======================================
+# Mount system filesystems
+#--------------------------------------
+baseMount
```
davidcassany: Same as before

build-tests/ppc/suse/test-image-vmx/config.sh
```diff
@@ -0,0 +1,89 @@
...
+
+#======================================
+# Remove yast if not in use
+#--------------------------------------
+suseRemoveYaST
+
+#======================================
+# Umount kernel filesystems
+#--------------------------------------
+baseCleanMount
```
davidcassany: same as before

build-tests/ppc/fedora/test-image-vmx/config.sh
schaefi: you are absolutely right. It was a left over from copying that stuff around. I deleted them also the exit 0 should not be there

build-tests/ppc/fedora/test-image-vmx/root/etc/login.defs
schaefi: I don't know either :) I copied those from a fedora description of the past

schaefi: maybe Neal has an idea if we still need it

@Conan-Kudo thoughts ?

Thanks

Conan-Kudo: Why do we need any of these? Shouldn't this be provided by Fedora packages already?

schaefi: To be honest I blindly copied the files as they were present on the other fedora images we have in the kiwi-descriptions repo. I could imagine that these files are provided by fedora packages but maybe the config is adapted for the needs of an image. I'm happy to reduce the number of overlay files, really

schaefi: So you are saying we better drop it ?

schaefi: I dropped the file and will merge the test descriptions now. As they do not server a customer purpose we can adapt when needed.

Issue #1262, opened 2019-11-13
Replacement concept for PXE


schaefi:
What if we add a new image type that it only produces a rootfs, a kernel and initrd. Following the spirit of ARI/AKI/AMI. Basically a building a rootfs in a filesystem without any bootloader and producing a dracut initrd and extracting the kernel for it. This could potentially fit many different PXEs scenarios for which just the appropriate dracut modules should used.

I believe this could be a simple type of image in which we could easily bypass the specific use case requirements to some dracut module (coded by someone else, in cooperation with us, etc.) that is **not part of KIWI**. I forsee PXE or cases like [1] or [2].

Also some documentation for the new type would be required. I believe that we could work in two use cases:

* Booting it with qemu (using the `-kernel`, `-initrd` and `-append` flags)
* Booting it as PXE and deploy on RAM or HD (the idea would be to showcase how to code/configure your own custom deployment)

[1] https://jira.suse.com/browse/SLE-7088
[2] https://firecracker-microvm.github.io/

PR #1330, opened 2020-02-05
Add .append file for PXE images

### LINKED ISSUE #1327: Include kernel command line as an output for PXE builds
## Problem description

After building a PXE image one of the results should be a file containing the kernel command line to boot the image. Specially relevant to include image specific data like root UUID. It shoud be customizable by uisng `kernelcmdline` attribute.

davidcassany:
This commit creates the .append file for PXE images. This file
is only created for initrd based images and includes the
UUID of the rootfs and any parameter added by the user within
the 'kernelcmdline' attribute.

Fixes #1327

### DIFF
--- a/kiwi/builder/filesystem.py
+++ b/kiwi/builder/filesystem.py
@@ -41,6 +41,7 @@ class FileSystemBuilder:
     **Filesystem image builder**
 
     :param str label: filesystem label
+    :param str root_uuid: UUID of the created filesystem (on block device only)
     :param str root_dir: root directory path name
     :param str target_dir: target directory path name
     :param str requested_image_type: configured image type
@@ -57,6 +58,7 @@ class FileSystemBuilder:
     """
     def __init__(self, xml_state, target_dir, root_dir):
         self.label = None
+        self.root_uuid = None
         self.root_dir = root_dir
         self.target_dir = target_dir
         self.requested_image_type = xml_state.get_build_type_name()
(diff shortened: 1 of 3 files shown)


davidcassany:
> The code looks good +1 Thanks, but I think we need to mention the new file in the docs somewhere

Yes good point, gonna add some docs here in this PR.

davidcassany:
> The code looks good +1 Thanks, but I think we need to mention the new file in the docs somewhere

@schaefi I have been thinking about it and I have some doubts... I am wondering how to document that since this is only for PXE images with `initrd_system="dracut"` and we have no documentation for PXE dracut based. All the documentation is for netboot code.

I'd say all this belongs to the #1262 and its documentation. I would add a new documentation page for the new type (which is, in fact, PXE dracut based image) and cover there some example/use case and include `.append` file explanation there. I don't want to mess with the old documentation of netboot code, I have the feeling I am going to create more confusion than anything else.

So I would merge this PR without docs and make the new documentation part of #1262 story. IMHO it makes more sense. Is it ok for you?

schaefi:
> and we have no documentation for PXE dracut based. All the documentation is for netboot code.

yes only in case of installpxe set for oem images we have documentation that references the dracut pxe code.

>I'd say all this belongs to the #1262 and its documentation. I would add a new documentation >page for the new type 

I agree that makes more sense :+1:

PR #1332, opened 2020-02-07
Move gitlab build_doc target to gitlab registry

schaefi:
Use the container from the kiwi-ci-containers project to
run the build_doc target

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,9 +2,10 @@ stages:
   - test
   - package
 
-image: fedora:latest
 variables:
   LANG: en_US.UTF-8
+  BUILD_IMAGES_PROJECT: kiwi3/kiwi-ci-containers
+  TUMBLEWEED_BUILD: buildenv-tumbleweed
 
 .install_deps: &install_system_deps
   before_script:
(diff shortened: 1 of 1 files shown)


schaefi:
I've added custom build envs in a new project on gitlab:

* https://gitlab.com/kiwi3/kiwi-ci-containers/container_registry

schaefi:
>Just curiosity how is being rebuild the image? does it happens on demand or there is some sort of >automation?

As far as I understood there is no auto push to the registry if the base container changes. In the
.gitlab-ci.yml I have this setup:

```yml
openSUSE/Tumbleweed:
  image: docker:latest
  tags:
  - shared
  - docker
  script:
  - ./build.sh $TUMBLEWEED_BUILD docker-opensuse-tumbleweed $CI_REGISTRY $PROJECT_GROUP
```

That build.sh script pushes the image to the docker registry if it runs from the master
branch. As far as I know gitlab does not trigger this pipeline target if the reference container in the Dockerfile changes its content. I'm not 100% sure though. But we should see this in the
pipeline.

If I look at https://gitlab.com/kiwi3/kiwi-ci-containers/pipelines I see a rebuild only if I did a change

schaefi:
I plan to add the fedora container which we use for the other tests as well. It speed up things a lot if you don't need to install additional software and can use the container just as it is. Also the ci file on the kiwi project will become a much easier read

PR #1333, opened 2020-02-08
Allow custom modules to be run after image dump

phoenixflare:
Currently the kiwi-dump module boots the installed system right after the image dump and verification.
This means that modules cannot be added after the image dump.

The kiwi-dump module has been split into kiwi-dump and kiwi-dump-reboot. This allows custom modules to be run after the image dump has been completed and before booting into the installed system, which effectively ends the install process.

### DIFF
--- a/doc/source/working_with_kiwi/customize_the_boot_process.rst
+++ b/doc/source/working_with_kiwi/customize_the_boot_process.rst
@@ -21,6 +21,10 @@ of KIWI the following dracut modules are used:
   This module is required if one of the attributes `installiso`, `installstick`
   or `installpxe` is set to `true` in the image type definition
 
+``kiwi-dump-reboot``
+  Serves to boot the system into the installed image after installation is
+  completed.
+
 ``kiwi-live``
   Boots up a KIWI live image. This module is required
   if the `iso` image type is selected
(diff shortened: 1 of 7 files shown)


schaefi:
Hi, thanks much for your effort. I like the idea :+1: 

A few things that came to my mind

* You changed the builder code obviously to the module change, could you also adapt the unit tests which are now failing. Let us know if you need advise in this regard. Thanks

* We provide the dracut module in an rpm package dracut-kiwi-oem-dump. The spec file template in package/python-kiwi-spec-template needs adaption too. Please let us continue to provide both modules in the same package, since we have to care for backward compatibility on the image descriptions that installs it

Thanks

phoenixflare:
Uh woops. Missed an fi at the end of the script. Just need to fix up the unit test to work with the new addition...

phoenixflare:
Should be good now. Could you double check the spec template for me to ensure I added it correctly?

schaefi:
Looks perfect to me, thanks. Just one thing. Could you merge your commits into one, then we are ready to go :)

phoenixflare:
Not sure what the proper way to do this is but hey it's done.

PR #1334, opened 2020-02-09
PPC oem integration test

schaefi:
Added suse OEM ppc integration test
    
Added obs integration test for building a simple disk image
to be started in a VM on power. Related to Issue #1325

Along with this change a change in kiwi was needed to support iso creation on ppc.
I can't tell if that makes sense for bare metal but it at least enables the builder to
produce iso/install iso files


schaefi:
See the test including the iso patch building the image here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_ppc:suse/test-image-oem

PR #1335, opened 2020-02-10
Move gitlab stages to use gitlab registry

schaefi:
The kiwi-ci-containers project on gitlab now builds all
containers that are needed to run all type of tests we
have in the gitlab pipeline for kiwi. This commit moves
the tests to fetch the containers from the gitlab
registry and avoids any requirement to install packages
prior to running the tests. This will speedup the
overall test time and makes the system also more robust
in case the repo servers hosting the packages can't
be accessed for some reason


schaefi:
first hack week outcome :-)

PR #1340, opened 2020-02-17
Turn btrfs requirement into a recommend

schaefi:
Only on suse systems btrfs is eligable to be a required package
because it's the default filesystem of that distribution. In any
other case it should be a recommended package. As a side effect
of this change we will be able to activate the CentOS-8 build
target

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -212,12 +212,16 @@ BuildRequires:  dracut
 Requires:       bc
 Requires:       cryptsetup
 %if 0%{?fedora} || 0%{?rhel} || 0%{?ubuntu} >= 1804 || 0%{?debian} >= 9
+%if 0%{?rhel} && 0%{?rhel} < 8
 Requires:       btrfs-progs
+%else
+Recommends:     btrfs-progs
+%endif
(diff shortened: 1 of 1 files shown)


package/python-kiwi-spec-template
```diff
@@ -212,12 +212,12 @@ BuildRequires:  dracut
 Requires:       bc
 Requires:       cryptsetup
 %if 0%{?fedora} || 0%{?rhel} || 0%{?ubuntu} >= 1804 || 0%{?debian} >= 9
-Requires:       btrfs-progs
+Recommends:     btrfs-progs
```
Conan-Kudo: Actually, we can't use `Recommends` on EL7, so this needs to be done slightly more carefully...

Conan-Kudo: Suggestion:

```rpm-spec
%if 0%{?rhel} && 0%{?rhel} < 8
Requires:       btrfs-progs
%else
Recommends:     btrfs-progs
%endif
```

schaefi: ah yes, I always forgot to remind me on that. Thanks and updated

PR #1341, opened 2020-02-17
Do not create empty macros file

### LINKED ISSUE #1316: /usr/lib/rpm/macros.d/macros.kiwi-bootstrap-config in every layer
When building a derived docker image, the `/usr/lib/rpm/macros.d/macros.kiwi-bootstrap-config` file is in all layers with the same content. It should be enough to just reuse the one in the lowest layer.

davidcassany:
This commit makes shure to empty RPM macros files are created during
the build.

Fixes #1316

### DIFF
--- a/kiwi/path.py
+++ b/kiwi/path.py
@@ -104,9 +104,10 @@ class Path:
 
         :param string path: path name
         """
-        Command.run(
-            ['mkdir', '-p', path]
-        )
+        if not os.path.exists(path):
+            Command.run(
+                ['mkdir', '-p', path]
+            )
 
     @staticmethod
     def wipe(path):
@@ -115,9 +116,10 @@ class Path:
 
         :param string path: path name
         """
-        Command.run(
-            ['rm', '-r', '-f', path]
-        )
+        if os.path.exists(path):
+            Command.run(
+                ['rm', '-r', '-f', path]
+            )
 
     @staticmethod
     def remove(path):
--- a/kiwi/utils/rpm.py
+++ b/kiwi/utils/rpm.py
@@ -98,10 +98,11 @@ class Rpm:
 
         Write bootstrap macro file to the custom rpm macros path
         """
-        Path.create(self.macro_path)
-        with open(self.macro_file, 'w') as macro:
-            for entry in self.custom_config:
-                macro.write('{0}{1}'.format(entry, os.linesep))
+        if self.custom_config:
+            Path.create(self.macro_path)
+            with open(self.macro_file, 'w') as macro:
+                for entry in self.custom_config:
+                    macro.write('{0}{1}'.format(entry, os.linesep))
 
     def wipe_config(self):
         """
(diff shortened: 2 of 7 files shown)


davidcassany:
@scheafi, could you have a look a again? Just did a very small change to include the directory creation inside the `if` clause, this way no spurious folder is created when there are no custom macros defined.

Thanks

schaefi:
Still looks ok to me. There is one reference for self.macro_path in self.macro_file which is used in the wipe_config() method. But it does not hurt if neither the file nor the directory is present when the cleanup is called.

If we want to do this super clean you can add an os.path.exists() to the path.wipe() method prior calling the rm command. That would be good imho also for other wipe calls. It safes is 0.00001ms :)

davidcassany:
> That would be good imho also for other wipe calls. It safes is 0.00001ms :)

Allright :) lets be clean and symetric.

davidcassany:
All right now I added the `os.path.exists` check in `Path.wipe` and `Path.create`. The benefit of it, rather than a potential ridiculous optimization is that we will get cleaner logs due to this change, needless directory creations and paths deletions won't be logged.

schaefi:
ok let's hope the macro change did not break anything :crossed_fingers: 

Thanks David

Vogtinator:
I wonder how this works - both the changes to `rm` and `mkdir` are AFAICT noops - they might change the atime, but not mtime.

davidcassany:
The change in `Path.create` and `Path.wipe` (thus the use of  `rm` and `mkdir`) it only affects the logs. With the change of this PR you will not see needless `mkdir` and `rm` records. The change that is actually affecting #1316 are visible here: https://github.com/OSInside/kiwi/blob/2b30e8d480d1238b4fa1e9f37a877a8359148bdd/kiwi/utils/rpm.py#L102-L105 
This way the macros files are not created unless they content some value, before we were creating empty files when there was not specific configuration to be written. In the tests I did this was solving the inclusion of the mentioned folder as the files written there were empty because the host and image were having an equivalent rpmdb configuration.

Vogtinator:
Ok, so `custom_config` is empty in the common case?

davidcassany:
> Ok, so `custom_config` is empty in the common case?

Speaking from memory, the macros depend on locale configuration (for the `%_install_langs`) and depend on rpmdb path match/mismatch on host vs image.

Vogtinator:
Does that mean if the kiwi file sets a locale, it'll still write the file even if it would not change anything?

davidcassany:
The `%_install_langs` macro is only applied if `<rpm-locale-filtering>` is set to `true`.

Vogtinator:
If this is set in both base and derived images, wouldn't this result in having this duplicated again?

davidcassany:
> If this is set in both base and derived images, wouldn't this result in having this duplicated again?

It should only duplicate the directory, because during the build a file gets included and deleted, in that process the mtime of the directory changes and I guess this is tracked by `umoci`. I don't think we can do more regarding that in KIWI.

Vogtinator:
Yes, IMO good enough. I wonder though whether ignoring mtime changes would allow some significant space savings, but I guess not much.

PR #1342, opened 2020-02-18
Fixup documentation for consistency

schaefi:
There is the legacy kiwi version and there is this kiwi(next generation).
From a documentation perspective there are several inconsistencies that
could confuse users. This commit makes the name for KIWI-NG consistent
across the entire documentation. At places where we point to older
documentation we use the term Legacy KIWI and a link to the documentation
that covers this part. All this is needed in preparation to cleanup the
documentation situation for the SUSE documentation but with respect to
the upstream doc sources, their layout and markup.

### DIFF
--- a/doc/source/building/build_pxe_root_filesystem.rst
+++ b/doc/source/building/build_pxe_root_filesystem.rst
@@ -12,7 +12,7 @@ Build a PXE Root File System Image
 .. sidebar:: Abstract
 
    This page explains how to build a file system image for use with
-   KIWI's PXE boot infrastructure. It contains:
+   {kiwi}'s PXE boot infrastructure. It contains:
 
    * how to build a PXE file system image
    * how to setup the PXE file system image on the PXE server
@@ -22,11 +22,11 @@ Build a PXE Root File System Image
 implementations. The protocol sends a DHCP request to get an IP
 address. When an IP address is assigned, it uses the `TFTP`_ protocol
 to download a Kernel and boot instructions. Contrary to other images
-built with KIWI, a PXE image consists of separate boot, kernel and root
+built with {kiwi}, a PXE image consists of separate boot, kernel and root
 filesystem images, since those images need to be made available in
 different locations on the PXE boot server.
 
-A root filesystem image which can be deployed via KIWI’s PXE
+A root filesystem image which can be deployed via {kiwi}'s PXE
 netboot infrastructure represents the system rootfs in a linux
 filesystem. A user could loop mount the image and access the
 contents of the root filesystem. The image does not contain
@@ -61,14 +61,14 @@ system. As diskless client, a QEMU virtual machine is used.
 
    * The PXE root filesystem image approach is considered to be a
      legacy setup. The required netboot initrd code will be maintained
-     outside of the KIWI appliance builder code base. If possible,
+     outside of the {kiwi} appliance builder code base. If possible,
      we recommend to switch to the OEM disk image deployment via
      PXE.
 
 1. Make sure you have checked out the example image descriptions,
    see :ref:`example-descriptions`.
 
-2. Build the image with KIWI:
+2. Build the image with {kiwi}:
 
    .. code:: bash
 
@@ -128,7 +128,7 @@ system. As diskless client, a QEMU virtual machine is used.
    to find the image and how to activate it. In this case the image
    will be deployed into a ramdisk (ram1) and overlay mounted such
    that all write operations will land in another ramdisk (ram2).
-   KIWI supports a variety of different deployment strategies based
+   {kiwi} supports a variety of different deployment strategies based
    on the rootfs image created beforehand. For details, refer
    to :ref:`pxe_client_config`
 
--- a/doc/source/commands.rst
+++ b/doc/source/commands.rst
@@ -1,9 +1,9 @@
-KIWI Commands
-=============
+Command Line
+============
 
 .. note::
 
-   This document provides a list of the existing KIWI commands
+   This document provides a list of the existing {kiwi-product} commands
    for version |version|.
 
 .. toctree::
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -1,8 +1,7 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #
-# kiwi documentation build configuration file, created by
-# sphinx-quickstart on Fri Feb  5 11:03:18 2016.
+# KIWI NG documentation build configuration file
 #
 import sys
 from os.path import abspath, dirname, join, normpath
@@ -74,11 +73,14 @@ prolog_replacements = {
     '{exc_repo}': 'obs://openSUSE:Leap:15.1/standard',
     '{exc_kiwi_repo}':
         'obs://Virtualization:Appliances:Builder/openSUSE_Leap_15.1',
-    '{schema_version}': '7.1'
+    '{schema_version}': '7.1',
+    '{kiwi}': 'KIWI NG',
+    '{kiwi-product}': 'KIWI Next Generation (KIWI NG)',
+    '{kiwi-legacy}': 'KIWI Legacy'
 }
 
 latex_documents = [
-    ('index', 'kiwi.tex', 'KIWI Documentation', 'Marcus Schäfer', 'manual')
+    ('index', 'kiwi.tex', 'KIWI NG Documentation', 'Marcus Schäfer', 'manual')
 ]
 latex_elements = {
     'papersize': 'a4paper',
@@ -117,8 +119,8 @@ master_doc = 'index'
 default_role="py:obj"
 
 # General information about the project.
-project = 'kiwi'
-copyright = '2019, Marcus Schäfer'
+project = 'KIWI NG'
+copyright = '2020, Marcus Schäfer'
 author = 'Marcus Schäfer'
 
 # The version info for the project you're documenting, acts as replacement for
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -1,23 +1,26 @@
 .. kiwi documentation master file
 
-KIWI Documentation
-==================
+{kiwi} Documentation
+=====================
 
-.. note::
+.. note:: {kiwi} vs. {kiwi-legacy}
 
-   This documentation covers KIWI |version|- the command line utility to
-   build Linux system appliances. For versions of KIWI older or equal to
-   v7.x.x please refer to `Legacy KIWI <https://doc.opensuse.org/projects/kiwi/doc>`__
-
+   This documentation covers {kiwi-product} |version|- the command line
+   utility to build Linux system appliances. {kiwi} is stable and all
+   new features, bugfixes, and improvements will be developed here.
+   Versions older or equal to v7.x.x are out of maintenance and do
+   not get any updates or bugfixes. If you still need this version,
+   refer to the documentation for
+   `{kiwi-legacy} <https://doc.opensuse.org/projects/kiwi/doc>`__
 
 .. toctree::
    :maxdepth: 1
 
    quickstart
    installation
+   overview
    working_with_kiwi
    working_with_images
-   overview
    building
    commands
    development
@@ -50,11 +53,11 @@ a pre-configured application for a specific use case. The appliance is
 provided as an image file and needs to be deployed to, or activated in
 the target system or service.
 
-KIWI can create appliances in various forms: beside classical installation
+{kiwi} can create appliances in various forms: beside classical installation
 ISOs and images for virtual machines it can also build images that boot via
 PXE or Vagrant boxes.
 
-In KIWI, the appliance is specified via a collection of human readable files
+In {kiwi}, the appliance is specified via a collection of human readable files
 in a directory, also called the `image description`. At least one XML file
 :file:`config.xml` or :file:`.kiwi` is required. In addition there may as
 well be other files like scripts or configuration data.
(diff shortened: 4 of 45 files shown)


doc/source/working_with_images/pxe_client_configuration.rst
```diff
@@ -6,13 +6,13 @@ PXE Client Setup Configuration
...
    This page provides further information for handling
-   PXE images built with KIWI and references the following
+   PXE images built with {kiwi} and references the following
    articles:
 
    * :ref:`build_pxe`
 
 All PXE boot based deployment methods are controlled by configuration files
-located in :file:`/srv/tftpboot/KIWI` on the PXE server. Such a configuration
+located in :file:`/srv/tftpboot/{kiwi}` on the PXE server. Such a configuration
```
davidcassany: I doubt this is correct. This should not be translated to `KIWI-NG`, right?

doc/source/working_with_images/pxe_client_configuration.rst
```diff
@@ -42,8 +42,8 @@ The configuration file has got the following general format:
 
     CONF="src;dest;srvip;bsize;[hash],...,src;dest;srvip;bsize;[hash]"
 
-    KIWI_BOOT_TIMEOUT="seconds"
-    KIWI_KERNEL_OPTIONS="opt1 opt2 ..."
+    {kiwi}_BOOT_TIMEOUT="seconds"
+    {kiwi}_KERNEL_OPTIONS="opt1 opt2 ..."
```
davidcassany: Same as above, those variables should not be translated to `KIWI-NG`

doc/source/working_with_images/pxe_client_configuration.rst
```diff
@@ -207,7 +207,7 @@ boot of the client the following setup is required:
 
 .. code:: bash
 
-   KIWI_KERNEL_OPTIONS="nomodeset"
+   {kiwi}_KERNEL_OPTIONS="nomodeset"
```
davidcassany: Same as above

doc/source/working_with_images/pxe_client_configuration.rst
```diff
@@ -223,7 +223,7 @@ the following setup is required.
 
 .. code:: bash
 
-   KIWI_BOOT_TIMEOUT="10"
+   {kiwi}_BOOT_TIMEOUT="10"
```
davidcassany: Same as above

doc/source/building/build_pxe_root_filesystem.rst
```diff
@@ -118,7 +118,7 @@ system. As diskless client, a QEMU virtual machine is used.
 
    .. code:: bash
 
-       $ vi /srv/tftpboot/KIWI/config.default
+       $ vi /srv/tftpboot/{kiwi}/config.default
```
davidcassany: I believe this should not be translated to `KIWI-NG`, just `kiwi`.

doc/source/building/build_pxe_root_filesystem.rst
```diff
@@ -105,9 +105,9 @@ system. As diskless client, a QEMU virtual machine is used.
 
    .. code:: bash
 
-       DEFAULT KIWI-Boot
+       DEFAULT {kiwi}-Boot
```
davidcassany: I guess the change doesn't really make a difference here. However I'd prefer to not use a variable here.

doc/source/working_with_kiwi/legacy_kiwi.rst
```diff
@@ -39,25 +39,25 @@ In order to address all of these the questions came up:
...
 
-This provides an opportunity for users to test the next generation KIWI
+This provides an opportunity for users to test the next generation {kiwi}
 with their appliance descriptions without risk. If it builds and works
-as expected, I recommend to switch to the next generation KIWI. If not,
+as expected, I recommend to switch to the next generation {kiwi}. If not,
 please open an issue on https://github.com/OSInside/kiwi.
 
-The legacy KIWI version will be further developed in maintenance mode.
+The legacy {kiwi} version will be further developed in maintenance mode.
```
davidcassany: I guess this is inconsistent `legacy KIWI-NG` :wink:

doc/source/working_with_kiwi/legacy_kiwi.rst
```diff
@@ -39,25 +39,25 @@ In order to address all of these the questions came up:
...
 There won't be any new features added in that code base though.
 Packages will be available at the known place:
-`Legacy KIWI packages <http://download.opensuse.org/repositories/Virtualization:/Appliances>`__
+`Legacy {kiwi} packages <http://download.opensuse.org/repositories/Virtualization:/Appliances>`__
```
davidcassany: Same as above, review the hole page, there are many `legacy {kiwi}` or `next generation {kiwi}` in this one. Probably we should rephrase some of those.

schaefi:
Oh Oh yes you are right, I will fix all of those. Thanks much for the detailed check

doc/source/index.rst
```diff
@@ -1,11 +1,11 @@
...
-==================
+{kiwi} Documentation
+=====================
 
 .. note::
 
-   This documentation covers KIWI |version|- the command line utility to
+   This documentation covers {kiwi} |version|- the command line utility to
    build Linux system appliances. For versions of KIWI older or equal to
    v7.x.x please refer to `Legacy KIWI <https://doc.opensuse.org/projects/kiwi/doc>`__
```
tomschr: I would rephrase the note to emphasize the difference between the two:


```rst
.. note:: KIWI-NG vs. Legacy KIWI

     This documentation covers {kiwi-full} ({kiwi}) |version| the command line
     utility to build Linux system appliances.
     {kiwi} is stable and all new ideas, bugfixes, and improvements
     will be developed here.
     Older {kiwi} versions older or equal to v7.x.x are out of
     maintenance and do not get any updates or bugfixes.
     If you still need this verson, refer to the documentation for
     Legacy KIWI.
```

doc/source/commands.rst
```diff
@@ -1,9 +1,9 @@
-KIWI Commands
-=============
+Command Line
+============
 
 .. note::
 
-   This document provides a list of the existing KIWI commands
+   This document provides a list of the existing {kiwi} commands
```
tomschr: As this will be a manpage at some point, I would add a short explanation about our acronym "NG", using the "full" name with the `{kiwi-full}` placeholder:

```suggestion
   This document provides a list of the existing {kiwi-full} ({kiwi}) commands
```

doc/source/conf.py
```diff
@@ -74,11 +73,12 @@ def setup(app):
...
     '{exc_kiwi_repo}':
         'obs://Virtualization:Appliances:Builder/openSUSE_Leap_15.1',
-    '{schema_version}': '7.1'
+    '{schema_version}': '7.1',
+    '{kiwi}': 'KIWI-NG'
 }
 
 latex_documents = [
-    ('index', 'kiwi.tex', 'KIWI Documentation', 'Marcus Schäfer', 'manual')
+    ('index', 'kiwi.tex', 'KIWI-NG Documentation', 'Marcus Schäfer', 'manual')
```
tomschr:
```suggestion
    ('index', 'kiwi.tex', 'KIWI NG Documentation', 'Marcus Schäfer', 'manual')
```

doc/source/conf.py
```diff
@@ -117,8 +117,8 @@ def setup(app):
 default_role="py:obj"
 
 # General information about the project.
-project = 'kiwi'
-copyright = '2019, Marcus Schäfer'
+project = 'KIWI-NG'
```
tomschr:
```suggestion
project = 'KIWI NG'
```

doc/source/conf.py
```diff
@@ -74,11 +73,12 @@ def setup(app):
     '{exc_repo}': 'obs://openSUSE:Leap:15.1/standard',
     '{exc_kiwi_repo}':
         'obs://Virtualization:Appliances:Builder/openSUSE_Leap_15.1',
-    '{schema_version}': '7.1'
+    '{schema_version}': '7.1',
+    '{kiwi}': 'KIWI-NG'
```
tomschr: From our native speaker, she would replace the dash with a space. Additionally, I would also define the full name, without the "NG" acronym:

```suggestion
    '{kiwi}': 'KIWI NG',
    '{kiwi-full}': 'KIWI Next Generation',
```

doc/source/working_with_images/pxe_client_configuration.rst
schaefi: fixed

doc/source/working_with_kiwi/legacy_kiwi.rst
schaefi: yeah this is just stupid :-) fixed now

doc/source/commands.rst
schaefi: yes done and also updated the man pages. kiwi-ng is the actual name for the binary. Other names (kiw, kiwi-ng-3) are symlinks that can be used too. But the man pages should use the main name

doc/source/index.rst
schaefi: I like it and used parts of it. Thanks

schaefi:
@davidcassany @tomschr I think I have covered all issues, could you take a look again ? Thanks

doc/source/working_with_kiwi/legacy_kiwi.rst
```diff
@@ -1,7 +1,7 @@
 .. _legacy_kiwi:
 
-Legacy KIWI vs. Next Generation
-===============================
+Legacy KIWI vs. {kiwi-product}
+==============================================
```
tomschr: You could remove some of the equal signs. :wink:

doc/source/working_with_kiwi/legacy_kiwi.rst
```diff
@@ -39,25 +40,25 @@ In order to address all of these the questions came up:
...
+the same appliance description with the {kiwi-legacy} and the 
+{kiwi-product}, if the distribution and all configured features are
+supported by the used version.
 
-This provides an opportunity for users to test the next generation KIWI
+This provides an opportunity for users to test the {kiwi-product}
 with their appliance descriptions without risk. If it builds and works
-as expected, I recommend to switch to the next generation KIWI. If not,
+as expected, I recommend to switch to the {kiwi-product}. If not,
 please open an issue on https://github.com/OSInside/kiwi.
```
tomschr: In my opinion, `{kiwi-product}` should be used to *introduce* the term. After you have used it, you just use `{kiwi}`.

doc/source/working_with_kiwi/legacy_kiwi.rst
schaefi: no you can't. {kiwi-product} is replaced with content prior the sphinx rendering. The content string is longer than the macro. If you delete the signs you break the rendering

doc/source/working_with_kiwi/legacy_kiwi.rst
```diff
@@ -39,25 +40,25 @@ In order to address all of these the questions came up:
...
 
-The legacy KIWI version will be further developed in maintenance mode.
+The {kiwi-legacy} version will be further developed in maintenance mode.
 There won't be any new features added in that code base though.
 Packages will be available at the known place:
-`Legacy KIWI packages <http://download.opensuse.org/repositories/Virtualization:/Appliances>`__
+`{kiwi-legacy} packages <http://download.opensuse.org/repositories/Virtualization:/Appliances>`__
 
 When Do I need to use the legacy kiwi
```
tomschr:
```suggestion
When Do I need to use the {kiwi-legacy}
```

doc/source/working_with_kiwi/legacy_kiwi.rst
schaefi: This applies to all other chapters but not the explanations we use in this legacy chapter. I suggest to build it and read the result. For me that looked ok

doc/source/working_with_kiwi/legacy_kiwi.rst
```diff
@@ -141,15 +142,15 @@ generation KIWI.
...
-the next generation KIWI does not provide, the build will fail with an
+the {kiwi-product} does not provide, the build will fail with an
 exception early. If the appliance description uses next generation
 features like the selection of the initrd system, it's not possible to
-build that with the legacy KIWI, unless the appliance description
+build that with the {kiwi-legacy}, unless the appliance description
 properly encapsulates the differences into a profile.
 
-The next generation KIWI also provides the `--compat` option and
+The {kiwi-product} also provides the `--compat` option and
```
tomschr: Again, the product is already introduced so `{kiwi}` should be enough distinction against Legacy KIWI.

tomschr:
@schaefi Looks great, I just found some minor details you can think about.

doc/source/working_with_kiwi/legacy_kiwi.rst
schaefi: sorry I got it wrong, yes I agree will update the text

schaefi:
@davidcassany @tomschr how about it now ? can we go from here

tomschr:
> @tomschr how about it now ? can we go from here

I've built it and it looks good. There is one thing which was a bit irritating. Not sure if I got it right, but... the main script name has been changed to `kiwi-ng` and there is only a link to `kiwi` right?

So your intention is that the user calls the script by  `kiwi-ng`  (and not `kiwi`), correct? In that case you might want to adapt the help output.

I'm looking at the file `kiwi/cli.py` and the file docstring contains `kiwi` instead of `kiwi-ng`. Maybe you want to change that. :wink:

Other than that, I haven't found anything else. So go for it! :rocket:

schaefi:
>I'm looking at the file kiwi/cli.py and the file docstring contains kiwi instead of kiwi-ng. Maybe you want to change that. ￼

That's hard to change because the docstring is used with docopt and the api structure is kiwi/... not kiwi-ng. The name kiwi-ng comes from setup.py

```python
'entry_points': {
        'console_scripts': [
            'kiwi-ng=kiwi.kiwi:main',
            'kiwicompat=kiwi.kiwi_compat:main'
        ]
    }
```

So don't get me wrong but I rather risk this type of inconsistency before I break the entire cli handling. Hope that makes sense to you too ?

schaefi:
There are also more docopt strings. See the code in kiwi/tasks/*

schaefi:
@davidcassany What's is your opinion on that ? The man pages with this changes here will document the kiwi call as "kiwi-ng ..." and therefore matches the entry point from setup.py. The docstrings relevant for docopt needs to follow the code structure iirc which is kiwi/... Thus a kiwi-ng --help or a kiwi-ng-3 --help or a kiwi --help will only talk about "kiwi"

I thought too long about this... so happy if someone else has some input

schaefi:
maybe we just revert the kiwi-ng change in the man pages...

tomschr:
> So don't get me wrong but I rather risk this type of inconsistency before I break the entire cli handling. Hope that makes sense to you too ?

Ahh right! Thanks for the explanation! :+1: I didn't realize that. If that is too difficult to change, then just leave it as is for the time being. We can change that (if really wanted) in a different pull request.

PR #1343, opened 2020-02-19
Use GPT table for Power ofw firmware

schaefi:
GPT partition table should be preferred for power systems these days

### DIFF
--- a/kiwi/firmware.py
+++ b/kiwi/firmware.py
@@ -70,10 +70,7 @@ class FirmWare:
             else:
                 return 'msdos'
         elif 'ppc64' in self.arch:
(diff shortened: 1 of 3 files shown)


schaefi:
I have to double check if this is the correct partition ID for the prep partition

```
   1            2048            4095   8.0 MiB     4100  p.prep
```

schaefi:
should be ok

schaefi:
The obs oem build also succeeded

davidcassany:
@schaefi Oh sorry I did not include my approval into the PR as I was waiting for the confirmation on partition ID, just in case there was a further commit.

PR #1345, opened 2020-02-21
Make squashfs compression configurable

### LINKED ISSUE #1315: Add compression algorithm options to squashfs images
## Problem description
Creating squashfs images (for livecd images) is quite slow due to using the XZ compression algorithm.
For development purposes, disabling compression will make experimentation and working with kiwi much faster.
## Expected behaviour
Having an option in the XML, for example "compression=gzip" or "compression=none". 

- Just build any image with squashfs

+ KIWI version: 9.19.8
+ Operating system: SLES 15.1
+ OBS version: None

davidcassany:
This commit adds the `squashfscompression` attribute in type element. It
can take `gzip`, `xz`, `lzo`, `lz4` or `none`. The default is `xz`.

Fixes #1315

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -435,7 +435,11 @@ class DiskBuilder:
         if self.root_filesystem_is_overlay:
             squashed_root_file = NamedTemporaryFile()
             squashed_root = FileSystemSquashFs(
-                device_provider=None, root_dir=self.root_dir
+                device_provider=None, root_dir=self.root_dir,
+                custom_args={
+                    'compression':
+                        self.xml_state.build_type.get_squashfscompression()
+                }
             )
             squashed_root.create_on_file(
                 filename=squashed_root_file.name,
@@ -748,7 +752,11 @@ class DiskBuilder:
             log.info('--> creating readonly root partition')
             squashed_root_file = NamedTemporaryFile()
             squashed_root = FileSystemSquashFs(
-                device_provider=None, root_dir=self.root_dir
+                device_provider=None, root_dir=self.root_dir,
+                custom_args={
+                    'compression':
+                        self.xml_state.build_type.get_squashfscompression()
+                }
             )
             squashed_root.create_on_file(
                 filename=squashed_root_file.name,
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -254,7 +254,11 @@ class LiveImageBuilder:
         live_container_image = FileSystem(
             name='squashfs',
             device_provider=None,
-            root_dir=self.live_container_dir
+            root_dir=self.live_container_dir,
+            custom_args={
+                'compression':
+                    self.xml_state.build_type.get_squashfscompression()
+            }
         )
         container_image = NamedTemporaryFile()
         live_container_image.create_on_file(
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -1388,6 +1388,15 @@ div {
             sch:param [ name = "attr" value = "filesystem" ]
             sch:param [ name = "types" value = "vmx oem pxe" ]
         ]
+	k.type.squashfscompression.attribute = 
+        ## Specifies the compression type for mksquashfs
+        attribute squashfscompression {
+            "uncompressed" | "gzip" | "lzo" | "lz4" | "xz" | "zstd"
+        }
+        >> sch:pattern [ id = "squashfscompression" is-a = "image_type"
+            sch:param [ name = "attr" value = "squashfscompression" ]
+            sch:param [ name = "types" value = "vmx oem pxe iso" ]
+        ]
     k.type.overlayroot.attribute =
         ## Specifies to use an overlay root system consisting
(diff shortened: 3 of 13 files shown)


davidcassany:
This PR might be a bit confusing, let me add some explanation. Squashfs is used for three different builders: `install`, `live` and `filesystem`. For the first two squashfs is used in conjunction with other filesystems like the underlying filesystem inside the squashfs image and the ISO filesystem containing the squashfs image.

In schema we already had the attribute `createfsoptions` which is a free text attribute to include all the custom flags we want for the filesystem creation, however this affects to the disk creation for oem images and it affects to the iso creation for live images. Only PXE and filesystems builds can use this attribute to control the `mksquash` call. Thats the reasoning to add a new attribute in schema.

The new attribute is not a free text input, it is just used to select the compression type. `mksquash` does not support no compression at all, however there are some options to disable the compression of certain specific parts (inode table, data blocks, fragment blocks and extended attributes). If the user selects the `none` compression all those specific parts will be uncompressed. I am not aware if there is anything else that is eligible for compression in squashfs, if so, it will use the `mksquashfs` default compression algorithm (currently `gzip` if no `-comp` flag is provided).

In case the user is building PXE or filesystem images for `squashfs` and it is using both `createfsoptions` and `squashfscompression` attributes, the `createfsoptions` takes precedence when the specific flags are colliding (e.g. `createfsoptions="-comp gzip"` *and* `squashfscompression="lzo"`, results to the use of `gzip`). This is a really corner case as it doesn't apply for the common squashfs use cases (OEM and ISO images).

kiwi/builder/disk.py
```diff
@@ -435,7 +435,11 @@ def create_disk(self):  # noqa: C901
         if self.root_filesystem_is_overlay:
             squashed_root_file = NamedTemporaryFile()
             squashed_root = FileSystemSquashFs(
-                device_provider=None, root_dir=self.root_dir
+                device_provider=None, root_dir=self.root_dir,
+                custom_args={
+                    'compression': self.xml_state.
+                    build_type.get_squashfscompression()
+                }
```
schaefi: For consistency and because used at other places in the code, can we change the line break to match the following style:

```python
custom_args={
    'compression':
```

kiwi/builder/disk.py
```diff
@@ -748,7 +752,11 @@ def _build_and_map_disk_partitions(self, disksize_mbytes):  # noqa: C901
             log.info('--> creating readonly root partition')
             squashed_root_file = NamedTemporaryFile()
             squashed_root = FileSystemSquashFs(
-                device_provider=None, root_dir=self.root_dir
+                device_provider=None, root_dir=self.root_dir,
+                custom_args={
+                    'compression': self.xml_state.
+                    build_type.get_squashfscompression()
+                }
```
schaefi: same here, line break style

kiwi/builder/live.py
```diff
@@ -254,12 +254,14 @@ def create(self):
...
+            custom_args={
+                'compression': self.xml_state.
+                build_type.get_squashfscompression()
+            }
         )
         container_image = NamedTemporaryFile()
-        live_container_image.create_on_file(
-            container_image.name
-        )
+        live_container_image.create_on_file(container_image.name)
```
schaefi: why was this changed ? just nitpick but the diff part has no meaning and I liked the style before better :)

kiwi/schema/kiwi.rnc
```diff
@@ -1388,6 +1388,15 @@ div {
             sch:param [ name = "attr" value = "filesystem" ]
             sch:param [ name = "types" value = "vmx oem pxe" ]
         ]
+	k.type.squashfscompression.attribute = 
+        ## Specifies the compression type for mksquashfs
+        attribute squashfscompression {
+            "none" | "gzip" | "lzo" | "lz4" | "xz"
+        }
```
schaefi: Do we want "none" to be handled as a value ? I'm asking because this kind of conflicts with none handling in python from a  readers perspective. Technically it's all ok but reading 'none' usually does not imply to be an option setting.
I thought about 'uncompressed" or "raw" or "flat" ... something in that area. Thoughts ?

kiwi/builder/live.py
davidcassany: I guess this is an unintended change that escaped me. I did several changes here and there before doing the commit and I guess I did not revert to the exact same line breaks. Gonna leave it as it was.

kiwi/schema/kiwi.rnc
davidcassany: Good point, I did not think about it. I think the `uncompressed` is the best one, long but clear and explicit, gonna change to that.

kiwi/builder/disk.py
davidcassany: :+1:

Conan-Kudo:
@davidcassany latest squashfs-tools supports `zstd` too. 😄

davidcassany:
> Do you want to add the other compression mode Neal suggested ?

:+1: Fine with me, I haven't seen it as an option in TW yet, but I saw it upstream. Just added it

PR #1350, opened 2020-02-25
Consistency kiwi binary name

schaefi:
Fixed docopt strings to use correct binary name
    
The kiwi binary from the entry_point configuration is
kiwi-ng. The docopt strings should use this name for
consistency. The alternative binary name kiwi is just
a symlink created on the rpm packaging level and is
not guaranteed to exist depending on how kiwi was
installed

### DIFF
--- a/doc/source/commands/image_info.rst
+++ b/doc/source/commands/image_info.rst
@@ -1,5 +1,5 @@
-kiwi image info
-===============
+kiwi-ng image info
+==================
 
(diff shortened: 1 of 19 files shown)


schaefi:
@tomschr your loving topic, consistency :)

schaefi:
Hmm, it seems the mock tool by fedora has problems. From the error message in the pipeline I can't tell what the issue is

PR #1353, opened 2020-02-26
Turn xorriso requirement into a recommend

schaefi:
xorriso was set required on the master python3-kiwi package.
This is not needed since it is only required if an ISO target
will be built. Especially on non x86 architectures the ISO
target is used rarely. Therefore this commit turns xorriso
into a recommended package on non x86 architectures. This is
related to Issue #1325

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -139,7 +139,15 @@ Requires:       gdisk
 %endif
 Requires:       dosfstools
 Requires:       e2fsprogs
+%if 0%{?rhel} && 0%{?rhel} < 8
 Requires:       xorriso
+%else
+%ifarch %{ix86} x86_64
+Requires:       xorriso
+%else
+Recommends:     xorriso
+%endif
+%endif
 Requires:       grub2
 Requires:       kiwi-tools
 Requires:       lvm2


Conan-Kudo:
@schaefi Also note that ppc64le workstations exist now (Talos II, for example), and you'd absolutely make ISOs to install on there.

schaefi:
> nothing stops you from making ppc64le ISOs that boot on KVM on ppc64le

This is of course correct and I even create one in the integration tests. However for arm and for s390 the iso target gains not much interest. On power you are right but basically starting with power8 and also have the problem that sle12 as one example distro doesn't provide xorriso. The later is my biggest problem right now which was the start of my thinking that the iso creation toolkit is better a recommended package instead of a requirement. For running with a virtual iso drive on power you also have to setup that in the BMC. By default there is no such "boot from iso" available on those architectures.

But I agree my main driver is the absence of the right iso creation toolkit on older enterprise distributions.

I still think a smaller requirement list would make sense

Thoughts ?

Conan-Kudo:
> However for arm and for s390 the iso target gains not much interest. 

This is not going to remain true for ARM. In server and VM contexts, ISOs are quite useful for ARM nowadays because of SBSA and EBBR standards being adopted.

As for s390, meh.

I do not think this change is worth the complexity, and it we *really* should just have xorriso available everywhere.

davidcassany:
> I still think a smaller requirement list would make sense

I think it makes sense too. Keeping requirements to the minimum expression is something I like and agree with. Then following this approach we should not depend on architecture, but on image type. Thus IMHO this should be a recommended dependency for any sort of arch and add a hard requirement for `kiwi-image-iso-requires` and `kiwi-image-oem-requires`. Probably we should add anything that we require in `kiwi-image-*` packages as a recommended dependency in `python3-kiwi`.

I don't think there a clean nice minimalist solution, so the decision we need to make is: should we try to make kiwi package to require everything it might require at any time or should we try to minimize its dependencies and just recommend all the tools that could be needed depending on the image description.

I am fine with the later and I think it works better in SUSE's context. Also in recent days, highly motivated my the micro services hype, the tendency is to minimize the dependencies as much as possible, even I doubt this is that relevant for KIWI.

Conan-Kudo:
Downgrading the dependency will have other knock-on effects. Multi-profile builds are simply not going to work properly in build systems, since they don't install weak dependencies. And not all build systems make it easy to force in another dependency.

I'd be more willing to consider something like this if we decided to restructure the kiwi packaging to something similar to how I have it in Fedora with the `kiwi-systemdeps` subpackage, maybe variants of it for different build types.

But I also think this just isn't worth the trouble. It's not like it isn't or shouldn't be available, it's just that you think you want a smaller set of mandatory dependencies. I think in doing so, you'd cause difficult to discover breakages for downstream users.

schaefi:
>But I also think this just isn't worth the trouble. It's not like it isn't or shouldn't be available, it's just >that you think you want a smaller set of mandatory dependencies. I think in doing so, you'd cause difficult to discover breakages for downstream users.

This is a fair point. If we keep the requirement and want the package in sle12 it would mean the following packages needs to make its way to sle12 (ppc64le):

* https://build.opensuse.org/project/show/Virtualization:Appliances:Staging

See the packages marked as link

With this available I can build the integration tests for sle12

* https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_ppc:sle12

xorriso is missing on sle12 completely, the other python packages except for cffi are not part of sle12 as well. cffi requires an update in its version which requires negotiations with all products based on sle12.

I'd like to be honest, it's very unlikely that all this can be resolved before EOL of sle12. I will leave this on business demand because that's the only way to make it happen. Until then I will live with the prepared packages in Staging

schaefi:
> I'd be more willing to consider something like this if we decided to restructure the kiwi packaging to  something similar to how I have it in Fedora with the kiwi-systemdeps subpackage, maybe variants  of it for different build types.

This is a good idea. Actually that concept is in place with internal (-requires) packages used by obs to setup the worker. It would be an option to think about a restructuring that makes this also publicly available. I'll add this on the idea project page

schaefi:
Closing on behalf of https://github.com/OSInside/kiwi/projects/2#card-33806223

PR #1354, opened 2020-02-26
Store Mock build logs and built RPMs as short-term artifacts

Conan-Kudo:
This makes it much easier to debug what is happening with mock when
failures occur.


schaefi:
yep :+1:

PR #1356, opened 2020-02-27
Make sure it properly validate the size of the partitions in the conf…

kr0nt4b:
…ig xml

Fixes # .
When using the following xml:

`<volume mountpoint="/home" name="LVhome" size="512M" />`

you get a warning in the log:

`/usr/lib/python3.6/site-packages/kiwi/xml_parse.py:4056: UserWarning: Value "b'512M'" does not match xsd pattern restrictions: [['^\\d+\|\\d+M\|\\d+G\|all$'`


schaefi:
Thanks much for finding this. The code change was done on an auto generated file. xml_parse is code generated by generateDS. The fix for the pattern should be in the schema file schema/kiwi.rnc. I moved your change over to that file here: #1357 

Could you please double check if that pull request does what you aim for ?

Would be great thanks much

schaefi:
closing on behalf of #1357

PR #1357, opened 2020-02-28
Fixed schema regexp validation

schaefi:
Regexp patterns used in the schema are translated into python
expressions by generateDS. It's required to use the XSD schema
to run generateDS, xsd however has some restrictions on pattern
use which leads to a warning message for the ones fixed here


kr0nt4b:
Hi @schaefi. This is what i indented. This will resolve the incorrect warning and pass the validation as mentioned in #1356

schaefi:
Thanks everybody for the feedback and review

PR #1359, opened 2020-03-02
Build and push SUSE documentation

schaefi:
On any pull request also build the suse documentation and
push changes to the OSInside/kiwi-suse-doc git repository.
The SUSE documentation team needs a repo with docbook sources
for the publishing procedure. In addition change the gitlab
pipeline to run in three stages: Test, Documentation and
Package. Let the style and unit tests run in parallel and
cleanup the dependency setup

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,6 +1,8 @@
 stages:
   - test
-  - package
+  - documentation
+  - sourcepackage
+  - packages
 
 variables:
   LANG: en_US.UTF-8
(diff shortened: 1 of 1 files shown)


schaefi:
@davidcassany the background for this change is a conversation I had with the doc team. Here is a summary:

The doc team requires a repository with docbook sources to run their publishing process. I have setup that repository here:

* https://github.com/OSInside/kiwi-suse-doc

The contents of this repository gets modified by each pull request of kiwi here which changes the documentation. The commit message on a change will use:

```
git config --global user.email "kiwi-internal@suse.de"
git config --global user.name "SUSE KIWI Team"
```

We will continue to maintain only one documentation source which is here in this git repo.
In theory it's possible to push to the kiwi-suse-doc repo but it wouldn't make any sense. Also the generated docbook source is so ugly that I'm convinced nobody wants to touch this :)

If this is yet to happen I will change the gitlab test to force push. It might also be a good idea to increase the README file in the kiwi-suse-doc repo to point to the fact that this repo only holds autogenerated docbook sources which originates from here.

Hope all this makes sense to you too.

Looking forward to your feedback

schaefi:
Also see the running pipeline:

* https://gitlab.com/kiwi3/kiwi-ci/pipelines/122564815

looks much better than before imho

schaefi:
On the security level I maintain a gitlab variable. If that variable gets compromised people would have access to all open repos in the OSInside organisation. A simple revoke of the doc_suse ssh-key however will solve this situation. As we are not dealing with critical software in that organisation I considered the level of security to be ok. As I'm not an expert in this area I'm of course open to any feedback or concerns. Thanks

schaefi:
good point, do you know how this can be stated in the gitlab setup. I know it can be done but have to search how first

davidcassany:
I think what we need is to just run it against master, so it runs on any new master commit. Thus I assume something like what is described [here](https://docs.gitlab.com/ee/ci/merge_request_pipelines/) should be enough.

So probably adding
```yaml
only:
  - master
```
for the docs stage does the trick. I have very limited knowledge on gitlab though, I can give it a try if you wish.

davidcassany:
Or even better
```yaml
only:
  - tags
```

schaefi:
ok I added the push only on master. I can only hope it will work because we will see this now after this PR got merged ;) Could you double check that I didn't do something stupid ? Thanks

schaefi:
keeping fingers crossed and give it a merge

.gitlab-ci.yml
```diff
@@ -51,38 +55,94 @@ build_doc:
...
+    - chmod 600 id_doc_suse
+    - git clone git@github.com:OSInside/kiwi-suse-doc.git /tmp/kiwi-suse-doc
+    - cp doc/DC-kiwi /tmp/kiwi-suse-doc/doc/build
+    - cp doc/source/.images/* /tmp/kiwi-suse-doc/doc/build/build/.images/color
+    - cp doc/build/xml/book.xml /tmp/kiwi-suse-doc/doc/build/xml
+    - git config --global user.email "kiwi-internal@suse.de"
+    - git config --global user.name "SUSE KIWI Team"
+    - cd /tmp/kiwi-suse-doc
+    - git add .
+    - git commit -a -m "upstream sync" && git push || true
```
tomschr: Probably a matter of taste, but... could you move that part into a separate shell script? IMHO, the GitLab Yaml file should be as simple as possible.

tomschr:
Ahh, too late. :wink:

PR #1368, opened 2020-03-10
Added new image type: wsl

### LINKED ISSUE #1235: implement WSL appx support
appx files are  used to deliver Linux distributions for WSL. Appx files are essentially zip files that contain an exe and some xml files, as well as a tarball of a Linux system. The latter part is what kiwi can already create. Just the wrapping missing. There's a [tool](https://github.com/facebook/fb-util-for-appx) to generate the appx. As a first step kiwi could simply call that tool. What we actually need is only the blocklist calculation though, so could be implemented directly in python later too.

See also https://lizards.opensuse.org/2019/10/09/opensuse-wsl-images-in-obs/

schaefi:
Support for building appx container images for
Windows Subsystem Linux (WSL). This Fixes #1235

### DIFF
--- a/kiwi/container/setup/appx.py
+++ b/kiwi/container/setup/appx.py
@@ -1,0 +1,167 @@
+# Copyright (c) 2020 SUSE Software Solutions Germany GmbH.  All rights reserved.
+#
+# This file is part of kiwi.
+#
+# kiwi is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# kiwi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kiwi.  If not, see <http://www.gnu.org/licenses/>
+#
+import os
+import platform
+from lxml import etree
+from xml.dom import minidom
+
+# project
+from kiwi.container.setup.base import ContainerSetupBase
+from kiwi.exceptions import KiwiContainerSetupError
+
+
+class ContainerSetupAppx(ContainerSetupBase):
+    """
+    WSL/Appx container setup
+    """
+    def setup(self):
+        """
+        Setup container metadata
+        """
+        self.appx_at_microsoft = 'http://schemas.microsoft.com/appx'
+        self.appx_manifest_file = os.sep.join(
+            [self.custom_args.get('metadata_path'), 'AppxManifest.xml']
+        )
+        self.appx_namespace_map = dict(
+            wsl='{0}/manifest/foundation/windows10'.format(
+                self.appx_at_microsoft
+            ),
+            uap='{0}/manifest/uap/windows10'.format(
+                self.appx_at_microsoft
+            ),
+            desktop='{0}/manifest/desktop/windows10'.format(
+                self.appx_at_microsoft
+            ),
+            uap3='{0}/manifest/uap/windows10/3'.format(
+                self.appx_at_microsoft
+            ),
+            mp='{0}/2014/phone/manifest'.format(
+                self.appx_at_microsoft
+            )
+        )
+        if not os.path.exists(self.appx_manifest_file):
+            raise KiwiContainerSetupError(
+                'Appx Manifest not found at: {0}'.format(
+                    self.appx_manifest_file
+                )
+            )
+        try:
+            self.appx_manifest = etree.parse(
+                self.appx_manifest_file
+            )
+            self._update_identity()
+            self._update_desktop_visual()
+            self._update_application_metadata()
+            xml_data_unformatted = etree.tostring(
+                self.appx_manifest.getroot()
+            )
+            xml_data_domtree = minidom.parseString(xml_data_unformatted)
+            xml_data_raw = xml_data_domtree.toprettyxml(
+                indent='    ', encoding="utf-8"
+            )
+            with open(self.appx_manifest_file, 'wb') as appx_manifest:
+                for line in xml_data_raw.split(b'\n'):
+                    if line.split():
+                        appx_manifest.write(line + b'\n')
+        except Exception as issue:
+            raise KiwiContainerSetupError(
+                'Failed to update Appx Manifest: {0}'.format(issue)
+            )
+
+    def _update_identity(self):
+        identities = self._get_xpath('//wsl:Identity')
+        history = self.custom_args.get('history') or {}
+        for identity in identities:
+            if self._get_microsoft_processor_architecture():
+                identity.set(
+                    'ProcessorArchitecture',
+                    self._get_microsoft_processor_architecture()
+                )
+            if history.get('package_version'):
+                identity.set('Version', history.get('package_version'))
+            if history.get('created_by'):
+                identity.set('Publisher', history.get('created_by'))
+            if history.get('author'):
+                identity.set('Name', history.get('author'))
+
+    def _update_desktop_visual(self):
+        history = self.custom_args.get('history') or {}
+        displaynames = self._get_xpath(
+            '//wsl:Properties/wsl:DisplayName'
+        )
+        published_displaynames = self._get_xpath(
+            '//wsl:Properties/wsl:PublisherDisplayName'
+        )
+        visuals = self._get_xpath(
+            '//wsl:Applications/wsl:Application/uap:VisualElements'
+        )
+        titles = self._get_xpath(
+            '//wsl:Applications/wsl:Application/'
+            'uap:VisualElements/uap:DefaultTile'
+        )
+        if self.custom_args.get('container_name'):
+            for displayname in displaynames:
+                displayname.text = \
+                    self.custom_args.get('container_name')
+            for published_displayname in published_displaynames:
+                published_displayname.text = \
+                    self.custom_args.get('container_name')
+        for visual in visuals:
+            if self.custom_args.get('container_name'):
+                visual.set(
+                    'DisplayName', self.custom_args.get('container_name')
+                )
+            if history.get('comment'):
+                visual.set('Description', history.get('comment'))
+        for titel in titles:
+            titel.set('ShortName', self.custom_args.get('container_name'))
+
+    def _update_application_metadata(self):
+        history = self.custom_args.get('history') or {}
+        applications = self._get_xpath('//wsl:Applications/wsl:Application')
+        extensions = self._get_xpath(
+            '//wsl:Applications/wsl:Application/'
+            'wsl:Extensions/uap3:Extension'
+        )
+        aliases = self._get_xpath(
+            '//wsl:Applications/wsl:Application/'
+            'wsl:Extensions/uap3:Extension/uap3:AppExecutionAlias/'
+            'desktop:ExecutionAlias'
+        )
+        for application in applications:
+            if history.get('application_id'):
+                application.set('Id', history.get('application_id'))
+            if history.get('launcher'):
+                application.set('Executable', history.get('launcher'))
+        for extension in extensions:
+            if history.get('launcher'):
+                extension.set('Executable', history.get('launcher'))
+        for alias in aliases:
+            if history.get('launcher'):
+                alias.set('Alias', history.get('launcher'))
+
+    def _get_xpath(self, expression):
+        return self.appx_manifest.getroot().xpath(
+            expression, namespaces=self.appx_namespace_map
+        )
+
+    def _get_microsoft_processor_architecture(self):
+        windows = {
+            'x86_64': 'x64'
+        }
+        return windows.get(platform.machine())
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -37,6 +37,7 @@ arch-name = xsd:token {pattern = "(x86_64|i586|i686|ix86|aarch64|arm64|armv5el|a
 portnum-type = xsd:token {pattern = "(\d+|\d+/(udp|tcp))"}
 grub_console = xsd:token {pattern = "(console|gfxterm|serial)( (console|gfxterm|serial))*"}
 fs_attributes = xsd:token {pattern = "(no-copy-on-write|synchronous-updates)(,(no-copy-on-write|synchronous-updates))*"}
+package-version-type =  xsd:token {pattern = "(0|[1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])(\.(0|[1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])){3}"}
 
 #==========================================
 # start with image description
@@ -1542,8 +1543,22 @@ div {
         attribute image {
             "btrfs" | "clicfs" | "cpio" | "docker" | "ext2" | "ext3" |
             "ext4" | "iso" | "oem" | "pxe" | "squashfs" | "tbz" |
-            "vmx" | "xfs" | "oci"
+            "vmx" | "xfs" | "oci" | "appx"
         }
+        >> sch:pattern [
+            id = "metadata_path_mandatory" is-a = "image_type_requirement"
+            sch:param [ name = "attr" value = "metadata_path" ]
+            sch:param [ name = "types" value = "appx" ]
+        ]
+    k.type.metadata_path.attribute =
+        ## Specifies a path to additional metadata required for
+        ## the selected image type or its tools used to create
+        ## that image type
+        attribute metadata_path { text }
+        >> sch:pattern [ id = "metadata_path" is-a = "image_type"
+            sch:param [ name = "attr" value = "metadata_path" ]
+            sch:param [ name = "types" value = "appx" ]
+        ]
     k.type.installboot.attribute =
         ## Specifies the bootloader default boot entry for the
         ## initial boot of a kiwi install image. This value is
@@ -1764,6 +1779,7 @@ div {
         k.type.force_mbr.attribute? &
         k.type.initrd_system.attribute? &
         k.type.image.attribute &
+        k.type.metadata_path.attribute? &
         k.type.installboot.attribute? &
         k.type.install_continue_on_timeout.attribute? &
         k.type.installprovidefailsafe.attribute? &
@@ -2091,7 +2107,7 @@ div {
         abstract = "true"
         id = "container_type"
         sch:rule [
-            context = "containerconfig"
+            context = "containerconfig[@$attr]"
             sch:assert [
                 test = "contains('$types', ../@image)"
                 "containerconfig($attr) is only available for the following "~
@@ -2106,7 +2122,7 @@ div {
         attribute name { text }
         >> sch:pattern [ id = "name" is-a = "container_type"
             sch:param [ name = "attr" value = "name" ]
-            sch:param [ name = "types" value = "docker oci" ]
+            sch:param [ name = "types" value = "docker oci appx" ]
         ]
     k.containerconfig.tag.attribute =
         ## Specifies a tag for the container. This is usually the
@@ -2374,9 +2390,27 @@ div {
         ## Specifies the 'author' history record.
         attribute author { text }
 
+    k.history.application_id =
+        ## Specifies a custom application ID, used for WSL containers
+        attribute application_id { text }
+
+    k.history.package_version =
+        ## Specifies a version information, used for WSL containers.
+        ## The version given here follows the package numbering rules
+        ## as documented in: https://docs.microsoft.com/en-us/windows/uwp/publish/package-version-numbering
+        attribute package_version { package-version-type }
+
+    k.history.launcher =
+        ## Specifies name of the container launcher program
+        ## Used for WSL containers
+        attribute launcher { text }
+
     k.history.attlist =
         k.history.created_by.attribute? &
-		k.history.author.attribute?
+		k.history.author.attribute? &
+        k.history.application_id? &
+        k.history.package_version? &
+        k.history.launcher?
 
     k.history =
         ## Provides details about the container history. Includes the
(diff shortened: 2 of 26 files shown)


schaefi:
For testing a useful WSL build we now need a package that provides the distro launcher and the other data as listed here:

```
AppxManifest.xml
Assets/LargeTile.scale-100.png
Assets/LargeTile.scale-125.png
Assets/LargeTile.scale-150.png
Assets/LargeTile.scale-200.png
Assets/LargeTile.scale-400.png
Assets/SmallTile.scale-100.png
Assets/SmallTile.scale-125.png
Assets/SmallTile.scale-150.png
Assets/SmallTile.scale-200.png
Assets/SmallTile.scale-400.png
Assets/Square150x150Logo.scale-100.png
Assets/Square150x150Logo.scale-125.png
Assets/Square150x150Logo.scale-150.png
Assets/Square150x150Logo.scale-200.png
Assets/Square150x150Logo.scale-400.png
Assets/Square44x44Logo.altform-unplated_targetsize-16.png
Assets/Square44x44Logo.altform-unplated_targetsize-256.png
Assets/Square44x44Logo.altform-unplated_targetsize-32.png
Assets/Square44x44Logo.altform-unplated_targetsize-48.png
Assets/Square44x44Logo.scale-100.png
Assets/Square44x44Logo.scale-125.png
Assets/Square44x44Logo.scale-150.png
Assets/Square44x44Logo.scale-200.png
Assets/Square44x44Logo.scale-400.png
Assets/Square44x44Logo.targetsize-16.png
Assets/Square44x44Logo.targetsize-24.png
Assets/Square44x44Logo.targetsize-24_altform-unplated.png
Assets/Square44x44Logo.targetsize-256.png
Assets/Square44x44Logo.targetsize-32.png
Assets/Square44x44Logo.targetsize-48.png
Assets/StoreLogo.png
Assets/Wide310x150Logo.scale-100.png
Assets/Wide310x150Logo.scale-125.png
Assets/Wide310x150Logo.scale-150.png
Assets/Wide310x150Logo.scale-200.png
Assets/Wide310x150Logo.scale-400.png
Resources.pri
Resources.scale-125.pri
Resources.scale-150.pri
Resources.scale-200.pri
Resources.scale-400.pri
images/icon.ico
openSUSE-Tumbleweed.exe
```

schaefi:
@lnussel Would be great if you could provide such a package, it will enable us to provide an integration test build along with the pull request and we can directly check if the result runs on Windows. I also appreciate your code review and reading of the doc chapter that is also part of this PR.

Thanks

Vogtinator:
The version number and ideally also the release are part of `AppxManifest.xml`.
As this file is provided by a separate package, this doesn't match the image version/release. To make that work, kiwi needs to either set that itself or replace some placeholder with the image version/release before building the .appx.

schaefi:
Thanks for the hint, I think we could implement reading of the AppxManifest.xml and update the needed information in that file and write it back. I'll check on that

Vogtinator:
The resulting image currently has `.wsl` as extension, but it needs to be `.appx` for OBS and Windows to recognize it correctly.

schaefi:
> The resulting image currently has `.wsl` as extension, but it needs to be `.appx` for OBS and Windows to recognize it correctly.

Oh, good point, going to fix it. Thanks

Vogtinator:
> @lnussel Would be great if you could provide such a package, it will enable us to provide an integration test build along with the pull request and we can directly check if the result runs on Windows. I also appreciate your code review and reading of the doc chapter that is also part of this PR.
> 
> Thanks

I have it building in my OBS ~, feel free to take whatever you need: https://build.opensuse.org/package/show/home:favogt:branches:Virtualization:WSL/wsl-appx

It's not as clean as it should be, but it works (tested on windows, with a patched kiwi to get a valid `.appx` file)

schaefi:
Several changes due to review and feedback. Thanks.

I have fixed the extension name, is now .appx

I have added support for setting the sections and attributes in the AppxManifest.xml file according to the placeholders I found there. Please double check if the mappings I did from the containerconfig section into the AppxManifest.xml makes sense to you too

I also updated the documentation. Please read or build (tox -e doc) the docs and give it a read.

Thanks much we are close

kiwi/container/setup/appx.py
```diff
@@ -0,0 +1,155 @@
...
+            )
+
+    def _update_identity(self):
+        identities = self._get_xpath('//wsl:Identity')
+        history = self.custom_args.get('history') or {}
+        for identity in identities:
+            identity.set('ProcessorArchitecture', platform.machine())
+            identity.set('Version', self.custom_args.get('image_version'))
+            if history.get('created_by'):
+                identity.set('Publisher', history.get('created_by'))
```
Vogtinator: `Publisher` has to equal what is set in the project certificate, so kiwi should either find that out by itself or not modify this.

Vogtinator: I suggest to set the values below `Properties` instead.

kiwi/container/setup/appx.py
```diff
@@ -0,0 +1,155 @@
...
+        except Exception as issue:
+            raise KiwiContainerSetupError(
+                'Failed to update Appx Manifest: {0}'.format(issue)
+            )
+
+    def _update_identity(self):
+        identities = self._get_xpath('//wsl:Identity')
+        history = self.custom_args.get('history') or {}
+        for identity in identities:
+            identity.set('ProcessorArchitecture', platform.machine())
```
Vogtinator: Does `platform.machine()` return the windows names for architectures?

schaefi: probably not if we run it on Linux... hmmm

kiwi/container/setup/appx.py
schaefi: ok so we don't set Publisher.  I don't see a generic way for kiwi to do this

> I suggest to set the values below Properties instead.

This is done, we set:

DisplayName
PublisherDisplayName

from Properties. Do we need more ?

schaefi: in the visuals update

kiwi/container/setup/appx.py
schaefi: Is there an easy way to add a mapping table ? I don't know windows platform names. Help is appreciated

Vogtinator: Me neither. All I know are `x86`, `x64` and `arm64`

lnussel:
I don't think it's a good idea to update AppxManifest.xml from kiwi. Normally one creates that file using Visual Studio and that would be the canonical source then. Those placeholders are therefore nothing MS anticipated, that's something we need to be able to automatically adopt the file when eg branching wsl-appx in OBS. Doing that in kiwi loses the flexibility we get from doing that in a spec file. What you can do is read properties from the file though.

Vogtinator:
There should be a way to have the image build number in the `AppxManifest.xml` though.

lnussel:
right. So that one to replace then. Well, the rest we could still do in wsl-appx after all even if kiwi has code for that as well. We'd just not leave those placeholders in. Never mind then :)

schaefi:
> I don't think it's a good idea to update AppxManifest.xml from kiwi. Normally one creates that file using Visual Studio and that would be the canonical source then. Those placeholders are therefore nothing MS anticipated, that's something we need to be able to automatically adopt the file when eg branching wsl-appx in OBS. Doing that in kiwi loses the flexibility we get from doing that in a spec file. What you can do is read properties from the file though.

If building in obs is the only built platform you are probably right. But the combination build in obs plus a spec file that does placeholder replacements is only one possible way to build the appx container. If you look at it from more global perspective people also want to build this outside of obs and not surrounded by a spec file. Thus all operations should at least be possible from an image description perspective.

As I also wrote in the documentation all the containerconfig information is optional so you can stay with the concept of the spec file that updates the data, or combine it as you see fit.

schaefi:
Anybody who knows the architecture names supported on Windows or a doc pointer to the contents of the AppxManifest ? I did not find a good doc source so far. Thanks

schaefi:
I have now also setup the integration test build. You can find it here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:suse/test-image-wsl

I used the package @Vogtinator build to provide the metadata information. The image description does not setup the publisher and I assume the arch setting will be wrong. Could someone give this a test on Windows ?

Thanks

Vogtinator:
The package seems to be stuck in `signing`...

Note that you can use `Substitute: kiwi-image:appx python3-kiwi wsl-appx` to only pull it in for appx builds.

schaefi:
> The package seems to be stuck in signing

yes don't know why, I guess that's some obs delay. It's however build you can fetch it with osc getbinaries

schaefi:
>Note that you can use Substitute: kiwi-image:appx python3-kiwi wsl-appx to only pull it in for appx builds.

Could point, I'll adapt the prjconf. Thanks

Vogtinator:
> > The package seems to be stuck in signing
> 
> yes don't know why, I guess that's some obs delay. It's however build you can fetch it with osc getbinaries

Nope - before signing is done there aren't any binaries. `osc getbinaries` only works when stuck in `finished`

schaefi:
Oh, you are right. I thought this was possible because osc buildlog images x86_64 showed results:

```
[  527s] build: extracting built packages...
[  527s] KIWI/WSL-openSUSE-Tumbleweed.x86_64-2.1.0-Build2.1.appx.sha256
[  527s] KIWI/WSL-openSUSE-Tumbleweed.x86_64-2.1.0-Build2.1.packages
[  527s] KIWI/WSL-openSUSE-Tumbleweed.x86_64-2.1.0-Build2.1.appx
[  527s] KIWI/WSL-openSUSE-Tumbleweed.x86_64-2.1.0-Build2.1.verified
```

but you are right, there is no way to fetch them...

schaefi:
guess there is nothing I can do to influence the signing ?

Vogtinator:
Don't think so. If a rebuild doesn't fix it, some help from the OBS team would be needed.

schaefi:
> guess there is nothing I can do to influence the signing ?

Actually there was something I could do. Thanks to mls who explained that an RSA key is required for WSL signing. The Virt:App project still had an old key and I had to recreate this one. I talked to the maintainers listed in the project and they did not object. The new key has been created and now the integration test also builds. You should be able to see results now. Thanks

schaefi:
@Vogtinator 

* https://download.opensuse.org/repositories/Virtualization:/Appliances:/Images:/Testing_x86:/suse/images/WSL-openSUSE-Tumbleweed.x86_64-2.1.0-Build2.2.appx

Vogtinator:
That `.appx` file doesn't have the "Digital Signatures" tab in the "Properties" dialog and the installation dialog complains about a "parse error".

That's probably because the publisher in `AppxManifest.xml` provided by wsl-appx is the one referencing my home project, so it gets rejected. So you need to rebuild that package in the same project which the appx is built in.

davidcassany:
> That's probably because the publisher in `AppxManifest.xml` provided by wsl-appx is the one referencing my home project, so it gets rejected. So you need to rebuild that package in the same project which the appx is built in.

We can verify this hypothesis, let me see if this is not a configuration mess adding it there. Anyway, if your hypothesis is correct won't it cause issues on the `CR` projects for continuous rebuilds? or in any other project setup were packages and images are not within the same OBS project?

Vogtinator:
> > That's probably because the publisher in `AppxManifest.xml` provided by wsl-appx is the one referencing my home project, so it gets rejected. So you need to rebuild that package in the same project which the appx is built in.
> 
> We can verify this hypothesis, let me see if this is not a configuration mess adding it there. Anyway, if your hypothesis is correct won't it cause issues on the `CR` projects for continuous rebuilds? or in any other project setup were packages and images are not within the same OBS project?

Only if those projects don't use the same signing key. Everything below `openSUSE:` in OBS (there are some exceptions) uses the same key, for instance.

davidcassany:
@Vogtinator just temporarly added the package on the same project [here](https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_x86:suse) just for testing purposes. The package is still pending to be published by OBS since a while and thus the image build blocked. But it should only be a matter of time having a new image with an updated wsl-appx package publisher.

@schaefi you'll see I added the package. This is just to verify Fabian's hypothesis, in any case, IMHO this package should not be there and should be deleted after the tests.

Vogtinator:
@davidcassany Now the publisher is correct, but now `ProcessorArchitecture` and `Version` are broken - there are very strict requirements for those.

schaefi:
I expected the arch to be broken because we still use the linux arch names. I need a mapping table but can't find the correct arch strings Windows would accept

Regarding version, who knows the restrictions on that one ? maybe by a regexp pattern

Thanks @Vogtinator for testing all this, much appreciated

Vogtinator:
The version regex is listed in `wsl-appx.spec`:

`# '(0|[1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])(\.(0|[1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])){3}'`

schaefi:
I have updated the broken version and arch implementation and build a new package in Staging. The new image will build automatically. Let's see if the new result is better than the old one

Thanks

Conan-Kudo:
@schaefi Can we add a Fedora test too? I generally like to see us having tests for both Fedora and SUSE distributions on all the types we have.

schaefi:
> @schaefi Can we add a Fedora test too? I generally like to see us having tests for both Fedora and SUSE distributions on all the types we have.

I would do so but I don't know if it's so easy. Outside of kiwi we would need the following from Fedora:

* fb-util-for-appx, or similar package providing the appx tool
* wsl-appx, a package that provides the launcher (windows exe file) plus all the metadata around, AppxManifest.xml and the logo files and such things

I also looking for feedback from the suse build if it actually works with the latest changes or if we still missing some things

Conan-Kudo:
@schaefi I can package these things, where are their sources?

lnussel:
in Factory but wsl-appx will change after this kiwi feature. So you can save yourself work by waiting until we have things settled.
Also you need the launcher built using mingw.

Conan-Kudo:
The MinGW thing will be no problem, as Fedora has a complete MinGW stack in the distro already.

lnussel:
https://en.opensuse.org/WSL/Building_in_OBS

Vogtinator:
> I have updated the broken version and arch implementation and build a new package in Staging. The new image will build automatically. Let's see if the new result is better than the old one
> 
> Thanks

Now it complains about `Id="1"` in the `<Application>` tag, which has to start with a letter.

schaefi:
@Vogtinator next build should fix this. In my test I set opensuse-tumbleweed as id, inspired by the id settings we do in etc/os-release. Is there a general recommendation for this setting ? if so we can add that to the documentation. Thanks

Vogtinator:
> @Vogtinator the actual build should fix this. In my test I set opensuse-tumbleweed as id, inspired by the id settings we do in etc/os-release. Is there a general recommendation for this setting ? if so we can add that to the documentation. Thanks

No idea what the Id actually means. It's probably not really relevant.

`opensuse-tumbleweed` is invalid though, as `-` is not alphanumeric.

schaefi:
> `opensuse-tumbleweed` is invalid though, as `-` is not alphanumeric.

aargh, I'm so stupid

schaefi:
@Vogtinator hope for next rebuild, current osc state is "succeeded*" I guess a new build will start soon

schaefi:
@Vogtinator could you give the latest build a short test again. I think this one should be ok
Thanks

Vogtinator:
> @Vogtinator could you give the latest build a short test again. I think this one should be ok
> Thanks

Unfortunately no such luck:

> WSL-openSUSE-Tumbleweed violates pattern constraint of '[^\\]+\.(([Ee][Xx][Ee])'.

schaefi:
@Vogtinator 

>Unfortunately no such luck:
>WSL-openSUSE-Tumbleweed violates pattern constraint of '[^\]+.([Ee][Xx][Ee])'.

Thanks for testing. The string WSL-openSUSE-Tumbleweed was used in the Manifest for the attributes ShortName and Alias. I wonder about the regexp to force that names to end with ```.exe/.EXE``` Is that really the issue ? Do you know which elements of the Manifest requires that pattern ?

Thanks

Vogtinator:
> @Vogtinator
> 
> > Unfortunately no such luck:
> > WSL-openSUSE-Tumbleweed violates pattern constraint of '[^]+.([Ee][Xx][Ee])'.
> 
> Thanks for testing. The string WSL-openSUSE-Tumbleweed was used in the Manifest for the attributes ShortName and Alias. I wonder about the regexp to force that names to end with `.exe/.EXE` Is that really the issue ? Do you know which elements of the Manifest requires that pattern ?

No idea. All I can tell is that the XML in the wsl-appx build works: https://build.opensuse.org/package/live_build_log/home:favogt:branches:Virtualization:WSL/wsl-appx/openSUSE_Tumbleweed/x86_64

> Thanks

schaefi:
Ah I see

```
<desktop:ExecutionAlias Alias="openSUSE-Tumbleweed.exe" />
```
The Alias has to match this, ok I'll fix that

schaefi:
@Vogtinator ok fixed now. The image should start to build in a minute. Could you give the result a test again. Really sorry to bother you with this but I hope this will be the last time... last famous words :-)

Vogtinator:
> @Vogtinator ok fixed now. The image should start to build in a minute. Could you give the result a test again. Really sorry to bother you with this but I hope this will be the last time... last famous words :-)

It's actually working fine this time!

schaefi:
:-) great, thanks for your patience. I think we are good to go this time

@davidcassany could you have a final look on this pull request ?

Thanks much

kiwi/schema/kiwi.rnc
```diff
@@ -2115,36 +2131,36 @@ div {
         attribute tag { text }
         >> sch:pattern [ id = "tag" is-a = "container_type"
             sch:param [ name = "attr" value = "tag" ]
-            sch:param [ name = "types" value = "docker oci" ]
+            sch:param [ name = "types" value = "docker oci appx" ]
```
davidcassany: tags are being used in WSL?

kiwi/schema/kiwi.rnc
```diff
@@ -2115,36 +2131,36 @@ div {
...
         ]
     k.containerconfig.additionaltags.attribute =
         ## Specifies additional tags for the container using a comma
         ## separated values string
         attribute additionaltags { text }
         >> sch:pattern [ id = "additionaltags" is-a = "container_type"
             sch:param [ name = "attr" value = "additionaltags" ]
-            sch:param [ name = "types" value = "docker oci" ]
+            sch:param [ name = "types" value = "docker oci appx" ]
```
davidcassany: Does it makes sense for WSL?

kiwi/schema/kiwi.rnc
```diff
@@ -2115,36 +2131,36 @@ div {
...
         ]
     k.containerconfig.maintainer.attribute =
         ## Specifies a maintainer for the container.
         attribute maintainer { text }
         >> sch:pattern [ id = "maintainer" is-a = "container_type"
             sch:param [ name = "attr" value = "maintainer" ]
-            sch:param [ name = "types" value = "docker oci" ]
+            sch:param [ name = "types" value = "docker oci appx" ]
```
davidcassany: I don't see maintainer being used in WSL configuration

kiwi/schema/kiwi.rnc
```diff
@@ -2115,36 +2131,36 @@ div {
...
         ]
     k.containerconfig.user.attribute =
         ## Specifies a user for the container.
         attribute user { text }
         >> sch:pattern [ id = "user" is-a = "container_type"
             sch:param [ name = "attr" value = "user" ]
-            sch:param [ name = "types" value = "docker oci" ]
+            sch:param [ name = "types" value = "docker oci appx" ]
```
davidcassany: same as above

kiwi/schema/kiwi.rnc
```diff
@@ -2115,36 +2131,36 @@ div {
...
         ]
     k.containerconfig.workingdir.attribute =
         ## Specifies the default working directory of the container
         attribute workingdir { text }
         >> sch:pattern [ id = "workingdir" is-a = "container_type"
             sch:param [ name = "attr" value = "workingdir" ]
-            sch:param [ name = "types" value = "docker oci" ]
+            sch:param [ name = "types" value = "docker oci appx" ]
```
davidcassany: same as above

kiwi/schema/kiwi.rnc
schaefi: It's not and I had the same idea when I implemented it. But I could not come across a way to specify this in the schema. If you take a look at the schema for <containerconfig> you see:

```xml
sch:pattern [
        abstract = "true"
        id = "container_type"
        sch:rule [
            context = "containerconfig"
            sch:assert [
                test = "contains('$types', ../@image)"
                "containerconfig($attr) is only available for the following "~
                "image types: $types"
            ]
        ]
    ]
```

This schematron rule is applied on any attribute of the containerconfig element even if one ore more attributes are NOT specified in the XML description

Now if I go and say

```xml
k.containerconfig.tag.attribute =
        ## Specifies a tag for the container. This is usually the
        ## the tag name of the container as read if the container
        ## image is imported via the docker load command
        attribute tag { text }
        >> sch:pattern [ id = "tag" is-a = "container_type"
            sch:param [ name = "attr" value = "tag" ]
            sch:param [ name = "types" value = "docker oci" ]
        ]
```

and leave out appx I will get a validation error that says:

```
 --> containerconfig(tag) is only available for the following image types: docker oci
[ ERROR   ]: 11:14:11 | KiwiDescriptionInvalid: Schema validation for build-tests/x86/suse/test-image-wsl/appliance.kiwi failed
```

and that happens even if you don't have set a tag in the XML description. So the schema validation in schematron is always applied on all possible elements and attributes. Given this I have no idea how I could specify what you are after

If you know a way I'm happy to add it

davidcassany: I believe I got it, there is a problem on the context definition of the schematron rules. Compare it the schematron rule for type element attributes constraints. 

On my tests it works fine with:

```
    sch:pattern [
        abstract = "true"
        id = "container_type"
        sch:rule [
            context = "containerconfig[@$attr]"
            sch:assert [
                test = "contains('$types', ../@image)"
                "containerconfig($attr) is only available for the following "~
                "image types: $types"
            ]
        ]
    ]
```
see the context definition.

With this and removing the `appx` image type for `tags`, `additionaltags`, `maintainer` and `workingdir` I can validate the WSL text XML you provided.

davidcassany: My bad I entered this containerconfig back in a time and since the two supported image types were actually using the same config it did not pop up.

schaefi: Aaargh I completely overlooked the context. Yes that works. Thanks much

schaefi:
@davidcassany rebased and tested, I think we are ready for merge. Let me know if you agree

PR #1369, opened 2020-03-13
Add kpartx dependency on dracut-kiwi-oem-dump package

### LINKED ISSUE #1364: dracut module packages should require all required tools in spec
## Problem description

`dracut-kiwi-oem-dump` package does not require `kpartx`, however this is a required tool for the module and dracut reports an error trying to install the module if the tool is not present in the image, however it does not fail.

Dracut produces something like:
```
dracut: *** Including module: kiwi-dump ***
dracut-install: ERROR: installing 'kpartx'
dracut: /usr/lib/dracut/dracut-install -D /var/tmp/dracut.Tnz2wl/initramfs -a tr lsblk dd md5sum head pv kexec basename awk kpartx
```

and then  `kiwi-dump` is not listed on the installed modules list! I am suprised that dracut is not failing here, specially when `kiwi-dump` is explicitly requested!

## Expected behaviour

Installing `dracut-kiwi-oem-dump` pulls `kpart` package.

Build the leap 15.1 oem image from [OSInside/kiwi-descriptions](https://github.com/OSInside/kiwi-descriptions) repo and make sure to add the `installpxe="true"` option.

+ KIWI version: master
+ Operating system: 
+ OBS version:

davidcassany:
This commit adds a missing dependency on dracut-kiwi-oem-dump
package. In images where The `kpartx` tool is missing the
dracut-kiwi-oem-dump was not applied for the initrd.

Fixes #1364

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -284,6 +284,7 @@ BuildRequires:  dracut
 Requires:       dracut-kiwi-lib
 Requires:       kexec-tools
 Requires:       gawk
+Requires:       kpartx
 License:        GPL-3.0-or-later
 Group:          %{sysgroup}
 


davidcassany:
I haven't been capable to catch any other dependency being missed in dracut modules. After all most of the tools we use are coming from `coreutils` and or `utils-linux` which are in turn already dependencies of `dracut` itself.

schaefi:
perfect :+1: thanks

PR #1370, opened 2020-03-14
Skip nodev devices for size calculations

### LINKED ISSUE #1363: Exception while setting du_call in system/size.py
## Problem description
Permission Denied messages in output while setting du_call in size.py error the build

```
[ INFO    ]: 02:29:59 | Precalculating required disk size
[ INFO    ]: 02:30:01 | Cleaning up BootImageDracut instance
[ ERROR   ]: 02:30:01 | KiwiCommandError: du: stderr: du: cannot access '/builds/user/project-kiwi/kiwi/img/build/image-root/proc/8606/task/8606/fd/4': No such file or directory
du: cannot access '/builds/user/project-kiwi/kiwi/img/build/image-root/proc/8606/task/8606/fdinfo/4': No such file or directory
du: cannot access '/builds/user/project-kiwi/kiwi/img/build/image-root/proc/8606/fd/3': No such file or directory
du: cannot access '/builds/user/project-kiwi/kiwi/img/build/image-root/proc/8606/fdinfo/3': No such file or directory
, stdout: 140740459628253	/builds/user/project-kiwi/kiwi/img/build/image-root

[ INFO    ]: 02:30:01 | Cleaning up BootImageDracut instance
section_end:1583375403:build_script
```

## Expected behaviour
du command runs, returns processed du -s type output and continues

Run the attached repr description with like: `sudo kiwi-ng --type vmx system build --description ../project --target-dir ./img`
I am building kiwi  using gitlab ci in this instance, log including build and install of kiwi from source included in zip. Attached a reproducing description which is largely based on the kiwi-descriptions' xenial description.
+ KIWI version: v9.20.2-4 (logs from .2, same result with .4)
+ Operating system: Fedora 31
+ du version: du (GNU coreutils) 8.31
+ OBS version: N/A

[kiwi-ng_output.txt](https://github.com/OSInside/kiwi/files/4290662/kiwi-ng_output.txt)
[repr.zip](https://github.com/OSInside/kiwi/files/4290637/repr.zip)

schaefi:
Added a static list of mountpoints used for devices that
are not associated with a physical device like /proc and
use that information in the exclude list for calculating
the image byte size. This Fixes #1363

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -221,6 +221,22 @@ class Defaults:
         ]
         return exclude_list
 
+    def get_exclude_list_for_non_physical_devices():
+        """
+        Provides the list of folders that are not associated
+        with a physical device. KIWI returns the basename of
+        the folders typically used as mountpoint for those
+        devices.
+
+        :return: list of file and directory names
+
+        :rtype: list
+        """
+        exclude_list = [
+            'proc', 'sys', 'dev'
+        ]
+        return exclude_list
+
     @staticmethod
     def get_failsafe_kernel_options():
         """
--- a/kiwi/system/size.py
+++ b/kiwi/system/size.py
@@ -15,6 +15,8 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+import os
+
 # project
 from kiwi.command import Command
 from kiwi.defaults import Defaults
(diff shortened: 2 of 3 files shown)


schaefi:
@Conan-Kudo I have moved all the tests with building in the container into an extra PR #1372 

I think the changes here are still good and should be done but they are only the result of the after effects that you see when building in a container and kpartx was used.

@davidcassany Do you agree it would be good to explicitly skip the nodev filesystems from the du call ? I think this makes sense even though it should not be needed because when kiwi calls du all nodev and other filesystem references are already umounted

Thanks

davidcassany:
> @davidcassany Do you agree it would be good to explicitly skip the nodev filesystems from the du call ? I think this makes sense even though it should not be needed because when kiwi calls du all nodev and other filesystem references are already umounted

Yes I do agree with this change, IMHO, regardless if this is strictly needed or not for our specific use case in KIWI, it makes sense to avoid those for size calculations.

PR #1373, opened 2020-03-17
Fix order in fstab

### LINKED ISSUE #1349: JeOS: wrong order: /boot/grub2/arm64-efi specified before /boot
## Problem description
`findmnt --verbose --verify` reports issue with mounts in /boot
See bug report for details: https://bugzilla.opensuse.org/show_bug.cgi?id=1164310

## Expected behaviour
`/boot` should probably be declared before `/boot/grub2/arm64-efi`

Boot a JeOS image built with btrfs and run `findmnt --verbose --verify`.

+ KIWI version: 9.19.16

davidcassany:
Any mount point directly under / should be just right after the root
mountpoint and before the custom mountpoints based on user's subvolume
configuration.

Fixes #1349 and bsc#1164310

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -876,6 +876,18 @@ class DiskBuilder:
             device_map['root'].get_device(), '/',
             custom_root_mount_args, fs_check_interval
         )
+        if device_map.get('boot'):
+            if self.bootloader == 'grub2_s390x_emu':
+                boot_mount_point = '/boot/zipl'
+            else:
+                boot_mount_point = '/boot'
+            self._add_generic_fstab_entry(
+                device_map['boot'].get_device(), boot_mount_point
+            )
+        if device_map.get('efi'):
+            self._add_generic_fstab_entry(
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi I have doubts if we should run some validation against volumes mount points and spare partition mountpoint. There are some insane cases that those could be just inconsistent, for instance, some subvolume having a mountpoint under the spare partition mountpoint. I guess that something like that shouldn't be surprising to the user to cause issues.

In any case, there is no logic on what goes first, the spart partition mountpoint or volume mountpoints. I found it a bit messy to add such a logic in that portion of the code, so that I haven't done it. Probably a runtime check just to early fail or just increase the disk builder complexity (which is already quite complex). None convinces me as trying to match insane cases in runtime checks is hard and prone to errors (we might not catch all that we wanted and prevent some rare configurations for corner cases) and increase disk builder complexity is something I don't like too, as it is really sensitive code.

My vote is leaving it as it is in this PR, with no additional logic.

PR #1374, opened 2020-03-17
Better fstab handling

schaefi:
This pull request is two fold

**Added Fstab class**
    
Handling of fstab should be done in its own namespace and class. The current handling of fstab entries is spread at several places. There should be only one code that writes the fstab entries and that code should also care for the correct canonical order of the mountpoints

**Validate and order final fstab file**
    
On write of the final fstab file read, validate and order the entries. This is related to Issue #1349


schaefi:
I have to check how and if this impacts the fstab modifications for MicroOS...

schaefi:
tested on MicroOS and worked fine

PR #1378, opened 2020-03-19
Adding OEM integration tests for Ubuntu

davidcassany:
This commit adds oem image type for Ubuntu integration tests


davidcassany:
After merging this PR and rerunning the OBS service three builds should be green https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:ubuntu/test-image-iso-oem-vmx

schaefi:
It seems we can't do much about the grub2 dependency problem. I see this more of a problem in the ubuntu distro or in the way obs included the distro.

PR #1381, opened 2020-03-20
Changed concept for checking oem resize to be done

schaefi:
The former implementation was based on checking for unallocated space on the disk compared to the current disk geometry used in the partition table. This actually is a clean way to check
the condition if the image got already resized to the real disk geometry. The tools used to check this condition were sfdisk and sgdisk. The problem is they are different in behavior and functionality compared across the distributions we support with kiwi. Thus there was a big risk that the check for unallocated space failed depending on the target distribution and release. Therefore I changed the concept to detect the condition and added an ID string to the root filesystem label.
If the ID string is present we consider it as an initial deployment, if not we consider the machine deployed and will not run the resize operation again. On resize kiwi's dracut code will rewrite the label and remove the ID. Users can re-activate the resize operation if they stick with the kiwi
oem dracut modules and append the _IMG id string to the rootfs label.

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-filesystem-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-filesystem-lib.sh
@@ -4,19 +4,34 @@ function resize_filesystem {
     local device=$1
     test -n "${device}" || return
     local resize_fs
+    local set_label="true"
     local mpoint=/fs-resize
     local fstype
+    local fslabel
+    local fslabel_final
     fstype=$(probe_filesystem "${device}")
+    fslabel=$(blkid "${device}" -s LABEL -o value)
+    if [[ "${fslabel}" == *"_IMG" ]];then
+        # The filesystem label is the one set from the image build
+        # this should indicate this is the first time this image
+        # got deployed on the target. Along with the resize of
+        # the filesystem that label will now change to its final
+        # state
+        fslabel_final=${fslabel%%_*}
+    fi
     case ${fstype} in
     ext2|ext3|ext4)
+        set_label="tune2fs -L ${fslabel_final} ${device}"
         resize_fs="resize2fs -f -p ${device}"
     ;;
     btrfs)
+        set_label="btrfs filesystem label ${device} ${fslabel_final}"
         resize_fs="mkdir -p ${mpoint} && mount ${device} ${mpoint} &&"
         resize_fs="${resize_fs} btrfs filesystem resize max ${mpoint}"
         resize_fs="${resize_fs};umount ${mpoint} && rmdir ${mpoint}"
     ;;
     xfs)
+        set_label="xfs_admin -L ${fslabel_final} ${device}"
         resize_fs="mkdir -p ${mpoint} && mount ${device} ${mpoint} &&"
         resize_fs="${resize_fs} xfs_growfs ${mpoint}"
         resize_fs="${resize_fs};umount ${mpoint} && rmdir ${mpoint}"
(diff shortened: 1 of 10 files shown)


schaefi:
@davidcassany this touches sensitive code parts and I'd like to ask for a careful review. I have tested oem image and deployment in a VM using the ext4 filesystem. It would be good to have a test with btrfs and one with xfs to check if the label setup works. I do not expect problems there though

The change here is the result of a conversation on the mailing list. ```[kiwi] Sizes of lvm are not according to the specification in config.xml```

dracut/modules.d/99kiwi-lib/kiwi-filesystem-lib.sh
```diff
@@ -4,19 +4,34 @@ function resize_filesystem {
...
+    local fslabel_final
     fstype=$(probe_filesystem "${device}")
+    fslabel=$(blkid "${device}" -s LABEL -o value)
+    if [[ "${fslabel}" == *"_IMG" ]];then
+        # The filesystem label is the one set from the image build
+        # this should indicate this is the first time this image
+        # got deployed on the target. Along with the resize of
+        # the filesystem that label will now change to its final
+        # state
+        fslabel_final=${fslabel%%_*}
```
seife: this is wrong. If the label is FOO_BAR, it gets FOO_BAR_IMG. But now you change it to FOO.

Better:
```
   fslabel_final=${fslabel%_*}
```
best:
```
   fslabel_final=${fslabel%_IMG}
```

OK, I see that only A-Za-z are allowed as label-name (*which is a backwards-compatibility breaking change!*), but still the code should be conservative.

kiwi/schema/kiwi.rnc
```diff
@@ -24,6 +24,7 @@ namespace db  = "http://docbook.org/ns/docbook"
 namespace sch = "http://purl.oclc.org/dsdl/schematron"
 namespace nul = ""
 
+label-name = xsd:token {pattern = "[a-zA-Z]{1,8}"}
```
seife: 8 characters is the smallest common possible label length, but is really short.
Again, this is a backwards compatibility breaking change.

seife: So while I do not have a better idea yet, I think this is not the best way to solve the issue.

dracut/modules.d/99kiwi-lib/kiwi-filesystem-lib.sh
schaefi: I see you already realized the connected schema change which only allows a-zA-Z labels. So the situation you constructed is correct but could not be created with kiwi. Of course there is an opportunity for the user to change the label as he wants but that could also be used to mess with the entire idea. I'm not sure if it makes sense to care for your side effect if there are still other side effects we can't control. In theory a user can set a new label as part of the config.sh script and invalidate the entire mechanism. That's the reason why I kept the code as simple as possible even though I admit it's only simple together with the restrictions on the label in the schema.

schaefi: > (which is a backwards-compatibility breaking change!)

Not necessarily. Before there was no idea for compatibility. We allowed { text } for the filesystem label. But every label of a filesystem has certain constraints, 11-16 characters lengths only a subset of symbols allowed. The constrain I set now 8 characters and only a-zA-Z is a safe label name across all kiwi supported filesystems. So this change is not backward imcompatible but protecting users from wrong label definitions :)

kiwi/schema/kiwi.rnc
schaefi: I agree it's not the best way. The implementation before was better but inconsistent and unstable across distributions. I couldn't come up with a better alternative so far. This includes restrictions on the label I could live with. In kiwi only rootfs_label can be customized via the XML description. From my perspective the restrictions with this PR should not hurt. I admit this is just best guess

Of course if there are better ideas how to solve the initial problem I'm open to adapt

schaefi:
I will do this differently

PR #1382, opened 2020-03-22
Make disk allocation check more robust

schaefi:
**Make disk allocation check more robust**
    
The tools used to check the disk allocation condition are sfdisk and sgdisk. The problem is that at least sfdisk is different in behavior and functionality compared across the distributions we support with kiwi. In addition the verification for the msdos table cannot be used to distinguish between intentionaly wanted free space on disk and a disk that has not yet been resized. Thus  this commit changes two parts:
    
* always report unallocated space available for the msdos  table to allow to run kiwi's resize code
    
* make sure the table type is taken into consideration It's important to run the verification based on the table type (DOS, GPT) where we know the tools to work. In any  other case we report the disk to have unallocated space  and give the resize code a chance

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -312,9 +312,15 @@ function disk_has_unallocated_space {
     local pt_table_type
     pt_table_type=$(get_partition_table_type "${disk_device}")
     if [ "${pt_table_type}" = "dos" ];then
-        sfdisk --verify "${disk_device}" 2>&1 | grep -q "unallocated"
-    else
+        # we can't distinguish from 'intentional free space left'
+        # and the 'already resized' condition. Thus assume it's
+        # not fully allocated to allow for resize
+        true
+    elif [ "${pt_table_type}" = "gpt" ];then
         sgdisk --verify "${disk_device}" 2>&1 | grep -q "end of the disk"
+    else
+        # assume it's not fully allocated and allow for resize
+        true
     fi
 }
 
--- a/dracut/modules.d/99kiwi-lib/module-setup.sh
+++ b/dracut/modules.d/99kiwi-lib/module-setup.sh
@@ -16,7 +16,7 @@ install() {
     declare moddir=${moddir}
     inst_multiple \
         blkid blockdev parted dd mkdir rmdir \
-        grep cut tail head tr bc \
+        grep cut tail head tr bc true false \
         basename partprobe sfdisk sgdisk mkswap readlink lsblk \
         btrfs xfs_growfs resize2fs \
         e2fsck btrfsck xfs_repair \


schaefi:
This should the better fix and replaces #1381

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -307,14 +307,19 @@ function relocate_gpt_at_end_of_disk {
...
 
-function disk_has_unallocated_space {
+function disk_is_fully_allocated {
     local disk_device=$1
     local pt_table_type
     pt_table_type=$(get_partition_table_type "${disk_device}")
     if [ "${pt_table_type}" = "dos" ];then
-        sfdisk --verify "${disk_device}" 2>&1 | grep -q "unallocated"
+        sfdisk --verify "${disk_device}" 2>&1 |\
+            grep -q "unallocated" && false
```
seife: This is always false:

```
seife@strolchi:~> echo "hello" | grep -q hello && false
seife@strolchi:~> echo $?
1
seife@strolchi:~> echo "hello" | grep -q hrmpf && false
seife@strolchi:~> echo $?
1
```
This is probably best written out explicitly:
```
  if sfdisk --verify $disk_device 2>&1 | grep -q "unallocated"; then
    false
  else
    true
  fi
```
Yes, it's long, but it is easily readable and works ;-)

There is also this (I found it on Stackoverflow):
```
seife@strolchi:~> ! echo "hello" | grep -q hello
seife@strolchi:~> echo $?
1
seife@strolchi:~> ! echo "hello" | grep -q hmpf
seife@strolchi:~> echo $?
0
```
But I personally think it's slightly obscure... probably many people will stumble over it and wonder what it does.

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -307,14 +307,19 @@ function relocate_gpt_at_end_of_disk {
...
+    elif [ "${pt_table_type}" = "gpt" ];then
+        sgdisk --verify "${disk_device}" 2>&1 |\
+            grep -q "end of the disk" && false
```
seife: same here

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
schaefi: I'm so stupid, thanks for pointing it out. Fortunately bash has !(not), that should make it relatively readable

schaefi:
For the msdos table the sfdisk verification has turned out to be useless. If there is space left on the partition table the verification command will always report unallocated space. Thus there is no way to distiunguish between the "intentional space left" and the "already resized" condition if the msdos table is in use.

That could only be made clear with the label approach from #1381

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -312,9 +312,15 @@ function disk_has_unallocated_space {
     local pt_table_type
     pt_table_type=$(get_partition_table_type "${disk_device}")
     if [ "${pt_table_type}" = "dos" ];then
-        sfdisk --verify "${disk_device}" 2>&1 | grep -q "unallocated"
+        # we can't distinguish from 'intentional free space left'
+        # and the 'already resized' condition. Thus assume it's
+        # not fully allocated to allow for resize
+        true
+    elif [ "${pt_table_type}" = "gpt" ];then
+        sgdisk --verify "${disk_device}" 2>&1 grep -q "end of the disk"
```
seife: Pipe missing :-)

schaefi: thanks, this code doesn't like me... or maybe the virus

schaefi: wonder why shellcheck did not complain, probably it did not expect that stupid mistakes ;)

schaefi:
ok let's go this route

PR #1384, opened 2020-03-24
No default boot partition for btrfs_root_is_snapshot

### LINKED ISSUE #1351: If snapshots are enabled, a /boot partition should be disabled by default
Currently even if btrfs and snapshots are enabled, it creates a separate partition for `/boot` unless `bootpartition="false"` is explicitly set.

For snapshots and rollbacks to work properly, `/boot` has to be part of the snapshots, so can't be on a separate partition. The default should be adjusted.

davidcassany:
This commit ensures no boot partition is used when root is set to be
a btrfs snapshot unless this is explicitly required by the user.

Fixes #1351

### DIFF
--- a/kiwi/storage/setup.py
+++ b/kiwi/storage/setup.py
@@ -189,10 +189,10 @@ class DiskSetup:
             return False
         if self.mdraid:
             return True
-        if self.volume_manager:
+        if self.volume_manager == 'lvm':
             return True
-        if self.filesystem == 'btrfs':
-            return True
+        if self.volume_manager == 'btrfs':
+            return False
         if self.filesystem == 'xfs':
             return True
         if self.root_filesystem_is_overlay:
--- a/test/unit/storage/setup_test.py
+++ b/test/unit/storage/setup_test.py
@@ -105,8 +105,8 @@ class TestDiskSetup:
 
     def test_need_boot_partition_btrfs(self):
         self._init_bootpart_check()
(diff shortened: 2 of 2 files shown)


davidcassany:
> Hmm, I would do this differently. Actually the use of btrfs in general should not trigger the use of a boot partition. This is an old artifact from times when grub could not read btrfs. Thus the change could be simpler if you would just change the value for btrfs in `storage/setup.py` in the `need_boot_partition` method

Yes I was thinking on making this default apply for any `btrfs` volume manager. But I followed for the conservative approach of just changing the specific reported case. Mostly because I saw affectations on some unit tests by doing so. 

Anyway I am opened of changing it and set bootpartition to false by default on `btrfs` volume manager. Gonna give it a try and make PR, discussing over the actual changes is easier.

davidcassany:
@schaefi now I disabled the boot partition by default in for `btrfs` volume managers. Note the criteria is not on the filesystem but on the volume manager. This is because we allow having LVM on top of btrfs (without subvolumes). Having a look at volume manager configuration I noted that for `btrfs` filesystem the volume manager is always set to `btrfs` unless it is explicitly set to LVM in `config.xml`. This can be seen here
https://github.com/OSInside/kiwi/blob/46cdcc7a2aab6469949b0a5ad0f8ee27f3406927/kiwi/xml_state.py#L1238-L1261

PR #1386, opened 2020-03-26
Added new oem-resize-once element

schaefi:
The new element controls the behavior of the repart/resize code
in the oem-repart dracut module.

```xml
<oemconfig>
    <oem-resize-once>true</oem-resize-once>
</oemconfig>
```

By default the repart/resize happens on every reboot and therefore also allows for disk geometry changes during the livetime of the machine. If the element is set to false the repart/resize operation happens only once and then never again. To check for this condition a new profile environment variable kiwi_rootpartuuid which holds the PARTUUID of the root partition has been added to the disk builder.

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -311,15 +311,19 @@ function disk_has_unallocated_space {
     local disk_device=$1
     local pt_table_type
     pt_table_type=$(get_partition_table_type "${disk_device}")
-    if [ "${pt_table_type}" = "dos" ];then
-        # we can't distinguish from 'intentional free space left'
-        # and the 'already resized' condition. Thus assume it's
-        # not fully allocated to allow for resize
-        true
-    elif [ "${pt_table_type}" = "gpt" ];then
+    if [ "${pt_table_type}" = "gpt" ];then
+        # GPT disks store a backup table at the end of the disk
+        # if the disk geometry changes the backup table is no
+        # longer at the end and this condition can be easily
+        # checked and used to detect that there is space
+        # unallocated due to a geometry change of the underlying
+        # block device layer
         sgdisk --verify "${disk_device}" 2>&1 | grep -q "end of the disk"
     else
-        # assume it's not fully allocated and allow for resize
+        # There is currently no method we could come up with
+        # to detect a geometry change for non GPT based disks.
+        # Thus we assume it's not fully allocated and allow
+        # for resize
         true
     fi
 }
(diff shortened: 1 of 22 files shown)


schaefi:
@davidcassany I have extensively tested this and so far things worked as expected

schaefi:
The code also covers a small refactoring of how the .profile environment gets handled. I needed to refactor this a but to allow for changes on the profile setup at different stages in the build process. The current way of doing it was too static and is one left over of the backport from the old perl kiwi times

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -497,3 +501,53 @@ function _parted_write {
...
+    # the image according to the configured constraints.
+    #
+    # If oem-resize-once is set to true we check if the
+    # system has been already resized compared to the
+    # original image PARTUUID and repart/resize the system
+    # only if the PARTUUID is still the original value.
+    # After resize a new PARTUUID will be written by the
+    # partitioner and that will result in the repart/resize
+    # operation to happen only once in the livetime of
+    # the image
```
davidcassany: Nice approach! :+1:

kiwi/xml_state.py
```diff
@@ -1808,6 +1809,20 @@ def set_derived_from_image_uri(self, uri):
             ''')
             log.warning(message.format(uri))
 
+    def set_root_partition_uuid(self, uuid):
```
davidcassany: I have to admit I am surprised you use xml_state instance to share the part uuid between DiskBuilder instance and the BootImage instance :thinking: I have no strong opinion against it but I find it unexpected, I am not sure the root partition uuid belongs there. I would expect that most, if no all, the data contained in the xml_state instance is somehow a result of XML file together a kiwi command like.

Wouldn't it make sense having a root_partuuid new attribute in BootImage?

schaefi: Well I was also thinking what would be a good namespace to store that information. The xml_state object transports stateful information about the image description. The state in this namespace covers information that is unique to the build, for example the selected profile. I thought it would not violate the scope of the instance if stateful information that does not directly come from the image description is stored there too. This has the advantage that the information is available across all major steps of the image build because the xml_state object is handed over to any major class of the build process.

The BootImage class is a reader of the information because it writes the .profile file.
The contents of the .profile file are created exclusively from the state of xml_state.
That's why I put the root_partuuid there too. kiwi itself does not need this information,
also the BootImage implementation does not need this information. Adding it to
BootImage felt odd to me as it is not needed.

I can understand the confusion you see when reading the BootImage changes.
I think I will change that code to make clear BootImage doesn't need root_partuuid

schaefi: Added a commit which hopefully makes it more clear ?

davidcassany: > I thought it would not violate the scope of the instance if stateful information that does not directly come from the image description is stored there too. This has the advantage that the information is available across all major steps of the image build because the xml_state object is handed over to any major class of the build process.

Right I buy it.

> Added a commit which hopefully makes it more clear ?

Yes thanks, I believe that with the current implementation I would have not been surprised or confused.

schaefi:
> Looks good to me  I like this new approach of having the choice to expand during the whole life cycle or just during the first deployment, nice!

Thanks, yes I hope this will be useful to users. I have run my tests again after the latest change and it still works as expected. Going to merge and write an e-mail on the mailing list

PR #1387, opened 2020-03-26
Add SECURE_BOOT parameter for grub2 in efi mode

davidcassany:
This commit adds the SECURE_BOOT parameter on bootloader sysconfig
for grub2.

Fixes bsc#1167746

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -483,6 +483,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         * LOADER_LOCATION
         * DEFAULT_APPEND
         * FAILSAFE_APPEND
+        * SECURE_BOOT
         """
         sysconfig_bootloader_entries = {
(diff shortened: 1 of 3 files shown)


Vogtinator:
IIRC `uefi` means secure boot while `efi` just means plain EFI without shim, which is also what is checked for in https://github.com/OSInside/kiwi/blob/e2cd4dfe9531f862411e712c59e1deb370977a9b/kiwi/bootloader/install/grub2.py#L241

So this should only be set to `yes` on `uefi` AFAICT. Otherwise it would fail on arm64, where `firmware="efi"`, but there's no `shim-install` yet.

schaefi:
yes this needs to be restricted to uefi

davidcassany:
Oh I see, I am fixing it, thanks for the insights

davidcassany:
@schaefi @Vogtinator ok, now I constrained the SECURE_BOOT parameter to `uefi` mode.

PR #1390, opened 2020-04-02
Allow overlay directories for profiles

schaefi:
In addition to the existing root/ overlay directory which
applies always there can now also be profile specific overlay
directories. If an overlay directory should be applied for
a specific profile this can now be done by placing this data
in a directory that is named the same as the profile name.


schaefi:
This feature will be needed for image descriptions that builds for multiple profiles and need to distinguish overlay data per profile. Right now this can only be done by providing tar archives that are referenced as ```<archive>``` in a profiled packages section of the XML description. As soon as you manage the description in a git there will be no binary archives anymore but directories with files in it. At this point latest you will need an opportunity to distinguish data that is connected to a profile.

PR #1393, opened 2020-04-11
Create qcow2 images in two steps

schaefi:
The creation of the qcow2 format was done in one qemu-img
convert call. That call instructs qemu to compress and
convert in one call. The downside of this approach is that
not all qcow2 options can be used. For example the setup
of:

```xml
<type ... formatoptions="preallocation=metadata"/>
```

failed the build with an error message that compression and
preallocation is not possible at the same time. Thus this patch
changes the way the qcow2 image is created to be done in two
steps. The first step converts the format without compression
and therefore allows for any format option to be used. The
second call only applies the compression and leads to the
final result.

### DIFF
--- a/kiwi/storage/subformat/qcow2.py
+++ b/kiwi/storage/subformat/qcow2.py
@@ -14,7 +14,7 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
-
+from tempfile import NamedTemporaryFile
 
 # project
 from kiwi.storage.subformat.base import DiskFormatBase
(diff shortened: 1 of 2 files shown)


schaefi:
David, I was experimenting with qcow2 performance settings for the box plugin and came across a better performance if we set preallocation=metadata. However that triggered the described problem in kiwi and I thought it's worth a general fix

schaefi:
@davidcassany if you could have a short look on that one too would be great. I think there is not much to say about it. I've tested the result in the box images and that works nicely. Thanks

davidcassany:
> @davidcassany if you could have a short look on that one too would be great.

Sure, sorry, this is on my list, too many things were needed early this week.

PR #1395, opened 2020-04-14
Call grub update wrapper script if present

schaefi:
grub reads its variables from /etc/default/grub but there
are also additional information in /etc/sysconfig. For
moving data to the right place some distributions uses
a wrapper script from perl-BootLoader. This commit checks
if the pbl script exists and checks if it can be called
with --refresh. If yes it's assumed this call moves grub
setup variables into to the right place such that the
follow up grub2-mkconfig call has access to all information
it needs to produce a correct grub.cfg file. This Fixes
bsc#1165960

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -24,6 +24,7 @@ import shutil
 from collections import OrderedDict
 
 # project
+from kiwi.utils.command_capabilities import CommandCapabilities
 from kiwi.bootloader.config.base import BootLoaderConfigBase
 from kiwi.bootloader.template.grub2 import BootLoaderTemplateGrub2
 from kiwi.command import Command
@@ -227,6 +228,24 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 self.boot_directory_name, 'grub.cfg'
             ]
         )
+        update_bootloader = self._get_update_bootloader_tool()
+        if update_bootloader and CommandCapabilities.has_option_in_help(
+            update_bootloader, '--refresh',
+            root=self.root_mount.mountpoint,
+            raise_on_error=False
+        ):
+            # Call wrapper script to update the bootloader setup
+            # The presence of the script with option --refresh
+            # makes calling it eligible. grub reads its variables from
+            # /etc/default/grub but there are also additional information
+            # in /etc/sysconfig. For moving data to the right place
+            # this update wrapper script exists.
+            Command.run(
+                [
+                    'chroot', self.root_mount.mountpoint,
+                    update_bootloader, '--refresh'
+                ]
+            )
         Command.run(
             [
                 'chroot', self.root_mount.mountpoint,
@@ -838,6 +857,11 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         for grub_mkconfig_tool in ['grub2-mkconfig', 'grub-mkconfig']:
             if Path.which(grub_mkconfig_tool, root_dir=self.root_dir):
                 return grub_mkconfig_tool
+
+    def _get_update_bootloader_tool(self):
+        update_bootloader = 'pbl'
+        if Path.which(update_bootloader, root_dir=self.root_dir):
+            return update_bootloader
 
     def _get_grub2_boot_path(self):
         return self.boot_dir + '/boot/' + self.boot_directory_name
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -410,10 +410,12 @@ class TestBootLoaderConfigGrub2:
     @patch.object(BootLoaderConfigGrub2, '_mount_system')
     @patch.object(BootLoaderConfigGrub2, '_copy_grub_config_to_efi_path')
     @patch('kiwi.bootloader.config.grub2.Command.run')
+    @patch('kiwi.bootloader.config.grub2.CommandCapabilities.has_option_in_help')
     def test_setup_disk_image_config(
-        self, mock_Command_run, mock_copy_grub_config_to_efi_path,
-        mock_mount_system
+        self, mock_has_option_in_help, mock_Command_run,
+        mock_copy_grub_config_to_efi_path, mock_mount_system
     ):
+        mock_has_option_in_help.return_value = True
         self.firmware.efi_mode = Mock(
             return_value='uefi'
         )
@@ -435,12 +437,20 @@ class TestBootLoaderConfigGrub2:
             mock_mount_system.assert_called_once_with(
                 'rootdev', 'bootdev', None, None
             )
-            mock_Command_run.assert_called_once_with(
-                [
-                    'chroot', self.bootloader.root_mount.mountpoint,
-                    'grub2-mkconfig', '-o', '/boot/grub2/grub.cfg'
-                ]
-            )
+            assert mock_Command_run.call_args_list == [
+                call(
+                    [
+                        'chroot', self.bootloader.root_mount.mountpoint,
+                        'pbl', '--refresh'
+                    ]
+                ),
+                call(
+                    [
+                        'chroot', self.bootloader.root_mount.mountpoint,
+                        'grub2-mkconfig', '-o', '/boot/grub2/grub.cfg'
+                    ]
+                )
+            ]
             mock_copy_grub_config_to_efi_path.assert_called_once_with(
                 'efi_mount_point', 'root_mount_point/boot/grub2/grub.cfg'
             )


Vogtinator:
I really do not think that this is necessary at all. AFAIK this is equivalent to `grub2-mkconfig` only.

schaefi:
> I really do not think that this is necessary at all. AFAIK this is equivalent to `grub2-mkconfig` only.

I thought so too but Robert observed a different behavior. @rjschwei are you really sure this is needed ?

schaefi:
@Vogtinator I would be really happy if we don't need this :)

Vogtinator:
In fact, there is no such option as `--refresh` to `pbl`, only the compat wrapper `update-bootloader` ignores unknown arguments and falls back to just updating the config. AFAICT all it does is:

```
    return 0 == $self->RunCommand (
        "/usr/sbin/grub2-mkconfig -o /boot/grub2/grub.cfg"
    );
```

Considering that this worked fine with the GA kiwi version, I really think that this is just bsc#1168480 again.

davidcassany:
I have doubts this will work.. but gonna give it a try. I believe the root cause is that `/etc/grub.d/10_linux` used by `grub2-mkconfig` is not suited for chroot environments as it runs checks against `/sys/firmware/efi`. This explains the failure on booting efi or secure efi, as `grub2-mkconfig` does not choose `linuxefi` command in OBS.

The more I think about it the more I think the fix should be placed in `/etc/grub.d/10_linux` even I understand this is a super sensitive area.

Vogtinator:
> I have doubts this will work.. but gonna give it a try. I believe the root cause is that /etc/grub.d/10_linux used by grub2-mkconfig is not suited for chroot environments as it runs checks against /sys/firmware/efi. This explains the failure on booting efi or secure efi, as grub2-mkconfig does not choose linuxefi command in OBS.

It must not choose `linuxefi` either, as then it won't boot on non-EFI anymore. That would break even with grub 2.04.

> The more I think about it the more I think the fix should be placed in /etc/grub.d/10_linux even I understand this is a super sensitive area.

That's been discussed somewhat and apparently not really possible for 2.02.

The issue is just that the config template grub ships with isn't compatible with "hybrid" images. That's only because of downstream patches though and with 2.04 the situation is much better already.

schaefi:
I also think the code here doesn't make sense. I just brought this up for testing and I agree with @Vogtinator. That on the other hand means there is not much we can do in kiwi to fix it

@rjschwei for the public cloud images we still have the opportunity to hotfix the initial grub.cfg. iirc that affects sle15-sp1 only ?

davidcassany:
> The issue is just that the config template grub ships with isn't compatible with "hybrid" images. That's only because of downstream patches though and with 2.04 the situation is much better already.

Well, feature wise, there are two different issues here:

1. Currently UEFI images are not booting on secure EFI if they are build in a non efi environment.
2. The provided grub.cfg on EFI/UEFI firmware is not compatible for all cases: non-efi, efi and efi secure.

Vogtinator:
Those two issues are actually one and the same.

The setting `SECURE_BOOT` in sysconfig only affects the behaviour of `update-bootloader`, which either runs `shim-install` or `grub2-install`. So it is only noticeable after updating the bootloader on a system.

davidcassany:
The patch provided here does not work on my environment (I am not using EFI on my host). Just run a UEFI build for SLE 15 SP1 (using JeOS image for kvm as reference) and the resulting image boots fine on non-efi, but fails to boot on secure efi (and I assume it also doesn't for efi).

If I manually edit the entry at boot time for efi I can easily boot, then run grub2-mkconfig and then it will work fine for efi boots, but fail on non efi. This is happening because we are missing the logic we used to have on the templated grub.cfg that was selecting `linux` or `linuxefi` at runtime. I don't think this was a good approach what it was doing the trick.

> Those two issues are actually one and the same.

I understand you point and probably both can be solved together as both point to the same weaknesses in grub.cfg generation. However from KIWI POV I could consider it reasonable that an EFI image does not boot in a non-efi environment, however an EFI image needs to be fine for EFI environments.

Vogtinator:
> However from KIWI POV I could consider it reasonable that an EFI image does not boot in a non-efi environment, however an EFI image needs to be fine for EFI environments.

At least for JeOS, it's a requirement that the images work both in non-EFI and (secure boot) EFI environments.

schaefi:
For that purpose we had this in the template

```
set linux=linux
set initrd=initrd
if [ "$${grub_cpu}" = "x86_64" -o "$${grub_cpu}" = "i386" ];then
    if [ "$${grub_platform}" = "efi" ]; then
        set linux=linuxefi
        set initrd=initrdefi
    fi
fi
```

but this can't be applied if we use the grub tool chain, unless we provide an /etc/grub.d/... script to include this logic. This can be done as part of an overlay file setup though

Vogtinator:
That's what @davidcassany proposed at https://bugzilla.suse.com/show_bug.cgi?id=1168480#c25 as well and I agreed that it would be the best option for now.

schaefi:
perfect, I will update our pubcloud images with that change

Vogtinator:
> perfect, I will update our pubcloud images with that change

AFAICT it can't be done at the image level as there is nothing (except the grub.d snippet itself) that runs after `grub2-mkconfig` to delete it again. It also has to be properly tested first...

schaefi:
yes I will add my thoughts in the Devel area first and then give it a test.

rjschwei:
> I also think the code here doesn't make sense. I just brought this up for testing and I agree with @Vogtinator. That on the other hand means there is not much we can do in kiwi to fix it
> 
> @rjschwei for the public cloud images we still have the opportunity to hotfix the initial grub.cfg. iirc that affects sle15-sp1 only ?

Yes, SLES 15 SP1 only, SLES 15 SP2 has grub 2.04 where this is not an issue.

rjschwei:
As far as update-bootloader is concerned, after digging a bit more I agree, it's a red herring, but I want to do more testing. So far I have 1 test where just running "grub2-mkconfig" failed to reboot when secure boot was enabled and one where it did succeed.

davidcassany:
> yes I will add my thoughts in the Devel area first and then give it a test.

@schaefi I did an ugly test of adding such a thing in KIWI, however it is quite ugly approach. It works according to the tests I did but it still results in hacky grub.cfg that is changed on any subsequent `grub2-mkconfig`, however I do not expect the image to change the firmware once it is already deployed, so it should not be a big deal feature wise, but still ugly IMHO.

schaefi:
> > yes I will add my thoughts in the Devel area first and then give it a test.
> 
> @schaefi I did an ugly test of adding such a thing in KIWI, however it is quite ugly approach. It works according to the tests I did but it still results in hacky grub.cfg that is changed on any subsequent `grub2-mkconfig`, however I do not expect the image to change the firmware once it is already deployed, so it should not be a big deal feature wise, but still ugly IMHO.

yeah I don't want something like that in kiwi. Thanks for the effort but I think it could be handled on the image level.

@rjschwei I have an update of the image description in Devel:PubCloud:GoogleCompute:ImagesSLE15/SLES15-SP1-GCE-BYOS and are testing this currently.

From my perspective this should work also with regards to kernel updates. It will most probably break on an update of the grub2 package. But in this case I hope that grub2 update updates to a version where we also don't have the issue.

I'll keep you posted

schaefi:
ok I have a working solution added to the following image description:

* https://build.suse.de/package/show/Devel:PubCloud:GoogleCompute:ImagesSLE15/SLES15-SP1-GCE-BYOS

I changed the following:

1) The root.tar.gz tarball
   
```
etc/grub.d/01_linuxefi
etc/grub.d/10_linux.patch
```

2) Apply of the grub.d patch above in config.sh

```bash
# Patch grub
pushd /etc/grub.d
patch -p0 < 10_linux.patch
popd
```

3) Added the patch program to the package list in config.kiwi

```xml
<package name="patch"/>
```

Together this does the following

* etc/grub.d/01_linuxefi
  handle linux vs. linuxefi and initrd vs. initrdefi dynamically on the grub level. when grub loads it decides for the right tool

* 10_linux.patch
  unfortunately needed to change the static use of linux and linuxefi to the use of a variable $linux

The result image has been tested as follows:

* Standard BIOS
  kvm -hda disk.raw -serial stdio

* EFI not secure boot
  kvm -hda disk.raw -bios /usr/share/qemu/ovmf-x86_64.bin -serial stdio

* EFI secure boot ($MS signed)
   kvm -hda disk.raw -bios /usr/share/qemu/ovmf-x86_64-ms.bin -serial stdio

Works in all three cases and is in my opinion safe for kernel updates. As said only a grub2 package update can kill it again

Let me know what you think

rjschwei:
@schaefi image tested and it boots with and without secure boot. The applied changes to the image work.

rjschwei:
On the update scenario; since `SECURE_BOOT=yes` is set even if grub gets updated everything is expected to work since in that case the generated grub.cfg file gets written with linuxefi.

Vogtinator:
> On the update scenario; since `SECURE_BOOT=yes` is set even if grub gets updated everything is expected to work since in that case the generated grub.cfg file gets written with linuxefi.

`SECURE_BOOT` has no impact on `linux`/`linuxefi`. That is decided based on whether EFI is detected by `grub2-mkconfig` and `GRUB_USE_LINUXEFI` in `/etc/default/grub`.

davidcassany:
@schaefi I like this approach, however this feels like manually applying a patch that could be part of grub2.

rjschwei:
> > On the update scenario; since `SECURE_BOOT=yes` is set even if grub gets updated everything is expected to work since in that case the generated grub.cfg file gets written with linuxefi.
> 
> `SECURE_BOOT` has no impact on `linux`/`linuxefi`. That is decided based on whether EFI is detected by `grub2-mkconfig` and `GRUB_USE_LINUXEFI` in `/etc/default/grub`.

Thanks. On running instances in GCE EFI is detected, have not checked on `GRUB_USE_LINUXEFI`

rjschwei:
@davidcassany I agree, but based on the discussion in bsc#1168480 it does not appear to be quite so easy. I am updating the relevant images to get the fix out the door.

schaefi:
@davidcassany you are completely right regarding the patch. I tried hard to avoid it but couldn't find an alternative solution. The patch as I did will for sure not be accepted on grub upstream. Yet another custom patch to the grub2 package for sles is in my opinion a step in the wrong direction. Given that grub on sle15-sp2 already works I think the only easy manageable way is at the image description level with the downside that each description needs to be touched that wants secure boot.

schaefi:
> > > On the update scenario; since `SECURE_BOOT=yes` is set even if grub gets updated everything is expected to work since in that case the generated grub.cfg file gets written with linuxefi.
> > 
> > 
> > `SECURE_BOOT` has no impact on `linux`/`linuxefi`. That is decided based on whether EFI is detected by `grub2-mkconfig` and `GRUB_USE_LINUXEFI` in `/etc/default/grub`.
> 
> Thanks. On running instances in GCE EFI is detected, have not checked on `GRUB_USE_LINUXEFI`

GRUB_USE_LINUXEFI is set by kiwi if the firmware is set to either efi or uefi.

We should be good now

@rjschwei I'm going to update all GCE pubcloud image and also the Azure images for SLE15-SP1 that uses uefi

Any objections to go down that route ?

Vogtinator:
Did @davidcassany's idea of only adding a new `grub.d` drop-in for `grub2-mkconfig` and deleting it again not work out?

I don't think that any approach which requires modifications to image configuration is acceptable.

Note that *all* EFI builds are broken, not just secure boot.

schaefi:
>Did @davidcassany's idea of only adding a new grub.d drop-in for grub2-mkconfig and deleting it >again not work out?

The grub.d drop in did a "set linux... / set initrd" but those variables are not used by 10_linux
but maybe I misunderstood the idea ?

> Note that all EFI builds are broken, not just secure boot.

In my tests efi always worked with linux and linuxefi on sle15-sp1. Did you observe a different behavior ?

> I don't think that any approach which requires modifications to image configuration is acceptable.

At the moment I don't see how else it should be done. We could also think of using a script
and use the editbootconfig attribute in kiwi to call it. In this case the image descriptions
would be extended by another script code

Vogtinator:
> > Did @davidcassany's idea of only adding a new grub.d drop-in for grub2-mkconfig and deleting it >again not work out?
> 
> The grub.d drop in did a "set linux... / set initrd" but those variables are not used by 10_linux
> but maybe I misunderstood the idea ?

Yes. https://bugzilla.suse.com/show_bug.cgi?id=1168480#c25 does not rely on variables.

> > Note that all EFI builds are broken, not just secure boot.
> 
> In my tests efi always worked with linux and linuxefi on sle15-sp1. Did you observe a different behavior ?

It's possible that without btrfs, grub is able to load `linux.mod` from the module directory. That would arguably be by accident though.

> > I don't think that any approach which requires modifications to image configuration is acceptable.
> 
> At the moment I don't see how else it should be done. We could also think of using a script
> and use the editbootconfig attribute in kiwi to call it. In this case the image descriptions
> would be extended by another script code

I can imagine multiple ways. Only `/boot/grub2/grub.cfg` in the image needs fixing, so in the worst case kiwi could just do some regex replacement in that file after calling `grub2-mkconfig`. There would be more elegant ways though.

schaefi:
>I can imagine multiple ways. Only /boot/grub2/grub.cfg in the image needs fixing, so in the worst >case kiwi could just do some regex replacement in that file after calling grub2-mkconfig. There >would be more elegant ways though.

ok you mean if we set GRUB_USE_LINUXEFI but the resulting grub.cfg does not use linuxefi/initrdefi we "fix" that. Actually that is a good idea and can work generically

davidcassany:
> The patch as I did will for sure not be accepted on grub upstream. Yet another custom patch to the grub2 package for sles is in my opinion a step in the wrong direction.

Completely agree, however note that linuxefi is not part of grub2 upstream, we are messing around with our own patches here. Anyway I can understand the resistance to touch this area as this is quite messy and hard evaluate and with high impact.

> Did @davidcassany's idea of only adding a new `grub.d` drop-in for `grub2-mkconfig` and deleting it again not work out?

Well the implementation is pretty ugly and not that simple. Take into account that other supported distros such as Fedora are not shipping linuxefi module anymore, thus this would require again some further checks beyond grub2 version check to verify if grub includes suse's downstream patches or not and according to that include some additional configuration file or not. I am not discarting it but it is not that simple. I can provide the PR for discussion.

> Note that _all_ EFI builds are broken, not just secure boot.

Yes you are right, this is still an issue

schaefi:
>Well the implementation is pretty ugly and not that simple. Take into account that other >supported distros such as Fedora are not shipping linuxefi module anymore, thus this would >require again some further checks beyond grub2 version check to verify if grub includes suse's >downstream patches or not and according to that include some additional configuration file or >not. I am not discarting it but it is not that simple. 

not really because all of that exists already. The condition under which we add GRUB_USE_LINUXEFI already exists

schaefi:
if we set GRUB_USE_LINUXEFI we can also validate if it really uses linuxefi and initrdefi commands and if not fix it. That should be a generic approach, shouldn't it ?

Vogtinator:
> if we set GRUB_USE_LINUXEFI we can also validate if it really uses linuxefi and initrdefi commands and if not fix it. That should be a generic approach, shouldn't it ?

Yup! Just prepending the code to set `$linux`/`$initrd` and doing `s/^(\s+)linux(\s)/\1$linux\2/` (also for `$initrd`) should work.

schaefi:
Here we go https://github.com/OSInside/kiwi/pull/1398

closing on behalf of #1397 1398

rjschwei:
> Did @davidcassany's idea of only adding a new `grub.d` drop-in for `grub2-mkconfig` and deleting it again not work out?
> 
> I don't think that any approach which requires modifications to image configuration is acceptable.
> 
> Note that _all_ EFI builds are broken, not just secure boot.

Our instances in Azure boot on Gen2 VMS and those use EFI, secure boot is not enabled. THus not sure about the "all EFI" image builds are broken.

Vogtinator:
> > Did @davidcassany's idea of only adding a new `grub.d` drop-in for `grub2-mkconfig` and deleting it again not work out?
> > I don't think that any approach which requires modifications to image configuration is acceptable.
> > Note that _all_ EFI builds are broken, not just secure boot.
> 
> Our instances in Azure boot on Gen2 VMS and those use EFI, secure boot is not enabled. THus not sure about the "all EFI" image builds are broken.

If secure boot is disabled and the grub modules are not on btrfs, it might work by accident as it loads `linux.mod`.

PR #1398, opened 2020-04-15
Validate use of GRUB_USE_LINUXEFI

schaefi:
Validate use of GRUB_USE_LINUXEFI
    
On systems that uses GRUB_USE_LINUXEFI with grub2 version
less than 2.04 there is no support for dynamic EFI
environment checking. In this condition we extend the grub
setup to add this support. The change kiwi does is as
follows:
    
* Apply only on grub < 2.04
   1. Modify 10_linux to set linux/initrd as variables
   2. Write hybrid setup as 01_efihybrid
    
This Fixes bsc#1165960

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+from string import Template
 import re
 import os
 import logging
@@ -24,6 +25,7 @@ import shutil
 from collections import OrderedDict
 
 # project
+from kiwi.utils.command_capabilities import CommandCapabilities
 from kiwi.bootloader.config.base import BootLoaderConfigBase
 from kiwi.bootloader.template.grub2 import BootLoaderTemplateGrub2
 from kiwi.command import Command
@@ -140,6 +142,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         self.cmdline = None
         self.iso_boot = False
         self.shim_fallback_setup = False
+        self.validate_use_of_linuxefi = False
 
     def write(self):
         """
@@ -234,6 +237,41 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 config_file.replace(self.root_mount.mountpoint, '')
             ]
         )
+
+        if self.validate_use_of_linuxefi:
+            # On systems that uses GRUB_USE_LINUXEFI with grub2 version
+            # less than 2.04 there is no support for dynamic EFI
+            # environment checking. In this condition we change the
+            # grub config to add this support as follows:
+            #
+            # * Apply only on grub < 2.04
+            #    1. Modify grub.cfg to set linux/initrd as variables
+            #    2. Prepend hybrid setup to select linux vs. linuxefi on demand
+            #
+            # Please note this is a one time modification done by kiwi
+            # Any subsequent call of the grub config tool will overwrite
+            # the setup and disables dynamic EFI environment checking
+            # at boot time
+            if not CommandCapabilities.check_version(
+                self._get_grub2_mkconfig_tool(),
+                version_waterline=(2, 4), raise_on_error=False
+            ):
+                with open(config_file) as grub_config_file:
+                    grub_config = grub_config_file.read()
+                    grub_config = re.sub(
+                        r'([ \t]+)linux[efi]*([ \t]+)', r'\1$linux\2',
+                        grub_config
+                    )
+                    grub_config = re.sub(
+                        r'([ \t]+)initrd[efi]*([ \t]+)', r'\1$initrd\2',
+                        grub_config
+                    )
+                with open(config_file, 'w') as grub_config_file:
+                    grub_config_file.write(
+                        Template(self.grub2.header_hybrid).substitute()
+                    )
+                    grub_config_file.write(grub_config)
+
         if self.root_reference:
             if self.root_filesystem_is_overlay or \
                Defaults.is_buildservice_worker():
@@ -581,6 +619,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 if use_linuxefi_implemented.returncode == 0:
                     grub_default_entries['GRUB_USE_LINUXEFI'] = 'true'
                     grub_default_entries['GRUB_USE_INITRDEFI'] = 'true'
+                    self.validate_use_of_linuxefi = True
         if self.xml_state.build_type.get_btrfs_root_is_snapshot():
             grub_default_entries['SUSE_BTRFS_SNAPSHOT_BOOTING'] = 'true'
         if self.custom_args.get('boot_is_crypto'):
--- a/kiwi/utils/command_capabilities.py
+++ b/kiwi/utils/command_capabilities.py
@@ -103,10 +103,11 @@ class CommandCapabilities:
         try:
             command = Command.run(arguments)
             for line in command.output.splitlines():
-                match = re.search('[0-9]+(\.[0-9]+)*', line)
-                if match:
+                matches = re.findall(r'([0-9]+(\.[0-9]+)*)', line)
+                if matches:
+                    match = max([m[0] for m in matches], key=len)
                     version_info = tuple(
-                        int(elt) for elt in match.group(0).split('.')
+                        int(elt) for elt in match.split('.')
                     )
                     break
             if version_info is None:
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -1,4 +1,5 @@
 import io
+import os
 import logging
 from mock import (
     patch, call, MagicMock, Mock
@@ -12,6 +13,7 @@ import kiwi
 from kiwi.xml_state import XMLState
 from kiwi.xml_description import XMLDescription
 from kiwi.bootloader.config.grub2 import BootLoaderConfigGrub2
+from kiwi.bootloader.template.grub2 import BootLoaderTemplateGrub2
 
 from kiwi.exceptions import (
     KiwiBootLoaderGrubPlatformError,
@@ -78,7 +80,9 @@ class TestBootLoaderConfigGrub2:
             return_value='0xffffffff'
         )
 
+        grub_template = BootLoaderTemplateGrub2()
         self.grub2 = Mock()
+        self.grub2.header_hybrid = grub_template.header_hybrid
         kiwi.bootloader.config.grub2.BootLoaderTemplateGrub2 = Mock(
             return_value=self.grub2
         )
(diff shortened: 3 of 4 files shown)


kiwi/bootloader/config/grub2.py
```diff
@@ -253,6 +254,22 @@ def setup_disk_image_config(
...
 
+        if self.validate_use_of_linuxefi:
+            # On Systems that uses GRUB_USE_LINUXEFI we check if the
+            # grub2 config creation tool was able to detect that
+            # linuxefi/initrdefi commands have to be used. If not
+            # we fix this condition
+            with open(config_file) as grub_config_file:
+                grub_config = grub_config_file.read()
+                grub_config = re.sub(
+                    r'([ \t]+)linux', r'\1linuxefi', grub_config
```
Vogtinator, line 265: This would replace `linuxefi` with `linuxefiefi`.

schaefi: jaha my bad, needs fixing, thanks

schaefi: should be better now

schaefi:
we are turning in circles. if this needs to be dynamic it's grub job to write a config that is dynamic and that brings us back to the concept of "set linux / set initrd" as variables in the grub config.

why can't we just submit grub 2.04 to sle15-sp1 which clears all that mess ?

Vogtinator:
> we are turning in circles. if this needs to be dynamic it's grub job to write a config that is dynamic and that brings us back to the concept of "set linux / set initrd" as variables in the grub config.

Yes, but due to os-prober that is not currently possible (based on a comment by mchang, I don't know anything about os-prober). What I described at https://github.com/OSInside/kiwi/pull/1395#issuecomment-613898804 would only require a handful of lines more than this.

> why can't we just submit grub 2.04 to sle15-sp1 which clears all that mess ?

https://bugzilla.suse.com/show_bug.cgi?id=1168480#c22 - you just closed the circle again :P

schaefi:
ok I got you, give me another 5 minutes

kiwi/bootloader/config/grub2.py
```diff
@@ -227,13 +230,54 @@ def setup_disk_image_config(
...
+                self._get_grub2_mkconfig_tool(),
+                version_waterline='2.04', raise_on_error=False
+            ):
+                linux_config = os.sep.join(
+                    [self.root_mount.mountpoint, 'etc/grub.d/10_linux']
+                )
+                with open(linux_config) as grub_linux_config_file:
+                    grub_linux_config = grub_linux_config_file.read()
+                    grub_linux_config = re.sub(
+                        r'([ \t]+)linux[efi]*([ \t]+)', r'\1\$linux\2',
```
Vogtinator, line 254: Would also match `linuxeieieif`, but shouldn't actually cause an issue AFAICT

schaefi:
@davidcassany @Vogtinator I've update the code, let's review again

I'm not sure if the version check is ok in any situation. 

```
grub2-mkconfig --version
grub2-mkconfig (GRUB2) 2.02
```

In any case which is less than 2.04 we apply the code from here.

Thanks

kiwi/bootloader/config/grub2.py
```diff
@@ -227,13 +230,54 @@ def setup_disk_image_config(
...
+
+        if self.validate_use_of_linuxefi:
+            # On systems that uses GRUB_USE_LINUXEFI with grub2 version
+            # less than 2.04 there is no support for dynamic EFI
+            # environment checking. In this condition we extend the grub
+            # setup to add this support. The change kiwi does is as
+            # follows:
+            #
+            # * Apply only on grub < 2.04
+            #    1. Modify 10_linux to set linux/initrd as variables
```
Vogtinator, line 242: This should run after `grub2-mkconfig` got called and only modify `grub.cfg`. There must not be any permanent changes inside `/etc/grub.d/` as they wouldn't be tracked.

Vogtinator:
> @davidcassany @Vogtinator I've update the code, let's review again
> 
> I'm not sure if the version check is ok in any situation.

Looking good to me, though it might make sense to move the check to the `self.validate_use_of_linuxefi = True` assignment.

kiwi/bootloader/config/grub2.py
schaefi: yes I know but thought it doesn't matter

kiwi/bootloader/config/grub2.py
schaefi: if we modify the written grub.cfg once and leave 10_linux as is then the next call of grub2-mkconfig will make the $linux variable setup ineffective.

Vogtinator: Yup, but at that point it's also not necessary anymore as it'll decide correctly between `linux`/`linuxefi` in the running system.

schaefi:
>I think you misunderstood the suggestion. Until the call of grub2-config, nothing changes. Only the generated grub.cfg needs to be modified, nothing else.

But how is this useful if the next grub2-mkconfig call overwrites all that ? Let's say you run on an EFI system and call grub2-mkconfig, it now changes the config to use linuxefi. Now reboot the system into CSM(BIOS) mode and it will fail due to the missing dynamic selection.

I think the reason for all of this is to keep the system bootable no matter what firmware is active at boot time. This requires the grub.cfg to dynamically check every time. grub2-mkconfig as it is right now doesn't do that, it creates the setup for the situation at call time.

Something is still weird

Vogtinator:
> > I think you misunderstood the suggestion. Until the call of grub2-config, nothing changes. Only the generated grub.cfg needs to be modified, nothing else.
> 
> But how is this useful if the next grub2-mkconfig call overwrites all that ? Let's say you run on an EFI system and call grub2-mkconfig, it now changes the config to use linuxefi. Now reboot the system into CSM(BIOS) mode and it will fail due to the missing dynamic selection.
>
> I think the reason for all of this is to keep the system bootable no matter what firmware is active at boot time. This requires the grub.cfg to dynamically check every time. grub2-mkconfig as it is right now doesn't do that, it creates the setup for the situation at call time.
> 
> Something is still weird

Yup, that's exactly how it's always been since the introduction of `linuxefi` in grub (which I still consider a bad thing...). The previous templated `grub.cfg` written by kiwi was also overwritten by `grub2-mkconfig` the same way.

Only the first boot of the image has to be "universal", as after that it's unlikely to switch boot mode. This isn't supported by other deployments (install with YaST or so) either.

schaefi:
> Only the first boot of the image has to be "universal",

ok in this case we can change the code

> linuxefi in grub (which I still consider a bad thing...

I fully agree, what a mess

schaefi:
@Vogtinator ok back to just grub.cfg modification. I also added a note in the comment of the kiwi code to make that clear.

Vogtinator:
Looking good, thanks! I'll give it a try

schaefi:
great, thanks for testing

Vogtinator:
Failed during image building:

```
[   93s] [ ERROR   ]: 13:00:45 | Unexpected error:
[   93s] Traceback (most recent call last):
...
[   93s]   File "/usr/lib/python3.6/site-packages/kiwi/utils/command_capabilities.py", line 120, in check_version
[   93s]     return version_info >= version_waterline
[   93s] TypeError: '>=' not supported between instances of 'tuple' and 'str'
```

I assume it should be `(2,4)` instead?

schaefi:
> I assume it should be (2,4) instead

yes must be a tuple (major,minor,release) I'll fix it

schaefi:
I build with the fix and wondered why the grub.cfg was not changed... then I realized I built against TW which has grub 2.04, so the version check worked :-)

schaefi:
needs testing on some older distro...

schaefi:
running Leap 15.1 test now

schaefi:
worked for me

kiwi/bootloader/config/grub2.py
```diff
@@ -234,6 +237,41 @@ def setup_disk_image_config(
...
+            #    1. Modify grub.cfg to set linux/initrd as variables
+            #    2. Prepend hybrid setup to select linux vs. linuxefi on demand
+            #
+            # Please note this is a one time modification done by kiwi
+            # Any subsequent call of the grub config tool will overwrite
+            # the setup and disables dynamic EFI environment checking
+            # at boot time
+            if not CommandCapabilities.check_version(
+                self._get_grub2_mkconfig_tool(),
+                version_waterline=(2, 4), raise_on_error=False
```
davidcassany, line 257: I was just writing about the use of a tuple instead of using a string for versioning when this got fixed :+1:. 

However I just noted the regular expression that checks the version is not robust enough for grub2. As it takes the `2` from the `grub2`. The issue can be seen with:

```python
match = re.search('[0-9]+(\.[0-9]+)*', 'grub2-mkconfig (GRUB2) 2.02')
```
This will match first digit found `2` thus it won't actually distinguish between 2.02 and 2.04. Probably the best solution is use `re.findall` instead of search and just keep the longest match. Is it robust enough? I'd like to avoid having complex regex here.

davidcassany: This could be implemented with something like

```python
matches = re.findall(r'([0-9]+(\.[0-9]+)*)', 'grub2-mkconfig (GRUB2) 2.2')
match = max(matches)[0]
```
This returns the longest match of the whole expression.

schaefi: @davidcassany ok could you do me a favor and directly push to this branch to fix the version checkup ? Thanks

davidcassany: sure

Vogtinator:
Tested SLE 15 SP1 JeOS, built and booted fine on legacy, EFI and EFI with secure boot. Also worked after `update-bootloader --reinit && reboot`.

kiwi/utils/command_capabilities.py
```diff
@@ -103,10 +103,11 @@ def check_version(
         try:
             command = Command.run(arguments)
             for line in command.output.splitlines():
-                match = re.search('[0-9]+(\.[0-9]+)*', line)
-                if match:
+                matches = re.findall(r'([0-9]+(\.[0-9]+)*)', line)
```
Vogtinator, line 106: Are the added `()` necessary?

kiwi/utils/command_capabilities.py
```diff
@@ -103,10 +103,11 @@ def check_version(
...
+                if matches:
+                    match = max(matches)[0]
```
Vogtinator, line 108: I'm not too fluent in python - does this do a length comparison and not a lexical sort?

kiwi/utils/command_capabilities.py
rjschwei: The () are for grouping in the regular expression which makes it easier to get stuff after any "." that are in the version.

kiwi/utils/command_capabilities.py
davidcassany: Good point over tuples of strings this is no clear which criteria is being used. I am fixing it.

kiwi/utils/command_capabilities.py
davidcassany: I have been also surprised by that see the following example:

```python
>>> import re
>>> re.findall(r'([0-9]+(\.[0-9]+)*)', 'grub2-mkconfig (GRUB2) v2.02')
[('2', ''), ('2', ''), ('2.02', '.02')]
>>> re.findall(r'[0-9]+(\.[0-9]+)*', 'grub2-mkconfig (GRUB2) v2.02')
['', '', '.02']
```

schaefi:
JFI: the broken gitlab pipeline is caused by gitlab not by us. They are working on it

kiwi/utils/command_capabilities.py
Vogtinator: I just checked: Unlike `re.search`, `re.findall` returns the groups only, not the full match. So the `()` on the outside are indeed necessary.

kiwi/utils/command_capabilities.py
Vogtinator: Just checked, it's a lexical sort:

```
max(["z", "aa"])
'z'
```

So probably not working as desired?

rjschwei: Finds the maximum, given that the re.findall() returns as list it will pick the item in the list that has the "highest" value. The way the regular expression is written that means max() will find first the tuple in the list that has the most populated entries and if there are multiple tuples with the same number of entries it will pick the one that has the highest numerical value in the first entry in the tuple, then compare the second entry and so on. It's a length first approach, yes. The way the regex is written it should be that way.

Vogtinator: @rjschwei The issue is that `"aaaaa" < "z"`, so the length didn't actually have an influence at all.

That's fixed now though.

rjschwei:
@Vogtinator yes but, the re.findall only returns numbers, as strings, in tuples in a list. Anyway updated way works as well.

Vogtinator:
> @Vogtinator yes but, the re.findall only returns numbers, as strings, in tuples in a list.

Yes, so it would've returned the highest (lexically sorted) version, while the intention was to pick the longest match:

> Probably the best solution is use re.findall instead of search and just keep the longest match.

rjschwei:
Well that's the way findall works

`re.findall(r'([0-9]+(\.[0-9]+)*)', 'grub2-mkconfig (GRUB2) 1.2.3.44')
[('2', ''), ('2', ''), ('1.2.3.44', '.44')]`

then max() with the index [0] as it was implemented returned the longest match. Anyway, same thing just a different way to get there.

Vogtinator:
> Well that's the way findall works
> 
> ```
> re.findall(r'([0-9]+(\.[0-9]+)*)', 'grub2-mkconfig (GRUB2) 1.2.3.44')
> [('2', ''), ('2', ''), ('1.2.3.44', '.44')]
> ```
> then max() with the index [0] as it was implemented returned the longest match.

Here it returns the highest (lexical) version:

```
>>> max([('2', ''), ('2', ''), ('1.2.3.44', '.44')])
('2', '')
>>> max([('2', ''), ('2', ''), ('1.2.3.44', '.44')])[0]
'2'
```

rjschwei:
Glad I concocted a "failing" example where once again the Python sorting algorithm gives me a surprise. Learned something new.

schaefi:
Ok looks like everybody is ok with this. I keep fingers crossed that we found a "solution" everybody is happy with. Thanks a ton for all your help, feedback and testing. I'll merge commits and let it settle until tomorrow, then I'll merge if we are still ok with it :)

schaefi:
I'll keep the commits, they make sense as they are. Thanks

schaefi:
I added a unit test to test the version check for non obvious data

PR #1399, opened 2020-04-15
PoC for adding a KIWI specific grub.d config file

davidcassany:
This PR adds a new `/etc/grub.d/05_linux_loader` configuration file just at runtime for disk builds on EFI firmware. It only applies if grub includes LINUXEFI support.


davidcassany:
note this is just a PoC and only covers the case the image is build in a non EFI host.

davidcassany:
Closing in favor of #1398

PR #1400, opened 2020-04-15
Fixed check for grub mkconfig capabilities

schaefi:
The check for the capabilities of the tool were applied to
the tool installed on the host but the later call of the
tool will be done with the tool inside the image root


schaefi:
@davidcassany at least something that makes sense to fix :)

PR #1402, opened 2020-04-16
Fixed return value from _get_grub2_mkconfig_tool

schaefi:
The method returned the basename of the tool if it could
be found by Path.which(). But the method's scope has been
changed in a way that the return value of the method must
be the result from Path.which() to allow working on the
full path name.


schaefi:
@davidcassany this is an important follow up on the change introduced in #1400

schaefi:
This one also must be part of the porting to SLE

schaefi:
Thanks for review, will merge it now. Remember we need this in the porting to sle12

PR #1403, opened 2020-04-16
Increase spare space on disk repart

schaefi:
The sizing of the virtual cylinders in parted seems to be unfavorable,
as with some disks and SD cards here the device size is not a multiple
of the cylinder size, so the last incomplete cylinder is wasted.
If this wasted space is more than 5MiB, kiwi tries to resize indefinitely.
Therefore min_additional_mbytes gets increased to prevent running
into this situation. This Fixes bsc#1165823


schaefi:
ok let's go for it

PR #1414, opened 2020-04-27
Restructure pxe doc

### LINKED ISSUE #1346: Documentation: OEM PXE boot images, .append and .config.bootoptions
## Problem description

The OEM documentation is describing how to configure the boot server, but there are some mismatches with the current status:

* For the boot system image is referencing the .config.bootoptions file:

```
scp LimeJeOS-Leap-15.1.x86_64-1.15.1.config.bootoptions PXE_SERVER_IP:/srv/tftpboot/image/
```

As today, this file is not present under this name, but with the `pxeboot.` prefix, that indicate that is part of the initial pxeboot process:

```
pxeboot.LimeJeOS-Leap-15.1.x86_64-1.15.1.config.bootoptions
```

There is also a new file, with extension `.append`, and there is missing documentation about where this file belongs, and where needs to be copied.

```
LimeJeOS-Leap-15.1.x86_64-1.15.1.append
```

Untar this one: https://download.opensuse.org/repositories/systemsmanagement:/yomi/images_pxe/

schaefi:
**Restructure pxe documentation**

Moved and merged chapters to isolate the netboot infrastructure to be one deployment method based on pxe which needs the images initrd, kernel and system to be created by kiwi

**Added a new KIS chapter**

A new chapter to document KIS images. KIS is an abbreviation
for Kernel, Initrd, System and defines a highly customizable
image consisting out of these components. This Fixes #1346


schaefi:
@davidcassany for review I suggest to build the docs `tox -e doc` and read the kis chapter as well as look into the org changes. The relevant changes on the code to support kis I will ad in an extra commit

doc/source/working_with_kiwi/xml_description.rst
```diff
@@ -192,8 +192,8 @@ sections):
 - `vmx`: build a virtual machine image, see: :ref:`vmx`
 - `oem`: results in an expandable image that can be deployed via a bootable
   installation medium, e.g. a USB drive or a CD. See :ref:`oem`
-- `pxe`: creates an image that can be booted via PXE (network boot), see
-  :ref:`build_pxe`
+- `pxe`: creates an image that can be booted from the network using the
+  legacy kiwi netboot infrastructure, see :ref:`build_legacy_pxe`
```
davidcassany, line 196: Shouldn't we include the `kis`type here?

doc/source/building/build_kis.rst
```diff
@@ -0,0 +1,81 @@
...
+
+   $ sudo kiwi-ng --type kis system build \
+       --description kiwi-descriptions/suse/x86_64/{exc_description} \
+       --target-dir /tmp/myimage
+
+The resulting image components are saved in the folder :file:`/tmp/myimage`.
+
+.. note::
+
+   For testing the components of a KIS image the customer needs to
```
davidcassany, line 75: Wouldn't make sense to include here an example of a qemu call to boot it? The only reason I see to not include it here is that the filesystem image might not be there, but isn't it still a good example for people to let them try the new type?

doc/source/working_with_kiwi/xml_description.rst
schaefi: done

doc/source/building/build_kis.rst
schaefi: Hmm, I did not include that part because we have intentionally no scope for the kis type.
Instead I added a note at the end that says:

```
.. note::

   For testing the components of a KIS image the customer needs to
   provide a deployment infrastructure and also the deployment
   process by themselve. One example of a deployment infrastructure
   using PXE is provided by {kiwi} with the `netboot` infrastructure.
   However that netboot infrastructure is no longer developed and
   only kept for compatibiliy reasons. For details
   see :ref:`build_legacy_pxe`
```

So what generic qemu testing method would you suggest ?

davidcassany: I'd suggest something like

```
qemu-kvm -kernel <kernel file> -initrd <initrd file> -append $(cat <append file>) -hda <filesystem file> -m 1024
```

The only thing that is probably clearly not generic enough here is the filesystem image as this is optional. I like the note and makes sense, but probably would also make sense to provide a simple test example as the above. The above should be valid for any of our testing images (assuming a kis type is added with filesystem). I like it because is simple and makes clear the purpose of the build binaries.

I just bring this up because this is what I would have done. I am not having a strong opinion regarding that. I understand that this does not cover the use case without a filesystem. That case falls in the note you wrote.

schaefi: ok that worked for me. I have tested booting a kis image through qemu and that worked. Interesting approach to attach a rootfs file as disk but qemu managed to handle it. ok I also updated the docs regarding this part

PR #1415, opened 2020-04-27
Added new build type kis

### LINKED ISSUE #1346: Documentation: OEM PXE boot images, .append and .config.bootoptions

schaefi:
KIS is an abbreviation for Kernel, Initrd, System and defines a
highly customizable image consisting out of these components.
This commit performs the changes documented in #1414 and
introduces the new kis type. From an image build perspective
kis is currently the same as pxe with restrictions for kis
on the schema level. A kis build uses dracut and does not allow
to use the legacy netboot initrd. The pxe type will therefore be
exclusively used to built for the legacy netboot infrastructure
and is on its way to deprecation in the future.
This Fixes #1346

### DIFF
--- a/doc/source/api/kiwi.builder.rst
+++ b/doc/source/api/kiwi.builder.rst
@@ -54,10 +54,10 @@ Submodules
     :undoc-members:
     :show-inheritance:
 
-`kiwi.builder.pxe` Module
+`kiwi.builder.kis` Module
 -------------------------
 
-.. automodule:: kiwi.builder.pxe
+.. automodule:: kiwi.builder.kis
     :members:
     :undoc-members:
     :show-inheritance:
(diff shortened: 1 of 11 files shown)


schaefi:
@davidcassany this makes the documented change effective in code. For the user this completely transparent, nothing changes on the image="pxe" level and there is a new type image="kis" that is handled by the same code that builds pxe. However I have renamed anything pointing to pxe in code now to be kis because the legacy netboot pxe stuff the pxe image type is associated with will be on its way to deprecation.

Hope this makes sense to you too

PR #1418, opened 2020-04-28
Fixed spec file header information

schaefi:
Fixed copyright date/company and url path to project

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -1,7 +1,7 @@
 #
 # spec file for package kiwi
 #
-# Copyright (c) 2016 SUSE LINUX GmbH, Nuernberg, Germany.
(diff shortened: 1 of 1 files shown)


package/python-kiwi-spec-template
```diff
@@ -1,7 +1,7 @@
 #
 # spec file for package kiwi
 #
-# Copyright (c) 2016 SUSE LINUX GmbH, Nuernberg, Germany.
+# Copyright (c) 2020 SUSE Software Solutions Germany GmbH, Nuernberg, Germany.
```
dcermak, line 4:
```suggestion
# Copyright (c) 2020 SUSE LLC
```

schaefi: ok

PR #1420, opened 2020-04-28
Do not set GRUB_DISTRIBUTOR

### LINKED ISSUE #1416: kiwi sets GRUB_DISTRIBUTOR again
## Problem description

kiwi sets GRUB_DISTRIBUTOR /etc/default/grub

## Expected behaviour

kiwi not setting GRUB_DISTRIBUTOR, that behavior was disabled previously:
https://bugzilla.suse.com/show_bug.cgi?id=1032119&GoAheadAndLogIn=1
by commit
https://github.com/OSInside/kiwi/commit/295a7faaf3c485d4bd9e49638b9b113a41050cde

however https://github.com/OSInside/kiwi/commit/162d5c9a4120860c49e7c0de36bef838123bd28d introduced it again

Build openSUSE:Leap:15.2:MicroOS/openSUSE-MicroOS, it produces a grub menu that says "openSUSE MicroOS" instead of "openSUSE MicroOS 15.2":
https://openqa.opensuse.org/tests/1247233#step/disk_boot/1

+ KIWI version: python3-kiwi-9.20.5-lp152.1.1.x86_64
+ Operating system: Leap 15.2
+ OBS version: ?

schaefi:
grub mkconfig creates a proper string based on /etc/os-release
if GRUB_DISTRIBUTOR is not set. We assume this to be true
for all distributions. This Fixes #1416

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -574,13 +574,11 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         * GRUB_CMDLINE_LINUX_DEFAULT
         * GRUB_GFXMODE
         * GRUB_TERMINAL
-        * GRUB_DISTRIBUTOR
         """
         grub_default_entries = {
             'GRUB_TIMEOUT': self.timeout,
             'GRUB_GFXMODE': self.gfxmode,
-            'GRUB_TERMINAL': '"{0}"'.format(self.terminal),
-            'GRUB_DISTRIBUTOR': '"{0}"'.format(self.get_menu_entry_title())
+            'GRUB_TERMINAL': '"{0}"'.format(self.terminal)
         }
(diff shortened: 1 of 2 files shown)


schaefi:
@Conan-Kudo could you double check if that doesn't create a problem on Fedora ? Thanks

Conan-Kudo:
@schaefi Code looks good, haven't checked if the resulting image is fine yet, though I think it will be.

Conan-Kudo:
cc: @martinezjavier @dwmccheyne

martinezjavier:
> cc: @martinezjavier @dwmccheyne

@schaefi looks good to me. We also don't add use `GRUB_DISTRIBUTOR` in Fedora and rely on `/etc/os-release`.

schaefi:
perfect, thanks for the feedback :+1:

PR #1423, opened 2020-04-30
Clean AnonymousUniqueId from the image

### LINKED ISSUE #1396: /var/lib/zypp/AnonymousUniqueId not deleted if packages are uninstalled
If there are packages in the `uninstall` section, `/var/lib/zypp/AnonymousUniqueId` is recreated after it is deleted in `config.sh` as part of `suseCleanup` or a plain call to `rm -rf`.

Not sure what could be done about this, but it might need to be documented.

davidcassany:
This commit is two fold.

For one side does a small refactor to move the deletion of custom RPM
macros to package manager level inside a cleaning method. This way only
RPM based package managers run RPM specific code and each package manager
can apply its own specific logic.

On the other hand for the zypper package manager the deletion of
/var/lib/zypp/AnonymousUniqueId file has been added as part of the
new cleaning method.

Fixes #1396

### DIFF
--- a/kiwi/package_manager/base.py
+++ b/kiwi/package_manager/base.py
@@ -205,6 +205,15 @@ class PackageManagerBase:
         """
         return True if returncode != 0 else False
 
+    def clean_leftovers(self):
+        """
+        Cleans package manager related data not needed in the
+        resulting image such as custom macros
+
+        Implementation in specialized package manager class
+        """
+        pass
+
     def cleanup_requests(self):
         """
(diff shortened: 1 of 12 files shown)


Vogtinator:
As it's now deleted here, you can drop it from https://github.com/OSInside/kiwi/blob/4c4d80ac1e39246357418a2cb93607f9e83e50d3/kiwi/config/functions.sh#L1116

PR #1426, opened 2020-05-04
Clean AnonymousUniqueId

davidcassany:
This is a follow up commit for 469f71144. There is not need to remove
the `/var/lib/zypp/AnonymousUniqueId` file from the clean up method
since this has been already integrated inside the python code base
and this is not needed anymore as a helper function.


davidcassany:
Thanks @vogtinator for noticing this leftover from the previous PR

schaefi:
yep :+1:

PR #1429, opened 2020-05-06
Fix support for multiple dracut modules

patjak:
Passing the dmsquash flag adds dmsquash-live and livenet modules to
dracut. This broke when adding a check to only add modules that dracut
reports as available. Use a list instead of a string to represent the
modules to add.

Fixes: 07ea23a4

### DIFF
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -185,15 +185,16 @@ class LiveImageBuilder:
 
         # create dracut initrd for live image
         log.info('Creating live ISO boot image')
-        live_dracut_module = Defaults.get_live_dracut_module_from_flag(
+        live_dracut_modules = Defaults.get_live_dracut_modules_from_flag(
             self.live_type
         )
-        self.boot_image.include_module('pollcdrom')
-        self.boot_image.include_module(live_dracut_module)
+        live_dracut_modules.append('pollcdrom')
+        for dracut_module in live_dracut_modules:
+            self.boot_image.include_module(dracut_module)
         self.boot_image.omit_module('multipath')
         self.boot_image.write_system_config_file(
             config={
-                'modules': ['pollcdrom', live_dracut_module],
+                'modules': live_dracut_modules,
                 'omit_modules': ['multipath']
             },
             config_file=self.root_dir + '/etc/dracut.conf.d/02-livecd.conf'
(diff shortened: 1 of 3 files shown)


kiwi/builder/live.py
```diff
@@ -185,15 +185,16 @@ def create(self):
...
-        self.boot_image.include_module('pollcdrom')
-        self.boot_image.include_module(live_dracut_module)
+        live_dracut_modules.append('pollcdrom')
+        for dracut_module in live_dracut_modules:
+            self.boot_image.include_module(dracut_module)
         self.boot_image.omit_module('multipath')
         self.boot_image.write_system_config_file(
             config={
-                'modules': ['pollcdrom', live_dracut_module],
+                'modules': live_dracut_modules,
```
davidcassany, line 197: You are missing the `pollcdrom` module here. I bet this is why unit tests are failing here.

patjak: Hi, thanks for review.

pollcdrom is appended to the list (line 191) and is listed as included when I'm building my image so should not be a problem here.

I rarely use python so not sure how to run the unit tests properly. I just did "python3 -m pytest test/unit/defaults_test.py" which succeeds for me. What did I miss?

Thanks
Patrik

davidcassany: Oh you are right, pollcdrom is there. I just quickly checked the unit tests and saw pollcdrom related failures and assumed you had missed it.

We manage the development environment using tox, thus for checking tests you can run `tox -e unit_py3_6` if you are on python3.6 or `tox -e unit_py3_7` if you are on python3.7 (those are the two python versions that we currently use, but we keep adding new ones and remove old ones as we need).

Then, regarding the specific failure after having a closer look I'd say the issue is the order. With your code the order of the included modules has changed (before pollcdrom was the first one) thus the unit test that checks the `boot_image.include_module` calls should be adapted to the new order. I'd say this could be solved inverting the order of https://github.com/OSInside/kiwi/blob/451d39a237e0f671600baf447d067dde78bcb47c/test/unit/builder/live_test.py#L215

patjak: Yes, that's it and I also had to change the order on line https://github.com/OSInside/kiwi/blob/451d39a237e0f671600baf447d067dde78bcb47c/test/unit/builder/live_test.py#L220
I'll send you a new PR

PR #1431, opened 2020-05-06
Fix support for multiple dracut modules (v2)

patjak:
Passing the dmsquash flag adds dmsquash-live and livenet modules to
dracut. This broke when adding a check to only add modules that dracut
reports as available. Use a list instead of a string to represent the
modules to add.

Fixes: 07ea23a4


schaefi:
Looks great to me :+1: good catch, thanks

@davidcassany could you double check ?

PR #1432, opened 2020-05-06
Adding support for Pacman package manager

### LINKED ISSUE #1098: Add support for Arch Linux images / Pacman package manager
## Problem description
Currently, kiwi supports neither Arch Linux nor Pacman package manager.
It will be nice to have such support :)

+ KIWI version: 9.17.39
+ Operating system: openSUSE Leap 15.1
+ OBS version: -

davidcassany:
This commit adds support for pacman package manager and, in
consequence, support for Arch Linux images. It also adds
the package definition for Arch Linux.

Fixes #1098

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,10 @@ build: clean tox
 	# update package version in spec file
 	cat package/python-kiwi-spec-template | sed -e s'@%%VERSION@${version}@' \
 		> dist/python-kiwi.spec
+	# update package version in PKGBUILD file
+	md5sums=$$(md5sum dist/python-kiwi.tar.gz | cut -d" " -f1); \
+	cat package/python-kiwi-pkgbuild-template | sed -e s'@%%VERSION@${version}@' \
+		-e s"@%%MD5SUM@$${md5sums}@" > dist/PKGBUILD
 	# provide rpm rpmlintrc
 	cp package/python-kiwi-rpmlintrc dist
 
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -672,6 +672,7 @@ class Defaults:
         """
         return [
             '/usr/share/syslinux',
+            '/usr/lib/syslinux/bios',
             '/usr/lib/syslinux/modules/bios',
             '/usr/lib/ISOLINUX'
         ]
@@ -1470,6 +1471,8 @@ class Defaults:
             return 'rpm'
         elif package_manager in deb_based:
             return 'dpkg'
+        elif package_manager == 'pacman':
+            return 'pacman'
 
     def get(self, key):
         """
--- a/kiwi/package_manager/__init__.py
+++ b/kiwi/package_manager/__init__.py
@@ -21,6 +21,7 @@ import logging
 from kiwi.package_manager.zypper import PackageManagerZypper
 from kiwi.package_manager.apt import PackageManagerApt
 from kiwi.package_manager.dnf import PackageManagerDnf
+from kiwi.package_manager.pacman import PackageManagerPacman
 
 from kiwi.exceptions import (
     KiwiPackageManagerSetupError
@@ -50,6 +51,8 @@ class PackageManager:
             manager = PackageManagerDnf(repository, custom_args)
         elif package_manager == 'apt-get':
             manager = PackageManagerApt(repository, custom_args)
+        elif package_manager == 'pacman':
+            manager = PackageManagerPacman(repository, custom_args)
         else:
             raise KiwiPackageManagerSetupError(
                 'Support for package manager %s not implemented' %
--- a/kiwi/package_manager/pacman.py
+++ b/kiwi/package_manager/pacman.py
@@ -1,0 +1,244 @@
+# Copyright (c) 2019 SUSE Linux GmbH.  All rights reserved.
+#
+# This file is part of kiwi.
+#
+# kiwi is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# kiwi is distributed in the hope that it will be useful,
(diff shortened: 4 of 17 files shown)


davidcassany:
Finally the Arch Pull Request is here :) This code can be seen in action and tested in this [OBS Project](https://build.opensuse.org/package/show/home:dcassany:Arch/archlinux-images). In that project the integration with OBS is done thanks to a build time service that prepares the local built time repositories for Arch. Equivalent changes in the `obs-build` project over the KIWI receipt should be applied in order to make the integration transparent and make the `obs-service-arch_repos_builder` unneeded, see [X]() pull request.

The example image descriptions are also included in [X] Pull Request.

A couple of notes regarding the presented Arch Linux support implementation:

### Repositories definition in image description XML

There are two different ways to define repositories for Pacman in KIWI schema:

1. Use components to identify repositories:

```xml
<!-- Compenents usage example in OBS-->
<repository components="Arch_Community_standard">
	<!-- Notice repository component matches the source path -->
  <source path="obs://Arch:Community/standard"/>
</repository>
```
```xml
<!-- Compenents usage example for builds common builds -->
<repository components="core extra community">
    <source path="http://ftp.heanet.ie/mirrors/ftp.archlinux.org/$repo/os/$arch"/>
 </repository>
```

2. Use alias to name repositories:

```xml
<!-- Compenents usage example in OBS -->
<repository alias="Arch_Community_standard">
	<!-- Notice repository component matches the source path -->
  <source path="obs://Arch:Community/standard"/>
</repository>
```

```xml
<!-- Compenents usage example for builds common builds -->
<repository alias="core">
    <source path="http://ftp.heanet.ie/mirrors/ftp.archlinux.org/$repo/os/$arch"/>
 </repository>
<!-- Compenents usage example for builds common builds -->
<repository alias="extra">
    <source path="http://ftp.heanet.ie/mirrors/ftp.archlinux.org/$repo/os/$arch"/>
 </repository>
```

How repositories are mapped in OBS (this could eventually change on behalf of [this PR](https://github.com/openSUSE/obs-build/pull/563)) is a bit tricky since the repository name (component or alias if component is not present) has to match the name of the repository database file.

In Arch Linux the repository name and the database file included within the repository must be the same. This creates a dependency between repository name and repository contents that doesn't happen in the other supported repository types (e.g. Apt, RPM). In order to keep repository code creation in OBS decoupled of the KIWI repository components and aliases a convention is required. Thus in OBS the repository name has to match the OBS project and repository names as in the above OBS examples.


### Kernel, initrd and dracut adoption

In Arch Linux dracut is not yet officially adopted, even it is available and there are plans to adopt it. In Arch Linux the kernel files and initrd files that are expected to be located in `/boot` are not owned by any package, in fact, those files are created or copied in post transaction hook during kernel related package installation. The hooks that actually doing so are part of the the `mkinitcpio` (the current Arch Linux initr creation tool), however the dracut package does not include equivalent hooks yet, see [dracut docs](https://wiki.archlinux.org/index.php/Dracut#Generate_a_new_initramfs_on_kernel_upgrade). Thus, in any case, the dracut installation and kernel installation is not sufficient to get `/boot` populated, custom hooks are required. In the example images a custom package is provided to work for KIWI images.

The examples suggested by Arch Linux wiki and Arch Linux comunity do are not functional for KIWI builds since they use a different criteria for kernel naming and initrd naming compared with other common linux distributions. In fact in distributions like openSUSE the kernel and initrd include a dash separated suffix that matches the kernel version. This is not the case for Arch Linux, the initrd and kernel, both include a dash separated suffix that matches the kernel package name, not kernel version. KIWI convention for initrd is hardcoded and has the following criteria: `<kernel_basename>-<kernel_version>` (e.g. `vmlinuz-5.5.7-1-default` ). In Arch Linux the criteria is `<kernel_basename>-<kernel_packagename>` (e.g. `vmlinuz-linux`). Initrd and kernel name must follow the same schema according the grub2 templates, thus, from KIWI PoV it means that the dracut hooks are required to follow the kernel version suffix criteria, otherwise grub2 will not match the kernel with the initrd. This is unfortunate, as I expect that some day mid term dracut is fully adopted in Arch Linux and if they keep using the same naming conventions as they have been doing up to now this will break KIWI builds :( To solve that KIWI would require more flexible initrd naming convention and a a slightly different criteria to determine the kernel version.

schaefi:
wow, thanks much David, a lot of effort at several places :+1: I haven't looked into it detailed but have one question. Do you plan to move https://build.opensuse.org/project/show/home:dcassany:Arch to Virtualization:Appliances:Images:Testing_x86:archlinux ? I'm asking because that would be the natural place for the integration testing. Also the name "Arch" should be better the long version "archlinux" just to avoid confusion with arch == processor-arch

It looks really great, I'll come back to it soon. Thanks

davidcassany:
> Do you plan to move https://build.opensuse.org/project/show/home:dcassany:Arch to Virtualization:Appliances:Images:Testing_x86:archlinux

Sure, if we merge it then tests will come with a follow up request. Also, I am pending on  openSUSE/obs-build#563 probably changes are still about to come in OBS area. As I mentioned more than once I do not want to merge this without having a clear OBS procedure for builds.

Take your time to have a look, I don't think there is any hurry :) 

My concern regarding Archlinux is the dracut adoption... it is not clear to me how it will end up. I believe their naming criteria for kernel file and initrd will hit us at some point, as Arch Linux does not append the kernel version in those files. Now I could sort it out as dracut is available in Arch but not yet fully adopted, so dracut installation needs to be customized anyways.

Anyway I am loving to see how a new distro can be added without being that intrusive in code, note that mostly I just needed to add new classes, no changes in KIWI logic.

schaefi:
As you solved the inconsistencies compared to other distributions by an extra dracut hook package I'm totally fine with this approach. I'm going to merge this now and submit a new staging package such that you can go ahead with the integration tests. Thanks much

Issue #1401, opened 2020-04-15
Add configurable GRUB serial settings


nickhuber:
## Problem description

The GRUB serial speeds are always set to 9600 baud. from

        self.header_serial = dedent('''
            serial --speed=9600 --unit=0 --word=8 --parity=no --stop=1
        ''').strip() + os.linesep

in bootloader/template/grub2.py

## Expected behaviour

The GRUB serial speeds should be configurable in the XML configuration.

## Steps to reproduce the behaviour

Make an ISO with ``console=ttyS0,115200n8`` on the kernelcmdline, and fail to see the GRUB menu because of the mismatched serial settings

## OS and Software information
+ KIWI version: 9.20.5
+ Operating system: Fedora 31
+ OBS version: N/A

schaefi:
yes you are right, we should make this more configurable

jloeser:
Would it make sense to introduce a bootloader element with respective attributes?

```
<type image="oem" bootloader="grub2" ...>
  <bootloader speed="115200" parity="false" ... timeout="0"/>
  ...
```
(this could also tackle https://github.com/OSInside/kiwi/issues/1404)

Alternatively we could use the kernel's parameter format:
```
console=         ...
                        Use the specified serial port.  The options are of
                        the form "bbbbpnf", where "bbbb" is the baud rate,
                        "p" is parity ("n", "o", or "e"), "n" is number of
                        bits, and "f" is flow control ("r" for RTS or
                        omit it).  Default is "9600n8".
```
```
<... bootloader_serial="115200n8" ...>
```

The former one looks more promising to me - but I am not that familiar with this XML scheme stuff. What's your guy's opinion on that?

davidcassany:
What I am wondering if it makes sense to use different parameters for the kernel command line and bootloader. Should we force them to be equivalent? If so I'd say that probably the best option is parsing the `kernelcmdline` and use the `console=` definition from there if present, if not use the current defaults.

But if there is not need to enforce serial console to be equivalent in kernel command and the bootloader I'd say we can add a new attribute to set if for the bootloader. I like the idea to use the kernel notation because I'd say users are already quite familiar with it and in most cases I'd say they want the kernel paramters to be equivalent with the bootloader paramters.

[...]

PR #1433, opened 2020-05-08
Refactor xml bootloader settings format

schaefi:
The bootloader settings are handled through attributes in the type element. Over the years some attributes were added and  there are requests for more settings. Therefore the bootloader   setup deservers its own section. With this commit the schema  changes from v7.1 to v7.2 and moves bootloader, bootloader_console   boottimeout and zipl_targettype into a new bootloader subsection below type. The commit also adds an auto transformation template   such that customers don't have to change their image descriptions.  This is related to Issue #1401

The new layout looks like this:

```xml
<type ...>
    <bootloader name="..." ...
</type>
```

Of course an auto translation of the former 7.1 schema into the new 7.2 schema will be done automatically

### DIFF
--- a/build-tests/arm/fedora/test-image-iso/appliance.kiwi
+++ b/build-tests/arm/fedora/test-image-iso/appliance.kiwi
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<image schemaversion="7.1" name="ISO-Fedora">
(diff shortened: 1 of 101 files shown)


schaefi:
From here we can go forward and fix the open grub settings issues. I know many files got touched but this is mostly update of XML data to the new schema. From a review perspective you can focus on the changes in code

PR #1435, opened 2020-05-11
Support grub timeout_style parameter

### LINKED ISSUE #1404: Support time out style setting for GRUB
## Problem description

Support time out style setting for GRUB
https://www.gnu.org/software/grub/manual/grub/html_node/timeout.html

schaefi:
Grub supports a style setting that influences the display of
the menu depending on the configured timeout value. With this
patch kiwi allows to specify the style via a new bootloader
parameter named timeout_style="hidden|countdown". If not set
the grub default applies which shows the menu in any case.
This Fixes #1404


kiwi/xml_parse.py
```diff
@@ -730,6 +730,7 @@ class k_packagemanager_content(object):
     ZYPPER='zypper'
```
schaefi, line 730: no review needed... auto generated code

schaefi: @davidcassany would still be good if you can double check because I realized generateDS also added information for the pacman support you added with #1432 and I was confused that this data got changed. Nevertheless all auto generated changes looks good to me

davidcassany: interesting, this is happening because I did not regenerate the xml_parse on my Arch Linux branch after 09e84933e. Before that commit adding a new choice element values had no impact at all on `xml_parse.py` generated code, thus xml_parse was not part of the Arch Linux branch diff. However, recent generateDS versions (we started using an updated version in 09e84933e) include the list of choices defined for elements.

The funny fact is that it never stopped working, thus I am wondering this code portion is actually being used in any way :P I kept rebasing from master on regular basis and never saw complains about not having this enum up to date. :man_shrugging:

schaefi: ok, really a funny one :)

PR #1436, opened 2020-05-11
Allow setup of serial line in bootloader

### LINKED ISSUE #1401: Add configurable GRUB serial settings

schaefi:
So far kiwi used a static serial line setup of the console
type was configured to use a serial console. However if
different parameters are needed there was no way to change
the static setup. This commit adds a new bootloader attribute
called serial_line which allows to specify the serial line
parameters for the grub bootloader e.g

    serial_line="serial --speed=9600 --unit=0 --word=8 --parity=no --stop=1"

Please note serial_line takes the value as it is given and
adds this as a line to the grub configuration or as content
to the variable GRUB_SERIAL_COMMAND. No further validation
of the information takes place by kiwi. This Fixes #1401


schaefi:
In new bootloader style this looks like like the following example:

```xml
<bootloader name="grub2" serial_line="serial --speed=9600 --unit=0 --word=8 --parity=no --stop=1"/>
```

Please note the complete serial command as supported by grub2 needs to be entered here.
I added it that way because the GRUB_SERIAL_COMMAND variable in ```/etc/default/grub``` handles it the same way.

I'm not super happy with it but this is the most flexible solution we can go with given that grub changes might come along our way

schaefi:
@davidcassany please give this a decent review as we are also changing behavior slightly. I double checked on the static value which was used before and it seems that was the default setting in grub anyway... not sure though

schaefi:
>no serial console line with defaults is appended, just an empty line is appended. In that case I would have appended only serial, without any flag. This way a simple serial configuration is also possible and, IMHO, it provides better backward compatibility for ISO images that include serial console. 

yes makes perfect sense. I forgot to use "serial" as the default setting needed in the template case. Will adapt. Thanks

schaefi:
> Note this is not happening for the non templated configs

yes because grub2-mkconfig is smart enough to handle that properly by the value of GRUB_TERMINAL

schaefi:
@davidcassany I pushed a patch to fix the issue you found

schaefi:
> This is not what I had in mind

yeah I was spinning my head if it would be wise to handle the serial setup with some clever code on our side. I really think this will not pay off. The serial setup of the kernel and the serial setup of the bootloader should be treated as two different components. I know the serial line is old and its parameters are the same for any subsystem. However as you can already see there is no standard at all in specifying speed, parity, stop-bit ... If we go that route and set the standard it will be our job to handle the information for the kernel and any bootloader that supports a serial setup. I think this is not kiwi's job and I would rather see repeating information in the description than dealing with the problem if at some point in time grub decides to incompatible change the parameters of the serial command.

I know that in the area of bootloaders the people don't care much for configuration standards or backward compatibility. I think this is because of the hardware agnostic area where not much high level programming standards apply. We will have a hard time keeping this consistent and functional and that's why I think we should not be the clever part at this point. If I look how much of a mess grub2-mkconfig or /etc/default/grub is I'm really happy that I'm just a user of it

I hope I don't stand on somebodies toes but from a setup perspective the low level bootloader tools have a lot room for improvement. As long as I don't see this there I'm not too enthusiastic trying to add logic on top of it

Thoughts ?

davidcassany:
Tests are failing due to a formatting issue see #1441

schaefi:
I know I know... too fast push

davidcassany:
> I know I know... too fast push

:sweat_smile: I did not wanted to stress anyone I was just leaving the same message in all failing PRs to avoid having more than one person looking at the same failure :pray:

davidcassany:
> I know that in the area of bootloaders the people don't care much for configuration standards or backward compatibility. I think this is because of the hardware agnostic area where not much high level programming standards apply. We will have a hard time keeping this consistent and functional and that's why I think we should not be the clever part at this point. If I look how much of a mess grub2-mkconfig or /etc/default/grub is I'm really happy that I'm just a user of it

Yes I agree, that's why I considered your implementation more robust to what I had in mind. I completely agree and with your approach we gain more flexibility and a more maintainable code. Hopefully in most regular x86_64 cases using the defaults will just be sufficient, so the `serial_line` attribute might not be needed in most cases.

PR #1437, opened 2020-05-11
Support multiple markup formats

schaefi:
Allow to read multiple markup formats. Supported are XML and YAML
The parsing is based on the anymarkup module. The code does not create a hard
dependency on the module and uses it if available

This is a site project of mine. As the current code changes are relatively small and the
opportunity to convert descriptions to the latest schema and or different formats is
nice I've opened this up.

To convert a given image description into latest schema and/or different markup you
can call the following command:

```bash
kiwi-ng image info --description kiwi-descriptions/suse/x86_64/suse-leap-15.2-JeOS/ \
    --print-xml
```

To do the same but also convert into yaml you can call

```bash
kiwi-ng image info --description kiwi-descriptions/suse/x86_64/suse-leap-15.2-JeOS/ \
    --print-yaml
```

Either result can be used as image description to build the image. For any other than an XML
description the name of the file that contains the description should now use the notation:

```
file.kiwi
```

As with anymarkup installed kiwi can read any markup :)


schaefi:
I've done some test builds and that worked amazingly good. The internal format of course always stays xml, which is also the reason why I translate the parse result from anymarkup to xml and work with the result like before. If you give it xml the anymarkup translation is doing nothing except for xml syntax validation on write

schaefi:
of course any validation and rule checks apply on the converted xml data, this could be confusing if yaml is used as input markup and a validation error from e.g jing tells you something about attributes and sections. But I think users will still be able to see what is wrong just line numbers things like that will not match the origin if it's not xml

schaefi:
An interesting behavior is the use of ordered maps. See https://pypi.org/project/anymarkup/.
I think we don't need this since the order is of no relevance in our use case. However I did not find a way how to prevent !!omap as documented here: https://yaml.org/type/omap.html

schaefi:
A roadblock is probably the issue that anymarkup does not exist in the distribution (looked up suse)

davidcassany:
Great! I see you  finally bought the idea of converting across markup formats :) This is fun a probably it helps on some people to get into KIWI. I'd say the main interesting formats are YAML, JSON (are they actually different beyond the look and feel?).

Gonna play with it a little and share my thoughts. What I like for this idea is that it can be done without being invasive as we can keep using the xml internally.

davidcassany:
Tests are failing due to a formatting issue see #1441

schaefi:
I've changed the code to be plugin based. Meaning if anymarkup module is installed you get full support for all formats, if not you need to stick with xml. I realized anymarkup is not packaged and I don't want to introduce another requirement.

Just play a bit with it, for me personally this is really not something the world needs. But the code is easy and I would be open to add it like that. Of course some words in the documentation is missing and to make it useful the "kiwi image info" command should have a "show" or similar command to allow people to display the description in different formats.

schaefi:
A ```kiwi image info description --format xml|yaml``` would also be nice to see the converted results, as it includes all style sheet processing. So maybe a useful thing in the end :)

schaefi:
> Great! I see you finally bought the idea of converting across markup formats :) 

well more driven by the opportunity to easily update old descriptions to the latest schema. But yes we now can handle any markup. After I solved the !!omap issue I think the yaml is a nice read. People can choose whatever they prefer

schaefi:
>The YaML <-> XML convertion is quite nice, obviously it suffers from some format conventions (@ for attributes) as yaml is lacking categories in comparison to XML. Repository and packages conversion are fine but curious.

right but that's what people have to accept when they want to go with yaml. The only way to change this is to deal with different schemata which I will not accept. There is only one schema and that's XML. It will stay the internal format which all our validation and stylesheet processing is based on. Also all data structures are generated based on that schema using the generateDS module. So whatever format is used to describe the image it will always be converted to match that schema.

I was also surprised how attributes and sections got differentiated in yaml but I think they found a very nice way to do it :-)

If at that point people don't want to use kiwi with yaml because it's aligned to an XML schema I must admit that I won't care any further. I see this as a matter of given and take and the solution provided here is enough accommodation from our side.

Do you agree with me on that end or do you think when we merge this there will be more requests that we should use a different layout for the yaml format which we then can only decline ? From my perspective the yaml is nice and self explaining enough to be used when converted from an existing description. Writing a description from scratch in yaml without knowing the XML structure behind is however not an easy task. I don't think people write image descriptions from scratch though and that's also why we maintain a descriptions git repository-

I'm waiting with the merge for your feedback regarding the above.

Thanks

davidcassany:
> Do you agree with me on that end or do you think when we merge this there will be more requests that we should use a different layout for the yaml format which we then can only decline ? From my perspective the yaml is nice and self explaining enough to be used when converted from an existing description. Writing a description from scratch in yaml without knowing the XML structure behind is however not an easy task. I don't think people write image descriptions from scratch though and that's also why we maintain a descriptions git repository-

Well, yes I agree that we can't change our underlaying XML and that this is our ground truth, all validations and description processes should happen from the XML. I have doubts about if we should advert it louder and encourage people to use yaml files, I'd say no, this should be more an alternative for users that definitely ask for it and prefer/need yaml files for some reason.

I think we might get request for different yaml layout if users start using it. In that case I'd say we have to be strict on that we only support XML schema layout and that yaml files are just an automated conversion of those XMLs. At most we can relay on that the converter tools evolve to something different or that they can be slightly configured, but this is not KIWI code.

So I think we are on the same page regarding that. IMHO we can merge that

schaefi:
Great :+1: thanks yes we are on the same page here. I'll merge it now

PR #1444, opened 2020-05-13
Cleanup use of machine platform name

### LINKED ISSUE #1438: Kiwi failed to create a 32-bit VDI from a raw disk image
## Problem description
Kiwi could not create a 32-bit VDI image. This occured first when using Kiwi in a chroot on a 64-bit host , but I reproduced the issue with a local `osc build` on another image.
In the second case Kiwi fails with (I cut out the lines of no interest to the problem):
```
[  305s] [ INFO    ]: 19:04:56 | Creating vdi Disk Format
[  305s] [ DEBUG   ]: 19:04:56 | EXEC: [qemu-img convert -f raw /usr/src/packages/KIWI-vmx/OpenSUSE-Tumbleweed-Kiwi.i686-0.1.raw -O vdi /usr/src/packages/KIWI-vmx/OpenSUSE-Tumbleweed-Kiwi.i686-0.1.vdi]
[  305s] [ DEBUG   ]: 19:04:56 | EXEC: Failed with stderr: qemu-img: Could not open '/usr/src/packages/KIWI-vmx/OpenSUSE-Tumbleweed-Kiwi.i686-0.1.raw': Could not open '/usr/src/packages/KIWI-vmx/OpenSUSE-Tumbleweed-Kiwi.i686-0.1.raw': No such file or directory
[  305s] , stdout: (no output on stdout)
[  305s] [ ERROR   ]: 19:04:56 | KiwiCommandError: qemu-img: stderr: qemu-img: Could not open '/usr/src/packages/KIWI-vmx/OpenSUSE-Tumbleweed-Kiwi.i686-0.1.raw': Could not open '/usr/src/packages/KIWI-vmx/OpenSUSE-Tumbleweed-Kiwi.i686-0.1.raw': No such file or directory
[  305s] , stdout: (no output on stdout)
[  305s] [ INFO    ]: 19:04:56 | Cleaning up BootImageDracut instance
[  305s] [ INFO    ]: 19:04:56 | Cleaning up BootImageDracut instance
[  305s] [ INFO    ]: 19:04:56 | Cleaning up BootImageDracut instance
[  306s] 
[  306s] host.domain failed "build OpenSUSE-Tumbleweed-Kiwi.kiwi" at Mon May 11 19:04:57 UTC 2020.
[  306s] 
```
## Expected behaviour
Kiwi should convert the raw disk image into a VDI.

I think creating a 32-bit VDI image will always fail (see below the results of my investigations).

+ KIWI version: 9.20.9
+ Host operating system: OpenSUSE Leap 15.1 64-bit (up to date 11/05/2020)
+ Build host (used by `osc build`): OpenSUSE Tumbleweed 32-bit (up to date 11/05/2020)
+ OBS version: 0.168.2 (`osc --version`)

## My understanding
Despite what it says a raw disk image is created, but it is named `OpenSUSE-Tumbleweed-Kiwi.ix86-0.1.raw` and not `OpenSUSE-Tumbleweed-Kiwi.i686-0.1.raw` as expected by `qemu-img`.
I think this is due to the fact that https://github.com/OSInside/kiwi/blob/947b741748ea0d5d0df8da26337bbb7070f7fdee/kiwi/storage/subformat/base.py#L50 uses directly `platform.arch()` whereas https://github.com/OSInside/kiwi/blob/14ee5ecbb7953de9cbfb099c21faf31f4da64de1/kiwi/builder/disk.py#L72-L74 modifies this value for i586 and i686 architectures.
The two should be in sync, if I am not mistaken. Maybe it would be worth doing a function and not using directly `platform.machine()`, because I saw a lot of lines identical to the second one all over kiwi code.

schaefi:
in kiwi we handle 32bit x86 architecture names as ix86. This is
done by checking i586 and i686 32bit arch names. However those
checks are spread over the entire kiwi code base and should be
consolidated into one method. The cleanup of those arch name
usage fixes an inconsistency between the arch name used in the
disk builder and the arch name used in the subformat image
formats. This Fixes #1438

### DIFF
--- a/kiwi/boot/image/base.py
+++ b/kiwi/boot/image/base.py
@@ -17,7 +17,6 @@
 #
 import re
 import os
-import platform
 import pickle
 import logging
 from collections import namedtuple
@@ -69,7 +68,7 @@ class BootImageBase:
         self.initrd_base_name = ''.join(
             [
                 self.xml_state.xml_data.get_name(),
-                '.' + platform.machine(),
+                '.' + Defaults.get_platform_name(),
                 '-' + self.xml_state.get_image_version(),
                 '.initrd'
             ]
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -17,7 +17,6 @@
 #
 import os
 import logging
-import platform
 from collections import namedtuple
 
 # project
@@ -45,7 +44,7 @@ class BootLoaderConfigBase:
         self.root_dir = root_dir
         self.boot_dir = boot_dir or root_dir
         self.xml_state = xml_state
-        self.arch = platform.machine()
+        self.arch = Defaults.get_platform_name()
 
         self.volumes_mount = []
         self.root_mount = None
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -19,7 +19,6 @@ from string import Template
 import re
 import os
 import logging
-import platform
 import glob
 import shutil
 from collections import OrderedDict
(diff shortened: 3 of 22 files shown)


schaefi:
Yeah the report from #1438 triggered all this. Actually he saw that this deserves a change. I'm not 100% sure no side effect for 32bit is introduced which we don't see by the unit testing but as 32bit x86 is also fading away I think we can fix on an individual base if an issue pops up. Thus I weight the refactoring to be more important also with regards to further arch handling.

Let's keep the fingers crossed :) Thanks for the review

PR #1445, opened 2020-05-14
Fixed spec file

### LINKED ISSUE #1413: Obsolete python2-kiwi in spec file
## Problem description

Before py2 was EOL kiwi supported both versions and built a sub-package named python2-kiwi. After we dropped support for py2 some version of this package is still around. Thus the current spec file should obsolete python2-kiwi

## Expected behaviour

python2-kiwi is not used anymore

schaefi:
This patch is two fold. First the py2 version of kiwi was
dropped since py2 is EOL. To indicate that correctly on the
package level python3-kiwi has to obsolete python2-kiwi.
The other part of the change is a file conflict of the
files:

*  etc/bash_completion.d/kiwi-ng.sh
*  /usr/share/doc/packages/python-kiwi/README

which were provided by the kiwi-man-pages sub-package but
were moved to be provided by the main python3-kiwi package
now. On update of the package with an older version of
kiwi that maintains this files to belong to kiwi-man-pages
a file conflict at install time appears. To solve this
python3-kiwi now conflicts with kiwi-man-pages < 9.20.5
This Fixes #1413 and Fixes bsc#1168973

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -80,6 +80,9 @@ and cloud systems like Xen, KVM, VMware, EC2 and more.
 %package -n python%{python3_pkgversion}-kiwi
 Summary:        KIWI - Appliance Builder Next Generation
 Group:          Development/Languages/Python
+Obsoletes:      python2-kiwi
+Conflicts:      python2-kiwi
+Conflicts:      kiwi-man-pages < %{version}
 %if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version} || 0%{?debian} || 0%{?ubuntu}
 Recommends:     jing
 %endif


Conan-Kudo:
@schaefi Perhaps this is the right time to explore unifying the Fedora spec with the upstream kiwi one. I did a significant reorganization of the the packaging when I packaged this in Fedora a while ago...

That said, I think this fix is fine for now.

schaefi:
@Conan-Kudo yeah I'm happy if we can do this. As you started that work it would be great if you can drive this change. Unification sounds great to me

package/python-kiwi-spec-template
```diff
@@ -80,6 +80,8 @@ and cloud systems like Xen, KVM, VMware, EC2 and more.
 %package -n python%{python3_pkgversion}-kiwi
 Summary:        KIWI - Appliance Builder Next Generation
 Group:          Development/Languages/Python
+Obsoletes:      python2-kiwi
```
davidcassany, line 83: IIRC with @Conan-Kudo we talked about including `Conflicts: python2-kiwi` together with the `Obsoletes` a while ago, as result of that it this approach was suggested in https://bugzilla.suse.com/show_bug.cgi?id=1156677 

Also would be nice to mention bsc#1156677 in the commit message.

package/python-kiwi-spec-template
```diff
@@ -80,6 +80,8 @@ and cloud systems like Xen, KVM, VMware, EC2 and more.
...
+Conflicts:      kiwi-man-pages < 9.20.5
```
davidcassany, line 84: I am wondering if it wouldn't be more generic to do something like `Conflicts: kiwi-man-pages < %{version}`. This way, in addition of solving the former issue, we force some alignment between kiwi-man-pages package and python3-kiwi which is, IMHO, not bad.

package/python-kiwi-spec-template
schaefi: ok done

package/python-kiwi-spec-template
schaefi: yes better

PR #1446, opened 2020-05-15
Fix signing keys management for APT

### LINKED ISSUE #1440: Kiwi cannot set keys for APT repositories
## Problem description
As discussed in the mailing list, Kiwi cannot add keys for APT repositories. 

## Expected behaviour
Kiwi should be able to add keys for APT repositories

Add keys on the command line when building an image based on Debian as follows
```
sudo kiwi --debug --color-output --type=vmx system build --description=KiwiJeOS-Debian-Buster --target-dir=/tmp/KiwiJeOS-Debian-Buster --signing-key KiwiJeOS-Debian-Buster/keys/debian-10-key.asc --signing-key KiwiJeOS-Debian-Buster/keys/obs.key
```
You should get the following error just after `debootstrap`
```
[ DEBUG   ]: 22:02:52 | EXEC: [debootstrap --no-check-gpg --variant=minbase --components=main,contrib,non-free buster /tmp/KiwiJeOS-Debian-Buster/build/image-root.debootstrap http://deb.debian.org/debian/]
[ WARNING ]: 22:04:02 | Extended attributes not supported for target: /tmp/KiwiJeOS-Debian-Buster/build/image-root
[ DEBUG   ]: 22:04:02 | EXEC: [rsync -a -H /tmp/KiwiJeOS-Debian-Buster/build/image-root.debootstrap/ /tmp/KiwiJeOS-Debian-Buster/build/image-root]
[ DEBUG   ]: 22:04:03 | EXEC: [chroot /tmp/KiwiJeOS-Debian-Buster/build/image-root apt-key add KiwiJeOS-Debian-Buster/keys/debian-10-key.asc]
[ DEBUG   ]: 22:04:03 | EXEC: Failed with stderr: E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation
[ DEBUG   ]: 22:04:03 | EXEC: [rm -r -f /tmp/KiwiJeOS-Debian-Buster/build/image-root.debootstrap]
[ ERROR   ]: 22:04:03 | KiwiDebootstrapError: KiwiCommandError: chroot: stderr: E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation
```

+ KIWI version:  9.18.16
+ Operating system: OpenSUSE Leap 15.1
+ OBS version: Not relevant

## Root cause
Using `apt-key` to register the keys to apt just after `debootstrap`ing does not work since it requires `gnupg` which is not installed by `deboostrap`.

## Workaround
I solved the issue by preloading the keys in `/var/cache/kiwi/apt-get/trusted.gpg` (which is again a keyring in `gnupg1` format).

### First solution
Implement the workaround in Kiwi.
### Second solution
Install `gnupg` using `debootstrap` APT configuration before using `apy-key`.

davidcassany:
This commit fixes the management of the trusted keyring for apt
repositories. It creates a `trusted.gpg` keyring with the provided
signing keys so APT can check against that the configured repositories.

Fixes #1440

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -134,14 +134,12 @@ class PackageManagerApt(PackageManagerBase):
             # debootstrap takes care to install apt-get
             self.package_requests.remove('apt-get')
         try:
-            if self.repository.unauthenticated == 'false':
-                log.warning(
-                    'KIWI does not support signature checks for apt-get '
-                    'package manager during the bootstrap procedure, any '
-                    'provided key will only be used inside the chroot '
-                    'environment'
-                )
-            cmd = ['debootstrap', '--no-check-gpg']
+            cmd = ['debootstrap']
+            if self.repository.unauthenticated == 'false' and \
+               os.path.exists(self.repository.keyring):
+                cmd.append('--keyring={}'.format(self.repository.keyring))
+            else:
+                cmd.append('--no-check-gpg')
             if self.deboostrap_minbase:
                 cmd.append('--variant=minbase')
             if self.repository.components:
@@ -161,10 +159,6 @@ class PackageManagerApt(PackageManagerBase):
                 options=['-a', '-H', '-X', '-A'],
                 exclude=['proc', 'sys', 'dev']
             )
-            for key in self.repository.signing_keys:
-                Command.run([
-                    'chroot', self.root_dir, 'apt-key', 'add', key
-                ], self.command_env)
         except Exception as e:
             raise KiwiDebootstrapError(
                 '%s: %s' % (type(e).__name__, format(e))
--- a/kiwi/repository/apt.py
+++ b/kiwi/repository/apt.py
@@ -16,6 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
+import logging
 from tempfile import NamedTemporaryFile
 from urllib.parse import urlparse
 
@@ -23,6 +24,9 @@ from urllib.parse import urlparse
 from kiwi.repository.template.apt import PackageManagerTemplateAptGet
 from kiwi.repository.base import RepositoryBase
 from kiwi.path import Path
+from kiwi.command import Command
+
+log = logging.getLogger('kiwi')
 
 
 class RepositoryApt(RepositoryBase):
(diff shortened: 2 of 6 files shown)


davidcassany:
@schaefi this PR fixes #1440 however it includes a couple of behavior changes that I'd like to discuss:

* The trusted keyring is created by using `gpg` command. This is a new dependency on the host for APT package manager if used with custom keys (`--signing-key` flag). I am not sure how should we treat that at packaging level. Probably adding gpg in `kiwi-packagemanager:apt` is enough. What do you think?

* The trusted keyring is not included into the resulting image, just used at buildtime, this differs from rpm based package manager (in rpm making this distinction in not obvious). Shall we reproduce a similar behavior? I'd say no, keep it simple and user can still include any custom keyring at config.sh.

schaefi:
@davidcassany  Thanks much for looking into this

>I am not sure how should we treat that at packaging level. Probably adding gpg in kiwi->packagemanager:apt is enough. What do you think? 

The code is now invoking gpg which I think is worth at least a recommends in the spec file like we do it for dnf and other packages that are used only on demand. Also gpg should be added to the box descriptions that are Debian based (the ubuntu box only atm)

>The trusted keyring is not included into the resulting image, just used at buildtime, this differs from rpm based package manager (in rpm making this distinction in not obvious). Shall we reproduce a similar behavior? I'd say no, keep it simple and user can still include any custom keyring at config.sh.

I agree let's keep this simple. Do you think it's worth to print a warning on the console that the gpg signature for the repos will not be permanent in the image ? What also came to my mind is; Do we have a chapter about signing-keys for repos and packages in the documentation ? If yes we should add a behavior comment in this case if not I think it's worth to have one (new issue)

davidcassany:
> The code is now invoking gpg which I think is worth at least a recommends in the spec file like we do it for dnf and other packages that are used only on demand.

Yes this sounds reasonable to me.

> I agree let's keep this simple. Do you think it's worth to print a warning on the console that the gpg signature for the repos will not be permanent in the image ?

Yes that might be a good idea, we used to have a warning to let users know keys were not being used in bootstrap phase, probably we should warn about the keyring not being included into the resulting image.

> What also came to my mind is; Do we have a chapter about signing-keys for repos and packages in the documentation ? If yes we should add a behavior comment in this case if not I think it's worth to have one (new issue)

I don't think we have a section for that. But this is an interesting topic, moreover I believe each package manager might behave slightly different.

davidcassany:
Ok, now I have added a warning to advise users the custom keyring is only present at buildtime and I also added the gpg tool as a recommended dependency.

schaefi:
@davidcassany  I think the commits can be merged, can't they ?

davidcassany:
> @davidcassany I think the commits can be merged, can't they ?

yes, I am merging

PR #1448, opened 2020-05-20
Refactor dracut disk re-partition library

### LINKED ISSUE #1406: Refactor disk resize
## Problem description

The current way to resize the disk is based on parted for machines with GPT and MSDOS tables. We put on parted because one tool could handle both table types. However using parted requires us to calculate start/end addresses ourselves and pass them along as parameters. Apart from the fact that this code in the initrd is very sensitive it should imho also be not the job of the resize code. I'd like to see the tooling to be able to validate size information properly which can be done with tools written for the dedicated partition tables, e.g gdisk for GPT and fdisk for msdos. The way the kiwi oem-repart module contains the partition size requests is already in form that allows easy consumption by those tools.

Also the distinguished partition tools have code to handle different types of storage disks that do not have a real C/H/S layout like it's the case for SSD's, SD cards, USB sticks and so on. The virtual C/H/S mapping is something parted does not deal well with.

For other architectures like s390 kiwi already runs into an extra thread that uses fdasd. Thus the concept to use a dedicated partitioner depending on the disk table type is already in place.

The meta format kiwi uses to describe the requested size information should not be changed. Only the tooling that validates that information and I expect the resulting code to be much smaller and more straight forward.

schaefi:
The kiwi repartition code for msdos and gpt tables was based
on the parted tool. However parted requires to give start and
end addresses to describe a space in the partition table. The
numbers were calculated in kiwi based on cylinders. The mapping
into cylinders is a virtual number that plays no role under
Linux and vanished completely with newer storage technologies.
Tools like sgdisk and sfdisk are better prepared to handle
disk partitioning and also makes this a more straight forward
code in kiwi. Thus this commit refactors the re-partition code
to drop parted and use a dedicated tool depending on the
partition table type. This Fixes #1406

### DIFF
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
 type lookup_disk_device_from_root >/dev/null 2>&1 || . /lib/kiwi-lib.sh
-type create_parted_partitions >/dev/null 2>&1 || . /lib/kiwi-partitions-lib.sh
+type create_partitions >/dev/null 2>&1 || . /lib/kiwi-partitions-lib.sh
 type resize_filesystem >/dev/null 2>&1 || . /lib/kiwi-filesystem-lib.sh
 type activate_volume_group >/dev/null 2>&1 || . /lib/kiwi-lvm-lib.sh
 type activate_mdraid >/dev/null 2>&1 || . /lib/kiwi-mdraid-lib.sh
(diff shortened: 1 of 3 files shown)


schaefi:
needless to say this is a very sensitive part of the oem code. I have tested gpt and msdos table resize and it worked for me so far.

schaefi:
>Not for this PR, but I was wondering while reviewing the code if it would make sense trying to >completely get rid of parted.

would be nice but looking at the kiwi code I see ```bootloader/config/zipl.py``` which reads a disk geometry based on CHS that really still exists on IBM mainframe

The other part is ./partitioner/msdos.py where we use parted to set the active flag. This can and should be moved to sfdisk like done here for the initrd code too

Are there other places you identified ?

davidcassany:
> Are there other places you identified ?

No, I did not find any further use of `parted` (the binary). I started checking dracut module dependencies and I realized that dracut-kiwi-live and dracut-kiwi-lib still require `parted` (the package). Then I noted that this is mostly because we use `partprobe`. However I could not find any `parted` (the binary) call in dracut modules. Then, in python code, I just found the call for setting some msdos flags and zipl ones as you mentioned.

schaefi:
ok let's move on with this one. I will merge this and create a new staging package. There are quite some changes in which should be tested by real boot tests.

PR #1452, opened 2020-05-22
Remmove any Recommends for CentOS7

davidcassany:
This commit fixes the package spec for CentOS 7. In CentOS 7 there is no
support for weak dependencies. In 9792cea1 a recommended dependency on
gpg tools was included for all builds and this caused a failure for
CentOS 7. With this commit, the recommended dependency, is omitted for
any pre CentOS 8 distro.


davidcassany:
@schaefi you can see a build of this change in https://build.opensuse.org/package/show/home:dcassany:Staging/python-kiwi

PR #1456, opened 2020-05-25
Fix scope for linuxefi grub config adaption

### LINKED ISSUE #1453: CentOS 7 UEFI images are not booting on UEFI
## Problem description

The OEM integration test image for CentOS does not boot on uefi. It faces the same issue we had in SLE about grub not being capable to determine wheather to use linuxefi or not at runtime.

Looks like the workaround introduced in a46c56a54632db3628a671247cece44f52bef5de doesn't do the job for CentOS 7. Would be interesting to test the same thing on CentOS 8, I haven't done it yet.

## Expected behaviour

The resulting image boots at least on UEFI regardless of the host configuration.

Use the OEM integration test image for CentOS and boot it as:

```
qmeu-kvm -m 2048 -hda ~/test.img -bios /usr/share/qemu/ovmf-x86_64-ms.bin
```

+ KIWI version: v9.20.16 (current master)
+ Operating system: 
+ OBS version:

schaefi:
In kiwi we support a one time patch for the grub config file
that changes the static use of linuxefi to be dynamic. In
grub2 >= 2.04 all this has already been fixed but for grub2
version older than this version we applied the patch. The
patch however was only applied based on the presence of a
grub setting named GRUB_USE_LINUXEFI. It has turned out that
this variable is a custom extension not part of grub upstream
which makes the test functional only on distributions that
supports this setting. The use of linuxefi however is code
that belongs to grub upstream. Therefore this patch changes
the scope of the one time patch to be only based on the
version of grub no matter if GRUB_USE_LINUXEFI is supported
by the distro or not. This Fixes #1453

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -145,7 +145,6 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         self.cmdline = None
         self.iso_boot = False
         self.shim_fallback_setup = False
-        self.validate_use_of_linuxefi = False
 
     def write(self):
         """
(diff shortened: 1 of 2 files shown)


schaefi:
Hope is last to die. I'll merge this and we can test the integration builds once a new staging package is out

PR #1458, opened 2020-05-25
Fixed dracut output file format detection

### LINKED ISSUE #1450: Initrd created by KIWI not in use for Debian/Ubuntu builds
## Problem description

Kiwi does not use the proper initrd name for Debian/Ubuntu distros. The resulting build ends with duplicate initrds, one created during packages installation by the distro tools and another one created by KIWI on later stages.

`grub-mkconfig` does not choose by default the initrd file created by KIWI. This is probably not obvious to note on simple builds as this affects only OEM and VMX images. For VMX is hard to realize, since KIWI, by default, does not perform any special configuration into the initrd, thus the default initrd created by the distribution tools during the `kernel` or `dracut` installation is mainly equivalent.

However in OEM images the modules for self expanding are KIWI custom modules and they are included only to the initrd created by KIWI. The problem is that the resulting root tree includes two initrds and gurb-mkconfig does not pick the appropriate one.

After an Ubuntu build in `/boot` there is:
```
-rw-------  1 root root  47M May 20 12:35 initramfs-5.4.0-31-generic.img
-rw-------  1 root root  56M May 20 12:34 initrd.img-5.4.0-31-generic
```

`initramfs-*` file is created by KIWI, the `initrd.im-*` is created by the distro tools. When running `grub-mkconfig` the one provided by the distro is used. Removing the `intird.img-*` file and calling again `grub-mkconfig` works fine.

Note that using installation media for OEM images works fine, as in that cases the firstboot is coming from templated grub configuration, thus for the first boot it uses the expected initrd.

## Expected behaviour

Grub is configured to use the initrd created by KIWI.

Create an OEM image without installation media and dd the result to a bigger disk. The image won't self expand during the first boot.

+ KIWI version: current master (9.20.12)
+ Operating system: not relevant
+ OBS version: not relevant

schaefi:
The current way to detect the dracut output file format was
based on a lookup of the format used in the dracut tool itself.
However there are distributions like Ubuntu which calls dracut
and passes the name of the initrd file as options to the call.
This invalidates the checking done by kiwi. The only chance
for kiwi to do the same than the distributions does is by
looking for an initrd file pre-created by the package
installations and use the same format. Only if no such initrd
file exists the former format detection code applies. The
additional code expects any initrd file to match the glob
pattern 'init*'. This Fixes #1450

### DIFF
--- a/kiwi/boot/image/base.py
+++ b/kiwi/boot/image/base.py
@@ -19,6 +19,7 @@ import re
 import os
 import pickle
 import logging
+import glob
 from collections import namedtuple
 
 # project
@@ -196,7 +197,9 @@ class BootImageBase:
                 'No kernel in boot image tree %s found' %
                 self.boot_root_directory
             )
-        dracut_output_format = self._get_boot_image_output_file_format()
+        dracut_output_format = self._get_boot_image_output_file_format(
+            kernel_info.version
+        )
         return boot_names_type(
             kernel_name=kernel_info.name,
             initrd_name=dracut_output_format.format(
@@ -380,15 +383,24 @@ class BootImageBase:
                     boot_description
             return boot_description
 
-    def _get_boot_image_output_file_format(self):
+    def _get_boot_image_output_file_format(self, kernel_version):
         """
         The initrd output file format varies between
         the different Linux distributions. Tools like lsinitrd, and also
         grub2 rely on the initrd output file to be in that format. Thus
         kiwi should use the same file format to stay compatible
-        with the distributions. The format is determined by the
-        outfile format used in the dracut initrd tool which is the
-        standard on all major linux distributions.
+        with the distributions. The format is determined in three
+        stages:
+
+        a) check for an existing initrd file and use this naming schema
+        b) if no initrd file is found check the dracut binary for its
+           default output file name schema
+        c) if no output file schema could be detected from the dracut
+           binary return a default output file format which is
+           initramfs-{kernel_version}.img
+
+        Let's hope all this mess can be deleted once all distros just
+        can agree on one initrd file name schema
         """
         if self.xml_state.get_initrd_system() == 'kiwi':
             # The custom kiwi initrd system is used only on SUSE systems.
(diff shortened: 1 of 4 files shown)


schaefi:
@davidcassany this is a start. I haven't done much tests with it so far. Could you also check if this would solve the initrd naming problem for Arch Linux ? I have looked up:

* suse/sle: initrd-<kernel_version>
* fedora/centos: initramfs-<kernel-version>.img
* ubuntu/debian: initrd.img-<kernel-version>

That we have to deal with this is really a pita

davidcassany:
> @davidcassany this is a start. I haven't done much tests with it so far. Could you also check if this would solve the initrd naming problem for Arch Linux ? I have looked up:
> 
>     * suse/sle: initrd-<kernel_version>
> 
>     * fedora/centos: initramfs-<kernel_version>.img
> 
>     * ubuntu/debian: initrd.img-<kernel_version>

I can do some tests in Arch Linux, however, in Arch Linux, the naming convention is different. In there the naming is:

* vmlinuz-<kernel_package_name> which turns to `vmlinuz-linux`
* initramfs-<kernel_package_name>.img which turns to `initramfs-linux.img`

This is the schema for the pre-dracut tool `mkinitcpio`. For dracut it still does not have official hooks, even in the Arch Linux dracut [documentation](https://wiki.archlinux.org/index.php/Dracut) the example follows this same schema.

This is not raising in my image examples as the hooks for creating the initrd are provided by a package of mine.

In any case, at a glance, I believe this solves half of the story for Arch Linux. At least the initramfs part.


For the kernel version I already coded a patch some time ago that could conclude the whole story. 
> 
> That we have to deal with this is really a pita

Absolutely agree...

So, from my PoV I'd like to run some tests with this. If it all goes green I'd say it is good to merge it. Regarding the Arch Linux I might do another PR for the kernel detection, however probably it is better to wait until there is an 'official' naming criteria for builds using dracut.

davidcassany:
I tested this on Ubuntu Focal with success :+1:

schaefi:
Great, thanks for testing and I'm pro what you have said about Arch :+1:

PR #1460, opened 2020-05-26
Fix the patch applied on grub.cfg

davidcassany:
This commit fixes the patch applied on grub.cfg when EFI mode is
selected and grub < 2.04. There are some distros that make use of the
`linux16` command instead of `linux` in grub configuration, this commit
extends the regex to also consider `linux16` command.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -261,11 +261,11 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 with open(config_file) as grub_config_file:
                     grub_config = grub_config_file.read()
                     grub_config = re.sub(
-                        r'([ \t]+)linux[efi]*([ \t]+)', r'\1$linux\2',
+                        r'([ \t]+)linux(efi|16)*([ \t]+)', r'\1$linux\3',
                         grub_config
                     )
                     grub_config = re.sub(
-                        r'([ \t]+)initrd[efi]*([ \t]+)', r'\1$initrd\2',
+                        r'([ \t]+)initrd(efi|16)*([ \t]+)', r'\1$initrd\3',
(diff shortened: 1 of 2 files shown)


kiwi/bootloader/config/grub2.py
```diff
@@ -261,11 +261,11 @@ def setup_disk_image_config(
                 with open(config_file) as grub_config_file:
                     grub_config = grub_config_file.read()
                     grub_config = re.sub(
-                        r'([ \t]+)linux[efi]*([ \t]+)', r'\1$linux\2',
+                        r'([ \t]+)linux[efi16]*([ \t]+)', r'\1$linux\2',
                         grub_config
                     )
                     grub_config = re.sub(
-                        r'([ \t]+)initrd[efi]*([ \t]+)', r'\1$initrd\2',
+                        r'([ \t]+)initrd[efi16]*([ \t]+)', r'\1$initrd\2',
```
schaefi, line 268: Maybe we should think of a regexp that really only matches 'linux, linuxefi and linux16'. My former regexp was already not the best because [efi] is a character class that matches any combination of those. Now the class is containing even more characters and would match more names without sense.

I think it will not hurt but it's not the best way of using the expression.

Thoughts ?

davidcassany: Yes I can try to come up with  a more strict regexp. I also noted the regex is turning less restrictive. Let me see if I can come to something more strict.

schaefi: great :+1:

davidcassany: Now it only allows `linux`, `linuxefi` and `linux16`.

schaefi: perfect, thanks

PR #1461, opened 2020-05-26
Added custom partition image integration test

schaefi:
There are still people who like many extra static inflexible
entries in a partition table instead of using LVM or filesystem
volume capabilities for sometimes good but often questionable
reasons. In kiwi we intentionally support partition table
entries on a restricted basis but that does not mean you can't
create an image with a highly customized partition table.
However it includes some bits of custom code as part of the
image description and that's what this reference implementation
of an image with custom partitions demonstrates. The image
described here adds three extra partitions, var, var/log and
var/audit. The concept to create custom partitions is based on
specifying a spare partition which is by default created before
the root partition. The spare space can now be sliced into as
many partitions as needed and that needs to be implemented by
the author of the image description. Of course the partition
table itself comes with limitations which has to be respected
depending on the partition table type.

### DIFF
--- a/build-tests/x86/suse/test-image-custom-partitions/appliance.kiwi
+++ b/build-tests/x86/suse/test-image-custom-partitions/appliance.kiwi
@@ -1,0 +1,74 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<image schemaversion="6.8" name="LimeJeOS-TW">
+    <description type="system">
+        <author>Marcus Schaefer</author>
+        <contact>ms@suse.de</contact>
+        <specification>
+            TW JeOS, is a small text based image
+        </specification>
+    </description>
+    <preferences>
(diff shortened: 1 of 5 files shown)


schaefi:
I wanted to come up with an image that implements the above just to have a reference for all those who think managing partitions in the static table is a good idea

Even though I'm 99% convinced this is not a good thing I'd like to give people a chance to go for what they want. I think we have explained relatively often why in the appliance world with flexible storage backends and ready to resize filesystems a static table always stands in your way but we didn't came across to provide users with an image description to demonstrate how they can do it anyway.

Parts of the description are not nice but that's the footprint of the limited table handling we support in kiwi intentionally

The integration test builds here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:suse/test-image-custom-partitions

PR #1462, opened 2020-05-26
Fix the boot path for initrd detection

davidcassany:
This commit fixes the path to check the initrd naming convetion from an
already existing initrd file in /boot.

### DIFF
--- a/kiwi/boot/image/base.py
+++ b/kiwi/boot/image/base.py
@@ -446,7 +446,7 @@ class BootImageBase:
     def _get_boot_image_output_file_format_from_existing_file(
         self, kernel_version
     ):
-        for initrd_file in glob.iglob(self.boot_root_directory + '/init*'):
+        for initrd_file in glob.iglob(self.boot_root_directory + '/boot/init*'):
             if not os.path.islink(initrd_file):
                 return os.path.basename(initrd_file).replace(
                     kernel_version, '{kernel_version}'


kiwi/boot/image/base.py
```diff
@@ -446,7 +446,7 @@ def _get_boot_image_output_file_format_from_dracut_code(self):
     def _get_boot_image_output_file_format_from_existing_file(
         self, kernel_version
     ):
-        for initrd_file in glob.iglob(self.boot_root_directory + '/init*'):
+        for initrd_file in glob.iglob(self.boot_root_directory + '/boot/init*'):
```
davidcassany, line 449: I am wondering if there is any use case in which this might not be the appropriate path. In the Disk tests I did this was correct.

davidcassany:
@schaefi sorry for this late amend on the previous PR. I did something wrong in my previous test that I had the impression it worked fine, but it didn't. I missed this detail on my previous revision, sorry.

kiwi/boot/image/base.py
schaefi: The reason why there is an extra self.boot_root_directory is because of the support for custom kiwi initrds. If you look at kiwi/boot/image/builtin_kiwi.py you see the difference in construction. However I think your change is correct in any way. self.boot_root_directory is always pointing to a linux root (/)

schaefi:
> @schaefi sorry for this late amend on the previous PR.

No problem at all. Actually it was me who coded it wrong. My only excuse is that I said that I haven't tested it so far :)

It's only Staging so nothing problematic happened

I'll create a new staging submission. I think we found most of the booting issues. How great is this :) Thanks for your effort on fixing this one

PR #1463, opened 2020-05-27
Add EFI firmware on OEM and ISO ubuntu tests

davidcassany:
This commits adds efi firmware for Ubuntu integration tests

### DIFF
--- a/build-tests/x86/ubuntu/test-image-iso-oem-vmx/appliance.kiwi
+++ b/build-tests/x86/ubuntu/test-image-iso-oem-vmx/appliance.kiwi
@@ -25,7 +25,7 @@
         <rpm-check-signatures>false</rpm-check-signatures>
     </preferences>
     <preferences profiles="Live">
-        <type image="iso" flags="overlay" hybridpersistent_filesystem="ext4" hybridpersistent="true"/>
+        <type image="iso" flags="overlay" hybridpersistent_filesystem="ext4" hybridpersistent="true" firmware="efi"/>
     </preferences>
     <preferences profiles="Virtual">
         <type image="vmx" filesystem="ext4" kernelcmdline="console=ttyS0 splash">
@@ -33,7 +33,7 @@
         </type>
     </preferences>
     <preferences profiles="Disk">
-        <type image="oem" filesystem="ext4" initrd_system="dracut" installiso="true">
+        <type image="oem" filesystem="ext4" initrd_system="dracut" installiso="true" firmware="efi">
             <bootloader name="grub2"/>
             <oemconfig>
                 <oem-swap>true</oem-swap>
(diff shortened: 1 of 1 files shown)


build-tests/x86/ubuntu/test-image-iso-oem-vmx/appliance.kiwi
```diff
@@ -74,6 +77,7 @@
         <package name="isc-dhcp-client"/>
         <package name="netbase"/>
         <package name="dbus"/>
+        <package name="btrfs-progs"/>
```
davidcassany, line 80: This is required for `dracut-kiwi-lib`. In our spec it is a recommended dependency for Ubuntu, but not a hard requirement. I can't recall why we did this way, I remember we had several issues in the past for Ubuntu, thus probably adding the hard requirement was an issue...not sure, I did not dig into that.

davidcassany:
Just validated that OBS does not resolve with the current packages list

davidcassany:
Ok I have the feeling we already tried that in the past and failed because of the same grub2 packaging issues... grub2 dependencies in Debian/Ubuntu seam to be inconsistent under some contexts...

schaefi:
> Ok I have the feeling we already tried that in the past and failed because of the same grub2 packaging issues... grub2 dependencies in Debian/Ubuntu seam to be inconsistent under some contexts...

yes I remember. when using efi this required the grub2 efi modules to be used. The package in the ubuntu space that provides those lead to solver problems when this is done in obs. I remember we gave up and postponed it. Seem like it's still an issue

davidcassany:
> yes I remember. when using efi this required the grub2 efi modules to be used. The package in the ubuntu space that provides those lead to solver problems when this is done in obs. I remember we gave up and postponed it. Seem like it's still an issue

Yes that is the case I think. In fact this is still an issue for local builds (at least for Debian), installing grub2 pulls in grub-pc which is incompatible with grub-efi-amd64, however grub-efi-amd64 can obsolete grub-pc, meaning installing grub-efi-amd64 on top of grub-pc works, but they can't be installed together at the same time (that's why we have grub2 at bootstrap for Debian). This mess breaks on OBS solver... probably I'll do a couple of new tests on Focal (the latest ubuntu version) to see if this improved and if not just abandon the idea.

davidcassany:
OK after reading the insane control file for grub2 I managed to come with a subset of package that actually contain the real binaries and that they do not conflict with each other. See the successful tests in https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:ubuntu/test-image-iso-oem-vmx

I did a couple of boots (vmx and oem) and so far it works. Once merged the `_service` file needs to be adapted to pull again from `master` branch.

Issue #1464, opened 2020-05-28
Proposal: bash to sh?


isbm:
If one builds a system with a busybox, then Bourne Shell isn't there, but Almquist Shell. How is the idea to substitute hardcoded `bash` all over the place with `sh` and fix the inconsistencies, if any? ("works on my laptop" TM tho, at the moment). For SUSE world, `/usr/bin/sh` is anyway a symlink to `/bin/bash`, in Debian, howerver, to `dash` (Debian Almquist Shell).

I am not sure this needs to be explicitly configurable, as the mere purpose is to simply run commands. But it gets problematic if coreutils aren't there.

Mostly to @schaefi :wink:

schaefi:
I think the invocations in kiwi we do with bash can also be done with any other shell. This would cover the following parts in code:

* ./solver/repository/base.py
* ./storage/disk.py
* ./package_manager/pacman.py
* ./package_manager/dnf.py
* ./iso_tools/base.py
* ./system/setup.py
* ./system/size.py
* ./utils/compress.py
* ./archive/cpio.py
* ./archive/tar.py
* ./bootloader/config/grub2.py
* ./bootloader/config/zipl.py
* ./partitioner/msdos.py
* ./container/oci.py
* ./defaults.py

There are the following places where I'm concerned if changes will have some negative side effects:

* ./config/functions.sh
   helper functions for shell scripts and it is assumed they use bash

* ./system/shell.py
   call out a script using bash

* ./system/profile.py
   read and export variable settings, not sure if that's handled the same for all shells. I know e.g tcsh does strange things with variables

Overall I'm pro to be not bash specific but I'd like to avoid introducing bugs. From your environment did you have identified where things fails ? if so we can start with those and go in small steps

Thoughts ?

schaefi:
Another area for the shell is dracut. All dracut modules are written using bash. If you look through /usr/lib/dracut/modules.d you see almost all modules to run with "#!/bin/bash". If your environment runs without bash it will be hard to create a dracut initrd because even dracut itself runs with "#!/bin/bash --norc". But all of that can be adapted on other distributions

isbm:
Actually! With [this PR](https://github.com/OSInside/kiwi/pull/1465) I could successfully kill Bash with Busybox and still have dracut & co (you don't need dracut at runtime anyway). So maybe this Issue we can actually close, as long as https://github.com/OSInside/kiwi/pull/1465 is in (after final polishing, rgiht now it is just WiP).

Issue #915, opened 2019-01-16
Add ability to produce containers without zypper/rpm in them


Vogtinator:
It should be possible to create container images without any package manager inside. Currently that's not easily possible without listing all packages manually in the delete section.

davidcassany:
Have you tried `uninstall` instead of `delete`? I believe this should be what are you looking for.

Consider something like

```xml
<packages type="uninstall">
  <package name="zypper"/>
</package>
```

This will produce a call of `chroot root_dir zypper rm -u --force-resolution zypper` after running `config.sh`. Isn't it sufficient?

Vogtinator:
That's what I tried, but it does not work:

```
[   77s] [ DEBUG   ]: 13:10:52 | system: ( 1/82) Removing zypper-1.14.19-1.1.x86_64 [.....done]
[   77s] [ DEBUG   ]: 13:10:52 | system: ( 2/82) Removing procps-3.3.15-1.2.x86_64 [.....done]
[   77s] [ DEBUG   ]: 13:10:52 | system: ( 3/82) Removing libzypp-17.10.3-1.1.x86_64 [.....done]
[   77s] [ DEBUG   ]: 13:10:52 | system: ( 4/82) Removing libaugeas0-1.11.0-1.1.x86_64 [.....done]
[   77s] [ DEBUG   ]: 13:10:52 | system: ( 5/82) Removing libprocps7-3.3.15-1.2.x86_64 [.....done]
[   77s] [ DEBUG   ]: 13:10:52 | system: ( 6/82) Removing libsolv-tools-0.7.2-1.1.x86_64 [.....done]
[   77s] [ DEBUG   ]: 13:10:52 | system: ( 7/82) Removing libproxy1-0.4.15-5.1.x86_64 [.....done]
[   77s] [ DEBUG   ]: 13:10:52 | system: ( 8/82) Removing libgpgme11-1.12.0-2.1.x86_64 [.....done]
[   77s] [ DEBUG   ]: 13:10:52 | system: ( 9/82) Removing libcurl4-7.63.0-1.2.x86_64 [.....done]
[   77s] [ DEBUG   ]: 13:10:52 | system: (10/82) Removing libboost_thread1_68_0-1.68.0-3.2.x86_64 [.....done]
[   77s] [ DEBUG   ]: 13:10:52 | system: (11/82) Removing libsystemd0-239-3.1.x86_64 [.....done]
[...]

PR #1465, opened 2020-05-28
Post-build config concept

isbm:
This is a concept of post-build config. The basic idea is that it is as same as `config.sh`, except it is called after everything is already on the image. Mostly this is useful for the embedded. Currently this is working on "root" only, so I think should be expanded to something like `bootloader-post-build.sh`, `root-post-build.sh` etc.

Ideas/suggestions/criticism encouraged.

*Upd*: It also helps #1464

P.S. I also started slowly adding annotations, because it is easier then to work with. It also would be nice if you, guys, decide on a standard code formatter (linting), because it is pain to work with modern Python editors, those now tends to enforce code formatting in Go fashion and then [this happens](https://github.com/OSInside/kiwi/pull/1465/commits/e2759c5471a466778eb4991ceedc313cb45489df).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,7 +24,7 @@ __pycache__/
 *~
 *.kate-swp
 *.~
-
+*vscode/
 
 # C extensions
 *.so
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+*mypy_cache/
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
--- a/kiwi/bootloader/install/grub2.py
+++ b/kiwi/bootloader/install/grub2.py
@@ -39,6 +39,7 @@ class BootLoaderInstallGrub2(BootLoaderInstallBase):
     """
     **grub2 bootloader installation**
     """
+
     def post_init(self, custom_args):
         """
         grub2 post initialization method
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -19,7 +19,7 @@ import os
 import logging
 import pickle
 from tempfile import NamedTemporaryFile
-from typing import Dict, Union
+from typing import Dict
 
 # project
 from kiwi.defaults import Defaults
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -67,6 +67,7 @@ class DiskBuilder:
         * signing_keys: list of package signing keys
         * xz_options: string of XZ compression parameters
     """
+
     def __init__(self, xml_state, target_dir, root_dir, custom_args=None):
         self.arch = Defaults.get_platform_name()
         self.root_dir = root_dir
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -19,6 +19,7 @@ import os
 import logging
 import pickle
 from tempfile import NamedTemporaryFile
+from typing import Dict, Union
 
 # project
 from kiwi.defaults import Defaults
@@ -46,6 +47,8 @@ from kiwi.utils.fstab import Fstab
 from kiwi.path import Path
 from kiwi.runtime_config import RuntimeConfig
 from kiwi.partitioner import Partitioner
+from kiwi.storage.mapped_device import MappedDevice
+from kiwi.mount_manager import MountManager
 
 from kiwi.exceptions import (
     KiwiDiskBootImageError,
@@ -466,6 +469,9 @@ class DiskBuilder:
         # set root filesystem properties
         self._setup_property_root_is_readonly_snapshot()
 
+        # This is same as config.sh, just after everything has been created on the disk.
+        self._call_post_build_config(device_map)
+
         # prepare for install media if requested
         if self.install_media:
             log.info('Saving boot image instance to file')
@@ -509,6 +515,32 @@ class DiskBuilder:
         )
 
         return self.result
+
+    def _call_post_build_config(self, device_map: Dict[str, MappedDevice]):
+        """
+        Call post-build.sh config script. This is the same as config.sh, except
+        it is run on already created rootfs for final tuning/adjusting.
+
+        This method is mostly meant for embedded Linux images, where some packages
+        and other adjustments needs to be explicitly removed or any other things
+        has to be done directly.
+        """
+
+        if not self.system_setup.has_post_build_script():
+            return
+
+        if "root" not in device_map:
+            log.error("Need to call post-build.sh script, but unable to find root device!")
+            raise KiwiDiskBootImageError("Root device was not found")
+
+        mnt_point = MountManager(device=device_map["root"].get_device())
+        mnt_point.mount()
+
+        log.info("Mounted root-fs for post-build config on %s at %s", mnt_point.device, mnt_point.mountpoint)
+        self.system_setup.call_post_build_config(root_dir=mnt_point.mountpoint)
+
+        mnt_point.umount()
+        log.info("Umounted root-fs for post-build config")
 
     def create_disk_format(self, result_instance):
         """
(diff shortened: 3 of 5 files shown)


schaefi:
Hey thanks for all your effort. Before I share my view on the suggested proposal let me comment
on the following:

>P.S. I also started slowly adding annotations, because it is easier then to work with. It also would be nice if you, guys, decide on a standard code formatter (linting), because it is pain to work with modern Python editors, those now tends to enforce code formatting in Go fashion

I think we are following the standard of the flake8 linter which is also running on every PR. See testenv:check in tox.ini. From my perspective this a good standard for python projects. Do you think this is not the case ?

If editors like code enforce something different this should be imho solved on the level of the editor setup and not by changing the code here to make the editor happy. Especially the rule enforcement of go applied to python is in my opinion questionable. The extra newlines as
well as the change of 80c lines to long lines in the scope of this PR makes it harder to review the code. If you don't mind if would be good to turn that back such that we can focus on the actual changes. Would that be possible ?

I admit the 80c lines is something I personally like because it makes reading code easy and it prevents over complicated code on one liners. I know there are people who see that differently 
and I respect that which is why we allow long lines, see setup.cfg "ignore = E501" from the linter. However the code I wrote in this project follows 80c lines and it would be nice if people respect that part when I also respect lone lines from other peoples contributions.

Overall I think the review of your idea will be easier without the linter changes as a start

schaefi:
Regarding the idea of a script that runs after the system gets transfered into the actual image loop disk. I like that idea and there are also other people who had this use case in the past. Two things on the current implementation which I'd like to discuss is

a) you said this script is the same than config.sh just called at another place in the process. This is only partial correct because config.sh is called in any case for all image types at the end of the creation of the root data. Your script is specific to the disk builder and also runs as part of the loop disk data. This means the modification the script hook does are not permanent to the image root data. This is not a problem but the script hook should indicate that clearly by its name and we also need to have that documented

b) the implementation of the script hook actively mounts the rootfs, this should not be done because the full mount of the system is not known to you at this point. A system can have an extra boot partition, it could have lvm volumes, it could have btrfs sub-volumes, it could be luks layered it could be raid layered etc etc... So what mounting of the system means should not be an action you do prior calling the script. If you look at the disk builder code you will find ```sync_data(``` calls. At this point the system is in the mount state you are after and right before the bootloader gets installed you can include the script call as chrooted operation in whatever device_map['root'] defines.

Does this makes sense ?

schaefi:
I also looked at the code that syncs the root data to the disk. There is a little inconsistency between the FileSystem class and the VolumeManager class. The first populates self.filesystem_mount
whereas the last has self.mountpoint. I suggest a little refactoring PR and will look into it

isbm:
Due to a) -- yes, absolutely. With b) -- that's why a big thanks for clarifying! Yes, I wasn't that sure exactly on that part, and that's currently more a try/hack than something final (hence I don't care ATM of broken tests). Also user would like _maybe_ also to do something else with other partitions, so currently I was experimenting with root only and hammering only active mount. Originally I thought that the system is mounted, but then I found it is not. Yes, I was looking exactly that `sync_data` part, I will try look into that in a minute and get back here.

schaefi:
> Yes, I was looking exactly that sync_data part,

perfect. we need a little refactoring to make that usable for your use case. Let me code that and I will send a PR. Also @davidcassany feedback is very much welcome

isbm:
> inconsistency between the FileSystem class and the VolumeManager class.

We will make it perfect. Let's do one step at a time: first proper concept, then proper interface with the usage, then polish/fix tests. Sounds like a plan?

P.S. Please think of MyPy in tox and annotations. I am using it all the time and it does help a lot (and this code needs it).

schaefi:
yep good plan

isbm:
@schaefi hm, now I remember. I actually been there already and it doesn't work (hence the mentioned Issue I've opened): dracut, bash etc, now I have that problem back. The very idea of the post-build is to do whatever needed after all other tooling done (dracut including). And that's quite important.

Any other ways to mitigate that? I am even OK to go after different types of remounts, if I need to, but generally I am not just after mounted system, but after "finished state of building before final image bake" state.

isbm:
> A system can have an extra boot partition, it could have lvm volumes, it could have btrfs sub-volumes, it could be luks layered it could be raid layered etc etc...

I was thinking to get the constraints first:
- Embedded and Btrfs do not match (rare exceptions, but you want to avoid that, if possible). So I would rather less worry about this part at this point.
- This is "advanced mode" and mostly manual at this point with intention "you know what you are doing". The very purpose of it is to undo what distribution did (e.g. remove zypper, libsolv, cracklib and other packages you never need in embedded, replace coreutils etc etc). So it is more like "shaping up" OS image with a final touch.
- Its main purpose for the finished rootfs alteration, something that normally nobody does it in a regular appliance mode.

So maybe we can go with active remounts after all and then just restrict it to specific file systems, reading the layout from the `config.xml` and/or collect that data before? I would be willing helping even to extend the "mount awareness" so the remount is correct then.

Ideas?

davidcassany:
The idea of having some hook that is chrooted into the disk looks reasonable to me. At the moment I fail to see if this has some potential for being used in other environments beyond disks (filesystem images, container trees, etc.), but I doubt it. Thus as it has been already said we should make explicit the scope of this feature.

Feature wise my only concern is the proliferation of hooks were the user is getting more chances to miss use it. I see we have `config.sh`, `images.sh`, boot scripts and now probably something like `disk-config.sh`. I'd say this is still maintainable, but still quite a few.

> Embedded and Btrfs do not match (rare exceptions, but you want to avoid that, if possible). So I would rather less worry about this part at this point.

Well, not so sure, probably not for the embedded plus btrfs tandem, but we should consider complex configurations as it is done for MicroOS images.

> The very purpose of it is to undo what distribution did (e.g. remove zypper, libsolv, cracklib and other packages you never need in embedded, replace coreutils etc etc). 

Note that we have this other opened issue to this sort of procedures #915 and our major issue in there is precisely that removing zypper and rpm stack can't be done easily done from a chrooted env. 

> I would be willing helping even to extend the "mount awareness" so the remount is correct then.

I like the idea of thinking of utilities to properly mount the disk device to chroot in it. I haven't had a close look though not sure right now how far we are or how difficult it is, I'd also like to avoid major refactors.

@isbm out of curiosity, which is the use case to remove busybox, coreutils and staff like that? Is it just a matter of space saving? Just trying to understand the underlying needs

isbm:
> out of curiosity, which is the use case to remove busybox

Ah. No. "Replace coreutils with busybox" -- sorry! (corrected comment too).

schaefi:
#1469 offers a way to run a disk.sh script from your image description inside of the fully loop mounted disk. I think this should serve the purpose regarding the actions you need. If the PR is acceptable I can submit a package to https://build.opensuse.org/project/show/Virtualization:Appliances:Staging from you can give it a try as well.

isbm:
I am not sure how that supposed to work:

1. Create `disk.sh` inside description.
2. Call build:

```
[ ERROR   ]: 13:43:31 | KiwiScriptFailed: disk.sh failed: chroot: failed to run command ‘/image/disk.sh’: No such file or directory
```

If I delete that script, no problems then. :wink:

schaefi:
Hmm, it worked in my test but I can reproduce the error you got too in another test. Still investigating...

schaefi:
found it :) btrfs related, the get_mountpoint method must be handled correctly for the btrfs volume manager like it's done for sync_data in this case. I missed that part. Will fix it in a minute

schaefi:
@isbm ok fixed now. Please find packages in https://build.opensuse.org/project/show/Virtualization:Appliances:Staging and also see the integration test for this running here: https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:suse/test-image-oem. It should turn to green as soon as the new package is done.

Thanks for testing

schaefi:
I think this one can be closed on behalf of #1469 

if not feel free to re-open

PR #1469, opened 2020-06-03
Disk hook

### LINKED ISSUE #1464: Proposal: bash to sh?

schaefi:
Added new post disk sync script hook
    
Allow to put a disk.sh script as part of the image description which is called for the disk image types `vmx` and `oem` only and runs after the synchronisation of the root tree into the  disk image loop file. At call time of the script the device name  of the currently mapped root device is passed as a parameter. The chroot environment for the call is the mounted disk itself  which makes this different from the config.sh/images.sh caller  environment. This Fixes #1464

Update user defined scripts documentation
    
Added information about new disk.sh script and reworked the entire chapter

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -16,19 +16,30 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
+import sys
 import glob
 from collections import namedtuple
 import platform
 from pkg_resources import resource_filename
 
 # project
-from .path import Path
-from .version import (
+from kiwi.path import Path
+from kiwi.version import (
     __githash__,
     __version__
 )
 
-from .exceptions import KiwiBootLoaderGrubDataError
+from kiwi.exceptions import KiwiBootLoaderGrubDataError
+
+# Default module variables
+this = sys.modules[__name__]
+
+this.post_disk_sync_script_name = 'disk.sh'
+this.post_prepare_script_name = 'config.sh'
+this.pre_create_script_name = 'images.sh'
+this.edit_boot_config_script_name = 'edit_boot_config.sh'
+this.edit_boot_install_script_name = 'edit_boot_install.sh'
+this.image_metadata_directory = 'image'
 
 
 class Defaults:
--- a/kiwi/system/profile.py
+++ b/kiwi/system/profile.py
@@ -53,6 +53,17 @@ class Profile:
         self._drivers_to_profile()
         self._root_partition_uuid_to_profile()
 
+    def get_settings(self):
+        """
+        Return all profile elements that has a value
+        """
+        profile = {}
+        for key, value in list(self.dot_profile.items()):
+            profile[key] = Shell.format_to_variable_value(value)
+        return collections.OrderedDict(
+            sorted(profile.items())
+        )
+
     def add(self, key, value):
         """
         Add key/value pair to profile dictionary
@@ -72,16 +83,13 @@ class Profile:
 
         :param str filename: file path name
         """
-        sorted_profile = collections.OrderedDict(
-            sorted(self.dot_profile.items())
-        )
+        sorted_profile = self.get_settings()
         temp_profile = NamedTemporaryFile()
         with open(temp_profile.name, 'w') as profile:
             for key, value in list(sorted_profile.items()):
-                if value:
-                    profile.write(
-                        format(key) + '=' + self._format(value) + '\n'
-                    )
+                profile.write(
+                    '{0}={1}{2}'.format(key, value, os.linesep)
+                )
         profile_environment = Shell.quote_key_value_file(temp_profile.name)
         with open(filename, 'w') as profile:
             for line in profile_environment:
@@ -382,13 +390,3 @@ class Profile:
                 return 'true'
             else:
                 return content
-
-    def _format(self, value):
-        """
-        Helper method to format bool profile values in the way
-        the boot code expects them
-        """
-        if value is True:
-            return 'true'
-        else:
-            return format(value)
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -18,11 +18,14 @@
 import glob
 import os
 import logging
+import copy
 from collections import OrderedDict
 from collections import namedtuple
 from tempfile import NamedTemporaryFile
 
 # project
+import kiwi.defaults as defaults
+
 from kiwi.mount_manager import MountManager
 from kiwi.system.uri import Uri
 from kiwi.repository import Repository
@@ -39,6 +42,7 @@ from kiwi.archive.tar import ArchiveTar
 from kiwi.utils.compress import Compress
 from kiwi.utils.command_capabilities import CommandCapabilities
 from kiwi.utils.rpm_database import RpmDataBase
+from kiwi.system.profile import Profile
 
 from kiwi.exceptions import (
     KiwiImportDescriptionError,
@@ -84,9 +88,15 @@ class SystemSetup:
         script helper methods
         """
         log.info('Importing Image description to system tree')
-        description = self.root_dir + '/image/config.xml'
-        log.info('--> Importing state XML description as image/config.xml')
-        Path.create(self.root_dir + '/image')
+        description = os.path.join(
+            self.root_dir, defaults.image_metadata_directory, 'config.xml'
+        )
+        log.info(
+            '--> Importing state XML description to {0}'.format(description)
+        )
+        Path.create(
+            os.path.join(self.root_dir, defaults.image_metadata_directory)
+        )
         with open(description, 'w', encoding='utf-8') as config:
             config.write('<?xml version="1.0" encoding="utf-8"?>')
             self.xml_state.xml_data.export(outfile=config, level=0)
@@ -95,12 +105,23 @@ class SystemSetup:
         self._import_custom_archives()
         self._import_cdroot_archive()
 
+    def script_exists(self, name):
+        """
+        Check if provided script base name exists in the image description
+        """
+        return os.path.exists(os.path.join(self.description_dir, name))
+
     def cleanup(self):
         """
         Delete all traces of a kiwi description which are not
         required in the later image
         """
-        Command.run(['rm', '-r', '-f', '/.kconfig', '/image'])
+        Command.run(
+            [
+                'chroot', self.root_dir,
+                'rm', '-rf', '.kconfig', defaults.image_metadata_directory
+            ]
+        )
 
     def import_repositories_marked_as_imageinclude(self):
         """
@@ -135,10 +156,10 @@ class SystemSetup:
             )
             if not repo_alias:
                 repo_alias = uri.alias()
-            log.info('Setting up image repository %s', repo_source)
-            log.info('--> Type: %s', repo_type)
-            log.info('--> Translated: %s', repo_source_translated)
-            log.info('--> Alias: %s', repo_alias)
+            log.info('Setting up image repository {0}'.format(repo_source))
+            log.info('--> Type: {0}'.format(repo_type))
+            log.info('--> Translated: {0}'.format(repo_source_translated))
+            log.info('--> Alias: {0}'.format(repo_alias))
             repo.add_repo(
                 repo_alias, repo_source_translated,
                 repo_type, repo_priority, repo_dist, repo_components,
@@ -201,8 +222,8 @@ class SystemSetup:
             archive.extract(self.root_dir)
 
         for profile in self.xml_state.profiles:
-            overlay_directory = os.sep.join(
-                [self.description_dir, profile]
+            overlay_directory = os.path.join(
+                self.description_dir, profile
             ) + os.sep
             if os.path.exists(overlay_directory):
                 self._sync_overlay_files(
@@ -227,8 +248,8 @@ class SystemSetup:
         Deleting the machine-id without the dracut initrd
         creating a new one produces an inconsistent system
         """
-        machine_id = os.sep.join(
-            [self.root_dir, 'etc', 'machine-id']
+        machine_id = os.path.join(
+            self.root_dir, 'etc', 'machine-id'
         )
 
         if os.path.exists(machine_id):
@@ -270,7 +291,7 @@ class SystemSetup:
         """
         if 'keytable' in self.preferences:
             log.info(
-                'Setting up keytable: %s', self.preferences['keytable']
+                'Setting up keytable: {0}'.format(self.preferences['keytable'])
             )
             if CommandCapabilities.has_option_in_help(
                 'systemd-firstboot', '--keymap',
@@ -305,7 +326,9 @@ class SystemSetup:
                 locale = '{0}.UTF-8'.format(
                     self.preferences['locale'].split(',')[0]
                 )
-            log.info('Setting up locale: %s', self.preferences['locale'])
+            log.info(
+                'Setting up locale: {0}'.format(self.preferences['locale'])
+            )
             if CommandCapabilities.has_option_in_help(
                 'systemd-firstboot', '--locale',
                 root=self.root_dir, raise_on_error=False
@@ -329,7 +352,7 @@ class SystemSetup:
         """
         if 'timezone' in self.preferences:
             log.info(
-                'Setting up timezone: %s', self.preferences['timezone']
+                'Setting up timezone: {0}'.format(self.preferences['timezone'])
             )
             if CommandCapabilities.has_option_in_help(
                 'systemd-firstboot', '--timezone',
@@ -356,7 +379,7 @@ class SystemSetup:
         for user in self.xml_state.get_users():
             for group in self.xml_state.get_user_groups(user.get_name()):
                 if not system_users.group_exists(group):
-                    log.info('Adding group %s', group)
+                    log.info('Adding group {0}'.format(group))
                     system_users.group_add(group, [])
 
     def setup_users(self):
@@ -366,7 +389,7 @@ class SystemSetup:
         system_users = Users(self.root_dir)
 
         for user in self.xml_state.get_users():
-            log.info('Setting up user %s', user.get_name())
+            log.info('Setting up user {0}'.format(user.get_name()))
             password = user.get_password()
             password_format = user.get_pwdformat()
             home_path = user.get_home()
@@ -399,7 +422,7 @@ class SystemSetup:
                 system_users.user_add(user_name, options)
                 if home_path:
                     log.info(
-                        '--> Setting permissions for %s', home_path
+                        '--> Setting permissions for {0}'.format(home_path)
                     )
                     # Emtpy group string assumes the login or default group
                     system_users.setup_home_for_user(
@@ -434,9 +457,13 @@ class SystemSetup:
         image_id = self.xml_state.xml_data.get_id()
         if image_id and os.path.exists(self.root_dir + '/etc'):
             image_id_file = self.root_dir + '/etc/ImageID'
-            log.info('Creating identifier: %s as %s', image_id, image_id_file)
+            log.info(
+                'Creating image identifier: {0} in {1}'.format(
+                    image_id, image_id_file
+                )
+            )
             with open(image_id_file, 'w') as identifier:
-                identifier.write('%s\n' % image_id)
+                identifier.write('{0}{1}'.format(image_id, os.linesep))
 
     def set_selinux_file_contexts(self, security_context_file):
         """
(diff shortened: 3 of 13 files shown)


schaefi:
There is a bug in ```get_mountpoint``` when used with the btrfs volume manager. I will fix this now

schaefi:
Also extended the oem integration test with a disk.sh example. Builds here and should be fixed now:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:suse/test-image-oem

schaefi:
@isbm my tests were successful so far. I think what needs to be clarified is, if the parameter(s) passed to the script suits you ? At the moment I only pass the device node to the root filesystem. Do you need more ?

Vogtinator:
I read the PR and diff and it's not entirely clear to me what environment the `disk.sh` script is running in. To support the desired functionality of removing stuff which was only needed during build, but not in the final root (such as dracut or grub themes), it would need to be running inside the mounted disk, but the documentation only says that the disk device is passed as parameter?

If this indeed runs in the target FS, I wonder what prompted you to do a 180° change in opinion compared to two years ago, where exactly this change was requested but vehemently objected: https://github.com/OSInside/kiwi/issues/625#issuecomment-366244972

davidcassany:
@Vogtinator the motivation of this hook is not removing packages, of course you can do so in there, the same as you can in `config.sh`. The motivation of this PR is providing a way to jump in over the actual loop device and apply changes there by running a chrooted script (thus it is still not functional to remove packages such as `rpm`). This opens a door for changes to elements that are not part of the root-tree (files attributes, bootloader, partition table, etc.).

> it would need to be running inside the mounted disk, but the documentation only says that the disk device is passed as parameter?

Yes, this is the case. It runs over the mounted disk and the device is passed as a parameter so the script can eventually make use of this information or not, this is just to provide context.

Vogtinator:
> @Vogtinator the motivation of this hook is not removing packages, of course you can do so in there, the same as you can in `config.sh`. The motivation of this PR is providing a way to jump in over the actual loop device and apply changes there by running a chrooted script (thus it is still not functional to remove packages such as `rpm`).

Well, https://github.com/OSInside/kiwi/issues/1464#issuecomment-635592011 implies that a script like this was used to remove dracut from the root fs after the initrd got created.

> > it would need to be running inside the mounted disk, but the documentation only says that the disk device is passed as parameter?
> 
> Yes, this is the case. It runs over the mounted disk and the device is passed as a parameter so the script can eventually make use of this information or not, this is just to provide context.

Great! This means the bootloader config and other files impacted by kiwi after config.sh can finally be edited.

schaefi:
>Great! This means the bootloader config and other files impacted by kiwi after config.sh can finally be edited.

It would be one use case for this script hook yes

schaefi:
As I wrote we need to clarify on the parameters that should be passed along to the script. It would be good if we get some feedback from the people who intent to use it. We will be able to add params but we can't change it incompatible once the feature is public. That's why I added the feedback label

isbm:
@schaefi yep, that works now just fine. I could happily _remove a lot of packages_ from the image just like that, including Zypper, RPM itself and many other stuff. :wink: I think we are good to go for now.

isbm:
@schaefi As of parameters, in fact, you could do not parameters, but env variables. So instead of having `$@` (frankly, a perfect mess-able way to run into incompatibilities, bad documentation, wrong refactorings, mistakes etc), we could instead access `$KIWI_<WHATEVER>` instead (`$KIWI_ROOT_NODE` for example?) and then this could be extended & documented much better than just an array of "stuff".

Vogtinator:
Currently I don't see the point of getting the disk as parameter, for two reasons:

- The script is running chrooted, so the partition is mounted. This means that most operations on that partition would not be a good idea
- It can easily be determined by `findmnt -no SOURCE /` anyway

> @schaefi As of parameters, in fact, you could do not parameters, but env variables. So instead of having `$@` (frankly, a perfect mess-able way to run into incompatibilities, bad documentation, wrong refactorings, mistakes etc), we could instead access `$KIWI_<WHATEVER>` instead (`$KIWI_ROOT_NODE` for example?) and then this could be extended & documented much better than just an array of "stuff".

+1 on that, it would also supersede the `/.profile` and `/.kconfig` files

schaefi:
ok good point, and makes sense. Thanks I'll adapt

schaefi:
ok I have done the change in code and did a test. my disk.sh script is now

```
env | grep kiwi
```

and that results in

```
INFO: 16:54:16 | Calling disk.sh script
DEBUG: 16:54:16 | --------------out start-------------
DEBUG: 16:54:16 | kiwi_rootpartuuid=6a716152-46da-490a-bd75-3923cd8a3b5d
DEBUG: 16:54:16 | kiwi_keytable=us
DEBUG: 16:54:16 | kiwi_type=oem
DEBUG: 16:54:16 | kiwi_install_volid=INSTALL
DEBUG: 16:54:16 | kiwi_cmdline=console=ttyS0 splash
DEBUG: 16:54:16 | kiwi_displayname=OEM-openSUSE-Tumbleweed
DEBUG: 16:54:16 | kiwi_btrfs_root_is_snapshot=true
DEBUG: 16:54:16 | SUDO_COMMAND=/home/ms/Project/kiwi/.tox/3.6/bin/kiwi-ng system build --description build-tests/x86/suse/test-image-oem --target-dir /tmp/mytest/ --allow-existing-root
DEBUG: 16:54:16 | kiwi_oemunattended=true
DEBUG: 16:54:16 | kiwi_firmware=efi
DEBUG: 16:54:16 | kiwi_bootloader_console=serial
DEBUG: 16:54:16 | kiwi_iname=OEM-openSUSE-Tumbleweed
DEBUG: 16:54:16 | kiwi_iversion=1.42.1
DEBUG: 16:54:16 | kiwi_oemswapMB=1024
DEBUG: 16:54:16 | kiwi_Volume_1=home|freespace:30|home
DEBUG: 16:54:16 | kiwi_Volume_2=LVRoot|size:all|
DEBUG: 16:54:16 | kiwi_loader_theme=openSUSE
DEBUG: 16:54:16 | kiwi_language=en_US
DEBUG: 16:54:16 | kiwi_bootloader=grub2
DEBUG: 16:54:16 | kiwi_startsector=2048
DEBUG: 16:54:16 | kiwi_splash_theme=bgrt
DEBUG: 16:54:16 | PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/ms/Project/kiwi/.tox/3.6/bin:/home/ms/Project/mash/.tox/3/bin:/home/ms/Project/kiwi/.tox/3/bin:/home/ms/Project/kiwi/.tox/3.6/bin
DEBUG: 16:54:16 | kiwi_timezone=Europe/Berlin
DEBUG: 16:54:16 | kiwi_initrd_system=dracut
DEBUG: 16:54:16 | kiwi_installboot=install
DEBUG: 16:54:16 | --------------out stop--------------
```

So as a user of this script hook you would have access to those variables.

@Vogtinator that way the .profile as a file can also be eliminated but I would do this in a follow up refactoring pull request at a later time

makes sense ?

isbm:
Perfect! I won't complain for the variables normally UPPERCASE, but I won't complain either if you address that too, if not much work. :wink:

kiwi/system/profile.py
```diff
@@ -53,6 +53,18 @@ def __init__(self, xml_state):
...
         self._root_partition_uuid_to_profile()
 
+    def get(self):
+        """
+        Return all profile elements that has a value
+        """
+        profile = {}
+        for key, value in list(self.dot_profile.items()):
+            if value:
+                profile[format(key)] = self._format(value)
```
isbm, line 63: This `format` doesn't do anything useful. And also this is a bug: it won't export value with an empty string, which is incorrect: empty string **is** a value. The `_format` function doesn't care about `False` to "false" either.

I would also start adding annotations.

kiwi/system/setup.py
```diff
@@ -95,6 +96,9 @@ def import_description(self):
         self._import_custom_archives()
         self._import_cdroot_archive()
 
+    def script_exists(self, name):
+        return os.path.exists(self.description_dir + '/' + name)
```
isbm, line 100: @schaefi why is that string concats all over the place? A typical way would be:

```python
def script_exists(self, name:str) -> bool:
    """
    I am documentation and this would be a failure, if PyLint would be in use.
    """
    return os.path.exists(os.path.join(self.description_dir, name))
```

kiwi/system/setup.py
```diff
@@ -892,14 +908,18 @@ def _import_custom_scripts(self):
...
+    def _call_script(self, name, option_list=None):
         script_path = os.path.join(self.root_dir, 'image', name)
         if os.path.exists(script_path):
+            options = option_list or []
             command = ['chroot', self.root_dir]
             if not Path.access(script_path, os.X_OK):
                 command += ['bash']
-            command += ['/image/' + name]
-            config_script = Command.call(command)
+            command += ['/image/' + name] + options
```
isbm, line 918: This does work, yes, but I would suggest to clean this up and make it it less Perl-ish:
```python
command.append(os.path.join("/image", name))
command.extend(option_list)
```

Ideally, replace "/image" string with the constant variable.

kiwi/system/setup.py
```diff
@@ -892,14 +908,18 @@ def _import_custom_scripts(self):
...
+            profile = Profile(self.xml_state)
+            caller_environment = os.environ
```
isbm, line 920: **Bug!** (and your tests didn't caught it!). If you want to copy the environment, then you should not make just another reference to the same memory address, but actually copy the whole thing like so:

```python
import copy
...
caller_environment = copy.deepcopy(os.environ)
```

Right now you do not even need that parameter, as you modifying it globally:

```
>>> os.environ["BLAH"]
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/os.py", line 675, in __getitem__
    raise KeyError(key) from None
KeyError: 'BLAH'
>>> call_env = os.environ
>>> call_env["BLAH"] = "blah"
>>> os.environ["BLAH"]
'blah'
```

kiwi/volume_manager/btrfs.py
```diff
@@ -287,6 +287,21 @@ def umount_volumes(self):
...
+
+        Effective use of the directory is guaranteed only after sync_data
+
+        :return: directory path name
+
+        :rtype: string
+        """
+        sync_target = self.mountpoint + '/@'
+        if self.custom_args['root_is_snapshot']:
+            sync_target = self.mountpoint + '/@/.snapshots/1/snapshot'
```
isbm, line 302: Redefinitions aren't nice, as well as accessing keys for checks unsafely and incrementing immutable strings. I also find that code confusing. I would suggest to write it that way:

```python
def get_mountpoint(self) -> str:
    sync_target:Dict[str] = [self.mountpoint, '@']
    if self.custom_args.get("root_is_snapshot"):  # This will fire if key is False or is not present at all
        sync_target.extend([".snapshots", "1", "snapshot"])
    return os.path.join(*sync_target)
```

kiwi/system/profile.py
```diff
@@ -53,6 +53,18 @@ def __init__(self, xml_state):
         self._drivers_to_profile()
         self._root_partition_uuid_to_profile()
 
+    def get(self):
```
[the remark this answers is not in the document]
schaefi: agreed I'll come up with a better name

kiwi/system/profile.py
schaefi: > This format doesn't do anything useful

yes, absolutely

> it won't export value with an empty string, 

yes, I think this was done intentional because the initrd code that makes use of the variables does not handle variables with no value. But I will allow it in the _format helper now as the origin of the content for those variables comes from the image description and all constraints for them are applied earlier.

> I would also uppercase it to be consistent with the environment overall

I can't do this because it would break the code that uses the variables. This would be the case in the dracut code for kiwi and it would break for users that use our variables in their custom scripts. Thus I'm afraid can't uppercase them. You should have in mind the variables you see here are from a .profile file which we offer to people to be sourced in their custom scripts. Also the same .profile file will be placed in the dracut modules and sourced there as well. Thus I have to care for not breaking this.

isbm: Uppercasing it is just "to be nice", but if that might break stuff, yeah, better not. Still, out of curiosity, why it cannot be done? You suppose to call any arbitrary script (`disk.sh` in this case, but so valid to all others) in their _own space_ (look at the bug below I pointed you out). So that said, you are exporting all the environment _exclusively to that script_, which should by no means never affect anything outside of it.

At least that would be my design...

kiwi/volume_manager/btrfs.py
schaefi: yes this is nice and with annotations

kiwi/system/setup.py
schaefi: Thanks, yes makes a lot sense

kiwi/system/setup.py
schaefi: ok after reading more parts of that code I think it deserves some cleanup. The /image location imho deserves to be a default setting as it's also used at other classes and I will apply your suggestions to be less Perl-ish ... man I've not done perl for years ;)

kiwi/system/profile.py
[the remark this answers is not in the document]
schaefi: I decided for get_settings

kiwi/system/setup.py
schaefi: I had some not so nice experience with os.path.join in the past. But let's assume this is no longer an issue with py3

schaefi:
@isbm I think I have addressed your review issues. Thanks much for taking the time and the in depth look to my crappy code :) How about it now ?

kiwi/system/profile.py
schaefi: > f you done it right, then the script disc.sh should see KIWI_STUFF and dracut should keep seeing kiwi_stuff. To do so, I would:

yes this is correct and we can uppercase the variables and keep the .profile file contents as they are right now. But wouldn't that be inconsistent ? Given that I can't change to uppercase in the .profile file for compat reasons we would need to stick with that difference for a long time, or am I mistaken

kiwi/system/profile.py
```diff
@@ -385,10 +393,14 @@ def _text(self, section_content):
...
-        the boot code expects them
+        Helper method to format None and Bool profile values in
+        the way the boot code expects them
         """
         if value is True:
             return 'true'
+        elif value is False:
+            return 'false'
+        elif not value:
+            return ''
```
isbm, line 404: It is better, but still a bug (even unlikely will be ever triggered). :wink:  If you get non-string value, e.g. `0`, it will return "". This is not what we want (hence why I am insisting on MyPy and anno's). So as long as we want it safe enough, I would do this way (could not find a better error exception type):

```python
from typing import Optional, Any
from collections.abc import Iterable

def _format(self, value: Optional[Any]) -> str:
    """
    Helper method to format None and Bool profile values in
    the way the boot code expects them.

    Returns: String value representation
    Raises: KiwiError if value is iterable (list, dict, tuple etc).
    """
    if isinstance(value, Iterable):
        raise KiwiError("Value cannot be %s", type(value))
    elif isinstance(value, bool):
        return str(value).tolower()
    return str(value or "")
```

Also I would move this function to `utils/misc.py`.

kiwi/volume_manager/btrfs.py
```diff
@@ -287,6 +287,21 @@ def umount_volumes(self):
...
+        """
+        Provides btrfs root mount point directory
+
+        Effective use of the directory is guaranteed only after sync_data
+
+        :return: directory path name
+
+        :rtype: string
+        """
+        sync_target: [list][str] = [self.mountpoint, '@']
```
isbm, line 300: [I would not use](https://www.python.org/dev/peps/pep-0484/#type-comments) direct built-in for lists. This is an old way but after 3.5 was out, this is gone. And if you will use MyPy, it will work better.

```python
from typing import List

sync_target: [List][str] ....
```

isbm:
> yes this is correct and we can uppercase the variables and keep the .profile file contents as they are right now. But wouldn't that be inconsistent ? Given that I can't change to uppercase in the .profile file for compat reasons we would need to stick with that difference for a long time, or am I mistaken

OK, at least _my personal_ reasoning (intuition-based, frankly) behind the uppercase is solely "How do I see Kiwi's stuff within that script, **regardless** what I defined in my own env additionally to it". So my logic would be to use those "KIWI_WHATEVER" variables as those that are injected at runtime of my script. Convenient and pleasant that my additional vars are are _also_ there.

But I have to admit that I do not know if there are any other use of randomcase :-) variables already somewhere. I am just saying that having it non-upper is very unnatural to a typical shell scripts and is not how it is [typically defined by GNU guys](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html). That said, I would (as user of my scripts) would definitely prefer them uppercase.

Now it is solely your call, I won't insist on it. But I were you, I would (below would be my personal strategy):

- Add a method that detects if there are any non-uppercase vars used in custom `config.sh` or whatever, and then nag-screen that very explicitly in logs and STDERR that in version XYZ that goes away so don't use anymore please. You probably still need that mechanism anyway, if you want to develop/refactor stuff and push everyone to adjust accordingly.

- Export variables alongside existing one if anything like that already exposed in other scripts.
- Do not export anything in randomcase to `disk.sh` and that's fresh start.

And `sed -ei ....` would do the job to migrate things, if any. :wink: Leaving as it is now will keep organically growing, but not sure if that is what you want. Anyways, your call!

kiwi/defaults.py
```diff
@@ -1526,6 +1526,26 @@ def get_default_packager_tool(package_manager):
...
+        return 'disk.sh'
+
+    @staticmethod
+    def get_image_metadata_directory():
+        """
+        Provides base directory to store the image description
+        and its metadata inside of the system root tree
+        """
+        return 'image'
+
```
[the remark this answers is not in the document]
schaefi: yes that's way better for static values.

kiwi/volume_manager/btrfs.py
schaefi: for sure a better read, thanks

kiwi/system/profile.py
schaefi: > if you get non-string value, e.g. 0, it will return "".

Uhh, yes you are right

schaefi: ```<class 'str'>``` is also an Iterable

schaefi: so the above will raise on strings

schaefi: ok moved _format into the Shell class format_to_variable_value()

test/unit/system/shell_test.py
```diff
@@ -0,0 +1,17 @@
...
+
+class TestSystemShell:
+    def test_format_to_variable_value(self):
+        assert Shell.format_to_variable_value('text') == 'text'
+        assert Shell.format_to_variable_value(True) == 'true'
+        assert Shell.format_to_variable_value(False) == 'false'
+        assert Shell.format_to_variable_value('42') == '42'
+        assert Shell.format_to_variable_value(0) == '0'
+        assert Shell.format_to_variable_value(42) == '42'
+        assert Shell.format_to_variable_value(None) == ''
```
isbm, line 15:
```python
assert Shell.format_to_variable_value(b"42") == "42"
```
:yum:

schaefi: binary stuff at this point, I really hope that this path is never reached :) ...anyway your test passes now ;)

kiwi/system/setup.py
isbm: @schaefi what problems, BTW? I actually never had any issues, but you need to be careful with slashes:

```
>>> os.path.join("this", "is", "/wrong")
'/wrong'
```
But this is also works as designed:

```
join(a, *p)
    Join two or more pathname components, inserting '/' as needed.
    If any component is an absolute path, all previous path components
    will be discarded.  An empty last part will result in a path that
    ends with a separator.
```

schaefi: I think I had some trouble with path.join on relative references. Something like "/foo/bar/../baz". But this is long time ago and I really can't remember good enough to open a conversation about it. I'm sure it behaved as designed but caused me some headaches.

schaefi: > as long as you get rid of those strings concats, this should do

yep we care for this, however I think you will find code that still does this. Whenever it comes up we fix it :)

schaefi: not here please :) that PR already has grown too big for more cleanup/refactoring

isbm:
Seems good to me. :partying_face:

schaefi:
Thanks much for the in depth review. Your comment how to deal with the variables is valuable and I don't want to loose it with the merge of this pull request history. I will open an issue and put your idea how to handle uppercase/vs lowercase variables and the compat concern into a new issue.

Thanks

isbm:
You will see me soon again. I haven't finished with you for another 20 rounds, no worries. 🤣🤣 Nice to work together again. 👍👍

PR #1472, opened 2020-06-04
Fixed size of msgbox dialog

schaefi:
The width of the dialog was set to a small value which
causes the message to be choped. I found this when testing
pxe deployments. The error messages on "Failed to fetch..."
were missing the interesting part

### DIFF
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -10,7 +10,7 @@ type fetch_file >/dev/null 2>&1 || . /lib/kiwi-net-lib.sh
 #--------------------------------------
 function report_and_quit {
     local text_message="$1"
-    run_dialog --timeout 60 --msgbox "\"${text_message}\"" 5 70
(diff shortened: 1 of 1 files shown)


dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
```diff
@@ -10,7 +10,7 @@ type fetch_file >/dev/null 2>&1 || . /lib/kiwi-net-lib.sh
 #--------------------------------------
 function report_and_quit {
     local text_message="$1"
-    run_dialog --timeout 60 --msgbox "\"${text_message}\"" 5 70
+    run_dialog --timeout 60 --msgbox "\"${text_message}\"" 5 100
```
Conan-Kudo, line 13: We probably can't widen this more than 80 columns due to potential limitations with serial consoles...

schaefi: I've tested on serial consoles and interesting enough it did not harm if the width is set wider than the actual console. I guess there is code in dialog that checks that. But 80 is fine by me too.

Issue #1421, opened 2020-04-28
port and update xml description docs from legacy kiwi


schaefi:
## Problem description

The current kiwi next generation has no proper documentation for the xml elements. The auto generated schema documentation does not serve this purpose because it shows available elements and attributes and their relations but no explanation about this information. This tasks has been done in the past by the openSUSE doc team and ended up in docbook sources in the legacy kiwi git repo.

With help from Tom I have done some experimental conversion using:

```
pandoc --from=docbook --to=rst xml/kiwi-doc-description.xml
```

That produces good results to be included into the kiwi next generation ReST documentation.
The result of course needs adaptions according to the current schema but is a good start.
As we also have an auto translation from ReST to the suse documentation we will automatically increase that part which is one of the roadmap items in my team (public cloud dev)

schaefi:
posting from Tom

```
$ cd doc/docbook
$ pandoc --from=docbook --to=rst xml/kiwi-doc-description.xml

Mir ist dabei folgendes aufgefallen:

* <indexterm>s bleiben erhalten, genauso <remark>s. Die könnte man vorher
  per XSLT rauswerfen.

* <filename>config.xml</filename> wird zu ``config.xml``, das ist gut.

* Ein Bild mittels <figure> wird korrekt in eine entsprechende RST-Direktive
  umgewandelt, inklusive Breite und Titel. Auch gut.

* Auch <screen> wird korrekt umgesetzt.

* Listen scheinen zu funktionieren.

* Wenn du <screen language="xml"> hast, überträgt er die Codeinformation
  korrekt nach RST. Syntaxhighlighting ist damit gerettet! \o/

* Ein <sgmltag> wird ohne alles umgesetzt. Das finde ich etwas suboptimal,
  aber sowas könnte man durch XSLT beheben.

* Ein <command> wird leider nicht nach :command:`...` umgewandelt. Ebenso
  nicht <option>. Bei letzteren wird allerdings die doppelten Backticks
  verwendet.

So auf den ersten Blick würde ich sagen sieht es recht gut aus. Man müsste
vermutlich nochmals richtig durchgehen und evlt. Querverweise korrekt auflösen
oder entfernen.
```

PR #1474, opened 2020-06-05
Add text about how to add pattern names

tomschr:

* Extend section `_adding-and-removing-packages` with how to add patterns
* Taken from the discussion "How to specify package patterns?" on 2020-06-05T10:18

### DIFF
--- a/doc/source/working_with_kiwi/xml_description.rst
+++ b/doc/source/working_with_kiwi/xml_description.rst
@@ -707,8 +707,10 @@ following paths types:
 Adding and removing packages
 ----------------------------
 
-Now that we have defined the repositories, we can define which packages
-should be installed on the image. This is achieved via the `packages`
+Now that we have defined the repositories, we can define which packages or
+patterns should be installed on the image.
+
+This is achieved via the `packages`
 element which includes the packages that should be installed, ignore or
 removed via individual `package` child elements:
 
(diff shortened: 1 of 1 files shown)


schaefi:
Thanks Tom, this was a good idea. As I plan to convert the xml description from the former docbook source as described in #1421 I fear this will conflict with the information you added here right now.

My suggestion is to keep this open and merge it into the converted area. I can do that once I'm done with it, ok ?

tomschr:
Thanks Marcus! All good, do whatever you think is useful. :+1: 
I just thought, it would be helpful to not forget it as I consider it useful information.

doc/source/working_with_kiwi/xml_description.rst
```diff
@@ -757,6 +759,30 @@ configured via the following attributes:
 We will describe the different child elements of `packages` in the following
 sections.
 
+If you want to install patterns, use the `namedCollection` element:
```
Conan-Kudo, line 762: This should be generalized. `namedCollections` is how you install "groups" in RH/Fedora and "patterns" in SUSE, so this should probably be reworded to account for that.

schaefi: yes I did that when I added the information to the current xml_description_doc_update branch. Thanks

schaefi:
@tomschr I have added this information in addition to the massive doc update I'm working on in the xml_description_doc_update branch

PR #1477, opened 2020-06-07
Py3 4 compat

schaefi:
Maintain custom type hinting patch
    
If we want the upstream code to function on Python 3.4 the  current use of type hints is not possible. The syntax used in our code is available for Python 3.5 or later. Distributions
like sle12-sp5 still use Python 3.4 and we claim to support  this target. However 3.4 won't be maintained after March 2019 (cf PEP 429) which means upstream we don't care for this Python version. In order to allow using the latest code I created a custom patch which we can maintain on the package level and also port to the enterprise packages if we want to renew the code base for e.g sle12-sp5 before the distro  reaches EOL.


schaefi:
I'm not really convinced about this solution but I don't see how else it could be done

Conan-Kudo:
@schaefi Have you considered cutting off SLE 12 from kiwi releases? Or perhaps just maintaining the patch in the SLE 12 maint tree in the SUSE org?

schaefi:
> Or perhaps just maintaining the patch in the SLE 12 maint tree in the SUSE org?

yes I thought about this part but then I realized we have integration tests for the upstream code running against sle12 for ppc here:

* https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_ppc:sle12

So without the patch in the package spec here this test will not be successful anymore. Actually I had not planned to run tests for the enterprise in combination with tests for the open platforms but this one was created for IBM and their cloud offering and they use enterprise distros exclusively at the moment. This is a tough issue:

* we can eliminate enterprise integration tests from the matrix
* we can delete py3 type hints and annotations

both is not really acceptable I think. I also want to see when the patch breaks as the code upstream evolves. So I think it's needed that we have it here.

I only know about sle12 as one enterprise distro that has py 3.4 on board until eol, maybe there are more which we claim to support ? Also in this case it would be good to keep the patch here

Conan-Kudo:
Ubuntu 16.04 is Python 3.5, RHEL 8 and SLE 15 is Python 3.6, Debian is Python 3.7. SLE 12 is the only one with Python 3.4, but we shouldn't care about it anymore here...

davidcassany:
I have to admit I don't like the solution. Wouldn't it be simpler just avoid using typing hints unit next code renew for SLE15-SP5? I know we have no specific plans for SLE12-SP5, but this will stay there for a while, so I think it would be reasonable to see a new KIWI release there at some point.

Probably we could plan the adoption of typing hints and place it at some point in time, instead of starting to use them now. So we just delay typing hints adoption.

We could also do the other way round, which is add this ugly patch and remove it after we make a new sle12-SP5 release, assuming that whatever need to be done afterwards in there should only be back ports. But IMHO this is a bit messy.

schaefi:
> I have to admit I don't like the solution.

I don't like it either

> Wouldn't it be simpler just avoid using typing hints unit next code renew for SLE15-SP5?

How do we handle sle12 based integration tests ?
* https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_ppc:sle12

I personally want to have that patch only in the kiwi_sle12 github project but that means we have to delete the above integration test which I'm willing to do since the same exists for sle15 as well.

But it would be a regression and I bet IBM will not wait long to approach us what happened with it

Given we put on type hints and annotation more and more the work on porting to sle12 will also increase. when maintaining a patch file here it will just grow as we develop and just applies on porting.

Overall I agree with you, I don't like it and I'm just trying to have an effortless porting process

Conan-Kudo:
Something has to give. This is a bad idea to implement and maintain, and you're really probably better off maintaining a kiwi version branch for SLE 12.

schaefi:
> Something has to give. This is a bad idea to implement and maintain, and you're really probably better off maintaining a kiwi version branch for SLE 12

yes we even maintain an extra git repo for sle12 and sle15 to follow suse policies.

I had a conversation with IBM and they are fine with moving the integration test to the internal buildservice. I did this today and moved the test to: QA:Maintenance:Images:12-SP5:kiwi-ng-testing_ppc/test-image-vmx-oem

The integration test has been deleted from kiwi.

Once we update kiwi in sle12 it is expected that the test in ibs fails and we need to create a patch in the sle12 repo that basically deletes all type hints and annotations. With a bit of luck there will be no kiwi update in sle12-sp5 needed anymore :)

PR #1479, opened 2020-06-10
Only modify /etc/vimrc if it exists in config.sh

dcermak:
An update of vim in Tumbleweed will move `/etc/vimrc` to `/usr/share/vim` as part of the /usr - /etc split. This makes the sed call fail because `/etc/vimrc` no longer exists.

However, the fix is not required anymore then, as the vim package dropped the `syntax on` line from the default vimrc.

### DIFF
--- a/build-tests/x86/suse/test-image-oem/config.sh
+++ b/build-tests/x86/suse/test-image-oem/config.sh
@@ -61,12 +61,6 @@ rm -rf /usr/share/doc/manual/*
 rm -rf /opt/kde*
 
 #======================================
-# only basic version of vim is
-# installed; no syntax highlighting
-#--------------------------------------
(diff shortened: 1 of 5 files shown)


davidcassany:
Thanks for the PR, however I would be more in favor of just dropping the offending lines. Because those tests or only against TW, thus there is no need to consider other scenarios, and also because I don't think those tests should care at all about vim configuration.

@schaefi what you think? IMHO we should just remove needless stuff from these scripts unless there is a particular feature we want test or exercise (I can't think of any now).

schaefi:
@davidcassany I agree to what you have said, just remove needless stuff

dcermak:
Marcus Schäfer <notifications@github.com> writes:

> @davidcassany I agree to what you have said, just remove needless
> stuff

Done.

PR #1482, opened 2020-06-12
Remove /etc/sysconfig/language support

### LINKED ISSUE #1471: Kiwi writes to `/etc/sysconfig/language` instead of `locale.conf` for languages settings
## Problem description
Kiwi writes to `/etc/sysconfig/language` instead of `locale.conf` for languages settings on OEM image.

## Expected behaviour
Kiwi should write to `locale.conf` for languages settings.

Build an image with `<locale>en_US</locale>`.

+ KIWI version: 9.20.9
+ Operating system: openSUSE Tumbleweed aarch64

davidcassany:
As of SLE15 and onwards /etc/sysconfig/language is considered to be
obsolete and just kept for compatibility purposes. Thus there is no
need to manage the file anymore.

Fixes #1471

### DIFF
--- a/doc/source/working_with_kiwi/shell_scripts.rst
+++ b/doc/source/working_with_kiwi/shell_scripts.rst
@@ -263,6 +263,30 @@ $kiwi_type
 Configuration Tips
 ------------------
 
+#. **Locale configuration:**
+
+  KIWI in order to set the locale relies on :command:`systemd-firstboot`,
+  which in turn writes the locale configuration file :file:`/etc/locale.conf`.
+  The values for the locale settings are taken from the description XML
+  file in the `<locale>` element under `<preferences>`.
+
+  KIWI assumes systemd adoption to handle these locale settings, in case the
+  build distribution does not honor `/etc/locale.conf` this is likely to not
+  produce any effect on the locale settings. As an example, in SLE12
+  distribution the locale configuration is already possible by using the
+  systemd toolchain, however this approach overlaps with SUSE specific
+  managers such as YaST. In that case using :command:`systemd-firstboot`
+  is only effective if locales in :file:`/etc/sysconfig/language` are
+  not set or if the file does not exist at all. In SLE12
+  :file:`/etc/sysconfig/language` has precendence over
+  :file:`/etc/locale.conf` for compatibility reasons and management tools
+  could still relay on `sysconfig` files for locale settings.
+
+  In any case the configuration is still possible in KIWI by using
+  any distribution specific way to configure the locale setting inside the
+  :file:`config.sh` script or by adding any additional configuration file
+  as part of the overlay root-tree.
+
 #. **Stateless systemd UUIDs:**
 
   Machine ID files are created and set (:file:`/etc/machine-id`,
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -338,13 +338,6 @@ class SystemSetup:
                     'chroot', self.root_dir, 'systemd-firstboot',
                     '--locale=' + locale
                 ])
-            if os.path.exists(self.root_dir + '/etc/sysconfig/language'):
-                Shell.run_common_function(
-                    'baseUpdateSysConfig', [
-                        self.root_dir + '/etc/sysconfig/language',
-                        'RC_LANG', locale
-                    ]
-                )
 
     def setup_timezone(self):
         """
--- a/test/unit/system/setup_test.py
+++ b/test/unit/system/setup_test.py
@@ -394,11 +394,10 @@ class TestSystemSetup:
             self.setup.setup_keyboard_map()
 
     @patch('kiwi.system.setup.CommandCapabilities.has_option_in_help')
-    @patch('kiwi.system.setup.Shell.run_common_function')
     @patch('kiwi.system.setup.Command.run')
     @patch('os.path.exists')
     def test_setup_locale(
-        self, mock_path, mock_run, mock_shell, mock_caps
+        self, mock_path, mock_run, mock_caps
     ):
         mock_caps.return_valure = True
         mock_path.return_value = True
@@ -411,24 +410,22 @@ class TestSystemSetup:
                 '--locale=locale1.UTF-8'
             ])
         ])
-        mock_shell.assert_called_once_with(
-            'baseUpdateSysConfig', [
-                'root_dir/etc/sysconfig/language', 'RC_LANG', 'locale1.UTF-8'
-            ]
-        )
 
-    @patch('kiwi.system.setup.Shell.run_common_function')
+    @patch('kiwi.system.setup.CommandCapabilities.has_option_in_help')
     @patch('kiwi.system.setup.Command.run')
     @patch('os.path.exists')
-    def test_setup_locale_POSIX(self, mock_path, mock_run, mock_shell):
+    def test_setup_locale_POSIX(self, mock_path, mock_run, mock_caps):
+        mock_caps.return_valure = True
         mock_path.return_value = True
         self.setup.preferences['locale'] = 'POSIX,locale2'
         self.setup.setup_locale()
-        mock_shell.assert_called_once_with(
-            'baseUpdateSysConfig', [
-                'root_dir/etc/sysconfig/language', 'RC_LANG', 'POSIX'
-            ]
-        )
+        mock_run.assert_has_calls([
+            call(['rm', '-r', '-f', 'root_dir/etc/locale.conf']),
+            call([
+                'chroot', 'root_dir', 'systemd-firstboot',
+                '--locale=POSIX'
+            ])
+        ])
 
     @patch('kiwi.system.setup.Command.run')
     def test_setup_timezone(self, mock_command):


davidcassany:
@schaefi I did the PR to open the discussion.

The problem with this PR is that it breaks locale configuration for SLE 12.x.

The `/etc/sysconfig/language` file is considered to be obsolete in SLE15, however it is kept just for compatibility reasons. Moreover, if set, it takes precedence over `/et/c/locale.conf` which is the correct file to manage. In KIWI for SLE15 and SLE12 we set both: `/etc/locale.conf` is set if the installed systemd version supports `systemd-firstboot --locale` and `/etc/sysconfig/language` file is edited if the file exists. Both conditions are true for SLE12 and SLE15, thus the result is setting both in both distros. `/etc/locale.conf` is ignored on SLE12 and this is fine, however in SLE15 this also causes `/etc/locale.conf` to be ignored becuase if both are set `/etc/sysconfig/language` takes precedence. 

My concern with this PR is that if at some point this gets into SLE12 stream this will break the locale configuration there (break it the sense that it will not be effective as expected). This is unfortunate because if at some point we renew the SLE12 code there won't be an integration test failing because of this behavior change, thus we are in risk of introducing a regression. I am not sure how to keep track of this change in a way it doesn't get into SLE12.

On the other side the alternate solution is adding additional code to discriminate SLE12 from SLE15 (by parsing the `/etc/os-release` or equivalent) just to keep compatibility with a distro we already now that is already no longer compatible with upstream code. It just feels dumb this approach.

Any thoughts?

schaefi:
> The problem with this PR is that it breaks locale configuration for SLE 12.x.

correct me when the following is wrong: if we merge this no /etc/sysconfig/language file will be created by kiwi anymore. A distribution that needs it can create the file as an overlay file with the appropriate settings correct ?

Maybe that's worth to be mentioned in our documentation, such that we can point people to a link when they stumble over it.

Other than that I really would like to avoid upstream merge of this kiwi to sle12. We are already incompatible with annotations and type hints and it continuous with this change. I assume there will be more to come that makes a 1:1 upstream merge to sle12 impossible. Thus we can say:

* we have to support the sle12 target in the scope of SUSE and our job there
* we do that for sle12 on an individual patch basis starting now
* this also means customers can't expect any upstream feature in sle12, only selective and based on bugzilla reports. Given sle12-sp5 is the last SP and sle12 is entering LTSS very soon this can be done
* kiwi in sle12 is then decoupled from upstream

Thoughts ?

davidcassany:
> correct me when the following is wrong: if we merge this no /etc/sysconfig/language file will be created by kiwi anymore. A distribution that needs it can create the file as an overlay file with the appropriate settings correct ?

This is mostly correct. The file is not created by KIWI, just modified by KIWI. KIWI just sets a variable that is unset by default. So anyone that needs to play around with this file I'd say the best approach is to run `baseUpdateSysConfig /etc/sysconfig/language RC_LANG <locale>` in config.sh as the file will mostly be already there (I am not sure which base package creates it).

> Thus we can say:
>
> * we have to support the sle12 target in the scope of SUSE and our job there
> * we do that for sle12 on an individual patch basis starting now
> * this also means customers can't expect any upstream feature in sle12, only selective and based on bugzilla reports. Given sle12-sp5 is the last SP and sle12 is entering LTSS very soon this can be done
> * kiwi in sle12 is then mostly decoupled from upstream

I am fine with this approach is matches what we discussed in the chat. Decoupling SLE12 from upstream seams to be the good approach to prevent additional tech debt on upstream.

> If we do it I think we have to write something in the docs

The question is where to add it? I just realized we do not have release notes in our docs (neither I am sure we need to produce release notes). Does it deserve another Note entry on the main docs page?

schaefi:
There is this chapter:

https://osinside.github.io/kiwi/working_with_kiwi/shell_scripts.html?highlight=function#configuration-tips

I think it deserves an extra space such that it can be found easier but I think such a tip page for those system configurations kiwi touches is useful

isbm:
If you, guys, want to get my few unbiased cents, then I would say this is a Bad Design TM. Because barebone Kiwi supposed to be as agnostic, as possible. So I would:

1. Rip off the whole locale stuff entirely from the Kiwi core code
2. Create a module that takes care of a specific distribution of the day and mess with that there
3. Repeat the above for everything else distro-specific and put all that into modules.

HTH.

Conan-Kudo:
@isbm We can rely on distro-agnostic configuration. And in general, our philosophy is to minimize or eliminate distro-aware modules in kiwi. The old kiwi had too much of this and made it impossible to maintain the code.

isbm:
@Conan-Kudo what kind of _modules_ you are talking about? I see quite a plenty of hard-coded stuff in the Kiwi. Would be nice to have something like `modules/` subdir and then something like:

```xml
<module name="locale" />
<module name="system.uboot" />
<module name="net.slxos.config>
  <param name="foo" value="bar" type="str" />
</module>
```

etc.

davidcassany:
@schaefi I added a locale settings configuration tip section to explicit that any distro that does not honor `/etc/locale.conf` will not get the locales configured by the standard KIWI build process.

I'd appreciate if you could have a look the little text I added. Code wise I believe that the PR is ready to be merged.

davidcassany:
@isbm I see your point, probably this is something to consider to some extend. KIWI is not distro aware, meaning that the build process is never acting depending on certain distro. This, of course, creates some issues here and there, as the procedure is not equivalent for all distros and causes us to add some hacks (even there aren't that many I'd say). On the other side we do not have a bunch of different procedures or modules to maintain independently, which is good.

In any case this discussion goes way beyond the scope of this PR. So I believe the current approach is fine to fix the actual problem. But the concept of distribution awareness might be something to think about (even I find it quite frightening, to me this looks like it could be easily be abused to duplicate relevant portions of the process).

isbm:
Ya, that is beyond PR, but it is always any related discussions are always beyond any specific PR and issues aren't usually most practical way to see it relevantly. :wink:

doc/source/working_with_kiwi/shell_scripts.rst
```diff
@@ -263,6 +263,25 @@ $kiwi_type
...
+  KIWI in order to set the locale relies on :command:`systemd-firstboot`,
+  which in turn writes the locale configuration file :file:`/etc/locale.conf`.
+  The values for the locale settings are taken from the description XML
+  file in the `<locale>` element under `<preferences>`.
+
+  KIWI assumes systemd adoption to handle these locale settings, in case the
+  build distribution does not honor `/etc/locale.conf` this is likely to not
+  produce any effect on the locale settings. Note this is the case for old
+  distributions such as SLE12 in which the locale configuration is expected
+  to happen under the :file:`/etc/sysconfig/language` file which is,
```
Conan-Kudo, line 277: Note `/etc/locale.conf` *does* work on SLE 12. It's just that if `/etc/sysconfig/language` exists, the SUSE-specific hacks kick in.

davidcassany: Yes you are right. I did not want to get into details of SLE12 behavior here but just mention it as an example of a distro that does not fully embrace the locale settings via systemd.

schaefi:
> If you, guys, want to get my few unbiased cents, then I would say this is a Bad Design TM. Because barebone Kiwi supposed to be as agnostic, as possible. So I would:
> 
> 1. Rip off the whole locale stuff entirely from the Kiwi core code
> 2. Create a module that takes care of a specific distribution of the day and mess with that there
> 3. Repeat the above for everything else distro-specific and put all that into modules.
> 4. When Kiwi creates an image, it will see "Aha! I am now on SLE 15" and will run all the chain of SLE15 specific modules.

You are right kiwi tries to be distribution independent as ever possible. You can also see that by watching the integration tests that builds images for many different distributions and that would not work if we design wise would not care for what you are aiming for. Just to "defend" a little bit on the design :)

Now the reality is a little bit different. You said you would go with the very simplified approach that kiwi is not responsible for any type of system configuration because that sucks and is different across the distributions. I'm 100% in agreement with you but there are the following system setup parts that are relevant at build time of an image:

* locale (packages script runs needs a locale setup)
* timezone (packages checks that at install time, pre/post scripts)
* users/groups (needed for packages and scripts)
* keymap (required certain tool integrations, e.g autoyast)
* plymouth theme (needed to boot with configured theme)
* selinux context (needed on selinux systems to bootup)
* fstab (needed to boot)

As you can see from the list kiwi needs to handle those as it is somehow relevant to create a useful image. I want kiwi to be able to produce workable images without additional components. Thus ansible, salt... you name it should be an option on top but not mandatory. That's why we touch base on those system configurations.

Of course we thought if all of that can be done common to all distributions. Given that systemd took over most of the setup tasks we were able to make the statement that this is done by systemd and those distros that does not use systemd or are down striped like embedded solutions will just skip the step but nothing stupid happens.

The change here done by David basically eliminates one left over source where we handled a system config file /etc/sysconfig/language by kiwi itself and I'm happy to get rid of it if we have an alternative available for our users which we have and documented. That's fair enough from my perspective

Last we can think about the keymap setup which seems to be a setup where kiwi handles /etc/sysconfig/keyboard which we probably also want to get rid of. If we want to add this change here I would be ok

Overall we want to give customers hook in opportunities via script hooks that allows them to add distro specific, project specific extensions but keep kiwi free from all that as we simply will not be able to handle that.

Your suggestion how to handle distro specifics is nice and a good design but actually we don't want to provide an infrastructure to handle distro specific modules as part of kiwi itself. From my experience customers are fine if they have a hook opportunity to add their extensions as they see fit

Does this makes sense to you too ?

Thanks

isbm:
>  the following system setup parts that are relevant at build time of an image:
>
> - locale (packages script runs needs a locale setup)
> - timezone (packages checks that at install time, pre/post scripts)
> - users/groups (needed for packages and scripts)
> - keymap (required certain tool integrations, e.g autoyast)
> - plymouth theme (needed to boot with configured theme)
> - selinux context (needed on selinux systems to bootup)
> - fstab (needed to boot)
> ...
> kiwi needs to handle those

OK. No. :wink: 

I might sound here hard, but no, absolutely not. These are all **modules**. E.g. `plymouth` has no use in embedded. Zero. I always have to rip that thing off entirely. SELinux is an optional thing, can be replaced with SMACK, for example. Or AppArmor. The `keymap` is an optional thing, sometimes not needed. The boot loader installation is also should be another module etc etc. You can, in fact, get rid of the package manager support as well and only delegate that to an underlying cfgmgmt.

Point is: current Kiwi design is like a capsule of an embedded mini-configuration orchestration system inside. While what you actually want is to really purge all that and just, partition it (also module BTW!), format an image (also module BTW!), bootstrap it, put all the playbooks inside along with some cfgmgmt and "provision" it by executing those. But doing so, you will get rid of a lot of code you don't have to support, because [it is there already, way bigger ocean than your lake](https://galaxy.ansible.com) and less unnecessary job for you. And you could just focus only on merely setting things up. That way you could build a system that bakes anything, even OpenBSD or Solaris on those systems if they need to.

Time to make [KiwiRG](https://mirror.uncyc.org/wiki/Windows_RG)? :wink:

schaefi:
> I might sound here hard, but no, absolutely not. These are all modules

That's unfortunately not true. You just need to take a look at what the packages that makes up a distribution do. Their scripting part creates the need for almost all mentioned topics. As you said yourself you are spending some time to get rid of parts of the system. In your ideal world this wouldn't be necessary. The simple call ```rpm -qa --scripts``` on an rpm based system gives you
a small impression what code runs when installing from packages. I know you don't care for packages, but we care, we have to because it's part of the kiwi design. It all starts with packages.

The config bits in kiwi are not mandatory at all. They are all based on tooling. If tools to run that setup step doesn't exist the steps are skipped. If you would be able to run a package install with a dependency resolution that does NOT pull systemd, locales, splash etc etc you get what you want already today. But this is not how it works as even you have to make use of the packages of a distribution to begin with and now suffer the consequences how to get rid of all the stuff you don't need.

You are trying hard to achieve all of that; no packages, no package managers, all based on a config management system, building of immutable systems and I understand your reasons but kiwi as an appliance builder for distributions is not designed for the embedded world. If you need this and can't live with the current design unless changing it fundamentally I fear we won't be able to follow you on that effort.

> Point is: current Kiwi design is like a capsule of an embedded mini-configuration orchestration system

This was not the design aspect I had in mind. The kiwi design is based on building OS images from distributions. An appliance build by kiwi for suse should look like a suse system, build for Fedora should look like a Fedora system and so on  This includes also an absolute minimum of a system setup, timezone, locale such that you can log in and at least know about a C locale or some sort of character set. We are looking at it from different angles so I think it's fair that we both have a different opinions. I just like you to understand how we work and what needs to stay a priority as it otherwise might lead you creating lots of code we won't be able to merge. That would be very sad as it's great to receive your contributions.

From a distribution with the flag "embedded" I would expect that it does not provide what it does not need. But that's not how it works. People take some distro and rip out things such that in the end it can be used for an embedded use case. This is all fine but will not lead to a design change in kiwi.
The resulting system for the embedded use case is then also far away from the original intend of that distribution. I'm pretty sure it's possible to address any aspect of the embedded world with kiwi as it is today. It might be cumbersome here and there but that's mostly because actions needs to be taken that are against the original intend of the distribution. As a user of kiwi in this regard I would not blame on the design of the appliance builder but rather be ok with the compromise that the used distribution is not an embedded distribution. I hope this point is fair ?

I'm sorry my thinking is very different at that point. It's not the image builder that needs to be able to turn a full blown distribution into an embedded distribution, it's the distribution that should be made for this use case. When some image builder then builds an appliance from this distribution no fancy disk.sh or whatever script would be even needed to rip out core components or do "nasty" things.

> Time to make KiwiRG

That's up to you I guess :) Maybe some of the existing api components from kiwi can be helpful.
Maybe for your use case it could be a solution to just use the parts you need from kiwi and write the appliance builder as you need it ? ```from kiwi.... import ...```

Let me know which direction would work better for you

Thanks

isbm:
> You just need to take a look at what the packages that makes up a distribution do. Their scripting part creates the need for almost all mentioned topics.
Well, yes. See, I rewrote Salt from scratch, called Sugar, and now doing it again with a different architecture. Say, you need to set a locale. So you have a module that does it so, and it knows how to do it on whatever platform and thus is using required logic accordingly. All your Kiwi needs to do is issue that module and say "set me locale Foo". The module should do the rest of the job automatically.

> You are trying hard to achieve all of that; no packages, no package managers, 

Nooooo, I think you completely misunderstood me. Happens. 

Definitely you need all that. Again. Think of Kiwi formats disks it needs. Then it bootstraps needed stuff for a distro you're building. Obviously, best if  you use the same distro as you are building (SUSE on SUSE, RH on RH, Deb on Deb, FreeBSD on FreeBSD etc). Now, you install boot loader you need, main skeleton, and start installing packages. Then you configure all that and finalise the disk. So till the bootstrap -- it is surely Kiwi the way it is (and, frankly that also can be modules, Kiwi only takes care of devices setup/mounts/routing/formatting). Say, why you need to implement zypper or dnf wrapper if you have all that in some configuration management already implemented? You need to just call it with a different root destination (`/tmp/somewhere/`). You don't even need to chroot in this case.

Or I will ask you the other way around: what is possible for Kiwi that is not possible for e.g. Ansible?

schaefi:
Sorry I didn't get your point and we are pretty off topic.

> But that's exactly where you limit yourself by adding more code. 

This PR removes code and we are not adding code that is distro specific. Thus I don't see where we limit ourselves. kiwi uses core components of modern distributions and to be honest I intentionally avoided to create a deep requirement to a configuration management subsystem like salt, ansible, cim(well only a model ok) just to provide a damn simple image builder for distributions.

kiwi is not more and also aims to be not more. Describe an OS by a human readable definition and allow users to integrate whatever subsystem they need in addition to that.

> Agreed. ￼ But I am trying to let it "Aha!" for you.

No Aha for me at this point sorry. kiwi is small, needs python and linux core tools. Can be extended relatively easy, can be used as api with other python implementations and has a plugin interface.

However if you can tell me one small not gigantic and stable plus available in all major distributions
configuration management system that allows me to bootstrap any system from an empty root
without any pre-setup actions similar like this pseudo code:

```python
from kiwi.repository import Repository
from kiwi.package_manager import PackageManager

repository = Repository(...)
repository.add_repo(...)
manager = PackageManager(repository, package_manager)
manager.request_package(...)
manager.process_only_required()
```

Then we are surely happy to replace the simple repo and package manager api from kiwi and just use that. If you think this makes sense you can open an issue for it. But please remember all decisions in kiwi to code something on our own instead of using existing interfaces had been done wisely and was not just taken out of thin air. At the level of the package manager there are many changes in the distributions, yum vs. dnf, zypper plugins, suse pattern concept, fedora modules concept. All of this requires to react quickly and I don't want to patch and wait for config management to step up. As I said before it's all about packages and kiwi is a damn simple appliance builder not more. The code around package and repo management belongs to kiwi.

As I'm sure you disagree with me here :) I'm happy to continue this conversation on Riot or on the mailing list since also others should share their thoughts about this.

Regarding this issue I'd like to come back to it now and go ahead

doc/source/working_with_kiwi/shell_scripts.rst
Conan-Kudo: Could you please tweak this to be more accurate, because I don't want to give the impression this doesn't work in SLE 12.

davidcassany: Sure, what about now? I believe this is explicit enough

Conan-Kudo: LGTM.

isbm:
> available in all major distributions configuration management system that allows me to bootstrap any system from an empty root without any pre-setup actions similar like this pseudo code

Actually, it is exactly where I am heading for. Wait for it. :wink:

PR #1483, opened 2020-06-12
Fixed check for root device in grub config

schaefi:
There is a code path that fixes the grub2-mkconfig used root device
when building in an environment that does not allow to resolve the
by-X path names, e.g an obs build worker without udev. For images
that explicitly defines a root=... value in the kernelcmdline
attribute the root device check was not called because the
_get_root_cmdline_parameter method returns None. This commit fixes
the method to return the expected root device in any case such that
the grub2-mkconfig root device check has a chance to fix what
grub2-mkconfig has created

### DIFF
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -16,6 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
+import re
 import logging
 from collections import namedtuple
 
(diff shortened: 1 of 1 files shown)


schaefi:
> shouldn't we have bugzilla issue to track if we need that for SLE?

yes good point. Should be a task by Robert but as he is on vacation I might create the bug

schaefi:
bugzilla reference

https://bugzilla.suse.com/show_bug.cgi?id=1172928

PR #1485, opened 2020-06-17
Global variables

isbm:
I just found that global variables were added run-time dynamically, as well as they are all lower case, which is not the best practice.


isbm:
What you mean? You mean squash them or?

schaefi:
> You have a button for it, no?  

yes, not everybody likes this. ok I'll go ahead

isbm:
Thanks!

PR #1486, opened 2020-06-17
Factory pattern mismatch

isbm:
This fixes ~100 static typing errors

### DIFF
--- a/kiwi/boot/image/__init__.py
+++ b/kiwi/boot/image/__init__.py
@@ -33,9 +33,8 @@ class BootImage:
     :param string root_dir: system image root directory
     :param list signing_keys: list of package signing keys
     """
-    def __new__(
-        self, xml_state, target_dir, root_dir=None, signing_keys=None
-    ):
+    @staticmethod
+    def new(xml_state, target_dir, root_dir=None, signing_keys=None):
         initrd_system = xml_state.get_initrd_system()
         if initrd_system == 'kiwi':
             return BootImageKiwi(
--- a/kiwi/bootloader/config/__init__.py
+++ b/kiwi/bootloader/config/__init__.py
@@ -34,9 +34,8 @@ class BootLoaderConfig:
     :param string root_dir: root directory path name
     :param dict custom_args: custom bootloader config arguments dictionary
     """
-    def __new__(
-        self, name, xml_state, root_dir, boot_dir=None, custom_args=None
-    ):
+    @staticmethod
+    def new(name, xml_state, root_dir, boot_dir=None, custom_args=None):
         if name == 'grub2':
             return BootLoaderConfigGrub2(
                 xml_state, root_dir, boot_dir, custom_args
--- a/kiwi/bootloader/install/__init__.py
+++ b/kiwi/bootloader/install/__init__.py
@@ -33,7 +33,8 @@ class BootLoaderInstall:
     :param object device_provider: instance of :class:`DeviceProvider`
     :param dict custom_args: custom arguments dictionary
     """
-    def __new__(self, name, root_dir, device_provider, custom_args=None):
+    @staticmethod
+    def new(name, root_dir, device_provider, custom_args=None):
         if name == 'grub2':
             return BootLoaderInstallGrub2(
                 root_dir, device_provider, custom_args
--- a/kiwi/builder/__init__.py
+++ b/kiwi/builder/__init__.py
@@ -33,7 +33,8 @@ class ImageBuilder:
     """
         image builder factory
     """
-    def __new__(self, xml_state, target_dir, root_dir, custom_args=None):
+    @staticmethod
+    def new(xml_state, target_dir, root_dir, custom_args=None):
         requested_image_type = xml_state.get_build_type_name()
         if requested_image_type in Defaults.get_filesystem_image_types():
             return FileSystemBuilder(
--- a/kiwi/builder/container.py
+++ b/kiwi/builder/container.py
@@ -114,7 +114,7 @@ class ContainerBuilder:
             log.info(
                 'Setting up %s container', self.requested_container_type
             )
-            container_setup = ContainerSetup(
+            container_setup = ContainerSetup.new(
                 self.requested_container_type, self.root_dir,
                 self.container_config
             )
--- a/kiwi/builder/container.py
+++ b/kiwi/builder/container.py
@@ -41,6 +41,7 @@ class ContainerBuilder:
     :param dict custom_args: Custom processing arguments defined as hash keys:
         * xz_options: string of XZ compression parameters
     """
+
     def __init__(self, xml_state, target_dir, root_dir, custom_args=None):
         self.custom_args = custom_args or {}
         self.root_dir = root_dir
@@ -130,7 +131,7 @@ class ContainerBuilder:
         log.info(
             '--> Creating container image'
         )
-        container_image = ContainerImage(
+        container_image = ContainerImage.new(
             self.requested_container_type, self.root_dir, self.container_config
         )
         self.filename = container_image.create(
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -323,7 +323,7 @@ class DiskBuilder:
                 'image_type':
                     self.xml_state.get_build_type_name()
             }
-            self.volume_manager = VolumeManager(
+            self.volume_manager = VolumeManager.new(
                 self.volume_manager_name, device_map,
                 self.root_dir + '/',
                 self.volumes,
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -536,7 +536,7 @@ class DiskBuilder:
         """
         if self.image_format:
             log.info('Creating %s Disk Format', self.image_format)
-            disk_format = DiskFormat(
+            disk_format = DiskFormat.new(
                 self.image_format, self.xml_state,
                 self.root_dir, self.target_dir
             )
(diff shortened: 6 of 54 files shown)


schaefi:
Why is using python's __new__ method no longer preferred to construct an instance of class ?
To my knowledge the __new__() method is always a static method of the class, even if no static method decorator is added.

Adding a staticmethod new() that needs to be called explicitly creates now two ways to construct an instance of class X by either a = X(...) or a = X.new(...) but both would behave differently, which seems odd to me for a Factory class

Can you explain why there was a static typing error before ?

Thanks

aplanas:
Yeah, I would stay with the `__new__` magic method here. This `.new()` in so many places can be a drawback. But I agree with Bo that this is not the usual use for the class constructor.

In any case, the convention is not to use `self` as a first parameter, but `cls`. At least I would change this.

schaefi:
> In any case, the convention is not to use self as a first parameter, but cls. At least I would change 
this.

Thanks, I would be fine with that change

isbm:
> Why is using python's `__new__()` method no longer preferred to construct an instance of class ?

Because it is not a class, but a factory where you supposed to use an interface for and return it instead, so then one knows what is type is returned. If you remember correctly, I suggested you to try MyPy and see that your code currently has about 500 errors and potential bugs, where you are mixing types. This is one of those, unfortunately. This pattern breaks the ability to navigate code properly and says "method not found" (because there is no method in the very class instance you're actually making).

> Adding a staticmethod new() that needs to be called explicitly creates now two ways to construct an instance of class X by either `a = X(...)` or `a = X.new(...)`

Sure. You should not use `X()` and block it by interfacing. You want factory pattern, and this is something else.

isbm:
> Yeah, I would stay with the `__new__` magic method here. This `.new()` in so many places can be a drawback. But I agree with Bo that this is not the usual use for the class constructor.
> 
> In any case, the convention is not to use `self` as a first parameter, but `cls`. At least I would change this.

@schaefi @aplanas  lol no, you guys are both wrong. :stuck_out_tongue: 

The class constructor consists of both `__new__` and `__init__`. So the `__new__` *is the usual* for the class _constructor_. Because the `__init__` is applied over the instance `__new__` is providing. And if `__new__` returns something else, then `__init__` is not called. 

However, we want factory instead. Current "factory" implementation is just hidden abusing the `__new__` and returning something else then the instance of self, so the `__init__` won't fire. That said, the method __new__ is called when instance is already created. This is only the method which will be called first then __init__ will be called to init the instance when you are creating one. So for example, application of ever touching `__new__` e.g. be useful to make a singleton or some instance customisation:

```python
from typing import Optional, Tuple, Any, Dict

class Foo:
    _instance:Optional["Foo"] = None
    
    def __new__(cls, *args: Tuple[Any], **kwargs: Dict[str, Any]) -> "Foo":
        if not cls._instance:
            cls._instance = Singleton.__new__(cls, *args, **kwargs)
        return cls._instance

assert id(Foo()) == id(Foo())
```

A proper way of Factory pattern (or one of those), what you actually want to do (including preventing `__init__`) is this:

```python
from abc import ABCMeta, abstractmethod

class iStuff:
    @abstractmethod
    def define(self) -> str:
        pass

class RealStuff(iStuff):
    def define(self) -> str:
        return self.__class__.__name__

class Factory(metaclass=ABCMeta):
    @abstractmethod
    def __init__(self) -> None:
        return None

    @staticmethod
    def new() -> iStuff:
        return RealStuff()
```

HTH.

aplanas:
>The class constructor consists of both `__new__` and `__init__`. So the `__new__` is the usual for the class constructor.

@isbm read better, because this is what I said : ))

> But I agree with Bo that this is not the usual use for the class constructor.

isbm:
> > The class constructor consists of both `__new__` and `__init__`. So the `__new__` is the usual for the class constructor.
> 
> @isbm read better, because this is what I said : ))

OK, but it is still not factory pattern anyway. To my defence... :)

schaefi:
ok, I still don't like introducing an extra new() method for a factory. It was not my understanding from oo design that a factory constructor is not eligible to use the same construction method than a class instance constructor do. Taking that into account and from what you have said before what would be wrong with the following suggestion:

```python
from abc import ABCMeta, abstractmethod
  
class FileSystemBase(metaclass=ABCMeta):
    @abstractmethod
    def __init__(self) -> None:
        return None

    def ok(self) -> None:
        print('OK')


class FileSystemExt2(FileSystemBase):
    def __init__(self) -> FileSystemBase:
        self.ok()


class FileSystemXFS(FileSystemBase):
    def __init__(self) -> FileSystemBase:
        self.ok()


class FileSystem(metaclass=ABCMeta):
    @abstractmethod
    def __init__(self) -> None:
        return None

    @staticmethod
    def __new__(cls, name: str) -> FileSystemBase:
        try:
            filesystem_object = globals()['{0}{1}'.format(cls.__name__, name)]
            return filesystem_object()
        except KeyError:
            print('No FileSystem handler for {0}'.format(name))


a = FileSystem('Ext2')
print(a.__class__.__name__)

b = FileSystem('foo')
print(b.__class__.__name__)

c = FileSystemXFS()
print(c.__class__.__name__)
```

isbm:
@schaefi that would work too, but that's not any factory pattern but just interfaces. :wink: The current problem you do have is that you are abusing class _instantiation_ to let it return non-interfaced object, and so then static compiler complains that you returning a spaghetti, so certain methods are not found.

Surely, you can fix your problem with the interfaces above too.

schaefi:
> are abusing class instantiation to let it return non-interfaced object, 

yes for static compilers a nightmare I agree. However I thought on not strong typed languages this is not an issue. I would have not coded a factory like this in C or C++ but how does it harm on python ? I just want to understand the negative effects and I'm surely willing to change the code into a new() method, but I probably miss the point other than a convention. All factory classes in kiwi are connected to Interface classes which is also a convention we use to ensure the "returning spaghetti" always has the same interface.

Thanks

isbm:
@schaefi finally start using MyPy: a must thing to every Python project. Then you will see. :wink:

schaefi:
"Optional static typing for Python" mypy, now I understand where all this comes from. I'm not sure if bumping over strong type rules on a language that has this optional is curse or blessing ;)

schaefi:
I'll take a look and your branch as the base to check base on this. Not in this week though, as we have the public cloud workshop, virtually of course ;)  ...didn't found a solution for virtual beer though :)

isbm:
> I'm not sure if bumping over strong type rules on a language that has this optional is curse or blessing ;)
If I tell you GvR is doing it, then it sounds like a blessing or a curse?

aplanas:
> I'll take a look and your branch as the base to check base on this. Not in this week though, as we have the public cloud workshop, virtually of course ;) ...didn't found a solution for virtual beer though :)

My unsolicited opinion: do not use MyPy for Kiwi. Type annotations in dynamic languages like Python can make sense as a contract to allow collaboration in big code bases, like Dropbox. But even they (the precursors of this), dropped this approach.

There is (IMHO) more value in constantly re-factorize the code, and more comprehensive tests.

Anyways :runner:

isbm:
@aplanas that's a bollocks, sorry. A proper answer is here:

https://mypy.readthedocs.io/en/stable/faq.html

Making maintainable and clear code is what Markus is passionate about. And misleading here is basically a crime.

The only downside of mypy is that shoddy programmers do not want to push outside their comfortable zone and having a half of a thousand errors in a small project might be a bit depressing (about 450, so to be precise). But if you would ever seriously using it, you would understand how much value it brings.

schaefi:
Hi guys, I see there are strong feelings about this topic :wink: 

There is respect for anything you both have said. I thought about the content and how to continue and like to
discuss how we can go from there. So here is what matters to me:

* Full freedom to refactor and improve. unit and integration testing leads to this and we definitely continue with this
* Maintainable and clear code. Yes 100%. Code review and other peoples contributions and opinions leads to that
* Stable API. At this point I think kiwi can do much better. The public part of the API could be more clear and also
  maintainable if we would put on type hints and annotations as @isbm suggested. However I don't like to drive the change in this area by mypy results leading and pushing us through it. That it showed you many issues regarding static type rules is not surprising as we didn't payed much attention to this part so far.

Therefore I suggest a step by step improvement based on the public interface. Let's start with all the factory classes as they also seem to have a behavior issue. Making the other API classes clear in their type and use should be the next step. Finally a run over mypy should show way less issues and can be worked on better than today.

I'll make a start with adapting the FileSystem factory in an extra PR and it would be great if you can have a look.
On agreement the move of the other factory classes should be straight forward and if we can share some work here
I'm happy.

Overall it will take some time to sort all this out and I like to avoid that any of those impacts the functionality of kiwi.
That having said we plan to put this on our roadmap for the next three months on a moderate priority.

@isbm @aplanas how about the plan ? thoughts, flamewars, hopefully not :smiley:

isbm:
LOL, @schaefi  I did not wanted to start 100500 comments :wink: Well, since I am strongly opinionated towards the static typing (where Python sucks, to be honest and where GvR and Google are trying to catch-up with PEP484 and MyPy), I would also strongly suggest to start using MyPy. Even it is small (relatively) it still should be clean and maintainable; without stuff like:

```python
my_great_something = "some string"

....
# Lots of code
...

my_great_something = ["troll", "face"]

```

As long one tries it, sees ~500 errors -- plenty of "refactorizing", as @aplanas expressed himself, opportunities! :wink:

schaefi:
> As long one tries MyPy, sees ~500 errors, then it renders plenty of "refactorizing", as @aplanas expressed himself, opportunities!

opportunities yes ;) so let's continue to reduce that number by focusing on the public api first. I made a start with the FileSystem refactor and think that goes into the right direction (at least no complains so far :)) We have this on our list for the next and the following quarter. There will be PRs popping up and we appreciate your review from time to time.
 
I'm now closing this on behalf of #1495. @isbm forgive me if this is not touching all of the code you wanted to change but I'd like to give all the static type adaptions a clear structure prior to "just" making mypy happy. Hope that works for everybody which I think it should ;)

Thanks everybody

schaefi:
I forgot to mention. To track the effort on factories and type hints issue #1498 will be used

schaefi:
Once we are done with this the next portion of the code can follow

PR #1487, opened 2020-06-17
Add disk.sh script payload

isbm:
Doing everything in `disk.sh` is not that perfect. For example, I would like to call, say, Ansible module locally. Or Ansible state. Or some Perl script. Or some binary in Go or Rust or C or whatever. I don't want to package it and install and then pinch. I want to just call it "as is".

This adds a value of keeping `disk.sh` merely a caller to more complex things, which can be setup without "scripting mess", say in declarative YAML etc.

This PR expects `payload` directory in the image description with whatever needs to be called on the target (in my case a set of rules and a binary modules for that) in order to maintain proper configuration management (for example). This content is then copied to `/tmp/payload` on the target image and then purged afterwards.

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -136,7 +136,7 @@ class SystemSetup:
         os.chroot("/")
         payload_dst: str = os.path.join(self.root_dir, "tmp", defaults.DISK_SYNC_DIR)
         if os.path.exists(payload_dst):
-            shutil.rmtree(os.path.join(self.root_dir, "tmp", defaults.DISK_SYNC_DIR), ignore_errors=False)
+            shutil.rmtree(payload_dst, ignore_errors=False)
 
     def import_repositories_marked_as_imageinclude(self):
         """
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -106,9 +106,11 @@ class SystemSetup:
         self._import_custom_archives()
         self._import_cdroot_archive()
 
-    def script_exists(self, name):
         """
         Check if provided script base name exists in the image description
+    def file_or_directory_exists(self, name):
+        """
+        Check if provided target base name exists in the image description
         """
         return os.path.exists(os.path.join(self.description_dir, name))
 
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -17,6 +17,7 @@
 #
 import glob
 import os
+import shutil
 import logging
 import copy
 from collections import OrderedDict
@@ -106,9 +107,15 @@ class SystemSetup:
         self._import_custom_archives()
         self._import_cdroot_archive()
 
+    def import_payload(self):
         """
-        Check if provided script base name exists in the image description
-    def file_or_directory_exists(self, name):
+        Import payload content.
+        """
+        if self.file_or_directory_exists(defaults.DISK_SYNC_DIR):
+            log.info("Importing disk.sh working payload")
+            payload_dst: str = os.path.join(self.root_dir, "tmp", defaults.DISK_SYNC_DIR)
+            shutil.copytree(os.path.join(self.description_dir, defaults.DISK_SYNC_DIR), payload_dst)
+
     def file_or_directory_exists(self, target: str) -> bool:
         """
         Check if provided target base name exists in the image description
@@ -126,6 +133,10 @@ class SystemSetup:
                 'rm', '-rf', '.kconfig', defaults.image_metadata_directory
             ]
         )
+        os.chroot("/")
+        payload_dst: str = os.path.join(self.root_dir, "tmp", defaults.DISK_SYNC_DIR)
+        if os.path.exists(payload_dst):
+            shutil.rmtree(os.path.join(self.root_dir, "tmp", defaults.DISK_SYNC_DIR), ignore_errors=False)
 
     def import_repositories_marked_as_imageinclude(self):
         """
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -71,6 +71,7 @@ class SystemSetup:
         image archives, overlay files
     :param str root_dir: root directory path name
     """
+
     def __init__(self, xml_state, root_dir):
         self.arch = Defaults.get_platform_name()
         self.xml_state = xml_state
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -109,10 +109,11 @@ class SystemSetup:
         """
         Check if provided script base name exists in the image description
     def file_or_directory_exists(self, name):
+    def file_or_directory_exists(self, target: str) -> bool:
         """
         Check if provided target base name exists in the image description
         """
-        return os.path.exists(os.path.join(self.description_dir, name))
+        return os.path.exists(os.path.join(self.description_dir, target))
 
     def cleanup(self):
         """
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -463,9 +463,7 @@ class DiskBuilder:
             )
 
         # run post sync script hook
-        if self.system_setup.script_exists(
-            defaults.post_disk_sync_script_name
-        ):
+        if self.system_setup.file_or_directory_exists(defaults.post_disk_sync_script_name):
             disk_system = SystemSetup(
                 self.xml_state, self.system.get_mountpoint()
             )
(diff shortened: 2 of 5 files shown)


isbm:
@schaefi those tests I will look afterwards in both PRs as they are merely need just tuning.

kiwi/system/setup.py
```diff
@@ -109,10 +109,11 @@ def import_description(self):
         """
         Check if provided script base name exists in the image description
     def file_or_directory_exists(self, name):
```
froh: should be removed, no?

schaefi:
>Doing everything in disk.sh is not that perfect. For example, I would like to call, say, Ansible module locally. Or Ansible state. Or some Perl script. Or some binary in Go or Rust or C or whatever. I don't want to package it and install and then pinch. I want to just call it "as is".

I understand your use case. I also still think the need for disk.sh at all is a special situation for doing let's say "special" things

> This adds a value of keeping disk.sh merely a caller to more complex things, which can be setup without "scripting mess", say in declarative YAML etc.

From an implementation point of view for disk.sh and what it should do I understand your reasons

> This PR expects payload directory in the image description 

At this point I have to admit that I'm not much in favor of adding yet another directory managed by kiwi itself in terms of code to allow "special" things.

You said you don't want to package the actual code called by dist.sh and uninstall the package after it has done it's job. Ok that's a fair point but a package would be a clean solution wouldn't it ?

Given you don't want a package you also have the following choices:

* pack it into a tarball and reference that tarball with
   ```xml
   <archive name="my-funny-disk-extension-code.tgz"/>
   ```

* add them as overlay files in your image description
   ```
   root/payload/...my-funny-disk-extension-code
   ```
   Typically an image description is stored in a git. That way you manage files as files and not as a tarball. So this would be the way I would go. In disk.sh at the end there would be an ```rm -rf /payload``` which looks not too cumbersome to me

To summarize this I think there are already plenty of opportunities to structure your disk extension code nicely and I'm not too enthusiastic to add python code in the builder to achieve this.

Thoughts ?

davidcassany:
I agree with @schaefi in the sense that all this is already possible by using the `disk.sh` together with packages or binaries that are not necessarily part of any package.

Feels like just adding alternate ways of managing specific and special hooks. You could even have the `disk.sh` inside the `payload` directory and it would be almost the same as we have now.

IMHO the idea of calling a hook script (like disk.sh, or config.sh) is already quite simple and still very flexible. Adding alternate procedures to avoid few lines of scripting feels like setting a process that can potentially not suit to everyone and end up having requests to adjust it because of A and B.

isbm:
> Feels like just adding alternate ways of managing specific and special hooks. You could even have the `disk.sh` inside the `payload` directory and it would be almost the same as we have now.

Maybe even do that instead (actually better). The thing is that the image description != configuration of it, but right now it is all static. Overlay might be the way of injecting the payload, but it feels/looks like a hacky way of abusing its very purpose: to actually _overlay_ the OS artifacts, keeping the overlay in Git as a part of image _description_ rather then also reusing it for _confinguration_.

schaefi:
>@schaefi Of course not. That's the point of a proper configuration management system not to be packaged and be as static as possible:

ok got you, you don't want packages. I guess it's the different worlds we are working on. Remember at suse and the distro everything even must be a package. So forgive me when I still think in packages. At least the initial rootfs still comes from a collection of packages. Thus I thought your disk.sh extension code somehow belongs to that source

isbm:
> ok got you, you don't want packages. I guess it's the different worlds we are working in. Remember at suse and the distro everything even must be a package. So forgive me when I still think in packages. At least the initial rootfs still comes from a collection of packages. Thus I thought your disk.sh extension code somehow belongs to that source

@schaefi actually even in SUSE world you do _not_ want to have packages for cfgmgmt. Right now you have Salt, but it more bites you than helps. This is the reason why modules in Ansible are transient and all the data is volatile: you install Ansible as a system, but each call has no packages on the target. Salt even wants now to go "encapsulated" mode where Python interpreter is wrapped into one single binary. While I won't comment on that, the principle is here: packaging that makes it a part of a system. Embedded wants it so minimal that even `/bin/ls` should not be there (because you know your system!). And in servers like typical your world right now: packaged cfgmgmt clients becomes a part of the system cycle update, which is not a good thing at the end.

schaefi:
> I cannot apply the same configuration to a different descriptions and vice versa.

why is this not possible with a disk.sh script hook that basically allows you to do everything ?
You could for example pull from a git to get things in if you think managing them as overlay files is conceptually wrong.

It would be interesting to understand how you actually build images. I think I have a misunderstanding here and it sounds odd and confusing to me

* you build from a collection of packages but the project specific extension code should not be in a package. So you also delete all of the package stack since it seems that's not wanted. ok. That raises the question where does the extension belong to. Not in a package I got that, but also not as part of the kiwi image description directory. And that I don't understand. It's disk.sh code that seems custom to the image you build, so it belongs to the image description. I understand if you don't want to duplicate that code and manage it properly. So why not put that into a git and let disk.sh fetch it ?

Something is odd

isbm:
> * you build from a collection of packages but the project specific extension code should not be in a package. So you also delete all of the package stack since it seems that's not wanted. ok. That raises the question where does the extension belong to. Not in a package I got that, but also not as part of the kiwi image description directory. And that I don't understand. It's disk.sh code that seems custom to the image you build, so it belongs to the image description. I understand if you don't want to duplicate that code and manage it properly. So why not put that into a git and let disk.sh fetch it ?

You are almost there. :wink: So. When you are building an image, you must think that after it is baked into a raw, it is read only and no more changes ever is possible anymore (not always the case, but let's focus on this). Think you just flashing it to a device, pour an epoxy on the PCB and turn it into a plastic brick with the sticking wires outside, OK? So all configuration mgmt should be done during the `disk.sh` before. Correct me if I am wrong, but I diffing between _image description_ (which defines my OS) and _OS configuration_, which is not a part "how to build OS", but is part of "how to configure OS". That said, things in overlay area (files, are overwritten on top at the end) is a part of "how to build OS". And thus after this is done, I want to have a _temporary_ data that comes from the concept of "how to configure OS" (it may be in a different Git repo even, so to speak).

Gets it more clear? We could also move this to IRC so not making huge noise here. I clearly see that my idea overlaps here with existing, but I am seeking to extend Kiwi, not to change it. Your conceptual help would be highly appreciated.

schaefi:
Thanks for the details. Ok this goes into the direction of "immutable systems".

We can discuss this further but I'm afraid I have to get some other work done, prepare for the public cloud workshop next week and am running a little bit out of time. 

Just one point which I think is crucial for the followup conversation.

> Correct me if I am wrong, but I diffing between image description (which defines my OS) and OS configuration, which is not a part "how to build OS", but is part of "how to configure OS". That said, things in overlay area (files, are overwritten on top at the end) are part of "how to build OS". 

At this point kiwi has no clear distinction if you overlay files that are configuration or if you overlay files that describes the OS. People use this in mixed ways.

I agree it might be worth thinking about it and if we come to the point that it would be important for kiwi itself to have a clear distinction between "defines my OS" and "configure my system" in terms of the use case of the overlay directory, then it makes sense to keep those two different.

No matter what we might add here, I think you will not be able to prevent people from using the overlay directory as they see fit. In the end this is handled pretty dumb by kiwi. "Everything in there will be part of the system".

> Gets it more clear? We could also move this to IRC so not making huge noise here. 

I think I got your point, and yes it's better to move the discussion to the channel as I also think other people have an opinion in this regard. That will be a fun conversation as everybody has different use cases ;)

isbm:
> Ok this goes into the direction of "immutable systems".

No, not really. You still have overlays etc. This is purely "keep things segregated". Maybe it is better to do even this layout division:

- Entities, that are part of permanent description.
- Configuration
- Volatile, transient data

So:
- Description:
  - `config.xml`
  - `root/`
- Configuration:
  - `config.sh`
  - `disk.sh`
  - (other `.sh`)
  - `some-playbook.yaml`
- Volatile, transient data:
  - pile of executable modules in any form, coming from outside
  - random tools, unrelated to the very image itself, but they do their usual job equally anywhere

isbm:
> That will be a fun conversation as everybody has different use cases ;)

That is precisely why I [wrote this](https://github.com/OSInside/kiwi/pull/1482#issuecomment-646127614).

schaefi:
I moved the essentials of this to an issue open for a discussion at #1497

PR #1492, opened 2020-06-24
Update documentation in content and structure

### LINKED ISSUE #1421: port and update xml description docs from legacy kiwi

schaefi:
This commit adds a documentation for the KIWI XML description. In contrast to the former auto generated code from the XML
schema this document can now finally be used as a reference. Along with that new chapter all auto generated and static html content as been deleted. Also all helper scripts around the proprietary oxygen tool and our schema doc generator has been deleted. Auto generating this information does not lead to a reference guide people can really work with. As a consequence to these changes this commit also includes some changes of the structure such that no information written by other people in the past gets lost. This Fixes #1421 and Fixes #1474


schaefi:
Hi All, this is a huge change that cannot be reviewed by reading the diff. I suggest you checkout the branch
and call:

```
tox -e doc
```

Then move your browser to doc/build/html/index.html

and take a look at the following:

a) Overall structure of the menu, makes sense ?
b) Read the new chapter: Image Description
c) quickly look through "Building Images for Supported Types" and "Working with Images"

Thanks a ton

tomschr:
Thank you very much @schaefi for this huge effort! 

So here are my observations:

---

Regarding the section "Image Description":

* Should the tags be sorted alphabetically?
  I know it's tempting to describe it hierarchically. :wink:  However, this is a reference. If someone is searching for an element, the user doesn't have the hierarchy in mind. Just the name of the element. As such, it's easier to find it when the sections are sorted alphabetically.

* The (hierarchical) structure
  I think, there is no right or wrong. IMHO we could use both, a hierarchical structure to learn and group things together, and a more alphabetically sorted flat structure for quick reference.
  Maybe it could help to "group" your elements into building blocks? Something like the [HTML Reference at Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference) did?

* Could we "formalize" the sections more?
  As a reference, it contains a list of all possible elements (and attributes) which are allowed by the KIWI schema. IMHO, such a reference should have a clear structure. For example, a reference for a schema should contain:
  * content model
  * attributes (required and optional)
  * additional constraints
  * description

  As an example, similar what the DocBook [`<book>`](https://tdg.docbook.org/tdg/5.1/book.html) reference element shows. Of course, it doesn't have to be identical, it's just an example. I know this is probably way more work. Maybe the initial work can be done with some tool and then edited manually.

* The "examples"
  Each section has an "example" which describes the element. However, the example is used with two functions in mind: to describe the "content model" and to show an example. I'm not sure if that really works. I would separate these two functions, because they act as distinct roles.

So just to sketch up this idea:

```
Section: `foo` element

One line about the purpose of the element

# Content model

  foo :=  (bar | baz)

# Attributes

   zoo (required): does more zoo in foo.

# Description

  add here more explanations

# Example

  add here your (typical) example

# Related

   Look also for ...
```

---

I haven't looked up the other sections you've mentioned. Maybe I will comment later. :wink:

schaefi:
@tomschr thanks for your thoughts. I tried to give it a bit more structure but I also think it will be even more work to get it done perfect. Especially on the (typical) examples this is tough because for an overview with a meaning usually several elements are playing together. This is also the reason why the documentation for e.g the bootloader Element only contains a reference to an extra chapter that only handles this part connected to a typical/real example.

I'd like to go with that structure for now as it otherwise ends up in rewriting all the docs for kiwi. We can't afford that and actually I'm glad that some sort of XML reference that makes sense now is present. Thus I suggest to go with this as a first shot and then refactor based on readers feedback.

The result for the suse doc by the way can be built with:

```
tox -e doc_suse
```

That is the result I'd like to publish. So I hope it fulfills the requirements and quality standards of the doc team and that's my first prio since we are lacking kiwi docs on documentation.suse.com for this version of kiwi completely.

Thoughts ?

davidcassany:
Few minor comments from my side:

1. I would devide the `Image Description` section in two, or divide its contents in two. I find it confusing having the explanation of the main image elements together right along with the list of xml elements available in the schema.

    So I would have expected a section division between the [Image Namespace](https://github.com/OSInside/kiwi/blob/xml_description_doc_update/doc/source/image_description.rst#image-namespace) and [\<image\>](https://github.com/OSInside/kiwi/blob/xml_description_doc_update/doc/source/image_description.rst#image). The first part of the image descriptions page explains in text the purpose or motivation of main elements, then without any advise, starts the list of available elements. IMHO there should be a distinction between these two parts.

    I think that @tomschr (if I got it right) was also pointing it when he suggested the structure:

```
Image Description
  +--- Elements Grouped by Categories
  +--- Element Reference
        +--- archive
        +--- ignore
```

2. About the `Build a Virtual Disk Image` and `Build an OEM Expandable Disk Image` sections I was wondering how would we reorganize that if we actually merge `vmx` and `oem` types. I know this is sidetracking a bit in this PR, I am also bringing this topic because some of the details of the schema described in `Build a Virtual Disk Image` can also apply for OEM images and by reading the docs you don't actually get this impression IMHO. I don't think this is something that necessarily needs to be reflected in this PR as might not have much sense staring to refactor this part if we foresee oem and vmx merge at midterm. Just raising it in case you have a different opinion.

3. A couple of very minor typos: 
  * there is a typo in `luks="passphras":` I guess there is a missing `e`
  * I believe the source path for OBS `obsrepositories://` is not correct, isn't it `obsrepositories:/`?

Finally, as an overall comment, I really like having the schema documentation as it is in this PR. I like having it as a not terribly long list of the major elements and then move the specifics to dedicated chapters (as it is done for the `oemconfig` for instance). I like this concept makes the documentation lighter and easier to navigate through IMHO.

schaefi:
> About the Build a Virtual Disk Image and Build an OEM Expandable Disk Image

I would touch this when we actually perform the change. I think code and doc changes should go in line here

schaefi:
@davidcassany thanks for reading it, I also fixed the typos :)

tomschr:
@schaefi thanks for the huge efforts! :+1: 

Yes, I'm aware, my suggestions were a "big" and I feel sorry that I put so much work on your plate. I agree, we can improve this part over time. It doesn't need to be perfect with that PR.

David also had some good points which I agree with.

schaefi:
> David also had some good points which I agree with.

ok I think I addressed them. My suggestion is to merge this one, wait until it's online and have a quick look. Then you @tomschr or someone from the doc team decides is this is good enough to be a doc on documentation.suse.com. If yes I'd like to get this there and for this process I need help which I would like to arrange a short meeting with you Tom and/or Frank or whomever would be the right person to talk to.

Does this makes sense ?

tomschr:
@schaefi  Thanks for all your hard work! :+1: 

> My suggestion is to merge this one, wait until it's online and have a quick look. 

I think, we could merge it for the time being. We can improve it over time, if there things need to be changed.

> decides [if] is this is good enough to be a doc on documentation.suse.com. If yes I'd like to get this there and for this process I need help

I think it's already good enough. :-)

Stefan created already a merge request on GitLab regarding the configuration of documentation.suse.com. When this is merged, the KIWI doc would be available under the existing SLE documentation. It would also be possible to create a separate page, if you prefer that.


> which I would like to arrange a short meeting with you Tom and/or Frank or whomever would be the right person to talk to.

Just talked with Frank. Currently, we are in some virtual doc team meetings, but next week would probably helpful. We could arrange that by private mail.

Thanks!

schaefi:
> I think it's already good enough. :-)

:) thanks, it's a start there are room for many improvements. 

> Just talked with Frank. 

Great, I saw your e-mail and will respond next.

Ok so I'll merge this one now...

PR #1493, opened 2020-06-25
Fixes live ISOs

### LINKED ISSUE #1489: Different blkid output on Arch Linux and Debian causes ISO to be unbootable
Referencing previous issue:
https://github.com/OSInside/kiwi-descriptions/issues/69

## Problem description
blkid on debian and arch linux spits out different output than what is expected.
Therefore the uuid string is formatted wrong, which leads to the image being unbootable.
## Expected behaviour
Boot
Create any arch linux or debian ISO.
+ KIWI version: KIWI (next generation) version 9.21.2
+ Operating system: Arch Linux
+ OBS version: ?

davidcassany:
This commit fixes iso images. Due to a change introduced in c7ed1cf
live ISOs were no longer booting as the rootfs.img filesystem was
copied to the squashfs container while being still mounted. Because of
that, at boot time, it refused to mount.

This commit adds umount method for the filesystem base class, so it
can be manually umounted before deleting the instance.

Fixes #1489

### DIFF
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -241,6 +241,8 @@ class LiveImageBuilder:
         live_filesystem.sync_data(
             Defaults.get_exclude_list_for_root_data_sync()
         )
+        live_filesystem.umount()
+
         log.info('--> Creating squashfs container for root image')
         self.live_container_dir = mkdtemp(
             prefix='live-container.', dir=self.target_dir
--- a/kiwi/filesystem/base.py
+++ b/kiwi/filesystem/base.py
@@ -164,6 +164,15 @@ class FileSystemBase:
             exclude=exclude
         )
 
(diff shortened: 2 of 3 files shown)


davidcassany:
I could also just delete the instance `del live_filesystem`, but I thought this was not explicit enough.

schaefi:
Thanks for the quick fix. There is also a bug for this here:

https://bugzilla.suse.com/show_bug.cgi?id=1173356

Can you add that in the commit message ?

Thanks

davidcassany:
> Thanks for the quick fix. There is also a bug for this here:
> 
> https://bugzilla.suse.com/show_bug.cgi?id=1173356
> 
> Can you add that in the commit message ?
> 
> Thanks

Sure, I just added it

PR #1499, opened 2020-07-02
Describe targets in tox.ini

tomschr:
No related issue.

* Make use of description to show them with "tox -av"
* Use comments as "description" lines
* Define minimal tox version

I think, the changes are minimal and non-destructive. :wink: With the above changes, you can run `tox` like this:

```
$ tox -av
using tox.ini: ... (pid 14991)
using tox-3.14.5 from /usr/lib/python3.6/site-packages/tox/__init__.py (pid 14991)
default environments:
check        -> Source code quality/integrity check
unit_py3_7   -> Unit Test run with basepython set to python3.7
unit_py3_6   -> Unit Test run with basepython set to python3.6
packagedoc   -> Documentation build run suitable for doc deployment in package(rpm)
devel        -> Test KIWI

additional environments:
doc          -> Documentation build suitable for local review
doc_travis   -> Documentation build suitable for doc deployment in travis env
doc_suse     -> Documentation build suitable for SUSE documentation
doc.html     -> Documentation build html result
doc.xml      -> Documentation build xml result
doc.latexpdf -> Documentation build PDF result
doc.man      -> Documentation build man pages
```

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,6 +11,7 @@
 #
 
 [tox]
+minversion = 3.5.0
 skip_missing_interpreters = True
 skipsdist = True
 envlist =
@@ -22,6 +23,9 @@ envlist =
 
 
 [testenv]
+description =
+    {unit_py3_6,unit_py3_7}: Unit Test run with basepython set to {basepython}
+    devel: Test KIWI
 whitelist_externals = *
 basepython =
     {check,devel,packagedoc,doc,doc_travis,doc_suse}: python3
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -11,6 +11,7 @@
 #
 
 [tox]
+minversion = 3.4.0
```
schaefi, line 14: minversion should be 3.5.0 as we use type hints and annotations which do not exists on py3.4

tomschr: Oh right. Good catch! I've corrected the line now.

Issue #1451, opened 2020-05-22
pip and wheels causing trouble


schaefi:
## Problem description

If you look at https://stackoverflow.com/a/46087545 it turns out recent versions of
pip (v >= 9.0.1) will try to compile a wheel even from a source distribution. This is a problem
for the way kiwi uses setup.py. In kiwi setup.py contains instructions to run code that e.g
installs man pages or the bash completion. If pip turns data into a static zip file (nothing else is a wheel) all dynamics are gone. Meaning the wheel gets created implicitly in the background and used. Using a wheel means it simply unpacks data but does not run any code. This means
no call of ```setup.py install``` will be issued. This leads to files that are expected to be installed into the system to be unpacked in the respective ```site-packages/``` directory which is wrong in the context of kiwi.

To avoid this people has to install kiwi with:

```
pip install kiwi-ng --no-binary :all:
```

Nobody will do this or remember that it is needed.

The following files are affected by the way pip works when installing kiwi:

* man pages: /usr/share/man/man8/...
* command completion: /etc/bash_completion.d/kiwi-ng.sh
* kiwi default config file: /etc/kiwi.yml
* package docs: /usr/share/doc/packages/kiwi-ng/...
     (kiwi.pdf, LICENSE, README)

kiwi stays functional without those but we should cleanup and come to a conclusion how we want to refactor setup.py in a way it seems to expected by the python community

tomschr:
If you apply the following patch, wheel can be built (however, you need to build the manpages first):


```diff
diff --git i/setup.py w/setup.py
index 50a1b2136..eb370c4e1 100755
--- i/setup.py
+++ w/setup.py
@@ -3,6 +3,7 @@
 
 from setuptools import setup
 from setuptools.command import sdist as setuptools_sdist
+from setuptools.command import install as setuptools_install
 
 from distutils.command import build as distutils_build
 from distutils.command import install as distutils_install
@@ -122,7 +123,7 @@ class clean(distutils_clean.clean):
         distutils_clean.clean.run(self)
 
 
-class install(distutils_install.install):
+class install(setuptools_install.install):
     """
     Custom install command
     Host requirements: make
```

schaefi:
Maybe also worth a look how other projects does it:

https://pyscaffold.org/
https://pypi.org/project/setuptools-scm/

They use setup.cfg to describe the project

schaefi:
Fixed by #1502

PR #1502, opened 2020-07-02
Cleanup setup py

schaefi:
The way kiwi uses setup.py assumes that pip runs this script  like a spec file in rpm is processed. However this is not the  case given that pip implicitly creates a static zip file called  wheel which looses all the code logic done in setup.py. Therefore  setup.py should not contain code that needs to run at install  time. Of course this change comes with an effect which is that  the following files will not be available when installing kiwi  from pip:
    
* man pages: /usr/share/man/man8/...
* command completion: /etc/bash_completion.d/kiwi-ng.sh
* kiwi default config file: /etc/kiwi.yml
* package docs: /usr/share/doc/packages/kiwi-ng/... (kiwi.pdf, LICENSE, README)
    
kiwi stays fully functional without this information. It is  expected that the installation of kiwi as a service will be done by a package and its package manager. When using kiwi  from pip it is designed to provide a python module but not  a complete user application. The way pip and wheels interact  with each other seems to demonstrate that pip is not a  package manager but more a python module manager.

This Fixes #1415


schaefi:
@tomschr while coding this part here I found the the minversion setup references the tox version not the python version :)

tools/dcounter.c
```diff
@@ -22,7 +22,7 @@ int main(int argc, char **argv) {
     }
     if (argc == 2 && strstr(*argv, "-l")) {
         newline = 1;
-        strncpy (prefix,*(argv+1),strlen(*(argv+1)));
+        strncpy (prefix, *(argv+1), strlen(prefix));
```
davidcassany, line 25: I haven't deeply looked the code here, but I guess this change is fine because source and destination strings are expected the have the same size. In any case now looks good.

tox.ini
```diff
@@ -11,7 +11,7 @@
 #
 
 [tox]
-minversion = 3.5.0
+minversion = 2.9.1
```
davidcassany, line 14: :+1: Great I faced that yesterday and did not dig into it yet.

davidcassany:
@schaefi I just noted that the issue #1451  was not closed after merging this, I assume the `Fixes #1415` of the commit is just a typo and it should have been `#1451`. I don't think it is worth amending the commit, so I'd say we'll have to live with it. I didn't notice the referenced issue was not the correct one at the time I did the review, sorry.

schaefi:
Ah my bad, yes a typo. I'm closing the issue manually and we just live with the mistake in the log

Thanks

PR #1504, opened 2020-07-03
Document output files KIWI produces

davidcassany:
This commits adds a chapter to describe the ouput files that are part
of any image build.

### DIFF
--- a/doc/source/building_images.rst
+++ b/doc/source/building_images.rst
@@ -54,6 +54,131 @@ KIS Root File System Image
   The use case for this component image type is highly customizable.
   Many different deployment strategies are possible.
 
+
+Image Results
+-------------
+
+{kiwi} execution results in an appliance image after a successful run of
+:ref:`kiwi_system_build` or :ref:`kiwi_system_create` command.
+The result is the image binary in addition a couple of metadata files that can
+be handy to describe or identify the resulting image. The output files follow
+this naming convention:
+
+`<image-name>.\<arch\>-\<version\>.\<extension\>`
+
+where `<image-name>` is the name stated in the :ref:`image-description` as an
+attribute of the :ref:`sec.image` element. The `<arch>` is the CPU
+architecture used for the build, `<version>` is the image version defined in
+:ref:`\<version\><sec.preferences>` element of the image description
+and the `<extension>` is dependent on the image type and its definition.
+
+Any {kiwi} appliance build results in, at least, the following output files:
+
+
+1. The image binary, `<image-name>.\<arch\>-\<version\>.\<image-extension\>`:
+
+   This is the file containig the actual image binary, depending
+   on the image type and its definition it can be a virtual disk image
+   file, and ISO image, a tarball, etc.
+
+2. The `<image-name>.<arch>-<version>.packages` file:
+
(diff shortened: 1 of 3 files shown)


davidcassany:
@schaefi this is not exactly a documentation of all possible output files but I was in doubt about how to do it. I think this quite enough as different image types produce different binary types, so this is the effective result for any image build. I have not documented the `image-root.log` file for the  `system build` command as I do not necessarily consider it part of the result and I do no see how it would fit in this chapter, this also depends a lot on the specific command line.

I only considered the output files for image builds, so `system build` and `system create`, should we consider something else? The other subcommands I thought they produce mostly data dropped to stdout, but no actual files. Also not so sure about the `kiwi.result` should we document it here too? Probably yes but this is not part of the image IMHO, this is kind of build history.

doc/source/building_images.rst
```diff
@@ -54,6 +54,64 @@ KIS Root File System Image
...
+   This file is the output of a verification done by the package manager
+   against the package data base. More specific it is the output of
+   the :command:`rpm` verification process or :command:`dpkg` verification 
+   depending on the packaging technology selected for the image.
+   In both cases the output follows the RPM verification syntax. This
+   provides an overview of all packages status right before any boot of
+   the image.
+
+In addition to these three files there is the OEM type that can eventually
+return and additional ISO image or an additional tarball including artifacts
```
schaefi, line 111: typo and/an

schaefi:
Thanks for writing this up :+1: I like the chapter describing the general naming approach. This is great.

At the end you mentioned a specialty of the oem type. I think at this point it would be nice if we provide a list of the image types as mentioned in image_description/elements.rst and quickly explain what users can expect as output files. That allows you to specify the possible output files from the oem type. Also if we have this list we could add one sentence referencing it at the end of the image type list described in image_description/elements.rst

Thoughts ?

schaefi:
You can use the names in the prolog as example file names:

```
prolog_replacements = {
    '{exc_image_base_name}': 'LimeJeOS-Leap-15.1',
    '{exc_description}': 'suse-leap-15.1-JeOS',
    '{exc_netboot}': 'netboot/suse-leap15.1',
    '{exc_os_version}': '15.1',
    '{exc_image_version}': '1.15.1',
    '{exc_repo}': 'obs://openSUSE:Leap:15.1/standard',
    '{exc_kiwi_repo}':
        'obs://Virtualization:Appliances:Builder/openSUSE_Leap_15.1',
    '{schema_version}': '7.2',
    '{kiwi}': 'KIWI NG',
    '{kiwi-product}': 'KIWI Next Generation (KIWI NG)',
    '{kiwi-legacy}': 'KIWI Legacy'
}

```

see doc/source/conf.py

So you don't need to invent file names and the writing up the list should be straight forward

davidcassany:
>  I think at this point it would be nice if we provide a list of the image types as mentioned in image_description/elements.rst and quickly explain what users can expect as output files.

Yes, I like the idea. Let's do it

davidcassany:
> Just a few typos on first read

Fixed thanks!

davidcassany:
@schaefi what about now? I added a little sentence for each image type with some specific example

PR #1505, opened 2020-07-08
Do not create a md5 checksum output file for archive images

davidcassany:
This commit removes the creation of the md5 file that includes a
checksum of the image binary for the `tbz` image type. Removing it
as this is the only image type that includes it as part of the result
and because the bundle procedure already creates a sha256 file out of
the results, so there is still the chance to produce validation
checksums.

### DIFF
--- a/kiwi/builder/archive.py
+++ b/kiwi/builder/archive.py
@@ -21,7 +21,6 @@ import logging
 from kiwi.defaults import Defaults
 from kiwi.archive.tar import ArchiveTar
 from kiwi.system.setup import SystemSetup
-from kiwi.utils.checksum import Checksum
 from kiwi.system.result import Result
 from kiwi.runtime_config import RuntimeConfig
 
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi while having  a look to the output documentation I realized that the `image="tbz"` is the only image type that produces a md5 file of the output, none of the others does and I could not fine any specific reason for that. So I just assumed this is a leftover. I checked the lxc container images on openSUSE (which use this image type [here](https://build.opensuse.org/package/live_build_log/Virtualization:containers:images:openSUSE-Tumbleweed/opensuse-tumbleweed-image:lxc/container/x86_64)) and verified that the file is not included as part of the bundle and that the bundle already creates a checksum file in its process, as in all other image types. So, to me, it looks like this md5 file in OBS is just not being used at all.

This is a behavior change, I don't think this image type is widely used and even less the md5 file that it produces, however this is still a potentially breaking change. The creation of this md5 file does no hurt, it is just inconsistent compared with other image types. So even if we agree that this md5 file is just a leftover I am not that sure we should remove it.

Any thoughts?

PR #1508, opened 2020-07-09
WIP: Add and update integration tests

### LINKED ISSUE #1468: Update integration tests
## Problem description

During some fixes on APT and EFI setup for non SUSE distros I realized we have our integration tests that are not running against the lastest versions of the non SUSE distros we support.

In fact, we have no integration tests for CentOS 8, Ubuntu 20.04, Debian Buster and probably latest Fedora.

I'd say it makes sense to keep centos7 and add centos8. However probably there is no specific need to keep multiple versions for other distros, as the change or gap between versions is not that big. I'd just move on from Ubuntu bionic to ubuntu focal, use latest stable fedora available in OBS and add Debian buster.

davidcassany:
* Add CentOS v8 build test
* Upgrade Ubuntu Bionic -> Ubuntu Focal build test
* Upgrade Fedora 30 -> Fedora Rawhide build test
* Add Debian Buster build test

Fixes #1468


davidcassany:
@schaefi just adding starting the PR about the integration tests upgrade even this is not final. Doing so as I found some unexpected issues (beyond setting the OBS project) that might turn to be actual bugs or misbehavior of KIWI in some corner cases.

The images are building in this project [home:dcassany:integration_tests](https://build.opensuse.org/project/show/home:dcassany:integration_tests). They all build but I found few issues, mostly related to bootloader, listed here:

* CentOS7 UEFI boot for OEM, fails after dumping while kexec the deployed kernel
* CentOS8 UEFI boot on live ISO fails (I could not even see grub2 output)
* CentOS8 UEFI boot fails on install media ISO (I guess equivalent issue as before, both ISOs with templated grub2 config)
* CentOS8 VM does not boot on bios and & UEFI because the `root=<device>` kernel parameter is wrong
* Fedora Rawhide Live network interface is not up (minor issue)
* Fedora Rawhide installation media fails checksum validation on UEFI & BIOS (this is surprising error)
* Fedora Rawhide VMX wrong root=device in kernel command line UEFI & BIOS (I guess the same error as in CentOS8)

I tested six boots per distro (live EFI & BIOS, the installation media EFI & BIOS and VM EFI & BIOS). I tested runnig qemu-kvm directly with combinations on the following command:

```
qemu-kvm -m 2048 -hda <disk> -cdrom <iso> -serial stdio -bios <omvf firmware>
```

So my take here is that the patches we have on the grub2 configuration look like they are not working for latest Fedora and CentOS8 inside the OBS. In addition the checksum failure I see on Fedora Rawhide OEM image is puzzling.

davidcassany:
Once the issues listed here are kind a clarified I'd just set the VM build for Fedora to make use of btrfs and set few subvolumes without snapshots and see what happens from there.

schaefi:
Thanks much for this great effort :+1:

schaefi:
> few issues, mostly related to bootloader,

It's always the same, isn't it

davidcassany:
> > few issues, mostly related to bootloader,
> 
> It's always the same, isn't it

indeed :sweat_smile:

davidcassany:
@schaefi I am quite unsure about how to follow up on that. I updated the list of issues above with my latest findings, which aren't that many unfortunately....

Regarding CentOS and Fedora issues I am quite disappointed as don't even see the same issues in both, no overlap in failures I experiment and some point to miss-configured grub2 files... we are just applying the default grub2 tools and running a couple of patches for OBS builds, I don't feel like adding extra hacks on grub2 config... So I am thinking on splitting this PR into three, one for Debian/Ubuntu (those are just working fine), another one for CentOS 8 and finally another one for Fedora Rawhide, so they can be merged and treated independently.

I haven't tested any of these builds locally, probably would be nice to figure out to what extend OBS env is being problematic here.

schaefi:
>So I am thinking on splitting this PR into three, one for Debian/Ubuntu (those are just working fine), another one for CentOS 8 and finally another one for Fedora Rawhide, so they can be merged and treated independently.

I like that plan

I wonder that there are so many grub related issues. As you said we just call grub2 tools and expect them to work. I think the absence of udev and other core services in a buildservice worker could be one reason for the unexpected behavior. However I agree completely, I don't want to add more hacks in kiwi that massage the generated grub.cfg file. From your findings it sounds like the problems we already had and covered by the extra code in kiwi. Thus it seems like an issue that this extra code path is not triggered or not operational on the systems you have tested. In this case I think it's worth to make that code to work with the other distros too but we should avoid adding yet another extra fix-my-grub-config entry point in the kiwi code base

Thoughts ?

davidcassany:
Closing this requests since all the tests are already part of the integration tests

Issue #1509, opened 2020-07-12
XFS filesystem created with corrupt superblock


christiansetzer:
## Problem description
I moved from Centos7 as build server to Centos8, because of the repository structure. There is no real error that kiwi produces and the vhdfixed file for Azure is created. But when I try to deploy a VM I get the following error

> "details":[{"code":"Conflict","message":"{\r\n \"status\": \"Failed\",\r\n \"error\": {\r\n \"code\": \"ResourceDeploymentFailure\",\r\n \"message\": \"The resource operation completed with terminal provisioning state 'Failed'.\",\r\n \"details\": [\r\n {\r\n \"code\": \"OSProvisioningTimedOut\",\r\n \"message\": \"OS Provisioning for VM 'xxx-efi-vm' did not finish in the allotted time. The VM may still finish provisioning successfully. Please check provisioning state later. Also, make sure the image has been properly prepared (generalized).\\r\\n *

I also have to add, that I tried
- firmware=efi
- firmware=uefi
and the classic one. I thought that the error had something to do with me missing to set the hyper-v gen1 or gen2 setting. Both do not work

## Expected behaviour
At least deployment should finish. The created VM does not even boot up.

## Steps to reproduce the behaviour
[image-root.log](https://paste.ubuntu.com/p/wb6s7RwYGW/)
[config.xml](https://paste.ubuntu.com/p/SbMBB6Sgg9/)

After the image is created:
- azcopy for upload
- az disk create
- az image create
- az sig image-version create

The VM is deployed from a shared image gallery. After that the above error comes up

## OS and Software information
+ KIWI version: 9.21.0
+ Operating system: Centos 8 (latest updates as of creation date of issue)
+ OBS version:

schaefi:
I don't think this is a kiwi problem. We are building for azure with that kiwi version as follows:

```xml
<type image="vmx" filesystem="xfs" bootloader="grub2" bootloader_console="serial" boottimeout="1" kernelcmdline="earlyprintk=ttyS0 console=ttyS0 rootdelay=300 net.ifnames=0 dis_ucode_ldr scsi_mod.use_blk_mq=1" devicepersistency="by-uuid" formatoptions="force_size" format="vhd-fixed" bootpartition="true" bootpartsize="1024" efipartsize="512" firmware="efi">
    <size unit="M">30720</size>
</type>
```

There are many constraints Azure requires for their vhds. The above type setup covers all of them. Also the 30GB size is a must have.

[...]

PR #1510, opened 2020-07-13
Include CentOS vendor in grub2 configuration

davidcassany:
This commit ensures the vendor files for grubenv consider CentOS
vendor.

Related to https://github.com/OSInside/kiwi/pull/1508#issuecomment-656221457

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -759,6 +759,8 @@ class Defaults:
         """
         efi_vendor_directories = [
             'EFI/fedora',
+            'EFI/redhat',
+            'EFI/centos',
             'EFI/opensuse'
         ]
         for efi_vendor_directory in efi_vendor_directories:


davidcassany:
I'd say we should include RHEL here too, but I don't have any RHEL at hand for testing. However I am not sure if this would be something like:
```python
efi_vendor_directories = [
    'EFI/fedora',
    'EFI/redhat',
    'EFI/centos',
    'EFI/opensuse'
]
```

@Conan-Kudo do you have a clue about how it should be to include Redhat?

schaefi:
I wonder if it is related to #1509

davidcassany:
> I wonder if it is related to #1509

I was wondering the same, but I'd say not. Since this change is related to the grub patches that get into play while building inside OBS.

davidcassany:
Googling I found some references to `EFI/redhat/grubenv`, thus I assume `EFI/redhat` is the correct path here.

Conan-Kudo:
@martinezjavier Could you answer what the EFI directory is for Red Hat Enterprise Linux? I believe it is `redhat`, but I'd like to double check.

martinezjavier:
> @martinezjavier Could you answer what the EFI directory is for Red Hat Enterprise Linux? I believe it is `redhat`, but I'd like to double check.

That's correct, it is `redhat`.

PR #1512, opened 2020-07-14
Rename CentOS 7 test and add CentOS 8 test

davidcassany:
This commit renames the CentOS 7 test and adds a new CentOS 8 test

Related to #1468


davidcassany:
@schaefi so here are the issues I currently have for CentOS 8. Sorry for the late feedback I delayed this report as on Friday I realized I was missing to run some tests on a non OBS env (in terms of boot configuration this seams to be a very important aspect). So I prepared a CentOS 8 example test [here](https://github.com/OSInside/kiwi-descriptions/pull/71) with equivalent configurations of what I did for the integration tests.

So far the issues for the CentOS 8 images (image downloads [here](https://build.opensuse.org/package/show/home:dcassany:integration_tests/test-image-iso-oem-vmx-centosv8)) are:

* CentOS8 VM does not boot on bios if set in EFI/UEFI mode.

  In that case grub2 does not properly evaluate the runtime configuration of grub2 and it does not include any `root=UUID=` parameter even it is already properly set as `default_kernelopts` in `/boot/grub2/grub.cfg`. Interestingly an equivalent setup build on my non-EFI Leap 15.1 host works just fine on both, EFI and Bios. I could not find any relevant difference, on my host I also see equivalent values for `default_kernelopts` and they properly taken into consideration when evaluating the grub2 menu entry at boot time.

* CentOS8 Live ISO does not boot on EFI/UEFI, it does on bios fallback.

  I can't explain that too. Builds on my Leap host work just fine on EFI and Bios, however if the build is done in OBS it doesn't. I was comparing grub2 configuration between host build and obs build and they really looked equivalent to me :man_shrugging: 

* CentOS8 Live ISO booted as disk it fails using the persistent partition, the partition is created but never mounted. It fails with:
   ```
   [   12.000976] dracut-initqueue[407]: Syntax error in mke2fs config file (<default>, line #22)
   [   12.001073] dracut-initqueue[407]:   Unknown code prof 17
   [   12.003418] dracut-initqueue[407]: Warning: Failed to setup persistent write space !
   [   12.003475] dracut-initqueue[407]: Warning: Falling back to temporary overlay
   ```

   I supose this might be fixed by proper configuration on the mke2fs config file, but I have no idea about what to tune or change. I can consistently reproduce it in any build done with OBS or not.

* CentOS8 Install ISO for OEM boots on UEFI/EFI and bios. I couldn't see any meaningful difference compared with Live ISO, strange. If I do bios deployment it seams to be functional but any reboot on the deployed OS fails as described for the VMX images.

So all in all it looks like building inside OBS with this its special build environment causes some issues in terms of grub2 configuration, something I could not trace. All device references looked good to me and I could not see wrong values in grub config for OBS builds...

On the good side, an equivalent build on my host using description from the https://github.com/OSInside/kiwi-descriptions/pull/71 worked fine in all cases. To me the most relevant issue is that Live images for UEFI/EFI build in OBS using UEFI/EFI firmware are not booting. I would appreciate another pair of eyes here. As mentioned I tried to compare the results building on my host from https://github.com/OSInside/kiwi-descriptions/pull/71 and the ones build in OBS and I could not see anything relevant...

davidcassany:
According to the latest changes in #1516 The only issue that still prevails is that the disk images (OEM and VMX) build in OBS for CentOS 8 do not boot. I really have doubts this is caused at all by a KIWI issue as an equivalent build outside OBS works fine.

Moreover I just noted that the grub2 configuration for in OBS and outside OBS builds just differs on the `/boot/loader/entries/*.conf`. For OBS builds we get:

```
title CentOS Linux (4.18.0-193.6.3.el8_2.x86_64) 8 (Core)
version 4.18.0-193.6.3.el8_2.x86_64
linux /boot/vmlinuz-4.18.0-193.6.3.el8_2.x86_64
initrd /boot/initramfs-4.18.0-193.6.3.el8_2.x86_64.img $tuned_initrd
options $kernelopts $tuned_params
id centos-20200610111729-4.18.0-193.6.3.el8_2.x86_64
grub_users $grub_users
grub_arg --unrestricted
grub_class kernel
``` 
Note the `$tuned_params` and `$tuned_initrd` variables. However for the builds done outside OBS we get:
```
title CentOS Linux (4.18.0-193.6.3.el8_2.x86_64) 8 (Core)
version 4.18.0-193.6.3.el8_2.x86_64
linux /boot/vmlinuz-4.18.0-193.6.3.el8_2.x86_64
initrd /boot/initramfs-4.18.0-193.6.3.el8_2.x86_64.img
options $kernelopts
id centos-20200610111729-4.18.0-193.6.3.el8_2.x86_64
grub_users $grub_users
grub_arg --unrestricted
grub_class kernel
```
Without any `$tuned_*` variable. I noticed that removing the `$tuned_` references in `/boot/loader/entries/*.conf` fixes the boot. The interesting thing is that in both cases (OBS builds and non OBS builds) the ` $tune_*` variables are set and defined in `/boot/grub2/grub.cfg` in the same exact way, an empty string:
```
### BEGIN /etc/grub.d/00_tuned ###
set tuned_params=""
set tuned_initrd=""
### END /etc/grub.d/00_tuned ###
```

To me this looks like a bug in variables expansion within the context of the [BootLoaderSpec](https://www.freedesktop.org/wiki/Specifications/BootLoaderSpec).

@Conan-Kudo @martinezjavier does this ring any bell for you? The image having this issue can be found [here](https://build.opensuse.org/package/binaries/home:dcassany:integration_tests/test-image-iso-oem-vmx-centosv8:Live/images_CentOS8)

davidcassany:
I consider this task complete.

@schaefi the test looks good enough from my PoV for this first round. What you think?

schaefi:
yes looks good, let's go for it :+1:

PR #1514, opened 2020-07-15
Use rsync in inplace transfer mode

schaefi:
Using the --inplace option in rsync helps to save space on
syncing the rootfs data and prevents e.g OBS workers from
running out of VM space when transfering root filesystem
data. Also using --inplace allows to keep hardlinks intact.
This is related to bsc#1096738

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -223,6 +223,17 @@ class Defaults:
         )).lstrip(os.sep)
 
     @staticmethod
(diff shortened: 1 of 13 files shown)


Vogtinator:
> Also using --inplace allows to keep hardlinks intact.

Note that this is only relevant for changed files, but as the target is empty there aren't any files which changed content. So no difference there.

PR #1516, opened 2020-07-16
Fixed grub setup in EFI/BOOT directory

### LINKED ISSUE #1491: Improper grub.cfg in /EFI/BOOT/grub.cfg
## Problem description

kiwi copies the grub.cfg created by grub2-mkconfig also to EFI/BOOT/grub.cfg such that the grub efi binary can find it. However there is a report that this grub.cfg should only contain the code to load the proper (vendor specific) grub.cfg file from the linux boot space

kiwi does this for some years and it was not an issue. On ARM and the changes there it seems to be an issue now and this needs to be investigated

schaefi:
kiwi copied the same grub.cfg file as it exists in boot/grub2
to the efi path. This is wrong as the setup in the efi boot
directory is used to enable normal grub loading and not providing
the user grub configuration. In addition the changes here makes
sure that the early grub boot code is placed into the system
in any EFI case except for secure boot when shim-install is
present. If shim-install is present it also creates the early
grub boot setup such that kiwi doesn't have to do it.
This Fixes #1491 and Fixes bsc#1172908

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -144,15 +144,13 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         self.root_reference = None
         self.cmdline = None
         self.iso_boot = False
-        self.shim_fallback_setup = False
+        self.early_boot_script_efi = None
 
     def write(self):
         """
         Write bootloader configuration
 
         * writes grub.cfg template by KIWI if template system is used
-        * copies grub config file to alternative boot path for EFI systems
-          in fallback mode
         * creates an embedded fat efi image for EFI ISO boot
         """
         if self.config:
@@ -164,9 +162,6 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 config.write(self.config)
 
             if self.firmware.efi_mode():
-                self._copy_grub_config_to_efi_path(
-                    self.boot_dir, config_file
-                )
                 if self.iso_boot:
                     self._create_embedded_fat_efi_image()
 
@@ -308,9 +303,9 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                         with open(vendor_grubenv_file, 'w') as vendor_grubenv:
                             vendor_grubenv.write(grubenv)
 
-        if self.firmware.efi_mode():
+        if self.firmware.efi_mode() and self.early_boot_script_efi:
             self._copy_grub_config_to_efi_path(
-                self.efi_mount.mountpoint, config_file
+                self.efi_mount.mountpoint, self.early_boot_script_efi
             )
 
     def setup_install_image_config(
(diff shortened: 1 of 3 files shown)


schaefi:
@davidcassany I have tested this change with efi and uefi setup for disk and iso (installiso) images in various ways and it worked for me. I have not tested no shim-install based distros like Fedora as the only addition is the early boot config replacement. Actually I expect this to solve some boot issues there too

schaefi:
why is all this grub stuff so broken :( Thanks a ton for setting up the builds with that change. I'll checkout your project and fetch the images from there and try to find the bugs

schaefi:
I found the problem... going to fix it

davidcassany:
> I found the problem... going to fix it

@schaefi awesome, just for the notes. I would consider as reference (for CentOS) the example image from [here](https://github.com/OSInside/kiwi-descriptions/tree/master/centos/x86_64/centos-08.0-JeOS) build from a regular host (it smoothly worked for me on Leap 15.1). Just to point that this image was functioning all fine in my builds. The OBS builds are always tricky and there things here and there...

I added you as a maintainer in home:dcassany:integration_tests projects, feel free to touch and commit anything in there if you need, there is no risk of data loss.

schaefi:
@davidcassany ok with that change my image tests were successful. Could you apply it to your home project and check as well ?

I found another glitch

```
error: ../../grub-core/commands/terminal.c:138:terminal `gfxterm' isn't found.
```

But I see that from any efi based build no matter if the shim efi image or the /boot/efi/EFI/fedora/grubx64.efi image is being used. From my perspective those prebuild images are missing some terminal stuff. It does not impact the boot process, however the gfxterm up to the point when the kernel inits the console is broken in my VM tests when booting with the emulated Tiano Core firmware. That part is not connected to the change here, it was broken before. Right now I don't think this is a kiwi issue

Thoughts ?

Thanks

davidcassany:
> @davidcassany ok with that change my image tests were successful. Could you apply it to your home project and check as well ?

Sure the update is on the way



> But I see that from any efi based build no matter if the shim efi image or the /boot/efi/EFI/fedora/grubx64.efi image is being used. From my perspective those prebuild images are missing some terminal stuff. It does not impact the boot process, however the gfxterm up to the point when the kernel inits the console is broken in my VM tests when booting with the emulated Tiano Core firmware. That part is not connected to the change here, it was broken before. Right now I don't think this is a kiwi issue

Yes I also saw this issue before. Few days ago I did a very quick google search to see if I could find an easy fix for that (config or package missing), however I did not see anything and considered it minor issue most probably related to the image definition. I did not spend time on trying to fix the graphics terminal neither the plymouth splash, in fact probably we could try to make the console work and that should be sufficient IMHO. So probably we should set boot for non SUSE images to a non graphical boot at all, so we avoid having to consider issues on this part. I don't think this sort of issues are actually KIWI bugs.

schaefi:
> So probably we should set boot for non SUSE images to a non graphical boot at all

I think for the functional testing we would like to perform with those images this makes a lot sense, also the console should respond to serial in all of them and we need a common way to access the system via ssh. But that's the next step I think

schaefi:
From my perspective this is now complete

schaefi:
> > So probably we should set boot for non SUSE images to a non graphical boot at all
> 
> I think for the functional testing we would like to perform with those images this makes a lot sense, also the console should respond to serial in all of them and we need a common way to access the system via ssh. But that's the next step I think

I've created #1518 to address this part

PR #1521, opened 2020-07-22
Make sure kernel options are respected by grub

schaefi:
Set options variable in grub loader/entries to match the
kiwi image description setup. This adaption currently
happens only when building in OBS because in this env
the options setup done by grub is a complete mess.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -298,8 +298,8 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                     with open(menu_entry_file) as grub_menu_entry_file:
                         menu_entry = grub_menu_entry_file.read()
                         menu_entry = re.sub(
-                            r'root=.*?(\s|$)(.*)',
-                            r'{0}\1\2'.format(self.root_reference),
+                            r'options (.*)',
+                            r'options {0}'.format(self.cmdline),
                             menu_entry
                         )
                     with open(menu_entry_file, 'w') as grub_menu_entry_file:
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -448,9 +448,28 @@ class TestBootLoaderConfigGrub2:
         self.bootloader.efi_mount.mountpoint = 'efi_mount_point'
         self.bootloader.early_boot_script_efi = 'earlyboot.cfg'
         with patch('builtins.open', create=True) as mock_open:
-            mock_open.return_value = MagicMock(spec=io.IOBase)
-            file_handle = mock_open.return_value.__enter__.return_value
-            file_handle.read.return_value = 'root=rootdev'
+            mock_open_grub = MagicMock(spec=io.IOBase)
+            mock_open_menu = MagicMock(spec=io.IOBase)
+
+            def open_file(filename, mode=None):
+                print(filename)
+                if filename == 'root_mount_point/boot/grub2/grub.cfg':
+                    return mock_open_grub.return_value
+                elif filename == 'some_entry.conf':
+                    return mock_open_menu.return_value
+                elif filename == 'grubenv':
+                    return mock_open_grub.return_value
+
+            mock_open.side_effect = open_file
+
+            file_handle_grub = \
+                mock_open_grub.return_value.__enter__.return_value
+            file_handle_menu = \
+                mock_open_menu.return_value.__enter__.return_value
+
+            file_handle_grub.read.return_value = 'root=rootdev'
+            file_handle_menu.read.return_value = 'options foo bar'
+
             self.bootloader.setup_disk_image_config(
                 boot_options={
                     'root_device': 'rootdev', 'boot_device': 'bootdev'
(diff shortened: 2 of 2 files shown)


schaefi:
I did debugging the whole day long and the result is this stupid code but I couldn't come up with anything better.
To be honest I have no clue about this new grub blscfg stuff but it feels worse than ever, sorry.

I think we don't want to merge this but it is a way to build images for Fedora and CentOS that actually can boot.
So let's at least discuss it

davidcassany:
I think the real issue here is that the `tuned` package is enabled by default and somehow executed at install time. Inside OBS this is causing issues. I bet that removing that package from the image description will make OBS images functional.

I had added this package in the list because this is a package that gets installed together with the `core` packages group. In my example images that I build in my host I make use of this pattern for simplicity. Then in OBS as we do not have groups/collection resolution for CentOS I just blindly included the package that are part of the `core` group.

The CentOS test are about to be updated in `home:dcassany:integration_tests` project, stay **tuned** :sweat_smile:

davidcassany:
All right I finally found what wrong with Fedora and CentOS. The OBS test from this PR #1524 are functional without this change, the it is a matter to decide how do we want to deal with this problem, by some tricks or workarounds in the image description or by some grub2 configuration hacks in KIWI code. None is satisfying. In order to help on the discussion these are the issues that we can currently see in Fedora RawHide and in CentOS 8:

* The CentOS case seams to have an easy but not reliable solution, which is preventing the `tuned` daemon installation. This is not reliable in the sense that this package is part of the `core` CentOS collection, thus one can expect to have in any regular install. This package at install time messes with BLS grub configuration and does wrong decisions most likely to be based on the current host setup. Excluding this package fixes CentOS builds in OBS, however this package did not cause issues on the builds run in my host. In any case this situation doesn't make me feel confident this is an OBS specific issue. IMHO this is a wrong behavior of CentOS, the boostrap procedure using the `core` group configures grub2 depending on the host setup, IMHO this is bad.

* In Fedora the issue is similar, but worse in the sense that the workaround solution in the image description is ugly. In fedora RawHide a new systemd/udev behavior to add and remove kernel and initramfs from /boot is being used, see the documentation [here](https://www.freedesktop.org/software/systemd/man/kernel-install.html). This behavior is provided by systemd-udev package, thus one can't expect to simply remove this package. The `kernel-install` utility instead of relaying on common grub2 configuration file it uses the file `/etc/kernel/cmdline` to define the options parameter for the BLS loader entry configuration. If the `/etc/kernel/cmdline` is not there it makes use of whatever is on `/proc/cmdline`, which is causing issues on KIWI builds as this is running in a chrooted env. Again, the problem of this is that this procedure happens at install time, thus the wrong grub2 loader entries are already there at the time we get into `config.sh` during a KIWI build. In order to make use of this configuration feature in KIWI we need to append into the image `/etc/kernel/cmdline` before the kernel is being installed, that means using an <archive name="kernelcmdline.tar.gz"> entry at bootstrap phase that includes this file. It works, but still quite annoying having to include a `/etc/kernel/cmdline` file this way.

davidcassany:
I have mixed feelings with this PR... since we are patching a file that is generated by the distro tools according some (arguable) criteria. And specially given that there are already in place mechanisms to alter this files from the distro tooling. So it looks like this PR should not be needed.

However I also agree that current CentOS and Fedora behavior is really bad on that regard (a simple packages installation on a bootstrapped root causes wrong grub2 configurations) and also the new BLS thing is IMHO wildly being spread across other tools. It is insane to see how `tuned` daemon can alter BLS configuration that was already created by a generator included in kernel package; and now, if it was not enough, systemd-udev can also replace the existing `/boot/loader/entries/*.conf` by adding some hooks on kernel installation and these hooks just introduce yet a new configuration file `/etc/kernel/cmdline` and ignore `/boot/grub2/grubenv` or `/etc/defaults/grub`. And the worst is that all this is happening at install time.

Probably you are right and it is just simpler to patch whatever is in `/boot/loader/entries/*.conf`. If at some point this is being adopted by other distros I'd say that we better add our on templates.

schaefi:
> So it looks like this PR should not be needed.

I fully agree but the current quality of the tooling is in such bad shape that I see ourselves to be forced to react on it.

> Probably you are right and it is just simpler to patch whatever is in /boot/loader/entries/*.conf.

I'm leaning towards merging this yes. We are simply replacing the kernel options line in that .conf file nothing more. kiwi knows exactly what needs to be there as we build the image. Furthermore the tools do not respect the custom options which we write into the grub defaults setup. Thus a user setting up a custom kernel option in the kiwi XML setup which lands in etc/default/grub will not land in the .conf file and will not be respected at all by the tooling. From my perspective a bug in the distribution and I'm sure there will be fixes/changes to this concept in the future.

In kiwi all of this "live patching" is handled in one place and can be dropped as soon as the tooling is fixed across all the distributions. The way it is done is just fixing errors. kiwi is not introducing its own concept. Because of that I don't think we are too intrusive or working against the native distribution concepts.

I'm going to merge this now as I think with your approval you agree with me. I totally agree with you that we should get rid of the code in kiwi as soon as possible. Nothing is ideal here but I don't see a better way to address this across all distros. On the bootloader level there are too many differences between the distros and I can only hope this will become better in the future. Until then we can manage the extra code in kiwi only. However we could think of refactor the code block a bit to make the adaptions more clear and maintainable. I'll add that as en extra issue

PR #1524, opened 2020-07-24
Add fedora rawhide test

davidcassany:
This PR upgrade the fedora 30 tests to run tests against Rawhide. In addition it also removes the `tuned` package from CentOS 8 integration test as this causing issues for the grub2 configuration.


davidcassany:
What's in this PR works without needing #1521, however I am far from satisfied. If we follow the #1521 idea there is no need to include this kernelcmline.tar.gz archive in the bootstrap phase.

davidcassany:
@schaefi I consider this to be sufficient. I cleared the hacks I did to workaround grub configuration images and now this is a clean simple image description. I validated VMX, OEM and ISOS, boot on EFI and BIOS. So grub wise we are done.

However I noticed the installation ISO of OEM fails at installation time, the raw image dump breaks before it completes and in consequence the checksum validation fails. I spent quite some time yesterday on it, looked like the raw image in the iso is corrupt. But today morning I realized Fedora RawHide, as Tumbleweed, uses kernel 5.8.0, so I assume we are facing #1531.

schaefi:
> so I assume we are facing #1531

I think so too

schaefi:
> So I'd say this is ready to be merged. 

It looks good to me, thanks much for the hard work. Are you going to update the label ? I don't dare to merge when don't merge is still set :-)

schaefi:
ah nevermind it just changed on the page I see here

PR #1528, opened 2020-08-10
Fixed doc structure for image types

### LINKED ISSUE #1520: Docu Site links to wrong page
## Problem description
The "Next" button at the bottom of https://osinside.github.io/kiwi/building_images/build_kis.html links to https://osinside.github.io/kiwi/working_with_images.html however the sidebar strongly suggests that one should continue with https://osinside.github.io/kiwi/building_images.html#image-types and the following sub-points of the https://osinside.github.io/kiwi/building_images.html# page

## Expected behaviour
The next page links to https://osinside.github.io/kiwi/building_images.html#image-types 

If it's only possible to link to subpages of building-images it might be worth to make `Images Types`, `Images Results` etc. in its own page(s).

schaefi:
The documentation did not differentiate correctly between
the explanation of the image types, their results and the
build host constraints compared to the documentation that
actually describes how to build an image for those types.
This Fixes #1520


schaefi:
The pull request also deletes the static table to show the supported distro vs. host distro table. This table can only be wrong as the distros are constantly changing and the information is typically outdated and probably wrong. The table was replaced with a recommendation how the build host should look like to prevent most of the errors.

Issue #1299, opened 2019-12-12
Decommission of legacy oemboot/netboot code


schaefi:
## Problem description

In KIWI we still support the pxe and oem image types to be connected to a custom KIWI initrd ```boot="netboot/..."``` or ```boot="oemboot/..."```

We kept the support because customers includes hook code which is supported by the netboot/oemboot initrd code but can't be used 1:1 when moving to dracut.

The time to phase out the support for custom KIWI initrd code has come because the distributions evolve and neither netboot nor oemboot does follow the changes anymore which will cause the code to be broken without having us noticing it.

Here is my suggestion to plan on removal of the netboot/oemboot descriptions and in the end the removal of the custom KIWI initrd build support:

- [x] Update test-image-oem-legacy to use an extra test package that provides the used oemboot description
- [x] Update test-image-pxe to be a test-image-kis or provide extra package that provides the used netboot description
- [x] Delete Leap15/15.1 netboot boot descriptions from custom_boot/ in the https://github.com/OSInside/kiwi-descriptions  repo. afaik nobody uses them
- [x] Announce the deletion of the TW oemboot descriptions from https://github.com/OSInside/kiwi-descriptions on the KIWI mailing list and Riot chat
- [x] Start a conversation with the OBS team who uses the netboot code for the buildservice workers. We would like them to take over the netboot code they use and package it for their own usecase if not already done. We would advise them to switch over to dracut and better implement a dracut module for the needs of a worker instead of using the netboot code, or switch to PXE disk image deployment for workers instead of filesystem images. That way they can plan what to do and how.
- [x] Send a drop request for kiwi-boot-descriptions in Factory. All people in suse who rely on it will now need to move. Start conversations where needed and arrange for an agreement in the same way as we did with the buildservice team
- [x] keep one version of netboot for reference, Delete custom_boot from https://github.com/OSInside/kiwi-descriptions git repo after 6months from the accept of the drop

Feedback welcome

schaefi:
I don't have a good idea how to deal with customers that uses SLES and oemboot/netboot.

schaefi:
I have sent an e-mail to the obs team regarding the take over of their netboot descriptions such that we can delete them from the custom_boot space but did not receive any feedback so far

schaefi:
- [x] Update test-image-oem-legacy to use an extra test package that provides the used oemboot description
- [x] Update test-image-pxe to be a test-image-kis or provide extra package that provides the used netboot description
[...]

PR #1535, opened 2020-08-27
Fixed legacy custom kiwi initrd based image builds

schaefi:
This patch is two fold:
    
* Image builds that uses the kiwi initrd system did not apply
   the grub config file fixes because the root= parameter is
   an optional information when using a kiwi initrd. However
   this information is required to apply the grub config file
   fixes. Therefore this patch simplifies the kernel commandline
   processing such that it is the same for dracut and custom
   kiwi initrd based systems. This means root= is passed in
   any case.
    
* The other part of the patch disables the method that strips
   unused libraries from the custom kiwi initrd. It has turned
   out that is safes us almost nothing but could causes corrupted
   initrds missing important libraries e.g libpam which is linked
   against udev and other tools. Therefore libpam is also added
   to the protected strip list

### DIFF
--- a/build-tests/x86/suse/test-image-oem-legacy/appliance.kiwi
+++ b/build-tests/x86/suse/test-image-oem-legacy/appliance.kiwi
@@ -63,6 +63,7 @@
         <package name="dracut-kiwi-oem-dump"/>
     </packages>
     <packages type="bootstrap">
+        <package name="gawk"/>
         <package name="udev"/>
         <package name="filesystem"/>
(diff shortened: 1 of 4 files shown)


schaefi:
@davidcassany I found this problem when testing the legacy builds while working on #1299

schaefi:
with this patch the legacy integration tests actually works fine on TW

davidcassany (approved):
Looks good, this is way is simpler :+1:

PR #1545, opened 2020-09-08
Explicitly set start sector for msdos repartition

### LINKED ISSUE #1543: Kiwi OEM resizing breaks root partition
Originally reported on https://lists.opensuse.org/opensuse-kubic/2020-09/msg00018.html

The OEM image has the partitions aligned at 512B sectors, but the target device can have a vastly different optimal I/O size.
This means that when deleting and recreating the partition, sfdisk aligns it differently and so it becomes useless.

Kiwi has to ensure that the start of the partition stays the *exact* same, no matter the alignment.
In addition, using a higher alignment by default (4kiB or 1MiB) would improve performance outside of virtual disks.

schaefi:
When using the sfdisk(>=2.26) tool the start and end of partitions
aligns to block-device I/O limits. This means it's hardware specific
what values sfdisk picks. In a repart operation this is however
dangerous for the start sector which must be the same to prevent
corruption of the data in the partition when kiwi resizes it.
Therefore this commit uses the sfdisk dump command to read the
start sector for deleted partitions and uses that start sector
if the same partition gets recreated such that only the end of
the partition changes and gets aligned properly.
This Fixes #1543

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -32,6 +32,7 @@ function create_msdos_partitions {
     local partid
     local part_size_end
     local cmd_list
+    local start_sector_from
     local cmd
 
     # put partition setup in a command list(cmd_list)
@@ -49,13 +50,16 @@ function create_msdos_partitions {
         "d")
             # delete a partition...
             partid=${cmd_list[$index + 1]}
+            start_sector_from[$partid]=$(
+                _get_msdos_partition_start_sector "${disk_device}" "${partid}"
+            )
             sfdisk --force --delete "${disk_device}" "${partid}"
         ;;
         "n")
             # create a partition...
             partid=${cmd_list[$index + 2]}
             part_size_end=${cmd_list[$index + 4]}
-            echo ",${part_size_end}" > /tmp/sfdisk.in
+            echo "${start_sector_from[$partid]},${part_size_end}" > /tmp/sfdisk.in
             sfdisk --force -N "${partid}" "${disk_device}" <  /tmp/sfdisk.in
         ;;
         "t")
@@ -404,6 +408,16 @@ function resize_wanted {
 #======================================
 # Methods considered private
 #--------------------------------------
+function _get_msdos_partition_start_sector {
+    # """
+    # read start sector from given partition
+    # """
+    local disk_device=$1
+    local partid=$2
+    sfdisk --dump "${disk_device}" |\
+        grep "${partid} :" | cut -f1 -d, | cut -f2 -d= | tr -d " "
+}
+
 function _to_guid {
     # """
     # convert two digit partition id to guid id


schaefi:
@davidcassany  I have to test this code more deeply prior merge

schaefi:
All tests worked

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -404,6 +408,16 @@ function resize_wanted {
...
 # Methods considered private
 #--------------------------------------
+function _get_msdos_partition_start_sector {
+    # """
+    # read start sector from given partition
+    # """
+    local disk_device=$1
+    local partid=$2
+    sfdisk --dump "${disk_device}" |\
+        grep "${partid} :" | cut -f1 -d, | cut -f2 -d= | tr -d " "
```
davidcassany, line 418: Just executed these commands in my host and I got something different than what I expected. This is what I got:

```
david@localhost:~> sudo sfdisk --dump /dev/sda 
label: gpt
label-id: CE4AA9A2-59DF-4DCC-B55A-A27A80676B33
device: /dev/sda
unit: sectors
first-lba: 34
last-lba: 500118158

/dev/sda1 : start=    17303552, size=   482814607, type=0FC63DAF-8483-4772-8E79-3D69D8477DE4, uuid=17B1F4B4-046E-4205-8592-A8284E5BC179
/dev/sda2 : start=      526336, size=    16777216, type=0657FD6D-A4AB-43C4-84E5-0933C84B4F4F, uuid=25D9EA9B-857B-470F-929D-6CD43842A71F
/dev/sda3 : start=        2048, size=      524288, type=C12A7328-F81F-11D2-BA4B-00A0C93EC93B, uuid=2C0AAEEA-455A-4CAD-8C2D-B034A4339769

david@localhost:~> sudo sfdisk --dump /dev/sda | grep 17B1 | cut -f2 -d= | tr -d " "
17303552,size
```
I would have expected `17303552` and not `17303552,size` as the output for this `_get_msdos_parition_start_sector` method. In fact I am surprised this is actually working. I am on Leap 15.2 (fresh install of today), sfdisk is from util-linux version 2.33.1.

davidcassany (commented):
I am not fully convinced about the "_get_msdos_partition_start_sector" method parsing, see my comment below. I'll give it a try later today or tomorrow early in the morning.

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
schaefi: This is not the grep command as I have coded it. Try:

```
sudo sfdisk --dump /dev/sda | grep "1 :" | cut -f2 -d= | tr -d " "
```

partid matches the number of the partition and is handed over at another place in the initrd code.
As I said I have done several tests with good results

davidcassany: Yes you are right I missed one `cut -f1 -d,` split on my test on the console, I did a wrong copy & paste. Yes I just verified it works as expected, on my host and on image. Sorry for the noise.

davidcassany (approved):
This is good I just have also verified it on a OEM msdos efi image. :+1:

schaefi:
Great thanks much for testing

PR #1546, opened 2020-09-08
Support dynamic linux/linuxefi in any case

schaefi:
Instead of restricting the dynamic linux vs. linuxefi setup
to a specific grub version, support this setup for any version
of grub. This Fixes bsc#1175729 and bsc#1176134

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -24,7 +24,6 @@ import shutil
 from collections import OrderedDict
 
 # project
-from kiwi.utils.command_capabilities import CommandCapabilities
 from kiwi.bootloader.config.base import BootLoaderConfigBase
 from kiwi.bootloader.template.grub2 import BootLoaderTemplateGrub2
 from kiwi.command import Command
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
Looks good :+1:

This is a never ending story. The dynamic substitution the `linux(efi)` at run time is a pain...

schaefi:
> This is a never ending story. The dynamic substitution the `linux(efi)` at run time is a pain...

yes but maybe they really decide to use linuxefi in any case which would finally solve the issue for us and we could also delete this magic from kiwi. It just reminds me that we want to move that extra code into methods such that we can easily delete them. I think I'm going to work on that next

Vogtinator:
> This is a never ending story. The dynamic substitution the linux(efi) at run time is a pain...

Fully agreed.

> > This is a never ending story. The dynamic substitution the `linux(efi)` at run time is a pain...
> 
> yes but maybe they really decide to use linuxefi in any case which would finally solve the issue for us and we could also delete this magic from kiwi.

Fortunately it looks like that's the current direction (yay, I finally convinced him!) so should be supported on 15.2+ in the future.

PR #1548, opened 2020-09-10
Fix target setup on grub install

### LINKED ISSUE #1547: Wrong grub2 installation in Ubuntu Focal
## Problem description

OEM images of Ubuntu Focal complain with a `can't find command save_env` error on boot, after about 10 seconds of this error msg pressing any key triggers the boot to continue normally. 

Apparenlty this might be caused by a wrong `grub-install` call. In fact in KIWI for any x86 architecture we use the `--target i386-pc`, however `x86_64-efi` (for x86_64 architectures) seams to be required for Ubuntu. I have no idea about how Ubuntu differs from other distros on that regard, but being on x86_64 with en EFI/UEFI setup using `x86_64-efi` definitely looks like the correct target. I am wondering why we are not seeing this issue in other distros. 

So after boot the effected image I can simply run

`$ grub-install --target x86_64-efi --boot-directory /boot /dev/sda`

and this solves the issue. Moreover I verified that if a reinstall grub again with:

`$ grub-install --target i386-pc --boot-directory /boot /dev/sda`

brings back the reported problem. Note that the grub configuration files were not touched in any case, thus I assume the grub config files generated by KIWI are correct.

## Expected behaviour

OEM image boots without error messages and `press any key to continue` interactive messages.

Boot the OEM image for the UBUNTU integration test.

schaefi:
The parameter --target in a grub install call specifies the
target platform. For the EFI target the setting was wrong.
Strange enough this did not cause problems on some distros
and was therefore not seen until now. This Fixes #1547

### DIFF
--- a/kiwi/bootloader/install/grub2.py
+++ b/kiwi/bootloader/install/grub2.py
@@ -129,12 +129,16 @@ class BootLoaderInstallGrub2(BootLoaderInstallBase):
             self.install_arguments.append('--removable')
 
         if Defaults.is_x86_arch(self.arch):
-            self.target = 'i386-pc'
+            if self.firmware and self.firmware.efi_mode():
+                self.target = 'x86_64-efi'
+            else:
+                self.target = 'i386-pc'
             self.install_device = self.device
             self.modules = ' '.join(
                 Defaults.get_grub_bios_modules(multiboot=True)
             )
             self.install_arguments.append('--skip-fs-probe')
+
         elif self.arch.startswith('ppc64'):
             if not self.custom_args or 'prep_device' not in self.custom_args:
                 raise KiwiBootLoaderGrubInstallError(
--- a/test/unit/bootloader/install/grub2_test.py
+++ b/test/unit/bootloader/install/grub2_test.py
@@ -279,7 +279,7 @@ class TestBootLoaderInstallGrub2:
         mock_exists.return_value = True
         mock_glob.return_value = ['tmp_root/boot/grub2/grubenv']
         mock_grub_path.return_value = \
-            self.root_mount.mountpoint + '/usr/lib/grub2/i386-pc'
+            self.root_mount.mountpoint + '/usr/lib/grub2/x86_64-efi'
         self.firmware.efi_mode.return_value = 'uefi'
         self.boot_mount.device = self.root_mount.device
 
@@ -296,9 +296,9 @@ class TestBootLoaderInstallGrub2:
         assert mock_command.call_args_list == [
             call([
                 'chroot', 'tmp_root', 'grub2-install', '--skip-fs-probe',
-                '--directory', '/usr/lib/grub2/i386-pc',
+                '--directory', '/usr/lib/grub2/x86_64-efi',
                 '--boot-directory', '/boot',
-                '--target', 'i386-pc',
+                '--target', 'x86_64-efi',
                 '--modules', ' '.join(
                     Defaults.get_grub_bios_modules(multiboot=True)
                 ),
@@ -340,7 +340,7 @@ class TestBootLoaderInstallGrub2:
         mock_exists.return_value = True
         mock_glob.return_value = ['tmp_root/boot/grub2/grubenv']
         mock_grub_path.return_value = \
-            self.root_mount.mountpoint + '/usr/lib/grub2/i386-pc'
+            self.root_mount.mountpoint + '/usr/lib/grub2/x86_64-efi'
         self.firmware.efi_mode.return_value = 'uefi'
         self.boot_mount.device = self.root_mount.device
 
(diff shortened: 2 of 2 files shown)


schaefi:
I haven't extensively tested it. Let's review and do some tests prior merge

schaefi:
ok so there are issues with this patch. Actually I know why we always used i386-pc when running grub2-install. The reason here is that a grub2-install that installs bios code to an MBR only runs for i386-pc and never for EFI because on EFI there is no boot code that actually needs to be installed.

So I wonder about that grub2-install call with target set to x86_64-efi as it in theory doesn't make any sense

davidcassany (commented):
I already tested this patch yesterday without success on Ubuntu. I did not test on SUSE though.

I noted is that configuring grub to EFI and calling `grub-install` on `i386-pc` target leads to an installation that includes both:

```
ls /boot/grub2/ -l
total 12
drwxr-xr-x 1 root root   22 Sep 10 06:53 fonts
-rw------- 1 root root 6175 Sep 10 11:29 grub.cfg
-rw-r--r-- 1 root root 1024 Sep 10 06:53 grubenv
drwxr-xr-x 1 root root 6538 Sep 10 06:53 i386-pc
drwxr-xr-x 1 root root   16 Sep 10 06:48 themes
drwxr-xr-x 1 root root 6200 Sep 10 06:48 x86_64-efi
```
See the i386-pc folder and the x86_64-efi folder in it. However if we apply a similar patch of what is in this PR, on Ubuntu, the i386-pc folder is gone after the build and it doesn't boot for non efi firmwares...

davidcassany:
> ok so there are issues with this patch. Actually I know why we always used i386-pc when running grub2-install. The reason here is that a grub2-install that installs bios code to an MBR only runs for i386-pc and never for EFI because on EFI there is no boot code that actually needs to be installed.
> 
> So I wonder about that grub2-install call with target set to x86_64-efi as it in theory doesn't make any sense

Ok so that explaing why using the `--target i386-pc` first and then reinstalling with `--targer x86_64-efi` fixes the issue on Ubuntu, as the second does not overwrite MBR so the non-EFI boot is still there. Then I guess we will have to figure out what else is doing `grub-install`

schaefi:
yes something is weird. The way we call grub2 seems correct to me (without the changes here)

schaefi:
I wonder what exactly is the problem on Focal ? I have tested the boot and the only weird thing I found is that you need to hit a key (Escape) to let it show the grub menu, otherwise it just boots without showing a menu

davidcassany:
There is a `save_env` function in the header script in `/etc/grub.d` which is being used in  `grub.cfg`, however for some reason in x86_64-efi boots the command seams not to be loaded. 

If I boot the OEM image of our integration tests (the disk not the install ISO) I see this issue if boot with KVM using the `-bios /usr/share/qemu/omvf-x84_86.bin` flag.

I noted that after booting I have this:

```
$ ls /boot/efi/EFI
BOOT
```

However after running `grub-install --target x86_64-efi` the result is different:

```
$ ls /boot/efi/EFI
BOOT grub

$ ls /boot/efi/EFI/grub/
grubx64.efi
```

However I am not sure this is actually making any difference as the `/boot/efi/EFI/BOOT/bootx64.efi` was already there and it is the same exact binary file as `/boot/efi/EFI/grub/grubx64.efi`

schaefi:
The save_env method is defined for x86 boot code as well as for efi binaries. If you take a look at a suse system you will find this here:

/usr/share/grub2/i386-pc/command.lst
    *save_env: loadenv

Implementation in /usr/share/grub2/i386-pc/loadenv.mod

for efi it is here:

/usr/share/grub2/x86_64-efi/command.lst
    *save_env: loadenv

Implementation in /usr/share/grub2/x86_64-efi/loadenv.module. For efi that module is usually compiled into the efi binary. That is the /boot/efi/EFI/grub/grubx64.efi file you found.

So if you don't see the issue when booting via EFI all is good on the EFI front. If you see the issue only when booting in bios mode I assume the grub2-install has not installed an mbr code that has this module.

If kiwi creates its own grub2 bios module we include the modules as setup here:

```
@staticmethod
    def get_grub_bios_modules(multiboot=False):
        """
        Provides list of grub bios modules

        :param bool multiboot: grub multiboot mode

        :return: list of module names

        :rtype: list
        """
        modules = Defaults.get_grub_basic_modules(multiboot) + [
            'part_gpt',
            'part_msdos',
            'biosdisk',
            'vga',
            'vbe',
            'chain',
            'boot'
        ]
        return modules
```

As you can see loadenv is not explicitly mentioned but that is usually also not needed because it depends on everything.
But it's worth checking

However in most cases the distro provided grub modules are used. You can check that from the log file of our integration test.
There should be a message:

```
'--> Using prebuilt bios image'
```

If you see this there seems to be a bug in the distro provided grub module. If not we might want to update our module list and see if that makes a difference.

Could you check all this ? I'm currently busy with pubcloud work and a showcase today

Thanks

davidcassany:
@schaefi thanks for the comments you left here, they lead me to the right place.

> So if you don't see the issue when booting via EFI all is good on the EFI front. If you see the issue only when booting in bios mode I assume the grub2-install has not installed an mbr code that has this module.

It is exactly the opposite, the EFI front is the one failing, the BIOS firmware and MBR are perfectly working fine. Our list of modules for EFI seams not to be complete enough. In Ubuntu and Debian we are creating the unsigned images for the grub as KIWI does not find any image provided by the distro, thus in that case it is just created with `grub-mkimage` and the default list of modules provided by `kiwi/defautls.get_grub_efi_modules`. Adding `loadenv` module in such a list seams to fix the issue.

This is something that went moslty silent as Ubuntu/Debian are the only images in our integration tests that actually make use of these features. These are the only images that create the unsigned grub image during the build.

Gonna provide a PR with the change

schaefi:
> Gonna provide a PR with the change

great, I'm closing this one here. Thanks much

PR #1549, opened 2020-09-11
Make oem image use dracut by default

schaefi:
Before this commit an oem image still had the kiwi initrd_system set as default.

As we are decommission the custom kiwi initrd concept the default should be changed. It is still possible to use a custom kiwi initrd but it needs to be explicitly requested via the initrd_system="kiwi" attribute. In addition to the changed default a runtime check was introduced that checks the presence of the boot= attribute which only makes sense in combination with the kiwi initrd_system. If boot= is set but initrd_system="kiwi" is not, a message is raised that explains the situation and either requests setting initrd_system properly or deleting the boot attribute. The change only affects people who still use oem with a boot="oemboot/..." setting and no explicit selection of kiwi as the initrd_system.

As these image type configurations should not be in use anyway because this is all legacy and announced to go away, we
need to make the next step and enforce a new default in code.

This is related to Issue #1299

### DIFF
--- a/build-tests/arm/suse/test-image-rpi-oem/appliance.kiwi
+++ b/build-tests/arm/suse/test-image-rpi-oem/appliance.kiwi
@@ -15,7 +15,7 @@
         <keytable>us</keytable>
         <timezone>Europe/Berlin</timezone>
         <rpm-excludedocs>true</rpm-excludedocs>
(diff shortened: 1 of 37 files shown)


davidcassany:
Looks good, the runtime test with the error message is nice. I consider this explicit enough for anyone that faces the request of having to add a `initrd_system="kiwi"` on an OEM image.

davidcassany (approved):

PR #1550, opened 2020-09-14
Include loadenv as a default module for grub images

### LINKED ISSUE #1547: Wrong grub2 installation in Ubuntu Focal

davidcassany:
This commit includes the 'loadenv' module to the list of basic grub
modules. This makes sure the module is included in any grub-mkimage
that KIWI does.

Fixes #1547

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -455,7 +455,8 @@ class Defaults:
             'crypto',
             'cryptodisk',
             'test',
-            'true'
+            'true',
+            'loadenv'
         ]
         if multiboot:
             modules.append('multiboot')
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -717,8 +717,8 @@ class TestBootLoaderConfigGrub2:
                     'gettext', 'font', 'minicmd', 'gfxterm', 'gfxmenu',
                     'all_video', 'xfs', 'btrfs', 'lvm', 'luks',
                     'gcry_rijndael', 'gcry_sha256', 'gcry_sha512', 'crypto',
-                    'cryptodisk', 'test', 'true', 'multiboot', 'part_gpt',
-                    'part_msdos', 'efi_gop', 'efi_uga', 'linuxefi'
+                    'cryptodisk', 'test', 'true', 'loadenv', 'multiboot',
+                    'part_gpt', 'part_msdos', 'efi_gop', 'efi_uga', 'linuxefi'
                 ]
             )
         ]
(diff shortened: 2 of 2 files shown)


davidcassany:
I only tested OEM Ubuntu/Debian images so far. I'll do few more tests of OEM disks and installation ISOs on other distros.

I believe that adding such module into the grub modules list should not hurt in any case. I assume it doesn't hurt on ISOs even look like this is completely useless in such a use case. I added it on the basic modules list, meaning this is included in EFI and BIOS grub images. I could have narrowed more the case where we actually noticed the issue, only on EFI setups, however it felt bad to me adding this module only for EFI when it is not something specifically related to EFI features.

schaefi (approved):
The change looks perfectly fine to me and I agree it will not hurt in any case. loadenv is a module that provides save_env function which is used in any grub2-mkconfig call. Thus it's mandatory for grub to provide that module for any supported platform.

The kiwi code also aims for using the pre-built version of grub binaries as they are provided by the distribution. Only for those distros that does not provide their own vision of such a grub binary kiwi creates its own one as it will become mandatory in this case.

So if we want to followup on this change it might be good to check if on Ubuntu the prebuilt version really doesn't exist or if we just did not look at the right place.

In any case I like this change here :+1:

davidcassany:
> So if we want to followup on this change it might be good to check if on Ubuntu the prebuilt version really doesn't exist or if we just did not look at the right place.

Yes, good point. I did a small research and trials on that regard and I could not find anything conclusive. Seams that Ubuntu/Debian are providing some binaries in `/usr/lib/grub/x86_64-efi/monolithic/` folder, however adding such a path in KIWI defaults resulted in using a prebuilt binary that lead to a broken grub installation (image build succeeded, but it was not booting). So I guess further investigations are needed beyond running `dpkg -L grub-efi-amd64-bin | grep "\.efi"` to identify any prebuild grub image.root

```
$ dpkg -L grub-efi-amd64-bin | grep "\.efi"
/usr/lib/grub/x86_64-efi/monolithic/gcdx64.efi
/usr/lib/grub/x86_64-efi/monolithic/grubnetx64.efi
/usr/lib/grub/x86_64-efi/monolithic/grubx64.efi
```

I tried adding `/usr/lib/grub*/*-efi/*/grubx64.efi` to the lookup path and the result did not work, the image build succeeded using the `grubx64.efi` file from that package but the result was not EFI bootble.

In any case I consider using the prebuild image as a nice to have. The current PR as it is solves a generic issue not strictly tied to Ubuntu, so I'd say this ready as it is. Figuring out if there is a prebuilt image ready to be used in Ubuntu seams to require some time a dedication.

schaefi:
ok thanks for investigating that. Let's go with this change, thanks much

PR #1552, opened 2020-09-15
Make oem be a superset of vmx

### LINKED ISSUE #1425: Making oem be a superset of vmx
## Problem description

This is more a query than a bug -- hopefully that's ok.  (I'd issues signing up for Google Groups when I tried to post to the mailing list). 

First up, I'm new to kiwi. I've been able to build some Centos 7 images superfast -- I'm impressed with what I've seen so far.  But it's not clear how to best create a swap partition on a vmx image.

I've seen some other queries about this, eg

https://groups.google.com/forum/#!searchin/kiwi-images/swap|sort:date/kiwi-images/2YrRhtM5KZ0/Z2DM9Q7tBwAJ

https://groups.google.com/forum/#!searchin/kiwi-images/swap|sort:date/kiwi-images/pr7ryvxjHCw/Bgs-CoxXDwAJ

They seem to suggest some workarounds, eg creating swap on first boot.

I seem to have been able to create a vmx swap partition directly from kiwi, without separate workaround commands.

After reading through the code I tried mixing in some `oemconfig` options into the example Centos 7 config.xml, like this:

```
        <type image="vmx" primary="true" format="vmdk" filesystem="ext3" kernelcmdline="rhgb no_timer_check console=tty0 console=ttyS0,115200 net.ifnames=0 biosdevname=0" bootloader="grub2" firmware="uefi">
            <size unit="G">8</size>
            <oemconfig>
                <oem-swap>true</oem-swap>
                <oem-swapsize>400</oem-swapsize>
            </oemconfig>
        </type>
```

On the VM I checked that this produced what I wanted: a live 400M swap partition: 

```
# swapon --show
NAME      TYPE      SIZE USED PRIO
/dev/vda3 partition 400M   0B   -2
```

For completeness, the partition table looked like this:

```
# fdisk -l /dev/vda
WARNING: fdisk GPT support is currently new, and therefore in an experimental phase. Use at your own discretion.

Disk /dev/vda: 8589 MB, 8589934592 bytes, 16777216 sectors
Units = sectors of 1 * 512 = 512 bytes
Disk label type: gpt
Disk identifier: EE3D4058-DDBF-4EB9-8B64-B38DD4B6CA92


#         Start          End    Size  Type            Name
 1         2048         6143      2M  BIOS boot       p.legacy
 2         6144        47103     20M  EFI System      p.UEFI
 3        47104       866303    400M  Linux swap      p.swap
 4       866304     16777182    7.6G  Linux filesyste p.lxroot
```

After reading the code I'm wondering if this may be working 'by accident'. As far as I can tell, only the first existing `oemconfig` snippet will be read from the xml file, so if you have an oem type with it's `oemconfig` section I think it's possibly only the first one will ever be read (I haven't tested that this is true).

In short: is using `oemconfig` with `vmx` in this way supported?

## Expected behaviour

From reading the documentation I think I'd have expected this to produce an xml schema error saying that `oemconfig` is not valid when used with type vmx.

Obviously, it's doing what I want though (!).

+ KIWI version: latest
+ Operating system: opensuse container
+ OBS version: centos 7

Thanks

schaefi:
A vmx image is the same disk as an oem just without the dracut
repart/resize feature. This difference is better handled with
an oemconfig parameter <oem-resize> which allows to switch resize
on or off. The extra image type vmx will be dropped and an XSLT
stylesheet automatically transforms a vmx description to be a
oem image with the resize feature switched off.
This Fixes #1425

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -265,16 +265,14 @@ image="iso"
   system storage components. A useful pocket system for testing
   and demo and debugging purposes.
 
-image="vmx"
-  An image representing the system disk, useful for cloud frameworks
-  like Amazon EC2, Google Compute Engine or Microsoft Azure.
-
 image="oem"
   An image representing an expandable system disk. This means after
   deployment the system can resize itself to the new disk geometry.
   The resize operation is configurable as part of the image description
   and an installation image for CD/DVD, USB stick and Network deployment
-  can be created in addition.
+  can be created in addition. For use in cloud frameworks like
+  Amazon EC2, Google Compute Engine or Microsoft Azure this disk
+  type also supports the common virtual disk formats.
 
 image="docker"
   An archive image suitable for the docker container engine.
@@ -434,29 +432,35 @@ spare_part="number":
   of the requested size. The attribute takes a size value
   and allows a unit in MB or GB, e.g 200M. If no unit is given
   the value is considered to be mbytes. A spare partition
-  can only be configured for the disk image types oem and vmx
+  can only be configured for the disk image type oem
 
 spare_part_mountpoint="dir_path":
   Specify mount point for spare partition in the system.
-  Can only be configured for the disk image types oem and vmx
+  Can only be configured for the disk image type oem
 
 spare_part_fs="btrfs|ext2|ext3|ext4|xfs":
   Specify filesystem for spare partition in the system.
-  Can only be configured for the disk image types oem and vmx
+  Can only be configured for the disk image type oem
 
 spare_part_fs_attributes="attribute_list":
   Specify filesystem attributes for the spare partition.
   Attributes can be specified as comma separated list.
   Currently the attributes `no-copy-on-write` and `synchronous-updates`
   are available. Can only be configured for the disk image
-  types oem and vmx
+  type oem
 
 spare_part_is_last="true|false":
   Specify if the spare partition should be the last one in
-  the partition table. Can only be configured for the vmx
-  disk image type. By default the root partition is the last
-  one and the spare partition lives before it. With this
-  attribute that setup can be toggled.
+  the partition table. Can only be configured for the `oem`
+  type with oem-resize switched off. By default the root
+  partition is the last one and the spare partition lives
+  before it. With this attribute that setup can be toggled.
+  However, if the root partition is no longer the last one
+  the oem repart/resize code can no longer work because
+  the spare part would block it. Because of that moving
+  the spare part at the end of the disk is only applied
+  if oem-resize is switched off. There is a runtime
+  check in the {kiwi} code to check this condition
 
 devicepersistency="by-uuid|by-label|by-path":
   Specifies which method to use for persistent device names.
@@ -473,7 +477,7 @@ overlayroot="true|false"
   out of a squashfs compressed read-only root system
   overlayed using the overlayfs filesystem into an
   extra read-write partition. Available for the disk
-  image types, vmx and oem
+  image type oem only
 
 bootfilesystem="ext2|ext3|ext4|fat32|fat16":
   If an extra boot partition is required this attribute
@@ -491,7 +495,7 @@ flags="overlay|dmsquash"
   module.
 
 format="gce|ova|qcow2|vagrant|vmdk|vdi|vhd|vhdx|vhd-fixed":
-  For disk image types vmx and oem, specifies the format of
+  For disk image type oem, specifies the format of
   the virtual disk such that it can run on the desired target
   virtualization platform.
 
@@ -610,8 +614,8 @@ element including references to their usage in a detailed type setup:
 
 <preferences><type><bootloader>
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Used to describe the bootloader setup in vmx or oem disk image types.
-For details see: :ref:`vmx-bootloader`
+Used to describe the bootloader setup in the oem disk image type.
+For details see: :ref:`disk-bootloader`
 
 <preferences><type><containerconfig>
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -621,12 +625,12 @@ image types. For details see: :ref:`building_docker_build` and:
 
 <preferences><type><vagrantconfig>
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Used to describe vagrant configuration metadata in vmx disk image
+Used to describe vagrant configuration metadata in a disk image
 that is being used as a vagrant box. For details see: :ref:`setup_vagrant`
 
 <preferences><type><systemdisk>
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Used to describe the geometry, partitions and volumes, in a vmx or oem
+Used to describe the geometry, partitions and volumes, in a
 disk image. For details see: :ref:`custom_volumes`
 
 <preferences><type><oemconfig>
@@ -636,14 +640,14 @@ For details see: :ref:`oem_customize`
 
 <preferences><type><size>
 ~~~~~~~~~~~~~~~~~~~~~~~~~
-Used to customize the size of the resulting disk image in a vmx or
-oem image. For details see: :ref:`vmx-the-size-element`
+Used to customize the size of the resulting disk image in an
+oem image. For details see: :ref:`disk-the-size-element`
 
 <preferences><type><machine>
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Used to customize the virtual machine configuration which describes
 the components of an emulated hardware.
-For details see: :ref:`vmx-the-machine-element`
+For details see: :ref:`disk-the-machine-element`
 
 .. _sec.repository:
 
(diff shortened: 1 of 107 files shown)


schaefi:
Hi guys. This was a big effort but I think I covered all aspects to make the move and have only one disk type oem in a way that our users are transparently upgraded without having the need to change anything in their descriptions.

@davidcassany As you can see it's a big change but I think still good to review as I mostly had to touch on the unit test setup which we really could do better ;) It would be great if you can have a close look to the change and maybe also build one or the other image.

@tomschr The XSLT is not perfect. I wrote you an e-mail that shows how the transformation can produce a wrong result. I don't want to merge this before the XSLT is perfect. Your help in that area is very much appreciated. Thanks much

kiwi/xsl/convert72to73.xsl
```diff
@@ -0,0 +1,54 @@
...
+                    <oem-resize>false</oem-resize>
+                </oemconfig>
+                <xsl:apply-templates  mode="conv72to73"/>
+            </type>
+        </xsl:when>
+        <xsl:otherwise>
+            <xsl:copy-of select="."/>
+        </xsl:otherwise>
+    </xsl:choose>
+</xsl:template>
```
schaefi, line 52: Here it would be great to get some help. If an oemconfig section existed before this stylesheet produces another one and the XML becomes invalid

davidcassany (approved):
Looks good to me. Great effort! I have just one comment about the validation between `sparte_partition_is_last` and `oem-resize` options.

kiwi/builder/disk.py
```diff
@@ -785,7 +787,8 @@ def _build_and_map_disk_partitions(self, disksize_mbytes):  # noqa: C901
             )
             disksize_used_mbytes += squashed_rootfs_mbsize
 
-        if self.spare_part_mbsize and self.spare_part_is_last:
+        if self.spare_part_mbsize and self.spare_part_is_last and not \
+           self.disk_resize_requested:
```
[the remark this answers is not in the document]
schaefi: > The consistency of spare partition options together with resizing options is validated on the runtime check. 

this is correct, but we also allow to disable runtime checks. Assuming the runtime check is disabled and the code above does not look at disk_resize_requested and someone configures spare_part_is_last together with oem-resize enabled it will lead to a system that will fail the repart step on boot. I think the failed repart will still lead the system to boot up. So the question is what is better:

a) disabled runtime check and invalid config leads to a system that fails the repart on boot
b) disabled runtime check and invalid config leads to a system that ignores spare_part_is_last

maybe option a) is better as it tries to setup the requested but fails and there is also an error message about it.
In this case I would just delete the self.disk_resize_requested check in the if conditions.

> Also wondering if the sanity check couldn't be done with some schematron rule.

That might be the best solution. I will look into it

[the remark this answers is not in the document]
schaefi: I have created a schematron rule which works for me. The only problem I see with that is that we can't display longer text information as it all prints it in one line. See the new commit: Thoughts ?

kiwi/xsl/convert72to73.xsl
tomschr: Just for the record, I've suggested this snippet:

```xml
<xsl:template match="type[@image = 'vmx'][not(oemconfig)]" mode="conv72to73">
      <type image="oem">
         <xsl:copy-of select="@*[local-name() != 'image']"/>
         <oemconfig>
         </oemconfig>
         <xsl:apply-templates mode="conv72to73"/>
      </type>
   </xsl:template>
```

When I use this input XML:

```xml
<type image="vmx"/>
```

I get this output:

```xml
<type image="oem">
  <oemconfig>
  </oemconfig>
</type>
```

A `type` element with an `oemconfig` is not affected by this template rule.

schaefi: Thanks. I think your template is more robust and easier to read but it does not cover all cases. Here is the list of cases we are facing:

```xml
<type image="vmx"/>
```

should be transformed to

```xml
<type image="oem">
  <oemconfig>
  </oemconfig>
</type>
```

This is covered nicely. Next there could also be the following:

```xml
<type image="vmx">
  <oemconfig>
    <oem-swap>true</oem-swap>
  </oemconfig>
</type>
```

It doesn't make sense as a vmx would not use the provided information but the schema out there would allow this setup and it should be transformed to:

```xml
<type image="oem">
  <oemconfig>
    <oem-swap>true</oem-swap>
  </oemconfig>
</type>
```

I think this part is not covered by the current stylesheet

doc/source/image_description/elements.rst
```diff
@@ -434,29 +432,35 @@ spare_part="number":
...
+  type oem
 
 spare_part_is_last="true|false":
   Specify if the spare partition should be the last one in
-  the partition table. Can only be configured for the vmx
-  disk image type. By default the root partition is the last
-  one and the spare partition lives before it. With this
-  attribute that setup can be toggled.
+  the partition table. Can only be configured for the oem
+  type with oem-resize switched off. By default the root
```
tomschr, line 455: I think, you should use backticks here to visualize that it not a concept but a value or an element:

```suggestion
  the partition table. Can only be configured for the `oem`
  type with `oem-resize` switched off. By default the root
```

doc/source/image_description/elements.rst
```diff
@@ -434,29 +432,35 @@ spare_part="number":
...
+  partition is the last one and the spare partition lives
+  before it. With this attribute that setup can be toggled.
+  However if the root partition is no longer the last one
```
tomschr, line 458:
```suggestion
  However, if the root partition is no longer the last one
```

doc/source/image_description/elements.rst
```diff
@@ -434,29 +432,35 @@ spare_part="number":
...
+  the oem repart/resize code can no longer work because
+  the spare part would block it. Because of that moving
+  the spare part at the end of the disk is only applied
+  if oem-resize is switched off. There is a runtime
+  check in the kiwi code to check this condition
```
tomschr, line 463:
```suggestion
  check in the {kiwi} code to check this condition
```

tomschr (commented):
Just some minor comments regarding typos, spell checking, and some markup stuff. Maybe you would like to add these. :wink:

kiwi/xsl/convert72to73.xsl
schaefi: and be aware oemconfig can contain many other elementes oem-swap was just taken as an example. So what we need is to keep the existing contents of oemconfig (except for oem-resize) and an explicit <oem-resize>false</oem-resize>

tomschr: Ok, I suggest an additional template rule like this:

```xml
<xsl:template match="type[@image = 'vmx']
                     [normalize-space(oemconfig/oem-swap) = 'true']" mode="conv72to73">
    <type image="oem">
         <oemconfig>
            <xsl:apply-templates select="oemconfig/*[not(self::oem-resize)]"
               mode="conv72to73"/>
         </oemconfig>
    </type>
</xsl:template>
```

This would match for the second example and copy all elements from `oemconfig` except `oem-resize`.

-----

Additionally, maybe we should rewrite the first template a bit. At the moment, it looks like this:

```xml
<xsl:template match="type[@image = 'vmx'][not(oemconfig)]" mode="conv72to73">
```

This would match any `type` element without an `oemconfig` part. If you really want to match *only* an empty `<type image="vmx" />` you can rewrite the match like this:

```xml
<xsl:template match="type[@image = 'vmx'][not(*)]" mode="conv72to73">
```

Not sure if this would be better or worse. :wink:

schaefi:
@tomschr thanks much for the doc review, all changes in now :)

kiwi/xsl/convert72to73.xsl
schaefi: With all the info you provided I think we have a working version now

kiwi/xsl/convert72to73.xsl
```diff
@@ -0,0 +1,55 @@
...
+                    <oem-resize>false</oem-resize>
+                </oemconfig>
+                <xsl:apply-templates select="*[not(self::oemconfig)]" mode="conv72to73"/>
+            </type>
+        </xsl:when>
+        <xsl:otherwise>
+            <xsl:copy-of select="."/>
+        </xsl:otherwise>
+    </xsl:choose>
+</xsl:template>
```
schaefi, line 53: @tomschr please double check if the apply-templates calls makes sense to you too. Thanks

schaefi: oh and I think I forgot to add your suggestion to get rid of the "when"

kiwi/xsl/convert72to73.xsl
```diff
@@ -0,0 +1,48 @@
...
+<xsl:template match="type[@image='vmx']" mode="conv72to73">
+    <type image="oem">
+        <xsl:copy-of select="@*[local-name() != 'image']"/>
+        <oemconfig>
+            <xsl:apply-templates select="oemconfig/*[not(self::oem-resize)]" mode="conv72to73"/>
+            <oem-resize>false</oem-resize>
+        </oemconfig>
+        <xsl:apply-templates select="*[not(self::oemconfig)]" mode="conv72to73"/>
+    </type>
+</xsl:template>
```
schaefi, line 46: @tomschr ok final version, how about this ?

kiwi/xsl/convert72to73.xsl
tomschr: Looks good to me. :+1: Can confirm it works also in my little example.

kiwi/xsl/convert72to73.xsl
tomschr: Looks good to me. +1 Can confirm it works also in my little example.

davidcassany:
I think that having the schematron rule to prevent inconsistent definitions is better than runtime checks. I see the issue with the user message and its friendliness, however I think this is such a very specific case in which the message is shown. I am fine with schema validation errors that clearly point to the offensive attributes or tags, and this is the case.

I understand your frustration as the previous message was way nicer, but frankly speaking I prefer having this validations in XML schema when ever is possible.

Regarding the XLS conversion I can't think of any other use case beyond what has bee already discussed. I only see the weird case in which there is some `<oem-config>` section within a VMX image type.

schaefi:
>I understand your frustration as the previous message was way nicer, but frankly speaking I prefer having this validations in XML schema when ever is possible.

There is no frustration :) I just realized showing a message in schematron should be small :) What you said makes complete sense. I was mostly interested if we can live with the mini concern of the message and I agree it's a corner case. So let's go with the rule in the schema. I'll merge the commits

The XSLT is fine now too and I think we are ready to merge it.

I will create a Staging submission and then we should boot test the images. I suggest to have a short test on at least one per distro.

Thanks much

stuart-mclaren-hpe:
Many thanks for following up on this.

PR #1554, opened 2020-09-16
Fix the early boot grub.cfg file

### LINKED ISSUE #1553: Boot fails on images with separate /boot partition
On Leap 15.1, booting fails: https://openqa.opensuse.org/tests/1386592

This is because `/EFI/boot/grub.cfg` on the ESP contains

```
set prefix=(${root})//grub2
configfile (${root})/boot/grub2/grub.cfg
```

As `$root` points to the boot partition here, the `/boot` part is wrong. grub2-install itself uses `source "${prefix}/grub.cfg"`, which should work here as well AFAICT.

davidcassany:
This commit makes sure that the early boot configuration files
for grub make use of the proper boot path and omitting the `/boot`
prefix if there is a dedicated boot partition.

Fixes #1553

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -888,8 +888,8 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 )
             )
             early_boot.write(
-                'configfile ($root)/boot/{0}/grub.cfg{1}'.format(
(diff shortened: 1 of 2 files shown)


schaefi (approved):
Uh yes my bad, good catch :+1:

Vogtinator:
Any particular reason why you don't just use `source "${prefix}/grub.cfg"`?

PR #1556, opened 2020-09-17
Change Appliance names to drop the name LimeJeOS

### LINKED ISSUE #1544: LimeJeOS should be replaced
## Problem description
There is a lot of LimeJeOS reference in the documentation and I think we should rename it openSUSE JeOS.

LimeJeOS is not used anymore by the openSUSE community, at least I don't find it in https://software.opensuse.org/distributions/leap but I see openSUSE JeOS images, AND no entry https://en.opensuse.org/LimeJeos but https://old-en.opensuse.org/LimeJeos, AND https://www.google.com/search?hl=en&q=limejeos%20suse is a time traveling to 2008.

## Expected behaviour
openSUSE JeOS instead of LimeJeOS

Since I'm more familiar with the regular SUSE doc structure and format, could you please let me know how should I proceed to replace LimeJeOS by openSUSE JeOS?

schaefi:
The name LimeJeOS was an invention of the SUSE Studio project.
Since the project does no longer exist, users have no idea
what the name means. Therefore the integration tests as well
as the documentation now changes the image names to provide
more clarity. This Fixes #1544

### DIFF
--- a/build-tests/x86/archlinux/test-image-iso-oem-vmx-kis/appliance.kiwi
+++ b/build-tests/x86/archlinux/test-image-iso-oem-vmx-kis/appliance.kiwi
@@ -3,11 +3,11 @@
 <!-- The line below is required in order to use the multibuild OBS features -->
 <!-- OBS-Profiles: @BUILD_FLAVOR@ -->
 
-<image schemaversion="7.3" name="LimeJeOS-Arch">
+<image schemaversion="7.3" name="Arch_appliance">
     <description type="system">
         <author>David Cassany</author>
         <contact>dcassany@suse.com</contact>
-        <specification>Arch JeOS</specification>
+        <specification>Arch Appliance</specification>
     </description>
     <profiles>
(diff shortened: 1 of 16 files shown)


schaefi:
@vmoutoussamy as I broke your changes I applied the patch from your forked project and created a new PR here.
Please note I have not applied the changes to the unit tests. I think it's not needed to change the unit tests as they are not exposed to users.

I also saw the documentation was changed. The names used here references the image descriptions in the following repository:

https://github.com/OSInside/kiwi-descriptions

Therefore before merge we also need to adapt that descriptions such that the documentation stays valid

Hope this is ok for you ?

Thanks

davidcassany (approved):
> Therefore before merge we also need to adapt that descriptions such that the documentation stays valid

Yes so we should merge OSInside/kiwi-descriptions#77 first and then this one.

schaefi:
yep the kiwi-descriptions PR is merged now

vmoutoussamy (approved):

PR #1561, opened 2020-09-21
Added consistency runtime check for the type setup

schaefi:
multiple type sections within one preferences section is allowed
in a kiwi image description. However, if multiple type sections
for the same image attribute are configured only the last type
configuration will be ever reachable. The proposed runtime check
in this commit detects this situation and raises an exception
showing the conflicting types including a solution suggestion
which needs to be based on profiles to distinguish between
types of the same image type name.

### DIFF
--- a/kiwi/runtime_checker.py
+++ b/kiwi/runtime_checker.py
@@ -20,6 +20,7 @@ import re
 from textwrap import dedent
 
 # project
+from io import StringIO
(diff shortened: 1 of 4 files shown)


davidcassany (approved):
Looks good to me! I like that the type sections appear on the message

schaefi:
Thanks

PR #1562, opened 2020-09-22
Cleanup grub adaptions code

### LINKED ISSUE #1527: Refactor grub config adaptions to be easy maintainable
## Problem description

In kiwi/bootloader/config/grub2.py::setup_disk_image_config() we have code to fix errors in the grub config due to
bugs in the grub tool chain when running in a loopback environment. We expect those extra code to go away when
the grub tools gets fixed. For us this means that extra code should be an easy read, as small as possible and also
easy to become deleted again. The current implementation has room for improvement in this regard

## Expected behaviour

refactor grub config patch code to be separated into private methods including good comments why we do this

schaefi:
The grub setup code has some after grub-mkconfig code that
fixes the written grub.cfg file on certain conditions. For
a better understanding and readability those conditions
and reasons are now put into private _fix* methods that
explains why we need to patch the written grub config file.
We all hope that those methods can go away when grub gets
fixed properly. This Fixes #1527

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -234,86 +234,15 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             ]
         )
 
-        if self.firmware.efi_mode():
-            # On systems that are configured to use EFI with grub2
-            # there is no support for dynamic EFI environment checking.
-            # In this condition we change the grub config to add this
-            # support as follows:
-            #
-            # * Apply on grub with EFI
-            #    1. Modify grub.cfg to set linux/initrd as variables
-            #    2. Prepend hybrid setup to select linux vs. linuxefi on demand
-            #
-            # Please note this is a one time modification done by kiwi
-            # Any subsequent call of the grub config tool will overwrite
-            # the setup and disables dynamic EFI environment checking
-            # at boot time
-            with open(config_file) as grub_config_file:
-                grub_config = grub_config_file.read()
-                grub_config = re.sub(
-                    r'([ \t]+)linux(efi|16)*([ \t]+)', r'\1$linux\3',
-                    grub_config
-                )
-                grub_config = re.sub(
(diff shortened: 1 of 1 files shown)


kiwi/bootloader/config/grub2.py
```diff
@@ -1123,3 +1052,101 @@ def _get_shim_install(self):
...
+                                'root={0}'.format(
+                                    boot_options.get('root_device')
+                                ),
+                                self.root_reference
+                            )
+                        with open(vendor_grubenv_file, 'w') as vendor_grubenv:
+                            vendor_grubenv.write(grubenv)
+
+    def _fix_grub_loader_entries_boot_cmdline(self):
+        if self.cmdline:
```
davidcassany, line 1123: This execution criteria is different than it was before, I guess this is intended. So we are now applying the patch to any existing `/boot/loader/entries/*.conf` file. I am fine with this change, in fact, seams to be simpler and easier to understand, I even would say that some cases not properly covered before.... :+1: for that change

davidcassany (approved):
Yes, way nicer to read now! :+1:

kiwi/bootloader/config/grub2.py
schaefi: yes I did that on purpose as I thought this is what we want as of now. The loader/entries files were so broken under several different conditions that I remember we agreed on just replacing the entire cmdline with the values we know that are correct. So the only criteria is if there is a cmdline bump it over all loader/entries

I'm happy when we can delete this though, as I think it should really work in grub independent of the host

schaefi: Thanks for the review. I'll merge and do a Staging commit. Would be great if we can check a fedora and ubuntu build if they work as before. Those are affected by the adaptions. On suse we only have the root dev update and the linuxefi vs. linux adaptions. I'll do some tests too.

Thanks

Issue #1498, opened 2020-07-01
Refactor Factory Classes


schaefi:
## Problem description

The way Factories are implemented in kiwi is from a strong typed API perspective
questionable and can be done more clear and stable like it was done in #1495 

In addition to the FileSystem factory the following other factory classes needs
adaption:

- [x] builder
- [x] boot/image
- [x] bootloader/config
- [x] bootloader/install
- [ ] container
- [ ] container/setup
- [x] filesystem
- [ ] iso_tools
- [x] markup
- [ ] oci_tools
- [x] package_manager
- [x] partitioner
- [x] repository
- [x] solver/repository
- [x] storage/subformat
- [x] system/root_import
- [x] volume_manager

schaefi:
Let's commit for the September sprint on the following:

- [ ] builder
- [ ] boot/image
- [ ] bootloader/config
- [ ] bootloader/install

schaefi:
Starting with builder today...

davidcassany:
Lets continue with

- [ ] solver/repository
- [ ] storage/subformat
- [ ] system/root_import
- [ ] volume_manager

for october/november

schaefi:
@davidcassany I'm taking over the first two now

schaefi:
Done for this sprint, moving it back to the backlog. We will touch base on this in the next sprint again

davidcassany:
Lets continue with:

- [ ] markup
- [ ] package_manager
- [ ] partitioner
- [ ] repository

schaefi:
Starting with the first two markup and package_manager

schaefi:
@davidcassany  Going to address the other two as well since you are busy with dracut code

schaefi:
done for this sprint. Next time we work on this we will complete it, yay

davidcassany:
Let's continue and finalize the task with:

- [ ] container
- [ ] container/setup
- [ ] iso_tools
- [ ] oci_tools
- [ ] package_manager

davidcassany:
I start to work on this last set of factory classes

PR #1563, opened 2020-09-22
Refactor ImageBuilder Factory

schaefi:
This commit refactors the ImageBuilder factory to be a real
factory and to add type hints such that its use from an api
perspective is clear and enforced. Related to Issue #1498

### DIFF
--- a/kiwi/builder/filesystem.py
+++ b/kiwi/builder/filesystem.py
@@ -39,23 +39,13 @@ class FileSystemBuilder:
     """
     **Filesystem image builder**
 
-    :param str label: filesystem label
-    :param str root_uuid: UUID of the created filesystem (on block device only)
-    :param str root_dir: root directory path name
+    :param obsject xml_state: Instance of :class:`XMLState`
     :param str target_dir: target directory path name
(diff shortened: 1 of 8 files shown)


kiwi/builder/filesystem.py
```diff
@@ -39,23 +39,13 @@ class FileSystemBuilder:
...
-    :param int blocksize: configured disk blocksize
-    :param object filesystem_setup: instance of :class:`FileSystemSetup`
-    :param object filesystems_no_device_node: List of filesystems which are
-        created from a data tree and do not require a block device e.g loop
-    :param dict filesystem_custom_parameters: Configured custom filesystem
-        mount and creation arguments
-    :param object result: instance of :class:`Result`
+    :param str root_dir: root directory path name
+    :param dict custom_args: Custom processing arguments defined as hash keys:
+        * None
```
davidcassany, line 46: Yes I like it :+1: I'll try follow this convention too, I also prefer adding only the parameters that are actually part of the init method rather than including some attributes too. This was likely to be confusing and probably not correctly maintained.

davidcassany (approved):
Looks good!

PR #1564, opened 2020-09-23
Followup fix to handle one disk type better

schaefi:
The vmx type is auto converted into an oem type with rootfs
resize disabled such that all disk images can be handled
under one disk type. However people who run kiwi on the
commandline and have selected --type vmx before now end
with an error message saying that there is no vmx type
because it was converted into an oem type. To handle this
more gracefully this commit changes the commandline
option --type vmx into --type oem if provided and prints
a warning message.

### DIFF
--- a/kiwi/cli.py
+++ b/kiwi/cli.py
@@ -65,6 +65,7 @@ global options for services: image, system
         image build type. If not set the default XML specified
         build type will be used
 """
+import logging
 import sys
 import os
 import pkg_resources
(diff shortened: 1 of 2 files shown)


schaefi:
@davidcassany I hope this completes the move to one disk type and allows our users to just continue as they were used to before

davidcassany (approved):
Makes sense :+1:

DattoDaveOps:
Worked for me. :) TY

PR #1565, opened 2020-09-23
Fix s390 offset setup for non dasd table

schaefi:
This change is two fold

**Fixed zipl targetoffset setup for non dasd devices**
    
It seems zipl expects a sector based information for the
targetoffset setting but kiwi provided blocks. This
Fixes bsc#1170863

**Fixed s390 Virtual disk integration tests**
    
Explicitly set 512byte block size for images that should run
in KVM. The default blocksize set by kiwi is 4k on s390


schaefi:
After testing on s390 I see we need to completely refactor this. Closing this PR and will come up with a new one

PR #1566, opened 2020-09-23
Refactor BootImage Factory

schaefi:
This commit refactors the BootImage factory to be a real
factory and to add type hints such that its use from an api
perspective is clear and enforced. Related to Issue #1498


davidcassany (approved):
Looks good :+1:

PR #1567, opened 2020-09-25
Fix/Refactor s390 support

schaefi:
This changes the s390 support on several stages:
    
1) On s390 the boot process is based on zipl which boots into an
     initrd from which a userspace grub process is started to support
     the grub capabilities. The implementation of this concept is
     provided via the grub2-s390x-emu package. Once installed the
     setup of the bootloader is done via the grub2-mkconfig and
     grub2-install commands and therefore from a caller perspective
     the same as with any other grub2 setup process. For kiwi this
     means no extra zipl bootloader target code is needed. Therefore
     this commit deletes the zipl setup from kiwi and puts on
     the standard grub2 process. This Fixes bsc#1170863
    
2) To support different targettypes the grub2-s390x-emu provided
     zipl template must be adapted. Parts of the former zipl bootloader
     setup therefore now applies to an update of the zipl2grub
     template file
    
3) Support for CDL/LDL DASD targets has been disabled in the schema
     When testing 4k devices and a respective zipl2grub template
     setup for CDL/LDL targettype it has turned out that grub2-install
     is not able to run on such a device. My assumption is that
     the device code in grub2-install does not work for 4k devices
     with an fdasd created partition table. As this needs further
     investigations and most probably adaptions on the grub toolchain
     for s390, we disabled the setup of these modes for now.
     emulated DASD (FBA) and SCSI targets stays supported.


schaefi:
I have done extensive testing with that code on the s390 test machines. So far things are working but the following issues needs to be discussed. I plan to ask Raymund Will as he knows much about the grub2-s390x-emu package

* How to setup zipl options, e.g targettype
  It seems to be needed to manually adapt /etc/default/zipl2grub.conf.in

* Options set in /etc/default/grub seems to be respected but I'm not sure. It looks like the tooling
  takes the values from /etc/default/grub and replaces it in the template from /etc/default/zipl2grub.conf.in. Is that correct ?

* If an extra boot partition is used the userspace grub failed
  If I just for fun setup a system that has ```bootpartition="true"``` the zipl initrd comes up but the system root seems to be
  not found. Even though the provided root= information is correct

schaefi:
I sent Raymund an e-mail and will come back to this PR once we clarified all open questions

schaefi:
I had a very good conversation with Raymund. Here is the result:

* How to setup zipl options, e.g targettype
  This is only possible if kiwi adapts the template file /etc/default/zipl2grub.conf.in prior calling grub2-install
  I will update this PR to achieve this

* Extra boot partition causes the boot to fail
  This is not a kiwi problem but a problem of the code in ```/usr/sbin/grub2-zipl-setup```
  The code there uses grub2-probe to detect extra boot partitions and most problably this fails in the loop based
  environment of a kiwi image build process. To support this case a fix to grub2-zipl-setup would be needed that works in any environment. e.g based on lsblk. At the moment we will not do any changes in this area as there is also no customer request or open business case. In general an extra /boot partition is unwanted these days

schaefi:
Finally building and booting

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_s390:sle15/test-image-disk

schaefi:
@davidcassany I think you will have a hard time to review this as it's pretty s390 specific and I have done all my tests on an s390 machine. Unless you see something super stupid that has potential to break non s390 code I would just merge it now, ok ?

schaefi:
For completeness I summarize the still open issues:

* Extra boot partition causes the boot to fail
  This is not a kiwi problem but a problem of the code in /usr/sbin/grub2-zipl-setup
  The code there uses grub2-probe to detect extra boot partitions and most problably this fails in the loop based
  environment of a kiwi image build process. To support this case a fix to grub2-zipl-setup would be needed that works in any 
  environment. e.g based on lsblk. At the moment we will not do any changes in this area as there is also no customer request or 
  open business case. In general an extra /boot partition is unwanted these days

* CDL/LDL DASD disks prevents grub2-install to work
  when using a 4k loop disk device partitioned with fdasd and respective zipl conf to support the CDL/LDL the call of grub2-install fails because the grub2-install code cannot find a grub device. This is also not a kiwi problem but a problem in grub2 on s390. We just disabled this mode in the kiwi schema for the moment

* Factory dracut zdev is broken
  Building s390 images on factory ends with a segmentation fault issued when dracut runs the zdev module code. Also not a kiwi problem and I have sent details to Mike and Ihno

So under certain conditions s390 images works :)

davidcassany (approved):
Indeed this includes quite some s390 specific stuff. Is good to see zipl classes gone and manage everything from gurb2. Looks good to me, I have not seen anything in the code that looks wrong, suspicious or a bad practice. About the config and install procedure I just trusted the steps you wrote.

schaefi:
Thanks much David for looking at it :+1:

PR #1568, opened 2020-09-26
Add support for package changelog info

schaefi:
In addition to the .packages file which shows details about
the installed packages in terms of version, license, etc...
we now also create a .changes file which contains the changelog
information of the installed packages. The file can be used
to compare the package changelogs between image builds.

### DIFF
--- a/kiwi.yml
+++ b/kiwi.yml
@@ -25,6 +25,10 @@
 #  # Specify if the bundle tarball should contain compressed results.
 #  # Note: already compressed result information will not be touched
 #  - compress: false
+#  # Specify if the image build result and bundle should contain
+#  # a .changes file. The .changes file contains the package changelog
+#  # information from all packages installed into the image.
+#  - has_package_changes: false
 
 
 # Setup behaviour of XZ compressor
--- a/kiwi/builder/archive.py
+++ b/kiwi/builder/archive.py
@@ -106,6 +106,15 @@ class ArchiveBuilder:
                 shasum=False
             )
             self.result.add(
+                key='image_changes',
+                filename=self.system_setup.export_package_changes(
+                    self.target_dir
+                ),
+                use_for_bundle=True,
+                compress=True,
+                shasum=False
+            )
+            self.result.add(
                 key='image_verified',
                 filename=self.system_setup.export_package_verification(
                     self.target_dir
--- a/kiwi/builder/container.py
+++ b/kiwi/builder/container.py
@@ -157,6 +157,15 @@ class ContainerBuilder:
             shasum=False
         )
         self.result.add(
+            key='image_changes',
+            filename=self.system_setup.export_package_changes(
+                self.target_dir
+            ),
+            use_for_bundle=True,
+            compress=True,
+            shasum=False
+        )
+        self.result.add(
             key='image_verified',
             filename=self.system_setup.export_package_verification(
                 self.target_dir
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -515,6 +515,15 @@ class DiskBuilder:
             shasum=False
         )
         self.result.add(
+            key='image_changes',
+            filename=self.system_setup.export_package_changes(
+                self.target_dir
+            ),
+            use_for_bundle=True,
+            compress=True,
+            shasum=False
+        )
+        self.result.add(
             key='image_verified',
             filename=self.system_setup.export_package_verification(
                 self.target_dir
--- a/kiwi/builder/filesystem.py
+++ b/kiwi/builder/filesystem.py
@@ -139,6 +139,15 @@ class FileSystemBuilder:
             shasum=False
         )
         self.result.add(
+            key='image_changes',
+            filename=self.system_setup.export_package_changes(
+                self.target_dir
+            ),
+            use_for_bundle=True,
+            compress=True,
+            shasum=False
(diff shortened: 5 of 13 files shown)


schaefi:
@davidcassany this add the additional changes info file. I noticed on Arch the people don't use changelog information in their packages (unbelievable). The pacman manager allows to have changes info and you can also query it, but it's always empty. Thus no info for Arch

schaefi:
@davidcassany I think I also found a small bug in the pacman call that reads the packages from the system. The way this was done looked liked it is reading the system package database but not the image database. Also the "-e" option shows only explicitly listed packages which ends up in a small list. I've modified the call as I see fit. Would be great if you can look at that change as well. Thanks much

schaefi:
I did a test and manually changed the .changes file by deleting some changelog entries from the top of a few packages. I think due to the policy the history must be chronological which means a diff should take the package name as anchor into account. The result looked like this:

```patch
--- Leap-15.1_appliance.x86_64-1.15.1.changes   2020-09-27 16:04:55.692524114 +0200
+++ Leap-15.1_appliance.x86_64-1.15.1.changes.orig      2020-09-27 16:03:40.652649285 +0200
@@ -1,4 +1,13 @@
 wicked|
+* Fri Apr 05 2019 mt@suse.de
+- version 0.6.54
+- switch to use systemd notify and prevent event backlog at start
+  by calling udevadm settle before starting wickedd (bsc#1118206)
+- dhcp6: don't discard confirm reply without status (bsc#1127340)
+- ethtool: set lro legacy flag and not txvlan (bsc#1123555)
+- init memory before use in ioctl
+- fsm: fix find pending worker loop segfault (boo#1106809)
+
 * Thu Jan 24 2019 rtorreromarijnissen@suse.de
 - version 0.6.53
 - dhcp: request hostname/fqdn option in the tester (bsc#1118378)
@@ -496,6 +505,15 @@
   nanny (bsc#900951)

 cracklib-dict-full|
+* Thu Aug 09 2012 mc@suse.de
+- correct License to LGPL-2.1
+
+* Sun Sep 18 2011 jengelh@medozas.de
+- Remove redundant tags/sections from specfile
+
+* Mon Sep 01 2008 mc@suse.de
+- do not zip the dict by default. A check takes too long.
+
 * Fri Jun 27 2008 mc@suse.de
 - update wordlist
 - add german wordlists
@@ -510,6 +528,10 @@
 - full cracklib dictionary - splitted out from cracklib

 fonts-config|
+* Thu Nov 23 2017 rbrown@suse.com
+- Replace references to /var/adm/fillup-templates with new
+  %_fillupdir macro (boo#1069468)
+
 * Wed Sep 21 2016 ftake@geeko.jp
 - version 20160921
   * add 59-family-prefer-lang-specific.conf to specify prefered
```

So I guess a simple diff between two .changes files should provide a useful information. The diff output will become not so easy readable if there are changes inside of the history of changes and not only on the top of the changelog info

Thoughts ?

davidcassany:
> @davidcassany this add the additional changes info file. I noticed on Arch the people don't use changelog information in their packages (unbelievable). The pacman manager allows to have changes info and you can also query it, but it's always empty. Thus no info for Arch

Yes I noted that in Arch changelogs are likely to be unset. In fact when I created the package it took me a while to figure out how include the changelog as all the examples I had at hand did not include any kind of changelog...

> @davidcassany I think I also found a small bug in the pacman call that reads the packages from the system. The way this was done looked liked it is reading the system package database but not the image database. Also the "-e" option shows only explicitly listed packages which ends up in a small list. I've modified the call as I see fit. Would be great if you can look at that change as well. Thanks much

Urks... Good chatch, yes looks like I forgot the chroot or to point explicitly to the image database path.

> So I guess a simple diff between two .changes files should provide useful information. The diff output will become not so easy readable if there are changes inside of the history of changes and not only on the top of the changelog info. Due to the chronological order requirement imho all distributions apply this should not be so much of a problem

Yes thats a good point, at least in openSUSE there is a clear criteria about having changelogs that append new data at top and avoid rewrites of history. Even though history changes are accepted when there is a clear justification and they are done in a clear and understandable way, this can easily happen when trying to sync SLE packages with TW.

davidcassany (approved):
Looks good to me! Thanks for the pacman fixes :+1:

schaefi:
Since obs plans to provide a .report file (remember the .packages conversation) we should make this a config option with the support for it disabled when running inside obs and enabled when running outside

schaefi:
I've added a runtime config value to setup the .changes creation

```yaml
bundle:
  - has_package_changes: true|false
```

The default when building in obs is: false
The default when building outside obs is: true

schaefi:
@davidcassany final review :) Thanks

davidcassany (approved):
LGTM I understand that better not mess with OBS output without having a clear and explicit conversation with autobuild, but I'd definitely like to see this file being part of the output in OBS.

schaefi:
once this is merged and the feature available in SLE I will have this conversation with obs, especially if they don't come up with the solution forecasted. Currently there is much in flux and even if we create the .changes it will not be part of the published results. I remember the discussion that obs should publish all data that comes from the "kiwi result bundle" call, but that was refused because it would allow another project (kiwi) to influence publishing results. I found the lack of trust at this point unfortunate but it has to be respected. 

Thanks for looking at this again, will merge it now

PR #1569, opened 2020-09-28
Refactor BootLoaderConfig

davidcassany:
This commit refactors BootLoaderConfig to turn it into a proper factory
class and to also include type hints to facilitate it's use from an API
POV.

Related to #1498

### DIFF
--- a/kiwi/bootloader/config/__init__.py
+++ b/kiwi/bootloader/config/__init__.py
@@ -15,17 +15,19 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
-# project
-from kiwi.bootloader.config.grub2 import BootLoaderConfigGrub2
-from kiwi.bootloader.config.isolinux import BootLoaderConfigIsoLinux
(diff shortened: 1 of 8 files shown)


kiwi/bootloader/config/__init__.py
```diff
@@ -34,22 +36,35 @@ class BootLoaderConfig:
     :param string root_dir: root directory path name
     :param dict custom_args: custom bootloader config arguments dictionary
     """
-    def __new__(
-        self, name, xml_state, root_dir, boot_dir=None, custom_args=None
+    @abstractmethod
+    def __init__(self) -> None:
+        return None                     # pragma: no cover
```
schaefi, line 41: This is nice, didn't know that way of skipping the code coverage. I have added it to the kiwi/test/unit/.coveragerc under report/omit. But your solution is better :+1:

schaefi (approved):

kiwi/bootloader/config/__init__.py
davidcassany: Yes I did some trials to avoid it having to omit any line, but I had no success :P I was also not aware of it way of omitting codes I discovered it yesterday trying to figure out some way to handle the abstract methods coverage.

PR #1573, opened 2020-09-29
Fix compat link for rpmdb location

davidcassany:
This commit fixes the symlink creation for `/var/lib/rpm`. More specific
for derived container images in which the base root tree already
included the `/var/lib/rpm` the link, the `ln` command was creating a
symlink inside the `/var/lib/rpm` folder givent that it was following
the already existing symlink. Adding the `--no-target-directory` force
`ln` command to treat `/var/lib/rpm` path as the fully qualified symlink name.

Fixes bsc#1176977

### DIFF
--- a/kiwi/repository/zypper.py
+++ b/kiwi/repository/zypper.py
@@ -201,7 +201,8 @@ class RepositoryZypper(RepositoryBase):
             )
             Command.run(
                 [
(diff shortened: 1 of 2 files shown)


Vogtinator:
LGTM!

schaefi (approved):
Thanks, looks great :+1:

PR #1574, opened 2020-10-02
Align dracut-kiwi-lib version with packages requiring it

### LINKED ISSUE #1529: Strict versioned RPM dependency on dracut-kiwi-lib for dracut-kiwi-oem-* packages
Some of the dracut-kiwi-* RPM packages currently have a simple "dracut-kiwi-lib" dependency, without specifying exact version of the package. 

During an image build, this would pull in the dracut-kiwi-lib package of any version, should there be multiple versions of this package available in configured repositories (usually highest version available), thus potentially breaking dracut-kiwi functionality with incompatible code.

The RPM packages which depend on dracut-kiwi-lib should therefore specify the exact version of the lib package in their dependencies instead.

davidcassany:
This commit enforces dracut-kiwi-oem-repart and
dracut-kiwi-oem-dump to require dracut-kiwi-lib of the same exact
version. This prevents dracut-kiwi-lib and the packages
dependent on it being installed on a image with inconsistent versions.

Fixes #1529

### DIFF
--- a/package/python-kiwi-pkgbuild-template
+++ b/package/python-kiwi-pkgbuild-template
@@ -42,14 +42,14 @@ package_dracut-kiwi-lib(){
 }
 
 package_dracut-kiwi-oem-repart(){
-  depends=(dracut-kiwi-lib)
+  depends=(dracut-kiwi-lib=${pkgver})
   cd kiwi-${pkgver}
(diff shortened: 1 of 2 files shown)


schaefi:
Looks good to me :+1:

PR #1576, opened 2020-10-05
Delete length limitation of image id attribute

schaefi:
For legacy reasons the <image id="..."/> attributes was
limited to 10digits. The contents of /etc/ImageID are now
free format and no longer strictly evaluated. Thus the
limitations on the id attribute can be deleted


davidcassany (approved):
LGTM

PR #1577, opened 2020-10-05
Make dependencies to dracut-kiwi-lib release specific

davidcassany:
This commit adds a dracut-kiwi-lib dependency to dracut-kiwi-oem-dump and
darcut-kiwi-oem-repart to match up to the release level. This way the
dependency ensures the pulled binaries they are all part of the same build.


Conan-Kudo (approved):

schaefi (approved):
perfect :+1: Thanks

PR #1578, opened 2020-10-05
Use custom tmpfs for managing overlays

### LINKED ISSUE #1558: Overlayfs write folder is placed in path with limited resources
## Problem description

Apparently the `/run/overlayfs` is not a good folder for storing the overlayfs in Live systems as systemd limits the capacity of `/run` folder to 20% of the overall RAM. 

This issue tracks [boo#1176709](https://bugzilla.opensuse.org/show_bug.cgi?id=1176709):

> 
> openQA test in scenario opensuse-Tumbleweed-GNOME-Live-x86_64-gnome-live@64bit fails in
> [zypper_in](https://openqa.opensuse.org/tests/1398257/modules/zypper_in/steps/12)
> 
> Kiwi stores the overlayfs in /run/overlayfs - which, taking systemd 246 into account, is not a good place.
> 
> systemd 246 limits /run to 20% of the RAM, under the premise that 'run' is not supposed to host large files, but should only reference small runtime-related files. /tmp is limited to 50% of RAM (as it used to be)
> 

## Expected behaviour

Overlyfs is storage is not that scrictly constrained on memory, 20% seams to be clearly limited for live images such desktop environments.

Can be seen on resource hungry Live images such as Tumbleweed GNOME or such with 2GB of RAM.

schaefi:
Provide a method to manage the base mount point for overlay and iso mount points and set the base directory to it because /run
has a RAM space limitation which is unfortunate to handle live systems. This Fixes #1558

### DIFF
--- a/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
@@ -1,6 +1,20 @@
 #!/bin/bash
 type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
 
+function getOverlayBaseDirectory {
+    # initialize and print overlay base directory below which
+    # the iso and overlayfs based mountpoints are managed
+    local overlay_base=/run/overlay
+    local overlay_size
+    mkdir -p "${overlay_base}"
+    if ! mountpoint -q "${overlay_base}";then
+        overlay_size=$(getarg rd.live.overlay.size)
+        [ -z "${overlay_size}" ] && overlay_size="50%"
+        mount -t tmpfs -o "size=${overlay_size}" tmpfs "${overlay_base}"
+    fi
+    echo "${overlay_base}"
+}
+
 function lookupIsoDiskDevice {
     local root=$1
     local iso_label=${root#/dev/disk/by-label/}
@@ -100,8 +114,10 @@ function initGlobalOptions {
 }
 
 function mountIso {
-    ln -s "${isodev}" /run/initramfs/isodev
-    local iso_mount_point=/run/initramfs/live
+    local overlay_base
+    overlay_base=$(getOverlayBaseDirectory)
+    ln -s "${isodev}" "${overlay_base}/isodev"
+    local iso_mount_point="${overlay_base}/live"
     mkdir -m 0755 -p "${iso_mount_point}"
     if ! mount -o ro -n -t "${isofs_type}" "${isodev}" "${iso_mount_point}"; then
         die "Failed to mount live ISO device"
@@ -111,7 +127,9 @@ function mountIso {
 
 function mountCompressedContainerFromIso {
     local iso_mount_point=$1
-    local container_mount_point=/run/initramfs/squashfs_container
+    local overlay_base
+    overlay_base=$(getOverlayBaseDirectory)
+    local container_mount_point="${overlay_base}/squashfs_container"
     squashfs_container="${iso_mount_point}/${live_dir}/${squash_image}"
     mkdir -m 0755 -p "${container_mount_point}"
     if ! mount -n "${squashfs_container}" "${container_mount_point}";then
@@ -122,8 +140,10 @@ function mountCompressedContainerFromIso {
 
 function mountReadOnlyRootImageFromContainer {
     local container_mount_point=$1
+    local overlay_base
+    overlay_base=$(getOverlayBaseDirectory)
     local rootfs_image="${container_mount_point}/LiveOS/rootfs.img"
-    local root_mount_point=/run/rootfsbase
+    local root_mount_point="${overlay_base}/rootfsbase"
     mkdir -m 0755 -p "${root_mount_point}"
     if ! mount -n "${rootfs_image}" "${root_mount_point}"; then
         die "Failed to mount live ISO root filesystem"
@@ -132,15 +152,19 @@ function mountReadOnlyRootImageFromContainer {
 }
 
 function prepareTemporaryOverlay {
-    mkdir -m 0755 -p /run/overlayfs/rw
-    mkdir -m 0755 -p /run/overlayfs/work
+    local overlay_base
+    overlay_base=$(getOverlayBaseDirectory)
+    mkdir -m 0755 -p "${overlay_base}/overlayfs/rw"
+    mkdir -m 0755 -p "${overlay_base}/overlayfs/work"
 }
 
 function preparePersistentOverlay {
     if [ -z "${isodiskmode}" ] || [ -z "${isodiskdev}" ]; then
         return 1
     fi
-    local overlay_mount_point=/run/overlayfs
+    local overlay_base
+    overlay_base=$(getOverlayBaseDirectory)
+    local overlay_mount_point="${overlay_base}/overlayfs"
     mkdir -m 0755 -p "${overlay_mount_point}"
     if [ "${isodiskmode}" = "disk_boot" ];then
         if ! preparePersistentOverlayDiskBoot "${overlay_mount_point}"; then
@@ -151,8 +175,8 @@ function preparePersistentOverlay {
             return 1
         fi
     fi
-    mkdir -m 0755 -p ${overlay_mount_point}/rw
-    mkdir -m 0755 -p ${overlay_mount_point}/work
+    mkdir -m 0755 -p "${overlay_mount_point}/rw"
+    mkdir -m 0755 -p "${overlay_mount_point}/work"
 }
 
 function preparePersistentOverlayLoopBoot {
--- a/doc/source/concept_and_workflow/customize_the_boot_process.rst
+++ b/doc/source/concept_and_workflow/customize_the_boot_process.rst
@@ -189,6 +189,14 @@ the available kernel boot parameters for this modules:
   OS deployment is `/dev/sdj`. With `rd.kiwi.oem.maxdisk=500G` the
   deployment will land on that RAID disk.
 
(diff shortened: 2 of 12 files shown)


schaefi:
I hope I catch all code parts and am still testing. This is a sensitive part and covers only the live system management

There is also the overlayfs support for disk images which is handled in ./90kiwi-overlay/kiwi-overlay-root.sh and not part of this PR

davidcassany:
> There is also the overlayfs support for disk images which is handled in ./90kiwi-overlay/kiwi-overlay-root.sh and not part of this PR

Is there any reason for that? The former issue will also raise in this module if anyone makes use of it. Is this because live module and the overlay module do not have lib dependency to share their overlayfs defaults?

schaefi:
> Is there any reason for that?

no specific reason, I just wanted to make live iso to work and if that works as expected I would just copy the change to the other lib. I'll mark the PR as wip, ok ?

davidcassany (approved):
LGTM, I also have not found any other references to `/run` we should care about. I was wondering about the dump image methods but I see only metadata, kernel and initrd are stored in `/run`, I think it is fair to have them there, they should not be too big.

Looks good to me, probably I would have included the overlayfs-root module change into this PR, but I am also fine by doing it in a followup PR.

schaefi:
I'm still testing, something tells me I overlooked something :)

schaefi:
of course it does not work :/ ...still investigating

schaefi:
I have no idea what the hell is going on here, the simple change broke everything and I don't understand it. It looks like code gets called twice for whatever reason. Please ignore commits for the moment I'm just trying step by step what in God's name could cause such a breakage by a simple path change...

davidcassany:
> Please ignore commits for the moment I'm just trying step by step what in God's name could cause such a breakage by a simple path change...

Sure, just let me know if at some point you want another pair of eyes

schaefi:
Ok apart from that I'm completely stupid there is also a thinking error. /tmp is not necessarily a tmpfs. On my system it's not and by default it's also not a tmpfs. We can't trust this to be present. If we overmount /tmp when it's not a tmpfs very very strange things on the liveOS rootfs happens. /run is handled by dracut/systemd and can be trusted to exist. So we need to create our own tmpfs space. I will check on that

schaefi:
On the right track now, but not complete

schaefi:
@davidcassany finally I'm finished with this. Tested live iso in hybrid, iso persistent, non persistent mode and also build a disk image with overlayroot="true" and tested that. I couldn't resist to add a small feature rd.kiwi.overlay.readonly which allows to boot an overlay disk in non-persistent-mode

Would be great if you can give it a short review and maybe a test build, just in case I forgot something

Thanks much

dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
```diff
@@ -1,6 +1,17 @@
 #!/bin/bash
 type getarg >/dev/null 2>&1 || . /lib/dracut-lib.sh
 
+function getOverlayBaseDirectory {
+    # initialize and print overlay base directory below which
+    # the iso and overlayfs based mountpoints are managed
+    local overlay_base=/overlay
+    mkdir -p "${overlay_base}"
+    if ! mountpoint -q "${overlay_base}";then
+        mount -t tmpfs tmpfs "${overlay_base}"
```
davidcassany, line 10: I am wondering if it would make sense to customize the tmpfs constraints for the non persistent overlayfs from the kernel command line, in fact, it means only adding the `-o size=<custom_value>` here. The default value is  `size=50%`. 

Probably something like `rd.kiwi.overlay.size=`. Also it could eventually be meaningful for persistent overlayfs. Anyways this is an enhancement not strictly related to the issue being solved in this PR.

davidcassany (approved):
Looks good :+1:  I am running a couple of builds to test it. Gonna test a VM with overlay and a Live image.

dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
schaefi: very good point. Yes let's add that

schaefi:
@davidcassany I also realized the mountpoint used for the tmpfs is hidden in the system later because it's not been moved to sysroot/ (mount --move). Do you think this is a problem ? I mean there is nothing you could do with the active mounts as they are busy anyway. So maybe this is from a protection point of view even a good setting but I'm not sure.

Thoughts ?

davidcassany:
> @davidcassany I also realized the mountpoint used for the tmpfs is hidden in the system later because it's not been moved to sysroot/ (mount --move). Do you think this is a problem ? I mean there is nothing you could do with the active mounts as they are busy anyway. So maybe this is from a protection point of view even a good setting but I'm not sure.
> 
> Thoughts ?

This is interesting, in fact I am having issues with the overlay module, I can't boot on persistent mode... I am wondering if this is related. I am just starting to have a closer look

davidcassany:
> > @davidcassany I also realized the mountpoint used for the tmpfs is hidden in the system later because it's not been moved to sysroot/ (mount --move). Do you think this is a problem ? I mean there is nothing you could do with the active mounts as they are busy anyway. So maybe this is from a protection point of view even a good setting but I'm not sure.
> > Thoughts ?
> 
> This is interesting, in fact I am having issues with the overlay module, I can't boot on persistent mode... I am wondering if this is related. I am just starting to have a closer look

Forget it, I did stupid tests, I forgot to apply the dracut changes on the image root-tree, I was no actually testing the right code :man_facepalming:

dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
schaefi: done. I used rd.live.overlay.size as this matches the other available options. For the overlayroot options I chose rd.root.overlay.size

I also updated the docs

davidcassany:
@schaefi Not sure what are you working on right now exaclty, but I realized that the VMs with a persistent overlay are broken, and I bet they were already broken before this PR. So far I noted we are not using the appropriate command line arguments to make it work. I fear this got broken at some point while refactoring grub in last months. The point is that the overlay module expects `root=overlay:UUID=uuid`. However I am currently getting `root=PARTUUID=b9d37ebb-7f4a-452e-a0b9-8768342b523a`. I'd say this root argument is provided by grub tools. 

I am wondering since how long the overlayroot is broken... :disappointed: I'd say the fix should happen mostly within the overlay module and not grub, since we can't easily control the command line grub is going to create. Probably add a new parameter `rd.kiwi.overlay.persistentpart=UUID` ? Dynamically discover it from the LABEL?

davidcassany:
Just confirmed that our integration test for overlayroot does not boot. Thus the issue I found is previous to this PR. 

@schaefi does it makes sense to work on it on a separate issue and PR? From this current PR POV I'd say that having the non presistent overalyroot support working is enough.

schaefi:
Thanks much for testing this. I bet it has something to do with that we build in obs and apply that root rewrite which is wrong when running on an obs worker without udev. I'll check it. Thanks

schaefi:
> Just confirmed that our integration test for overlayroot does not boot. 

yes the problem is grub2-mkconfig now writes PARTUUID reference instead of /dev/mapper/loop... therefore the matching in _fix_grub_root_device_reference does not apply. We need to make this more generic. I agree this should be done in a new PR

davidcassany:
> yes the problem is grub2-mkconfig now writes PARTUUID reference instead of /dev/mapper/loop... therefore the matching in _fix_grub_root_device_reference does not apply. We need to make this more generic. I agree this should be done in a new PR

yes agree, lets discuss this issue in separate PR, I still can't see how it is causing a problem in OBS and in my localhost in the very same way (does now grub2 resolves to something different than the loop device in OBS?).

schaefi:
> does now grub2 resolves to something different than the loop device in OBS?

this is what I assume

Vogtinator:
This was merged but it's still failing in openQA. Missing submission to openSUSE:Factory?

schaefi:
v9.21.19 is on its way to Virt:App:Builder that should auto kick the submission to factory. This was delayed as we are currently working on running functional tests from the integration test builds in openQA.

Vogtinator:
Any news here? There's still no submission.

schaefi:
I've done the submission some time ago to Virt:App:Builder. Meanwhile v9.21.20 was released. I thought this auto submits to Factory ? Let us know if a manual submission is required. Thanks

Vogtinator:
> I've done the submission some time ago to Virt:App:Builder. Meanwhile v9.21.20 was released. I thought this auto submits to Factory ?

It looks like you commited directly instead of submitting + forwarding. The only autosubmission in that case is after a week of inactivity.

> Let us know if a manual submission is required. Thanks

There's no submission pending: https://build.opensuse.org/package/requests/openSUSE:Factory/python-kiwi

schaefi:
> The only autosubmission in that case is after a week of inactivity.

a week of inactivity is required ? that's new to me. I thought auto submission from the devel project to Factory depends on content change and successful build results.

Anyway did it manually now

* created request id 845379

Vogtinator:
> a week of inactivity is required ? that's new to me. I thought auto submission from the devel project to Factory depends on content change and successful build results.

That was implemented back in 2012: https://github.com/openSUSE/obs-autosubmit/commit/77f29e220922d0215f1ef769ae3de90795cc17da#diff-8c22f29b9cb32a5f7bcea3ca5d5bb92e5a4b88ad7add193ee5cdc056d73bf97b

PR #1580, opened 2020-10-08
Enhance scope of _fix_grub_root_device_reference

schaefi:
In addition to the wrong root=/dev/mapper/loop... reference
fixing, written by grub2-mkconfig when used in obs there is
also the case that grub2-mkconfig writes root=PARTUUID which
is also unwanted and needs fixing.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -1142,10 +1142,19 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 # the correct value.
                 with open(config_file) as grub_config_file:
                     grub_config = grub_config_file.read()
-                    grub_config = grub_config.replace(
-                        'root={0}'.format(boot_options.get('root_device')),
-                        self.root_reference
+                    # The following expression matches any of the following
+                    # grub mkconfig root= settings and replaces it with a
+                    # correct value
+                    # 1. root=LOCAL-KIWI-MAPPED-DEVICE
+                    # 2. root=.*(lazy)=ANY-LINUX-BY-ID-VALUE
+                    grub_config = re.sub(
+                        r'(root=.*?=[a-zA-Z0-9:\.-]+)|(root={0})'.format(
+                            boot_options.get('root_device')
+                        ),
+                        '{0}'.format(self.root_reference),
+                        grub_config
                     )
+
                 with open(config_file, 'w') as grub_config_file:
                     grub_config_file.write(grub_config)
 
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -542,7 +542,7 @@ class TestBootLoaderConfigGrub2:
             file_handle_grubenv = \
                 mock_open_grubenv.return_value.__enter__.return_value
 
-            file_handle_grub.read.return_value = 'root=rootdev'
+            file_handle_grub.read.return_value = 'root=PARTUUID=xx'
             file_handle_grubenv.read.return_value = 'root=rootdev'
             file_handle_menu.read.return_value = 'options foo bar'
 
@@ -576,7 +576,7 @@ class TestBootLoaderConfigGrub2:
                     '    fi\n'
                     'fi\n'
                 ),
-                call('root=rootdev'),
+                call('root=PARTUUID=xx'),
                 # second write of grub.cfg, setting overlay root
                 call('root=overlay:UUID=ID')
             ]


schaefi:
This is my first suggestion and I haven't tested it in obs... will submit a staging build to run the tests

kiwi/bootloader/config/grub2.py
```diff
@@ -1146,6 +1146,11 @@ def _fix_grub_root_device_reference(self, config_file, boot_options):
                         'root={0}'.format(boot_options.get('root_device')),
                         self.root_reference
                     )
+                    grub_config = re.sub(
+                        r'root=PARTUUID=[a-z0-9-]+',
+                        r'{0}'.format(self.root_reference),
+                        grub_config
+                    )
```
davidcassany, line 1153: Rather than increasing the cases we try to guess it I am wondering if it wouldn't be better to apply one of the two ideas I had:

1. Always apply our `root=` for the OBS and/or overlay builds
2. Set some strategy to discover the write partition in https://github.com/OSInside/kiwi/blob/6cdc6bbdd350120d6984e89f058afaa160e6d72d/dracut/modules.d/90kiwi-overlay/kiwi-overlay-root.sh#L23 and forget about having to parse `root=overlay:UUID`. That would make overlay module transparent to grub2-mkconfig

schaefi: I also thought about always replacing root= with the correct value when we run in obs. That will make the code easier but we will then always patch the mkconfig result... which could live with to be honest

The second approach I'm not sure about. The reason for root=overlay= is because there is a dracut module that handles this special setup. So we need a trigger for it and as it works nicely I don't think the lack of obs-worker capabilities servers the purpose for a design change

If ok I would go with 1)

Thoughts ?

PS:

This patch here worked, for me. The current overlay based images in the integration tests now boots again. Could double check on them too and maybe give a standard image (no overlay) also a short test ?

Thanks

davidcassany: Also there is another alternative which a is a variant of 1). What about setting always the `root=` for the overlay case (within or not OBS) and then keep the OBS case exactly as it was. I'd say this is not very intrusive and also opens the door to the possibility of not touching the configuration inside OBS if unneeded (probably this is already the case in TW :man_shrugging: )

In any case I am fine with 1)

davidcassany: > This patch here worked, for me. The current overlay based images in the integration tests now boots again. Could double check on them too and maybe give a standard image (no overlay) also a short test ? 

I am giving a quick run

davidcassany: JFYI I did few boots of overlay VM and regular VM disks and they all successfully booted.

schaefi: Thanks much. I think I will think about our options a bit more and let this one rest until Monday (or maybe the weekend). Thanks for all your feedback, much appreciated.

schaefi:
@davidcassany I spent some time to think about our options and come up with one reqular expression to handle all cases (inline documented) I limited the scope of the runtime to the condition we already used (for overlayroot images and in obs). It would be possible to call this code in any case but I think it should not be opened up that way.

I still don't want to change the overlayroot detection in the dracut module just to workaround a grub weakness. So if there are no other reasons that makes a change in the dracut code eligible I'd like to go with the change from here

Thoughts ?

kiwi/bootloader/config/grub2.py
```diff
@@ -1142,10 +1142,19 @@ def _fix_grub_root_device_reference(self, config_file, boot_options):
...
+                    # The following expression matches any of the following
+                    # grub mkconfig root= settings and replaces it with a correct
+                    # value
+                    # 1. root=LOCAL-KIWI-MAPPED-DEVICE
+                    # 2. root=.*(lazy)=ANY-LINUX-BY-ID-VALUE
+                    grub_config = re.sub(
+                        r'(root=.*?=[a-zA-Z0-9:\.-]+)|(root={0})'.format(
+                            boot_options.get('root_device')
+                        ),
+                        r'{0}'.format(self.root_reference),
```
davidcassany, line 1154: nit: is this `r` before the string needed? I'd say no, this is not a regular expression.

davidcassany (approved):
I like this approach. Yes, I think it is a quite refined parsing and regular expression. It follows the spirit of patching always, but it does some checks on the expected syntax. Looks generic enough to not cause issues. I also agree that we should not open this patch to any kind of build, there is no need for that.

kiwi/bootloader/config/grub2.py
schaefi: nope, not needed, you are right. I change it

schaefi:
I think we are good to go now :)

schaefi:
I'll submit this to Staging and re-test the overlay

PR #1581, opened 2020-10-11
Fix profile docs

davidcassany:
This commit fixes the profiles documentation. The example KIWI-NG
command was using wrong flags order. This commit fixes the `--profile`
flag order in documentation.


schaefi:
Yes I guess this was my fault back in time. Thanks for catching it :+1:

PR #1584, opened 2020-10-13
Fixed strncopy in dcounter helper

### LINKED ISSUE #1579: dcounter option -l does not work
## Problem description

The recent change in dcounter breaks option handling:

https://github.com/OSInside/kiwi/blame/master/tools/dcounter.c#L25

```strncpy (prefix, *(argv+1), strlen(prefix));```

prefix is set to zeros, so strlen(prefix) is also zero, so nothing is copied.

schaefi:
The max size of the copy operation was always set to zero
because the strlen of an empty buffer is zero. Bad mistake
from my side :( This Fixes #1579

### DIFF
--- a/tools/dcounter.c
+++ b/tools/dcounter.c
@@ -6,12 +6,13 @@
 int main(int argc, char **argv) {
     int i, percent = -1;
     int newline = 0;
-    char prefix[512];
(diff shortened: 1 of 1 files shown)


davidcassany (approved):
LGTM

PR #1589, opened 2020-10-15
Fixed device_array structure in get_selected_disk

### LINKED ISSUE #1588: kiwi-dump-image.sh get_disk_list output is not parsed correctly in get_selected_disk
## Problem description
get_selected_disk() is used to display to the user all available disks but does only shows the first and fourth disks, skipping the second and third installed disks.

I believe this bug was introduced with commit 4780f444e535b129820d38977f44ea1f20ac4d71.  The list_item return value was changed from "disk size status" TO "disk size"
```
-     list_items="${list_items} ${disk_device} ${disk_size} ${item_status}"
-     item_status=off
+     list_items="${list_items} ${disk_device} ${disk_size}"
```

Testing has shown that changing $((count %3)) to $((count % 2)) in the below subroutine does resolve this issue.
```
        for entry in ${disk_list};do
            if [ $((count % 3)) -eq 0 ];then
                device_array[${device_index}]=${entry}
                device_index=$((device_index + 1))
            fi
            count=$((count + 1))
        done
```

list_items is a space separated list of the disks and size of the disk. e.g. "/dev/sda 500G /dev/sdb 250G /dev/sdc 750G /dev/sdd 50G"
Currently this routine is setting device_index= "/dev/sda 250G /dev/sdd"
The correct result should be device_index= "/dev/sda /dev/sbd /dev/sdc /dev/sdd"

## Expected behaviour
get_selected_disk() should correctly parse all disks in disk_list=$(get_disk_list)

This will happen on every system with two or more disks installed.

+ KIWI version: 9.20.5
+ Operating system: leap15.2
+ OBS version:

schaefi:
In reference to Issue #880 a bug was introduced that broke
the contents of the device_array which causes issues on
installations with two or more attached disks. The change
in the mentioned PR reduced the tuple for each disk
from 3 elements to 2 elements. Therefore the loop that
iterates over the disk tuples via modulo 3 was broken.
This commit fixes the modulo operation to correctly
parse the disk_list. Fixes #1588

### DIFF
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -137,7 +137,7 @@ function get_selected_disk {
         local count=0
         local device_index=0
         for entry in ${disk_list};do
-            if [ $((count % 3)) -eq 0 ];then
+            if [ $((count % 2)) -eq 0 ];then
                 device_array[${device_index}]=${entry}
                 device_index=$((device_index + 1))
             fi


schaefi:
@davidcassany this is not part of the current sprint but was severe enough to get some attention. Also the reported did such a great job in describing the issue that it was easy to come up with a fix.

davidcassany (approved):
Looks good to me :+1:

@dstiles314 good catch, thanks much.

PR #1590, opened 2020-10-15
Limit the use of linuxefi command

### LINKED ISSUE #1559: Arch Linux EFI boots fail
## Problem description

On Arch Linux images EFI boot is not working as expected. In short the grub.cfg patch that decides weather to use `linuxefi` or `linux` command is applied despite the `linuxefi` module was not found and in consequece not included in the `grub2-mkimage` command.

For completion I believe there are two tighty related but different issues here:

1 Check if there are prebuilt efi images for Arch, if so check if the path can be nicely included in the efi image lookup path in KIWI.

2 Make sure the patch that includes `linuxefi` usage is not applied when the command `linuxefi` is not available.

For the second this looks more like a generic issue rather than Arch specific.

Probably this is a side effect of 988404ff9f, the criteria on the grub version was probably not suited. I'd say that if we can validate that there is the actual `linuxefi` command somehow this should be the preferred criteria to apply the patch or not.

For for the notes but probably some decision criteri like in https://github.com/OSInside/kiwi/blob/c2f7d7fade8e1ecb97527c3bf7930a2e61432839/kiwi/bootloader/config/grub2.py#L817-L818 could prevent applying the patch when no linuxefi available, needs further verification.

## Expected behaviour

Arch Linux EFI images boot on EFI environments.

Make use of the VMX build of the integration tests for Arch Linux. At boot time it fails trying to run `linuxefi`, which is not there for Arch. Editing to boot entry at boot time to switch `$linux` to the actual command `linux` solves the issue.

davidcassany:
This commit makes sure that the linuxefi grub2 command is only being
used by the KIWI grub2 config patches if the linuxefi.mod file is
found. Otherwise the config patch to alternate linux and linuxefi
commands at runtime is not applied.

Fixes #1559

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -299,7 +299,10 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             )
         else:
             log.info('--> Using standard boot install template')
-            hybrid_boot = True
+            hybrid_boot = False
+            module_path = self._get_efi_modules_path(self.root_dir)
+            if os.path.exists(module_path + '/linuxefi.mod'):
+                hybrid_boot = True
             template = self.grub2.get_install_template(
                 self.failsafe_boot, hybrid_boot, self.terminal,
                 self.continue_on_timeout
@@ -358,7 +361,10 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             )
         else:
             log.info('--> Using standard boot template')
-            hybrid_boot = True
+            hybrid_boot = False
+            module_path = self._get_efi_modules_path(self.root_dir)
+            if os.path.exists(module_path + '/linuxefi.mod'):
+                hybrid_boot = True
             template = self.grub2.get_iso_template(
                 self.failsafe_boot, hybrid_boot,
                 self.terminal, self.mediacheck_boot
@@ -1113,21 +1119,23 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             # Any subsequent call of the grub config tool will overwrite
             # the setup and disables dynamic EFI environment checking
             # at boot time
-            with open(config_file) as grub_config_file:
-                grub_config = grub_config_file.read()
-                grub_config = re.sub(
-                    r'([ \t]+)linux(efi|16)*([ \t]+)', r'\1$linux\3',
-                    grub_config
-                )
-                grub_config = re.sub(
-                    r'([ \t]+)initrd(efi|16)*([ \t]+)', r'\1$initrd\3',
-                    grub_config
-                )
-            with open(config_file, 'w') as grub_config_file:
-                grub_config_file.write(
-                    Template(self.grub2.header_hybrid).substitute()
-                )
-                grub_config_file.write(grub_config)
+            module_path = self._get_efi_modules_path(self.root_dir)
+            if os.path.exists(module_path + '/linuxefi.mod'):
+                with open(config_file) as grub_config_file:
+                    grub_config = grub_config_file.read()
+                    grub_config = re.sub(
+                        r'([ \t]+)linux(efi|16)*([ \t]+)', r'\1$linux\3',
+                        grub_config
+                    )
+                    grub_config = re.sub(
+                        r'([ \t]+)initrd(efi|16)*([ \t]+)', r'\1$initrd\3',
+                        grub_config
+                    )
+                with open(config_file, 'w') as grub_config_file:
+                    grub_config_file.write(
+                        Template(self.grub2.header_hybrid).substitute()
+                    )
+                    grub_config_file.write(grub_config)
 
     def _fix_grub_root_device_reference(self, config_file, boot_options):
         if self.root_reference:
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -472,13 +472,18 @@ class TestBootLoaderConfigGrub2:
             True, 'gfxterm', None
         )
 
+    @patch('kiwi.bootloader.config.grub2.os.path.exists')
+    @patch('kiwi.bootloader.config.grub2.Defaults.get_grub_path')
     @patch.object(BootLoaderConfigGrub2, '_copy_grub_config_to_efi_path')
     def test_setup_live_image_config_standard(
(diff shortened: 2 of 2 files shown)


davidcassany:
I did some tests around this code [here](https://build.opensuse.org/project/show/home:dcassany:integration_tests). Again changes in the very senstive grub2 area. The tests I did make me feel comfortable enough to do final verification on Staging over all the main distros if this PR gets approved. This change should stay for a while in staging for some proper functional verification.

Conan-Kudo:
Wouldn't this fail in the event that you're installing the signed grub2 EFI binaries? The modules are not required to be installed when using those, since it's all compiled into the binaries...

davidcassany:
> Wouldn't this fail in the event that you're installing the signed grub2 EFI binaries? The modules are not required to be installed when using those, since it's all compiled into the binaries...

mmm that's a good point... if the linuxefi module comes from different package and the efi images are also provided from a different package too this is likely to fail if we are not beyond grub >= 2.04 ... all this stuff is so disappointing.

In fact I'd need to verify linuxefi command existance, is that possible at runtime? Gonna investigate it but I am not very happy scripting inside grub.cfg. I could not find any strategy to verify available/valid commands at build time.

schaefi:
Hmm, yes the implementation has potential to fail if the decision is based on the presence of the module as you don't know how the pre-compiled efi binary was built by the distributor. As we in kiwi prefer the pre-built versions of the EFI binaries this could still lead to a problem.

I was looking if there is an opportunity to check this in the grub.cfg file but I cannot came across a method to differentiate a command :/ Still investigating

schaefi:
Hmm, I don't think there is an easy solution. Maybe we can agree on a mapping table based on the efi vendor path:
e.g efi/EFI/BOOT/opensuse. See defaults/get_efi_vendor_directory. Based on that info we can decide for linuxefi y|n

I know this is also not ideal as the distros can change this from one version to another. In this case we might think of

```python
@staticmethod
    def get_os_release():
        with open('/etc/os-release', 'r') as handle:
            keys, values = zip(
                *[
                    (key.lower(), value.strip('\'"')) for (key, value) in (
                        line.strip().split('=', 1) for line in
                        handle.read().strip().split(os.linesep)
                    )
                ]
            )
            return namedtuple('OSRelease', keys)(*values)
```

and add a mapping table based on that information using good defaults and remap when needed.

SUSE is also changing such that "linux" handles both cases which then will lead to distros that has linux/linuxefi and linux/linux
So maybe the distro agnostic mapping of features is inevitable ?

I don't like it either

Thoughts ?

davidcassany:
> Hmm, I don't think there is an easy solution.

Agree, seams this linux/linuxefi pain is here to stay.

> SUSE is also changing such that "linux" handles both cases which then will lead to distros that has linux/linuxefi and linux/linux
> So maybe the distro agnostic mapping of features is inevitable ?

Yes, I was thinking about this yesterday. This issue I was trying to fix is not Arch Linux specific, meaning it just happened to appear there first, but it could easily replicate to other distros as they evolve and it looks like we are still quite far away of having a common cross-distro playground in that area. Looks like we will have to live for a while with the linux vs linuxefi for a quite a while.

Doing one step back. One of the main problems we had related to linux/linuefi is that grub2-mkconfig (at least for SUSE) relies on the values of `/sys/firmware/efi` to create the grub.cfg. Meaning that EFI builds on non efi hosts produces inconsistent results (aka OBS). I am wondering if we couldn't have the concept of setting some sort of chroot environment that mocks or configures the /sys in some known state. However I have no idea about how to achieve such a thing, I have doubts this is feasible.

> Thoughts ?

Not many unfortunately. So far having distro specific code is something that could work, I don't like either, so I'd keep it as one of the last options. Sounds like something that might be hard to maintain (as there are several derivative distros: Mageia, Mint, Oracle Linux, etc.). If we'd go that path I'd say the idea would be to make plain installations (no manual tweaks) and only apply certain changes on files based on the mentioned map. So the map would include a list of required patches per distro. If the distro is unknown, none is applied. Said this way sounds simple and easy however I am afraid we find some situation where it is not a simple as running the standard config and then apply few tweaks depending on the `/ets/os-release`. Take into accound we should also apply this distro concept for the templated configs.

schaefi:
> grub2-mkconfig (at least for SUSE) relies on the values of /sys/firmware/efi

This is so stupid. These tools that makes implicit assumptions from the host, we have that everywhere in grub. Look at how the menuentries are created, by looking at /proc/cmdline of the host. This is all crap and just causing pain. None of those guys ever built an image or ran their stuff on a host that is not the target :-((( so frustrating

schaefi:
we already have a bit distro specific code in defaults:get_efi_vendor_directory(). I hoped there will be no need to extend it

Thinking further we could also think of managing this on the image description level

Meaning our arch build can have a simple ```<type ... editbootconfig="script"/>``` that changes the grub.cfg as it's needed for this distribution.

The current way covers most of the distros, exception Arch. But I agree more exceptions are likely to follow

This is really a pain

davidcassany:
> Meaning our arch build can have a simple `<type ... editbootconfig="script"/>` that changes the grub.cfg as it's needed for this distribution.
> 
> The current way covers most of the distros, exception Arch. But I agree more exceptions are likely to follow

I am absolutely fine to fix Arch EFI use case at description level for now. I definitely prefer not changing the code rather than running another grub2 refactor that can easily lead to a not clearly better situation plus some additional breaks. The change I had in mind appeared to be simple at the beginning for the `_fix_*` method only, but then realized that templates were having the same issue, thus it already became bigger, then @Conan-Kudo realized that for precompiled efi images this was a bad choice and who knows what else we have overlooked. So I am in favor or not touching anything unless there is a clear solution candidate, something we do not have for the time being.

If this is fine for you I can close this PR and propose a new PR at description level.

schaefi:
I completely agree to your summary. In addition a fix on the description level can only break that description and as it is a distribution specific topic (every distro is somehow different here) I would say we should fix it on the description level until it
can be done in a generic way in code.

From my perspective we can only loose if we try to manage these type of differences in the builder code. The history of this report nicely shows the many pitfalls and maybe we even haven't thought about all of them.

> If this is fine for you I can close this PR and propose a new PR at description level.

Fine for me and thanks much for your effort. I'm sorry that you coded for no merge in this case but it has helped a lot to identify the complexity of the situation.

Thanks

PR #1591, opened 2020-10-16
Add support for s390 CDL DASD disks

schaefi:
On s390 and in CDL mode (4k DASD) the call of grub2-install
does not work because grub2-install is not able to identify
a 4k fdasd partitioned device as a grub supported device
and fails. As grub2-install is only used to invoke
grub2-zipl-setup and has no other job to do we can
circumvent this problem by directly calling grub2-zipl-setup
instead. Also delete LDL mode support, IBM no longer supports
this


schaefi:
almost there, but needs the root device fix as grub2-zipl-setup has the same root=/dev/mapper/.. bug like grub2-install. Why are all the grub tools soo broken

davidcassany (approved):
Looks good to me.

I am not into the details of the DASD grub2 configuration thus I trust the syntax on the config files.

schaefi:
Thanks for reviewing this part. I had a conversation with Mike yesterday and we still have the problem that the DASD device doesn't get activated in the initrd. There is a dracut module named: /usr/lib/dracut/modules.d/95zdev/parse-zdev.sh. I took a look into it and it does the following

```
zdev_fw_file="/sys/firmware/sclp_sd/config/data"
zdev_base_args="--force --yes --no-root-update --no-settle --auto-conf --quiet"

...

chzdev --import "$zdev_fw_file" $zdev_base_args
```

This chzdev call makes /dev/dasdX available and also triggers the creation of the typical udev device names e.g /dev/disk/by-uuid/...

The call did not happen because /sys/firmware/sclp_sd/config/data is empty and that part I don't understand. How to feed data in there and if that is a job of kiwi or some other component.

If we issue the chzdev command manually in a rescue shell in the initrd everything works. But obviously that doesn't help.

So this PR might still be incomplete.

schaefi:
@davidcassany I probably will need some time in the s390 are before I will come back to our sprint items. Just if you wonder about slower response times

schaefi:
we are getting closer, but still the problem of the modified zipl/config file that is only correct at image build time but not on the running machine

schaefi:
All right, all parts of the change are done now and have been tested together with Mike, Ihno and Mark, great people :)

schaefi:
Image build tests ready for a final test here:

https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_s390:sle15/test-image-disk

mfriesenegger:
I tested kiwi-test-image-disk.s390x-1.15.1-Physical-Build14.1.raw.xz and it booted successfully.

Issue #1555, opened 2020-09-16
Functional testing of integration tests


schaefi:
in the past and in several different situation we would have been in a better position if a functional test of an image, meaning
an image that is actually really booted, would have existed before we release.

Recently we saw it on grub2 and secure boot and the changes the people did. Result: GCE images broken for secure boot.
Consequence: A bug in a hurry, testing of the fix, starting a new release, keeping fingers crossed nothing else broke due to
the hurry.

For me this is a situation that becomes more and more inacceptable and we should put on a system that does automated functional tests of our integration test images

dcermak:
openQA could definitely provide some of the necessary features for the kiwi test builds:
- it can launch arbitrary VMs and boot from virtual machine disk images
- tests can be partitioned into individual steps that can be combined into test suites (e.g. have a different boot up workflow for isos and virtual machine disks but run the same tests once the VM is up and running)
- it can provide a comprehensive overview of one full test run of all test suites belonging to a single build, e.g. https://openqa.opensuse.org/tests/overview?distri=microos&distri=opensuse&version=Tumbleweed&build=20200915&groupid=1
- openQA can be run locally as well, so all tests or just a subset of them can be re-executed locally if desired
- there are already some kiwi tests on in the SLE tests, we could leverage these for even less work

on the other hand, openQA lacks in the following regard:
- there are no public instances like with travis or github actions: we'd either have to host our own or use openqa.opensuse.org (provided that the o3 maintainers agree). In any case, it would mean that the test cases will **not** be part of this repository.
- Triggering builds automatically from OBS is afaik not that straight forward. However, as this would be probably intended for a weekly verification run (?) a manual trigger would be definitely doable via a script.
- openQA only really supports qemu as a backend. Other types of virtualization can be used by relying on nested virtualization, but that is a huge pain in the backside as nested virtualization adds so many problems.

dcermak:
Another advantage of openQA is the ability to run tests on a plethora of architectures: o3 for executes tests on x86, aarch64 and ppc64. In theory there are also s390x runners, but none of the tests executed in the last months made it, so I assume that the runner has been dead for a while now.

schaefi:
There is also a system called kanku: https://kanku.suse.de

The downside is there seems to be no obs/osc integration and the test recipes cannot be taken from a source service

schaefi:
[...]

PR #1595, opened 2020-10-20
Rename image build tests

schaefi:
To use the image builds in openQA they have to have a unique
name such that it cannot happen that a cached version of an
image in openQA is used. The current names matched openQA
cached images e.g openSUSE-Tumbleweed and in addition different
image build tests used the same name. This commit uses the
name of the image as it is organized in its directory structure
prepending "kiwi-" to be unique in openQA when it fetches
the image. This is realted to Issue #1555


davidcassany (approved):
I agree aligning the image name with the OBS package name is a good practice.

The container builds in OBS using KIWI follow this approach.

PR #1596, opened 2020-10-20
Allow custom root volume name setup

### LINKED ISSUE #1530: Allow custom root LVM volume name
## Problem description
The root LVM volume name is currently hardcoded to "LVRoot".

## Expected behaviour
Instead of using "@root" in \<systemdisk>\<volume name="...">, allow putting in a real volume name in the name attribute, together with mountpoint="/" attribute to indicate root volume.

+ KIWI version: 9.21.3
+ Operating system: SLES15
+ OBS version: not using OBS

schaefi:
In addition to the custom size of the root volume it's now
also possible to setup the name of the root volume as follows:

```xml
<volume name="@root=rootlv"/>
```

If no name for the root volume is specified the default
name: LVRoot applies as before. This Fixes #1530

### DIFF
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -23,6 +23,8 @@ from collections import namedtuple
 from textwrap import dedent
 
 # project
+import kiwi.defaults as defaults
+
 from kiwi import xml_parse
 from kiwi.system.uri import Uri
 from kiwi.defaults import Defaults
@@ -1223,7 +1225,8 @@ class XMLState:
                         mountpoint=path,
                         fullsize=True,
                         label=volume_label,
-                        attributes=['no-copy-on-write']
+                        attributes=['no-copy-on-write'],
+                        is_root_volume=True|False
                     )
                 ]
 
@@ -1243,7 +1246,8 @@ class XMLState:
                 'mountpoint',
                 'fullsize',
                 'label',
-                'attributes'
+                'attributes',
+                'is_root_volume'
             ]
         )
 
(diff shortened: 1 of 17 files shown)


schaefi:
This needs extensive testing. I'm still running the tests

schaefi:
All tests good so far

kiwi/xml_state.py
```diff
@@ -1243,7 +1245,8 @@ def get_volumes(self):  # noqa C901
                 'mountpoint',
                 'fullsize',
                 'label',
-                'attributes'
+                'attributes',
+                'is_root_volume'
```
davidcassany, line 1249: nit: the is_root_volume tuple attribute is not included in the above docs comment.

davidcassany (approved):
Nice! I like it :+1: I just added a comment regarding the docs of the volume tuple that has not been updated accordingly.

kiwi/xml_state.py
schaefi: yep, will fix it :)

schaefi: done

ksanjeet:
Can we similarly rename the LVSwap in the xml file?

schaefi:
> Can we similarly rename the LVSwap in the xml file?

nope, the setup of swap is still using a fixed name. If you need this to be configurable please open an issue for it. An explanation why the name of the swap volume is relevant for your use case would be appreciated. Thanks

ksanjeet:
> nope, the setup of swap is still using a fixed name. If you need this to be configurable please open an issue for it. An explanation why the name of the swap volume is relevant for your use case would be appreciated. Thanks

I have opened a new issue https://github.com/OSInside/kiwi/issues/1638

My customer(s) want to standardize the swap volume names. One of them have their existing swaps are on /dev/system/swap. They want the new servers created by the images to have same name. It would be good to have swap volume name configurable.

Currently I am using a OneShot systemd unit file to do a lvrename. I would want to avoid that if possible.
Thanks.

PR #1597, opened 2020-10-21
Obsolete config functions baseMount/baseCleanMount

### LINKED ISSUE #1536: baseCleanMount is not compatible with set -e
It's good practice to use `set -e` in `config.sh` and related scripts to ensure that there is no silent breakage.
However, that breaks the call to `baseCleanMount` at the end:

```
[  111s] + baseCleanMount
[  111s] + for path in /proc/sys/fs/binfmt_misc /proc /dev/pts /sys
[  111s] + '[' -d /proc/sys/fs/binfmt_misc ']'
[  111s] + umount /proc/sys/fs/binfmt_misc
[  111s] 
[  111s] [ INFO    ]: 11:26:51 | Cleaning up SystemPrepare instance
```

Full log: https://build.opensuse.org/package/live_build_log/home:favogt:kiwisete/kiwi-templates-JeOS:kvm-and-xen/openSUSE_Tumbleweed/x86_64

schaefi:
The above methods are obsolete since kiwi handles these
mount/umount processes as part of the core builder code.
This Fixes #1536

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -296,27 +296,14 @@ function baseGetProfilesUsed {
 # baseCleanMount
 #--------------------------------------
 function baseCleanMount {
-    for path in /proc/sys/fs/binfmt_misc /proc /dev/pts /sys;do
-        [ -d "${path}" ] && umount "${path}" &>/dev/null
-    done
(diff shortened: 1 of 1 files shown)


davidcassany (approved):
Yes, I think this is a great solution. current kiwi-ng already takes care of all this in a nicer way.

PR #1598, opened 2020-10-26
Improve debootstap execution

### LINKED ISSUE #1587: debian: bootstrap packages handled like other stages, falsely assuming that the chroot is ready yet
## Problem description

The general steps for building a Debian rootfs are:
- debootstrap a basic chroot
- configure basic system (repositories, dns, ...)
- invoke apt to install custom package selection

In kiwi, `debootstrap` is run for a packages block of type *bootstrap*. Next, all configured repositories are configured and apt is used to install all packages listed in the bootstrap block.
What's wrong here is the assumption that all repos and apt are already usable at this point.

There are corner cases where repositories of e.g. type https require additional packages in the chroot before they are usable;
or maybe additional signing keys need to be installed ...

## Expected behaviour

usability of all repos / apt is not assumed for the bootstrap stage, by either:
- use debootstrap cli option for additional packages
- only activate one marked repository for the bootstrap packages step

Sample *config.xml* that uses a repo with https, and tries to bootstrap with the required packages for that:
```
<image name="sr-imx6">
...
	<repository type="apt-deb" distribution="buster" components="main non-free" repository_gpgcheck="false">
		<source path="http://deb.debian.org/debian"/>
	</repository>
	<repository type="apt-deb" repository_gpgcheck="false">
		<source path="https://repo.solid-build.xyz/debian/buster/bsp-any"/>
	</repository>
	<packages type="bootstrap">
...
	</packages>
	<packages type="image">
...
	</packages>
</image>
```

Sample log:
```
[ INFO    ]: 14:48:05 | Using package manager backend: apt-get
[ INFO    ]: 14:48:05 | Installing bootstrap packages
[ INFO    ]: 14:48:05 | --> collection type: onlyRequired
[ INFO    ]: 14:48:05 | --> package: apt-get
[ INFO    ]: 14:48:05 | --> package: apt-transport-https
[ INFO    ]: 14:48:05 | --> package: ca-certificates
[ ERROR   ]: 14:54:49 | KiwiCommandError: chroot: stderr: W: https://repo.solid-build.xyz/debian/buster/bsp-any/./InRelease: No system certificates available. Try installing ca-certificates.
```

+ KIWI version: KIWI (next generation) version 9.21.17
+ OBS version: -

davidcassany:
This PR refactors the bootstrap procedure of APT based images. Since APT is not capable to bootstrap the `debootstrap` tool is used. However `debootstrap` overlaps with KIWI on certain functionalities (for instance the mounts of kernel file systems for a proper chroot environment). In order to make debootstrap call as transparent as possible compared with non APT package managers small refactor is suggested.

Fixes #1587

### DIFF
--- a/kiwi/system/root_bind.py
+++ b/kiwi/system/root_bind.py
@@ -98,6 +98,19 @@ class RootBind:
                 '%s: %s' % (type(e).__name__, format(e))
             )
 
+    def umount_kernel_file_systems(self):
+        """
+        Umount kernel filesystems
+
+        :raises KiwiMountKernelFileSystemsError: if some kernel filesystem
+            fails to mount
+        """
+        umounts = [
+            mnt for mnt in self.mount_stack if mnt.device
+            in self.bind_locations
+        ]
+        self._cleanup_mounts(umounts)
+
     def mount_shared_directory(self, host_dir=None):
         """
         Bind mount shared location
@@ -266,19 +279,22 @@ class RootBind:
 
                 shutil.move(config_rpm_new, self.root_dir + config)
 
-    def _cleanup_mount_stack(self):
-        for mount in reversed(self.mount_stack):
-            if mount.is_mounted():
+    def _cleanup_mounts(self, umounts):
+        for umount in reversed(umounts):
+            if umount.is_mounted():
                 try:
-                    mount.umount_lazy()
+                    umount.umount_lazy()
+                    self.mount_stack.remove(umount)
                 except Exception as e:
                     log.warning(
                         'Image root directory %s not cleanly umounted: %s',
                         self.root_dir, format(e)
                     )
             else:
-                log.warning('Path %s not a mountpoint', mount.mountpoint)
+                log.warning('Path %s not a mountpoint', umount.mountpoint)
 
+    def _cleanup_mount_stack(self):
+        self._cleanup_mounts(self.mount_stack)
         del self.mount_stack[:]
 
     def _cleanup_dir_stack(self):
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -143,16 +143,22 @@ class PackageManagerApt(PackageManagerBase):
                 cmd.append('--no-check-gpg')
             if self.deboostrap_minbase:
                 cmd.append('--variant=minbase')
+            if self.package_requests:
+                cmd.append(
+                    '--include={}'.format(','.join(self.package_requests))
+                )
             if self.repository.components:
                 cmd.append(
                     '--components={0}'.format(
                         ','.join(self.repository.components)
                     )
                 )
+            self.cleanup_requests()
             cmd.extend([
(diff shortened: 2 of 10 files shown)


davidcassany:
@schaefi this PR comes with two flavors, one per commit. Let me explain.

There are two commits, both are solving the former issue #1587. So probably a look by commits might me interesting for the review and discussion. The first commit is just adding the bootstrap packages into the `debootstarp` call, in order to do so and keep the Command.call return structure I do a little hack, around Command.call -> wait it finalizes -> data sync -> return the Cammand.call object even having the process finalized. Simple and it works, but feels artificial having to add this little hack around Command.call.

This is required becuase KIWI expects that bootstrap procedure is being called in a non blocking subprocess and then the stdout is being processed on real time byt KIWI to provide the progress bar. With the former desing of debootstrap this is not possible since debootstrap is not called into the target folder, so after a debootstrap call a DataSync step is required before exiting the bootstrap installation method.

In oder to solve that, a second commit is applied to refactor that part and make debootstrap call happen directly over the target folder removing the need of the later DataSync step (so a faster bootstrap is also a concequence). Note this is not strictly required to solve #1587 and this implies a little API refactor. This refactor was spinning in my head since quite a while and never actually tried to code it since there were no issues in that area. This refactor should also allow to use `--allow-existing-root` flag on APT based images (even I have not really tested it). Probably there are other ways to perform this kind of refactor, I tried to be as less invasive as possible at the time I tried to keep the API meaningful.

davidcassany:
@Josua-SR I'd appreciate if you could give a try of this PR. I did some tests on my local environment but I could not verify if this actually did the trick to solve your certificates problem and gpg verification for repositories.

kiwi/system/root_bind.py
```diff
@@ -98,6 +98,28 @@ def mount_kernel_file_systems(self):
...
+            if umount.is_mounted():
+                try:
+                    umount.umount_lazy()
+                    self.mount_stack.remove(umount)
+                except Exception as e:
+                    log.warning(
+                        'Image root directory %s not cleanly umounted: %s',
+                        self.root_dir, format(e)
+                    )
+
```
schaefi, line 122: This duplicates some code that it already present in the private _cleanup_mount_stack() method. Maybe you could consider to extend _cleanup_mount_stack such that you can pass mountpoints to umount. This would then result in a smaller umount_kernel_file_systems that might look like this:

```python
def umount_kernel_file_systems(self):
    self._cleanup_mount_stack(self.bind_locations)
```

schaefi (approved):
This is a great refactoring and improve on use of debootstrap. I like it. There is only part I noticed where I think we could do better regarding code duplication

kiwi/system/root_bind.py
davidcassany: Good point, gonna fix it.

schaefi: This is great now :+1:

PR #1600, opened 2020-11-02
Set GRUB_DISTRIBUTOR if requested via displayname

### LINKED ISSUE #1575: Regression in Grub menu options - user text - oem
## Problem description
Grub user text, on reboot of oem type installs, has not customisation of user facing grub options text. Previously both the initial boot of the installer (still working) and all subsequent reboots and updates, would maintain the user facing text customisation applied by kiwi.

I.E. Previously generated oem installers from our relatively unchanged config would result in an initial grub screen, in our case, depending on profile related config edits (name=) of:

> Rocsktor-Leap15.2-RaspberryPi4 [ OEM ]
> Advanced options for Rockstor-Leap15.2-RaspberryPi4 [ OEM ]
> Start bootloader from a read-only snapshot

or with no edits the first line would read:

> Rocsktor-NAS [ OEM ]

But more recent kiwi-ng generated oem type installers have the above expected customisation as expected during install but upon booting into the resulting install the Grub text is reverted to the upstream 'openSUSE' text thus:

 > openSUSE Leap 15.2
> Advanced options for openSUSE Leap 15.2
> Start bootloader from a read-only snapshot

So the installer's evident grub text customisations are lost, i.e. don't persist to the resulting install, and we have inadvertent use of on openSUSE "Mark" and loss of the OEM bit.

Sorry but I haven't bisected when this regression happened but I think it may have been around 1-2 months or so ago. This issue is not profile/architecture dependant in our case as it has been replicated in our x86_64 profile, again a 'Built on openSUSE" Leap 15.2 variant.

## Expected behaviour
The Grub text Kiwi-ng customisation is I am assuming intended to persist in the resulting install. This we rely upon in order that our resulting installer and installs conform to the Trademark guidelines as per:

https://en.opensuse.org/openSUSE:Trademark_guidelines

As the user facing "openSUSE" in the resulting install's grub option looks like we are claiming to be "openSUSE" which is obviously not the case. Hence the prior welcome 'capability' within kiwi-ng to have this auto customised.

Build an oem type image using recent kiwi-ng, the config used in this case was from our as yet unreleased in binary form:

https://github.com/rockstor/rockstor-installer

Leap 15.2 host:
```
sudo zypper addrepo http://download.opensuse.org/repositories/Virtualization:/Appliances:/Builder/openSUSE_Leap_15.2/ appliance-builder
sudo zypper --gpg-auto-import-keys refresh
sudo zypper install git python3-kiwi btrfsprogs gfxboot
git clone https://github.com/rockstor/rockstor-installer.git
cd rockstor-installer/
sudo kiwi-ng --profile=Leap15.2.x86_64 --type oem system build --description ./ --target-dir /home/kiwi-images/
```
The resulting installer is found to exhibit this issue's behaviour, although given the referenced config is relatively vanilla it is assumed that all oem configs share this more recent regression.

+ KIWI version: More recent versions exhibit this behaviour, prior ones did not. First noticed on 9.21.7-lp152.1.1 Virtualization_Appliances_Builder (arm64) but could well have appeared before that. Also seen on 9.21.7-lp152.5.7.1 x86_64 Main Update Repository. And just now on 9.21.15-lp152.1.1 Virtualization_Appliances_Builder x86_64.
+ Operating system: Leap 15.2 host and target
+ OBS version: OBS is not used in this case as the sighted config is stand alone and hosted on a local openSUSE Leap 15.2 instance.

Apologies if this is down to a slip up in our own kiwi-ng configuration as per the indicated repo, but we have made no relevant changes to this config (bar default name changes etc) that would impact on this to my knowledge. But I have noticed a lot of work going on in kiwi-ng re grub in this suspected regression period.

@schaefi Sorry to bring yet another grub issue up here as I know it's not one of your most favoured elements but we find ourselves in a spot re inadvertent user facing use of an openSUSE Trademark and with our own multi year 'Build on openSUSE' endeavour nearing it's binary installer download release stage I am somewhat concerned with this regression. And from what I can tell this would affect all oem type installs.

Bar some prior issue reports/pre-release-fix-confirmations, I have no working knowledge of the kiwi-ng code but I'm happy to have a go if you could point me in the right direction. Although I'm assuming/hoping here, given it's prior function, this issue is a simple regression rather than a change in how this 'capability' is intended to work..

Thanks to all involved for the kiwi-ng project. It is almost singularly responsible for our ability to have our very own custom installer at all. A key element in 'The Rockstor Project''s effort to achieve sustainability in our open source endeavour.

schaefi:
If the image description explicitly specifies a displayname
it is expected that the bootloader shows this in the menu.
Therefore in case displayname is set GRUB_DISTRIBUTOR will
be set if grub2 is in use. This partially reverts #1420
and Fixes #1575

### DIFF
--- a/build-tests/x86/suse/test-image-disk-simple/appliance.kiwi
+++ b/build-tests/x86/suse/test-image-disk-simple/appliance.kiwi
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<image schemaversion="7.3" name="kiwi-test-image-disk-simple">
+<image schemaversion="7.3" name="kiwi-test-image-disk-simple" displayname="Simple Disk">
     <description type="system">
         <author>Marcus Schäfer</author>
         <contact>ms@suse.com</contact>
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -94,6 +94,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         self.timeout = self.get_boot_timeout_seconds()
         self.timeout_style = \
             self.xml_state.get_build_type_bootloader_timeout_style()
+        self.displayname = self.xml_state.xml_data.get_displayname()
         self.serial_line_setup = \
             self.xml_state.get_build_type_bootloader_serial_line_setup()
         self.continue_on_timeout = self.get_continue_on_timeout()
@@ -594,12 +595,17 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         * GRUB_CMDLINE_LINUX_DEFAULT
         * GRUB_GFXMODE
         * GRUB_TERMINAL
+        * GRUB_DISTRIBUTOR
         """
         grub_default_entries = {
             'GRUB_TIMEOUT': self.timeout,
             'GRUB_GFXMODE': self.gfxmode,
             'GRUB_TERMINAL': '"{0}"'.format(self.terminal)
         }
+        if self.displayname:
+            grub_default_entries['GRUB_DISTRIBUTOR'] = '"{0}"'.format(
+                self.displayname
+            )
         if self.timeout_style:
             grub_default_entries['GRUB_TIMEOUT_STYLE'] = self.timeout_style
         if self.cmdline:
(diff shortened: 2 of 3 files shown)


davidcassany (commented):
Nice! :+1: I have not tested or verified it, but it looks really clean and simple. Grub config should be all that simple

schaefi:
Yeah I think at least one test should set displayname such that we can verify it. done now

davidcassany (approved):
Looks good

phillxnet:
@schaefi Thanks for adding this option. Much appreciated.

PR #1601, opened 2020-11-02
Refactor solver and storage factory

schaefi:
Refactor DiskFormat and SolverRepository

Refactors DiskFormat and SolverRepository to turn them into proper factory classed and to also include type hints to facilitate it's use from an API POV. Related to #1498


schaefi:
@davidcassany my share on the factory refactoring for this month :) Let me know if you want me to take on the other two as well.

PR #1602, opened 2020-11-02
Do not exclude filesystem folders in OCI images

davidcassany:
This commit does not exclude the /dev folder during the rsync call
in OCI images. It has been noted that including an empty /dev folder
does not hurt and it can eventually help to work around some limitations
of container related tools such as buildah.

Fixes bsc#1176129

### DIFF
--- a/kiwi/container/oci.py
+++ b/kiwi/container/oci.py
@@ -114,10 +114,9 @@ class ContainerImageOCI:
         :param string base_image: archive used as a base image
         """
         exclude_list = Defaults.get_exclude_list_for_root_data_sync()
-        exclude_list.append('boot')
-        exclude_list.append('dev')
-        exclude_list.append('sys')
-        exclude_list.append('proc')
+        exclude_list.append('dev/*')
+        exclude_list.append('sys/*')
+        exclude_list.append('proc/*')
 
         oci = OCI()
         if base_image:
--- a/test/unit/container/oci_test.py
+++ b/test/unit/container/oci_test.py
@@ -134,7 +134,7 @@ class TestContainerImageOCI:
         mock_oci.sync_rootfs.assert_called_once_with(
             'root_dir', [
                 'image', '.profile', '.kconfig', '.buildenv',
-                'var/cache/kiwi', 'boot', 'dev', 'sys', 'proc'
+                'var/cache/kiwi', 'dev/*', 'sys/*', 'proc/*'
             ]
         )
         mock_oci.repack.assert_called_once_with({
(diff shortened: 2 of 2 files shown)


davidcassany:
@scheafi this helps to workaround an upstream buildah issue. Even I am not in favor of altering the code to facilitate workarounds for very specific tools and use cases it has been noted by the container's core engineers (@cyphar in this case) that including an empty `/dev` folder does not hurt. In fact, by looking at the repository history I could not see any specific reason to not include `/dev` folder, I just saw that it has been this way since the very beginning.

So given there are no clear technical motivations to omit this folder and that the change is straight forward I am fine by including this patch.

you can see the details in [bsc#1176129](https://bugzilla.suse.com/show_bug.cgi?id=1176129)

kiwi/container/oci.py
```diff
@@ -115,7 +115,7 @@ def create(self, filename, base_image):
         """
         exclude_list = Defaults.get_exclude_list_for_root_data_sync()
         exclude_list.append('boot')
-        exclude_list.append('dev')
+        exclude_list.append('dev/*')
         exclude_list.append('sys')
```
Vogtinator, line 119: All of those folders are owned by `filesystem`, so excluding them makes the RPM verification fail.

Could you drop the exclude of `/boot` completely and add `/*` to the others? Are there any hidden files anywhere which `*` wouldn't catch?

davidcassany: I did not see hidden files in `/dev`, gonna verify about the others. Yes, I agree we should leave this folders untouched.

Vogtinator:
LGTM, just the title is now somewhat misleading. Is there a test container with this built somewhere?

davidcassany:
> LGTM, just the title is now somewhat misleading. Is there a test container with this built somewhere?

Yes, just rewrote the title. I have only done a couple of local builds to test the changes, no OBS builds. Basically I created a docker image, converted to OCI with skopeo and unpacked the rootfs with umoci to validate the emptyness of the relevant folders.

cyphar:
This seems reasonable and even in the worst case where some hidden device inode snuck into `/dev`, containers are run under the devices cgroup by default which would block most dangerous device inode access anyway.

schaefi (approved):
The set of changes looks good to me. I haven't done a test build but it seems we are in agreement that this is the way to go. Thanks everybody :+1:

davidcassany:
> Is there a test container with this built somewhere?

@Vogtinator Here is a test for SLE 12 SP5 https://build.suse.de/package/show/QA:Maintenance:Images:12-SP5:kiwi-ng-testing/test-image-docker

Vogtinator:
> > Is there a test container with this built somewhere?
> 
> @Vogtinator Here is a test for SLE 12 SP5 https://build.suse.de/package/show/QA:Maintenance:Images:12-SP5:kiwi-ng-testing/test-image-docker

For some reason, both `/proc` and `/sys` have mode 755 instead of 555 as expected by the `filesystem` rpm. Can you check why?

PR #1604, opened 2020-11-02
Refactor root image and volume manager factory

schaefi:
Refactor VolumeManager / RootImport
    
This commit refactors VolumeManager and RootImport to turn them into proper
factory class and to also include type hints to facilitate it's
use from an API POV. Related to #1498


schaefi:
@davidcassany going to run a build test right now, but should just work

schaefi:
tests ok

davidcassany (approved):
Looks good

PR #1606, opened 2020-11-03
Fixed install of bash completion

### LINKED ISSUE #1603: bash-completion does not recognize files with suffix
`/usr/share/bash-completion/bash_completion` does not recognize completion addons if they end end with a suffix, like `kiwi-ng.sh`.

````bash
    for dir in "${dirs[@]}"; do
        [[ -d $dir ]] || continue
        for compfile in "$cmd" "$cmd.bash" "_$cmd"; do
````
I think `kiwi-ng.sh` must be installed as `kiwi`.

schaefi:
The kiwi completion was installed as kiwi-ng.sh below
/usr/share/bash-completion/completions. This is wrong
because the completion does not pick up files with
a suffix like .sh. This commit changes the completion
file to be installed as kiwi-ng without the suffix
and Fixes #1603

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # completion
-completion/kiwi-ng.sh
+completion/kiwi-ng
 
 # Schema formats
 kiwi/schema/kiwi.xsd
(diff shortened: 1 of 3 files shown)


davidcassany (approved):
:+1:

PR #1608, opened 2020-11-04
Fix quick start guide build command

quatran:
The kiwi-descriptions were reorganized in profiles (See OSInside/kiwi-descriptions@788b919ea2500b9d495622c8140e618938634306).
However the build command in the quick start guide was not updated appropriately and therefore the build fails.

This commit will update the build command.


schaefi:
Yes you are right, thanks much :+1: I fear there might be more of this

PR #1612, opened 2020-11-04
Move usrmerge package out of the bootstrap section

davidcassany:
Currently bootstrap phase on APT package manager makes use of the
debootstrap tool. However debootstrap is limited to execute  the
bootstrap using a single repository. This is causes several limitations
in OBS builds, such as the impossibility of using update repositories or
the inclusion of any package that is not part of the standard OBS
repository.

Usrmerge package is part of the universe repository in OBS which is not
te one used by debootstrap, so it can't be installed on bootstrap phase.


schaefi:
perfect thanks :+1:

PR #1613, opened 2020-11-09
Update tox and travis setup for python 3_8

schaefi:
Move latest python test target to 3.8 and also change
the deploy travis target to use python 3.8

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -36,12 +36,12 @@ unit_py37:
   image: $CI_REGISTRY/$BUILD_IMAGES_PROJECT:$FEDORA_BUILD
   stage: test
   script:
-    - export PYTHON=python3.7
-    - tox -e unit_py3_7 "-n $(nproc)"
+    - export PYTHON=python3.8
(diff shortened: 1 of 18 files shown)


schaefi:
I realized the pypi deployment on travis requires python 3.8 to operate. That's why this change was made. I also think we should move the latest python test target to 3.8

davidcassany:
> I realized the pypi deployment on travis requires python 3.8 to operate. That's why this change was made. I also think we should move the latest python test target to 3.8

Yes I agree and python 3.8 is also the current python3 version for TW.

davidcassany (approved):
I have no strong arguments to increase or decrease the complexity level of the flake check. Having it low can be helpful when coding big chunks of code to enforce us to try to split it with as many as possible meaningful pieces. However I also believe human reviews are the best way to keep code maintainable and readable, so I don't have the impression that we will lower the code quality by relaxing complexity limit of flake rules.

schaefi:
>I also believe human reviews are the best way to keep code maintainable and readable, so I don't have the impression that we will lower the code quality by relaxing complexity limit of flake rules.

Right that's what I also think

PR #1614, opened 2020-11-09
Add editbootinstall script for Arch Linux tests

### LINKED ISSUE #1559: Arch Linux EFI boots fail

davidcassany:
This commit adds the editbootinstall script to Arch Linux OEM
integration tests. The provided script removes the use of linuxefi and
initrdefi commands on grub configuration since Arch does not support
linuxefi module.

Fixes #1559

### DIFF
--- a/build-tests/x86/archlinux/test-image-live-disk-kis/appliance.kiwi
+++ b/build-tests/x86/archlinux/test-image-live-disk-kis/appliance.kiwi
@@ -31,12 +31,12 @@
         <rpm-check-signatures>false</rpm-check-signatures>
     </preferences>
     <preferences profiles="Live">
-        <type image="iso" flags="overlay" hybridpersistent_filesystem="ext4" hybridpersistent="true" firmware="efi" kernelcmdline="console=ttyS0">
+        <type image="iso" flags="overlay" hybridpersistent_filesystem="ext4" hybridpersistent="true" kernelcmdline="console=ttyS0">
             <bootloader name="grub2" console="serial"/>
         </type>
     </preferences>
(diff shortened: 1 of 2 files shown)


schaefi (approved):
Looks great :+1:

PR #1615, opened 2020-11-09
Update packaging spec

### LINKED ISSUE #1503: Packaging
## Problem description

Among other dubious deps, `zypper install python3-kiwi` asks also these dependencies:

```
bea-stax
bea-stax-api
git-gui
gitk
java-11-openj9-headless
javapackages-filesystem
javapackages-tools
timezone-java
tk
```

My instance:
```
cat /etc/os-release 
NAME="openSUSE Leap"
VERSION="15.2 Beta"
```

So seems like not "just Python", but also "git-gui" and "tk" with "java"... :-) Overall, about 200Mb installation. I'd say, it is a bit of too much, no?

schaefi:
Get rid of spec file Recommends because this is not implemented
consistently across all distro rpm versions. In addition the
recommended software components only makes sense if used in
an image build context. To create context a collection of
kiwi-host-setup-for-CONTEXT sub packages are added which
pulls in the context required software components. This
Fixes #1503

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -352,7 +352,7 @@ Group:          %{sysgroup}
 %description -n kiwi-man-pages
 Provides manual pages to describe the kiwi commands
 
-%package -n kiwi-host-setup-for-building-containers
+%package -n kiwi-systemdeps-containers
 Summary:        KIWI - host requirements for container images
 Group:          System/Management
 Obsoletes:      kiwi-image-docker-requires
@@ -368,11 +368,11 @@ Requires:       buildah
 Requires:       skopeo
 
 %if 0%{?fedora} || 0%{?rhel} || 0%{?suse_version}
-%description -n kiwi-host-setup-for-building-containers
+%description -n kiwi-systemdeps-containers
 Host setup helper to pull in all packages required/useful on
 the build host to build container images e.g docker, wsl
 
-%package -n kiwi-host-setup-for-building-iso-media
+%package -n kiwi-systemdeps-iso-media
 Summary:        KIWI - host requirements for live and install iso images
 Group:          System/Management
 Obsoletes:      kiwi-image-iso-requires
@@ -387,11 +387,11 @@ Requires:       gfxboot
 %endif
 %endif
 
-%description -n kiwi-host-setup-for-building-iso-media
+%description -n kiwi-systemdeps-iso-media
 Host setup helper to pull in all packages required/useful on
 the build host to build live and install iso images.
 
-%package -n kiwi-host-setup-for-building-filesystem-images
+%package -n kiwi-systemdeps-filesystem-images
 Summary:        KIWI - host requirements for filesystems
 Group:          System/Management
 Obsoletes:      kiwi-image-pxe-requires
@@ -416,24 +416,24 @@ Requires:       squashfs-tools
 Requires:       squashfs
 %endif
 
-%description -n kiwi-host-setup-for-building-filesystem-images
+%description -n kiwi-systemdeps-filesystem-images
 Host setup helper to pull in all packages required/useful on
 the build host to build filesystem images
 
-%package -n kiwi-host-setup-for-building-disk-images
+%package -n kiwi-systemdeps-disk-images
 Summary:        KIWI - host requirements for disk images
 Obsoletes:      kiwi-image-oem-requires
 Obsoletes:      kiwi-image-vmx-requires
 Group:          System/Management
 Provides:       kiwi-image:oem
 Provides:       kiwi-image:vmx
-Requires:       kiwi-host-setup-for-building-filesystem-images
+Requires:       kiwi-systemdeps-filesystem-images
 
-%description -n kiwi-host-setup-for-building-disk-images
+%description -n kiwi-systemdeps-disk-images
 Host setup helper to pull in all packages required/useful on
 the build host to build disk images
 
-%package -n kiwi-host-setup-for-image-validation
+%package -n kiwi-systemdeps-image-validation
 Summary:        KIWI - host requirements for handling image descriptions better
 Group:          System/Management
 %if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version} || 0%{?debian} || 0%{?ubuntu}
@@ -444,7 +444,7 @@ Requires:       python%{python3_pkgversion}-solv
 %endif
 Requires:       python%{python3_pkgversion}-anymarkup
 
-%description -n kiwi-host-setup-for-image-validation
+%description -n kiwi-systemdeps-image-validation
 Host setup helper to pull in all packages required/useful on
 the build host to handling image descriptions better. This also
 includes reading of image descriptions for different markup
@@ -566,19 +566,19 @@ fi
 %endif
 
 %if 0%{?fedora} || 0%{?rhel} || 0%{?suse_version}
-%files -n kiwi-host-setup-for-building-containers
+%files -n kiwi-systemdeps-containers
 %defattr(-, root, root)
 
-%files -n kiwi-host-setup-for-building-iso-media
+%files -n kiwi-systemdeps-iso-media
 %defattr(-, root, root)
 
-%files -n kiwi-host-setup-for-building-filesystem-images
+%files -n kiwi-systemdeps-filesystem-images
 %defattr(-, root, root)
 
-%files -n kiwi-host-setup-for-building-disk-images
+%files -n kiwi-systemdeps-disk-images
 %defattr(-, root, root)
 
-%files -n kiwi-host-setup-for-image-validation
+%files -n kiwi-systemdeps-image-validation
 %defattr(-, root, root)
 %endif
 
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -44,7 +44,7 @@
 
 Name:           python-kiwi
 Version:        %%VERSION
-Provides:       kiwi-schema = 7.2
+Provides:       kiwi-schema = 7.3
 Release:        0
 Url:            https://github.com/OSInside/kiwi
 Summary:        KIWI - Appliance Builder Next Generation
@@ -85,9 +85,6 @@ Group:          Development/Languages/Python
 Obsoletes:      python2-kiwi
 Conflicts:      python2-kiwi
 Conflicts:      kiwi-man-pages < %{version}
-%if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version} || 0%{?debian} || 0%{?ubuntu}
-Recommends:     jing
-%endif
 %if 0%{?ubuntu} || 0%{?debian}
 Requires:       python%{python3_pkgversion}-yaml
 %else
@@ -107,9 +104,6 @@ Requires:       python%{python3_pkgversion}-pyxattr
 %ifarch x86_64
 Requires:       grub2-x86_64-efi
 %endif
-%ifarch %{ix86} x86_64
-Recommends:     gfxboot
-%endif
 Requires:       qemu-tools
 Requires:       squashfs
 Requires:       gptfdisk
@@ -124,15 +118,8 @@ Requires(postun): chkconfig
 Requires:       qemu-img
 Requires:       squashfs-tools
 Requires:       gdisk
-%if 0%{?fedora} || 0%{?rhel} >= 8
-Recommends:     gnupg2
 %endif
-%endif
-%if 0%{?suse_version}
-# If it's available, let's pull it in
-Recommends:     dnf
-Recommends:     gpg2
-%endif
+Provides:       kiwi-image:tbz
 %if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version} >= 1550
 Provides:       kiwi-packagemanager:microdnf
 Requires:       microdnf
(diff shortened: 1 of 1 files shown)


schaefi:
Hi, this is my first draft on adding the provides tags in sub-packages of the kiwi spec such that it is useful for any user not only obs. I currently struggle with two issues:

a) Should any of the kiwi-host-setup-for-X packages require kiwi ?
b) Is the name kiwi-host-setup-for-validation good or bad  ? see the description for it

Other than that I'd like to ask you @Conan-Kudo for a closer look on the required packages with regards to the other distros and if we need to add exceptions and conditions.

Overall this is one result from the conversation we had regarding the consolidation of packaging.

With these sub-packages providing the provides tags from the kiwi-boot-descriptions we will also have an opportunity to drop them from kiwi-boot-descriptions at least for Factory.

All this needs to be reviewed carefully. Thanks much

package/python-kiwi-spec-template
```diff
@@ -364,6 +344,102 @@ Group:          %{sysgroup}
...
+Provides:       kiwi-filesystem:xfs
+Requires:       e2fsprogs
+Requires:       xfsprogs
+%if 0%{?fedora} || 0%{?rhel}
+Requires:       btrfs-progs
+Requires:       squashfs-tools
+%else
+Requires:       btrfsprogs
+Requires:       squashfs
+%endif
```
Conan-Kudo, line 411: This is still wrong for Debian/Ubuntu and RHEL 8.

package/python-kiwi-spec-template
```diff
@@ -364,6 +344,102 @@ Group:          %{sysgroup}
...
+Host setup helper to pull in all packages required/useful on
+the build host to build disk images
+
+%package -n kiwi-host-setup-for-validation
+Summary:        KIWI - host requirements for handling image descriptions better
+Group:          System/Management
+Requires:       jing
+%if 0%{?suse_version}
+Requires:       Requires:       python%{python3_pkgversion}-solv
+%endif
```
Conan-Kudo, line 436: Why do we require the Python bindings for libsolv? Also, `Requires: Requires:` is not valid.

package/python-kiwi-spec-template
```diff
@@ -364,6 +344,102 @@ Group:          %{sysgroup}
...
+Requires:       Requires:       python%{python3_pkgversion}-anymarkup
```
Conan-Kudo, line 437: `Requires: Requires:` is not valid.

package/python-kiwi-spec-template
```diff
@@ -364,6 +344,102 @@ Group:          %{sysgroup}
...
+%package -n kiwi-host-setup-for-live-media
+Summary:        KIWI - host requirements for live iso images
+Group:          System/Management
+Provides:       kiwi-image:iso
+Requires:       checkmedia
+Requires:       dosfstools
+Requires:       xorriso
+%ifarch %{ix86} x86_64
+Requires:       syslinux
+Requires:       gfxboot
```
Conan-Kudo, line 377: `gfxboot` doesn't exist outside of SUSE.

package/python-kiwi-spec-template
```diff
@@ -364,6 +344,102 @@ Group:          %{sysgroup}
...
+%description -n kiwi-host-setup-for-wsl
+Host setup helper to pull in all packages required/useful on
+the build host to build wsl/appx images
+
+%package -n kiwi-host-setup-for-docker
+Summary:        KIWI - host requirements for docker images
+Group:          System/Management
+Provides:       kiwi-image:docker
+Requires:       skopeo
+Requires:       umoci
```
Conan-Kudo, line 362: `umoci` doesn't exist outside of SUSE.

Conan-Kudo (changes requested):
I'm not terribly enthusiastic about how this is structured now, but here's some first pass feedback. Additionally, there's missing `%files` sections to correspond with these new subpackages, which means they'll never get built. Finally, Debian/Ubuntu aren't accounted for very well now. Note that we can't have those Provides on Debian/Ubuntu systems...

schaefi:
Thanks for the feedback. I'm looking into this soon. But before:

> I'm not terribly enthusiastic about how this is structured now

It's just a first suggestion from my side. I would love to know how you would structure it ? Please remember this should be useful to users. The way this was done just targets a service (obs) now it targets users and it would be great if you can share your thoughts how this could be useful. My thinking was this:

A user wants to build cloud images. The user knows this requires virtual disks. He installs:

```
$ package_manager install python-kiwi kiwi-host-setup-for-disk
```

And after that the user can build disk images.

Thoughts ?

schaefi:
My goal is also not to change these packages anymore once they have settled. That way the names can be used in the documentation in a clear and straight forward way

package/python-kiwi-spec-template
schaefi: libsolv is needed if you call:

kiwi-ng image info --description foo --resolve-package-list

very handy for debugging and in case of errors. That's why I put it into the kiwi-host-setup-for-validation sub package

davidcassany:
> A user wants to build cloud images. The user knows this requires virtual disks. He installs:
> 
> ```
> $ package_manager install python-kiwi kiwi-host-setup-for-disk
> ```
> 
> And after that the user can build disk images.

I am in favor of this structure. In chat @schaefi identified three blocs to divide packages:

```
    a python module
    a collection of dracut extensions
    a collection of host requirements
```

That makes absolute sense to me. I'd say that the `collection of host requirements` is not more that a list of requirements (probably recommends also) and obsoletes, no files associated at all.

I'd say we should include in the new subpackages that track host requirements an `Obsoletes` to any other package from `kiwi-descriptions/custom_boot/package` where there is a `Provides` overlap. IMHO both packages should not be installed at the same time.

package/python-kiwi-spec-template
davidcassany: There is the `kiwi-ng image info --resolve-package-list` command that resolves the packages that are required to build the image without actually build it. This is helpful for debugging the required list of packages without actually having to launch a build for some smoke test verification.

package/python-kiwi-spec-template
```diff
@@ -123,16 +117,9 @@ Requires:       qemu-img
 Requires:       squashfs-tools
 Requires:       gdisk
 Requires:       dnf
+Provides:       kiwi-image:tbz
```
davidcassany, line 120: I have doubts about this provides list, probably I'd put clearly together the requirement and the provides, so it is easy to spot which distro package build provides, requires and recommends each package manager.

davidcassany (commented):
Good for start, I miss Obsoletes to the old packages used in OBS.

package/python-kiwi-spec-template
schaefi: I put in this provides to the main package because it is actually true. The ability to create simple tarballs of the root tree is provided by the main package. The main package also has a requirement to tar set.

Currently this provides is used in "kiwi-image-tbz-requires" as it comes from kiwi-boot-descriptions. From a user perspective it does not make sense because the support for the "tbz" image type is provided by just installing python3-kiwi as this package already requires tar and nothing more is needed on the host to create tarballs from a root tree. This is the reason why I put this provides flag to the main package.

Not good ?

davidcassany: The provides for `kiwi-image:tbz` is good. Sorry my comment was not clear enough and now I realized that it was under the wrong line. I wanted to comment to the  `kiwi-packagemanger:*` provides. Which have not really been modified by your PR. Due to the lack of context in the diff yesterday I got confused by those provides as I did not spot that they were already defined only defined for Fedora and RHEL.

I see the `kiwi-packagemanger:apt` is missing but IIRC there was an issue with the spec trying to add it next to the requirement of the debootstrap. Either dpkg or the spec to deb converter did not like having colons in the provides.

schaefi:
@davidcassany I've added the obsoletes setup

schaefi:
I've done the changes regarding the review and a bit more. The current model provides the following sub-packages:

* kiwi-host-setup-for-building-containers
  All you need to build containers

* kiwi-host-setup-for-building-iso-media
  All you need to build basically iso images

* kiwi-host-setup-for-building-filesystem-images
  All you need to build kiwi supported filesystem images

* kiwi-host-setup-for-building-disk-images
  All you need to build virtual disk images and their variants, lvm, crypto etc...

* kiwi-host-setup-for-image-validation
  All you need to validate and check base on image descriptions, jing, libsolv, etc...

Let me know what you think ?

Thanks

davidcassany (approved):
Looks good to me, Iike the current approach.

Something to keep in mind is that we should eventually review our integration tests configuration in OBS to properly verify it. So we should keep in mind that adding this on Virt:App:Staging is some work that only rebuilding the package including this PR. As far as I see it, we should be capable to simplify some of the prj configs with this and also a good chance to do some cleaning there.

Conan-Kudo:
I'm going to take some time this weekend to fully review this and offer an alternative or at least enhancements to this.

schaefi:
> we should be capable to simplify some of the prj configs with this and also a good chance to do some cleaning there.

@davidcassany I agree it would be a good consequence of the change.

@Conan-Kudo I'm eager to see what changes you have in mind. From your response so far it feels like you are not happy with all of that ;) No worries about that but could you share your concerns and explain the issues you see with it ? Much appreciated. Thanks

Conan-Kudo:
The problem with this is that it doesn't solve any of the problems described in the commit. We still need to deal with `Recommends` and different package names, and it makes it more complicated for folks who are trying to understand what they need to support _everything_.

Another thing I'm slightly concerned about is _what_ functionality we want to consider baseline. In my opinion, we should be considering image description validation as a baseline function, but this reorganizes it to make it more like explicitly optional functionality.

So I guess my problem with this is that we lack consensus on what we _want_ by default.

There are also a couple of technical problems that we don't have solutions for:
* We still need conditional usage of weak dependencies because of platform differences (at least until RHEL 7 is dropped in KIWI 10)
* We still need to deal with weaknesses in the deb format and condition some stuff out (like Provides with "invalid" characters)
* We still need to deal with different package names across distributions we support (this might go away with KIWI 10, who knows?)

schaefi:
> We still need to deal with Recommends

Why ? I wanted to get rid of any recommends as the concept is not available on all distributions and from my experience also pointless. At the time you find out you are missing something it's no longer recommended it becomes mandatory for the customer use case. So our consideration of recommended actually means "depending on the use case". That's why this commit deletes all Recommends and make them a requirement for the desired use-case specific sub package

> and it makes it more complicated for folks who are trying to understand what they need to support everything

In pseudo code this means:

```
$ package-manager install kiwi-host-setup-for-*
```

If the package manager supports globing that would be the way to get *everything*.  I would write this also as words
in the documentation if we can come to an agreement.

> Another thing I'm slightly concerned about is what functionality we want to consider baseline.

That's a good point. My consideration here is based on the minimal functionality. The absolute minimum is the capability to create a new root tree and pack it as an archive. This is what the PR here implements. I was thinking in the direction of embedded system users and that it's easier to install "minimal plus X" instead of install "something-we-think-is-minimal".
I think we will only loose at this point because there will always be people who need something different than what we
think. The creation of a new root tree is a must have for image creation, there is no way out and that should be the baseline
in my thinking. Everything else is use case specific.

> In my opinion, we should be considering image description validation as a baseline function, but this reorganizes it to make it more like explicitly optional functionality.

This is partially correct. The validation functionality automatically comes with libXML2 which is a requirement. The problem here is that errors reported from libXML2 are hard to understand. Therefore kiwi adds some comfort with jing which reports much better what and where in the document there are issues. Also the extra kiwi command to pre-solve packages list is a comfort function to do early validation of consistency. So nothing of this I consider a mandatory part. Also with jing and the solver extra dependencies and even huge ones (java) are pulled in. So from my perspective we all want this as optional requirements. Maybe the name chosen for this (kiwi-host-setup-for-image-validation) is misleading and I'm open for alternatives

> So I guess my problem with this is that we lack consensus on what we want by default.

I think we are not so far apart

> We still need conditional usage of weak dependencies because of platform differences (at least until RHEL 7 is dropped in KIWI 10)

yes and I don't see that we would change this with this PR

> We still need to deal with weaknesses in the deb format and condition some stuff out (like Provides with "invalid" characters)

right and I also did not touch this part to my knowledge

> We still need to deal with different package names across distributions we support (this might go away with KIWI 10, who knows?)

yes unfortunately and inevitable by now. I think I did not break the various %if conditions in the spec to handle it

Thoughts ?

Conan-Kudo:
@schaefi There was some breakages for some distributions. I'll try to provide some feedback on this as I complete running some test package builds locally to verify my findings.

schaefi:
per conversation on the matrix channel:

1. Name change similar to the suggestion from @Conan-Kudo 

    ``` 
    %package -n kiwi-systemdeps-containers
    %package -n kiwi-systemdeps-iso-media
    %package -n kiwi-systemdeps-filesystem-images
    %package -n kiwi-systemdeps-disk-images
    %package -n kiwi-systemdeps-image-validation
    ```

2. Only apply this for rpm based distributions. We would like to refactor the packaging for kiwi 10 if there is also an opportunity to break backward compatibility

schaefi:
@davidcassany  @Conan-Kudo did anybody of you remember for what reason was gpg included in the spec ? Thanks

davidcassany:
> @davidcassany @Conan-Kudo did anybody of you remember for what reason was gpg included in the spec ? Thanks

I yes I do. I believe I included this dependency because this is needed to create the key rings for Deb repositories. See this code:

https://github.com/OSInside/kiwi/blob/fa283ef4431d81b44dc6f012cdbd23b306e89774/kiwi/repository/apt.py#L210-L231

schaefi:
I see, thanks. So from my perspective this should be a requirement then for the deb build

schaefi:
Closing on behalf of a better solution provided by @Conan-Kudo

PR #1617, opened 2020-11-11
NBD/AoE overlay support

schaefi:
This one came in as report from the kiwi descriptions and a user of the legacy netboot system. The resulting simple extension of the kiwi-overlay module now allows to use remote root filesystem images as described:

**Added support for nbd and aoe root overlay**
    
The kiwi-overlay dracut module can also be used as standalone module that is not connected to a disk image. In this case  it's needed to specify the location for the root filesystem and optionally the device to write data (default is ram space).  This commit adds the opportunity to specify a nbd/aoe location  for the root filesystem on the kernel cmdline like in the  following examples:
    
* root=overlay:nbd=192.168.100.42:exportname
* root=overlay:aoe=e0.1
    
An optional write space, if it should not be ram space, can be  provided through the rd.root.overlay.write option on the kernel cmdline.

This Fixes: OSInside/kiwi-descriptions#78

### DIFF
--- a/doc/source/building_images/build_kis.rst
+++ b/doc/source/building_images/build_kis.rst
@@ -64,7 +64,7 @@ build the image:
 
 .. code:: bash
 
-   $ sudo kiwi-ng --type kis system build \
+   $ sudo kiwi-ng --profile KIS system build \
        --description kiwi-descriptions/suse/x86_64/{exc_description} \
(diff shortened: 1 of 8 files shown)


davidcassany (approved):
Looks good :+1: quite some learning by reading it :)

PR #1621, opened 2020-11-12
Create relative boot link for extra boot partition

### LINKED ISSUE #1611: Centos8 with xfs not producing bootable image.
## Problem description
CentOS 8 image built with xfs filesystem fails to boot.  Same settings for an ext4 system boots successfully.

The gets to grub, but grub fails to find vmlinuz.  If you drop to the grub command line and manually point to the vmlinuz and initrd file, you can get the system to fully boot.  Once I booted the system, I ran grub2-mkconfig, but was still unsuccessful in getting the image to boot.  There are two profiles in the xml, Virtual-ext4 and Virtual-xfs the only different being the filesystem type for the image.  The Virtual-ext4 boots successfully without issue, only the Virtual-xfs image is affected.

## Expected behaviour
Image boots successfully.

Build using the attached config.xml.

Image was tested on a kvm host built on ubuntu 18.04.

+ KIWI version: python3-kiwi-9.20.5-lp152.4.1.x86_64 
+ OBS version: 
[config.zip](https://github.com/OSInside/kiwi/files/5490417/config.zip)

schaefi:
If an extra boot partition is used the grub toolchain
still references files from that partition as /boot/...
which fails because they are now at the toplevel. To
avoid this and keep any /boot/some-file reference still
valid we create a symlink 'boot -> .' This Fixes #1611

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -242,6 +242,20 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 config_file.replace(self.root_mount.mountpoint, '')
             ]
         )
+        if boot_options.get('root_device') != boot_options.get('boot_device'):
+            # Create a boot -> . link on the boot partition.
+            # The link is useful if the grub mkconfig command
+            # references all boot files from /boot/... even if
+            # an extra boot partition should cause it to read the
+            # data from the toplevel
+            bash_command = [
+                'cd', os.sep.join([self.root_mount.mountpoint, 'boot']), '&&',
+                'rm', '-f', 'boot', '&&',
+                'ln', '-s', '.', 'boot'
+            ]
+            Command.run(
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
Interesting issue I am surprised it did not raise until now. I am not very enthusiastic of having to create such link, but yes, the fix is simple and it should not hurt. 

Is this CentOS specific? I gave a quick shoot in Leap having set bootpartition="true" and grub.cfg was properly configured without the need of such a symlink.

schaefi:
Thanks for the review :+1: 

>Is this CentOS specific? I gave a quick shoot in Leap having set bootpartition="true" and grub.cfg was properly configured without the need of such a symlink.

I think it's grub tooling specific. I also did not see this issue on Leap which is probably the reason why it was not spotted so far.
I also plan to come up with a followup PR that disables the need for a boot part when XFS is selected. This is from times when grub was not able to read from xfs

PR #1626, opened 2020-11-16
Added microdnf support in XML schema

schaefi:
The XML schema did not allow to specify microdnf as
supported package manager


schaefi:
This microdnf support is lacking some fundamentals when it comes to real usage

Conan-Kudo (approved):

Conan-Kudo:
@schaefi Gack, we will need to bump the schema version now that the schema has changed.

PR #1627, opened 2020-11-16
Better error reporting if jing is missing

schaefi:
On validation error we use jing to report detailed error
messages. However if jing is not present no validation
errors are displayed. There is a error_log variable as
part of the relaxNG object which holds the library error
log. This information is not as good as the jing report
but better than nothing

### DIFF
--- a/kiwi/exceptions.py
+++ b/kiwi/exceptions.py
@@ -161,7 +161,7 @@ class KiwiCommandError(KiwiError):
     """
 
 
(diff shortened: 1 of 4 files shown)


schaefi:
I found this when playing with the microdnf support

davidcassany (approved):
Nice, I think I coded that part :P yes makes much more sense to include the former error message in addition to the jing warning. Even if the lxml error is this hard to read or does not include much meaningful data is better than nothing.

PR #1625, opened 2020-11-16
Added microdnf integration test

schaefi:
Add an integration test for the recently added microdnf support

PR #1629, opened 2020-11-18
Fixed microdnf support

schaefi:
The installroot argument must be used together with --config
and additionally with --noplugins, as well as --setopt for
cachedir, reposdir and varsdir. Related to #1625

### DIFF
--- a/kiwi/package_manager/microdnf.py
+++ b/kiwi/package_manager/microdnf.py
@@ -102,7 +102,16 @@ class PackageManagerMicroDnf(PackageManagerBase):
         bash_command = [
             'microdnf'
         ] + ['--refresh'] + self.dnf_args + [
-            '--installroot', self.root_dir
+            '--installroot', self.root_dir, '--noplugins',
+            '--setopt=cachedir={0}'.format(
+                self.repository.shared_dnf_dir['cache-dir']
+            ),
+            '--setopt=reposdir={0}'.format(
+                self.repository.shared_dnf_dir['reposd-dir']
+            ),
+            '--setopt=varsdir={0}'.format(
+                self.repository.shared_dnf_dir['vars-dir']
+            )
         ] + self.custom_args + ['install'] + self.package_requests
         self.cleanup_requests()
         return Command.call(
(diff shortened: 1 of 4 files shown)


kiwi/repository/dnf.py
```diff
@@ -80,7 +80,8 @@ def post_init(self, custom_args=None):
         self.shared_dnf_dir = {
             'reposd-dir': manager_base + '/repos',
             'cache-dir': manager_base + '/cache',
-            'pluginconf-dir': manager_base + '/pluginconf'
+            'pluginconf-dir': manager_base + '/pluginconf',
+            'vars-dir': manager_base + '/vars'
```
Conan-Kudo, line 84: We should add `varsdir` setting to the generated `dnf.conf` file, as that can be used by DNF too, and I could see reasons for us to want to populate that in the future.

Conan-Kudo (changes requested):

kiwi/package_manager/microdnf.py
```diff
@@ -102,7 +102,16 @@ def process_install_requests_bootstrap(self, root_bind=None):
         bash_command = [
             'microdnf'
         ] + ['--refresh'] + self.dnf_args + [
-            '--installroot', self.root_dir
+            '--installroot', self.root_dir, '--noplugins',
+            '--setopt=cachedir={0}'.format(
+                self.repository.shared_dnf_dir['cache-dir']
```
Conan-Kudo, line 107: Does `self.repository` exist at this point?

Conan-Kudo (changes requested):

kiwi/repository/dnf.py
schaefi: ok will adapt regarding this

kiwi/package_manager/microdnf.py
schaefi: sure, the package manager object can only be created by passing in a repository instance

Conan-Kudo: Ah, okay.

schaefi:
```python
class PackageManagerMicroDnf(PackageManagerBase):
```

and the base class has

```python
def __init__(self, repository, custom_args=None):
```

kiwi/repository/dnf.py
schaefi: done

Conan-Kudo (approved):

PR #1632, opened 2020-11-23
Refactor markup and package manager factory

schaefi:
Refactor PackageManager and Markup
    
This commit refactors the PackageManager and Markup classes and turns them into proper factory classes which also include type hints to facilitate their use from an API POV.

Related to #1498


davidcassany (approved):
Looks good!

I am sorry I did not have the change to participate much on this refactoring

PR #1633, opened 2020-11-24
Avoid using generators in pre-mount hooks

davidcassany:
This commit deletes the generator that was creating the sysroot.mount
unit for ramdisk deployments. Generators, specially the sysroot.mount is
expected to be created on very early stages of the boot procedure as
this has impact on relevant targets such as initrd-root-fs.target, which
does not depend on sysroot.mount if the unit is not there.

In ramdisk deployments some data is known on pre-mount stage as
as it is downloaded from the PXE server. At this stage it is not safe to
generate a sysroot.mount unit that depends on initrd-root-fs.target as
the target is close to finalize or even finalized already and could
potentially skip sysroot.mount exection.

Instead we create a sysroot.mount unit which is tied to initrd-fs.target

This fixes bsc#1178670


davidcassany:
@fbuihuu Could you also have a look to this PR so this is a sane fix for [bsc#1178670](https://bugzilla.suse.com/show_bug.cgi?id=1177560)

davidcassany:
@schaefi I did some tests for it using the tumbleweed examples from kiwi-descriptions and including 

```xml
        <type image="oem" primary="true" filesystem="ext4" initrd_system="dracut" bootloader="grub2" firmware="efi" installpxe="true" kernelcmdline="rd.kiwi.ramdisk ramdisk_size=2097152 rd.shell rd.debug">
            <size unit="G">2</size>                                             
            <oemconfig>                                                         
                <oem-skip-verify>true</oem-skip-verify>                         
                <oem-unattended-id>/dev/ram1</oem-unattended-id>                
                <oem-swap>false</oem-swap>                                      
            </oemconfig>                                                        
        </type>
```

And followed the KIWI docs to set a PXE server on a Leap 15.2 VM I have.

I did another test using the same server to validated the original image facing the issue reported in [bsc#1178670](https://bugzilla.suse.com/show_bug.cgi?id=1177560). Image based on SLE15-SP3 Beta

dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
```diff
@@ -33,17 +33,36 @@ function boot_installed_system {
...
+        while read -r -d ' ' opt; do echo "${opt}";done < /config.bootoptions |\
+        grep root= | cut -f2- -d=
+    )
+
+    [ -z "${root_uuid}" ] && \
+        die "'root=' argument not found in config.bootoptions"
+
+
+    image_target=$(get_selected_disk)
+    kpartx -s -a "${image_target}"
```
schaefi, line 53: This part of the code looks mixed up and I got confused ;)

1. The reading of config.bootoptions is explicit to PXE deployments and done in code only if if getargbool 0 rd.kiwi.install.pxe; is true. This code imho doesn't belong here

2. The kpartx mapping of the devices happened as an exceptional case if getargbool 0 rd.kiwi.ramdisk; is true. Only in this case we need to directly map the partitions of the ramdisk block device after deployment. udev does not do this on ramdisk block devices. I also don't see this code to belong here

dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
```diff
@@ -33,17 +33,36 @@ function boot_installed_system {
...
 
 if getargbool 0 rd.kiwi.ramdisk; then
     # For ramdisk deployment a kexec boot is not possible as it
     # will wipe the contents of the ramdisk. Therefore we prepare
-    # the switch_root from this deployment initrd. Also see the
-    # unit generator: dracut-kiwi-ramdisk-generator
-    image_target=$(get_selected_disk)
-    kpartx -s -a "${image_target}"
+    # the switch_root from this deployment initrd.
+    mount_sysroot
```
schaefi, line 65: So mount_sysroot is called in the ramdisk case only. The code from here has been moved into mount_sysroot and hides the scope. It feels like mount_sysroot can be used generically, but this is not the case. The mount_sysroot method itself contains code parts from here and code parts from PXE. Sorry I'm very confused about all that

schaefi (changes requested):
Thanks for the effort. Using systemd-mount seems a good solution to the generator race condition problem. However I'm concerned and also confused about the code changes here from a design perspective and am also relatively sure it breaks other non pxe based deployments

schaefi:
So this commit tries to refactor the use of generators for the 90kiwi-dump module. The concept of generators for sysroot is also used here:

* 90kiwi-overlay/kiwi-generator.sh
* 90kiwi-live/kiwi-generator.sh

I think for completeness the same change should apply there as well. Thoughts ?

davidcassany:
> So this commit tries to refactor the use of generators for the 90kiwi-dump module. The concept of generators for sysroot is also used here:
> 
>     * 90kiwi-overlay/kiwi-generator.sh
> 
>     * 90kiwi-live/kiwi-generator.sh
> 
> 
> I think for completeness the same change should apply there as well. Thoughts ?

No, I think you are missing the root cause of the reported issue. Let me try to explain it (it took me a while to actually realize it, kudos to @fbuihuu for a great and detailed report on bugzilla).

The problem with the kiwi dump module when using it for ramdisk deployments is that the generator for sysroot.mount is called twice and with different results. All generators are executed in parallel on systemd start and they set the basis for the very basic transactions for fundamental units such as `initrd-root-fs.target`. At this staget he ramdisk deployment generator is called and does not produce any result since the `config.bootoptions` file is not yet downloaded. Since there is no `sysroot.mount` unit this is not added as dependency to `initrd-root-fs.target`, meaning the target can be reached without mounting `/sysroot`.

Witht he current approach `sysroot.mount` unit get created once the ramdisk generator is called again in pre-mount hook after downloading `config.bootoptions` and calling `kpartx` (I assume this triggers again the generator execution). Since `config.bootoptions` is there this second time the generatror succeeds and produces the `sysroot.mount`, but then there are few problems, on one side `initrd-root-fs.target` could have been already finalized when the new unit is actually evaluated; on the other side there is no explicit  `systemd daemon-reload` to ensure systemd is actually aware of the changes.

`90kiwi-overlay` generator and  `90kiwi-live` generator do no face this problem as they are executed as soon as systemd is launched and they already create consistent units and there are no further changes during the boot. The problem of the ramdisk deployment is that the generator is expected to execute the right unit during pre-mount hook and this is something to avoid as it dynamically changes ongoing transactions with hard to foresee results.

My patch tries to simply mount `/sysroot` when required by using systemd tools (so systemd is aware of it). As far as I could understand `systemd-mount` utility generates a mount unit independent of `initrd-root-fs.target` that gets evaluated before `switch-root.service`.

dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
davidcassany: Right I agree the method name is missleading. This method is ramdisk deployment specific, the method name should reflect this fact. What about `mount_ramdisk`?

davidcassany: Also not that my patch on affects the ramdisk deployment execution path

schaefi:
>At this staget he ramdisk deployment generator is called and does not produce any result since the config.bootoptions file is not yet downloaded.

Oh I see, yes sorry I missed that part. Ok I got mostly confused about the generic mount_sysroot method.

dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
davidcassany: > This part of the code looks mixed up and I got confused ;)
> 
>     1. The reading of config.bootoptions is explicit to PXE deployments and done in code only if if getargbool 0 rd.kiwi.install.pxe; is true. This code imho doesn't belong here

Probably this can be done differently, note this code is borrowed from the ramdisk generator and it is only used to get the `root=` argument from the `config.bootoptions` to generate the `sysroot.mount` unit required for ramdisk deployments. Probably we could load the `root_uuid` variable in previous stages and just test `root_uuid` exists here. I have no strong preference.

>     2. The kpartx mapping of the devices happened as an exceptional case if getargbool 0 rd.kiwi.ramdisk; is true. Only in this case we need to directly map the partitions of the ramdisk block device after deployment. udev does not do this on ramdisk block devices. I also don't see this code to belong here

I think the method name confused you, this method is ramdisk specific and it mounts the ramdisk, it is not being used in any other scenario. As you say we need kpartx here to get the ramdisk device mapped as a regular disk so we can use the syntax from `config.bootoptions` and refer to the root device by it's UUID.

dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
schaefi: yes please rename it, mount_ramdisk_sysroot or similar would be good imho

schaefi:
I now also fully understand the problem. That we are reading config.bootoptions in the generator creates a dependency to this file but as you said generators are called under systemd control and I did not care for correct dependency handling in this case.

schaefi:
ok all good then, sorry for the confusion

schaefi (approved):
Given we rename the method all looks good to me, thanks much :+1:

davidcassany:
> Given we rename the method all looks good to me, thanks much +1

Method renamed to `mount_ramdisk`

dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
```diff
@@ -33,17 +33,36 @@ function boot_installed_system {
...
+        while read -r -d ' ' opt; do echo "${opt}";done < /config.bootoptions |\
+        grep root= | cut -f2- -d=
+    )
+
+    [ -z "${root_uuid}" ] && \
+        die "'root=' argument not found in config.bootoptions"
+
+
+    image_target=$(get_selected_disk)
+    kpartx -s -a "${image_target}"
```
schaefi, line 53: One thing I'd like to mention is that the method could reduce duplication a bit and avoid double empty lines, e.g

```bash
function mount_ramdisk_sysroot {
￼    local image_target
￼    local root_uuid
    local boot_options=/config.bootoptions
￼
￼    if [ ! -e "${boot_options}" ]; then
        die "Missing ${boot_options} config file"
    fi
￼    root_uuid=$(
￼        while read -r -d ' ' opt; do echo "${opt}";done < "${boot_options}" |\
￼        grep root= | cut -f2- -d=
￼    )
￼    if [ -z "${root_uuid}" ];then
￼        die "'root=' argument not found in ${boot_options}"
    fi
￼    image_target=$(get_selected_disk)
￼    kpartx -s -a "${image_target}"
```

I'm suggesting this because in case of an error the debug information is a nicer read. Thoughts ?

schaefi:
@davidcassany looks all great now :) I think we are waiting for feedback prior merge right ? if so feel free to merge when all is set. Thanks

dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
```diff
@@ -33,17 +33,40 @@ function boot_installed_system {
...
+        grep root= | cut -f2- -d=
+    )
+
+    if [ -z "${root_uuid}" ]; then
+        die "'root=' argument not found in ${boot_options}"
+    fi
+
+    image_target=$(get_selected_disk)
+    kpartx -s -a "${image_target}"
+    systemd-mount --fsck=no --no-block "${root_uuid}" /sysroot
```
fbuihuu, line 58: Well `systemd-mount` seems still not appropriate to me:

- the transient mount unit created here has no dependencies on `initrd-root-fs.target`, so you can't really tell for sure when PID1 will schedule it.
- since `--no-block` is used you actually don't even wait for the transient unit to be added to the list of current jobs before returning from `mount_ramdisk_sysroot()`.

I don't really understand why you want to use `systemd-mount` here especially since this tool is supposed to be used when file-systems needs to be mounted asynchronously. 

Why don't you simply use mount(8) here ?

PS: I would also add a comment explaining how /sysroot is mounted as it's not really obvious ;)

schaefi: > Why don't you simply use mount(8) here ?

I guess this was done to let systemd control the mount process for root. From my perspective the asynchronous behavior is ok at that point. The problem with the former generator for the sysroot unit was that the generator was depending on a file to be downloaded from the network in case of the remote (PXE) use case. If the generator was called before the code here has downloaded the file from the network no sysroot unit was created. With the changes David did we make sure the systemd-mount process is called not before any needed data is present. After that it does not play a role when PID1 schedule it.

I think it is part of the design that systemd should know about /sysroot. We can of course just mount it but wouldn't it be better if systemd controls this ?

I agree regarding the --no-block. I think that should not be used since I assume the call to return as soon as systemd has created the unit file or am I mistaken ?

Thanks much for your feedback

schaefi: It was a stupid idea from my side to read a file within system-generators/dracut-kiwi-ramdisk-generator code when I could not be sure this file exists no matter at what time the generator code is called.

davidcassany: @fbuihuu thanks for sharing you view, this is really valuable.

> I don't really understand why you want to use systemd-mount here especially since this tool is supposed to be used when file-systems needs to be mounted asynchronously.

The motivation of using `systemd-mount` as @scheafi said was to make systemd aware of `/sysroot` mount, as it thought this is something desirable. Probably for a ramdisk deployment this not that much of an issue.

> the transient mount unit created here has no dependencies on initrd-root-fs.target, so you can't really tell for sure when PID1 will schedule it.

That's correct but there is the `dracut-mount.service` right after `dracut-pre-mount.service` that actually waits for /sysroot being mounted, otherwise it does not continue as far as I could understand by reading the code.

> since --no-block is used you actually don't even wait for the transient unit to be added to the list of current jobs before returning from mount_ramdisk_sysroot()

I added it to sort out a deadlock, apparently the transient mount unit gets scheduled after the `pre-mount-dracut.service`, so this was causing a dead lock.

So I see four options to proceed:

1. Leave it as is in this PR
2. Run simple mount command in `mount_ramdisk_sysroot()` method
3. Include a mount hook (they are executed by `dracut-mount.service`) that only operates on `rd.kiwi.ramdisk` argument presence
4. In `mount_ramdisk_sysroot()` we create our own custom `sysroot.mount` unit manually (with the dependencies that suit our needs to not mess any running transaction) and run `systemctl daemon-reload`.

Probably the second and the third are the simplest ones and for a ramdisk deployment this might be sufficient.

schaefi: I would go with option 2 for this use case. I don't see how a dracut mount hook would be called when there is no /sysroot mountpoint. Isn't that a chicken&egg problem.

Regarding mount points: These are usually all handled by systemd through unit files. There is an fstab generator, there is a sysroot generator and so on. systemd does this to have control over all mount points of the system, which is a good concept. Running a manual mount circumvents this concept but is not a problem when the system was deployed into a ramdisk. That's because no data is persistent and on reboot everything is gone and you need to start from scratch. So it's ok to manually mount /sysroot in this ramdisk case.

But to be honest it's not nice and it's an exception and it would be great if we wouldn't need that.

If the deployment of an image happens into a real disk, not a ramdisk, the standard sysroot target created by systemd through the information of root=... works without any problems. The chain of events in this case is like this:

* dump of raw disk image to a storage device
* kexec into the new system
* udev automatically recognizes the new devices, e.g root partition
* udev creates all devices according to its rules
* the systemd generator for sysroot identifies the device provided via the root= kernel cmdline and is able to create a mount unit

done

Now the process using a ramdisk device

* dump of raw disk image to a brd device. kiwi has to manually load the brd driver prior dump and does this if rd.kiwi.ramdisk is set. Once brd module is loaded some /dev/ramX devices exists
* kexec boot is not possible because that would wipe the ramdisk contents. Therefore the boot into the system must be done from the environment that has dumped the image. In order to get partition devices kiwi calls: ```kpartx -s -a "${image_target}"```. This call kicks udev and all devices e.g root partition are created
* For some reason the systemd generator for sysroot has not created a unit file. This was the reason why I added kiwi-ramdisk-deployment-generator in pre-mount formerly

Maybe we should look for a different solution to the problem. Is it possible to tell systemd to re-run the generator for sysroot after we called ``kpartx -s -a "${image_target}"``` ? I was thinking at this code path:

```bash

... make sure the config file exists

if getargbool 0 rd.kiwi.ramdisk; then
    # For ramdisk deployment a kexec boot is not possible...
    image_target=$(get_selected_disk)
    kpartx -s -a "${image_target}"
    # make sure systemd sysroot generator is kicked...
else
    ...
fi
```

and no extra mount code from our side needed

Thoughts ?

davidcassany: > I would go with option 2 for this use case. I don't see how a dracut mount hook would be called when there is no /sysroot mountpoint. Isn't that a chicken&egg problem.

This does not relay on mount units, this executes custom user provided hook (as the pre-mount hook) scripts. So, if we actually mount that manually, which I am not a fan of, I believe this is the point where it should be mounted. See dracut code here 
https://github.com/dracutdevs/dracut/blob/7c2521091f30d7c14caaf96850208afa2fe1d811/modules.d/98dracut-systemd/dracut-mount.sh#L16-L29

dracut-mount.service executes all mount hooks one after the other until sysroot is mounted or until the timeout is reached. So I assume if we follow the path of manually mounting sysroot, dracut would expect us to do so in a mount hook. Anyway I am opened to other solutions based on systemd as I am not a fan of this solution.

>  systemd does this to have control over all mount points of the system, which is a good concept

yes I agree, that's why I initially tried the `systemd-mount` approach.

> For some reason the systemd generator for sysroot has not created a unit file. This was the reason why I added kiwi-ramdisk-deployment-generator in pre-mount formerly

It is not created because there is no `root=` argument to pass. As a result systemd continues without having any dependency on `sysroot.mount` and it is `dracut-mount.sh` and probably other steps that check if `/sysroot` is mounted or not, without relaying in systemd status.

> Maybe we should look for a different solution to the problem. Is it possible to tell systemd to re-run the generator for sysroot after we called ``kpartx -s -a "${image_target}"``` ? I was thinking at this code path:

This is the current released approach. The generator was currently re executed in pre-mount hook and modifying the initrd-root-fs.target transaction (which was about to finalize or even finalized at that point). I'd say that if we want to go with the systemd way we should do something like creating a `sysroot.mount` unit (not from a generator, just create it within the pre-mount hook itself) without the common `intird-fs-root.target` dependecy and make it start before `dracut-mount.service`, so we are certain that at least during the execution of `dracut-mount.service` the mount unit will be handled.

Between, I found another potential issue on the current generator for sysroot on ramdisk deployments. The unit file does not include `DefaultDependencies=no` otherwise dependencies are included on earlier services [see the docs](https://www.freedesktop.org/software/systemd/man/systemd.mount.html#Default%20Dependencies).

So I agree on using a sysroot.mount, but I don't think we should do that based on a generator. I would just directly dump the `sysroot.mount` file within the pre-mount hook and make sure the dependencies of this unit are sane in that boot stage to no mess with ongoing transactions.

schaefi: >So I agree on using some sysroot.mount created by us, but I am not sure doing that using a generator is the best way to do it. I would just directly dump the sysroot.mount file within the pre-mount hook and make sure the dependencies of this unit are sane in that boot stage to no mess with ongoing transactions. Basically this is my fourth option.

For me this would make sense as it is the closest approach to the default behavior

davidcassany:
@schaefi @fbuihuu ok I done the PR and now a mount unit is created within the `dracut-pre-mount.service` and added as a dependency to `initrd-fs.target`.

I do not run any specific execution to `systemctl daemon-reload` as this is already happening as part of the boot process right after the `dracut-pre-mount.service`, more specific, this is done by the `initrd-parse-etc.service`, the service that runs just after the pre-mount stage which only does a `daemon-reload` and starts `initrd-fs.target`. I tested it TW and SLE-15-SP3-Beta and worked fine.

fbuihuu:
> @schaefi @fbuihuu ok I done the PR and now a mount unit is created within the `dracut-pre-mount.service` and added as a dependency to `initrd-fs.target`.

Well `initrd-fs.target` is the wrong target to use here. It's used as synchronization point by all services that mount fs listed in /sysroot/etc/fstab, see man systemd.special(7).

The boot ordering is described in dracut.bootup(7) man page BTW.

You still chose to generate sysroot.mount during boot and to include it in the initial transaction. Again it's not the easiest solution, you could simply use the "mount" hook to mount /sysroot (by using mount(8))  since apparently this hook can also be used to run /sysroot, see https://github.com/dracutdevs/dracut/blob/7c2521091f30d7c14caaf96850208afa2fe1d811/modules.d/98dracut-systemd/dracut-mount.sh#L13.

The fact that `sysroot.mount` doesn't exist in your case doesn't really matter I believe. It's the mechanism used to mount /sysroot when `root=` is specified in the command line but for other cases you can use any other mean.  The only thing that matters is `initrd-root-fs.target`. It's the target to use when units wants to be scheduled after /sysroot is mounted.

`sysroot.mount` has no special meaning in initrd and doesn't exist anymore once the system has switched to rootfs.

davidcassany:
@fbuihuu allright, thanks again for your feedback. I don't think there is a chance to mount sysroot via systemd then if `initrd-fs.target` sync point is already too late in the boot process, as before that the obvious one is `initrd-root-fs.target` and this what we are trying to avoid.

Gonna make a third proposal using a mount hook.

fbuihuu:
It's probably possible to generate a unit in pre-mount hook which would be ordered before `initrd-root-fs.target` and started from either `pre-mount` or `mount` hook. But as I wrote I don't see any advantage over mounting sysroot directly from the mount hook, it just adds unnecessary complexity it seems.

davidcassany:
> It's probably possible to generate a unit in pre-mount hook which would be ordered before `initrd-root-fs.target` and started from either `pre-mount` or `mount` hook. But as I wrote I don't see any advantage over mounting sysroot directly from the mount hook, it just adds unnecessary complexity it seems.

Starting the mount unit on pre-mount stage is something I already tried yesterday and it did not really work (I am not sure I added `Before=initrd-root-fs.target` though, I think I did it without any dependency at all). The command `systemctl start sysroot.mount` executed in `pre-mount` hook resulted in a dead lock and adding the  `--no-block` seams to be just what we are trying to avoid. Probably doing so in a `mount` just works fine, but I agree that feels complex adding a mount hook just to `systemctl daemon-reload; systemctl start sysroot.mount`

I just did simple mount execution in a mount-hook now and seams pretty clean. I also verified that the booted systems seams to have the equivalent basic units as a regular disk boot. In the sense there is no `sysroot.mount`  present after switch root and the new root includes a service entry listed as `-.mount    loaded active    Root Mount`.

davidcassany:
@schaefi what you think about the current suggestion? Finally I think this is a reasonable option.

schaefi:
> @schaefi what you think about the current suggestion? Finally I think this is a reasonable option.

yes this looks clean to me and really calls at call time. Let's go for it

aplanas:
\o/ thanks!

PR #1634, opened 2020-11-24
Fixed spec file microdnf requirement

schaefi:
SUSE/SLES doesn't provide microdnf within the official channels

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -136,7 +136,7 @@ Recommends:     gnupg2
 Recommends:     dnf
 Recommends:     gpg2
 %endif
-%if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version} >= 1550 || 0%{?sle_version} >= 150200
+%if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version} >= 1550
 Provides:       kiwi-packagemanager:microdnf
 Requires:       microdnf
 %endif


Conan-Kudo (changes requested):
NACK. This is intentional for openSUSE. If SLE 15 needs different, please strip it in the appropriate downstream repo.

Conan-Kudo:
Alternatively, we can change `0%{?sle_version} >= 150200` to `0%{?sle_version} >= 150300`, so that we get this for SLE 15 SP3.

schaefi:
Hmm, we have it the way it is done in this PR for the dnf package, why is microdnf different ? I don't see it to be part of SLE and if it should be there this requires a jira ticket and a negotiation with the SLE people that we need it in SLE. As the conversation about dnf are currently at the beginning I doubt we can have microdnf in SLE

So we have this for dnf

```
%if 0%{?fedora} || 0%{?rhel}
Requires:       dnf
```

I thought it should be the same for microdnf 

confused ?

schaefi:
oh yes and I see it to be present in openSUSE. So this only bites me on SLE

Conan-Kudo:
@schaefi The difference is that I'm going through and actually implementing stuff in openSUSE to make the transition possible. As it currently stands, I have to ignore SUSE Linux Enterprise while I work through the effort on the openSUSE side, since I have received no useful, actionable feedback to move things forward.

Conan-Kudo:
Also, to note: there's no way to have a difference between openSUSE Leap and SLE anymore with openSUSE Jump being implemented. I'm reasonably confident we can have microdnf pulled into SLE 15.3 (it's already part of PackageHub for SLE 15.2) as I work through the implementation for openSUSE.

Conan-Kudo (approved):

schaefi:
> Also, to note: there's no way to have a difference between openSUSE Leap and SLE anymore with openSUSE Jump being implemented. I'm reasonably confident we can have microdnf pulled into SLE 15.3 (it's already part of PackageHub for SLE 15.2) as I work through the implementation for openSUSE.

I see everything available on openSUSE, but I have trouble for SLE and I'm sure new software still has to follow the process, which means an ECO that can take a long time. We should not require for SLE as long as it does not exist. This will block our 3months SLE refresh.

How about the change now ?

Conan-Kudo (changes requested):
Actually, we need the following for DNF/YUM:

```
%if 0%{?fedora} || 0%{?rhel} || 0%{?suse_version} >= 1550
Provides:       kiwi-packagemanager:dnf
Provides:       kiwi-packagemanager:yum
Requires:       dnf
%endif

Since we need this for older RHEL releases too...

schaefi:
Once I see dnf/microdnf in the SLE channels for customers we can increase the scope of the requirement

Conan-Kudo:
@schaefi The DNF/YUM dependency needs to be tweaked into a separate conditional block, but otherwise that's fine.

schaefi:
> @schaefi The DNF/YUM dependency needs to be tweaked into a separate conditional block, but otherwise that's fine.

ok will do

schaefi:
@Conan-Kudo ok so this should be ok now right ?

Conan-Kudo (approved):

PR #1635, opened 2020-11-24
Add DNF as a proper dependency for openSUSE

Conan-Kudo:
This is required so that OBS can build openSUSE containers and appliances
using DNF as the package manager.


schaefi:
ok :+1:

PR #1641, opened 2020-12-02
Fixed multipath disk device assignment in kiwi lib

schaefi:
The former lookup of the multipath mapped disk device contained
a race condition. If the lookup of the device mapper files happened
before multipathd has finished the initialization, kiwi continues
with the unix node name and fails when the device mapper keeps
a busy state on it. This commit changes the code such that in case
of an explicit request to use multipath the lookup of the mapped
device becomes a mandatory process that runs until the
DEVICE_TIMEOUT is reached. Default timeout is set to 60 sec.
This references Issue SUSE-Enceladus/azure-li-services#255

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-lib.sh
@@ -19,9 +19,34 @@ function get_root_map {
     echo "${root_map}"
 }
 
+function get_mapped_multipath_disk {
+    declare DEVICE_TIMEOUT=${DEVICE_TIMEOUT}
+    local disk_device=$1
+    local check=0
+    local limit=30
+    if [[ "${DEVICE_TIMEOUT}" =~ ^[0-9]+$ ]]; then
+        limit=$(((DEVICE_TIMEOUT + 1)/ 2))
+    fi
+    udev_pending &>/dev/null
+    while true;do
+        for wwn in $(multipath -l -v1 "${disk_device}");do
+            if [ -e "/dev/mapper/${wwn}" ];then
+                echo "/dev/mapper/${wwn}"
+                return
+            fi
+        done
+        if [ "${check}" -eq "${limit}" ]; then
+            die "Multipath map for ${disk_device} did not show up"
+        fi
+        check=$((check + 1))
+        sleep 2
+    done
+}
+
 function lookup_disk_device_from_root {
     declare root=${root}
     declare kiwi_RaidDev=${kiwi_RaidDev}
+    declare kiwi_oemmultipath_scan=${kiwi_oemmultipath_scan}
     local root_device=${root#block:}
     local disk_device
     local wwn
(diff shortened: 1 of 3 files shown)


schaefi:
@davidcassany I need to test this in an image with 

```xml
<oem-multipath-scan>true</oem-multipath-scan>
```

Therefore I created the PR and if you agree I create a Staging build from this branch such that I can test it in an image

For review it would be great if you can look at the loop code for any obviously broken code

Thanks

davidcassany:
> Therefore I created the PR and if you agree I create a Staging build from this branch such that I can test it in an image

Sure, lets add this to staging and test it.

dracut/modules.d/99kiwi-lib/kiwi-lib.sh
```diff
@@ -19,9 +19,34 @@ function get_root_map {
...
+    udev_pending &>/dev/null
+    while true;do
+        for wwn in $(multipath -l -v1 "${disk_device}");do
+            if [ -e "/dev/mapper/${wwn}" ];then
+                echo "/dev/mapper/${wwn}"
+                return
+            fi
+        done
+        if [ "${check}" -eq "${limit}" ]; then
+            die "Multipath map for ${disk_device} did not show up"
```
davidcassany, line 39: This is a behavior change. Could it impact to images that are abusing of the multipath disks mapping when unneeded?

davidcassany (commented):
I have just one doubt about this slightly different behavior of this implementation.

dracut/modules.d/99kiwi-lib/kiwi-lib.sh
schaefi: > If I recall correctly the default value for the multipath scan is true, so it might be triggered in many images out there even when this is not actually really needed.

Good point. In the code we have:

```
if self.root_filesystem_is_multipath is False:
```

so only an explicit false setting omits the module. This means if multipath is installed it gets activated.

> This is a behavior change.

This is correct and wanted. With the race condition the behavior right now is basically undefined. The system could go with the mapped device or with the unix node, depending on what is present at lookup time. With the patch here we expect multipath and fail if not present.

So you are right, if multipath is installed and used on non multipath systems with this patch it will fail.

I think we should make the multipath scan set to default = false

All images I know about that are used in multipath envs have this explicitly set to true, and I think the simple install of the package should not lead to a multipath enabled image since the package gets installed by default on suse which is strange as it maps any device into a multipath

Thoughts ?

schaefi: I have done some tests including the latest commit

1.  image with <oem-multipath-scan> set to false, and also without set it all. Results in the same (default is false)

     The image initrd has no multipath and boots up using the standard unix node according to the root UUID. The multipath-tools package was installed but ```multipath -ll``` lists nothing which is expected

2. image with <oem-multipath-scan> set to true

    The image initrd has multipath and performs the lookup of the disk mapped against the device mapper node from multipath. ```multipath -ll``` lists a QEMU disk assigned to a multipath map and provides the expected setup (just one path, of course)

3. image with <oem-multipath-scan> set to true but no multipath-tools installed

   The image fails at boot time, after the 60sec timeout because no multipath map exists due to the missing tools in the initrd

From my perspective this should be the desired behavior. I disliked the implicit behavior and it also had a race condition

davidcassany: > I think we should make the multipath scan set to default = false

I am 100% on that. I believe that if not set better not mess with it, so avoid it unless it is explicitly requested. Having this default modified would have also avoided some of the issues we have been asked time to time about not being capable to mount certain devices when multipath is in use.

schaefi: I think so too

PR #1646, opened 2020-12-08
Fixed validation of bool value in dracut module

schaefi:
The oem-multipath-scan setup results in a bool variable inside
of the initrd code. The variable kiwi_oemmultipath_scan is
therefore either set to "true" or "false". A check in code
of the form [ -n ... ] is stupid since the variable always
contains text. This commit fixes the validation to make use
of the bool() method provided for these type of variables


schaefi:
@davidcassany please review on high prio. I'm so stupid

davidcassany (approved):
LGTM

PR #1648, opened 2020-12-09
Make debootstrap repository explicit

### LINKED ISSUE #1593: debian: wrong "distribution" used for bootstrap stage
## Problem description

When building a Debian system with kiwi from multiple repositories that especially refer to multiple values of the "distribution" field in repository urls, kiwi will call debootstrap with the repository value of the last specified repsoitory, rather than a (configurable?) base.

```
[ ERROR   ]: 15:41:03 | KiwiDebootstrapError: debootstrap script for buster/updates distribution not found
[ INFO    ]: 15:41:03 | Cleaning up SystemPrepare instance
```

## Expected behaviour

debootstrap will be called for the base distribution, e.g. "buster".
Personally I'd like a way for declaring this value, e.g. by
```
<packages type="bootstrap" distribution="buster">
...
</packages>
```

or perhaps mark the repository for bootstrapping
```
<repository type="apt-deb" distribution="buster-updates" components="main non-free" repository_gpgcheck="true" bootstrap="true">
        <source path="http://deb.debian.org/debian"/>
</repository>
```

Sample config.xml:
```
<image name="sr-imx6">
        <description type="system">
                <author>Josua Mayer</author>
                <contact>support@solid-run.com</contact>
                <specification>text</specification>
        </description>
        <preferences>
                <version>0.1</version>
                <packagemanager>apt-get</packagemanager>
                <type image="ext4"/>
                <keytable>us-intl</keytable>
                <timezone>CET</timezone>
                <locale>en_GB</locale>
                <rpm-check-signatures>true</rpm-check-signatures>
        </preferences>
        <users>
                <user name="root" groups="root" pwdformat="plain" password="linux" home="/root"/>
                <user name="debian" groups="debian" pwdformat="plain" password="debian" home="/home/debian"/>
        </users>
        <repository type="apt-deb" distribution="buster" components="main non-free" repository_gpgcheck="true">
                <source path="http://deb.debian.org/debian"/>
        </repository>
        <repository type="apt-deb" distribution="buster-updates" components="main non-free" repository_gpgcheck="true">
                <source path="http://deb.debian.org/debian"/>
        </repository>
        <repository type="apt-deb" distribution="buster/updates" components="main non-free" repository_gpgcheck="true">
                <source path="http://security.debian.org/"/>
        </repository>
        <packages type="bootstrap">
                <!-- additional, packages for debootstrap -->
                <package name="busybox"/>
                <package name="curl"/>
                <package name="locales"/>
                <package name="initramfs-tools"/>
                <package name="systemd-sysv"/>
        </packages>
        <packages type="image">
                <!-- system packages: debian -->
                <package name="alsa-utils"/>
                <package name="command-not-found"/>
                <package name="connman"/>
                <package name="i2c-tools"/>
                <package name="iw"/>
                <package name="linux-image-armmp"/>
                <package name="mtd-utils"/>
                <package name="nano"/>
                <package name="ntfs-3g"/>
                <package name="openssh-server"/>
                <package name="pciutils"/>
                <package name="psmisc"/>
                <package name="rfkill"/>
                <package name="usbutils"/>
                <package name="wget"/>
        </packages>
</image>
```

+ KIWI version: KIWI (next generation) version 9.21.17
+ OBS version: -

davidcassany:
This commit introduces the use_for_debootstrap attribute for
repositories of type apt-deb. This is a boolean attribute to specify
the repository that will be used for bootstrapping in apt-deb based
images. Only one can be selected and if none is specified KIWI just
makes use of the last one in the list.

Fixes #1593

### DIFF
--- a/kiwi/repository/base.py
+++ b/kiwi/repository/base.py
@@ -64,7 +64,7 @@ class RepositoryBase:
     def add_repo(
         self, name, uri, repo_type, prio, dist, components,
         user, secret, credentials_file, repo_gpgcheck, pkg_gpgcheck,
-        sourcetype
+        sourcetype, use_for_bootstrap=False
     ):
         """
         Add repository
@@ -83,6 +83,7 @@ class RepositoryBase:
         :param bool repo_gpgcheck: unused
         :param bool pkg_gpgcheck: unused
         :param str sourcetype: unused
+        :param bool use_for_bootstrap: unused
         """
         raise NotImplementedError
 
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -906,6 +906,18 @@ div {
 # common element <repository>
 #
 div {
+    sch:pattern [
+        abstract = "true"
+        id = "repo_type"
+        sch:rule [
+            context = "repository[@$attr]"
+            sch:assert [
+                test = "contains('$types', @type)"
+                "$attr attribute is only available for the following "
+                "repository types: $types"
+            ]
+        ]
+    ]
     k.repository.profiles.attribute = k.profiles.attribute
     k.repository.type.attribute =
         ## Type of repository
@@ -970,6 +982,25 @@ div {
         attribute sourcetype {
             "baseurl" | "metalink" | "mirrorlist"
         }
+    k.repository.use_for_bootstrap.attribute =
+        ## Specify whether this repository should be the one used for
+        ## bootstrapping or not. False by default. Only a single repository
+        ## is allowed to be used for bootstrapping. If none is set the
+        ## last one is picked.
+        attribute use_for_bootstrap { xsd:boolean }
+        >> sch:pattern [ id = "use_for_bootstrap" is-a = "repo_type"
+            sch:param [ name = "attr" value = "use_for_bootstrap" ]
+            sch:param [ name = "types" value = "apt-deb" ]
+        ]
+        >> sch:pattern [ id = "single_deboostrap_repo"
+            sch:rule [ context = "image"
+                sch:assert [ 
+                    test = "count(repository[@use_for_bootstrap='true'])<=1"
+                    "There can only be a single repository set for "
+                    "bootstrap ('use_for_bootstrap' attribute)"
+                ] 
+            ]
+        ]
     k.repository.attlist =
         k.repository.type.attribute? &
         k.repository.profiles.attribute? &
@@ -985,7 +1016,8 @@ div {
         k.repository.package_gpgcheck.attribute? &
         k.repository.priority.attribute? &
         k.repository.password.attribute? &
-        k.repository.username.attribute?
+        k.repository.username.attribute? &
+        k.repository.use_for_bootstrap.attribute?
     k.repository =
         ## The Name of the Repository
         element repository {
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -768,6 +768,13 @@ sourcetype="baseurl|metalink|mirrorlist"
   list, the configured package manager needs to be setup appropriately.
   By default the source is expected to be a simple repository baseurl.
 
+use_for_bootstrap="true|false"
+  Used for Debian (apt) based repositories only. It specifies whether
+  this repository should be the one used for bootstrapping or not.
+  It is set to 'false' by default. Only a single repository is allowed
+  to be used for bootstrapping, if no repository is set for the bootstrap
+  the last one in the description XML is used.
+  
 <repository><source>
 ~~~~~~~~~~~~~~~~~~~~
 The location of a repository is specified by the path attribute of the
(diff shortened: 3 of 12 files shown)


davidcassany:
@schaefi note I did not code a runtime check since the schematron rule to limit the number of `use_for_debootstrap="true"` seamed to be pretty simple and robust.

Conan-Kudo:
Actually, we can genericize this functionality:

* Changing this to `use_for_bootstrap` and plumb through the ability to optionally specify repos for bootstrap phase _only_.

Conan-Kudo (changes requested):
Let's make this generically useful for all package managers.

kiwi/repository/base.py
```diff
@@ -64,7 +64,7 @@ def runtime_config(self):
     def add_repo(
         self, name, uri, repo_type, prio, dist, components,
         user, secret, credentials_file, repo_gpgcheck, pkg_gpgcheck,
-        sourcetype
+        sourcetype, use_for_debootstrap
```
schaefi, line 67: Regarding the interface character of the API I think that parameter should be generalized, e.g ```use_for_bootstrap=True|False```

schaefi (changes requested):
Overall looks great :+1: I would like to see the interface to stay generic though

kiwi/repository/base.py
davidcassany: I am fine with it, change done. I just added it without the default at base class level as it is the case for the other parameters, but I agree that would be good to also include the default at the base class, for clarity and to keep API consistent.

davidcassany:
> Actually, we can genericize this functionality by changing this to `use_for_bootstrap` and plumb through the ability to optionally specify repos for bootstrap phase _only_. All package managers have a two phase image build, and giving the flexibility to specify this stuff for bootstrap could come in handy.

I have a couple of concerns with this idea. First, implementation wise, this is is not that simple, as all package managers have a two phase (bootstrap and chrooted install), but the package manager config is build only once and used for both phases. Second, I am struggling to figure out a use case for it.

In any case, if we have a clear use case I'd be certainly happy to consider implementing this. Do get me wrong, so far what I see is a complex feature without a use case.

What I'd do now is proceed with the current PR, specific for apt-deb repositories, and open a new issue to discuss making it generic and extend it to the other package managers. 

@Conan-Kudo @schaefi  does it make sense?

davidcassany:
Gonna change the attribute name to `use_for_bootstrap` so this might be easily reused later on for non debian based distros if we see fit. I missed this detail on my first read of @Conan-Kudo's comment.

Conan-Kudo:
@davidcassany The main case I have for bootstrap repos is having repos available only during the bootstrap phase for custom package manager things (which should not be available in the final image).

schaefi:
>What I'd do now is proceed with the current PR, specific for apt-deb repositories, and open a new issue to discuss making it >generic and extend it to the other package managers.
>@Conan-Kudo @schaefi does it make sense?

makes sense to me

>The main case I have for bootstrap repos is having repos available only during the bootstrap phase for custom package >manager things (which should not be available in the final image).

This would introduce another layer of repos and is in my opinion a bigger topic to discuss. From an integrity perspective it feels odd to me to install packages into a system that comes from a set of repos which are unknown to the end-user running the image. It's the same approach like:

```
add-repo A
install package-from-repo-A
delete-repo A
...let the poor package manager mess with this situation now
```

I'm happy to discuss this further in an extra issue :smiley: 

Other than that I think the proposed patch will solve the issue of the reporter in #1593 and would be good to become effective

schaefi (approved):

kiwi/schema/kiwi.rnc
```diff
@@ -970,6 +982,25 @@ div {
         attribute sourcetype {
             "baseurl" | "metalink" | "mirrorlist"
         }
+    k.repository.use_for_bootstrap.attribute =
+        ## Specify whether this repository should be the one used by de
+        ## debootstrap or no. False by default. Only a single repository
```
Conan-Kudo, line 987: Please remove references to debootstrap.

Conan-Kudo (changes requested):

kiwi/schema/kiwi.rnc
davidcassany: Those are left on purpose since they are pure literals for the user and currently this is only being used to workaround debootstrap limitations.

Conan-Kudo: as a generic property that we can extend elsewhere, we should not have those assumptions exposed.

davidcassany: ok updated

schaefi:
@davidcassany one thing I noted. There is no update of the documentation regarding the new use_for_bootstrap attribute.
Could you please add a note in the docs, maybe here: doc/source/image_description/elements.rst

Thanks

davidcassany:
Documentation added.

schaefi:
@davidcassany great thanks :+1:

This is good, let's make it effective. I'm going to merge it now

PR #1649, opened 2020-12-09
Add vagrant setup function

### LINKED ISSUE #1127: Provide a vagrant_setup() shell script for use in config.sh
## Problem description

Vagrant has certain expectations towards the layout of the boxes, namely:
- the user `vagrant` exists, has passwordless sudo rights and it's password is `vagrant`
- ssh is running
- the [vagrant insecure key](https://github.com/hashicorp/vagrant/blob/master/keys/vagrant.pub) is in `/home/vagrant/.ssh/authorized_keys`.
- (optional, but usually the case) a shared folder is set up in `/vagrant/`

The necessary steps are explained in the kiwi docs, but currently every box creator has to apply them themselves. I think we should provide a shell script for usage in `config.sh` that will apply these settings.

I've had a function like that in the original PR that added virtualbox vagrant support, but that got rejected. I'm asking for that to be reconsidered, since such a function would be useful for most people. Those who have different requirements, will likely know that and won't use it.

dcermak:
Fixes #1127

* Add helper functions `baseVagrantSetup` and `suseVagrantSetup` to run common vagrant setup tasks
* Document both functions
* fix `Debug` failing to run when `config.sh` is run with `set -u` and `DEBUG` is not set

### DIFF
--- a/doc/source/concept_and_workflow/shell_scripts.rst
+++ b/doc/source/concept_and_workflow/shell_scripts.rst
@@ -206,7 +206,7 @@ suseSetupProductInformation
 
 Debug {message}
   Helper function to print the supplied message if the variable DEBUG is
-  set to 1.
+  set to 1 (it is off by default).
 
 Echo {echo commandline}
   Helper function to print a message to the controlling terminal.
--- a/doc/source/concept_and_workflow/shell_scripts.rst
+++ b/doc/source/concept_and_workflow/shell_scripts.rst
@@ -204,6 +204,15 @@ suseSetupProductInformation
   and installs all product specific packages. This function fails
   when :command:`zypper` is not the appliances package manager.
 
+baseVagrantSetup
+  Configures the image to work as a vagrant box by performing the following
+  changes:
+  - add the ``vagrant`` user to :file:`/etc/sudoers` or
+    :file:`/etc/sudoers.d/vagrant`
+  - insert the insecure vagrant ssh key, apply recommended ssh settings and
+    start the ssh daemon
+  - create the default shared folder :file:`/vagrant`
+
 Debug {message}
   Helper function to print the supplied message if the variable DEBUG is
   set to 1 (it is off by default).
--- a/doc/source/concept_and_workflow/shell_scripts.rst
+++ b/doc/source/concept_and_workflow/shell_scripts.rst
@@ -86,8 +86,9 @@ Functions
 ^^^^^^^^^
 
 The :file:`.kconfig` file provides a common set of functions.  Functions
-specific to SUSE Linux begin with the name ``suse``, functions applicable
-to all Linux distributions start with the name ``base``.
+specific to SUSE Linux Enterprise and openSUSE begin with the name
+``suse``, functions applicable to all Linux distributions start with the
+name ``base``.
 
 The following list describes all functions provided by :file:`.kconfig`:
 
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -1137,4 +1137,39 @@ function baseQuoteFile {
     mv "${conf}" "${file}"
 }
 
+#======================================
+# baseVagrantSetup
+#--------------------------------------
+function baseVagrantSetup {
+    # insert the default insecure ssh key from here:
+    # https://github.com/hashicorp/vagrant/blob/master/keys/vagrant.pub
+    mkdir -p /home/vagrant/.ssh/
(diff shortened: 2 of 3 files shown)


dcermak:
cc @Vogtinator, this might be useful for you

doc/source/concept_and_workflow/shell_scripts.rst
```diff
@@ -204,9 +205,28 @@ suseSetupProductInformation
...
+
+suseVagrantSetup
+  Additionally to ``baseVagrantSetup``, this function also performs the
+  following actions:
+  - if the package ``virtualbox-guest-tools`` is installed, load the
+    ``vboxsf`` module
+  - remove all udev rules for networking that match
+    :file:`/etc/udev/rules.d/*-net.rules` so that the wired network is called
+    ``eth0`` if the kiwi profile is not called ``virtualbox``
+  - setup DHCP on ``eth0``
```
Conan-Kudo, line 225: Why is `suseVagrantSetup` needed? Why isn't the package providing this setup like the Fedora `virtualbox-guest-additions` package does?

Conan-Kudo (changes requested):

doc/source/concept_and_workflow/shell_scripts.rst
dcermak: At least the udev rules removal & dhcp config is really only applicable for vagrant boxes (but it's distro specific, I have not checked if this would work in Fedora).

My guess concerning the `virtualbox-guest-additions` package is, that not everyone who installs virtualbox, wants to have the `vboxsf` kernel module.

kiwi/config/functions.sh
```diff
@@ -1137,4 +1137,65 @@ function baseQuoteFile {
...
+    # start vboxsf service only if the guest tools are present
+    if Rpm -q virtualbox-guest-tools 2> /dev/null; then
+        echo vboxsf > /etc/modules-load.d/vboxsf.conf
+    fi
+
+    # drop any network udev rules for libvirt, so that the networks are called
+    # ethX
+    # this is not required for Virtualbox as it handles networking differently
+    # and doesn't need this hack
+    if [ "${kiwi_profiles}" != "virtualbox" ]; then
```
Vogtinator, line 1190: I don't think this should behave differently depending on profile names. What about a `suseSetupDHCP` function which the script can call?

kiwi/config/functions.sh
```diff
@@ -1137,4 +1137,65 @@ function baseQuoteFile {
...
+        Rm -f /etc/udev/rules.d/*-net.rules
```
Vogtinator, line 1191: `net.ifnames=0` as kernel parameter might be better than deleting udev rules. IMO `rm` calls like this aren't a good idea as they'll just come back after the next update installation.

kiwi/config/functions.sh
dcermak: That step is only required for vagrant-libvirt, as that uses dnsmasq for assigning IPs. Virtualbox does that differently and the modifications are not necessary.

kiwi/config/functions.sh
dcermak: Hm, after digging into this a bit I've found out that I already have this setting in the image description and it was (probably) not sufficient.

You're right concerning updates, but that's not really a typical workflow for vagrant boxes. These tend to get redeployed instead of getting updates.

schaefi: I agree to what Fabian said, deleting udev rules that way is unexpected

kiwi/config/functions.sh
schaefi: The generic method should not depend on a profile name. The setup of a profile is an optional setting and it can be used as part of script code that belongs to the rest of an image description which has the profile configured. However for a generic method considered a helper for people building vagrant images you can't expect the profile "virtualbox" to exist.

Furthermore the code would kick in if a profile is named "virtualbox" but it's not guaranteed that this has something to do with your context. Agreed this is a constructed example and very unlikely to happen but any code provided as generic helpers should not be image description specific

schaefi: > What about a suseSetupDHCP function which the script can call?

sounds like a good approach to me too

doc/source/concept_and_workflow/shell_scripts.rst
schaefi: > that not everyone who installs virtualbox, wants to have the vboxsf kernel module.

I think if the image description has provider="virtualbox" virtualbox_guest_additions_present="true" set and vboxsf kernel module exists it is wanted. The kiwi image description is available below /image when your script runs. It would be possible to check base on those values to come to a decision

> At least the udev rules removal 

This should not be done imho

> dhcp config 

If you consider to put that code into it's own method you could make only the dhcp setup distro specific.
Regarding that network setup I'm not sure if we want to provide code to do it. The network setup is very
different between all distributions and I would rather document that step instead of writing extra code.
Alternatively we could utilize existing tools to create the network setup if possible.

In the end I think suseVagrantSetup should be avoided and we should aim for a generic method as much
as possible. Just my2cents

kiwi/config/functions.sh
Conan-Kudo: Is there a reason we don't just use NetworkManager and have it autoconfigure network stuff inside?

schaefi: as I said there are several ways to setup a network and every distro has their own concepts. I think network setup is out of scope and should be an individual tasks of the author of the image description. With regards to the vagrant setup a network is probably mandatory so this aspect and maybe an example should be ok as documentation

dcermak:
I have dropped the `suseVagrantSetup` function, as I think it does not belong into kiwi as you guys have pointed out: it's just too specific to how I build the vagrant boxes.

Conan-Kudo (approved):

PR #1652, opened 2020-12-11
Custom swap volume name

### LINKED ISSUE #1638: Allow custom swap volume name setup
## Problem description
Currently swap volume name is fixed. User wants to standardize the swap volume name but it doesn't seem to be configurable.

## Expected behaviour
One should be able to define the swap volume name.
Allow custom root volume name setup

+ KIWI version: python3-kiwi-9.21.26
+ Operating system: SLES15SP2
+ OBS version:

schaefi:
Added a new optional element inside ```<oemconfig>```:

```xml
<oemconfig>
    <oem-swapname>swap_volume_name</oem-swapname>
</oemconfig>
```

The setting allows to specify a name for the swap volume in case the LVM volume manager
is used. The default if not specified continuous to stay at: LVSwap.

This Fixes #1638

### DIFF
--- a/doc/source/building_images/build_expandable_disk.rst
+++ b/doc/source/building_images/build_expandable_disk.rst
@@ -355,6 +355,14 @@ oemconfig.oem-swap Element
   swap partition will be created. This value is represented
   by the ``kiwi_oemswap`` variable in the initrd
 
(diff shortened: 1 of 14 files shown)


davidcassany (approved):
Looks good

PR #1654, opened 2020-12-14
Move CI system to gitlab

schaefi:
Travis made questionable changes to their open statement
and I'm not following this anymore.


schaefi:
The gh-pages provides is still missing... need to check how we can do this with gitlab

helper/deploy_pages.sh
```diff
@@ -0,0 +1,44 @@
...
+repo_uri="https://x-access-token:${access_token}@github.com/OSInside/kiwi.git"
+remote_name="origin"
+target_branch="gh-pages"
+
+# set git user and email
+git config user.name "${user_name}"
+git config user.email "${user_mail}"
+
+# build kiwi documentation suitable for github pages
+tox -e doc_travis
```
davidcassany, line 19: I find it confusing... I understand we are not using travis to build docs, just the `travis-sphinx` python tool. We use it only to build the docs, not even to push them to giuthub as I understand this is done later on in this script.

To me all these references to travis are confusing, unless you really pay attention to the code it looks like docs are all build on travis. I'd say we should just keep `travis` literal for the `travis-sphinx` tool and name `tox` job something different than `doc_travis` and use a different temporary folder than `build_travis`.

davidcassany (changes requested):
I am not sure if we can avoid using the `travis-sphinx` tool, looks weird using it without travis around, but I can live with it. What I believe we should change is the docs build tox stages name and temporary folders. All those references to Travis are confusing to me. If there is no Travis around we should change those otherwise it is really misleading to anyone having a look to the code.

PR #1655, opened 2020-12-14
Fixed source rpm check in gitlab

schaefi:
Fedora30 is EOL, thus there is no provider for
mock/fedora-30-x86_64.cfg anymore


Conan-Kudo (approved):

schaefi:
yay works :)

davidcassany (approved):
Now the pipe is fixed! :+1:

PR #1656, opened 2020-12-15
Delete traces of travis integration

schaefi:
Rename and clarify code that was still using the name travis.
Delete all references to the travis CI system from kiwi


davidcassany (approved):
Yes, this is what I had in mind :+1:

PR #1659, opened 2020-12-16
Add 'arch' attribute to preferences

### LINKED ISSUE #1640: Allow arch-specific <preferences/>
The use-case is having a single `.kiwi` file which builds various flavors for various architectures, e.g.

```
kvm-and-xen: aarch64, x86_64
RaspberryPi3: aarch64
```

The `<preferences/>` element contains arch-specific info, like the kernel parameters and disk setup. So it's necessary to separate `<preferences/>` for each platform.

Currently it's possible to work around it by using profiles, but this is not straightforward:

```
<profile name="kvm-and-xen_x86_64" description="kvm-and-xen for x86_64" arch="x86_64"/>
<profile name="kvm-and-xen_aarch64" description="kvm-and-xen for aarch64" arch="aarch64"/>
<profile name="kvm-and-xen" description="kvm-and-xen" arch="x86_64">
    <requires profile="kvm-and-xen_x86_64"/>
</profile>
<profile name="kvm-and-xen" description="kvm-and-xen" arch="aarch64">
    <requires profile="kvm-and-xen_aarch64"/>
</profile>
<profile name="RaspberryPi" description="RaspberryPi (3+, aarch64)" arch="aarch64"/>
[...]
<preferences profiles="kvm-and-xen_aarch64">
    [...]
</preferences>
<preferences profiles="kvm-and-xen_x86_64">
    [...]
</preferences>
<preferences profiles="RaspberryPi">
    [...]
</preferences>
```

It also requires that each profile is essentially defined twice, just the arch and dependencies are different.

It would be simpler like this:

```
<profile name="kvm-and-xen" description="kvm-and-xen" arch="aarch64,x86_64"/>
<profile name="RaspberryPi" description="RaspberryPi (3+, aarch64)" arch="aarch64"/>
[...]
<preferences profiles="kvm-and-xen" arch="x86_64">
    [...]
</preferences>
<preferences profiles="kvm-and-xen" arch="aarch64">
    [...]
</preferences>
<preferences profiles="RaspberryPi">
    [...]
</preferences>
```

(The `arch` attribute of the profile could also be omitted then, as it can be inferred from the preferences instead.

davidcassany:
This commits adds the attribute 'arch' to preferences. It works
as any other 'arch' attribute within the schema. Preferences defined
with architectures that do not match the host are ignored. If no 'arch'
is provided it matches all any host architecture.

Fixes #1640

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -2811,8 +2811,10 @@ div {
 #
 div {
     k.preferences.profiles.attribute = k.profiles.attribute
+    k.preferences.arch.attribute = k.arch.attribute
     k.preferences.attlist =
-        k.preferences.profiles.attribute?
+        k.preferences.profiles.attribute? &
+        k.preferences.arch.attribute?
     k.preferences =  
         ## Configuration Information Needed for Logical Extend
         ## All elements are optional since the combination of appropriate
--- a/kiwi/schema/kiwi.rng
+++ b/kiwi/schema/kiwi.rng
@@ -4333,10 +4333,18 @@ or plusRecommended</a:documentation>
     <define name="k.preferences.profiles.attribute">
       <ref name="k.profiles.attribute"/>
     </define>
+    <define name="k.preferences.arch.attribute">
+      <ref name="k.arch.attribute"/>
+    </define>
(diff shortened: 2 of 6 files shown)


schaefi (approved):
This is great :+1:

Issue #1644, opened 2020-12-07
Public API Strong Typing


schaefi:
## Problem description

For using kiwi as API at least the public interface should be strong typed such that users of the API know how to call the methods and that python can verify and insist on the call to be correct. In kiwi we haven't covered much code of the public API
to be annotated with type hints. The following list creates groups of KIWI code which we will start to update to become strong typed in the order of their entries:

- [x]  Config reading/writing, xml,markup, logging
- [x] Preparation of new root, system/prepare/setup/profile,root_bind
- [x] Storage processing, loop/disk/partitions/filesystems/volumes
- [x] Package handling, package_manager/repositories
- [x] Disk builder
- [x] Filesystem builder
- [x] KIS builder
- [x] Install media builder
- [x] Live builder
- [ ] Container builder, containers, archive, root_import
- [ ] Bootloader, grub,isolinux
- [ ] Utils

schaefi:
Startung with 1. covers:

- [ ] logger

schaefi:
working on the first shot

schaefi:
First round done, card moved back to the backlog. We are revisiting this one in January

isbm:
LOL :heart:

schaefi:
Continue with

* Preparation of new root, system/prepare/setup/profile,root_bind

schaefi:
- [ ] system/prepare.py
- [ ] system/profile.py
- [ ] system/setup.py
- [ ] system/root_bind.py
- [ ] system/root_init.py
- [ ] system/identifier.py
- [ ] system/kernel.py
- [ ] system/result.py
- [ ] system/shell.py
- [ ] system/size.py
- [ ] system/uri.py
- [ ] system/users.py

schaefi:
I'll work on the first three now (prepare/profile/setup)

jesusbv:
I will work on `users`, `uri` and `size`

jesusbv:
I will work on `root_bind`, `root_init` and `identifier`

jesusbv:
I will work on `kernel`, `result` and `shell`

schaefi:
Thanks much @jesusbv, well done :+1: For this sprint we are good. Thus I moved the issue back to the backlog. We will agree and continue on the next step in the February sprint meeting.

schaefi:
For the next sprint I think we should cover:

- [ ] storage/device_provider.py
- [ ] storage/disk.py
- [ ] storage/loop_device.py
[...]

PR #1660, opened 2020-12-16
Strong typing first round

schaefi:
    
* logger
* logger_color_formatter
* logger_filter
* xml_description
* xml_state
* markup/any
* markup/base
* markup/xml
    
This references Issue #1644

### DIFF
--- a/kiwi/markup/base.py
+++ b/kiwi/markup/base.py
@@ -15,14 +15,11 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
-import os
 from tempfile import NamedTemporaryFile
 from lxml import etree
 
 # project
 from kiwi.defaults import Defaults
-
-from kiwi.exceptions import KiwiDescriptionInvalid
 
 
 class MarkupBase:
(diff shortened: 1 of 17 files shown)


schaefi:
@davidcassany it took longer than I expected to get there. I know we agree to not change the behavior of the code but in one point I was not able to follow it.

In kiwi we supported reading the description as file (str) and as xml_content (bytes). All this in the same parameter. Of course this cannot be put into one type and was also bad practice. The reading from xml_content is not used by people and only tested in terms of the schema extension feature. It's easy to use a temporary file to handle raw content and so I decided to get rid of this concept and only allow reading a file

I hope this is ok and you will see it on review

schaefi:
Looking on the codacy report next. This is going to be a hard ride but in the end the code will be better

davidcassany:
>  I know we agreed to not change the behavior of the code but in one point I was not able to follow it.

Well, we new there was the change of facing some inconsistences. I think it is good if we sort those our even at the cost of slightly modifying the API. I'd say this is one of the motivations to make use of typing for the public API. So we make sure we have a consistent use of variables and also make the API explicit which is good for anyone using it.

kiwi/markup/base.py
```diff
@@ -15,54 +15,39 @@
...
-                self.description = description
-            else:
-                self.description_temporary = NamedTemporaryFile()
-                with open(self.description_temporary.name, 'wb') as outfile:
-                    outfile.write(description)
-                self.description = self.description_temporary.name
-        except Exception as issue:
-            raise KiwiDescriptionInvalid(issue)
+    def __init__(self, description: str):
+        self.description = description
```
davidcassany, line 34: Makes sense this, I think it is a good idea to keep this description str variable as a single purpose. I see not much gain on keeping it just for extensions management. All in all the extensible XML feature is something to probably rethink, it turns to be quite complex extending the XML, I would even say far from obvious in a python virtual environment.

davidcassany (approved):
Looks good! I am glad we started doing so :+1:

PR #1662, opened 2020-12-17
Use the generic Dict type

davidcassany:
This commit makes use of the Dict type in the container factory classes
so these type hints aligned with the other dict related type hints in
KIWI code. This commit improves the refactor done in 99be52ba.


davidcassany:
@schaefi I saw I did not properly typed the dicts on my factory refactor while reviewing your first round of typing changes :+1:

schaefi (approved):
good catch :+1:

PR #1663, opened 2020-12-17
Set min version python requirement

schaefi:
The use of new features like type hinting and annotations
requires a python version >= 3.6

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -58,7 +58,7 @@ Source:         %{name}.tar.gz
 Source1:        %{name}-rpmlintrc
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildRequires:  gcc
-BuildRequires:  python%{python3_pkgversion}-%{develsuffix}
+BuildRequires:  python%{python3_pkgversion}-%{develsuffix} >= 3.6
 BuildRequires:  python%{python3_pkgversion}-setuptools
 %if 0%{?fedora} || 0%{?suse_version}
 BuildRequires:  fdupes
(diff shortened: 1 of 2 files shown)


schaefi:
@Conan-Kudo xUbuntu_16.04 is the only distro that fails in the matrix with this setting

Do you have a solution ?

Conan-Kudo (changes requested):
Ugh, well, I guess we have to do this. Since we do, we also need to set `python_requires` to be `>=3.6` in `setup.py`...

setup.py
```diff
@@ -44,6 +44,7 @@ def run(self):
 
 config = {
     'name': 'kiwi',
+    'python_requires': '>=3.6.2',
```
Conan-Kudo, line 47: Please make this `3.6`, not `3.6.2`

Conan-Kudo (changes requested):

setup.py
schaefi: done

package/python-kiwi-spec-template
```diff
@@ -58,6 +58,7 @@ Source:         %{name}.tar.gz
 Source1:        %{name}-rpmlintrc
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildRequires:  gcc
+BuildRequires:  python%{python3_pkgversion} >= 3.6
```
Conan-Kudo, line 61: Instead of adding this line, just change the line below it to have `>= 3.6` constraint. We require `python%{python3_pkgversion}-%{develsuffix}` at build-time.

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: ok

davidcassany (approved):

PR #1666, opened 2020-12-18
Fixed missing provides tag

schaefi:
When building WSL images the image type is set to appx.
Therefore obs is looking for what provides kiwi-image:appx
This provides tag was missing

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -127,7 +127,7 @@ Provides:       kiwi-image-wsl-requires = %{version}-%{release}
 Obsoletes:      kiwi-image-wsl-requires < %{version}-%{release}
 %if "%{_vendor}" != "debbuild"
 Provides:       kiwi-image:docker
-Provides:       kiwi-image:wsl
+Provides:       kiwi-image:appx
 %endif
 %if 0%{?suse_version}
 Requires:       umoci


package/python-kiwi-spec-template
```diff
@@ -127,6 +127,7 @@ Provides:       kiwi-image-wsl-requires = %{version}-%{release}
 Obsoletes:      kiwi-image-wsl-requires < %{version}-%{release}
 %if "%{_vendor}" != "debbuild"
 Provides:       kiwi-image:docker
+Provides:       kiwi-image:appx
 Provides:       kiwi-image:wsl
```
davidcassany, line 131: I'd say this one needs to go away, isn't it? there is no `wsl` image supported in our schema

davidcassany (changes requested):
I'd say that `kiwi-image-wsl-requires` package was never functional as there is no `wsl` image type and OBS just does `Require: kiwi-image:<image_type>`. I would probably get rid of the `kiwi-image:wsl` provides.

package/python-kiwi-spec-template
schaefi: makes sense

davidcassany (approved):

Conan-Kudo (approved):

PR #1667, opened 2020-12-18
Update kiwi-systemdeps-image-validation

schaefi:
Make python anymarkup to be only recommended. The package
does not exist on all distributions, e.g suse does not
provide it and for kiwi it's an optional plugin

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -277,7 +277,9 @@ Recommends:     jing
 %if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version}
 Requires:       python%{python3_pkgversion}-solv
 %endif
-Requires:       python%{python3_pkgversion}-anymarkup
+%if ! (0%{?rhel} && 0%{?rhel} < 8)
+Recommends:     python%{python3_pkgversion}-anymarkup
+%endif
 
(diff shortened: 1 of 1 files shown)


schaefi:
when using ```support: kiwi-systemdeps``` in obs prjconf I realized it failed for the above reason.

davidcassany (approved):
Looks good

package/python-kiwi-spec-template
```diff
@@ -277,7 +277,7 @@ Recommends:     jing
 %if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version}
 Requires:       python%{python3_pkgversion}-solv
 %endif
-Requires:       python%{python3_pkgversion}-anymarkup
+Recommends:     python%{python3_pkgversion}-anymarkup
```
Conan-Kudo, line 280: This needs to be:

```
%if ! 0%{?el7}
Recommends:     python%{python3_pkgversion}-anymarkup
%endif
```

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: thanks, updated

Conan-Kudo (approved):

PR #1668, opened 2020-12-21
Modify lsblk flags for a consistent output across distros

davidcassany:
This commit modifies the lsblk flags to make use of the list format
instead of raw output. `--list` flag seams to keep the geometry order
and produces a consistent output in several distros.

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -179,7 +179,7 @@ function get_partition_node_name {
     local part
     udev_pending
     for partnode in $(
-        lsblk -p -r -o NAME,TYPE "${disk}" | grep -E "part|md$" | cut -f1 -d ' '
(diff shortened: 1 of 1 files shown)


davidcassany:
@schaefi this fixes the dracut-kiwi-oem-repart module in Fedora Rawhide. I am not very pleased with the solution as I could not find any obvious combination of flags according to lsblk man pages to ensure the output is sorted as we want.

I did not want to dive into different approaches though. The `partid` parameter of `get_partition_node_name` is set in the python code at build time so following a different approach to identify root partition is easily leading to some sensitive refactor I wanted to avoid.

I have changes only a this call of `lsblk` despite the `--raw` flag is widely used in other places, however as I could see the call lsblk call here is the only one where order matters.

schaefi:
Thanks this looks good to me. On suse using -l or -r in the way we do it here doesn't change the output in any way.
Pretty interesting that this behavior is different on other distros... and concerning. According to the man page the change here using ```--list``` is correct as this is also the only place I could find were lsblk explicitly states the output to be sorted

Ok let's go for it. Thanks much

schaefi (approved):
:+1:

PR #1673, opened 2020-12-27
force ascii encoding on checksum files

Josua-SR:
Python open function can fail when a file is not found encoded in
utf-8, depending on the execution environment. In particular on
Debian 10 an error message regarding utf8 was encountered.

Since checksum files always use only ascii characters, force it.

### DIFF
--- a/kiwi/utils/checksum.py
+++ b/kiwi/utils/checksum.py
@@ -18,6 +18,7 @@
 import os
 from collections import namedtuple
 import hashlib
+import encodings.ascii
 
 # project
 from kiwi.command import Command
@@ -58,7 +59,7 @@ class Checksum:
         """
         if not os.path.exists(filename):
             return False
-        with open(filename) as checksum_file:
+        with open(filename, encoding=encodings.ascii) as checksum_file:
             checksum_from_file = checksum_file.read()
             # checksum is expected to be stored in the first field
             # separated by space, other information might contain


schaefi:
I'm fine forcing ascii on the checksum file since I agree there will be no other data in it. However I'm not sure if this is a generic fix for systems not using an utf-8 encoding.

@davidcassany would be interested in your feedback as well. From my perspective this one is good to be merged

Thoughts ?

schaefi (approved):

davidcassany (approved):
Looks good to me, there aren't non ascii characters in the checksum so I agree this is a safe code and if it prevents some confusions under certain environments lets go for it.

davidcassany:
Thinking it twice probably we could consider having the same criteria in 
https://github.com/OSInside/kiwi/blob/36c90c3fc01305eb96d97805580d790ee3e230db/kiwi/utils/checksum.py#L129 

this way we are consistent with the checksum file we write and read. This should mitigate issues in systems that do not make use of utf-8 enconding.

Josua-SR:
> Thinking it twice probably we could consider having the same criteria in
> 
> https://github.com/OSInside/kiwi/blob/36c90c3fc01305eb96d97805580d790ee3e230db/kiwi/utils/checksum.py#L129
> 
> this way we are consistent with the checksum file we write and read. This should mitigate issues in systems that do not make use of utf-8 enconding.

makes sense to me, I only caught the case that triggered during my particular use.

schaefi:
yep opening for writing with ascii encoding is the logical result of this :) @Josua-SR will you add the change in addition to what you already did ? Thanks much

schaefi:
I'd like to continue with this one and will merge it now

@davidcassany I'll create a follow up PR to address what you suggested

PR #1679, opened 2021-01-11
Delete obsolete ddb.adapterType patching

schaefi:
When building a vmdk image with pvscsi as adapter type, kiwi
implicitly changed the adapter_type from pvscsi to lsilogic
because the pvscsi driver was missing. The real solution to
this problem is to include the now existing vmw_pvscsi kernel
driver into the initrd with a custom root overlay file
etc/dracut.conf.d/pvscsi.conf. This Fixes bsc#1180539

### DIFF
--- a/kiwi/storage/subformat/base.py
+++ b/kiwi/storage/subformat/base.py
@@ -131,15 +131,6 @@ class DiskFormatBase:
         if custom_args:
             ordered_args = OrderedDict(sorted(custom_args.items()))
             for key, value in list(ordered_args.items()):
-                if key == 'adapter_type=pvscsi':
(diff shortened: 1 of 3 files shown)


schaefi:
The code can be reviewed but the merge needs to wait for confirmation of the driver inclusion suggestion in

* https://bugzilla.suse.com/show_bug.cgi?id=1180539

schaefi:
The feedback from bugzilla was not promising. Without the hack ripped out in this PR, qemu cannot create the requested disk format because qemu does not support the pvscsi adapter type

schaefi:
this will not be merged. See details in the bug report associated with the issue here

PR #1685, opened 2021-01-13
Refactor default shared cache location

### LINKED ISSUE #1671: kiwi/defaults.py depends on the kiwi client command line parameters
## Problem description

The `kiwi/defaults.py` class relies on command line parameters for some methods. More specific this is the case for the `get_shared_cache_location` method which depends on `--shared-cache-dir` parameter. This is a bad design as defaults class should be usable without any kiwi client being present. This is causing problems from an API perspective since it causes weird issues when using core classes such as `SystemPrepare`. In fact the following snipped throwing the usage dialog on stdout 

```python
#!/usr/bin/env python3
from kiwi.system.prepare import SystemPrepare
from kiwi.xml_description import XMLDescription
from kiwi.xml_state import XMLState

description = XMLDescription('sr-imx8-debian-10/config.xml').load()
state = XMLState(description)
system = SystemPrepare(xml_state=state, root_dir='/tmp/build', allow_existing=True)
print('done')
```

With the given code the default `shared-cache-dir` should just be applied without running any check against the kiwi-client command line parameters.

+ KIWI version: master
+ Operating system: 
+ OBS version:

jesusbv:
Default shared cache location does not depend on CLI parameters
Kiwi CLI default value for shared cache directory depends on Defaults.

Fixes #1671

### DIFF
--- a/kiwi/cli.py
+++ b/kiwi/cli.py
@@ -60,7 +60,7 @@ global options for services: image, system
         specify an alternative shared cache directory. The directory
         is shared via bind mount between the build host and image
         root system and contains information about package repositories
-        and their cache and meta data. [default: /var/cache/kiwi]
+        and their cache and meta data.
     --type=<build_type>
         image build type. If not set the default XML specified
         build type will be used
@@ -81,6 +81,7 @@ from kiwi.exceptions import (
 from kiwi.path import Path
 from kiwi.version import __version__
 from kiwi.help import Help
+from kiwi.defaults import Defaults
 
 log = logging.getLogger('kiwi')
 
@@ -206,6 +207,10 @@ class Cli:
                         'vmx type is now a subset of oem, --type set to oem'
                     )
                     value = 'oem'
+                if arg == '--shared-cache-dir' and not value:
+                    value = os.sep + Defaults.get_shared_cache_location()
+                if arg == '--shared-cache-dir' and value:
+                    Defaults.set_shared_cache_location(value)
                 result[arg] = value
         return result
 
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -38,6 +38,7 @@ EDIT_BOOT_CONFIG_SCRIPT = 'edit_boot_config.sh'
 EDIT_BOOT_INSTALL_SCRIPT = 'edit_boot_install.sh'
 IMAGE_METADATA_DIR = 'image'
 ROOT_VOLUME_NAME = 'LVRoot'
+SHARED_CACHE_DIR = '/var/cache/kiwi'
 
 
 class Defaults:
@@ -182,6 +183,16 @@ class Defaults:
         return '/var/tmp/kiwi/satsolver'
 
     @staticmethod
+    def set_shared_cache_location(location):
+        """
+        Sets the shared cache location once
+
+        :param str location: a location path
+        """
+        global SHARED_CACHE_DIR
+        SHARED_CACHE_DIR = location
+
+    @staticmethod
     def get_shared_cache_location():
         """
         Provides the shared cache location
@@ -196,9 +207,8 @@ class Defaults:
 
         :rtype: str
         """
-        from .cli import Cli
         return os.path.abspath(os.path.normpath(
-            Cli().get_global_args().get('--shared-cache-dir')
+            SHARED_CACHE_DIR
         )).lstrip(os.sep)
 
     @staticmethod
--- a/test/unit/defaults_test.py
+++ b/test/unit/defaults_test.py
@@ -44,15 +44,6 @@ class TestDefaults:
     def test_get_default_shared_cache_location(self):
         assert Defaults.get_shared_cache_location() == 'var/cache/kiwi'
 
-    def test_get_custom_shared_cache_location(self):
-        sys.argv = [
-            sys.argv[0],
-            '--shared-cache-dir', '/my/cachedir',
-            'system', 'prepare',
-            '--description', 'description', '--root', 'directory'
-        ]
-        assert Defaults.get_shared_cache_location() == 'my/cachedir'
-
     @patch('kiwi.defaults.Path.which')
     def test_get_grub_boot_directory_name(self, mock_which):
         mock_which.return_value = 'grub2-install-was-found'


kiwi/cli.py
```diff
@@ -206,6 +207,8 @@ def get_global_args(self):
                         'vmx type is now a subset of oem, --type set to oem'
                     )
                     value = 'oem'
+                if arg == '--shared-cache-dir' and not value:
+                    value = os.sep + Defaults.get_shared_cache_location()
```
schaefi, line 211: I wonder why ```os.sep + ...``` is needed here ? Do you had a particular reason why defaults.py can't simply provide ```/var/cache/kiwi``` with the leading "/" ? Actually the former default setting from docopt did that too

schaefi (changes requested):
Looks great, just one question to clarify

kiwi/cli.py
jesusbv: The former default in cli returned `/var/cache/kiwi` and in `get_shared_cache_location` the leading '/' was [stripped off](https://github.com/OSInside/kiwi/blob/5a0d13f233eafce325c307c5cc4a3cbec64268ce/kiwi/defaults.py#L185), returning `var/cache/kiwi`

Given the comment, I thought to preserve that behavior would be a good idea unless the reasons for stripping off the lead '/' are no longer valid or have changed.

davidcassany (changes requested):
I have doubts this is sufficient.

Note that `Defaults.get_cache_shared_location()` is being used in several places:

* boot/image/builtin_kiwi.py
* builder/disk.py
* system/root_bind.py
* system/root_init.py
* runtime_checker.py

With this implementation it looks like on these calls the result will always be `var/cache/kiwi` regardless of the client flag value, isn't it?

I believe this task is a bit more complex. To me it looks like we do not have any structure in KIWI to keep the state of global arguments and expose them to the API. Probably the closest thing we have is the xml_state class.

At a glance the only way I see to solve it without modifying the API is by moving the client dependency we used to have in defaults to the above mentioned classes, this is something I really dislike. 

We could embed the cache location value in XMLState class, however I don't think it belongs there and root_init and root_bind classes do not have access to the XMLState instance, this would require an API change for those classes.

Alternatively we could add this cache location as an optional init parameter of the above classes that defaults to the current default unless some location is given. This is like to imply changes to carry over the cache location value in further classes than the list above.

I guess this is why we used to take the short cut of having non stateless defaults :sweat_smile: 

I don't see any elegant solution that does not require some refactoring.

Do you have any thoughts or ideas?

jesusbv:
> With this implementation it looks like on these calls the result will always be `var/cache/kiwi` regardless of the client flag value, isn't it?
No, this [line](https://github.com/OSInside/kiwi/blob/4661d996931aa32c92d19857878da0ac2acb15c2/kiwi/cli.py#L211) makes sure there is a *default* value but if client flag has any other value, e.g. `/another/path/from/flag` that value will be used.

So, default shared cache location is exactly that, a default, if that location requires a different path, it is enough to fill `--shared-cache-dir` flag in CLI.

jesusbv:
> I have doubts this is sufficient.
> 
> Note that `Defaults.get_cache_shared_location()` is being used in several places:
> 
>     * boot/image/builtin_kiwi.py
> 
>     * builder/disk.py
> 
>     * system/root_bind.py
> 
>     * system/root_init.py
> 
>     * runtime_checker.py
> 
> 
> With this implementation it looks like on these calls the result will always be `var/cache/kiwi` regardless of the client flag value, isn't it?
> 

> With this implementation it looks like on these calls the result will always be `var/cache/kiwi` regardless of the client flag value, isn't it?

Yes, this [line](https://github.com/OSInside/kiwi/blob/4661d996931aa32c92d19857878da0ac2acb15c2/kiwi/cli.py#L211) makes sure there is a **default** value but if client flag has any other value, e.g. `/another/path/from/flag` that value will be used within the CLI context, default will be default.

> I don't see any elegant solution that does not require some refactoring.
> 
> Do you have any thoughts or ideas?

The problem of using `Defaults.get_shared_cache_location()` is that it will return a default value, from what I understand (correct me if wrong), we want to give the option of using a different shared cache location in the list of places you mentioned which said new/different cache location value could be populated in the CLI flag but it should work without the CLI.

If that is the case, what I think is a new `Defaults.custom_shared_cache_location(self, path)` + `Defaults.get_custom_shared_cache_location()`
is needed (or a way to set custom location) and then when using `Defaults.get_shared_cache_location()` check if custom has any value and use that otherwise, use `var/cache/kiwi`

schaefi:
> With this implementation it looks like on these calls the result will always be var/cache/kiwi regardless of the client flag value, isn't it?

Hmm, right I haven't thought about the behavior change that this patch involves. I still think the changes Jesus did are the right approach. We want the default method to return a static information.

On the CLI side however more code is needed. I think every code point that uses ```Defaults.get_cache_shared_location()``` needs to check if the setting was changed by the CLI and prefers it when set.

An alternative solution from my side would be the following:

1. In kiwi/defaults.py set
    ```
    SHARED_CACHE_DIR = '/var/cache/kiwi'
    ```

2. In kiwi/defaults.py delete the method get_shared_cache_location()

3. In kiwi/defaults.py create a new method
    ```
    def set_shared_cache_location(location):
        SHARED_CACHE_DIR=location;
    ```

4. In the kiwi/cli.py make use of set_shared_cache_location()

5. In all places of kiwi that used the former get_shared_cache_location replace the code with
    ```
    import kiwi.defaults as defaults

    defaults.SHARED_CACHE_DIR
    ```

Thoughts ? @davidcassany @jesusbv

schaefi:
I know this is an incompatible change when we delete get_shared_cache_location(), so we can (and maybe should keep it)
it will then change to

```python
def get_shared_cache_location():
    return SHARED_CACHE_DIR
```

jesusbv:
> I know this is an incompatible change when we delete get_shared_cache_location(), so we can (and maybe should keep it)
> it will then change to
> 
> ```python
> def get_shared_cache_location():
>     return SHARED_CACHE_DIR
> ```

Yes, I agree and planned to update it like that for the next change.

jesusbv:
> On the CLI side however more code is needed. I think every code point that uses `Defaults.get_cache_shared_location()` needs to check if the setting was changed by the CLI and prefers it when set.
> 

I dont think this is a good solution as still depends on CLI

> An alternative solution from my side would be the following:
> 
>     1. In kiwi/defaults.py set
>        ```
>        SHARED_CACHE_DIR = '/var/cache/kiwi'
>        ```
> 
>     2. In kiwi/defaults.py delete the method get_shared_cache_location()
> 
>     3. In kiwi/defaults.py create a new method
>        ```
>        def set_shared_cache_location(location):
>            SHARED_CACHE_DIR=location;
>        ```
> 
>     4. In the kiwi/cli.py make use of set_shared_cache_location()
> 
>     5. In all places of kiwi that used the former get_shared_cache_location replace the code with
>        ```
>        import kiwi.defaults as defaults
>        
>        defaults.SHARED_CACHE_DIR
>        ```
> 
> 
> Thoughts ? @davidcassany @jesusbv

I like this one, as we thought [similarly](https://github.com/OSInside/kiwi/pull/1685#issuecomment-760125664)

I think we need to make sure the custom location is set only once

davidcassany:
> If that is the case, what I think is a new `Defaults.set_custom_shared_cache_location(self, path)` + `Defaults.get_custom_shared_cache_location()`
> is needed (or a way to set custom location) and then when using `Defaults.get_shared_cache_location()` check
> 
> if CLI has any value => use CLI value
> else if custom has any value => use that
> otherwise, use `var/cache/kiwi`
> 
> This way it would work without forcing to have CLI at all and option to have a different cache location and default means that,a default value.
> 
> thoughts on that approach ?

I mostly agree on the functionality, yes. We need a default and we need a way to set a custom cache location. IMHO, the ideal scenario is that the client just sets the custom location the same way you would do it through the API according to the given flags.

> An alternative solution from my side would be the following:
> 
>     1. In kiwi/defaults.py set
>        ```
>        SHARED_CACHE_DIR = '/var/cache/kiwi'
>        ```
> 
>     2. In kiwi/defaults.py delete the method get_shared_cache_location()
> 
>     3. In kiwi/defaults.py create a new method
>        ```
>        def set_shared_cache_location(location):
>            SHARED_CACHE_DIR=location;
>        ```
> 
>     4. In the kiwi/cli.py make use of set_shared_cache_location()
> 
>     5. In all places of kiwi that used the former get_shared_cache_location replace the code with
>        ```
>        import kiwi.defaults as defaults
>        
>        defaults.SHARED_CACHE_DIR
>        ```

Yes, this seams to be reasonable approach if we keep the original method as later suggested. To me it is weird to change a default constant value at runtime, I did not even consider that :P but thinking about it I can't see any problem with it now.

With this approach the snipped from #1671 should be capable to set the custom cache location by adding

```python
Defaults.set_cache_location('mycustompath')
```

on the very beginning, looks good to me.

jesusbv:
>        SHARED_CACHE_DIR = '/var/cache/kiwi'
>        ```
Regarding the leading '/' in the location, any agreement on [that](https://github.com/OSInside/kiwi/pull/1685#discussion_r556675246) ?

kiwi/defaults.py
```diff
@@ -181,6 +182,17 @@ def get_solvable_location():
...
+    @staticmethod
+    def set_shared_cache_location(location):
+        """
+        Sets the shared cache location once
+
+        :param str location: a location path
+        """
+        global SHARED_CACHE_DIR
+        if SHARED_CACHE_DIR == 'var/cache/kiwi':
+            SHARED_CACHE_DIR = location
```
jesusbv, line 194: Thoughts about setting the cache location once ?

schaefi:
> > ```
> >    SHARED_CACHE_DIR = '/var/cache/kiwi'
> >    ```
> > ```
> 
> Regarding the leading '/' in the location, any agreement on [that](https://github.com/OSInside/kiwi/pull/1685#discussion_r556675246) ?

I vote for the default should be set with the leading "/". 

```
SHARED_CACHE_DIR = '/var/cache/kiwi'
```

The method get_shared_cache_location should have some code to remove the leading /. The method currently
has this comment block and that should stay valid
```
        Provides the shared cache location

        This is a directory which shares data from the image buildsystem
        host with the image root system. The location is returned as an
        absolute path stripped off by the leading '/'. This is because
        the path is transparently used on the host /<cache-dir> and
        inside of the image imageroot/<cache-dir>
```

kiwi/defaults.py
schaefi: did not understand the question ?

davidcassany: I have no strong opinion to be honest. Probably I would have let the user just set the custom cache at any time as many times as he or she wants. I can't see why someone would like to change the location more than once and I agree that this is most likely a bad practice, but just to keep the code simple and flexible. I'd assume that someone using the API and setting custom cache location in some strange way is likely to know she or he wants.

Anyway, as I said, no strong opinion.

schaefi:
@davidcassany 

>To me it is weird to change a default constant value at runtime, I did not even consider that :P but thinking about it I can't see any problem with it now.

It might be questionable but the Defaults class is not stateless. There is other code that returns different value e.g depending on the distribution. Think about the grub path, for one it's /usr/lib/grub2 for others it's /usr/lib/grub. Given that the nature of defaults.py is NOT const it's ok to have a setter method for a setting that is considered configurable. If you run kiwi with the standard default for the shared cache dir changed it will effectively become the default for all actions of this build. Furthermore it must stay as the default if you want to repeat the build and benefit from the setting. So I see the default space as stateful information that depends on the build environment and parameters. From an API perspective this seems also clean to me as a user has an opportunity to set its own default prior any further action.

What might be misleading is the name Defaults, it's more of a BuildConfig, but we are not going to change that :)

kiwi/boot/image/builtin_kiwi.py
```diff
@@ -165,7 +166,7 @@ def create_initrd(self, mbrid=None, basename=None, install_initrd=False):
             # during the process of creating an image but not when the
             # image is actually booting with this initrd
             exclude_from_archive = [
-                '/' + Defaults.get_shared_cache_location(),
+                '/' + defaults.SHARED_CACHE_DIR,
```
davidcassany, line 169: I would keep the get_shared_cache_location so these files remain exactly the same and the change is transparent.

schaefi: yes, all of these calls to get_shared_cache_location should stay

schaefi: @jesusbv sorry if I confused you with my first suggestion how to access the default variable. For compatibility and transparency the method get_shared_cache_location should stay, such that all code parts that calls it also stays untouched. The method itself changes and needs to do the following:

* strip an eventual leading / from the SHARED_CACHE_DIR variable
* return that value

kiwi/defaults.py
jesusbv: are there any problems/pitfalls if cache location is set multiple times in different places ?
My guess is that there isn't any major risks because it is the cache, please let me know otherwise.

davidcassany (approved):
Looks good to me :+1:

schaefi (approved):
perfect :+1:

schaefi:
I set the don't merge label just because we will submit to Factory tomorrow and don't want master to change now

PR #1688, opened 2021-01-15
Set dracut modules list for install initrds

### LINKED ISSUE #1676: oem-unattended (microos) pxe images won't boot (missing sysroot.mount)
## Problem description
oem-unattended (microos) pxe images will fail to boot in the initial initrd boot sequence with:
```
[    4.458063] systemd[1]: Failed to isolate default target: Unit sysroot.mount not found.
[!!!!!!] Failed to isolate default target.
[    4.462835] systemd[1]: Freezing execution.
```
This problem occured after merging https://github.com/OSInside/kiwi/pull/1633

We only tested this issue against microos, but probably also other distros are affected here, due to its supposed problem in the kiwi-dump module.

## Expected behaviour

Device should boot up and start downloading/writing the new system image to the harddisk.

Build a microos pxe image with kiwi or obs.
```
<type image="oem" initrd_system="dracut" installpxe="true" [...]>
[...]
```
Setup a pxe server and boot a vm or real device with the built image.

The initrd parameter are kept to the minimum:
```
[/srv/tftpboot/pxelinux.cfg/default]
[...]
label linux
  kernel boot/linux
  append initrd=boot/initrd rd.kiwi.install.pxe rd.kiwi.install.image=http://192.168.100.1/images/MicroOS.x86_64-20210103.xz
```
## Steps to workaround the behaviour

If we readd the removed [systemd generator](https://github.com/OSInside/kiwi/blob/c71e0ce944318df9df895de63e7bfaf24251f471/dracut/modules.d/90kiwi-dump/kiwi-ramdisk-deployment-generator.sh) to the initrd file, everything is back to normal.

```
pxe:/srv/tftpboot/boot # mkdir {pre-ramdisk-patch,post-ramdisk-patch,tmp}
pxe:/srv/tftpboot/boot/pre-ramdisk-patch # /usr/lib/dracut/skipcpio ../pxeboot.pre-ramdisk-patch.initrd.xz | unxz | cpio -i
pxe:/srv/tftpboot/boot/post-ramdisk-patch # /usr/lib/dracut/skipcpio ../pxeboot.post-ramdisk-patch.initrd.xz | unxz | cpio -i
pxe:/srv/tftpboot/boot/tmp # cat ../pxeboot.post-ramdisk-patch.initrd.xz | cpio -idmv
[...]
7134 blocks
pxe:/srv/tftpboot/boot # dd if=pxeboot.post-ramdisk-patch.initrd.xz bs=512 count=7134 of=early
pxe:/srv/tftpboot/boot/post-ramdisk-patch # cp ../pre-ramdisk-patch/usr/lib/systemd/system-generators/dracut-kiwi-ramdisk-generator usr/lib/systemd/system-generators/
pxe:/srv/tftpboot/boot/post-ramdisk-patch # find . | cpio --create --format='newc' > ../rootfs
pxe:/srv/tftpboot/boot # gzip rootfs
pxe:/srv/tftpboot/boot # cat early rootfs.gz > initrd
```

So far we did not dive deeper into the issue to fix the problem yet, but we wanted indicate this topic, which might help others trying to boot / install their systems using pxe.

+ KIWI version: 9.22.1-1.1
+ Operating system: openSUSE Tumbleweed / MicroOS

### LINKED ISSUE #1683: Use only required dracut modules for the installation initrd
## Problem description

The installation initrd for kiwi is expected to only run with the ```dracut-kiwi-oem-dump``` provided code. The dracut call to create this initrd runs inside of the root system and would include other dracut modules that are present or for which a config is provided. This however can lead to bad side effects. There is for example the MicroOS image which installs additional dracut modules that does something in an initrd. We don't want them to be part of the kiwi initrd that should only call the dumping
code.

We already omit other kiwi modules to be installed in the kiwi initrd but we have no handle for other "unknown" dracut modules that are just imported but unwanted.

I don't know if a good solution can be found to this problem

davidcassany:
This commit makes the installation initrds to include only the
dracut modules that are required by kiwi installation process.

Fixes #1676
Fixes #1683

### DIFF
--- a/doc/source/building_images/build_expandable_disk.rst
+++ b/doc/source/building_images/build_expandable_disk.rst
@@ -389,3 +389,28 @@ oemconfig.oem-unattended Element
   automatically without requiring user interaction. If multiple
   possible target devices are discovered the image is deployed to
   the first device. ``kiwi_oemunattended`` in the initrd
+
+.. _installmedia_customize:
+
+Installation Media Customization
+--------------------------------
+
+The installation media created for OEM network or CD/DVD deployments can
+be customized with the `installmedia` section which is a child section of the `type`
+element as it appears in the following example:
+
+.. code:: xml
+
+   <installmedia>
+     <initrd action="omit">
+       <dracut module="network-legacy"/>
+     </initrd>
+   </installmedia>
+
+The `installmedia` is only available for OEM image types that includes the
+request to create an installation media.
+
+The `initrd` child element of `installmedia` lists dracut modules, they
+can be omitted, added or staticaly set the list of included ones. This is
+specified with the `action` attribute and can take `action="omit"`,
+`action="add"` or `action="set"` values. 
--- a/kiwi/boot/image/dracut.py
+++ b/kiwi/boot/image/dracut.py
@@ -51,6 +51,8 @@ class BootImageDracut(BootImageBase):
         self.included_files_install = []
         self.modules = []
         self.install_modules = []
+        self.add_modules = []
+        self.add_install_modules = []
         self.omit_modules = []
         self.omit_install_modules = []
         self.available_modules = self._get_modules()
@@ -76,10 +78,10 @@ class BootImageDracut(BootImageBase):
         """
         warn_msg = 'module "{0}" not included in initrd'.format(module)
         if self._module_available(module):
-            if install_media and module not in self.install_modules:
-                self.install_modules.append(module)
-            elif module not in self.modules:
-                self.modules.append(module)
+            if install_media and module not in self.add_install_modules:
+                self.add_install_modules.append(module)
+            elif module not in self.add_modules:
+                self.add_modules.append(module)
         else:
             log.warning(warn_msg)
 
@@ -94,6 +96,18 @@ class BootImageDracut(BootImageBase):
             self.omit_install_modules.append(module)
         elif module not in self.omit_modules:
             self.omit_modules.append(module)
+
+    def set_static_modules(self, modules, install_media=False):
+        """
+        Set static dracut modules list for boot image
+
+        :param list modules: list of the modules to include
+        :param bool install_media: lists the modules for install initrds
+        """
+        if install_media:
+            self.install_modules = modules
+        else:
+            self.modules = modules
 
     def write_system_config_file(self, config, config_file=None):
         """
@@ -166,9 +180,12 @@ class BootImageDracut(BootImageBase):
             if install_initrd:
                 included_files = self.included_files_install
                 modules_args = [
-                    '--add', ' {0} '.format(' '.join(self.install_modules))
+                    '--modules', ' {0} '.format(' '.join(self.install_modules))
                 ] if self.install_modules else []
-                omit_modules_args = [
+                modules_args += [
+                    '--add', ' {0} '.format(' '.join(self.add_install_modules))
+                ] if self.add_install_modules else []
+                modules_args += [
                     '--omit', ' {0} '.format(
                         ' '.join(self.omit_install_modules)
                     )
@@ -176,14 +193,16 @@ class BootImageDracut(BootImageBase):
             else:
                 included_files = self.included_files
                 modules_args = [
-                    '--add', ' {0} '.format(' '.join(self.modules))
+                    '--modules', ' {0} '.format(' '.join(self.modules))
                 ] if self.modules else []
-                omit_modules_args = [
+                modules_args += [
+                    '--add', ' {0} '.format(' '.join(self.add_modules))
+                ] if self.add_modules else []
+                modules_args += [
                     '--omit', ' {0} '.format(' '.join(self.omit_modules))
                 ] if self.omit_modules else []
             dracut_initrd_basename += '.xz'
-            options = self.dracut_options + modules_args +\
-                omit_modules_args + included_files
+            options = self.dracut_options + modules_args + included_files
             dracut_call = Command.run(
                 [
                     'chroot', self.boot_root_directory,
--- a/kiwi/builder/install.py
+++ b/kiwi/builder/install.py
@@ -388,6 +388,14 @@ class InstallImageBuilder:
                 self.boot_image_task.omit_module(
                     'multipath', install_media=True
                 )
+            for module in self.xml_state.get_installinitrd_modules('add'):
+                self.boot_image_task.include_module(module, install_media=True)
+            for module in self.xml_state.get_installinitrd_modules('omit'):
+                self.boot_image_task.omit_module(module, install_media=True)
+            self.boot_image_task.set_static_modules(
+                self.xml_state.get_installinitrd_modules('set'),
+                install_media=True
+            )
         self.boot_image_task.create_initrd(
             self.mbrid, 'initrd_kiwi_install',
             install_initrd=True
@@ -430,6 +438,14 @@ class InstallImageBuilder:
                 self.boot_image_task.omit_module(
                     'multipath', install_media=True
                 )
+            for module in self.xml_state.get_installinitrd_modules('add'):
+                self.boot_image_task.include_module(module, install_media=True)
+            for module in self.xml_state.get_installinitrd_modules('omit'):
+                self.boot_image_task.omit_module(module, install_media=True)
+            self.boot_image_task.set_static_modules(
+                self.xml_state.get_installinitrd_modules('set'),
+                install_media=True
+            )
             self._add_system_image_boot_options_to_boot_image()
         self.boot_image_task.create_initrd(
             self.mbrid, 'initrd_kiwi_install',
--- a/kiwi/builder/install.py
+++ b/kiwi/builder/install.py
@@ -388,12 +388,12 @@ class InstallImageBuilder:
                 self.boot_image_task.omit_module(
                     'multipath', install_media=True
                 )
-            for module in self.xml_state.get_installinitrd_modules('add'):
-                self.boot_image_task.include_module(module, install_media=True)
-            for module in self.xml_state.get_installinitrd_modules('omit'):
-                self.boot_image_task.omit_module(module, install_media=True)
+            for mod in self.xml_state.get_installmedia_initrd_modules('add'):
+                self.boot_image_task.include_module(mod, install_media=True)
+            for mod in self.xml_state.get_installmedia_initrd_modules('omit'):
+                self.boot_image_task.omit_module(mod, install_media=True)
             self.boot_image_task.set_static_modules(
-                self.xml_state.get_installinitrd_modules('set'),
+                self.xml_state.get_installmedia_initrd_modules('set'),
                 install_media=True
             )
         self.boot_image_task.create_initrd(
@@ -438,12 +438,12 @@ class InstallImageBuilder:
                 self.boot_image_task.omit_module(
                     'multipath', install_media=True
                 )
-            for module in self.xml_state.get_installinitrd_modules('add'):
-                self.boot_image_task.include_module(module, install_media=True)
-            for module in self.xml_state.get_installinitrd_modules('omit'):
-                self.boot_image_task.omit_module(module, install_media=True)
+            for mod in self.xml_state.get_installmedia_initrd_modules('add'):
+                self.boot_image_task.include_module(mod, install_media=True)
+            for mod in self.xml_state.get_installmedia_initrd_modules('omit'):
+                self.boot_image_task.omit_module(mod, install_media=True)
             self.boot_image_task.set_static_modules(
-                self.xml_state.get_installinitrd_modules('set'),
+                self.xml_state.get_installmedia_initrd_modules('set'),
                 install_media=True
             )
             self._add_system_image_boot_options_to_boot_image()
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -1012,7 +1012,7 @@ class XMLState:
             return container_config_sections[0]
         return None
 
-    def get_installinitrd_modules(self, action: str) -> List[str]:
+    def get_installmedia_initrd_modules(self, action: str) -> List[str]:
         """
         Gets the list of modules to append in installation initrds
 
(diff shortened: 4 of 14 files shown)


davidcassany:
I added the **Don't merge** label as I am not fully satisfied with the implementation and I want to discuss the pros and cons.

This PR is an implementation of the idea from #1683 to limit the dracut modules that are being used within the installation initrd. The implementation as seen in this PR is pretty simple, however it has few down sides from my point of view:

* As a consequence we have to maintain a list of dracut modules, which is not obvious and it might differ between distros and distro versions. It is likely to become tricky in the long run. I was hoping a list as short as `' base bash systemd '` plus the kiwi modules would be enough to pull all relevant dependencies, but this is not the case.

* This code also affects the installation images. In KIWI, as is in this PR, we do not distinguish between installation initrds for ISOs and installation initrds for PXE. We could probably consider making this distinction, not sure if it would be worth thought.

* The installation initrd is no longer as customizable as it used to be. This implies that users doing customization for the deployments of oem images via PXE will struggle with it.

One idea that came to my mind is that in order to have full flexibility for customizations what we should encourage to do is treating the installation image and the disk image to install as independent images. So we could eventually build an OEM image and then a separate KIS image to actually deploy and install it. The cool thing is that both descriptions could live in a single XML. This way we could still keep the project in a single place and also clearly divide the installation part from the disk image. Probably we could have an example for that en an entry in `Working with images` section in docs to explain it.

Any thoughts regarding this topics?

schaefi:
Thanks David for kicking this off. I was spinning my head around this solution. My biggest concern is the static list of modules. This will strike back :) Because of that, how about the following:

1. Maintain the static list of modules on the Defaults area. If need be it can be made distro specific at that place but if ever possible I'd like to avoid it

2. Add a commandline line option ```--dracut-installation-image-modules```. If not specified our list is used, if specified the user has a chance to influence the static list

Thoughts ?

> we do not distinguish between installation initrds for ISOs and installation initrds for PXE

This is correct. The code for the installation procedure is exactly the same. But there are two independent
self.boot_image_task.create_initrd() calls in builder/install.py. I think it would be good if we maintain two lists of modules
and if we agree on the option approach also two different options to configure them

>The installation initrd is no longer as customizable as it used to be. This implies that users doing customization for the >deployments of oem images via PXE will struggle with it.

I agree and we should avoid this. The change suggested here is a behavior change and I'm not sure if people will like it. Remember the reason why we are doing this is for a relatively special MicroOS case. Therefore I was also thinking if we
could just add options like ```--install-iso-dracut-modules / --install-kis-dracut-modules``` to implement the limitation of
modules to the given list, but if not specified keep the current behavior. Something like:

a) --install-iso-dracut-modules=static
     Uses the required kiwi module plus a static list we maintain in the defaults space into account

b) --install-iso-dracut-modules=kiwi
     Uses the required kiwi modules and is flexible with whatever dracut modules and config is present at call time 

c) --install-iso-dracut-modules="a b c"
     Uses the required kiwi module plus the given list

with b) as the default

Thoughts ?

schaefi:
The downside of options is that it's very hard to manage this in obs builds

schaefi:
Therefore we can also think to extend the schema with this information

schaefi:
Something like this

```xml
<installiso initrd="static|custom">
    <dracut module="foo"/>
    <dracut module="bar"/>
</installiso>
```

* not specified, run like we do now
* Set as:
  ```xml
  <installiso initrd="static"/>
  ```
  Apply our static list (plus the required kiwi dracut modules)
  Optional given module list is added
* Set as:
  ```xml
  <installiso initrd="custom">
      <dracut module="foo"/>
  </installiso>
  ```
  Apply the given list (plus the required kiwi dracut modules)

schaefi:
Also allow this for ```<installkis>```

davidcassany:
>  My biggest concern is the static list of modules. This will strike back

Yes this is also a concern of mine. I agree maintaining such a list will be problematic in the long run.

> The more I think about it the more I think we should not change the current behavior but allow the suggested adaption

If we can find a decent way of keeing the changes transparent in a way that we keep the current behavior and we also let user to specifically modify the installation initrd in some way I am totally for it.



> Therefore I was also thinking if we
> could just add options like `--install-iso-dracut-modules / --install-kis-dracut-modules` to implement the limitation of
> modules to the given list, but if not specified keep the current behavior. Something like:
> 
>     1. --install-iso-dracut-modules=static
>        Uses the required kiwi module plus a static list we maintain in the defaults space into account
> 
>     2. --install-iso-dracut-modules=kiwi
>        Uses the required kiwi modules and is flexible with whatever dracut modules and config is present at call time
> 
>     3. --install-iso-dracut-modules="a b c"
>        Uses the required kiwi module plus the given list

I am not so sure, wouldn't it be easier to just map the three dracut modules flags of the command line? Something like find a way to set:

* omit_modules: if defined it adds `--omit` with he given list
* modules: if defined it adds `--modules` with the given list (this is the flag used in this PR)
* add_modules: if defined it adds `--add` flag with the given list.

So having something like:

```xml
<installiso initrd="add|omit|static">
    <dracut module="foo"/>
    <dracut module="bar"/>
</installiso
```

This way we do not need to maintain any list of modules anywhere. If some user wants a very static list of modules, just write it down here as a `<installiso initrd="static">`. What you thing? I'd really like to avoid maintaining a modules list.

qby-wenzel:
I generally like this last flexible idea. I'm just unsure about how to keep a note for others regarding the `combustion` case or similar _interfering_ dracut modules, which are installed by the future user or distro. If I get the last idea right, something like
```
<installiso initrd="omit">
    <dracut module="combustion"/>
    <dracut module="other-interfering-module"/>
</installiso>
```
would be crucial to add, which is not obvious in the first case, when the systemd (#1676) error occurs. It might be a rare case, nevertheless should it be at least mentioned in an kiwi example?

davidcassany:
> would be crucial to add, which is not obvious in the first case, when the systemd (#1676) error occurs. It might be a rare case, nevertheless should it be at least mentioned in an kiwi example?

I agree that this is not obvious to solve, however MicroOS is a very specific use case. Anyway, if we implement such a behavior I'd vote to keep an example making use of a static list of modules for the installation iso and kis.

schaefi:
>Anyway, if we implement such a behavior I'd vote to keep an example making use of a static list of modules for the installation iso and kis.

yes docs is a must have in this regard :)

schaefi:
>This way we do not need to maintain any list of modules anywhere. If some user wants a very static list of modules, just write it down here as a <installiso initrd="static">. What you thing? I'd really like to avoid maintaining a modules list.

I like this idea. This sounds good to me :+1:

davidcassany:
right I am going to adapt the code to include this new elements. I'd made them under the `<type>` sections as they depend on `installiso` and `installpxe` attributes. However I have doubts if we should include those under `<oem-config>` or not.

Any opinion regarding that? I'd say not as they do not config the runtime config, they alter the build time config of the oem image and they are not evaluated during first boot.

schaefi:
> right I am going to adapt the code to include this new elements. I'd made them under the `<type>` sections as they depend on `installiso` and `installpxe` attributes. However I have doubts if we should include those under `<oem-config>` or not.
> 
> Any opinion regarding that? I'd say not as they do not config the runtime config, they alter the build time config of the oem image and they are not evaluated during first boot.

I think this belongs under ```<type>``` and only for the disk type as schematron rule

schaefi (commented):
Thanks for this effort :+1: I was looking at it from a logic point of view first. Two things I noted:

1. If we don't specify installinitrd modules in the config it looks like we build the initrd with
    ```--modules _emptylist_``` That imho can't work and would also be a behavior change.
    iirc we wanted to keep the current behavior unless somebody specifies installinitrd
    or am I mistaken ?

2. Documentation is still missing

kiwi/builder/install.py
```diff
@@ -388,6 +388,14 @@ def _create_pxe_install_kernel_and_initrd(self):
                 self.boot_image_task.omit_module(
                     'multipath', install_media=True
                 )
+            for module in self.xml_state.get_installinitrd_modules('add'):
```
davidcassany, line 391: The `self.xml_state.get_installationinitrd_modules` always returns a list (an empty one of no module is defined in XML) so it can be easily typed and the result is always iterable.

kiwi/xml_state.py
```diff
@@ -1012,6 +1012,22 @@ def get_build_type_containerconfig_section(self) -> Any:
...
+
+        :rtype: list
+        """
+        modules = []
+        installinitrd_sections = self.build_type.get_installinitrd()
+        for installinitrd_section in installinitrd_sections:
+            if installinitrd_section.get_action() == action:
+                for module in installinitrd_section.get_dracut():
+                    modules.append(module.get_module())
+        return modules
```
davidcassany, line 1029: Always returns a list. An empty one if no module for the given type (`add|omit|set`) is defined.

davidcassany:
>     1. If we don't specify installinitrd modules in the config it looks like we build the initrd with
>        `--modules _emptylist_` That imho can't work and would also be a behavior change.
>        iirc we wanted to keep the current behavior unless somebody specifies installinitrd
>        or am I mistaken ?

I think this is not the case. At least this is certainly not intended. See the comments I have done in line with the code.



>     2\. Documentation is still missing

Sure, wanted to agree on the functionality and its implementation first. I agree the documentation should be there before merging.

davidcassany:
@schaefi in any case, I understand you are fine with the functionality and the fundamentals of the implementation so I can proceed with some documentation, right?

kiwi/boot/image/dracut.py
```diff
@@ -166,24 +180,29 @@ def create_initrd(self, mbrid=None, basename=None, install_initrd=False):
             if install_initrd:
                 included_files = self.included_files_install
                 modules_args = [
-                    '--add', ' {0} '.format(' '.join(self.install_modules))
+                    '--modules', ' {0} '.format(' '.join(self.install_modules))
                 ] if self.install_modules else []
```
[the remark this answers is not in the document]
schaefi: Ohh now I see it, the ```else []```. ok I overlooked that part of the code. All good

kiwi/boot/image/dracut.py
```diff
@@ -166,24 +180,29 @@ def create_initrd(self, mbrid=None, basename=None, install_initrd=False):
...
                     '--omit', ' {0} '.format(
                         ' '.join(self.omit_install_modules)
                     )
                 ] if self.omit_install_modules else []
             else:
                 included_files = self.included_files
                 modules_args = [
-                    '--add', ' {0} '.format(' '.join(self.modules))
+                    '--modules', ' {0} '.format(' '.join(self.modules))
                 ] if self.modules else []
```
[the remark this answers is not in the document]
schaefi: yes

kiwi/builder/install.py
schaefi: got it

kiwi/xml_state.py
schaefi: got it

kiwi/builder/install.py
```diff
@@ -388,6 +388,14 @@ def _create_pxe_install_kernel_and_initrd(self):
...
+                self.boot_image_task.include_module(module, install_media=True)
+            for module in self.xml_state.get_installinitrd_modules('omit'):
+                self.boot_image_task.omit_module(module, install_media=True)
+            self.boot_image_task.set_static_modules(
+                self.xml_state.get_installinitrd_modules('set'),
```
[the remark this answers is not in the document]
schaefi: got it

schaefi:
> I think this is not the case. At least this is certainly not intended. See the comments I

yes agreed, I overlooked the else condition.

schaefi:
> @schaefi in any case, I understand you are fine with the functionality and the fundamentals of the implementation so I can proceed with some documentation, right?

yes it implements nicely what we talked about. The section name ```installinitrd``` is however something I would change. The reason is that it doesn't tell the user much. From a user perspective you are building an install media when you configure installiso/installstick/installpxe. Therefore I think the section to manage install media should point that out. I was thinking on something like

```xml
<installmedia">
    <initrd action="add">
        <dracut name="foo"/>
    </initrd>
</installmedia>
```

It would give it a bit more scope and clarity when and where this information is being used

Thoughts ?

schaefi:
we also don't know what other surprises will come into play with the install media built by kiwi. Having that in an extra namespace allows us to extend the section

davidcassany:
> The section name `installinitrd` is however something I would change. The reason is that it doesn't tell the user much. From a user perspective you are building an install media when you configure installiso/installstick/installpxe. Therefore I think the section to manage install media should point that out. I was thinking on something like
> 
> ```
> <installmedia>
>     <initrd action="add">
>         <dracut name="foo"/>
>     </initrd>
> </installmedia>
> ```
> 
> It would give it a bit more scope and clarity when and where this information is being used
> 
> Thoughts ?

This is a nice propasal I like it. Yes I think it matches best the purpose. The only downside is that there is an extra level of elements, however, as you noted, this can be a good thing in case we want to add additional installation media parameters in the future (probably a installation ISO embedded in a partition as a recovery system :) ).

davidcassany:
Gonna adapt it and write the documentation

davidcassany:
I am wondering if we should also include an integration test having `<installmedia>`, probably we should update the MicroOS test to align it with the current released image and include installation media on it. @schaefi what you think? If so, I'd prefer doing it in a separate PR, MicroOS is no obviously related with this functionality.

doc/source/building_images/build_expandable_disk.rst
```diff
@@ -389,3 +389,30 @@ oemconfig.oem-unattended Element
...
   possible target devices are discovered the image is deployed to
   the first device. ``kiwi_oemunattended`` in the initrd
+
+.. _installmedia_customize:
+
+Installation Media Customization
+--------------------------------
+
+The installation media created for OEM network or CD/DVD deployments can
+be customized with the `installmedia` which is a child section of the `type`
```
schaefi, line 399: ...customized with the installmedia section which is ...

I think a word is missing

doc/source/building_images/build_expandable_disk.rst
```diff
@@ -389,3 +389,30 @@ oemconfig.oem-unattended Element
...
+.. code:: xml
+
+   <installmedia>
+     <initrd action="omit">
+       <dracut module="network-legacy"/>
+     </initrd>
+   </installmedia>
+
+The `installmedia` is only available for OEM image types including some
+installation media which, in turn, is build when the `type` section includes
```
schaefi, line 411: I would rephrase this like:

is only available for OEM image types that includes the request to create an installation media ...

schaefi (approved):
Nice doc section :+1: just some small comments

schaefi:
@davidcassany  Really great work, thanks. I'm going to merge this now. If the other open PR with the grub problem is ready for a merge I will update the Staging area with a new package. Could you prepare a new PR to update the MicroOS integration test ? It would directly benefit from the new package in Staging. Thanks much

schaefi:
@davidcassany Do you want to keep the 3 commits or do you want to squash them a bit ?

davidcassany:
We can keep the three commits, they are pretty atomic

schaefi:
ok

PR #1690, opened 2021-01-16
Fix disk build with <bootloader name="custom"/>

petrpavlu:
* Fix disk build with `<bootloader name="custom"/>`.
* Move logic to sync system data into a separate method in order to reduce complexity of `DiskBuilder.create_disk()`. This is needed because the fix for `<bootloader name="custom"/>` otherwise causes that `create_disk()` reaches the current `max-complexity=18` limit set for flake8.

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -251,17 +251,18 @@ class DiskBuilder:
         )
 
         # create the bootloader instance
-        self.bootloader_config = BootLoaderConfig.new(
-            self.bootloader, self.xml_state, root_dir=self.root_dir,
-            boot_dir=self.root_dir, custom_args={
-                'targetbase':
-                    self.loop_provider.get_device(),
-                'grub_directory_name':
-                    Defaults.get_grub_boot_directory_name(self.root_dir),
-                'boot_is_crypto':
-                    self.boot_is_crypto
-            }
-        )
+        if self.bootloader != 'custom':
+            self.bootloader_config = BootLoaderConfig.new(
+                self.bootloader, self.xml_state, root_dir=self.root_dir,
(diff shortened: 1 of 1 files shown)


schaefi (approved):
Looks good to me. Thanks for catching this. In fact creating an instance of BootLoaderConfig with the 'custom' bootloader ends in raising an exception. There should be no BootLoaderConfig instance for the custom bootloader setting. Also thanks for the refactoring bits.

:+1:

kiwi/builder/disk.py
```diff
@@ -251,17 +251,19 @@ def create_disk(self):
...
-            boot_dir=self.root_dir, custom_args={
-                'targetbase':
-                    self.loop_provider.get_device(),
-                'grub_directory_name':
-                    Defaults.get_grub_boot_directory_name(self.root_dir),
-                'boot_is_crypto':
-                    self.boot_is_crypto
-            }
-        )
+        self.bootloader_config = None
```
schaefi, line 254: One comment on this. I think it's not needed to have self.bootloader_config initialized as None. All code accessing bootloader_config is only reached if if ```self.bootloader != 'custom``` and in that case we never want to run into the situation that self.bootloader_config is None. I think we rather want the error of the python stack trace in this case.

davidcassany (approved):
Looks good to me, thanks for the fix and refactor.

I agree with Marcus that there is not need to set the bootloader_config to None.

kiwi/builder/disk.py
petrpavlu: Removed in the updated patch.

petrpavlu:
Thank you both for the review. Updated patch removes the extra line with `self.bootloader_config = None`.

schaefi:
Great thanks :)

PR #1692, opened 2021-01-18
Only write GRUB_CMDLINE_LINUX_DEFAULT with content

### LINKED ISSUE #1650: Only write into GRUB_CMDLINE_LINUX_DEFAULT if necessary
Some packages and scripts modify `/etc/default/grub` by adding/modifying options in `GRUB_CMDLINE_LINUX_DEFAULT`.

Even when not specifying `kernelcmdline` in `<type/>`, kiwi overwrites the options in the file:

```
[  122s] [ INFO    ]: 10:51:33 | Writing grub2 defaults file
[  122s] [ INFO    ]: 10:51:33 | --> GRUB_CMDLINE_LINUX_DEFAULT:""
[  122s] [ INFO    ]: 10:51:33 | --> GRUB_DISTRIBUTOR:"openSUSE MicroOS"
```

This means that there has to be the full `kernelcmdline` set for each `<type/>`, which also means that for each profile which needs a different cmdline there has to be a separate `<type/>`, which gets messy quickly.

AFAICT there's no workaround available.

The specific use case here is to enable SELinux in some profiles of `openSUSE-MicroOS` images, which requires adding `security=selinux` to the kernel cmdline.

schaefi:
Only write GRUB_CMDLINE_LINUX_DEFAULT in the grub defaults
file if there are custom options set via the kernelcmdline
attribute. This Fixes #1650

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -616,15 +616,18 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             'GRUB_GFXMODE': self.gfxmode,
             'GRUB_TERMINAL': '"{0}"'.format(self.terminal)
         }
+        grub_final_cmdline = re.sub(
+            r'root=.* |root=.*$', '', self.cmdline
+        ).strip()
         if self.displayname:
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
Looks good

PR #1693, opened 2021-01-18
Mark Micro DNF as a RPM based package manager

Conan-Kudo:
Without doing this, KIWI won't generate the correct output files
for verification of image content.


schaefi (approved):
Looks good

davidcassany (approved):

PR #1694, opened 2021-01-19
Handle checksum files in ascii encoding

schaefi:
Follow up fix for #1673. Handle reading/writing of all
supported checksum variants in ascii encoding


davidcassany (approved):
looks good

PR #1695, opened 2021-01-19
Fixed use of encoding in open calls

schaefi:
The use of encodings.ascii in open calls was wrong. Open expects
an encoding string but encodings.ascii returns a module reference

### DIFF
--- a/kiwi/utils/checksum.py
+++ b/kiwi/utils/checksum.py
@@ -18,7 +18,7 @@
 import os
 from collections import namedtuple
 import hashlib
-import encodings.ascii
+import encodings.ascii as encoding
 
 # project
 from kiwi.command import Command
@@ -43,6 +43,7 @@ class Checksum:
             )
         self.source_filename = source_filename
         self.checksum_filename = None
+        self.ascii = encoding.getregentry().name
 
     def matches(self, checksum, filename):
         """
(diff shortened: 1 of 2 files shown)


schaefi:
I pushed a staging build and realized many builds fails because of

```
[  187s]     with open(filename, encoding=encodings.ascii, mode='rb') as source:
[  187s] TypeError: open() argument 4 must be str or None, not module
[  187s] [ INFO    ]: 13:38:57 | Cleaning up SystemPrepare instance
```

I wonder how it worked for the original patch. Let's do a more careful review now. Thanks

davidcassany:
Damn it.

> I wonder how it worked for the original patch. Let's do a more careful review now. Thanks

I guess mocking was hiding this basic error.

The issue is still not fixed now I get
```
with open(filename, encoding=self.encoding, mode='rb') as source:
ValueError: binary mode doesn't take an encoding argument
```
This is raising at https://github.com/OSInside/kiwi/blob/db3f7f01ebe9ccfab863ac3392594ad1ca38bd3b/kiwi/utils/checksum.py#L156-L159

which makes sense as the file is opened in binary mode. I'd say we should remove the encoding argument for this call.

davidcassany (changes requested):
There are still little issues

schaefi:
yes, we shouldn't have touched this :( I'll delete all encoding stuff from the binary stuff, that was my fault

davidcassany (approved):
I don't see further issues now. Looks good

PR #1697, opened 2021-01-20
Include missing qemu-tools dependency for iso images

davidcassany:
This commit includes a missing qemu-tools dependency for iso
image type.

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -156,6 +156,15 @@ Requires:       xorriso
 %ifarch %{ix86} x86_64
 Requires:       syslinux
 %endif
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
We use qemu-img for ISO builds? Why?

Conan-Kudo (approved):

davidcassany:
> We use qemu-img for ISO builds? Why?

We use `qemu-img` to set the images to apply the embedded filesystem

seife:
> We use qemu-img for ISO builds? Why?

`truncate -s $size $filename` would also work instead of  `quemu-img create $filename $size` and coreutils is most likely installed anyway.

PR #1698, opened 2021-01-20

schaefi:
* kiwi/system/prepare.py
* kiwi/system/profile.py
* kiwi/system/setup.py

The changes here also lead to a small refactoring for the handling
of the package manager. In my opinion it doesn't make sense to allow
a None type package manager from the stateful XML instance. As without
any package manager nothing can be done. As it also turns into an
issue for the PackageManager API which does not allow for an empty
value here I thought it's better to come up with a default package
manager (set to dnf) if no one is explicitly specified

This references Issue #1644

### DIFF
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -17,14 +17,19 @@
 #
 import os
 import logging
+from typing import (
+    List, Any
+)
 from textwrap import dedent
 
 # project
+from kiwi.xml_state import XMLState
 from kiwi.system.root_init import RootInit
 from kiwi.system.root_import import RootImport
 from kiwi.system.root_bind import RootBind
 from kiwi.repository import Repository
 from kiwi.package_manager import PackageManager
+from kiwi.package_manager.base import PackageManagerBase
 from kiwi.command_process import CommandProcess
 from kiwi.system.uri import Uri
 from kiwi.archive.tar import ArchiveTar
@@ -38,7 +43,7 @@ from kiwi.exceptions import (
     KiwiPackagesDeletePhaseFailed
 )
 
-log = logging.getLogger('kiwi')
+log: Any = logging.getLogger('kiwi')
 
 
 class SystemPrepare:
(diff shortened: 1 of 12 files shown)


Conan-Kudo (approved):

kiwi/system/prepare.py
```diff
@@ -94,18 +98,21 @@ def __init__(
         # in order to delay the Uri destructors until the System instance
         # dies. This is needed to keep bind mounted Uri locations alive
         # for System operations
-        self.uri_list = []
+        self.uri_list: List[Uri] = []
 
-    def setup_repositories(self, clear_cache=False, signing_keys=None):
+    def setup_repositories(
+        self, clear_cache: bool = False, signing_keys: List = None
```
davidcassany, line 104: Would it be worth to include the type of the list elements? so instead of `signing_keys: List = None` -> `signing_keys: List[str] = None`. AFAIK this is a valid syntax, isn't it. I am also fine if we agree not to reach that level of detail, but since I saw the  `List[Uri]` few lines above I am wondering about the criteria around this topic.

kiwi/system/root_import/__init__.py
```diff
@@ -47,7 +48,8 @@ class RootImport(metaclass=ABCMeta):
     def __init__(self) -> None:
         return None  # pragma: no cover
 
-    def new(root_dir: str, image_uri: str, image_type: str):
+    @staticmethod
+    def new(root_dir: str, image_uri: Uri, image_type: str):
```
davidcassany, line 52: Good catch! This was wrong and also with an inconsistent type :+1:

davidcassany (approved):
Looks good to me, I am fine with the refactor, makes sense from an API perspective. 

I just added a comment regarding the criteria to follow about typing container objects.

kiwi/system/prepare.py
jesusbv: +1 on this, before noticing prepare was taken, I was following that approach.

kiwi/system/root_import/__init__.py
schaefi: yes you'll find all those glitches with mypy :)

kiwi/system/prepare.py
schaefi: I agree, I was just not sure if this is always a list of str ? can we be sure here ? In the other case I knew it for sure

schaefi: ok this ends in an import of a key which is a file path. Thus List[str] is good. I'll adapt

schaefi: done

PR #1699, opened 2021-01-20

jesusbv:
* kiwi/system/size.py
* kiwi/system/uri.py
* kiwi/system/users.py

Update the test path for users_test.py

This references #1644

### DIFF
--- a/kiwi/system/users.py
+++ b/kiwi/system/users.py
@@ -15,6 +15,9 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+from typing import List
+
+# project
 from kiwi.command import Command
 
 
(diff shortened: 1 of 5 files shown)


kiwi/system/users.py
```diff
@@ -82,7 +87,8 @@ def user_modify(self, user_name, options):
             ['chroot', self.root_dir, 'usermod'] + options + [user_name]
         )
 
-    def setup_home_for_user(self, user_name, group_name, home_path):
+    def setup_home_for_user(self, user_name: str, group_name: str,
+                            home_path: str) -> None:
```
schaefi, line 91: Just nitpick. Can we keep the indentation style to this format

```python
def setup_home_for_user(
    self, user_name: str, group_name: str, home_path: str
) -> None:
```

Thanks

schaefi (approved):
Looks great to me, thanks much :+1:  Just one indentation issue

prior merge, could you squash the commits into one ? We have done this as a one block commit so far and I think it's good if we keep this consistent as well. Thanks

schaefi:
great :)

PR #1700, opened 2021-01-21

jesusbv:
* kiwi/system/root_bind.py
* kiwi/system/root_init.py
* kiwi/system/identifier.py

This references issue #1644


schaefi (approved):
Fantastic :) thanks, also appreciate the update of the test code

PR #1701, opened 2021-01-22

jesusbv:
* kiwi/system/kernel.py
* kiwi/system/result.py
* kiwi/system/shell.py

This references issue #1644

### DIFF
--- a/kiwi/system/kernel.py
+++ b/kiwi/system/kernel.py
@@ -15,14 +15,30 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
-import os
 import re
-from collections import namedtuple
+import os
+from typing import (
+    NamedTuple, Optional
+)
 
 # project
 from kiwi.command import Command
 
 from kiwi.exceptions import KiwiKernelLookupError
+
+kernel_type = NamedTuple(
+    'kernel_type', [
+        ('name', str),
+        ('filename', str),
+        ('version', str)
+    ]
+)
+xen_hypervisor_type = NamedTuple(
+    'xen_hypervisor_type', [
+        ('filename', str),
+        ('name', str)
+    ]
+)
 
 
 class Kernel:
@@ -35,11 +51,13 @@ class Kernel:
         file so that we do not have to search for many different
         names in this code
     """
-    def __init__(self, root_dir):
+    def __init__(self, root_dir: str):
         self.root_dir = root_dir
         self.kernel_names = self._setup_kernel_names_for_lookup()
 
-    def get_kernel(self, raise_on_not_found=False):
+    def get_kernel(
+        self, raise_on_not_found: bool = False
+    ) -> Optional[kernel_type]:
         """
         Lookup kernel files and provide filename and version
 
@@ -50,7 +68,7 @@ class Kernel:
             and kernel is not found
         :return: tuple with filename, kernelname and version
 
-        :rtype: namedtuple
+        :rtype: tuple|None
         """
         for kernel_name in self.kernel_names:
             kernel_file = os.sep.join(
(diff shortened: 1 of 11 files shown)


kiwi/system/kernel.py
```diff
@@ -15,15 +15,23 @@
...
 from kiwi.command import Command
 
 from kiwi.exceptions import KiwiKernelLookupError
 
+kernel_tuple = NamedTuple('kernel', [('name', str),
+                                     ('filename', str),
+                                     ('version', str)])
+xen_hyperv = NamedTuple('xen', [('filename', str),
+                                ('name', str)])
+
```
schaefi, line 34: Regarding the use of a tuple as type there is a hidden policy in kiwi to name them XX_type and use the same indentation style as everywhere else. Therefore can be above be written as

```python
kernel_type = NamedTuple(
    'kernel_type', [
        ('name', str),
        ('filename', str),
        ('version', str)
    ]
)

xen_hypervisor_type = NamedTuple(
    'xen_hypervisor_type', [
        ('filename', str),
        ('name', str)
    ]
)
```

kiwi/system/kernel.py
```diff
@@ -109,14 +114,16 @@ def copy_kernel(self, target_dir, file_name=None):
         """
         kernel = self.get_kernel()
         if kernel:
-            if not file_name:
+            if file_name is None:
```
schaefi, line 117: This is a behavior change. ```not str``` matches empty strings and None. ```str is None``` only matches if the string is of the None type. An empty string ```str=""``` would not be matched. Any particular reason for this change ?

kiwi/system/kernel.py
```diff
@@ -128,7 +135,7 @@ def copy_xen_hypervisor(self, target_dir, file_name=None):
         """
         xen = self.get_xen_hypervisor()
         if xen:
-            if not file_name:
+            if file_name is None:
```
schaefi, line 138: same here, behavior change

kiwi/system/result.py
```diff
@@ -15,45 +15,53 @@
...
 from kiwi.exceptions import (
     KiwiResultError
 )
 
 log = logging.getLogger('kiwi')
 
 # must be global to allow pickle to find it
-result_file_type = namedtuple(
-    'result_file_type', ['filename', 'use_for_bundle', 'compress', 'shasum']
+ResultFileType = NamedTuple(
```
schaefi, line 34: Please keep the naming as ```result_file_type``` and the indentation as suggested in former comments

kiwi/system/result.py
```diff
@@ -15,45 +15,53 @@
...
+    'ResultFileType', [('filename', str),
+                       ('use_for_bundle', bool),
+                       ('compress', bool),
+                       ('shasum', bool)]
 )
+ResultType = TypeVar('ResultType', bound='Result')
```
schaefi, line 40: I did not understand why we need a new global ResultType variable from the Result class. Is a self reference to the Result class not possible when used as return type in ```def load(filename: str) -> ResultType:``` ?

schaefi (changes requested):
Overall looks good, thanks :+1: I was concerned about one behavior change as I think and having one question regarding the ResultType setting

Thoughts ?

kiwi/system/result.py
schaefi: meaning ```def load(filename: str) -> Result:``` does not work ?

kiwi/system/kernel.py
jesusbv: I missed the empty string check but the reason for the change is that when deciding the changes for strong typing I got an
error asking for explicit check in those lines. I will revert it as it is wrong and not explicit check needed anymore

kiwi/system/result.py
jesusbv: I will change it.

The reason for not choosing snake case naming was that I get an info message from mypy
```
 Class name "result_file_type" doesn't conform to PascalCase naming style
 ```

kiwi/system/kernel.py
jesusbv: Done.

schaefi (approved):
Looks great :+1:

PR #1704, opened 2021-01-27
Fixed creation of custom bios grub image

schaefi:
The last commit moved the grub mkimage call into the chroot.
As a side effect and when creating install media the earlyboot
script could no longer be found. This commit fixes it

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -558,8 +558,8 @@ class Defaults:
         lookup_list = []
         for grub_name in ['grub2', 'grub']:
             for install_dir in install_dirs:
-                grub_path = os.sep.join(
-                    [root_path, install_dir, grub_name, filename]
+                grub_path = os.path.join(
+                    root_path, install_dir, grub_name, filename
                 )
                 if os.path.exists(grub_path):
                     return grub_path
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -854,6 +854,11 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         self._create_early_boot_script_for_mbrid_search(
             early_boot_script, mbrid
         )
+        shutil.copy(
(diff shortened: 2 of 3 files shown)


schaefi:
@davidcassany this fixes the currently F:stated integration tests which are not failing because of the obs bugs

kiwi/defaults.py
```diff
@@ -560,7 +560,7 @@ def get_grub_path(root_path, filename, raise_on_error=True):
             for install_dir in install_dirs:
                 grub_path = os.sep.join(
                     [root_path, install_dir, grub_name, filename]
-                )
+                ).replace('//', '/')
```
davidcassany, line 563: Just a matter of taste, but isn't more generic and easier to just use `os.path.normpath` instead of this replacement?

davidcassany (approved):
I have just a small comment on a small specific detail, I leave it up to you. Looks good to me :+1:

kiwi/defaults.py
schaefi: I wanted to do that, but os.path.normpath does not delete double slashes at the beginning :( I think this is a bug there (or intentional). It works fine if the double slashes are in between path elements but not at the beginning. You can give it a try with a simple script. I did the same and was astonished. That's why it ended up with this stupid line of code

schaefi: I'm happy if you know a nicer solution. I found the file.Path class which works nicely but this would create another dependency which I don't want

schaefi: but maybe the above can be changed to use os.path() I'll give it a try

schaefi: @davidcassany  ok better solution, pushed. How about it now ?

davidcassany:
Nice solution. I am also surprised that `os.path.normpath` does not deal with starting `//`... I would not have expected that. Probably there is some insane use case preventing it in some platform :/ This sort of path utilities tend to be come a nightmare once they start supporting multiple platforms.

PR #1705, opened 2021-01-27
Update MicroOS build test

davidcassany:
This commit alignes the MicroOS tests with the MicroOS images build
for kvm and xen. In addition it adds the installation media request and
custom initrd modules configuration for the installation media.

### DIFF
--- a/build-tests/x86/suse/test-image-MicroOS/appliance.kiwi
+++ b/build-tests/x86/suse/test-image-MicroOS/appliance.kiwi
@@ -7,16 +7,12 @@
         <specification>MicroOS disk test build</specification>
     </description>
     <preferences>
-        <version>1.42.1</version>
+        <version>16.0.0</version>
         <packagemanager>zypper</packagemanager>
-        <locale>en_US</locale>
-        <keytable>us</keytable>
-        <timezone>Europe/Berlin</timezone>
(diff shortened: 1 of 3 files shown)


davidcassany:
Quite a relevant change, bigger than what I expected at a glance. I added the `Don't merge` as I'll try to run it before, too many changes to be confident it will succeed from the first run.

schaefi:
* python-kiwi 9.23.9 has landed in Staging which supports the new sections
* I've updated the _service file for the MicroOS integration test to point to this branch

... let's wait until everything has settled and see if it builds

schaefi:
For reference

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:suse/test-image-MicroOS

schaefi (approved):
The update of the description looks good to me :+1: 

I've prepared Staging such that this test can build

davidcassany:
> The update of the description looks good to me +1
> 
> I've prepared Staging such that this test can build

Awesome thanks

davidcassany:
The build now succeeds and I verified I can use the installation iso without issues. The current test actually creates a qcow2 VM that can eventually be deployed using the installation media, in that case the image self expands using the MicroOS tools and concepts and not using KIWI's expandable features.

PR #1706, opened 2021-01-28
Added decorators to help with API management

schaefi:
The lifetime of API methods could be limited due to
the development of kiwi. To allow for a deprecation
process the following helper methods has been added

### DIFF
--- a/kiwi/api_helper.py
+++ b/kiwi/api_helper.py
@@ -1,0 +1,81 @@
+# Copyright (c) 2021 SUSE Software Solutions Germany GmbH. All rights reserved.
+#
+# This file is part of kiwi.
+#
+# kiwi is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# kiwi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kiwi.  If not, see <http://www.gnu.org/licenses/>
+#
+import warnings
+from functools import wraps
+from typing import Callable
+
+
+def obsolete(decommission_at: str, version: str) -> Callable:
+    """
+    Decorator method to mark an API method as obsolete
+
+    Methods marked with this decorator are still called.
+    If the provided decommission date is reached it is
+    allowed to decorate the obsoleted method as
+    decommissioned.
+
+    Example:
+
+        .. code:: python
+
+            @obsolete(decommission_at='2025-01-28', version='1.2.3')
+            def method():
+                print('Method implementation...')
+
+    :param str decommission_at: decommision date string.
+    :param str version: the first version when this function was deprecated.
+
+    :return: decorated function which is marked as deprecated
+    """
+    def decorate_obsolete(func: Callable) -> Callable:
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            warnings.warn(
+                f'Function {func.__name__!r} is marked obsolete '
+                f'since version {version!r} '
+                f'and will be decommissioned at: {decommission_at}'
+            )
+            return func(*args, **kwargs)
+        return wrapper
+    return decorate_obsolete
+
+
+def decommissioned(func: Callable) -> Callable:
+    """
+    Decorator method to mark an API method as decommissioned
+
+    Methods marked with this decorator are no longer called
+    and raises DeprecationWarning when used.
+    The method implementation is allowed to be replaced by
+    a simple pass statement to get rid of old code
+
+    Example:
+
+        .. code:: python
+
+            @decommissioned
+            def method():
+                pass
+
+    """
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        raise DeprecationWarning(
+            f'Function {func.__name__!r} has been decommissioned'
+        )
+    return wrapper
--- a/test/unit/api_helper_test.py
+++ b/test/unit/api_helper_test.py
@@ -1,0 +1,33 @@
+from pytest import raises
+from mock import patch
+from kiwi.api_helper import (
+    obsolete,
+    decommissioned
+)
+
+
+@decommissioned
+def method_decommissioned():
+    # Implementation has been deleted
+    pass
+
+
+@obsolete(decommission_at='2025-01-28', version="1.2.3")
+def method_obsolete():
+    return 'I still exist'
+
+
+class TestApiHelpers:
+    def test_method_decommissioned(self):
+        with raises(DeprecationWarning):
+            method_decommissioned()
+
+    @patch('warnings.warn')
+    def test_method_obsolete(self, mock_warnings_warn):
+        warning_message = (
+            "Function 'method_obsolete' is marked obsolete "
+            "since version '1.2.3' and will be decommissioned "
+            "at: 2025-01-28"
+        )
+        assert method_obsolete() == 'I still exist'
+        mock_warnings_warn.assert_called_once_with(warning_message)


schaefi:
Inspired by the python class I came up with some decorators that should help with the deprecation process for python API methods. Now that we are also adding types and annotations to the public members we should not break other peoples code without a good deprecation policy which we can express in code and manage

What do you think ?

davidcassany (approved):
Interesting, I like it. I am wondering if we could apply a similar approach on the XML schema somehow. I don't think there is any straight method to do so at xml validation level though.

schaefi:
Maybe some words about the process. I would be very interested in your feedback if my thinking makes sense.

1. There is a method that we decided gets deprecated. We will have a pull request that among others marks it as @obsolete
    and we give it a date when it will no longer be supported. I suggest 6 months from the day of the pull request.
    Users will now get a warning

2. We might come up with a little script that checks the kiwi code for the decommission dates and tells us when it's time to 
    decommission an obsoleted method. At that time a pull request is made that sets @decommissioned and replaces the
    implementation with a simple pass block.
    Users will now get a raise condition

decommissioned code we keep as decorators with empty blocks which is a pretty low overhead to drive things forward.

Thoughts ?

kiwi/api_helper.py
```diff
@@ -0,0 +1,82 @@
...
+
+        .. code:: python
+
+            @obsolete(decommission_at='2025-01-28')
+            def method():
+                print('Method implementation...')
+
+    :param str decommission_at: date string
+    """
+    def obsolete_method(func: Callable) -> Callable:
```
tomschr, line 46: I think the `method` in the function name could be a bit misleading and overloaded. :wink:  When I write decorators, I usually follow this naming scheme:

```python
def mydecorator(an_argument):
    """ ... """
    def decorate_mydecorator(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
             # ...
             return func(*args, **kwargs)
         return wrapper
     return decorate_mydecorator
```

Applying this principle to your code, I would suggest:

```suggestion
    def decorate_obsolete(func: Callable) -> Callable:
```

kiwi/api_helper.py
```diff
@@ -0,0 +1,82 @@
...
+        @wraps(func)
+        def function(*args, **kwargs):
```
tomschr, line 48: Applying the above naming scheme, this would lead to:

```suggestion
        def wrapper(*args, **kwargs):
```

kiwi/api_helper.py
```diff
@@ -0,0 +1,82 @@
...
+            nonlocal decommission_at
+            log.warning(
+                f'Function {func.__name__!r} is marked obsolete '
+                f'and will be decommissioned at: {decommission_at}'
+            )
+            return func(*args, **kwargs)
+        return function
```
tomschr, line 55:
```suggestion
        return wrapper
```

kiwi/api_helper.py
```diff
@@ -0,0 +1,82 @@
...
+    return obsolete_method
```
tomschr, line 56:
```suggestion
    return decorate_obsolete
```

kiwi/api_helper.py
```diff
@@ -0,0 +1,82 @@
...
+            @obsolete(decommission_at='2025-01-28')
+            def method():
+                print('Method implementation...')
+
+    :param str decommission_at: date string
+    """
+    def obsolete_method(func: Callable) -> Callable:
+        @wraps(func)
+        def function(*args, **kwargs):
+            nonlocal decommission_at
```
tomschr, line 49: The `nonlocal` keyword is **only** needed if you need write access to the enclosing scope. As you are only read the variable, Python finds it by looking up the scope chain.

```suggestion
```

kiwi/api_helper.py
```diff
@@ -0,0 +1,82 @@
...
+            log.warning(
+                f'Function {func.__name__!r} is marked obsolete '
+                f'and will be decommissioned at: {decommission_at}'
```
tomschr, line 52: Just an idea, you could use the `warnings` module. With the `warnings.simplefilter` you could filter such deprecation messages.

kiwi/api_helper.py
```diff
@@ -0,0 +1,82 @@
...
+            @decommissioned
+            def method():
+                pass
+
+    """
+    @wraps(func)
+    def function(*args, **kwargs):
+        raise KiwiDecommissionedApiFunctionCalled(
+            f'Function {func.__name__!r} has been decommissioned'
+        )
```
tomschr, line 81: Shouldn't you raise a `DeprecationWarning`?

kiwi/exceptions.py
```diff
@@ -795,3 +795,10 @@ class KiwiShellVariableValueError(KiwiError):
     Exception raised if a given python value cannot be converted
     into a string representation for use in shell scripts
     """
+
+
+class KiwiDecommissionedApiFunctionCalled(KiwiError):
```
tomschr, line 800: Maybe derive this from `DeprecationWarning` instead of `KiwiError`?

tomschr (commented):
Great idea, thanks Marcus! :+1:

As you can see, I've added my comments below. As always, some of them are a matter of taste. :wink: 

From a more general point of view, I have some additional ideas for you:

* Does it make sense to add further arguments (defaults, optional and/or required)?
   Maybe you could add some default arguments. That would need a small rewrite to allow that.
   However, it would make the decorator more flexible.
* Would it help to add the `warnings` module?
   If I remember correctly, the warnings can be caught and passed to the logging module. I thought it was a bit more appropriate than using the logging module.
   Additionally, you can call Python with `python -wd` to view all of these deprecation warnings (by default, they are suppressed). It probably depends on your use case.

I've created such a decorator [`@deprecated`](https://github.com/python-semver/python-semver/blob/master/src/semver/_deprecated.py) some time ago in the semver project. The decorator can be called with or without arguments. You need to be careful here as, for example,  `@deprecated` and `@deprecated()` is basically the same. However, it's easy to forget the parenthesis which could be error-prone.

For example, here is a small example:

```python
@deprecated(version="2.10.0")
def parse(version):
    # ...

@deprecated(replace="semver.Version.parse", version="2.10.0")
def parse_version_info(version):
    # ...
```

Hope that gives you some food for thoughts. :wink:

kiwi/api_helper.py
schaefi: I have seen this style at many places but was not so happy with the name "wrapper". It's actually a wrapper but the decorator itself is also a wrapper. And if you are using decorators with params you have the wrapper of the wrapper. But as I see this style at several places I will use it too. It seems to be kind of a common standard. Thanks will update the code

tomschr: You are basically right, I had the same sentiments. I explained it to myself that `wrapper` is the inner wrapper that is really applied to the decorated function. The others are just, well, some supportive "wrappers". :smile:

schaefi: done now

kiwi/api_helper.py
schaefi: done

kiwi/api_helper.py
schaefi: Ah ok, good to know. Thanks

kiwi/api_helper.py
schaefi: The kiwi api provides a logging class to handle all types of logging module controlled messages. I have implemented the use of the logging system such that you can register the kiwi logging class. If you do that as a user you benefit from the logging work done in kiwi. If you register your own logging class you can do whatever you like, if you don't setup logging you get the default python logging behavior which includes the WARN type. The idea was to handle log messages through pythons logging facility. I do not understand how the warnings module will help if the idea was to transport everything through the logging facility ?

kiwi/api_helper.py
schaefi: yes that's better. Thanks

kiwi/exceptions.py
schaefi: I think we don't need an extra class that handles the deprecation in the KiwiError namespace. Actually those classes exists to handle errors of kiwi as an application. If used as an API the deprecation conditions should be handled on the python level. So I dropped the above on behalf of DeprecationWarning

schaefi:
@tomschr 

```python
def deprecated(
    func=None,
    *,  # we support keyword-only arguments
    replace: str = None,
    version: str = None,
    category=DeprecationWarning):
    """
    Decorates a function to output a deprecation warning.

    :param func: the function to decorate.
    :param replace: the function to replace (use the full qualified
        name like ``module.function``.
    :param version: the first version when this function was deprecated.
    :param category: allow you to specify the deprecation warning class
        of your choice. By default, it's  :class:`DeprecationWarning`, but
        you can choose :class:`PendingDeprecationWarning` or a custom class.
    :return: decorated function which is marked as deprecated
    """
```

With regards to the above definitionI have some questions. In my suggestion I only allow to set a date from which a method marked obsolete is allowed to be moved to decommissioned and its implementation becomes an empty block.

Your policy in addition allows for:

* replace
  I wonder in which condition this will be ever used ? For example if you rename a public method, do you use a decorator
  to actually call the renamed method ? I think such an action deserves an extra decorator e.g @renamed

* version
  This provides the information about the first version when this method got deprecated. I wonder if this information is used in python API's. You can only install _one_ version of a python module. In that version a method is deprecated and it can't be another. Can you explain where you have used this information in practice ?

* category
  This specifies a warning class. What does it actually do ? Does it allow you to specify the raise target ?
  instead of ```raise DeprecationWarning``` it's ```raise given_category_class``` ? In what way have you used this
  feature in practice ? I think a user does not prepare its own code for catching deprecation errors.

Thanks for your insights of use cases to these extra options

tomschr:
After my serious power and Internet outage, I can comment again now. :wink: 

> With regards to the above definition I have some questions.

Sure! :slightly_smiling_face: 

The `replace` and `version` arguments are solely used for creating a useful warning message. No other magic involved. :wink: As such, there are no other decorators needed.

The `version` is the version of the first deprecated function and `replace` is used as a hint for the developer what to use instead.

For example, if someone calls these deprecated functions:

```python
print(semver.parse('3.4.5'))
print(semver.format_version(1, 2, 3))
```

you will get these two warning messages:

```
my-warning.py:6: DeprecationWarning: Function 'semver.parse' is deprecated. Deprecated since version 2.10.0.  This function will be removed in semver 3. Use the respective 'semver.Version.parse' instead.

my-warning.py:7: DeprecationWarning: Function 'semver.format_version' is deprecated. Deprecated since version 2.10.0.  This function will be removed in semver 3. Use 'str(versionobject)' instead.
```

By accident, both functions where deprecated at the same version. This is not always the case.

As you can see, sometimes it is not just renaming one function name into another. The developer has to use a different construct (see the warning message with `str(versionobject)`). That's why there is this `replace` argument which gives you a little bit more freedom to express the replacement. Both are just used to provide the developers with hopefully useful information what to do.

The `category` argument is different. It allows you to set a class that is used for creating the deprecation warnings. Usually you have these options (taken from the Python documentation, emphasize by me):

* [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning): Base class for warnings about deprecated features when those warnings are intended for other Python developers.
* [PendingDeprecationWarning](https://docs.python.org/3/library/exceptions.html#PendingDeprecationWarning): Base class for warnings about features which are obsolete and **expected to be deprecated in the future**, but are not deprecated at the moment.

   This class is rarely used as emitting a warning about a possible upcoming deprecation is unusual, and DeprecationWarning is preferred for already active deprecations.

This `category` argument is rarely used. However, it's possible to distinguish what to use if you have to. Or you want (for some reason) use a custom class. This argument gives you this flexibility.
Regardless what class you use, it is passed into this construct: `warnings.warn_explicit(... category=category, ...)`.

Does that explain the idea a little bit better?

kiwi/api_helper.py
tomschr: > I do not understand how the warnings module will help if the idea was to transport everything through the logging facility ?

Yes, I know you can emit warning messages with the logging module too. :slightly_smiling_face: The idea is a slightly different use case.

It boils down to this: Warnings from the logging module are **targeted for the normal user**. Warnings from the warnings module are **targeted to developers**. 

Sure, you can mix both. But sometimes you want to separate these two concerns.

The warnings module can help _developers_. By default, warning messages are disabled. For simple scripts it can be enabled by starting the Python script with `python -Wd SCRIPTNAME` and you can see all (even from other modules!). It's even possible to filter specific warnings.

Don't know if this is something which is really a use case for KIWI. My text was just providing you with this information, it's up to you. :slightly_smiling_face: 

The distinction between `logging.warning` and `warnings.warn` is described in this StackOverflow post: https://stackoverflow.com/a/37979724

If you want to know how to use warnings and logging together, I'd recommend to read the little section [Integration with the warnings module](https://docs.python.org/3/library/logging.html#integration-with-the-warnings-module) in the Python documentation.

tomschr: Oh, and I forge to mention another difference between these two: warnings show up only once. Quoted from official doc:

>    Repetitions of a particular warning for the same source location are typically suppressed.

Example:

```python
>>> import warnings
>>> warnings.warn("foo")
__main__:1: UserWarning: foo
>>> warnings.warn("foo")
>>> warnings.warn("foo")
```

schaefi: ok the differentiation between logging (users) and warnings(developers) makes sense to me. I will use this concept. Thanks

schaefi:
@tomschr  Thanks for the details. I agree to the usefulness of the version information. I did not thought about the fact that several deprecated methods could exists since different versions of kiwi and that it's useful to know this. I've added that part

> and replace is used as a hint for the developer what to use instead.

I'd like us to properly document that in the replacement code (docstring) if there is any and in the documentation. I think the information about a replacement can be relatively complex and usually is the result of a design change or refactoring. Thus I did not apply for this information in the obsolete data.

schaefi:
> This category argument is rarely used. In most cases, the DeprecationWarning class is enough. 

I think so too. As it comes with a default (DeprecationWarning) we will have an opportunity to add this setting at a later point in time should it turn out to be needed/requested. In this first implementation I would go without it

schaefi:
@tomschr How does it look like now ? ready for a merge ?

Thanks much

Conan-Kudo (approved):

tomschr (approved):
@schaefi Looks great! :+1: Thanks!

> I'd like us to properly document that in the replacement code (docstring) if there is any and in the documentation.

For my little project is was enough, but for KIWI it's a different story. Good idea! :+1:

schaefi:
Great thanks much for the very detailed feedback and review :smiley:

PR #1708, opened 2021-02-01
Fixed debootstrap device node conflict

### LINKED ISSUE #1675: creat debian10 arm64 demo fail
## Problem description
Host debian9 arm64
Compile the source code successfully wiki version 9.23.3
'kiwi-ng --profile=Disk --debug --type oem system build     --description kiwi-descriptions/debian/arm64/debian-buster     --target-dir ./debian10'
Print as follows ...
[ DEBUG   ]: 16:24:27 | bootstrap: I: Extracting libuuid1...
[ DEBUG   ]: 16:24:27 | bootstrap: I: Extracting libzstd1...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting login...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting mawk...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting mount...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting ncurses-base...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting ncurses-bin...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting passwd...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting perl-base...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting sed...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting sysvinit-utils...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting tar...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting tzdata...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting util-linux...
[ DEBUG   ]: 16:24:28 | bootstrap: I: Extracting zlib1g...
[ INFO    ]: Processing: [########################################] 100%
[ DEBUG   ]: 16:24:28 | command get error code 1   //my add print debootstrap commd error code 
[ ERROR   ]: 16:24:28 | KiwiBootStrapPhaseFailed: Bootstrap package installation failed: 
[ INFO    ]: 16:24:28 | Cleaning up SystemPrepare instance
[ DEBUG   ]: 16:24:28 | EXEC: [mountpoint -q /root/debian10/build/image-root/var/cache/kiwi]
[ DEBUG   ]: 16:24:28 | EXEC: [mountpoint -q /root/debian10/build/image-root/var/cache/kiwi]
[ DEBUG   ]: 16:24:28 | EXEC: [umount -l /root/debian10/build/image-root/var/cache/kiwi]
[ DEBUG   ]: 16:24:28 | EXEC: [rmdir --ignore-fail-on-non-empty /root/debian10/build/image-root/var/cache/kiwi]
[ DEBUG   ]: 16:24:28 | EXEC: [rmdir --ignore-fail-on-non-empty /root/debian10/build/image-root/var/cache]
[ DEBUG   ]: 16:24:28 | EXEC: [rmdir --ignore-fail-on-non-empty /root/debian10/build/image-root/var]
[ DEBUG   ]: 16:24:28 | EXEC: [rm -f /root/debian10/build/image-root/etc/resolv.conf]
[ DEBUG   ]: 16:24:28 | EXEC: [rm -f /root/debian10/build/image-root/etc/hosts]

Compare debian10 x86 normal installation log 
The place of failure should be 'Installing core packages...'

I tried before
1. Run alone
debootstrap --no-check-gpg --variant=minbase --include=grub-theme-starfield,usrmerge --components=main,contrib,non-free buster /root/debian10/build/image-root http://ftp.cn.debian.org/debian 
--Base system installed successfully.

2. Install related packages
$ pip3 install --upgrade pip
$ pip3 install numpy --upgrade --ignore-installed
$ pip install scipy --upgrade --ignore-installed
$ pip install scikit-learn --upgrade --ignore-installed

3. Change file attributes
$ chown root:root ./debian10

None of the above works...

## Expected behaviour
'kiwi-ng --profile=Disk --debug --type oem system build     --description kiwi-descriptions/debian/arm64/debian-buster     --target-dir ./debian10' 
No error code 1 occurs when kiwi-ng  calling external commands  debootstrap ..

schaefi:
debootstrap creates its own device node tree and fails
if a node it creates itself already exists. This commit
introduces a list of conflicting device nodes and deletes
them prior kiwi calling debootstrap. This Fixes #1675

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -139,8 +139,13 @@ class PackageManagerApt(PackageManagerBase):
         # chroot environment, thus we need to umount the kernel file systems
         # before calling debootstrap and remount them afterwards.
         root_bind.umount_kernel_file_systems()
-        # debootsrap will create its own dev/fd device
-        os.unlink(os.path.normpath(os.sep.join([self.root_dir, 'dev/fd'])))
+        # debootsrap will create its own dev/fd devices
+        debootstrap_device_node_conflicts = [
+            'dev/fd',
+            'dev/pts'
+        ]
+        for node in debootstrap_device_node_conflicts:
+            os.unlink(os.path.normpath(os.sep.join([self.root_dir, node])))
 
         if 'apt-get' in self.package_requests:
             # debootstrap takes care to install apt-get
--- a/test/unit/package_manager/apt_test.py
+++ b/test/unit/package_manager/apt_test.py
@@ -1,9 +1,10 @@
 import logging
-from mock import patch
+from mock import (
+    patch, call, Mock
+)
 from pytest import (
     raises, fixture
 )
-import mock
 
 from kiwi.package_manager.apt import PackageManagerApt
 
@@ -19,14 +20,14 @@ class TestPackageManagerApt:
         self._caplog = caplog
 
     def setup(self):
-        repository = mock.Mock()
+        repository = Mock()
         repository.root_dir = 'root-dir'
         repository.signing_keys = ['key-file.asc']
         repository.keyring = 'trusted.gpg'
         repository.unauthenticated = 'false'
         repository.components = ['main', 'restricted']
 
-        repository.runtime_config = mock.Mock(
+        repository.runtime_config = Mock(
             return_value={
                 'apt_get_args': ['-c', 'apt.conf', '-y'],
                 'command_env': ['env'],
@@ -78,7 +79,7 @@ class TestPackageManagerApt:
         self.manager.request_package('apt-get')
         mock_call.side_effect = Exception
         mock_exists.return_value = True
-        mock_root_bind = mock.Mock()
+        mock_root_bind = Mock()
         with raises(KiwiDebootstrapError):
             self.manager.process_install_requests_bootstrap(mock_root_bind)
 
@@ -90,10 +91,10 @@ class TestPackageManagerApt:
     ):
         self.manager.request_package('apt-get')
         self.manager.request_package('vim')
-        call_result = mock.Mock()
+        call_result = Mock()
         call_result.process.communicate.return_value = ('stdout', 'stderr')
         mock_call.return_value = call_result
-        mock_root_bind = mock.Mock()
+        mock_root_bind = Mock()
         mock_exists.return_value = True
         self.manager.process_install_requests_bootstrap(mock_root_bind)
         mock_call.assert_called_once_with(
@@ -104,11 +105,14 @@ class TestPackageManagerApt:
                 'root-dir', 'xenial_path'
             ], ['env']
         )
-        mock_unlink.assert_called_once_with('root-dir/dev/fd')
+        assert mock_unlink.call_args_list == [
+            call('root-dir/dev/fd'),
+            call('root-dir/dev/pts')
+        ]
         mock_root_bind.umount_kernel_file_systems.assert_called_once_with()
 
     def test_post_process_install_requests_bootstrap(self):
-        mock_root_bind = mock.Mock()
+        mock_root_bind = Mock()
         self.manager.post_process_install_requests_bootstrap(mock_root_bind)
         mock_root_bind.mount_kernel_file_systems.assert_called_once_with()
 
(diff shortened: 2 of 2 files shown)


schaefi:
@davidcassany since I was working on the boxes for the box plugin for other archs I came across this one which is on the sprint. I know you are also working on this problem and I hope I did not a conflicting change. I've created the PR because I'm not sure if this should be the solution. At the moment we maintain a static list of conflicting nodes for debootstrap. However @piqao solved this by removing the entire dev/ tree prior debootstrap.

That's why I've opened the PR to have a conversation about it.

I'm still working on the boxes to build for arm and I had no chance to actually test the fix here on arm

davidcassany (approved):
Looks good to me

davidcassany:
@schaefi this what I had in mind as an obvious and immediate fix. I was trying to figure out why this issue was not raising in x86_64 as it looks arch independent. However I could not really figure it out.

schaefi:
> @schaefi this what I had in mind as an obvious and immediate fix. I was trying to figure out why this issue was not raising in x86_64 as it looks arch independent. However I could not really figure it out.

Same for me which is why I think we should have a successful test on arm prior merge. Maybe  @piqao can help

davidcassany:
Another idea that I was considering is making the `root_init`, `root_import` and `root_bind` classes aware of or dependent on the package manager, as the package manager is likely to constraint or define details of the bootstrap procedure. So we could eventually assume a different bootstrap procedure for apt and instead of undoing some the predefined steps we could just omit them from the very begining.

This implies a refactor and changes from API perspective even the implementation, at glance, doesn't look like something complex. For me the down side is that adding a flag for the package manager in the mentioned classes is not that natural or obvious.

schaefi:
>Another idea that I was considering is making the root_init, root_import and root_bind classes aware of or dependent on the >package manager, 

My problem with this approach is that root_init, root_import and root_bind are not connected to a certain package manager. From a design perspective I want them to be decoupled even for the price to pay that some package manager has a problem with it. The current way to explicitly fix the inconsistency in the code that needs the fix is imho better than skipping code that imho belongs to (root_init, root_import, root_bind). From my perspective there is no code in the mentioned root_* classes that should be skipped with regards to their responsibility. 

The problem here is because we use debootstrap and it doesn't know that we have already prepared the environment. debootstrap assumes it has to mess with device nodes and such things and that causes the conflict. Therefore I'm ok to solve the conflict at the code part that uses debootstrap, it makes this conflict explicit and that's imho what we want to see in there.

Thoughts ?

davidcassany:
> The problem here is because we use debootstrap and it doesn't know that we have already prepared the environment. debootstrap assumes it has to mess with device nodes and such things and that causes the conflict. Therefore I'm ok to solve the conflict at the code part that uses debootstrap, it makes this conflict explicit and that's imho what we want to see in there.

Yes I agree, that's why I suggested to just undo some of the stuff inside the apt package manager classes from the very beginning. I am just afraid and wondering if further issues will appear because of that. It can happen we nearly undo everything that has been already done in the rootfs before calling debootstrap. In any case I am fine with the current setup as this easy to read and the issue is isolated in apt package manager land.

schaefi:
> It can happen we nearly undo everything that has been already done in the rootfs before calling debootstrap.

if we reach that point we can think about a "skip" design that allows us to set this as a flag. This way the classes stay decoupled from knowing about a package manager. But of course the decision has to be made at another place, probably at the builder stage and will leave us with the question how to make the decision for root_bind yes|no.

I also think debootstrap could be much smarter when it sees an environment that already has what it needs it doesn't have to complain and fail. So overall I think we have to deal with it but it should not make us to destroy some of the kiwi design aspects.
For the moment I'd say we go with what we have, it's still easy enough to understand why and what's going on

schaefi:
Problem here is that there is no feedback from the issue reported. I don't have a debian10 arm64 description and no time to create one. Therefore I asked if the issue reporter can provide it... but no feedback so far

I'll give this one time until Monday. Without feedback this gets merged and we keep our fingers crossed that it fixes the issue. If not I'm sure people will come back :)

Does this makes sense to you too ?

Thanks

Conan-Kudo (approved):
Something about this bothers me, but Debian bootstrap is sufficiently weird that I'm willing to let it go.

davidcassany:
Yes makes perfect sense to me too. Thanks

Conan-Kudo:
Gack, no, I accidentally merged this instead of something else... 😭

PR #1710, opened 2021-02-02
Added option to allow reading obs project repos

schaefi:
If kiwi is used in OBS many people setup the repositories
in the OBS project metadata. In the KIWI image description
only a placeholder called obsrepositories needs to be added
to indicate that OBS is managing the repositories. This is
nice but prevents people from simply building this image
outside of OBS. This commit adds a new caller option:

```
    --set-obs-repos=<user,password,project,name,profile,arch,repo>
```

which allows to read and use the list of repositories as OBS
would use it but in a local build process. KIWI will send a _buildinfo
request to the OBS API and reads the provided result. For this
call to work the user needs to provide its obs account information.
The account data is only used in a HTTPBasicAuth request and
not stored or used in any other way.

### DIFF
--- a/kiwi/obs.py
+++ b/kiwi/obs.py
@@ -1,0 +1,238 @@
+# Copyright (c) 2021 SUSE Software Solutions Germany GmbH.  All rights reserved.
+#
+# This file is part of kiwi.
+#
+# kiwi is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# kiwi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kiwi.  If not, see <http://www.gnu.org/licenses/>
+#
+import os
+import logging
+from lxml import etree
+import requests
+from requests.auth import HTTPBasicAuth
+from tempfile import NamedTemporaryFile
+from typing import List
+
+# project
+from kiwi.xml_state import XMLState
+from kiwi.runtime_config import RuntimeConfig
+from kiwi.solver.repository.base import SolverRepositoryBase
+from kiwi.system.uri import Uri
+from kiwi.command import Command
+
+from kiwi.exceptions import (
+    KiwiUriOpenError,
+    KiwiOBSBuildInfoError,
+    KiwiOBSProjectError,
+    KiwiOBSSourceError
+)
+
+log = logging.getLogger('kiwi')
+
+
+class OBS:
+    """
+    **Implements methods to access the Open Build Service API**
+    """
+    def __init__(
+        self, image_path: str, user: str, password: str,
+        profiles: List[str], arch: str, repo: str
+    ):
+        """
+        Initialize OBS API access for a given project and package
+
+        :param str image_path: OBS project/package path
+        :param str user: OBS account user name
+        :param str password: OBS account password
+        :param list profiles: OBS image multibuild profile list
+        :param str arch: OBS architecture
+        :param str repo: OBS image package build repository name
+        :param str api_server:
+            OBS api server to use, defaults to https://api.opensuse.org/build
+        """
+        runtime_config = RuntimeConfig()
+        try:
+            (self.project, self.package) = image_path.split(os.sep)
+        except ValueError:
+            raise KiwiOBSProjectError(f'Invalid image path: {image_path}')
+        self.user = user
+        self.profile = profiles[0] if profiles else None
+        self.password = password
+        self.arch = arch or 'x86_64'
+        self.repo = repo or 'images'
+        self.api_server = runtime_config.get_obs_api_server_url()
+
+    def fetch_obs_image(self, checkout_dir) -> str:
+        """
+        Fetch image description from the obs project
+
+        :return: temp local path to image description
+
+        :return: path to local image checkout
+
+        :rtype: str
+        """
+        log.info('Checking out OBS project:')
+        if os.path.exists(checkout_dir):
+            raise KiwiOBSSourceError(
+                f'OBS source checkout dir: {checkout_dir!r} already exists'
+            )
+        log.info(f'--> {self.project}/{self.package}')
+        package_link = os.sep.join(
+            [
+                self.api_server, 'source',
+                self.project, self.package
+            ]
+        )
+        request = self._create_request(package_link)
+        package_source_xml_tree = OBS._import_xml_request(request)
+        package_source_contents = package_source_xml_tree.getroot().xpath(
+            '/directory/entry'
+        )
+        if not package_source_contents:
+            raise KiwiOBSSourceError(
+                f'OBS source for {self.package!r} package is empty'
+            )
+        source_files = []
+        for entry in package_source_contents:
+            source_files.append(entry.get('name'))
+
+        if '_service' in source_files:
+            raise KiwiOBSSourceError(
+                f'OBS source for {self.package!r} package is a source service'
+            )
+        Command.run(['mkdir', '-p', checkout_dir])
+        for source_file in source_files:
+            log.info(f'--> {source_file}')
+            request = self._create_request(
+                os.sep.join([package_link, source_file])
+            )
+            with open(os.sep.join([checkout_dir, source_file]), 'wb') as fd:
+                fd.write(request.content)
+        return checkout_dir
+
+    def add_obs_repositories(self, xml_state: XMLState) -> None:
+        """
+        Add repositories from the obs project to the provided XMLState
+
+        :param XMLState xml_state: XMLState object reference
+        """
+        if not OBS._delete_obsrepositories_placeholder_repo(xml_state):
+            # The repo list does not contain the obsrepositories flag
+            # Therefore it's not needed to look for repos in the OBS
+            # project configuration
+            return None
+
+        package_name = self.package if not self.profile \
+            else f'{self.package}:{self.profile}'
+        log.info(f'Using OBS repositories from {self.project}/{package_name}')
+        buildinfo_link = os.sep.join(
+            [
+                self.api_server, 'build',
+                self.project, self.repo, self.arch,
+                package_name, '_buildinfo'
+            ]
+        )
+        request = self._create_request(buildinfo_link)
+        buildinfo_xml_tree = OBS._import_xml_request(request)
+        repo_paths = buildinfo_xml_tree.getroot().xpath(
+            '/buildinfo/path'
+        )
+        if not repo_paths:
+            raise KiwiOBSBuildInfoError(
+                f'OBS buildinfo for {package_name} has no repo paths'
+            )
+        repo_prio_ascending = 0
+        repo_prio_descending = 501
+        repo_alias = None
+        for repo_path in repo_paths:
+            repo_url = repo_path.get('url')
+            if not repo_url:
+                repo_url = 'obs://{0}/{1}'.format(
+                    repo_path.get('project'), repo_path.get('repository')
+                )
+            if repo_url:
+                log.info(f'OBS Repo: {repo_url}')
+                try:
+                    repo_uri = Uri(repo_url)
+                    repo_url = repo_uri.translate(
+                        check_build_environment=False
+                    )
+                    request = requests.get(repo_url)
+                    request.raise_for_status()
+                except Exception as issue:
+                    log.warn(
+                        f'--> Unreachable repo ignored: {issue}'
+                    )
+                    continue
+
+                repo_check = SolverRepositoryBase(repo_uri)
+                repo_type = repo_check.get_repo_type()
+                if not repo_type:
+                    log.warn('--> Unknown repo type ignored')
+                    continue
+
+                if repo_type == 'rpm-md':
+                    repo_prio_ascending += 1
+                    repo_prio = repo_prio_ascending
+                else:
+                    repo_prio_descending -= 1
+                    repo_prio = repo_prio_descending
+
+                log.info('--> OK')
+                xml_state.add_repository(
+                    repo_url, repo_type, repo_alias, f'{repo_prio}'
+                )
+
+    def _create_request(self, url):
+        try:
+            request = requests.get(
+                url, auth=HTTPBasicAuth(self.user, self.password)
+            )
+            request.raise_for_status()
+        except Exception as issue:
+            raise KiwiUriOpenError(
+                f'{type(issue).__name__}: {issue}'
+            )
+        return request
+
+    @staticmethod
+    def _delete_obsrepositories_placeholder_repo(xml_state):
+        """
+        Delete repository sections which uses the obsrepositories
+        placeholder repo and keep the rest for the configured profiles
+
+        :return:
+            Returns True if a placeholder repo got deleted
+            otherwise False
+
+        :rtype: bool
+        """
+        has_obsrepositories = False
+        repository_sections_to_keep = []
+        repository_sections = xml_state.get_repository_sections()
+        for xml_repo in repository_sections:
+            repo_source = xml_repo.get_source().get_path()
+            if 'obsrepositories' in repo_source:
+                has_obsrepositories = True
+            else:
+                repository_sections_to_keep.append(xml_repo)
+        xml_state.xml_data.set_repository(repository_sections_to_keep)
+        return has_obsrepositories
+
+    @staticmethod
+    def _import_xml_request(request):
+        download = NamedTemporaryFile()
+        with open(download.name, 'wb') as request_info:
+            request_info.write(request.content)
+        return etree.parse(download.name)
--- a/kiwi/obs.py
+++ b/kiwi/obs.py
@@ -46,7 +46,7 @@ class OBS:
     """
     def __init__(
         self, image_path: str, user: str, password: str,
-        profiles: List[str], arch: str, repo: str
+        ssl_verify: bool, profiles: List[str], arch: str, repo: str
     ):
         """
         Initialize OBS API access for a given project and package
@@ -68,6 +68,7 @@ class OBS:
         self.user = user
         self.profile = profiles[0] if profiles else None
         self.password = password
+        self.ssl_verify = ssl_verify or True
         self.arch = arch or 'x86_64'
         self.repo = repo or 'images'
         self.api_server = runtime_config.get_obs_api_server_url()
@@ -197,7 +198,8 @@ class OBS:
     def _create_request(self, url):
         try:
             request = requests.get(
-                url, auth=HTTPBasicAuth(self.user, self.password)
+                url, auth=HTTPBasicAuth(self.user, self.password),
+                verify=self.ssl_verify
             )
             request.raise_for_status()
         except Exception as issue:
--- a/doc/source/commands/system_build.rst
+++ b/doc/source/commands/system_build.rst
@@ -30,6 +30,7 @@ SYNOPSIS
    kiwi-ng system build --obs-image=<path> --obs-user=<name> --target-dir=<directory>
        [--obs-arch=<arch>]
        [--obs-repo=<repo>]
+       [--obs-ssl-no-verify]
    kiwi-ng system build help
 
 .. _db_kiwi_system_build_desc:
@@ -139,6 +140,10 @@ OPTIONS
   set to any custom name, it's only required to specify the
   used repository name if another than the OBS default
   name is used.
+
+--obs-ssl-no-verify
+
+  Dont't verify SSL server certificate when connecting to OBS
 
 --set-repo=<source,type,alias,priority,imageinclude,package_gpgcheck>
 
--- a/doc/source/commands/system_build.rst
+++ b/doc/source/commands/system_build.rst
@@ -27,6 +27,9 @@ SYNOPSIS
        [--set-container-tag=<name>]
        [--add-container-label=<label>...]
        [--signing-key=<key-file>...]
+   kiwi-ng system build --obs-image=<path> --obs-user=<name> --target-dir=<directory>
+       [--obs-arch=<arch>]
+       [--obs-repo=<repo>]
    kiwi-ng system build help
 
 .. _db_kiwi_system_build_desc:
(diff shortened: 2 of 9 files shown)


schaefi:
@davidcassany because I was tired of copying around repos when rebuilding projects from obs I finally added the above code. Frankly speaking I don't like ```osc build``` as it just adds another layer which I think is not needed just for the repo handing.
So instead of:

```
$ osc co Virtualization:Appliances:Images:Testing_x86:suse
$ cd Virtualization:Appliances:Images:Testing_x86:suse/test-image-live
$ osc build .. plus-some-yadayada-foo-foo-kvm options
```

you can go like this

```
$ kiwi-ng system build --description build-tests/x86/suse/test-image-live \
       --set-obs-repos 'user,pass,Virtualization:Appliances:Images:Testing_x86:suse,test-image-live'
```

schaefi:
And the same of course will work as a boxbuild :)

kiwi/xml_state.py
```diff
@@ -1622,6 +1631,71 @@ def delete_repository_sections_used_for_build(self) -> None:
...
+                    # we will support real rpm-md repos only. The check for
+                    # the timestamp of the primary repodata object is an
+                    # indicator If the repo can be used, if not the repo
+                    # will be ignored.
+                    log.debug(
+                        f'Repo: {repo_url} is not a valid rpm-md type: {issue}'
+                    )
+                    use_repo = False
+                if use_repo:
+                    self.add_repository(repo_url, 'rpm-md')
```
davidcassany, line 1697: What about having some sort of increasing priority. So first has highest priority, this way the result is likely to be closer to what OBS is doing. This is specially sensitive when testing specific patches that relay on its own project repository being appended on top of the repos list.

davidcassany:
>  I was tired of copying around repos when rebuilding projects from obs I finally added the above code

Yes I understand that, on SLE SPs this is pretty annoying that list can be terribly long. This is definitely a handy feature for OBS users.

davidcassany (commented):
Overall looks good to me. Nice feature! I am just wondering about the repositories priority, Shoudn't we try to mimic a little the OBS behavior? It won't be still the same as the OBS solver is just different, but we would be closer.

kiwi/xml_state.py
schaefi: Oh yes this is a good idea. The problem is how should this be done consistently across the package managers. For zypper a low prio is 99 and a high prio is 1. So in this case it could be just "1 2 3 4". I think dnf is the same according to https://dnf.readthedocs.io/en/latest/conf_ref.html. How is it done for apt and pacman... still don't know

schaefi: Oh and it probably doesn't matter for non rpm-md repos because the current code only uses rpm-md repo, which could be a problem if the image is build for arch or debian based distros. Hmm a good question actually

schaefi: ok for an ubuntu build a path element could look like this

```xml
<path project="Ubuntu:20.04" repository="universe" url="https://download.opensuse.org/repositories/Ubuntu:/20.04/universe/"/>
```

And very interesting is that the provided URL is not reachable and gives a 404

schaefi: The Pin-Priority for apt works the other way round, low number == low prio

schaefi: ok so looking closer opens some questions for me:

* obs buildinfo reports urls which are not accessible from the outside like https://download.opensuse.org/repositories/Ubuntu:/20.04/universe/ but obs can build with them ??
* the prio setting is different per package manager, how to handle this
* the current code does a repo_check based on a timestamp. This only works for rpm-md repos because kiwi/solver/repository has no implementation for apt and pacman repotypes. Maybe there is a better check to find out the repo type ?

davidcassany: Well I believe it doesn't really matter for apt, pacman and probably for dnf too, at least for SUSE's OBS instances. So I believe the key factor is that only published repositories are going to be functional if I understand it correctly. Meaning that third party repositories won't work and unpublished repositories too (in OBS they can be unpublished but still used for build).

schaefi: @davidcassany I think I can add a get_repotype method to the solver code. Can you help me with the identification ?
for rpm-md the presence of repodata is a good check, for apt I would look for Packages.gz and for pacman I would look for _Linux.db. Does this sound correct ? pacman would be worst because the meta data file names are not static. Thoughts ?

davidcassany: >     * obs buildinfo reports urls which are not accessible from the outside like https://download.opensuse.org/repositories/Ubuntu:/20.04/universe/ but obs can build with them ??

I believe this is similar to what happens when using repositories that not published. OBS first tries to fetch them using the URL and in case it fails or it is not found then it tries to download them using the OBS API.

schaefi: > Well I believe it doesn't really matter for apt, pacman and probably for dnf too, at least for SUSE's OBS instances. So I believe the key factor is that only published repositories are going to be functional if I understand it correctly. Meaning that third party repositories won't work and unpublished repositories too (in OBS they can be unpublished but still used for build).

yes and the code will find out if the repo can be reached or not and will add a debug log. That should be better a warning message. I will change that. But we still need a better detection of the repo type. Only rpm-md is not enough because in obs you can decided if you publish or not as you said. For example someone building against Virt:App:Builder for ubuntu will use the apt repo with python3-kiwi in it which is also reachable publicly.

davidcassany: I am thinking of another approach to do something similar. What if instead of making KIWI using some remote repositories we use some local ones created by `osc`. This way we use the same exact repositories and we are likely to achieve pretty equivalent results.

Gonna investigate if it is possible to make `osc` run the solver and build the repositories without actually build the image.

schaefi: or maybe easier, we let the package manager decide for the type and just add the repo if it's reachable. But obs also contains things like this

```xml
<path project="openSUSE:Tumbleweed" repository="dod" url="https://download.opensuse.org/repositories/openSUSE:/Tumbleweed/dod/"/>
```

which I also don't understand

schaefi: but this is also not public, so can be ignored. I'll update the code and then we need to find a way how to handle prios

davidcassany: Yet another idea. The command `osc build --preload images x86_64` just runs the solver and creation of repositories. What if we code that as a new KIWI subcommand, so we could have an extra command that runs this preload. Probably it is not a crazy idea trying to code our own `osc build --preload images x86_64` to avoid `osc` dependencies. The solver is executed remotely as you did here and this PR. What if we also download the packages? The XML provides the list of packages (including arch, project and repo for each), so probably we can simply run OBS requests to download them.

schaefi: Hmm, so you are saying we create a new sub-command that does:

* send OBS api request to the get the _buildinfo
* use the _buildinfo data and download all packages. This would then be a _service:download_url: (not sure) api call
* collect all packages and create a repo out of them. Here I see a difficulty. for rpm-md it's a simple createrepo call but how to handle other repo-types ?
* let kiwi build with --set-repo _the-collective-repo_ instead of obsrepositories

Something like this ?

How to handle caching of downloaded packages ?

I'm not sure if we should do this. Actually some of the problems we don't like from the obs approach (that different solver and the re-creation of repos) would be again implemented but now in kiwi. I think this is not needed because if this is the plan people can use "osc build" with all its consequences and side-effects.

With this PR here my idea was to allow using the list of repositories from obs and all the rest is disconnected from obs as a service. I agree with you this requires the repos to be published and if they are not another option can be used. Options in this case are:

* find the public repos if they exist
* make use of "osc build"

But coding kiwi's own osc build implementation ties it strictly to the obs api, requires a credentials setup and re-implements quite some code from osc itself.

Do you think this is worth the effort ?

schaefi: ```osc build --preload images x86_64``` for this to work you also have to checkout the project from obs, correct ?

davidcassany: > `osc build --preload images x86_64` for this to work you also have to checkout the project from obs, correct ?

yes

davidcassany: > How to handle caching of downloaded packages ?

IIRC we are already doing that. There is no caching for local repositories.

>     * collect all packages and create a repo out of them. Here I see a difficulty. for rpm-md it's a simple createrepo call but how to handle other repo-types ?

Yes that's a good point and an issue I underestimated, I was thinking of plain rpm-dir, but yes this is likely to be not sufficient in many cases.

> I think this is not needed because if this is the plan people can use "osc build" with all its consequences and side-effects.

We are approaching this feature slightly different and I see your point. My idea was more in the direction of `osc build` is not trustable and sometimes fails locally and it doesn't remotely, so what if we find and functional alternatively.

> But coding kiwi's own osc build implementation ties it strictly to the obs api, requires a credentials setup and re-implements quite some code from osc itself.

Yes that's true. Yes the real issue of this is the repo creation which is package manager specific, I was just thinking of only downloading binaries from OBS using some simple structured folder setup <repo>/<package>, but yes this is not sufficient and the missing part is getting into the burden of OSB build procedures.

schaefi: @davidcassany ok I've added a very simple get_repo_type() method that tries to lookup repo metada on the server. The implementation is poor for Arch as they don't seem to have static repo metadata files like anyone else has.

Together with the information about the repo type I added the repo priority setting as you suggested.

Given the repos are available publicly (or can be accessed by the configured OBS download_url in the kiwi.yml) this feature can be used to build with the obs repos and prios

BUT

I'm not sure if we should merge it

schaefi:
@davidcassany I think I'm complete with what I wanted to achieve but your feedback makes me think about it.
Actually I think you are right. It would be nice if we can just say

```
$ kiwi system build --obs-project ...
```

* the repository problem if obsrepositories is used would be solved by the code here
* we would only need to add a temporary project checkout to fetch the image

What I don't like for the mentioned reasons is that local creation of extra repos like osc build does it. I'd like to use
the package manager backend to build the image. From my perspective all nasty hacks to an image description to make it build in obs are because of that "we solve our own way and we create our own local repos" concept. The good news is that from my perspective the hacks are only needed into one direction. Meaning an image that builds in obs always builds outside of obs with the native package manager given the repos can be setup correctly.

From that perspective what do you think about the idea ?

Thanks

davidcassany:
> It would be nice if we can just say
> 
> ```
> $ kiwi system build --obs-project ...
> ```
> 
> * the repository problem if obsrepositories is used would be solved by the code here
> * we would only need to add a temporary project checkout to fetch the image
 
So basically you suggest keep the current repositories resolutions as you coded here in this PR and add extra logic to checkout the project code. So repository wise we would still have the limitations we already detected in this PR but we would prevent having to have a pre-checkout of the project. So from the user perspective the OBS checkout would not be needed. Looks nice, it is pretty straight forward from user POV. It's like providing `--obs-project` as an alternative to `--description`, looks good.

> Meaning an image that builds in obs always builds outside of obs with the native package manager given the repos can be setup correctly.

I would change _always builds_ to _is likely to build_ since you can cheat a lot in _prjconf_ regarding dependencies. I agree that usually all the hacks we add in _prjconf_ are meant to solve issues in the build environment, so they do not affect us in this case. So I agree this is the case for the very very vast majority of cases.

Regarding the repositories resolution I just realized that we should also resolve the base container for derived images. In any case I'd say this can be done in a followup step. I think we can start with your last proposal of `--obs-project` flag instead of `--description` and assume it has it's limitations. Then we can try to further work on it to sort out some of these limitations.

> From that perspective what do you think about the idea ?
> 

I like it, the only down side I see is having to provide the credentials of OBS, I guess we can't avoid that. From a user POV is not very nice having to add a password in the command line in clear text. But I don't see any other straight forward solution now.

schaefi:
Regarding the credentials there is no other way when talking to the api server, you are right. We could code it in a way that kiwi reads the password from stdin. This allows for further options to "push-it-in"

schaefi:
> I would change always builds to is likely to build since you can cheat a lot in prjconf regarding dependencies.

yes you are right, I was too motivated when writing it :-)

schaefi:
ok so a call could look like this

```
$ kiwi-ng --profile Kernel system build --obs-project Virtualization:Appliances:SelfContained:suse \
    --obs-user sax2 --obs-image box
```

The call will then ask for the API password for user sax2 once

The already existing --profile option is used to name the multibuild profile (in case of multibuild) which is the same name

The API call to checkout the sources is easy. There is just one issue. If a source service is used it would be an extra effort to get to the sources. I know obs holds a copy but I don't know how to fetch it. Do you know that ?

Conan-Kudo:
@schaefi I think we need a way to pass authentication parameters to kiwi outside of the CLI arguments. That's too risky for things like CI environments.

schaefi:
> @schaefi I think we need a way to pass authentication parameters to kiwi outside of the CLI arguments. That's too risky for things like CI environments.

yes agreed

davidcassany:
> The API call to checkout the sources is easy. There is just one issue. If a source service is used it would be an extra effort to get to the sources. I know obs holds a copy but I don't know how to fetch it. Do you know that ?

No I don't... And yes, obs _services like scm service or any other buildtime service will break with this approach and it is likely to cause the build to fail. Man, there are many corner cases and limitations beyond the happy path.

Regarding SCM services, in general, they are on disabled mode as having them activated is not permitted in any official namespace (either SUSE or openSUSE). So in most cases SCM won't be a problem. It will for our integration tests, but this is something we could consider changing. Having the SCM source disabled has some benefits, and of course some downsides too. Meaning there are decent alternatives to work around the SCM issue on OBS side. So I'd probably would not stress much on finding a way apply SCM services on KIWI (or download the result of it).

The more I think about it the more I believe it would make a lot of sense building a KIWI plugin that actually imports osc project and just reuses osc api and let osc handle all OBS interaction. The _only_ big problem is the `osc` code that is not really designed to be used as a library and I'd say the implementation is chaotic and probably a source of issues by itself. But if `osc` would be in great shape I think it would be reasonable to create such a plugin :man_shrugging:

Conan-Kudo:
Yeah, it probably makes sense to make this a KIWI plugin, because then you could also make plugins to support workflows with Koji or any other build system too.

schaefi:
I've also thought to have something like ```from osc ... import``` but hey guys it really sucks. This osc code is made as an application and not to be used as an API.

The plugin idea is nice though

schaefi:
creating a plugin is a bit more work and overhead. Let me first code something to share with you and then make it a plugin if we think it's worth it

schaefi:
ok, I've finished the new --obs-image system build option. An example could look like this

```
$ sudo kiwi-ng --profile Azure system build \
    --obs-image Devel:PubCloud:CrossCloud:SLE15-SP1/SLES15-SP1-CHOST-BYOS \
    --obs-user sax2 --target-dir /tmp/mytest/
```

This will ask for the user password once from stdin. To control the open buildservice location you can set custom values
in the kiwi config file. For example to point to the suse internal server this would be:

```yaml
obs:
  - public: false
  - api_url: https://api.suse.de
  - download_url: https://download.suse.de/ibs
```

And when written into a file this can be used with the new --config option:

```
$ sudo kiwi-ng --config suse_obs.yml ...
```

of course only useful with a vpn connection into the suse network. By default kiwi uses the public open build service
location, but it can be any location also other build service instances.

Now the question is:

* Useful feature ? I think it is, since it finally allows me to run local builds from obs images
* Source services are not supported. As David said for anything released this will not be a problem as at least suse's obs does not allow source services for released customer data. But might be useful to read and process source services that uses git ?
* Plugin or not ? If this code is moved into a plugin it must be rewritten such that the plugin registers a complete new command. It's not possible to just add a new option into an existing command (build in this case) via the plugin infrastructure. This means we can move the code into a plugin and provide a new command e.g ```kiwi-ng system obs_checkout ...``` which checks out the project from obs such that a followup ```kiwi-ng system build ...``` can actually build it. So it moves into a two pass concept.

Feedback welcome

Thanks

kiwi/obs.py
```diff
@@ -0,0 +1,240 @@
...
+        if not package_source_contents:
+            raise KiwiOBSSourceError(
+                f'OBS source for {self.package!r} package is empty'
+            )
+        source_files = []
+        for entry in package_source_contents:
+            source_files.append(entry.get('name'))
+
+        if '_service' in source_files:
+            raise KiwiOBSSourceError(
```
davidcassany, line 112: This will raise in any sort of `_service` file exists in sources right? I would not do so. I would just throw a warning, it is common practice in OBS having the `_service` file in sources with all the services in `disabled` mode. So in case there is the presence of a `_service` I would just throw a big warning to let the user now the _service file won't be processed and if this is needed for the build it is most likely to fail at some point.

davidcassany (commented):
About the implementation logic I have just one comment regarding the `_service` file detection. I believe raising an exception is present is probably too strict.  On the other hand I understand this limits significantly the range of non-obvious failures.

davidcassany:
> Now the question is:
> 
> * Useful feature ? I think it is, since it finally allows me to run local builds from obs images

Yes, I also believe it is useful. i'd say there is no discussion here.

> * Source services are not supported. As David said for anything released this will not be a problem as at least suse's obs does not allow source services for released customer data. But might be useful to read and process source services that uses git ?

Regarding that see my comment on the review. For released stuff source services are not allowed in **active** mode, if disabled there is no problem, and this is the case for a wide range of packages in OBS and possibly images too. In any case I would not consider trying to solve any sort of `_service` for now, probably we can consider it later, I think now it would add too much complexity for an uncertain gain.

> * Plugin or not ? If this code is moved into a plugin it must be rewritten such that the plugin registers a complete new command. It's not possible to just add a new option into an existing command (build in this case) via the plugin infrastructure. This means we can move the code into a plugin and provide a new command e.g `kiwi-ng system obs --obs-image ...` which checks out the project from obs such that a followup `kiwi-ng system build ...` can actually build it. So it moves into a two pass concept.

For the plugin approach I like the following:

* As a plugin we are drawing a line between KIWI and OBS, to my eyes this is beneficial. I know we already have some specifics of OBS within the KIWI code, but my desire would be reducing it not increasing it.
* We can decouple this feature from KIWI releases and modify and alter it without having to mess with KIWI releases. So we not that coupled to it in case we want to dramatically modify it after a while.
* Since this is decoupled from KIWI I would consider easier to then embrace the support for other build services out there without having to actually support them as part of KIWI core.

However from the plugin approach I dislike:

* It requires more code
* It adds another piece to maintain and keep track off, more repos, more OBS packages, etc. so it probably increases the maintenance overhead compared to a native KIWI feature.
* It  might not be that obvious or simple to use compared with a native feature that show up with a simple `kiwi-ng system build --help`

 
Probably I have some preference to code it as a plugin, however I can see the simplicity of coding it natively in KIWI which is also something to consider.

davidcassany:
Oh and I have just realized two more cases that are not covered by this PR.

* Derived container images. The `obsrepositories:/` approach is also used for the URI of the base container image when building derived containers. This is is niche case and I believe solving that in KIWI is possible however is again re-implementing some OBS specific logic. Probably this is not worth to take it now as most container images are have some `builtime` services in place, so even if the URI of the base image is solved it still won't work.

* What happens when the root overlay is downloaded a CPIO archive instead of a `root.tar.gz` tarball? This is the case of our integration tests for which the root overlay is downloaded as a `root.obscpio` file. AFAIK `*.obscipo` files are just expanded at build time right before actually starting the build.

For both cases I would not care now. As tyring to solve these limitations leads us into reimplementing part of the OBS build logic.

kiwi/obs.py
schaefi: yes makes sense. I would even go that far and resolve the _service if it uses ```<service name="obs_scm">``` and ```<param name="scm">git</param>``` That should be really use to implement. In any other case we just take the file as you suggested

schaefi:
Thanks much for your detailed feedback. I think we should the plugin way. The plus points you mentioned imho weight more than the overhead a plugin creates. I've prepared a new plugin project within an hour, this is not much work :)

* https://github.com/OSInside/kiwi-obs-plugin/pulls

I also think we should use the github actions for testing approach. It's really nice and for plugins only the pypi release is a bit more work.

>Derived container images. The obsrepositories:/ approach is also used for the URI of the base container image when building derived containers.

yeah that I haven't tested at all. At the moment the checkout will not adapt the derived_container=... attribute. So the local checkout will most probably not be able to build. I also don't know if this is even resolvable by an api lookup ? The current way to resolve obsrepositories is by looking in the _buildinfo API call. I don't think this is sufficient for derived_container resolution ?
But this could be added later... maybe ?

> What happens when the root overlay is downloaded a CPIO archive instead of a root.tar.gz tarball?

right this happens when _service is in use. So currently this will raise an error. I know you said we should not mess with the source service but resolving the scm service with git is something I would implement. This will allow us to build any of the source service controlled image builds we setup (integration tests) and also public cloud images ... maybe some of the stuff other people use. Any other source-service-feature we would then just ignore ... or improve the plugin on demand

What do you think ?

I handle this as a side project, since it was inspired by an itch ;)

Conan-Kudo:
@schaefi I like the plugin approach too! The only issue I see so far is that I don't see a way to dynamically inject credentials for it (for example, in CI workflows). There's API server and user, but not password or token.

schaefi:
> @schaefi I like the plugin approach too! The only issue I see so far is that I don't see a way to dynamically inject credentials for it (for example, in CI workflows). There's API server and user, but not password or token.

@Conan-Kudo I currently use python's getpass() method which reads from stdin. In a CI I think a common approach for credentials are "secure" variables which can be used to pass in data from stdin. I though something like

```
$ kiwi-ng ...  < $PWD
```

should work with the CI system to care for not leaking the information in logs or other areas.

I'm certainly willing to implement something different but at the moment I don't know what would be a preferred solution ?

Thanks

schaefi:
@Conan-Kudo sorry what I wrote was wrong. getpass() uses the terminal and does **not** read from stdin.

schaefi:
So I agree we need an additional way to allow credentials handling for CI systems. Suggestions welcome

schaefi:
I have finished a first version of the plugin and released it to pypi. You can try the following:

```
$ sudo pip install kiwi-obs-plugin
```

For a non interactive user experience create the following kiwi config file (e.g opensuse.yml)

```yaml
obs:
  - user:
      - your_obs_user: 'credentials'
```

You can now fetch image descriptions from obs which you can build directly with kiwi on your local host or via boxbuild
For example:

```
$ kiwi-ng --config opensuse.yml image obs \
    --image Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-MicroOS \
    --target-dir /tmp/mytest/obs

$ kiwi-ng system build --description /tmp/mytest/obs ...
```

The current plugin code knows how to deal with git source services and can resolve obs urls into public urls if they exist.
The plugin code also translate the repo-order-prio from obs into prio numbers for the package manager.

Would be great if you can play a bit with it and let me know your thoughts.

Thanks

PR #1711, opened 2021-02-03
Dolly integration

mslacken:
[dolly](https://github.com/aginies/dolly) is small program do distribute files/images in parallel on a cluster. This is done with a tree like structure, sending and getting the files/images can also be done in parallel and so having a minimal performance impact on the master node.
The witch between dolly and curl is done via the download location of the image, so for all methods, except dolly://image.tar.xz, curl is used.

### DIFF
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -354,17 +354,17 @@ function get_remote_image_source_files {
 
     image_uri=$(getarg rd.kiwi.install.image=)
     image_md5_uri=$(
-        echo "${image_uri}" | awk '{ gsub("\\.xz",".md5", $1); print $1 }'
+      echo "${image_uri}" | awk '{ gsub("\\.xz",".md5", $1); gsub("dolly:","tftp:", $1); print $1 }'
     )
     image_initrd_uri=$(
-        echo "${image_uri}" | awk '{ gsub("\\.xz",".initrd", $1); print $1 }'
+      echo "${image_uri}" | awk '{ gsub("\\.xz",".initrd", $1); gsub("dolly:","tftp:", $1); print $1 }'
     )
     image_kernel_uri=$(
-        echo "${image_uri}" | awk '{ gsub("\\.xz",".kernel", $1); print $1 }'
+      echo "${image_uri}" | awk '{ gsub("\\.xz",".kernel", $1); gsub("dolly:","tftp:", $1); print $1 }'
     )
     image_config_uri=$(
         echo "${image_uri}" | \
-        awk '{ gsub("\\.xz",".config.bootoptions", $1); print $1 }'
+        awk '{ gsub("\\.xz",".config.bootoptions", $1); gsub("dolly:","tftp:", $1); print $1 }'
     )
 
     # if we can not access image_md5_uri, maybe network setup
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -353,14 +353,17 @@ function get_remote_image_source_files {
     mkdir -p "${metadata_dir}"
 
     image_uri=$(getarg rd.kiwi.install.image=)
+    # make sure the protocol type is tftp for metadata files. There is no need for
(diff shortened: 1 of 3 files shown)


davidcassany:
@mslacken thanks much for the contribution. However I struggling to find this `dolly` tool out there. The link you posted leads me to a 404 github page. Where can I find this tool, a quick search in google was not clearly leading me to anywhere.

This dracut modules are pretty sensitive code, failures and issues there lead to a non booting system and probably other terrible side effects. So preferably we'd like to avoid including extra stuff in that area. As far as I understood dolly does not require anything else that using the correct scheme in the urls that curl is going to use right? If so, the PR doesn't look bad to me as long as the `dolly` is well maintained stable tool.

mslacken:
@davidcassany I have corrected the [dolly](https://github.com/aginies/dolly) link and I plan to maintain it, as it will be part of openSUSE.

dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
```diff
@@ -354,17 +354,17 @@ function get_remote_image_source_files {
 
     image_uri=$(getarg rd.kiwi.install.image=)
     image_md5_uri=$(
-        echo "${image_uri}" | awk '{ gsub("\\.xz",".md5", $1); print $1 }'
+      echo "${image_uri}" | awk '{ gsub("\\.xz",".md5", $1); gsub("dolly:","tftp:", $1); print $1 }'
```
schaefi, line 357: Can we keep the four space indentation model. I'm getting older and my eyes too :)

If you like to break the long line you can do that by e.g

```bash
    echo "${image_uri}" |\
        awk '{ gsub("\\.xz",".md5", $1); gsub("dolly:","tftp:", $1); print $1 }'
```

But maybe the following comment makes this one obsolete...

dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
```diff
@@ -354,17 +354,17 @@ function get_remote_image_source_files {
...
+      echo "${image_uri}" | awk '{ gsub("\\.xz",".initrd", $1); gsub("dolly:","tftp:", $1); print $1 }'
     )
     image_kernel_uri=$(
-        echo "${image_uri}" | awk '{ gsub("\\.xz",".kernel", $1); print $1 }'
+      echo "${image_uri}" | awk '{ gsub("\\.xz",".kernel", $1); gsub("dolly:","tftp:", $1); print $1 }'
     )
     image_config_uri=$(
         echo "${image_uri}" | \
-        awk '{ gsub("\\.xz",".config.bootoptions", $1); print $1 }'
+        awk '{ gsub("\\.xz",".config.bootoptions", $1); gsub("dolly:","tftp:", $1); print $1 }'
```
schaefi, line 367: If I understand the concept right you want to use ```dolly:``` only for the system image but for the "metadata" we stick with tftp. If so I prefer if the replacement will be done only once and also explained as it's not obvious why it will be done

```bash
# make sure the protocol type is tftp for metadata files. There is no need for
# complex protocol types on small files and for standard PXE boot operations
# only tftp can be guaranteed
image_uri=echo "${image_uri}" | awk '{ gsub("dolly:","tftp:", $1); print $1 }'

...keep the rest of the code untouched
```

schaefi (changes requested):
Thanks much for adding the dolly support. Please bear with me but on the level of the dracut code we are really picky as it's very sensitive code and a small change can cause a huge nightmare :) From this perspective the following comments to discuss

dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
schaefi: don't see that this was resolved

mslacken: Nope, we can't manipulate the `$image_uri` variable as the 'dolly' keyword inside it is used in another function to detect if we have to use dolly or curl.

schaefi: Oh I see, yes we return ```echo "${image_uri}|${image_md5}"``` sorry I overlooked that part.

> but I will fix the spaces and resubmit after the weekend.

Thanks

mslacken:
closing this request and opening a new one, as the git order was messed up

PR #1712, opened 2021-02-07
Refactor RuntimeConfig class

schaefi:
The former implementation read the runtime config file every
time a new RuntimeConfig instance was created. The runtime
config is however static during the runtime of an image build
process and not supposed to change. Therefore the file should
be read in once and any new instance of RuntimConfig should
just use what has been read in at the first invocation. This
commit refactors the RuntimeConfig class to hold an application
global RUNTIME_CONFIG variable and allows to reread the
config on explicit request.


davidcassany (approved):
Looks good to me. Interesting to see how many times we were re-reading it again and again. I like this refactor :+1:

schaefi:
Thanks. I will revert the commit which pins cryptography. I don't know why this was needed but without it pip tried to rebuild the cryptography module from source and fails in the rust compiler. kiwi also has no direct dependency to this module so it was pulled in as dependency from something else. In any case this commit should not be needed. I have delete it and tested locally with a new tox env and it worked. Let's see if it works in the gitlab pipeline too...

PR #1713, opened 2021-02-08
Dolly

mslacken:
added dolly to download the disk image, md5 sums and rest is handled by tftp

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh
@@ -18,6 +18,9 @@ function fetch_file {
     else
         fetch="curl -f ${source_url}"
     fi
+    if _is_dolly "${source_url}";then
+        fetch="${dolly}"
+    fi
     if _is_compressed "${source_url}";then
         fetch="${fetch} | xz -d"
     fi
@@ -72,4 +75,9 @@ function uri_fragment {
 function _is_compressed {
     local source_url=$1
     [[ ${source_url} =~ .xz$ ]]
+}
+
+function _is_dolly {
+  local source_url=$1
+  [[ ${source_url} =~ ^dolly ]]
 }
--- a/dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh
@@ -11,15 +11,21 @@ function fetch_file {
     local fetch_info=/tmp/fetch.info
     local fetch
     local curl_options
+    local dolly_options
     curl_options=$(getarg rd.kiwi.install.pxe.curl_options=)
+    dolly_options=$(getarg rd.kiwi.install.pxe.dolly_options=)
     if [ -n "${curl_options}" ]; then
         curl_options=$(str_replace "${curl_options}" "," " ")
         fetch="curl ${curl_options} -f ${source_url}"
     else
         fetch="curl -f ${source_url}"
     fi
+    if [ -n "${dolly_options}" ]; then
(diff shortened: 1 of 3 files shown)


schaefi (approved):
Looks pretty good and clean to me :+1:

schaefi:
@davidcassany can you have a look again too... this is boot code :-)

dracut/modules.d/99kiwi-lib/module-setup.sh
```diff
@@ -22,7 +22,7 @@ install() {
         e2fsck btrfsck xfs_repair \
         vgs vgchange lvextend lvcreate lvresize pvresize \
         mdadm cryptsetup dialog \
-        pv curl xz \
+        pv curl xz dolly\
```
davidcassany, line 25: I would not do that. The initrd creation will fail if dolly is not installed in the rootfs. This will affect many image descriptions out there unless we add the `dolly` as a dependency to `dracut-kiwi-oem-dump` package and I think this is something we can't effort. `curl` is out there in every single distro and part of the very basic tools each distro provides and likely to be installed by default in most cases. I doubt this is the case for `dolly`.

 `dracut` provides an alternative for this kind of situations:
 
 ```
inst_multiple [-o] <file> [ <file> ...]
           installs multiple binaries and files. If executables are
           specified without a path, dracut will search the path
           PATH=/usr/sbin:/sbin:/usr/bin:/bin for the binary. If the
           option "-o" is given as the first parameter, a missing file
           does not lead to an error.
 ```
 So I'd say this lines should be changed to:
 ```bash
...
     pv curl xz dmsetup
 inst_multiple -o dolly
if [[ "$(uname -m)" =~ s390 ]];then
...
 ```
 
 This way we ensure the initrd creation does not break if dolly is not present in the system. I assume that anyone using the `dolly://` scheme in URIs will be well aware the ` dolly` binaries are required to be part of the rootfs. If someone uses `dolly://` URI without `dolly` binaries will just fail at run time with a command-not-found sort of issue and I bet build logs will also show a warning from dracut for the missing `dolly` command.

davidcassany (changes requested):
I don't think we should add a hard requirement for `dolly` within the initrd.

dracut/modules.d/99kiwi-lib/module-setup.sh
schaefi: This is interesting I thought dracut will not exit with an error if the file is missing. iirc when people configured an image to be used with lvm but did not install lvm2 the dracut call did not fail for the missing lvcreate and friends.

In any case I agree, a missing dolly should not harm the initrd creation process.

Thanks for catching this one

mslacken: Good point. Thanks for bringing this up. Its now optional.

dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh
```diff
@@ -73,3 +82,8 @@ function _is_compressed {
     local source_url=$1
     [[ ${source_url} =~ .xz$ ]]
 }
+
+function _is_dolly {
+  local source_url=$1
```
schaefi, line 87: and four space indentation please :) Thanks

mslacken: I know, I am greedy even with whitespaces...
Fixed.

davidcassany (approved):
Looks good to me now :+1:

dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh
schaefi: :) thanks

Issue #1684, opened 2021-01-13
initrd network is not usable on TW with new network-wicked


schaefi:
## Problem description

When using the kiwi installer for remote deployment:

The network-wicked dracut module (the default in most cases for TW) is not functional, the network interfaces are not ready when our boot code tries to download the disk image.

davidcassany:
The `network-wicked` dracut module is included used when wicked service is part of the image. We have some tests that do no include wicked hence for those the initrd includes the `network-legacy` module. This can be reproduced by building an oem image with `installpxe="true"` based on the tumbleweed example image we have in [OSInside/kiwi-descriptions](https://github.com/OSInside/kiwi-descriptions) repository.

Vogtinator:
I guess you're hitting https://bugzilla.opensuse.org/show_bug.cgi?id=1182227?

JulianBane:
Hi. Not sure if this exactly the same thing but I started a discussion on the openSUSE forum when trying to deploy openSUSE Tumbleweed via PXE. I must admit I am out of my depth as to what exactly is going on but here is a link to the discussion that is quite short and contains all the important facts:

https://forums.opensuse.org/showthread.php/553245-PXE-boot-of-an-ISO-live-distrbution

PR #1715, opened 2021-02-09
Fix ramdisk deployment

schaefi:
**Fixed ramdisk deployment from install ISO**
    
The dracut mount hook is only called if NO root= information is provided from the cmdline. In a PXE deployment this is the case but in an install ISO deployment not. This commit re-adds the sysroot mount generator approach but only applies on the condition rd.kiwi.ramdisk is set and root= is specified. Only in this case we know config.bootoptions is present in the initrd and will not be provided from an outside service


schaefi:
Here is the test building an image using the suggested patch

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:suse/test-image-disk-ramdisk

davidcassany:
@schaefi this is what I had in mind yesterday :+1: Let me run a couple of tests and reread it again, but so far it looks good to me.

build-tests/x86/suse/test-image-disk-ramdisk/appliance.kiwi
```diff
@@ -0,0 +1,75 @@
...
+        <version>1.42.1</version>
+        <packagemanager>zypper</packagemanager>
+        <locale>en_US</locale>
+        <keytable>us</keytable>
+        <timezone>Europe/Berlin</timezone>
+        <rpm-excludedocs>true</rpm-excludedocs>
+        <rpm-check-signatures>false</rpm-check-signatures>
+        <bootsplash-theme>bgrt</bootsplash-theme>
+        <bootloader-theme>openSUSE</bootloader-theme>
+        <type image="oem" filesystem="ext4" firmware="uefi" installiso="true" installboot="install" kernelcmdline="rd.kiwi.ramdisk ramdisk_size=8192000 console=ttyS0">
```
davidcassany, line 19: `ramdisk_size` parameter and `<size>` section they should match up to my understanding. Otherwise we are wasting memory, right? I'd say a perfect match is the only sane setup as OEM ramdisk images do not expand, so creating a disk image that matches the ramdisk size looks like the best approach to me. So I'd go with 

```
<size unit="G">2</size>
```
and `ramdisk_size=2097152`. This i way smaller and easier to test in host that already has some running VMs and some browser consuming insane amount of memory :)

Also I don't think it makes much sense try  to find out some strategy to allow self expansion on ramdisk deployments. I was thinking about it when we coded the mount hook few months ago. I am not sure it is worth the effort, I'd say use cases are damn limited.

build-tests/x86/suse/test-image-disk-ramdisk/appliance.kiwi
```diff
@@ -0,0 +1,75 @@
...
+            <size unit="G">4</size>
```
davidcassany, line 20: Why? I'd prefer having some smaller tests if possible, what about 2G, it will definitely be enough for a JeOS.

schaefi:
@davidcassany good news, my first tests worked :)

dracut/modules.d/90kiwi-dump/kiwi-ramdisk-deployment-generator.sh
```diff
@@ -0,0 +1,55 @@
...
+root_uuid=$(
+    while read -r -d ' ' opt; do echo "${opt}";done < /config.bootoptions |\
+    grep root= | cut -f2- -d=
+)
+
+[ -z "${root_uuid}" ] && exit 1
+
+{
+    echo "[Unit]"
+    echo "Before=initrd-root-fs.target"
```
davidcassany, line 38: I'd say we are missing the `DefaultDependencies=no` here. See this systemd issue here https://github.com/systemd/systemd/issues/12231 which was solved by adding 

```c
/* All mounts under /sysroot need to happen later, at initrd-fs.target time. IOW, it's not
 * technically part of the basic initrd filesystem itself, and so shouldn't inherit the default
 * Before=local-fs.target dependency. */                                
if (in_initrd() && path_startswith(where, "/sysroot"))                  
    fprintf(f, "DefaultDependencies=no\n");
```
into the fstab generator (this generator creates `sysroot.mount`).

We also fixed that a while ago in live generator here 4238b7eac9, but I guess we missed the ramdisk generator back at that time.

davidcassany (changes requested):
We are missing `DefaultDependencies=no` in the generated unit and also I'd prefer to include lighter test, since this is a ramdisk deployment I'd prefer having it with some small disk so we easily boot it locally without having to give up to 8G to the VM running it.

build-tests/x86/suse/test-image-disk-ramdisk/appliance.kiwi
schaefi: > ramdisk_size parameter and <size> section they should match up to my understanding

yes but you can't make them 100% identical because one byte in difference can lead the size checker code to say "not enough space available". Thus ramdisk_size setting should be a bit higher than the image file size. The test here did it absolutely "stable" to avoid non test related problems. But we can go down a bit

schaefi: done now (2G / 2048M image file size) 2192M ramdisk_size .. I don't really care about the ram waste for the purpose of a test, ok ?

build-tests/x86/suse/test-image-disk-ramdisk/appliance.kiwi
schaefi: remember you need some space to work with the image. It's all in ram this means if the image has ~2G of files you don't have anything left for the RAM of the OS. That's why I used 4G of image file size which in fact means 2G are free to work with the system. But for the purpose of the test it's ok to go down. I try with 2G but I think I know why I selected 4G in the end

dracut/modules.d/90kiwi-dump/kiwi-ramdisk-deployment-generator.sh
schaefi: yep good catch, this was never done in the live ramdisk generator. I'll adapt it

schaefi: done

schaefi:
I'm going to rebuild with the smaller file size... I doubt it will be enough to actually work with the system. I vote for at least 3g but let me test first. I had not much trouble to work with a 4G image deployed to ram... do you have any particular reason why you argue with me about the size ? Thanks

schaefi:
probably I'm working for too long with Microsoft and high mem machines 8TB and more :-))

davidcassany:
@schaefi by the way, did you test PXE deployment? I am facing #1684 :facepalm: I just realized I need to adapt the installation initrd to make use of the old legacy module :/

schaefi:
I just tested the the ramdisk from an iso deployment. I thought you might have a quick opportunity to test pxe as you did some changes there lately and maybe have the pxe setup still running on your machine

> I just realized I need to adapt the installation initrd to make use of the old legacy module :/

yes the network stuff is a bad side effect in this testing as well. Do you want me to create the integration test for an older distro ? I think with leap the network problem does not "yet" exist

schaefi:
ok the test works with the smaller sizes as well. Actually you are right, the deployment is quicker with less memory.

I found another issue which I will come up with a new PR. when using console="serial" in bootloader it will not have an effect for install iso media because there we still use the grub template and that setting is not applied

god... bootloaders and boot code

davidcassany:
> yes the network stuff is a bad side effect in this testing as well. Do you want me to create the integration test for an older distro ? I think with leap the network problem does not "yet" exist

No need I just made use of

```
<installmedia>
  <initrd action="add">
    <dracut module="network-legacy"/>
  </initrd>
</installmedia>
```
I really like it, it is quite handy.

The test went fine for PXE and ISO, both are booting TW successfully. Just for the note, the PXE ramdisk deployment of the TW image from kiwi-descriptions weights only 774MB. See `df -h` output:

```
Filesystem          Size  Used Avail Use% Mounted on
devtmpfs            1.9G     0  1.9G   0% /dev
tmpfs               2.0G     0  2.0G   0% /dev/shm
tmpfs               785M   56M  729M   8% /run
tmpfs               4.0M     0  4.0M   0% /sys/fs/cgroup
/dev/mapper/ram1p3  2.0G  774M  1.1G  42% /
tmpfs               2.0G     0  2.0G   0% /tmp
/dev/mapper/ram1p2   20M  1.2M   19M   6% /boot/efi
tmpfs               393M     0  393M   0% /run/user/0
```

2G is way enough and VM had 4G of memory and at runtime right after boot 1.5G of memory are still available. I really like this little JeOS :)

davidcassany:
> I thought you might have a quick opportunity to test pxe as you did some changes there lately and maybe have the pxe setup still running on your machine

Yes I do have such an environment. Finally I managed to have a nice libvirt env for that matter. Probably one day I'll create the server image description in OBS and just try to maintain it there. This is one of the things I'd love to have prebuilt every time I have to prepare the env from scratch it takes me a while.

davidcassany (approved):
Looks good to me.

Just one nitpick comment. I saw the file permissions for kiwi-mount-ramdisk.sh changed to make it executable. I don't see anything bad with it, just wanted to point this is not needed as dracut does not execute the hook, it sources it.

schaefi:
yep I know about the sourcing aspect from hook code in dracut. There is however nothing wrong in calling the script for e.g the purpose of testing. I needed to call it in this particular issue and debugging session. As you already mentioned it does not hurt and I changed perms intentionally I would like to keep it as such.

schaefi:
> No need I just made use of
> 
> ```
> <installmedia>
>   <initrd action="add">
>     <dracut module="network-legacy"/>
>   </initrd>
> </installmedia>
> ```
> 
> I really like it, it is quite handy.

I like this too, that was a good move :smiley:

PR #1716, opened 2021-02-10
Clean some copy and paste leftover from doc strings

davidcassany:


davidcassany:
@schaefi some embarrassing leftovers I did not clean properly when I coded it :/

schaefi (approved):
good catch :+1: I usually make more mistakes like this ;)

PR #1720, opened 2021-02-11
Fixed use of BIOS grub image on ISO media

schaefi:
Not all systems (e.g Debian) creates the boot/grub|grub2 directory.
In kiwi when we need to create a custom grub image because we did
not find the distro provided one, an earlyboot.cfg file is created
and embedded into the grub image. The locaton to store that file
is below boot/grub|grub2. However if the directory does not exist
the build will fail as with the current Debian Live integration
test.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -883,11 +883,14 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             early_boot_script, mbrid
         )
         early_boot_script_on_media = os.sep.join(
-            [self.root_dir, 'boot', self.boot_directory_name]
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
Looks good :+1: I am wondering how it is that it did not pop up before

schaefi:
I think Debian is the only distro that has the directory missing and the :debian integration tests were not part of the build_status helper so it felt off my radar

schaefi:
want to wait for the tests to be successful prior merge

schaefi:
Integration test to watch:

* https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_x86:debian

PR #1722, opened 2021-02-11
Followup fix for RuntimeConfig refactor

schaefi:
The refactor of the RuntimeConfig made sure the runtime config
file is read in only once. But if the file exists and is empty
after yaml.safe_load like in the kiwi package provided
/etc/kiwi.yml which contains all config options as comments,
the code still reads in the file with every new instance of
RuntimeConfig. This commit fixes this condition

### DIFF
--- a/kiwi/runtime_config.py
+++ b/kiwi/runtime_config.py
@@ -50,7 +50,7 @@ class RuntimeConfig:
     def __init__(self, reread=False):
         global RUNTIME_CONFIG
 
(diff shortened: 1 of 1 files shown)


schaefi:
A funny one, guess how long it took to find this out :)

davidcassany (commented):
Wow good catch :)  These sort of errors caused by empty vs none are evil. They are hard to debug.

davidcassany (approved):

PR #1723, opened 2021-02-12

jesusbv:
kiwi/storage/device_provider
kiwi/storage/disk
kiwi/storage/loop_device
kiwi/storage/luks_device
kiwi/storage/mapped_device

This references issue #1644

### DIFF
--- a/kiwi/storage/device_provider.py
+++ b/kiwi/storage/device_provider.py
@@ -37,7 +37,7 @@ class DeviceProvider:
             'No storage device(s) provided'
         )
 
-    def get_uuid(self, device):
+    def get_uuid(self, device: str) -> str:
         """
         UUID of device
 
@@ -52,7 +52,7 @@ class DeviceProvider:
         )
         return uuid_call.output.rstrip('\n')
 
-    def get_byte_size(self, device):
+    def get_byte_size(self, device: str) -> int:
         """
         Size of device in bytes
 
@@ -67,7 +67,7 @@ class DeviceProvider:
         )
         return int(blockdev_call.output.rstrip('\n'))
 
-    def is_loop(self):
+    def is_loop(self) -> bool:
         """
         Check if device provider is loop based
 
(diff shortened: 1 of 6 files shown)


davidcassany (approved):
Looks good to me. I am just wondering if we should review the undefined variables to avoid `Optional` usage. Probably in some places we could actually enforce certain type and make our code more robust. I also think that this can be done later in a separate step so we could eventually better isolate and bisect any potential regression.

jesusbv:
> Looks good to me. I am just wondering if we should review the undefined variables to avoid `Optional` usage. 
Not sure exactly what you mean. To clarify, `Optional` keyword means that type accepts `None`

So if a variable's value is `str` or `None` `Optional` fixes that.

davidcassany:
> > Looks good to me. I am just wondering if we should review the undefined variables to avoid `Optional` usage.
> 
> Not sure exactly what you mean. To clarify, `Optional` keyword means that type accepts `None`
> 
> So if a variable's value is `str` type or `None` `Optional` fixes that.

I was just thinking if it would make sense to modify the code to only accept `str` type and return an empty string instead of None. As it happened in #1722 sometimes the None vs empty difference is tricky. Continuing with the str type example if the return type is `str` (without `Optional`) form an API point of view you can relay on it being a str and blindly apply string methods on it without any sanity check.

Anyway, I am not saying we should change it now, neither this is needed. Just something to consider.

jesusbv:
> I was just thinking if it would make sense to modify the code to only accept `str` type and return an empty string instead of None. 

I got what you meant now.

> As it happened in #1722 sometimes the None vs empty difference is tricky. 

Agreed

> Continuing with the str type example if the return type is `str` (without `Optional`) form an API point of view you can relay on it being a str and blindly apply string methods on it without any sanity check.
> 
> Anyway, I am not saying we should change it now, neither this is needed. Just something to consider.

I agree, it would be nice to rely strictly on the type returned and not to have to do sanity checks. 

For now, I will add the type hinting/enforcement and we can think about/discuss a second refactor on those types

will merge then, if ok.

schaefi (approved):
Looks great thanks :+1:

schaefi:
> will merge then, if ok.

yep :)

PR #1724, opened 2021-02-12
Allow to specify OBS credentials in runtime config

schaefi:
In preparation to the new obs kiwi plugin this commit adds
an opportunity to store obs credentials such that the plugin
could also be used in a non-interactive way

### DIFF
--- a/kiwi/runtime_config.py
+++ b/kiwi/runtime_config.py
@@ -114,6 +114,22 @@ class RuntimeConfig:
         return obs_api_server_url if obs_api_server_url else \
             Defaults.get_obs_api_server_url()
 
+    def get_obs_api_credentials(self):
+        """
+        Return OBS API credentials if configured:
+
+        obs:
+          - user:
+              - user_name: user_credentials
+
+        :return: List of Dicts with credentials per user
+
+        :rtype: list
+        """
+        obs_users = self._get_attribute(element='obs', attribute='user') or []
+        if obs_users:
+            return obs_users
+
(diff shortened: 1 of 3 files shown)


schaefi:
@Conan-Kudo regarding the CI use of the obs plugin in a non-interactive way I created this commit which allows to store obs credentials in a custom runtime config file. In a CI system the credentials could be stored as protected variable and in the CI task this variable could be used to construct the runtime config from which the obs plugin could fetch the information.

Is this an adequate solution or should this be done differently/more secure ... etc ?

Thanks

Conan-Kudo:
I need a way to noninteractively pass in the password too. When there's no terminal at all, I can't do stdin tricks. :(

schaefi:
> I need a way to noninteractively pass in the password too. When there's no terminal at all, I can't do stdin tricks. :(

that's what will become possible with this PR. It allows you to create a file:

```yaml
obs:
   - user:
       - neal: pwd
```

in the obs plugin you would then say

```
kiwi-ng --config kiwi_config_with_obs_credentials.xml image obs ...
```

and you will not be asked anything

Would that work for you ?

Conan-Kudo:
Works for me! As long as both interactive and non-interactive credentials work, I'm good with it!

Conan-Kudo (approved):

schaefi:
> Works for me! As long as both interactive and non-interactive credentials work, I'm good with it!

yes this will be the case. Ok great, thanks

PR #1726, opened 2021-02-13
Fixed packaging metadata for pypi

schaefi:
Include the README as long description in the metadata
for pypi. The change causes the description on pypi.org
to show the ReST rendered README instead of a message
that the author of the module hasn't provided a description


davidcassany (approved):
Good catch :+1:

PR #1729, opened 2021-02-15
Add strong typing for filesystem classes

davidcassany:
This commit strong typing for FileSystem* classes.

Part of #1644

### DIFF
--- a/kiwi/filesystem/base.py
+++ b/kiwi/filesystem/base.py
@@ -18,12 +18,16 @@
 import os
 import logging
 import copy
+from typing import (
+    Dict, List, Optional
+)
 
 # project
 from kiwi.defaults import Defaults
 from kiwi.utils.sync import DataSync
 from kiwi.mount_manager import MountManager
 from kiwi.command import Command
+from kiwi.storage.device_provider import DeviceProvider
 
 from kiwi.exceptions import (
     KiwiFileSystemSyncError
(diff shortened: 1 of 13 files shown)


schaefi (approved):
Looks great :+1: 

and with any of these type settings some inconsistencies are found and fixed, I like it :)

kiwi/filesystem/base.py
```diff
@@ -43,7 +47,10 @@ class FileSystemBase:
     :param string root_dir: root directory path name
     :param dict custom_args: custom filesystem arguments
     """
-    def __init__(self, device_provider, root_dir=None, custom_args=None):
+    def __init__(
+        self, device_provider: DeviceProvider,
+        root_dir: str = None, custom_args: Dict = None
```
jesusbv, line 52: can we set the types of `Dict` ? or they are not known ?

kiwi/filesystem/clicfs.py
```diff
@@ -82,7 +87,7 @@ def create_on_file(self, filename, label=None, exclude=None):
             ['mkclicfs', clicfs_container_filesystem, filename]
         )
 
-    def _get_container_filesystem_size_mbytes(self):
+    def _get_container_filesystem_size_mbytes(self) -> int:
```
jesusbv, line 90: not a correction/nitpicking, question for me:
I understood the methods for the strong typing were the API ones,
are helpers included too ?

jesusbv (approved):
Only the Dict question, other than that, LGTM

kiwi/filesystem/clicfs.py
davidcassany: My bad on this one, I was just adding type blindly to every method. You are right no need to add it here.

kiwi/filesystem/base.py
davidcassany: Well I did not do that because it is not in a very strict format. This dict format should be something like `Dict[str, List or Dict or str]` I did not even check if it is possible to express it. I thought it already enough stating it is a dict.

See there is an example of a custom_args dict in `post_init` method few lines below.

PR #1732, opened 2021-02-15
Fix lsblk flags to get sorted output

davidcassany:
This commit modifies the lsblk command flags to get a sorted output
according to the disk layout.

This is related to 176c7eab commita and it fixes bsc#1182264

### DIFF
--- a/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
@@ -224,7 +224,7 @@ function preparePersistentOverlayDiskBoot {
             return 1
         fi
         local write_partition
-        write_partition=$(lsblk "${isodiskdev}" -p -r -n -o NAME | tail -n1)
+        write_partition=$(lsblk "${isodiskdev}" -p -l -n -o NAME | tail -n1)
         if ! mkfs."${cow_filesystem}" -L cow "${write_partition}"; then
             return 1
         fi
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -255,7 +255,7 @@ function get_free_disk_bytes {
     local part_uuids
     udev_pending
     for part in $(
-        lsblk -p -r -o NAME,TYPE "${disk}" | grep -E "part|md$" | cut -f1 -d ' '
+        lsblk -p -l -o NAME,TYPE "${disk}" | grep -E "part|md$" | cut -f1 -d ' '
     );do
         current_part_uuid=$(get_partition_uuid "${part}")
         for part_uuid in ${part_uuids[*]};do


davidcassany:
@schaefi this fixes bsc#1182264, however I'd like to test it before with the actual reported KDE Live image (up to now I just tested it with our test-image-live) as I could not understand why it managed to run the fallback non persistent boot for the first boot but failed for any later boot. I did not see this behavior on our test-image-live, so probably there is some other detail we might miss. Gonna further debug it tomorrow.

In addition I am wondering if we should review completely our usage of lsblk command, as probably we should stop using the `-r` flag in favor of `-l`. It is a sensitive hard to check change though...

I believe the issue we say few months ago in Fedora just reached TW now. Back at a time I tried to figure out if there were additional lsblk calls that were depending on the order and it seams I missed this one. :disappointed:

schaefi:
> I believe the issue we say few months ago in Fedora just reached TW now.

yes I agree :(

Let's check the code for lsblk and change where needed. I'll will have a look and send you an e-mail with my findings

schaefi:
I've done some research and post it here, rather than mail

```
90kiwi-overlay/kiwi-overlay-root.sh
========================================>
    root_disk=$(
        lsblk -p -n -r -s -o NAME,TYPE "${write_partition}" |\
        grep disk | cut -f1 -d ' '
    )

    This is imho independent of the order as it lists
    the block info of a partition device which looks like
    this:

    lsblk -p -n -r -s -o NAME,TYPE /dev/sda1

    /dev/sda1 part
    /dev/sda disk

    There can only be one disk associated with the partition

    read_only_partition=$(
        lsblk -p -r --fs -o NAME,FSTYPE "${root_disk}" |\
        grep squashfs | cut -f1 -d ' '
    )

    This code depends on the order and will be an issue of there
    are multiple devices with a squashfs filesystem. However this
    case cannot happen with a kiwi built image at the moment.
    If it happens the order will not fix the issue because
    if there is more than one squashfs device we are in trouble
    which one to use


90kiwi-live/kiwi-live-lib.sh
========================================>
    function lookupIsoDiskDevice {
        ...
        for disk in $(lsblk -p -n -r -o NAME,TYPE | grep disk | cut -f1 -d' ');do

    This code depends on the order but the order should not matter
    becaue the devices are checked for their block LABEL.
    There can also be one loopback mounted ISO such that this case
    can only exist once

    function preparePersistentOverlayDiskBoot {
        write_partition=$(lsblk "${isodiskdev}" -p -r -n -o NAME | tail -n1)

    !!! Here we have an order problem !!!

    function _partition_count {
        ...
        lsblk "${isodiskdev}" -p -r -n -o TYPE | grep -c part

    This is independent of the order, it just counts the entries


./99kiwi-lib/kiwi-lib.sh
========================================>
    function lookup_disk_device_from_root {
        ...
        for disk_device in $(
        lsblk -p -n -r -s -o NAME,TYPE "${root_device}" |\
            grep -E "disk|raid" | cut -f1 -d ' '
    ); do

    This is not order dependent from my perspective as it
    reads the block entries for the given root device. The
    device can only have one disk|rais associated with it

    Thus I expect the grep statement to return exactly
    one entry. If this would not be the case we are in
    trouble, but not related to the order imho

./99kiwi-lib/kiwi-partitions-lib.sh
========================================>
    function get_partition_node_name {
        ...
        lsblk -p -l -o NAME,TYPE "${disk}" | grep -E "part|md$" | cut -f1 -d ' '

    !!! This depends on the order and got already fixed by you !!!

    function get_free_disk_bytes {
        ...
        for part in $(
            lsblk -p -r -o NAME,TYPE "${disk}" | grep -E "part|md$" | cut -f1 -d ' '
        );do

    This code walks through the partitions of a disk and assigns them
    a number. It's order dependent and I think has to be fixed.
    It also might be the reason for the persistency problem

    !!! needs fixing !!!
```

schaefi:
I think I identified the following code that needs fixing:

* 99kiwi-lib/kiwi-partitions-lib.sh::get_free_disk_bytes
* 90kiwi-live/kiwi-live-lib.sh::preparePersistentOverlayDiskBoot

parts of it are covered by this PR

davidcassany:
>     * 99kiwi-lib/kiwi-partitions-lib.sh::get_free_disk_bytes

Good catch! I went through this one yesterday and I did not consider it order dependent as I thought i was just checking the UUID for each of them. I did not realize it was also mapping bytes cound to a list of UUIDs using part_count as the index.

Fixed now.

davidcassany:
I tested this PR in this project here https://build.opensuse.org/package/show/home:dcassany:openSUSE-Tumbleweed/livecd-tumbleweed-kde

schaefi (approved):
Looks good :+1: 

Let's hope this has no unexpected after effects ;)

PR #1733, opened 2021-02-19
Deprecate iso source type

### LINKED ISSUE #1678: SUSE product ISOs can't be simply used as iso repo due to module concept
## Problem description
Hello,

I try to build an SLES 15 SP2 appliance based on kiwi description "kiwi-templates-JeOS-15.2-14.8.noarch.rpm".
But it fails.

[ ERROR   ]: 18:23:44 | KiwiBootStrapPhaseFailed: Bootstrap package installation failed: No provider of 'ca-certificates-mozilla' found.
No provider of 'filesystem' found.
No provider of 'glibc-locale-base' found.
No provider of 'sles-release' found.
No provider of 'udev' found.
No provider of 'zypper' found.

Build with "SLE-15-SP2-Full-x86_64-GM-Media1.iso"

<packages type="bootstrap">
    <package name="udev"/>
    <package name="filesystem"/>
    <package name="sles-release"/>
</packages>

<repository type="rpm-dir" alias="dvd-1">
	<source path="iso:///kiwi/iso/SLE-15-SP2-Full-x86_64-GM-Media1.iso"/>
</repository>

ERROR... see Problem description

Build with "SLE-15-SP2-Full-x86_64-OU1-Media1.iso"

<packages type="bootstrap">
    <package name="udev"/>
    <package name="filesystem"/>
    <package name="sles-release"/>
</packages>

<repository type="rpm-dir" alias="dvd-1">
	<source path="iso:///kiwi/iso/SLE-15-SP2-Full-x86_64-OU1-Media1.iso"/>
</repository>

ERROR... see Problem description

Build with "SLE-15-SP2-Full-x86_64-GM-Media1.iso"

<packages type="bootstrap">
    <package name="udev"/>
    <package name="filesystem"/>
    <package name="sles-release"/>
</packages>

<repository type="rpm-dir" alias="dvd-1">
	<source path="iso:///kiwi/iso/SLE-15-SP1-Packages-x86_64-GM-DVD1.iso"/>
</repository>

OKAY

SUSE has changed with SLES 15 SP2 the DVDs. Package DVDs are no more available. They are integrated into Full DVDs.
But kiwi-ng is not working this new Full DVDs. Please fix it.

Or is it possible to use OBS Repos for SLES 15 SP2 with a local kiwi? I found no URL.

+ KIWI version: KIWI (next generation) version 9.22.1
+ OBS version: ---

schaefi:
This patch is two fold

**Document use of SUSE media**
    
Add chapter to describe how to use the SUSE product media in a kiwi build process. This Fixes #1678

**Delete support for generic iso:// source type**
    
The generic iso:// media type mounts the given iso file and expect its root to provide a repository that can be used 1:1 with a package manager. This concept is broken since some time and it can't be fixed in a generic way. All distribution media comes with a certain layout and basically needs extra handling to become fully usable as repository. The current implementation of the iso type which simply mounts the iso and expects its root to be a known repo is not useful. Therefore the support for it will be decommissioned. Instead we will provide a documentation chapter that documents how to incorporate distro ISO media for building images.

When using kiwi with this patch and a user has a repo definition like this

```xml
<repository>
    <source path="iso:///path/to/file.iso"/>
</repository>
```

The following error will be raised

```
[ INFO    ]: 11:35:11 | Setting up repository iso://path/to/file.iso
[ INFO    ]: 11:35:11 | --> Type: None
[ ERROR   ]: 11:35:11 | KiwiUriStyleUnknown: URI schema iso://path/to/file.iso not supported
[ INFO    ]: 11:35:11 | Cleaning up SystemPrepare instance
```

and the build stops. From my perspective the error is good information but we can also try to add some reference to the docs about the use of ISO images. However we currently prevented to use doc links in code as they can also change. I would leave it that way but am eager to get your feedback. Thanks

### DIFF
--- a/doc/source/working_with_images.rst
+++ b/doc/source/working_with_images.rst
@@ -12,19 +12,24 @@ Working with Images
 
    working_with_images/iso_to_usb_stick_deployment
    working_with_images/iso_to_usb_stick_file_based_deployment
+
    working_with_images/disk_setup_for_ec2
    working_with_images/disk_setup_for_azure
    working_with_images/disk_setup_for_google
-   working_with_images/setup_network_bootserver
-   working_with_images/setup_yast_on_first_boot
    working_with_images/disk_setup_for_vagrant
-   working_with_images/network_live_iso_boot
+   working_with_images/disk_setup_for_luks
    working_with_images/disk_ramdisk_deployment
+
    working_with_images/custom_partitions
    working_with_images/custom_volumes
-   working_with_images/custom_fstab_extension
-   working_with_images/disk_setup_for_luks
-   working_with_images/build_with_profiles
-   working_with_images/build_in_buildservice
+
+   working_with_images/setup_network_bootserver
    working_with_images/legacy_netboot_root_filesystem
    working_with_images/network_overlay_boot
+   working_with_images/network_live_iso_boot
+
+   working_with_images/setup_yast_on_first_boot
+   working_with_images/custom_fstab_extension
+   working_with_images/build_with_profiles
+   working_with_images/build_in_buildservice
+   working_with_images/use_suse_media
--- a/doc/source/working_with_images/use_suse_media.rst
+++ b/doc/source/working_with_images/use_suse_media.rst
@@ -1,0 +1,59 @@
+.. _suse_media:
+
+Using SUSE Product ISO To Build
+===============================
+
+.. sidebar:: Abstract
+
+    This page provides information how to use the SUSE
+    media ISO with {kiwi}
+
+When building an image with {kiwi}, the image description usually
+points to a number of public/private package source repositories
+from which the new image root tree will be created. Alternatively
+the vendor provided product ISO image(s) can be used. The contents
+of the ISO (DVD) media also provides package source repositories
+but organized in a vendor specific structure. As a user it's
+important to know about this structure such that the {kiwi}
+image description can consume it.
+
+To use a SUSE product media the following steps are required:
+
+1. Mount the ISO media from file or DVD drive:
+
+.. code:: bash
+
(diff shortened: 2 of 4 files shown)


Conan-Kudo:
What is the `dir://` protocol?

schaefi:
> What is the `dir://` protocol?

It just says, "use the following as a directory". In theory this could also work without a mime type but as we don't want to play the guessing game like in the old kiwi it needs to be made specific

schaefi:
btw, dir:// is not new, it always existed in kiwi

Conan-Kudo:
`dir://` is not a protocol supported by most package managers, there's `file://` since that's a real protocol, but not `dir://`.

schaefi:
> `dir://` is not a protocol supported by most package managers, there's `file://` since that's a real protocol, but not `dir://`.

This information is not passed to the package manager. The code has this:

```
elif uri.scheme == 'dir':
            return self._local_path(uri.path)
```

and that returns the local abs path of the given source path. So this is not a problem for package managers

davidcassany:
> `dir://` is not a protocol supported by most package managers, there's `file://` since that's a real protocol, but not `dir://`.

In fact, after a quick google search, it looks like the `dir://` schema is not an standard URI schema, however `file://` is. I'd say KIWI makes not difference between `dir://` and `file://`, as in both cases the uri is translated to a local path. So the good thing is that I am 90% confident that the exact same setup works fine with both `file://` and `dir://`. 

Probably it is worth documenting `file://` instead of `dir://` and keep `dir://` around for compatibility reasons and because, up to my understanding, it doesn't hurt.

Beyond this little `dir://` vs `file://` detail it all looks good to me :+1:

davidcassany (approved):

schaefi:
agreed, for the sake of compatibility I'll change the mime type

PR #1735, opened 2021-02-21
Added deb support to sat solver module

schaefi:
Add support to create SAT solvables from Debian repos via
deb2solv tool from libsolv. This allows image info --resolve-package-list
to work with Debian/Ubuntu image descriptions. Please note
by default libsolv is not compiled with support for Debian.
Therefore the following compile flags must be set on libsolv

* -DENABLE_DEBIAN=1
* -DMULTI_SEMANTICS=1

If libsolv does not provide the needed capabilities kiwi will
fail on either the repository solvable creation due to missing
deb2solv or on call of setdisttype() due to missing MULTI_SEMANTICS
feature in libsolv

### DIFF
--- a/kiwi/solver/repository/deb.py
+++ b/kiwi/solver/repository/deb.py
@@ -1,0 +1,38 @@
+# Copyright (c) 2015 SUSE Linux GmbH.  All rights reserved.
+#
+# This file is part of kiwi.
+#
+# kiwi is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# kiwi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kiwi.  If not, see <http://www.gnu.org/licenses/>
+#
+# project
+from kiwi.solver.repository.base import SolverRepositoryBase
+
+
+class SolverRepositoryDeb(SolverRepositoryBase):
+    """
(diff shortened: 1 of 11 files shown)


schaefi:
@davidcassany I have done the solver support for debian in preparation of the package list reducer support which will come in the future. I found a glitch with our ```<packagemanager>apt-get</packagemanager>``` setting were I think this should be changed to just ```apt```. The reason is that I can't find an apt-get package in Debian/Ubuntu. apt-get is just the tool name and the package is named ```apt```. This was causing a problem on the resolver because the repo metadata has no knowledge about a package named apt-get. In kiwi we use the packagemanager setting to extend the packages list by that name. For apt-get this fails and so I had to add a hack in here.

When building images we don't see this issue because we let debootstrap to the work and we have this code
in kiwi/package_manager/apt.py

```python
if 'apt-get' in self.package_requests:
            # debootstrap takes care to install apt-get
            self.package_requests.remove('apt-get')
```

I suggest we change the package manager to be just ```apt``` including a schema change and code change to cleanup the situation as a follow up to this PR

davidcassany:
> I suggest we change the package manager to be just `apt` including a schema change and code change to cleanup the situation as a follow up to this PR

Yes, makes sense to me, just use `apt` so it maps the actual package as it is the case for all others.

davidcassany (approved):
Looks good :+1:

kiwi/solver/repository/deb.py
```diff
@@ -0,0 +1,38 @@
...
+from kiwi.solver.repository.base import SolverRepositoryBase
+
+
+class SolverRepositoryDeb(SolverRepositoryBase):
+    """
+    **Class for SAT solvable creation for apt-deb type repositories.**
+    """
+    def _setup_repository_metadata(self):
+        """
+        Download repo metadata for rpm-md specific repositories
```
Conan-Kudo, line 28: This is wrong, since this running against `apt-deb` repositories, not `rpm-md` ones.

schaefi: yes my bad, fixed it. Thanks

PR #1739, opened 2021-02-23

schaefi:
* storage/subformat/vagrant_base.py
* storage/subformat/vagrant_libvirt.py
* storage/subformat/vagrant_virtualbox.py
* storage/subformat/vdi.py
* storage/subformat/vhd.py
* storage/subformat/vhdfixed.py
* storage/subformat/vhdx.py
* storage/subformat/vmdk.py

This references Issue #1644

### DIFF
--- a/kiwi/storage/subformat/vagrant_base.py
+++ b/kiwi/storage/subformat/vagrant_base.py
@@ -19,7 +19,12 @@ import json
 import os.path
 
 from tempfile import mkdtemp
+from typing import (
+    Dict, Optional, List
+)
 
+from kiwi import xml_parse
+from kiwi.system.result import Result
 from kiwi.storage.subformat.base import DiskFormatBase
 from kiwi.storage.subformat.template.vagrant_config import VagrantConfigTemplate
 from kiwi.command import Command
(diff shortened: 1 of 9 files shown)


kiwi/storage/subformat/vagrant_base.py
```diff
@@ -222,7 +220,7 @@ def get_additional_vagrant_config_settings(self):
         :return: additional vagrant settings
         :rtype: str
         """
-        pass
+        return ''
 
     def _create_box_metadata(self):
         metadata = self.get_additional_metadata() or {}
```
jesusbv, line 226: unnecessary `or {}` ? It returns an empty dict now

schaefi: yes any implementation of the function in a child class returns a dict. The abstract interface class however returned None. So the type is broken. The same for the other interface class which is expected to return a string

schaefi: see implementations of vagrant_libvirt.py and vagrant_virtualbox.py for details

schaefi: pythons ```if not``` operator will continue to work in the same way as with the None type before. Thus there was no behavior change and the base class methods are not called in the context of vagrant anyway

schaefi: That's why I thought the change is ok and imho the only way to make static type checks happy. An alternative would be ```->  Optional[...]``` but did not like it that much in this case. Thoughts ?

davidcassany: I am fine with this current implementation. I also prefer avoiding `Optional[...]`

davidcassany (approved):

PR #1745, opened 2021-02-26
Refactor disk and install builders and dracut boot image

### LINKED ISSUE #873: Refactor the use of BootImage for installation initrds

davidcassany:
This commit mainly refactors the DiskBuilder class. Then install builder
and dracut boot image refactors are mostly to drop no longer needed
code. The refactors tries to make the build simpler and easier to read and
also prevent the reuse of the BootImage instance used for the main image
when creating the installation media (this aplies only for
dracut based images).

Fixes #873


davidcassany:
This is my take to solve #873 and rephrase the DiskBuild creation process. These are sensible changes, that's why I added the discussion label. Before eventually merging anything from this PR we should properly discuss if this makes sense and properly test it, specially for images using custom initrds.

Initially I just wanted to make suggestion about what I'd change rather than actually coding, but I realized I could not understand the former code, or at least, I could not understand why it was coded the way it was. So my only argument was reduced to: _I believe it can still work with less and simpler code_. So I had to actually give it a try.

I am pretty sure there was a good reasoning behind and I could not see it, that makes me feel very cautious regarding this PR. Probably this setup was no longer needed with the current kiwi-ng status? :man_shrugging: All in all I opted to code it as my initial hacky tests were promising.

kiwi/builder/disk.py
```diff
@@ -175,33 +173,15 @@ def create(self):
         Image types which triggers this builder are:
 
         * image="oem"
-        * image="vmx"
 
         :return: result
 
         :rtype: instance of :class:`Result`
         """
```
davidcassany, line 180: This create method implementation is the root cause of all this refactor. I can't really understand why extra instances were required...

kiwi/builder/disk.py
```diff
@@ -554,9 +527,13 @@ def create_install_media(self, result_instance):
...
         if self.install_media:
+            if self.initrd_system == 'dracut':
+                boot_image = None
+            else:
+                boot_image = self.boot_image
```
davidcassany, line 533: Reusing the initrd instance for kiwi custom boots makes sense as otherwise it would make the image build way longer and rebuild the same (or almost the same) initrd twice. Given the initrd creation using KIWI is suboptimal all makes sense here.

schaefi: I think I can explain the reasons that lead to this concept. In the boot image initrd ```self.boot_image``` there needs to be information which is created before the actual initrd creation process as part of the disk builder process. This information is persistently stored in self.boot_image such that it's not required to add it when the install media is created. So the reason when we need this is in builder/disk.py here

```python
# prepare for install media if requested
if self.install_media:
    log.info('Saving boot image instance to file')
    self.boot_image.dump(
        self.target_dir + '/boot_image.pickledump'
    )
```

Information that is placed in self.boot_image which we potentially use later is this:

```
if self.boot_is_crypto:
    self.boot_image.include_file(...)

...

if self.initrd_system == 'dracut':
    if self.root_filesystem_is_multipath is False:
        self.boot_image.include_module('kiwi-overlay')
        self.boot_image.write_system_config_file(
            config={'modules': ['kiwi-overlay']}
        )
    if self.disk_resize_requested:
        self.boot_image.include_module('kiwi-repart')

...

def _write_partition_id_config_to_boot_image(self):
    ...
    self.boot_image.include_file(...)

def _write_raid_config_to_boot_image(self):
    ...
    self.boot_image.include_file(...)

def _write_crypttab_to_system_image(self):
    ...
    self.boot_image.include_file(...)

def _write_generic_fstab_to_boot_image(self, device_map):
    ...
    if self.initrd_system == 'kiwi':
        self._write_generic_fstab(device_map, self.boot_image.setup)

# this one can be considered obsolete...
def _write_recovery_metadata_to_boot_image(self):
    ...
    self.boot_image.include_file()
```

So overall if we wouldn't have the need to preserve data for the install initrd creation process none of the pickle()
code would be needed.

I have not yet looked at your refactor but I hope this explains the reasons why I implemented it that way ?

Actually a better implementation allows to transport this information from the disk builder instance to the
BootImage instance such that there is a loose coupling. I implemented that coupling by creating the BootImage instance first and serialize it via object pickle such that I can re-create that instance at any later point in time. I admit this was not the best idea ;) and kiwi also has changed a lot since that time

Does this give you a better idea why it was done this way ?
If yes and you look at your refactor code, do you still think it does not create regressions on install media creation, for example on crypto images to include /etc/crypttab in the install image initrd ?

Thanks much for this effort

schaefi (changes requested):
Ok, I've reviewed the code and from my perspective it confirms my fear that some information in install initrd will now be missing. Depending on the setup of the system image the install initrd can work but under certain conditions it will miss data.

I tried to summarize which data this is and under which conditions I now think we would create regressions with this refactoring in my former comment.

Let me know if this makes sense to you too.

From my perspective there is data collected in the system image build (builder/disk.py) which is also relevant inside of the install initrd. Therefore we need a way to preserve/handover this information when install image is built. The refactoring code here just deletes all of this concept and I think this is causing regressions

schaefi:
Building an image with

```xml
<type image="oem" ... luks=linux installiso="true"/>
```

will show issues with this refactor due to the missing /etc/crypttab in the initrd of the installation media

davidcassany:
> Ok, I've reviewed the code and from my perspective it confirms my fear that some information in install initrd will now be missing. Depending on the setup of the system image the install initrd can work but under certain conditions it will miss data.

Agree, however I believe it might be way less data than you probably though at first. Probably we can even narrow the data that gets lost. Note this only apply for dracut builds, which are way easier to trace.

>I tried to summarize which data this is and under which conditions I now think we would create regressions with this refactoring in my former comment.

Yes adding my comments in line in each of those places:

```python
if self.boot_is_crypto:
    self.boot_image.include_file(...)
```
Yes this part requires attention and checking how the change affects it. However note that the former code only included the file into the installation initrd in case there was an `include_file` method called with `install_media=True`. Which was not the case here. Also note that installation initrd does not necessarily need to know about crypto, since there will be `kexec` call to the new image after dump. There might issues for ramdisk deployments, in which case I bet some of those we already face them now with the current implementation. My refactor was based on top of `install_media=True` parameter usage, my first surprise was that it was never used inside the disk builder class.

```python
if self.initrd_system == 'dracut':
    if self.root_filesystem_is_multipath is False:
        self.boot_image.include_module('kiwi-overlay')
        self.boot_image.write_system_config_file(
            config={'modules': ['kiwi-overlay']}
        )
    if self.disk_resize_requested:
        self.boot_image.include_module('kiwi-repart')
```
I don't think this is problematic at all. Modules for dracut were already separated, BootImage was holding separate independent lists of modules for the install initrd and for the image initrd.


```python
def _write_partition_id_config_to_boot_image(self):
    ...
    self.boot_image.include_file(...)
```
This one I don't think is problematic. The partition id config I think is not part of the install initrd, isn't it. Repartition happens on kexec after dump. And no repartition is happening for ramdisk deployaments, it doesn't make sense there.

```python
def _write_raid_config_to_boot_image(self):
    ...
    self.boot_image.include_file(...)

def _write_crypttab_to_system_image(self):
    ...
    self.boot_image.include_file(...)
```
Agreed we need to pay attention to this use cases, they might be tricky. However once again, this `include_file` here was never part of the initrd, if we check dracut creation code for install initrds only files included with `install_media=True` are considered. This is not the case here.


```python
def _write_generic_fstab_to_boot_image(self, device_map):
    ...
    if self.initrd_system == 'kiwi':
        self._write_generic_fstab(device_map, self.boot_image.setup)
```
This one should not be affected, since the bootimage instance is kept alive for the kiwi custom initrds and the image intird is reused for the installation media.

While this checks are all valid, I believe they should be easy to sort them out or even they can be ignored since they are not affecting the installation initrd in any way.

My real concern about all this refactor are the changes applied on `builder/disk.create` method. There is where I can't really understand the former implementation, why a new instance of was needed Disk?

https://github.com/OSInside/kiwi/blob/cf936b3063abd0084048ca110845c2175e9bed7b/kiwi/builder/disk.py#L169-L204

By the way I created a reference project for testing in my home [here](https://build.suse.de/package/show/home:dcassany:refactor_image_install/sle_15_image) There is  a build for dracut and a build for custom kiwi boot. Both based on SLE_15_SP2

davidcassany:
The test image based on SLE15_SP2 with dracut and encrypted disk seams to work out of the box as I expected. The installation media does not require the crypttab or any luks related that, it just dumps and kexec. It remains to be test the ramdisk deployment for which I bet it will fail, with or without this refactor.

For the custom kiwi based initrd the image is failing to install, there is checksum failure... I really did not expect this sort of issues, so I'll give it a quick look, but I am afraid I am seeing some unrelated failure :S

schaefi:
I agree with you since we introduced kexec to distinguish between install and first boot, most of the issues I had mind should no longer be a problem.

kiwi/builder/disk.py
```diff
@@ -175,33 +173,15 @@ def create(self):
...
 
         :rtype: instance of :class:`Result`
         """
-        disk = DiskBuilder(
-            self.xml_state, self.target_dir, self.root_dir, self.custom_args
-        )
-        result = disk.create_disk()
-
-        # cleanup disk resources taken prior to next steps
-        del disk
```
schaefi, line 190: This deletion of the disk instance needs to stay or some resource free must be issued. The reason is that when creating the install media or the disk format we should prevent open loop/mount setup into the system image at call time of these operations. There is a lot potential for strange effects if a loop is done over a loop or a squashfs filesystem is created over an active mounted filesystem. I know you might not have seen bad effects in your tests but this cleanup is there for a reason of past issues we had in that regard. I suggest to keep it

schaefi:
> My real concern about all this refactor are the changes applied on builder/disk.create

I hope my last comment explains this part a bit

schaefi:
> It remains to be test the ramdisk deployment for which I bet it will fail, with or without this refactor.

I think if ramdisk deployment works or not highly depends on the used features. As you know for ramdisk deployment we can't kexec therefore the same initrd that dumps it also has to drive the boot and that means that one initrd has to hold ALL required information. I agree with you that with this refactor the situation will become worse but it's probably already not ideal. So yes ramdisk deployment has its limitations and I can live with that. I would say we should have a simple candidate ```./test-image-disk-ramdisk/appliance.kiwi``` which should persist to continue, but we should not spend more time on supporting other features (crypto, lvm, etc... ) with ramdisk deployments. Would that work for you ?

kiwi/builder/disk.py
davidcassany: Thanks for the comments, I realized there are a couple of issues with this refactor. Indeed, DiskBuiler creates some runtime structures that are never cleared (`LoopDevice` and `Disk`) unless the whole instance is cleared. I'll do another round of this refactor, however I am afraid this leads to bigger refactor, not in terms the amount of code, but in terms of sensitive changes.

davidcassany:
> I would say we should have a simple candidate `./test-image-disk-ramdisk/appliance.kiwi` which should persist to continue, but we should not spend more time on supporting other features (crypto, lvm, etc... ) with ramdisk deployments. Would that work for you ?

Yes that absolutely works for me.

kiwi/builder/disk.py
```diff
@@ -225,6 +186,24 @@ def create_disk(self) -> Result:
...
+        system_boot: Optional[FileSystemBase] = None
+
+        # an instance of a class with the sync_data capability
+        # representing the boot/efi area of the disk
+        system_efi: Optional[FileSystemBase] = None
+
+        # an instance of a class with the sync_data capability
+        # representing the spare_part_mountpoint area of the disk
+        system_spare: Optional[FileSystemBase] = None
+
```
davidcassany, line 206: They are only used within the scope of `create_disk` method, thus having them here not as object attributes is a way to ensure they are cleared once `create_disk` is done.

The counter part is having to pass them all to the utility methods as parameters.

davidcassany:
I updated the tests I have in IBS to run dracut based images and custom initrd images to build luks based OEM and plain simple OEM images. In addition I added an extra repository that runs the exact same 4 images using the upstream KIWI (without this refactor) from `Virtualization:Appliances:Builder`, so we can easily compare behavior in case of failures.

Builds are [here](https://build.suse.de/package/show/home:dcassany:refactor_image_install/sle_15_image)

kiwi/builder/disk.py
```diff
@@ -184,35 +163,17 @@ def create(self) -> Result:
...
-            self.xml_state, self.target_dir, self.root_dir
-        )
-
-        disk_format.append_unpartitioned_space()
-        result = disk_format.create_disk_format(result)
-
-        return result
-
-    def create_disk(self) -> Result:
+    def create_disk(self):
```
schaefi, line 176: It still returns Result, doesn't it ?

kiwi/builder/disk.py
schaefi: ok, makes sense to me, even with the enforcement to add them explicitly

kiwi/builder/disk.py
davidcassany: By bad, I took the changes from my branch on this method while solving the conflicts that raised after adding types master and I did not see that I should have picked the method signature. Thanks for catching it

schaefi:
@davidcassany I've looked over the latest changes again. There are quite some changes but this was expected with the last conversation we had. From your mind, do you think you finished the refactor or is there something you still want to change. As far as I can see you touched all of it. We also agreed on the limitations when using custom boot images. So can you give me a short status update on this one ?

Thanks much

davidcassany:
> @davidcassany I've looked over the latest changes again. There are quite some changes but this was expected with the last conversation we had. From your mind, do you think you finished the refactor or is there something you still want to change. As far as I can see you touched all of it. We also agreed on the limitations when using custom boot images. So can you give me a short status update on this one ?
 
Yes, the refactor is finished form my POV. This is all I had in mind. Testing is now what is left. I did few tests in IBS against SLE 15 as I mentioned in https://github.com/OSInside/kiwi/pull/1745#issuecomment-819370501 

I believe before merging we could probably push it to staging and check oem boots for other distros too. I you agree with that I can rebase, fix the typing detail you saw, push it to staging and run oem boots for all distros. Then, if all goes well I'd say we are ready to merge.

schaefi (approved):
Sounds like a great plan to me :+1:

schaefi:
Done, kiwi with this patch is building in Staging. Let's see what breaks ;)

schaefi:
I'll set the "don't merge" label until we have clarity if the integration tests work, my special interest is on:

* test-image-disk-legacy

davidcassany:
Right, I successfully boot the disks (or the install iso if any) from:

TW:
* test-image-disk-legacy (interactive test!)

Leap:
* test-image-luks
* test-image-disk
* test-image-disk-simple

Fedora:
* test-image-live-disk (the installation iso requires interaction on grub, installation option is not default!)

Ubuntu:
* test-image-live-disk (the installation iso requires interaction on grub, installation option is not default!)

Centos v7:
* test-image-live-disk (the installation iso requires interaction on grub, installation option is not default!)


However it failed for CentOS v8:

* test-image-live-disk-v8 (also requires interaction at boot): it gets stalled dumping the image to disk reaches 100% and stays there forever.

schaefi:
>test-image-live-disk-v8 (also requires interaction at boot): it gets stalled dumping the image to disk reaches 100% and stays there forever.

I'll give this a test after the rebuild of the updated integration tests

PR #1748, opened 2021-03-03
Refactor grub2 installation

davidcassany:
This commit refactors grub2 installation method to split it in two
parts. Former grub2.install method was meant to run the grub2-install
tool, however, in addition it was also running the secure boot
installation shim-install. The install method in KIWI is skipped for
those architectures and firmware combinations for which bios support
doesn't exist. This was leading to skip the secure boot installation.

The current approach strips the secure boot installation logic from the
grub2.install method, so skipping the install method does not
automatically result in skipping the secure boot installation.

Fixes bsc#1182211

### DIFF
--- a/kiwi/bootloader/install/base.py
+++ b/kiwi/bootloader/install/base.py
@@ -58,3 +58,11 @@ class BootLoaderInstallBase:
         Implementation in specialized bootloader install class required
         """
         raise NotImplementedError
+
+    def secure_boot_install(self):
+        """
+        Run shim-install in self.device for secure boots
+
+        Implementation in specialized bootloader install class required
+        """
+        raise NotImplementedError
--- a/kiwi/bootloader/install/grub2.py
+++ b/kiwi/bootloader/install/grub2.py
@@ -119,7 +119,7 @@ class BootLoaderInstallGrub2(BootLoaderInstallBase):
             return False
         return True
 
-    def install(self):  # noqa: C901
+    def install(self):
         """
         Install bootloader on disk device
         """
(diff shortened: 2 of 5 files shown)


davidcassany:
@schaefi this is a sensitive change. Despite the refactor and many line changes the logic mostly remains the same, it is more a reorder and restructure logic sections rather than new coding.

Apparently we were supporting pure EFI grub setup for  `arm` but not UEFI. Apparently now there is support for secure boot in `arm` with the EDK2 firmware, so now I am waiting for feedback to verify this actually solves the issue.

There is a testing build in IBS in my home see `home:dcassany:armjeos`

Conan-Kudo:
We should make sure the Fedora tests are extended to cover this case, since Fedora 34+ uses UEFI for 32-bit ARM now: https://fedoraproject.org/wiki/Changes/uEFIforARMv7

Conan-Kudo:
cc: @pbrobinson

Conan-Kudo:
cc: @pbrobinson

Conan-Kudo:
cc: @nullr0ute

Conan-Kudo:
cc: @martinezjavier

Vogtinator:
> We should make sure the Fedora tests are extended to cover this case, since Fedora 34+ uses UEFI for 32-bit ARM now: https://fedoraproject.org/wiki/Changes/uEFIforARMv7

So did openSUSE for basically ever, both on 32bit and 64bit ARM. The difference is that now for 64bit ARM there's secure boot (which kiwi calls "uefi" for some reason). It seems like this  PR also enables `firmware="uefi"` for 32bit ARM, which doesn't exist, but also shouldn't hurt.

davidcassany:
> It seems like this PR also enables `firmware="uefi"` for 32bit ARM, which doesn't exist, but also shouldn't hurt.

Yes, you are right, this is something to improve.

davidcassany:
> We should make sure the Fedora tests are extended to cover this case, since Fedora 34+ uses UEFI for 32-bit ARM now:

Ok, I don't think we have a 34 test yet. Probably we can give a quick test on RawHide and check the status, however I'd say to tackle the Fedora test update in a separate issue, since we might not be capable to set it until 34 is out.

schaefi (approved):
This looks great to me :+1: nice refactor

Conan-Kudo:
> > We should make sure the Fedora tests are extended to cover this case, since Fedora 34+ uses UEFI for 32-bit ARM now: https://fedoraproject.org/wiki/Changes/uEFIforARMv7
> 
> So did openSUSE for basically ever, both on 32bit and 64bit ARM. The difference is that now for 64bit ARM there's secure boot (which kiwi calls "uefi" for some reason). It seems like this PR also enables `firmware="uefi"` for 32bit ARM, which doesn't exist, but also shouldn't hurt.

Yes, but Fedora and openSUSE do GRUB differently, and that's bitten us enough times that I try to remind @schaefi and @davidcassany to test for it to make sure things actually work for both. 😉

Conan-Kudo:
> > We should make sure the Fedora tests are extended to cover this case, since Fedora 34+ uses UEFI for 32-bit ARM now:
> 
> Ok, I don't think we have a 34 test yet. Probably we can give a quick test on RawHide and check the status, however I'd say to tackle the Fedora test update in a separate issue, since we might not be capable to set it until 34 is out.

@davidcassany It's in Rawhide now too, so you can test with that. 😉

davidcassany:
> @davidcassany It's in Rawhide now too, so you can test with that. wink

Awesome, I am working on a followup PR adding integration tests for secure boot in arm for Leap, TW and Rawhide.

PR #1752, opened 2021-03-05
Make sure man pages are part of sdist tarball

### LINKED ISSUE #1746: man pages missing from latest PyPI release of kiwi
When trying to update to kiwi 9.23.16 in Fedora, I noticed that the documentation files are missing from the tarball uploaded to PyPI. This causes the package build to fail because those files are missing so the Makefile doesn't work.

schaefi:
The current tarball when uploaded to pypi via gitlab does
not contain the manual pages because the doc target to build
them is not called. This commit adds a doc_man tox target
which is called prior pypi release. This Fixes #1746

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -174,7 +174,7 @@ pypi_upload:
   script:
     - base64 --decode "$PYPI_CONFIG" > .pypirc
     - export PYTHON=python3.6
-    - tox -e release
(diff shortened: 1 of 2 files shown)


schaefi:
@Conan-Kudo this should bring back man pages to the sdist tarball

Conan-Kudo (approved):

PR #1756, opened 2021-03-06
Fixed API vs. CLI inconsistency

### LINKED ISSUE #1755: API vs. Cli conflict
## Problem description

Pretty bad bug. The feature that allows to read a config file from the command line option --config-file imports
```from kiwi.cli import Cli``` and create an object of ```cli = Cli()``` in the RuntimeConfig constructor. This causes
any code that is using kiwi as an api to fail because the docopt call in Cli() expects a valid docstring which only
exists in cli.py

## Expected behaviour

API use must be independent from command line interface

schaefi:
when using kiwi as API the program fails with a usage message
from the Cli class. The kiwi.cli module should not be imported
except for kiwi comandline tasks. It has turned out that the
RuntimeConfig class which is used in several places in different
API classes imports Cli and creates an instance of it to check
for a global option. This causes major issues for all programs
which uses the kiwi API but not the command line interface because
the docopt call in Cli() expects a valid docstring which only
exists in kiwi's cli.py. This commit fixes the inconsistency
and allows people to use the kiwi API independent of any
command line interface. Fixes #1755

### DIFF
--- a/kiwi/cli.py
+++ b/kiwi/cli.py
@@ -218,6 +218,8 @@ class Cli:
                     value = os.sep + Defaults.get_shared_cache_location()
                 if arg == '--shared-cache-dir' and value:
                     Defaults.set_shared_cache_location(value)
+                if arg == '--config' and value:
+                    Defaults.set_custom_runtime_config_file(value)
                 result[arg] = value
         return result
 
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -39,6 +39,7 @@ EDIT_BOOT_INSTALL_SCRIPT = 'edit_boot_install.sh'
 IMAGE_METADATA_DIR = 'image'
 ROOT_VOLUME_NAME = 'LVRoot'
 SHARED_CACHE_DIR = '/var/cache/kiwi'
+CUSTOM_RUNTIME_CONFIG_FILE = None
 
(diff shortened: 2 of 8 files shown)


schaefi:
@davidcassany similar problem with cli we had last time with the shared cache dir stuff. It was again me who introduced the trouble :/

davidcassany:
@schaefi  Oh, these are easy to oversee and hard to find afterwards. Just out of curiosity, did you face it while using the API?

Just wondering about KIWI's API use cases out there :)

davidcassany (approved):

schaefi:
> @schaefi Oh, these are easy to oversee and hard to find afterwards. Just out of curiosity, did you face it while using the API?
> 
> Just wondering about KIWI's API use cases out there :)

yes I found it when using the API. I was in the DNF meeting last Friday and someone had the issue how to detect if something changes in the dependency resolution according to a change in the repo metadata. I suggested it would be relatively easy (just a couple lines of code) to use the kiwi solver api to perform this job and send an e-mail with an example. Guess what on that example I found the API issue :) As this was in the DNF meeting the most obvious solution is to use the libdnf API but it guess this will be a bit more work. Anyway the conversation helped to find this bug :)

Thanks for the review

PR #1758, opened 2021-03-09
Fix use of by-label device persistency in grub

### LINKED ISSUE #1757: devicepersistency="by-label" not working in grub configuration
## Problem description
devicepersistency="by-label" is set.
As expected, all entries of /etc/fstab are set correctly.
But the settings for Grub in /boot/grub2/grub.cfg still refer to a device via UUID in the kernel command line.

excerpt from fstab: /etc/fstab:
```
LABEL=ROOT / btrfs defaults 0 0
LABEL=EFI /boot/efi vfat defaults 0 0
```
and so on...

exverpt from /boot/grub2/grub.cfg:
```
linux /boot/vmlinuz-5.3.18-22-defaut root=UUID=fd49f3.....
```

## Expected behaviour
excerpt from /boot/grub2/grub.cfg:
```
linux /boot/vmlinuz-5.3.18-22-defaut root=LABEL=ROOT.
```

excerpt from kiwi descriptionfile:
```
  <preferences profiles="VMware">
    <version>15.2</version>
    <bootsplash-theme>SLE</bootsplash-theme>
    <bootloader-theme>SLE</bootloader-theme>
    <rpm-excludedocs>false</rpm-excludedocs>
    <keytable>de</keytable>
    <locale>en_US</locale>
    <timezone>Europe/Berlin</timezone>
    <type
      image="vmx"
      filesystem="btrfs"
      format="vmdk"
      bootloader="grub2"
      firmware="uefi"
      efipartsize="33"
      kernelcmdline="plymouth.enable=0 console=ttyS0,115200 console=tty0 consoleblank=0"
      bootpartition="false"
      bootkernel="custom"
      devicepersistency="by-label"
      rootfs_label="ROOT"
      btrfs_quota_groups="true"
      btrfs_root_is_snapshot="true"
    >
      <systemdisk>
        <volume name="home"/>
        <volume name="root"/>
        <volume name="tmp"/>
        <volume name="opt"/>
        <volume name="srv"/>
        <volume name="usr/local"/>
        <volume name="var" copy_on_write="false"/>
      </systemdisk>
      <size unit="G" unpartitioned="2">60</size>
    </type>
  </preferences>
```

+ KIWI version: 9.20.5-3.18.1
+ Operating system: obs-worker (opensuse leap 15.1)
+ OBS version: 2.10.8-lp151.56.2

schaefi:
If devicepersistency="by-label" is set in the KIWI description
it will correctly operate on the fstab values but still uses
the UUID based setting for root= in the grub config. This commit
allows to set root=LABEL=... in the grub config in case the
devicepersistency requested it. In order for this to work this
commit also had to increase the scope of the grub helper
method _fix_grub_root_device_reference which is now called in
any case. This Fixes #1757

### DIFF
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -25,6 +25,7 @@ from kiwi.mount_manager import MountManager
 from kiwi.storage.setup import DiskSetup
 from kiwi.path import Path
 from kiwi.defaults import Defaults
+from kiwi.utils.block import BlockID
 
 from kiwi.exceptions import (
     KiwiBootLoaderTargetError
@@ -549,6 +550,7 @@ class BootLoaderConfigBase:
 
     def _get_root_cmdline_parameter(self, uuid):
         cmdline = self.xml_state.build_type.get_kernelcmdline()
+        persistency_type = self.xml_state.build_type.get_devicepersistency()
         if cmdline and 'root=' in cmdline:
             log.info(
                 'Kernel root device explicitly set via kernelcmdline'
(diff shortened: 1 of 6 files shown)


davidcassany:
> With this commit I increased the scope of the helper method _fix_grub_root_device_reference(). I had to do it because if you set GRUB_DISABLE_LINUX_UUID in the etc/default/grub it will lead to grub no longer being able to work with loop mapped devices correctly.

I see, yes if grub-mkconfig can't be instructed to use LABELs (I could not find anything like that) there must be some sort of custom procedure. I am fine by reusing this already existing patch and execute it in any case. This also facilitates testing, there will be less paths to cover. I am wondering what happens on the installed system with the `GRUB_DISABLE_LINUX_UUID=true` when the grub configuration is regenerated. I guess it will just use the device path of the current root as it does for when executed on a loop mount device.

I think this is a good move.

Nonfilin78:
Short question: Will this patch be available in the openSUSE:Factory in the short term?

schaefi:
yes, I'll submit to Factory this week

PR #1759, opened 2021-03-09
Exclude /run and /tmp to be synched into the image

### LINKED ISSUE #1744: Built images have content in /run
`/run` is for runtime content and is generally in a ramdisk. However, images built with kiwi currently have this fully populated, like a running system. This can be seen by either mounting the image manually or if it uses btrfs snapshots, by `ls /.snapshots/1/snapshot/run`.

davidcassany:
This commit makes sure the contents of /run and /tmp are ignored when
synchronizing the generated root tree into the image.

Fixes #1744

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -246,7 +246,7 @@ class Defaults:
         :rtype: list
         """
         exclude_list = [
-            'image', '.profile', '.kconfig',
+            'image', '.profile', '.kconfig', 'run/*', 'tmp/*',
             Defaults.get_buildservice_env_name(),
             Defaults.get_shared_cache_location()
         ]
--- a/test/unit/builder/archive_test.py
+++ b/test/unit/builder/archive_test.py
@@ -61,7 +61,8 @@ class TestArchiveBuilder:
         )
         archive.create_xz_compressed.assert_called_once_with(
             'root_dir', exclude=[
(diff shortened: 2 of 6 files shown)


davidcassany:
@schaefi after the prepare step there are some little left overs in `/run`, files like `zypper.pid` and few others. On the test I did it was only about 80kB. According to what I saw on FHS I'd say the best it to definitely omit any `/run` content to be synched. However, now I was wondering if we shouldn't extend it to `/tmp` too, on Fedora and TW this is a tmpfs mountpoint as `/run` and I think that we could easily claim that having images with an empty `/tmp` is good.

What you think?

schaefi (approved):
Looks good to me :+1:

schaefi:
> now I was wondering if we shouldn't extend it to /tmp too

I think this would be a good move. Starting with an empty /tmp is imho a good situation

davidcassany:
> > now I was wondering if we shouldn't extend it to /tmp too
> 
> I think this would be a good move. Starting with an empty /tmp is imho a good situation

Right, then I am adding `/tmp` into this PR too. :+1:

Conan-Kudo:
@davidcassany Could you update the commit title to indicate `/tmp` is included too?

davidcassany:
> @davidcassany Could you update the commit title to indicate `/tmp` is included too?

Oh you are absolutely right, I just updated the body and not the title. Done now.

Conan-Kudo (approved):

PR #1760, opened 2021-03-11
Add a requirement for kiwi-systemdeps-iso-media on disk images

davidcassany:
This commit adds a requirement for `kiwi-systemdeps-iso-media` in
`kiwi-systemdeps-disk-images`. This is to ensure that installing
`kiwi-systemdeps-disk-images` is enough to build OEM images including
install media.


Conan-Kudo (approved):

schaefi:
Hmm, I remember I intentionally did not add this requirement because building installation media is optional.
On the other hand I assume most people building an oem have a deployment strategy and that is often based on our install media.
So I'm ok

schaefi (approved):

PR #1761, opened 2021-03-11
Require qemu-img in any filesystem based image

davidcassany:
This commit moves the qemu-img requirement into the
`kiwi-systemdeps-filesystems` to ensure ISO, OEM and PXE images include
it in the build service. Also this is required for images that are
simple root-trees in a filesystem (image=ext4).

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -156,15 +156,6 @@ Requires:       xorriso
 %ifarch %{ix86} x86_64
 Requires:       syslinux
 %endif
-%if "%{_vendor}" == "debbuild"
-Requires:       qemu-utils
-%else
-%if 0%{?suse_version}
-Requires:       qemu-tools
-%else
-Requires:       qemu-img
-%endif
-%endif
 Requires:       kiwi-systemdeps-core = %{version}-%{release}
 Requires:       kiwi-systemdeps-filesystems = %{version}-%{release}
 Requires:       kiwi-systemdeps-bootloaders = %{version}-%{release}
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Do we really have such a mode where we can make non-bootable root trees on a filesystem disk image?

davidcassany:
> Do we really have such a mode where we can make non-bootable root trees on a filesystem disk image?

Yes we do, and this is basically what we do for `image=pxe` or for the new alternative `image=kis`. There is where I actually found the issue.

PXE images are simple root-trees on a filesystem basically. Kernel and initrd are provided as separate artifacts.

davidcassany:
@schaefi @Conan-Kudo I am seeing this issues (#1761 #1760) in `kiwi-systemdeps-*` by following the `Prefer: *` approach versus the `Support: *` approach in OBS. 

So with `Support` we tell the OBS that the supported package needs to installed in the build environment, so in our tests we have configs like:

```
Support: kiwi-systemdeps-containers
Support: kiwi-systemdeps-iso-media
Support: kiwi-systemdeps-filesystems
Support: kiwi-systemdeps-disk-images
```

Which is mostly installing everything regardless the image type. This results into a not really useful tests of  `kiwi-systemdeps-*` in terms of dependencies, since this easily hides leftovers like the  `qemu-img` we had for ISOs or the `xorriso` for OEMs.

So if we would use `Prefer: kiwi-systemdeps-iso-media` instead, we tell OBS to install the `iso-media` variant whenever there is a requirement for `kiwi-image:iso`, which would only happen for images with `image=iso`. So having the following in our project config results into appending only the needed one for each image type and is more likely to reveal details as the mentioned ones.

 ```
Prefer: kiwi-systemdeps-containers
Prefer: kiwi-systemdeps-iso-media
Prefer: kiwi-systemdeps-filesystems
Prefer: kiwi-systemdeps-disk-images
 ```

schaefi:
ok reading the conversation and the explanation here this makes sense to me. Thanks

schaefi (approved):

Conan-Kudo (approved):

PR #1762, opened 2021-03-11
Public API Strong Typing for kiwi package_manager

KeithMnemonic:
Apply Strong Typing to KIWI Public APIs for package_manager files

 package_manager/base.py
 package_manager/pacman.py
 package_manager/microdnf.py
 package_manager/apt.py
 package_manager/dnf.py
 package_manager/zypper.py

This references Issue #1644

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -18,11 +18,13 @@
 import re
 import os
 import logging
+from typing import List
 
 # project
 from kiwi.command import Command
 from kiwi.path import Path
 from kiwi.package_manager.base import PackageManagerBase
+from kiwi.system.root_bind import RootBind
 from kiwi.exceptions import (
     KiwiDebootstrapError,
     KiwiRequestError
@@ -41,7 +43,7 @@ class PackageManagerApt(PackageManagerBase):
     :param dict command_env: apt-get command environment from repository
         runtime configuration
     """
-    def post_init(self, custom_args=None):
+    def post_init(self, custom_args: List = None) -> None:
         """
         Post initialization method
 
@@ -52,7 +54,7 @@ class PackageManagerApt(PackageManagerBase):
         self.custom_args = custom_args
         if not custom_args:
             self.custom_args = []
-        self.deboostrap_minbase = True
+        self.deboostrap_minbase: bool = True
 
         runtime_config = self.repository.runtime_config()
         self.apt_get_args = runtime_config['apt_get_args']
@@ -60,7 +62,7 @@ class PackageManagerApt(PackageManagerBase):
         self.distribution = runtime_config['distribution']
         self.distribution_path = runtime_config['distribution_path']
 
-    def request_package(self, name):
+    def request_package(self, name: str) -> None:
         """
         Queue a package request
 
@@ -68,7 +70,7 @@ class PackageManagerApt(PackageManagerBase):
         """
         self.package_requests.append(name)
 
-    def request_collection(self, name):
+    def request_collection(self, name: str) -> None:
         """
         Queue a collection request
 
@@ -80,7 +82,7 @@ class PackageManagerApt(PackageManagerBase):
             'Collection(%s) handling not supported for apt-get', name
         )
 
-    def request_product(self, name):
+    def request_product(self, name: str) -> None:
         """
         Queue a product request
 
@@ -92,7 +94,7 @@ class PackageManagerApt(PackageManagerBase):
             'Product(%s) handling not supported for apt-get', name
         )
 
-    def request_package_exclusion(self, name):
+    def request_package_exclusion(self, name: str) -> None:
         """
         Queue a package exclusion(skip) request
 
@@ -104,7 +106,7 @@ class PackageManagerApt(PackageManagerBase):
             'Package exclusion for (%s) not supported for apt-get', name
         )
 
-    def process_install_requests_bootstrap(self, root_bind=None):
+    def process_install_requests_bootstrap(self, root_bind: RootBind = None) -> None:
         """
         Process package install requests for bootstrap phase (no chroot)
         The debootstrap program is used to bootstrap a new system with
@@ -180,7 +182,7 @@ class PackageManagerApt(PackageManagerBase):
                 '%s: %s' % (type(e).__name__, format(e))
             )
 
-    def post_process_install_requests_bootstrap(self, root_bind=None):
+    def post_process_install_requests_bootstrap(self, root_bind: RootBind = None) -> None:
         """
         Mounts the kernel file systems to the chroot environment is
         ready after the bootstrap procedure
@@ -190,7 +192,7 @@ class PackageManagerApt(PackageManagerBase):
         """
         root_bind.mount_kernel_file_systems()
 
-    def process_install_requests(self):
+    def process_install_requests(self) -> None:
         """
         Process package install requests for image phase (chroot)
 
@@ -218,7 +220,7 @@ class PackageManagerApt(PackageManagerBase):
             apt_get_command, self.command_env
         )
 
-    def process_delete_requests(self, force=False):
+    def process_delete_requests(self, force: bool = False) -> None:
         """
         Process package delete requests (chroot)
 
@@ -262,7 +264,7 @@ class PackageManagerApt(PackageManagerBase):
             apt_get_command, self.command_env
         )
 
-    def update(self):
+    def update(self) -> None:
         """
         Process package update requests (chroot)
 
@@ -281,7 +283,7 @@ class PackageManagerApt(PackageManagerBase):
             apt_get_command, self.command_env
         )
 
-    def process_only_required(self):
+    def process_only_required(self) -> None:
         """
         Setup package processing only for required packages
         """
@@ -289,7 +291,7 @@ class PackageManagerApt(PackageManagerBase):
             self.custom_args.append('--no-install-recommends')
         self.deboostrap_minbase = True
 
-    def process_plus_recommended(self):
+    def process_plus_recommended(self) -> None:
         """
         Setup package processing to also include recommended dependencies.
         """
@@ -297,7 +299,7 @@ class PackageManagerApt(PackageManagerBase):
             self.custom_args.remove('--no-install-recommends')
         self.deboostrap_minbase = False
 
-    def match_package_installed(self, package_name, apt_get_output):
+    def match_package_installed(self, package_name: str, package_manager_output: str) -> bool:
         """
         Match expression to indicate a package has been installed
 
@@ -306,33 +308,33 @@ class PackageManagerApt(PackageManagerBase):
         be false positives due to sub package names starting with
         the same base package name
 
-        :param list package_list: list of all packages
-        :param str log_line: apt-get status line
+        :param str package_name: package_name
+        :param str package_manager_output: apt-get status line
 
-        :returns: match or None if there isn't any match
+        :returns: True|False
 
-        :rtype: match object, None
+        :rtype: bool
         """
-        return re.match(
-            '.*Unpacking ' + re.escape(package_name) + '.*', apt_get_output
-        )
+        return bool(re.match(
+            '.*Unpacking ' + re.escape(package_name) + '.*', package_manager_output
+        ))
 
-    def match_package_deleted(self, package_name, apt_get_output):
+    def match_package_deleted(self, package_name: str, package_manager_output: str) -> bool:
         """
         Match expression to indicate a package has been deleted
 
-        :param list package_list: list of all packages
-        :param str log_line: apt-get status line
+        :param str package_name: package_name
+        :param str package_manager_output: apt-get status line
 
-        :returns: match or None if there isn't any match
+        :returns: True|False
 
-        :rtype: match object, None
+        :rtype: bool
         """
-        return re.match(
-            '.*Removing ' + re.escape(package_name) + '.*', apt_get_output
-        )
+        return bool(re.match(
+            '.*Removing ' + re.escape(package_name) + '.*', package_manager_output
+        ))
 
-    def _package_requests(self):
+    def _package_requests(self) -> List:
         items = self.package_requests[:]
         self.cleanup_requests()
         return items
--- a/kiwi/package_manager/base.py
+++ b/kiwi/package_manager/base.py
@@ -15,6 +15,10 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+from typing import List
+
+from kiwi.system.root_bind import RootBind
+from kiwi.repository.base import RepositoryBase
 
 
 class PackageManagerBase:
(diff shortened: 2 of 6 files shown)


kiwi/package_manager/base.py
```diff
@@ -28,17 +32,18 @@ class PackageManagerBase:
...
-        self.exclude_requests = []
+        self.package_requests: List[str] = []
+        self.collection_requests: List[str] = []
+        self.product_requests: List[str] = []
+        self.exclude_requests: List[str] = []
 
         self.post_init(custom_args)
 
-    def post_init(self, custom_args=None):
+    def post_init(self, custom_args: list = None) -> None:
```
schaefi, line 46: Should this not also use the List type from Typing ?

kiwi/package_manager/apt.py
```diff
@@ -41,7 +43,8 @@ class PackageManagerApt(PackageManagerBase):
     :param dict command_env: apt-get command environment from repository
         runtime configuration
     """
-    def post_init(self, custom_args=None):
+
```
schaefi, line 46: Any reason for the extra empty line ?

kiwi/package_manager/base.py
```diff
@@ -124,73 +129,82 @@ def process_delete_requests(self, force=False):
...
+    def process_plus_recommended(self) -> None:
         """
         Setup package processing to also include recommended dependencies
 
         Implementation in specialized package manager class
         """
         raise NotImplementedError
 
-    def match_package_installed(self, package_list, log_line):
+    def match_package_installed(self, package_list: list, log_line: str) -> Union[Match, None]:
```
schaefi, line 156: list vs List ?

kiwi/package_manager/dnf.py
```diff
@@ -36,7 +38,7 @@ class PackageManagerDnf(PackageManagerBase):
     :param dict command_env: dnf command environment from repository runtime
         configuration
     """
-    def post_init(self, custom_args=None):
+    def post_init(self, custom_args: list = None) -> None:
```
schaefi, line 41: list vs. List

kiwi/package_manager/dnf.py
```diff
@@ -211,21 +213,21 @@ def update(self):
...
-    def process_plus_recommended(self):
+    def process_plus_recommended(self) -> None:
         """
         Setup package processing to also include recommended dependencies.
         """
         if '--setopt=install_weak_deps=False' in self.custom_args:
             self.custom_args.remove('--setopt=install_weak_deps=False')
 
-    def match_package_installed(self, package_name, dnf_output):
+    def match_package_installed(self, package_list: list, log_line: str) -> Union[Match, None]:
```
schaefi, line 230: list vs. List

kiwi/package_manager/dnf.py
```diff
@@ -242,10 +244,10 @@ def match_package_installed(self, package_name, dnf_output):
         :rtype: match object, None
         """
         return re.match(
-            '.*Installing  : ' + re.escape(package_name) + '.*', dnf_output
+            '.*Installing  : ' + re.escape(package_list) + '.*', log_line
         )
 
-    def match_package_deleted(self, package_name, dnf_output):
+    def match_package_deleted(self, package_list: list, log_line: str) -> Union[Match, None]:
```
schaefi, line 250: list vs. List

kiwi/package_manager/microdnf.py
```diff
@@ -39,7 +42,7 @@ class PackageManagerMicroDnf(PackageManagerBase):
     :param dict command_env: microdnf command environment from repository runtime
         configuration
     """
-    def post_init(self, custom_args=None):
+    def post_init(self, custom_args: list = None) -> None:
```
schaefi, line 45: list vs. List

kiwi/package_manager/microdnf.py
```diff
@@ -209,21 +212,21 @@ def update(self):
...
-    def process_plus_recommended(self):
+    def process_plus_recommended(self) -> None:
         """
         Setup package processing to also include recommended dependencies.
         """
         if '--setopt=install_weak_deps=0' in self.custom_args:
             self.custom_args.remove('--setopt=install_weak_deps=0')
 
-    def match_package_installed(self, package_name, dnf_output):
+    def match_package_installed(self, package_list: list, log_line: str) -> Union[Match, None]:
```
schaefi, line 229: list vs. List

kiwi/package_manager/microdnf.py
```diff
@@ -240,10 +243,10 @@ def match_package_installed(self, package_name, dnf_output):
         :rtype: match object, None
         """
         return re.match(
-            '.*Installing  : ' + re.escape(package_name) + '.*', dnf_output
+            '.*Installing  : ' + re.escape(package_list) + '.*', log_line
         )
 
-    def match_package_deleted(self, package_name, dnf_output):
+    def match_package_deleted(self, package_list: list, log_line: str) -> Union[Match, None]:
```
schaefi, line 249: list vs. List

kiwi/package_manager/pacman.py
```diff
@@ -37,7 +39,7 @@ class PackageManagerPacman(PackageManagerBase):
     :param dict command_env: pacman command environment from repository
         runtime configuration
     """
-    def post_init(self, custom_args=None):
+    def post_init(self, custom_args: list = None) -> None:
```
schaefi, line 42: list vs. List

kiwi/package_manager/pacman.py
```diff
@@ -194,23 +196,23 @@ def update(self):
...
+    def process_plus_recommended(self) -> None:
         """
         Setup package processing to also include recommended dependencies.
 
         There is no such concept of recommended dependencies in pacman.
         """
         pass
 
-    def match_package_installed(self, package_name, pacman_output):
+    def match_package_installed(self, package_list: list, log_line: str) -> Union[Match, None]:
```
schaefi, line 215: list vs. List

kiwi/package_manager/pacman.py
```diff
@@ -227,10 +229,10 @@ def match_package_installed(self, package_name, pacman_output):
         :rtype: match object, None
         """
         return re.match(
-            '.* installing ' + re.escape(package_name) + '.*', pacman_output
+            '.* installing ' + re.escape(package_list) + '.*', log_line
         )
 
-    def match_package_deleted(self, package_name, pacman_output):
+    def match_package_deleted(self, package_list: list, log_line: str) -> Union[Match, None]:
```
schaefi, line 235: list vs. List

kiwi/package_manager/zypper.py
```diff
@@ -38,7 +41,7 @@ class PackageManagerZypper(PackageManagerBase):
     :param dict command_env: zypper command environment from repository
         runtime configuration
     """
-    def post_init(self, custom_args=None):
+    def post_init(self, custom_args: list = None) -> None:
```
schaefi, line 44: list vs. List

kiwi/package_manager/zypper.py
```diff
@@ -202,21 +205,21 @@ def update(self):
...
-    def process_plus_recommended(self):
+    def process_plus_recommended(self) -> None:
         """
         Setup package processing to also include recommended dependencies.
         """
         if '--no-recommends' in self.custom_args:
             self.custom_args.remove('--no-recommends')
 
-    def match_package_installed(self, package_name, zypper_output):
+    def match_package_installed(self, package_list: list, log_line: str) -> Union[Match, None]:
```
schaefi, line 222: list vs. List

kiwi/package_manager/zypper.py
```diff
@@ -233,10 +236,10 @@ def match_package_installed(self, package_name, zypper_output):
         :rtype: match object, None
         """
         return re.match(
-            '.*Installing: ' + re.escape(package_name) + '.*', zypper_output
+            '.*Installing: ' + re.escape(package_list) + '.*', log_line
         )
 
-    def match_package_deleted(self, package_name, zypper_output):
+    def match_package_deleted(self, package_list: list, log_line: str) -> Union[Match, None]:
```
schaefi, line 242: list vs. List

schaefi (changes requested):
Thanks, this is a great start :+1: 

See my comments, mostly nitpick but there is also one inconsistency which I think was handled incorrectly. More details in an extra comment

schaefi:
@KeithMnemonic I saw you found an inconsistency in the package manager code for the match_package_installed and match_package_deleted functions. Great catch :+1: I however think it was solved in the wrong direction. If you look at ```kiwi/command_process.py``` you see how those methods are called. There is poll_show_progress() which iterates over a list of items (usually package names) and calls for each of them _update_progress(). In that update method you see

```python
match_method(item, command_output)
```

item is always a string and command_output is as well.

Thus the wrong information in the match_package_* methods is the comment block and not the parameters. Therefore I think the type hints should be changed to for example:

```python
def match_package_installed(self, package_name: str, zypper_output: str):
```

and the comment blocks in that functions that tells something about ```:param list package_list: list of all packages``` needs to be fixed, that is a wrong information from my perspective

Thanks much

KeithMnemonic:
> @KeithMnemonic I saw you found an inconsistency in the package manager code for the match_package_installed and match_package_deleted functions. Great catch 👍 I however think it was solved in the wrong direction. If you look at `kiwi/command_process.py` you see how those methods are called. There is poll_show_progress() which iterates over a list of items (usually package names) and calls for each of them _update_progress(). In that update method you see
> 
> ```python
> match_method(item, command_output)
> ```
> 
> item is always a string and command_output is as well.
> 
> Thus the wrong information in the match_package_* methods is the comment block and not the parameters. Therefore I think the type hints should be changed to for example:
> 
> ```python
> def match_package_installed(self, package_name: str, zypper_output: str):
> ```
> 
> and the comment blocks in that functions that tells something about `:param list package_list: list of all packages` needs to be fixed, that is a wrong information from my perspective
> 
> Thanks much

Markus, I can change it back to package_name and fix the comments block. I am not sure zypper_output will work as it seems the "Codacy" check was complaining and failing about renaming the parameters from the base class base.py and the individual subclasses such as dnf.py,  zypper.py

See this

https://app.codacy.com/gh/OSInside/kiwi/pullRequest?prid=7088745

and let me know how to get around it

schaefi:
>the "Codacy" check was complaining and failing about renaming the parameters from the base class base.py and the individual subclasses such as dnf.py, zypper.py

yes, it's good that this gets cleaned up now. My suggestion would be. In the base class

```python
def match_package_installed(self, package_name: str, package_manager_output: str) -> bool:
        """
        Match expression to indicate a package has been installed

        Implementation in specialized package manager class

        :param str package_name: unused
        :param str package_manager_output: unused
        """
        raise NotImplementedError
```

and the implementations for the package managers follows this. I'm not sure how mypy reacts on the return statement.
Actually the best solution would be to refactor the base class into a virtual class using ```from abc import ABCMeta, abstractmethod``` But I think this should be an extra pull request as it imho would be too much for the scope of adding type hints in the first place

The other error from codacy ```Method could be a function``` can be solved by using the ```@staticmethod``` decorator

```python
@staticmethod
def has_failed(returncode):
    ...
```

Does this help ?

KeithMnemonic:
> > the "Codacy" check was complaining and failing about renaming the parameters from the base class base.py and the individual subclasses such as dnf.py, zypper.py
> 
> yes, it's good that this gets cleaned up now. My suggestion would be. In the base class
> 
> ```python
> def match_package_installed(self, package_name: str, package_manager_output: str) -> bool:
>     """
>     Match expression to indicate a package has been installed
> 
>     Implementation in specialized package manager class
> 
>     :param str package_name: unused
>     :param str package_manager_output: unused
>     """
>     raise NotImplementedError
> ```
> 
> and the implementations for the package managers follows this. I'm not sure how mypy reacts on the return statement.
> Actually the best solution would be to refactor the base class into a virtual class using `from abc import ABCMeta, abstractmethod` But I think this should be an extra pull request as it imho would be too much for the scope of adding type hints in the first place
> 
> The other error from codacy `Method could be a function` can be solved by using the `@staticmethod` decorator
> 
> ```python
> @staticmethod
> def has_failed(returncode):
>     ...
> ```
> 
> Does this help ?

Marcus,

from the earlier mypy and codacy failurees, if we are using 

return re.match(
            '.*Unpacking ' + re.escape(package_list) + '.*', log_line
        )

I do  not think ->bool will work in the base class per 

re.match(pattern, string, flags=0)
If zero or more characters at the beginning of string match the regular expression pattern, return a corresponding match object. Return None if the string does not match the pattern; note that this is different from a zero-length match.

that is why I needed the union. Do you want to change all of the subclasses to return True if 'match' and False if 'None'?

schaefi:
>I do not think ->bool will work in the base class per

>re.match(pattern, string, flags=0)
>If zero or more characters at the beginning of string match the regular expression pattern, return a corresponding match object. >Return None if the string does not match the pattern; note that this is different from a zero-length match.

>that is why I needed the union. Do you want to change all of the subclasses to return True if 'match' and False if 'None'?

I agree with you. I think the code should change to return a bool value. From my perspective the bool operator in python should do the trick

```python
return bool(
    re.match('...', package_manager_output)
)
```

Thanks

kiwi/package_manager/apt.py
KeithMnemonic: fixed in latest push

kiwi/package_manager/dnf.py
KeithMnemonic: fixed in latest push

kiwi/package_manager/pacman.py
KeithMnemonic: fixed in latest push

kiwi/package_manager/zypper.py
KeithMnemonic: fixed in latest push

kiwi/package_manager/dnf.py
KeithMnemonic: fixed in latest push

kiwi/package_manager/microdnf.py
KeithMnemonic: fixed in latest push

kiwi/package_manager/pacman.py
KeithMnemonic: fixed in latest push

kiwi/package_manager/microdnf.py
KeithMnemonic: fixed in latest push

kiwi/package_manager/microdnf.py
KeithMnemonic: fixed in latest push

KeithMnemonic:
> > I do not think ->bool will work in the base class per
> 
> > re.match(pattern, string, flags=0)
> > If zero or more characters at the beginning of string match the regular expression pattern, return a corresponding match object. >Return None if the string does not match the pattern; note that this is different from a zero-length match.
> 
> > that is why I needed the union. Do you want to change all of the subclasses to return True if 'match' and False if 'None'?
> 
> I agree with you. I think the code should change to return a bool value. From my perspective the bool operator in python should do the trick
> 
> ```python
> return bool(
>     re.match('...', package_manager_output)
> )
> ```
> 
> Thanks

Markus, please re-review and let me know if this looks good now.

kiwi/package_manager/base.py
```diff
@@ -124,73 +129,82 @@ def process_delete_requests(self, force=False):
...
         Match expression to indicate a package has been installed
 
         Implementation in specialized package manager class
 
-        :param list package_list: unused
-        :param str log_line: unused
+        :param str package_name: unused
+        :param str package_manager_output: unused
+
+        :return: Match|None
```
schaefi, line 165: I think that should be better ```:return: True|False```

kiwi/package_manager/base.py
```diff
@@ -124,73 +129,82 @@ def process_delete_requests(self, force=False):
...
+
+        :rtype: Match or None
```
schaefi, line 167: return type is ```:rtype: bool```

schaefi (commented):
That looks very good now

just nitpick on the doc string describing the return and rtype values. I think after you changed that to return bool(...) this information should be adapted in all methods that are affected

kiwi/package_manager/base.py
KeithMnemonic: please check again with latest push

kiwi/package_manager/base.py
KeithMnemonic: please check again with lastest push

schaefi (approved):

PR #1764, opened 2021-03-15
Fix apt package manager setting

### LINKED ISSUE #1738: Fix packagemanager setting for apt-get
## Problem description

With reference to #1735 we found an issue with the setting:

```xml
<packagemanager>apt-get</packagemanager>
```

In kiwi we use the name of the section as package name to install this package manager capability.
However on Debian based distros there is no package named ```apt-get```. There is only a package named ```apt```
which provides a tool named apt-get. To avoid inconsistencies like we had it in #1735 and to bring this setting
in line with all other packagemanager settings I suggest to move this to

```xml
<packagemanager>apt</packagemanager>
```

In the schema and the code

The change will include:

* schema change
* XSLT transformation template
* code changes mostly in comments but also some cleanup

schaefi:
**Change packagemanager setting from apt-get to apt**
    
In kiwi we use the name of the section as package name to install this
package manager capability. However on Debian based distros there is
no package named apt-get. There is only a package named apt which
provides a tool named apt-get. To avoid inconsistencies like we had it
in Issue #1735 and to bring this setting in line with all other
packagemanager settings the setting was moved to just apt.
This Fixes #1738

**Added XSLT transformation schema v73 -> v74**
    
Update schema version and change to package manager 'apt' if 'apt-get' was set

**Update unit- integration tests to schema v74**

### DIFF
--- a/build-tests/arm/fedora/test-image-live/appliance.kiwi
+++ b/build-tests/arm/fedora/test-image-live/appliance.kiwi
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<image schemaversion="7.3" name="kiwi-test-image-live">
+<image schemaversion="7.4" name="kiwi-test-image-live">
     <description type="system">
         <author>Marcus Schaefer</author>
         <contact>ms@suse.com</contact>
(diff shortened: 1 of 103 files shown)


davidcassany (approved):
Nice cleanup :+1:

Issue #1747, opened 2021-03-02
devicepersistency="by-path" not working


t-ru:
## Problem description

devicepersistency="by-path" is set.

Result /etc/fstab:

/dev/system/swap swap swap defaults 0 0
UUID=162a6a80-2097-449a-9ec5-71b4199c66b3 / xfs defaults 0 0
UUID=96D1-9B0D /boot/efi vfat defaults 0 0

Im using xfs on lvm for root. Mounting lvm disks with uuid is a bad, bad idea.
If you make a lvm snapshot the snapshot gets exactly the same uuid as the original volume.
If you now booting. The snapshot will be mounted instead the original volume. 

see also https://xan.manning.io/2017/05/29/best-practice-for-mounting-an-lvm-logical-volume-with-etc-fstab.html

## Expected behaviour
/dev/system/swap swap swap defaults 0 0
/dev/system/root / xfs defaults 0 0
UUID=96D1-9B0D /boot/efi vfat defaults 0 0

OR

/dev/system/swap swap swap defaults 0 0
/dev/system/root / xfs defaults 0 0
/dev/sda1 /boot/efi vfat defaults 0 0

## OS and Software information
+ KIWI version: KIWI (next generation) version 9.23.12
+ Operating system: openSUSE Tumbleweed
+ OBS version: ---

schaefi:
I was looking into this issue and have several questions. First I built an image that uses xfs and LVM to double check on the behavior. My type was as follows:

```xml
<type image="vmx" filesystem="xfs" firmware="efi">
    <systemdisk>
        <volume name="tmp"/>
        <volume name="home"/>
    </systemdisk>
</typer>
```

This results in the following fstab

```
UUID=5f19e4aa-6d99-4956-a419-7b9433e06c52 / xfs defaults 0 0
UUID=c99122e8-e449-4aa6-bcb0-765411d5fb19 /boot xfs defaults 0 0
/dev/systemVG/home /home xfs defaults 0 0
/dev/systemVG/tmp /tmp xfs defaults 0 0
UUID=CBAA-1065 /boot/efi vfat defaults 0 0
```

So for LVM volumes the persistent volume path is used. For devices outside the LVM the default UUID persistend device info is used. For the root(/) device kiwi also uses the UUID and if I understood you correctly you want us to change this to the LVM device. In my example this would be ```/dev/systemVG/LVRoot / xfs defaults 0 0```. And you are suggesting this due to the snapshot example you made. Is this correct ?

[...]

PR #1765, opened 2021-03-16
Fixed LVM root volume inclusion in fstab

schaefi:
The root(/) volume in a LVM setup was included as UUID device
path. This is inconsistent compared to the non root volumes
which are included by their LVM representation:

    /dev/VGROUP_NAME/VOLUME_NAME

This commit makes sure the root volume is included by its
persistent LVM device representation too. This is related
to Issue #1747


davidcassany (approved):
Looks good :+1:

PR #1767, opened 2021-03-16
Add API Strong Typing to the repository APIs

KeithMnemonic:
Strong Typing has been added to the following files:

 repository/base.py
 repository/pacman.py
 repository/apt.py
 repository/dnf.py
 repository/template/apt.py
 repository/zypper.py

References: #1644

### DIFF
--- a/kiwi/repository/apt.py
+++ b/kiwi/repository/apt.py
@@ -19,6 +19,7 @@ import os
 import logging
 from tempfile import NamedTemporaryFile
 from urllib.parse import urlparse
+from typing import List, Dict
 
 # project
 from kiwi.repository.template.apt import PackageManagerTemplateAptGet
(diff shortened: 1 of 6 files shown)


kiwi/repository/apt.py
```diff
@@ -120,7 +121,7 @@ def use_default_location(self):
             os.sep.join([self.manager_base, 'preferences.d'])
         self._write_runtime_config(system_default=True)
 
-    def runtime_config(self):
+    def runtime_config(self) -> dict:
```
schaefi, line 124: Please use ```from typing import Dict``` for this type and the other places where this was used. Thanks

schaefi (changes requested):
Looks great, just the dict type should be used from the typing module. Thanks much :+1:

kiwi/repository/apt.py
KeithMnemonic: resolved in latest push

schaefi (approved):

PR #1769, opened 2021-03-16
Add Strong Typing to builder APIs

KeithMnemonic:
This PR add Strong Typing for the builder API to the following files:

builder/container.py
 builder/kis.py
 builder/filesystem.py
 builder/disk.py
 builder/live.py
 builder/install.py
 builder/archive.py

References #1644.

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -19,6 +19,7 @@ import os
 import logging
 import pickle
 from tempfile import NamedTemporaryFile
+from typing import Dict, List
 
 # project
 import kiwi.defaults as defaults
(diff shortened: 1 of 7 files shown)


kiwi/builder/disk.py
```diff
@@ -688,7 +690,7 @@ def _build_boot_filesystems(self, device_map):
             )
             self.system_boot = filesystem
 
-    def _build_and_map_disk_partitions(self, disksize_mbytes):
+    def _build_and_map_disk_partitions(self, disksize_mbytes: float) -> dict:
```
schaefi, line 693: Should be Dict from typing

schaefi (commented):
Great :+1: just one small place where I think Dict from typing should be used

kiwi/builder/disk.py
KeithMnemonic: resolved in latest push

PR #1774, opened 2021-03-19
Delete yum from packagemanager in schema

### LINKED ISSUE #1768: Building minimal CentOS 7 images is broken when "yum" package manager is used
## Problem description
Some time ago, we aliased `yum` to `dnf` in the code. However, I found out that when requesting `yum` for CentOS image descriptions, it attempts to install `yum` when going into the "image" phase. This is fine on CentOS 8, but broken on CentOS 7 (and openSUSE Leap 15.x), where legacy `yum` still exists. Consequently, it breaks because then `dnf` is not installed.

## Expected behavior
DNF gets installed and consistently used, even when `yum` is set in the XML.

Try to build a CentOS 7 image using [the example description](https://github.com/OSInside/kiwi-descriptions/blob/master/centos/x86_64/centos-07.0/config.xml) _without_ `nextgen-yum4` installed in the bootstrap phase.

+ KIWI version: 9.23.20
+ Operating system: Fedora 33
+ OBS version: N/A

schaefi:
Auto convert yum to dnf if set as packagemanager. This allows
to delete the yum handling from code parts in kiwi where this
was still present. In addition this fixes the inclusion of yum
into the packagelist. This Fixes #1768

### DIFF
--- a/build-tests/x86/centos/test-image-live-disk-v7/appliance.kiwi
+++ b/build-tests/x86/centos/test-image-live-disk-v7/appliance.kiwi
@@ -16,7 +16,7 @@
     </profiles>
     <preferences>
         <version>1.3.0</version>
-        <packagemanager>yum</packagemanager>
+        <packagemanager>dnf</packagemanager>
         <bootsplash-theme>charge</bootsplash-theme>
         <locale>en_US</locale>
         <keytable>us</keytable>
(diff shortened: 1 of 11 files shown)


Conan-Kudo:
@schaefi This will require a schema bump, right?

schaefi:
> @schaefi This will require a schema bump, right?

nope. I did a schema bump from v73 -> v74 in #1738. The kiwi package has not been released since then and the version also has not changed. So I consider this change here as a followup pull request to the changed done in #1738

davidcassany (approved):

PR #1776, opened 2021-03-22
Refactor building of root cmdline setting

schaefi:
Creating the root= cmdline parameter was based on methods
that deals with the uuid. However, it's also possible to
use a label information for the root= cmdline. To support
this kiwi issued a 'blkid --uuid' command but that requires
udev device names to be present on the host. The open
buildservice workers do not run udev and fails apart the
standard. This commit refactors the root cmdline setup
to work with the device node as it exists during build
time such that the blkid call runs against that device
node.

### DIFF
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -77,11 +77,11 @@ class BootLoaderConfigBase:
         """
         raise NotImplementedError
 
-    def write_meta_data(self, root_uuid=None, boot_options=''):
+    def write_meta_data(self, root_device=None, boot_options=''):
         """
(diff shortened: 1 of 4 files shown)


schaefi:
I've submitted a version to Staging to check on the integration tests... hope they will all build again, as they all failed again due to obs workers being no standard systems without udev. I don't understand why they insist to stay in the outer world with obs workers

schaefi:
codacy didn't like the rebase ...

schaefi:
All integration tests ok again, also checked the result root= setting from a few builds

davidcassany (approved):
Looks good

PR #1781, opened 2021-03-29
Add support for riscv64

andreas-schwab:


Conan-Kudo (approved):

Conan-Kudo (changes requested):
Can you add to the unit tests to cover riscv64?

schaefi:
would be good, something like this

```python
@patch('platform.machine')
def test_post_init_riscv64_platform(self, mock_machine):
    xml_state = MagicMock()
    xml_state.build_type.get_firmware = Mock(
        return_value=None
    )
    mock_machine.return_value = 'riscv64'
    bootloader = BootLoaderConfigGrub2(xml_state, 'root_dir')
    assert bootloader.arch == mock_machine.return_value
```

actually I wonder why there is no coverage complain...

schaefi (approved):
Looks good to me :+1: Thanks

Conan-Kudo (approved):

PR #1788, opened 2021-04-09
Complete static type checks

### LINKED ISSUE #1644: Public API Strong Typing

schaefi:
**Complete strong typing for API methods**
    
Added required code changes to let mypy pass when running from the toplevel kiwi namespace. This now finally
Fixes #1644

**Added mypy call to tox target**
    
For the tox unit_pyX targets, mypy static type checking is now called prior tests.

**Decommission obsolete code reaching EOL**
    
Use the @decommissioned decorator to raise for API methods  that a over the obsoletion period

### DIFF
--- a/kiwi/boot/image/__init__.py
+++ b/kiwi/boot/image/__init__.py
@@ -23,9 +23,9 @@ from abc import (
     abstractmethod
 )
 
-from kiwi.exceptions import (
-    KiwiBootImageSetupError
-)
+from kiwi.xml_state import XMLState
+
+from kiwi.exceptions import KiwiBootImageSetupError
 
 
 class BootImage(metaclass=ABCMeta):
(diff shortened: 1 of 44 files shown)


schaefi:
The changes includes small code refactoring at places were we used different types in the same scope, e.g None type vs. str

schaefi:
I've applied this version to the Staging area and found some issues regarding the type changes. Going to fix them per failed test

schaefi:
> I've applied this version to the Staging area and found some issues regarding the type changes. Going to fix them per failed test

all good now

PR #1790, opened 2021-04-12
Added openssl to the core requires

schaefi:
openssl is used in kiwi to construct a password hash
if the plaintext password feature for user settings
is used. This Fixes bsc#1184128

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -113,6 +113,7 @@ Requires:       kiwi-tools
 Requires:       mtools
 Requires:       rsync
 Requires:       tar >= 1.2.7
+Requires:       openssl
 
 %description -n kiwi-systemdeps-core
 This metapackage installs the necessary system dependencies


schaefi:
I did this change because of the conversation in https://bugzilla.suse.com/show_bug.cgi?id=1184128

@Conan-Kudo I'm sure this is not complete and I appreciate your help with regards to the other distros. Thanks much

schaefi:
The openssl package is pretty small and has no other requirements which we have not yet already set in kiwi. Thus I thought it does not hurt to have it as a core requirement to the package.

package/python-kiwi-spec-template
```diff
@@ -113,6 +113,7 @@ Requires:       kiwi-tools
 Requires:       mtools
 Requires:       rsync
 Requires:       tar >= 1.2.7
+Requires:       openssl
```
dcermak, line 116: Not sure if OBS handles this well, but you could also require `/usr/bin/openssl` directly, if kiwi just needs the openssl binary. (IIRC OBS does not handle file provides very well though and not sure if this really simplifies things though…)

schaefi: I'd like to avoid that. obs doesn't handle file provides at all to my knowledge which was the reason in the past (and present) that some packages must be mentioned explicitly in the kiwi XML description because the obs worker and dependency solver did not pull them in as they are only a file provides.

schaefi: zypper can resolve them but obs doesn't as far as I know. We also need to make sure the dependency is correct for the other distributions and their package management system. I hope for Neal to check that part :)

Conan-Kudo:
I can't see the referenced bug. 😩

davidcassany (approved):
LGTM, I am fine with the Requires even if it is not always, even rarely, used. I we don't have any `kiwi-systemdep-*` subpackage that relates to unhashed passwords and I also believe it doesn't make sense to create and ad-hoc one for that. So adding it as a core dependency sounds good to me.

schaefi:
> I can't see the referenced bug.

do you have a bugzilla account name ? I'll add you to CC

Conan-Kudo:
> > I can't see the referenced bug.
> 
> @Conan-Kudo do you have a bugzilla account name ? I'll add you to CC

Yep, you can add me with my personal email address as CC.

Conan-Kudo (approved):

package/python-kiwi-spec-template
Conan-Kudo: This is fine for Fedora and Debian/Ubuntu.

PR #1791, opened 2021-04-14
Added option to set the image target architecture

schaefi:
The option --target-arch allows to set the architecture
used to build the image. By default this is the host
architecture. Please note, if the specified architecture
name does not match the host architecture and is therefore
requesting a cross architecture image build, it's important
to understand that for this process to work a preparatory
step to support the image architecture and binary format
on the building host is required and is not considered a
responsibility of kiwi. There will be a followup effort
on providing a plugin for kiwi which should be used to
manage the needed binfmt settings for cross arch image
builds

### DIFF
--- a/doc/source/commands/kiwi.rst
+++ b/doc/source/commands/kiwi.rst
@@ -24,6 +24,7 @@ SYNOPSIS
        result <command> [<args>...]
    kiwi-ng [--profile=<name>...]
            [--shared-cache-dir=<directory>]
+           [--target-arch=<name>]
            [--type=<build_type>]
            [--logfile=<filename>]
            [--debug]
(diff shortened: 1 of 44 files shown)


schaefi:
@davidcassany this is the first addition of some others to follow to help the OBS team with their cross image build effort. From my perspective this change is also a good refactoring as we will now have only one call of platform.machine() in the entire kiwi code and the option to overwrite that value. So regardless of the OBS effort I think this change is good

Once merged I will add a very small change to the zypp.conf as we write it, such that the architecture is set in the config file.

After that there might be a new plugin to handle the binfmt changes required for cross arch image builds. We will setup the basic plugin structure but it will be the OBS people that maintains it.

All these changes together will then be used in OBS for their cross image build feature and we will see how many problems that will create ;)

schaefi:
Most of the changes are on the testing code because we had many places which mocks platform.machine. The actual code change is straight forward imho

davidcassany (approved):
Looks good to me. The tests implementation is also nicer now :+1:

davidcassany:
> this is the first addition of some others to follow to help the OBS team with their cross image build effort. From my perspective this change is also a good refactoring as we will now have only one call of platform.machine() in the entire kiwi code and the option to overwrite that value. So regardless of the OBS effort I think this change is good

Yes I agree, this refactor is a good move regardless the additional client option and potential use cases, having the target architecture managed through defaults is a good move.

PR #1792, opened 2021-04-14
Update zypp.conf architecture setting

schaefi:
Make sure the architecture is set as parameter in the
zypp.conf file used for building the image. This is needed
to allow differentiation between host arch and image arch
for cross image build environments

### DIFF
--- a/kiwi/repository/zypper.py
+++ b/kiwi/repository/zypper.py
@@ -21,6 +21,8 @@ from tempfile import NamedTemporaryFile
 from typing import List, Dict
 
 # project
+import kiwi.defaults as defaults
+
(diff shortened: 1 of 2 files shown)


schaefi:
depends on #1791 to be merged first

davidcassany (approved):
Looks good to me

Conan-Kudo:
@schaefi We're probably going to need an equivalent to this for DNF using `arch` and setting `ignorearch`:

* https://dnf.readthedocs.io/en/latest/conf_ref.html#arch-label
* https://dnf.readthedocs.io/en/latest/conf_ref.html#ignorearch-label

PR #1793, opened 2021-04-15
Add support for a custom exclude file

schaefi:
The new optional metadata file image/exclude_files.yaml can
be placed inside of the local image root tree. At creation time of
the image binary the file contents are used to extend the default
exclude list with additional information. The structure of the
file must be as follows:

```yaml
exclude:
  - exclude-name-used-in-rsync
```

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -15,11 +15,14 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+import logging
 import os
 import glob
 from collections import namedtuple
 import platform
+import yaml
 from pkg_resources import resource_filename
+from typing import List
 
 # project
 from kiwi.path import Path
@@ -42,6 +45,8 @@ ROOT_VOLUME_NAME = 'LVRoot'
 SHARED_CACHE_DIR = '/var/cache/kiwi'
 CUSTOM_RUNTIME_CONFIG_FILE = None
 PLATFORM_MACHINE = platform.machine()
+
+log = logging.getLogger('kiwi')
 
 
 class Defaults:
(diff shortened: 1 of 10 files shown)


schaefi:
This feature is needed for the cross arch image build support from the obs team. The new kiwi plugin from here https://github.com/OSInside/kiwi-crossprepare-plugin prepares the image root tree with qemu binfmt file which should not be part of the final image. To allow for a concept to specify custom files to exclude I came up with this PR

schaefi:
@davidcassany This should be the last change in kiwi related to the cross image build feature in obs. I expect the rest to live in other areas

kiwi/defaults.py
```diff
@@ -286,6 +291,37 @@ def get_exclude_list_for_root_data_sync():
...
+        optional metadata file image/exclude_files.yaml
+
+        :return: list of file and directory names
+
+        :param string root_dir: image root directory
+
+        :rtype: list
+        """
+        exclude_file = os.sep.join(
+            [root_dir, 'image', 'exclude_files.yaml']
```
davidcassany, line 307: I find this location not obvious at a glance. Probably I am not understanding how it will be used though.

From an API perspective, at a glance, this seams a bit convoluted that I have to write the yaml file in this location which is created at runtime, shouldn't this be part of the description (provided by user) or some application static config (provided by the plugin)? I understand this is a pretty specific use case and that probably there is no desire to expose this feature to KIWI itself. But if then wouldn't it be easier to pass the yaml file as an argument and simply return an empty list if the file doesn't exit? so from an API POV the user is free to embed this file where ever it suits best for the specific application or use case. Anyway this is just an idea.

davidcassany (commented):
I just have doubts about the yaml file path and how to provide it.

I am fine with the excluding files concept though.

kiwi/defaults.py
schaefi: I can understand your concern. The idea here is however, not providing an "exclusion feature" on the API or user level. I don't see a big value or need for users to specify a file via the image description or on the command line to express a list of files are allowed in the root tree but are not eligible to be part of the final image.

Because of that I just added handling of that information from the image/ metadata. If it turns out to be a general useful feature we can add an option or the code that copies a file from the description to the image/ metadata. That's how all files and script are handled. For example config.sh is copied to image/config.sh in the new root tree and from there it gets picked up.

So with this PR I intentionally did not expose this feature to the command line or user level but follow the same concept as we have with the other metadata files. Does this makes sense ?

One user of this feature is the new kiwi-crossprepare plugin. From a plugin perspective I think this is a clean workflow as I expect a plugin to be allowed to add to the image/ metadata directly. Or do you think this is not a good workflow ?

Thanks

schaefi: So if we would turn exclude_files.yaml as a user feature we would need to register it here
https://github.com/OSInside/kiwi/blob/master/kiwi/system/setup.py#L909

And that I have not done intentionally

schaefi: But the plugin code knows kiwi handles this and does it here
https://github.com/OSInside/kiwi-crossprepare-plugin/blob/master/kiwi_crossprepare_plugin/tasks/system_crossprepare.py#L127

So this the direct injection which I think is ok

schaefi: Hope this makes sense ?

davidcassany: Yes makes sense, thanks for the clarification :+1:

davidcassany (approved):

PR #1795, opened 2021-04-16
Do not apply default subcommand for derivate containers

davidcassany:
This commit does not apply the default subcommand for derivate
containers.

Fixes bsc#1184823


schaefi (approved):
Looks good :+1:

Issue #1717, opened 2021-02-10
Update kiwi in SLE15


schaefi:
## Problem description

Submit upstream kiwi to SLE15

davidcassany:
There was a positive feedback on [bsc#1182963](https://bugzilla.suse.com/show_bug.cgi?id=1182963) (@schaefi for noticing the relation with the `lsblk` issue), so I believe there is nothing preventing this submission now.

davidcassany:
For referece I have already done these maintenance requests:

* [mr#237952](https://build.suse.de/request/show/237952)
* [mr#237953](https://build.suse.de/request/show/237953)

schaefi:
Thanks much David. I checked on the requests and it seems they are in the work queue. There is a planned release date set to 2021-04-19 for the SP[1/2] release. The GA version has no planned release date set but I also consider this a lower prio

I'm closing the issue here since all work is done and accepted from my perspective

davidcassany:
@schaefi I re-opened the issue since a new submission was required.

The packaging refactor had impact on SUSE Manager preventing the release to be completed on time according to the SP3 schedule. SUSE Manager will adapt its salt states to include the  `kiwi-systemdeps-*` packages for the build hosts ([bsc#1184271](https://bugzilla.suse.com/show_bug.cgi?id=1184280)), this change has to happen together with the kiwi release, hence the delay.

To workaround the delay a temporary submission has been done to SP3 ([sr#238860](https://build.suse.de/request/show/238860)) and a [unbranch issue](https://bugzilla.suse.com/show_bug.cgi?id=1184280) has been created to delete kiwi from SP3 once the maintenance release of SP1 is concluded.

davidcassany:
In relation to that we should be aware that now with the current status on SLE we have no `kiwi-systemdeps` package, thus the `python3-kiwi` recommends to it will never be accomplished. It is not a functional issue, however it means that any simple installation of `python3-kiwi` is likely to be incomplete unless someone also installs the related `kiwi-systemdeps-*`. All this is a consequence of having some specific dependencies spread over different SLE modules (in fact the problem comes from the containers tool chain).

[...]

PR #1796, opened 2021-04-16
Fix appx manifest for WSL containers

### LINKED ISSUE #1780: WSL appx image in OBS required author contact as email results in invalid appx schema
## Problem description

kiwi image definition fields `author` and `contact` overwrite prepared wsl-appx metadata for `AppxManifest.xml` `Identity` `Name`. `Properties` elements `DisplayName` and `PublisherDisplayName` are also overwritten by static string `system-container`,

I am working on direct kiwi WSL appx image generation in OBS. Package `wsl-appx` provides a complete `AppxManifest.xml` and associated metadata. When kiwi image builder `kiwi-images-wsl` creates the `.appx` image, required fields `author` and `contact`:

``` xml
 <image schemaversion="6.1" name="opensuse-wsl">
   <description type="system">
     <author>SUSE WSL Team</author>
     <contact>wsl@suse.com</contact>
     <specification>openSUSE WSL Image</specification>
   </description>
   <preferences>
```
overwrite the prepared `Identity` attribute `Name` with an email address:

``` diff
--- wsl-appx-input-AppxManifest.xml
+++ kiwi-wsl-images-output-AppxManifest.xml
-  <Identity Name="REDACTED.openSUSELeap15.3Beta"      Version="15.3.2645.0" Publisher="E=home:(...), CN=home:(...)" ProcessorArchitecture="x64"/>
+  <Identity Name="SUSE WSL Team &lt;wsl@suse.com&gt;" Version="15.3.2645.0" Publisher="E=home:(...), CN=home:(...)" ProcessorArchitecture="x64"/>

   <Properties>
-    <DisplayName>openSUSE Leap 15.3 Beta</DisplayName>
-    <PublisherDisplayName>SUSE</PublisherDisplayName>
+    <DisplayName>system-container</DisplayName>
+    <PublisherDisplayName>system-container</PublisherDisplayName>
```

This results in invalid character schema validation (angle brackets, spaces) when attempting to run the `.appx` image in Windows, as well as loss of the image name data prepared in package `wsl-appx` using information from file `os-release` and other informaiton. This information is necessary to publish in the Windows Store:

```
PS C:\Users\User> Add-AppxPackage .\opensuse-wsl.x86_64-1.0.0-Build29.1.appx
Add-AppxPackage : Deployment failed with HRESULT: 0x80080204, The Appx package's manifest is invalid.
error 0xC00CE169: App manifest validation error: The app manifest must be valid as per schema: Line 3, Column 15,
Reason: 'SUSE WSL Team <wsl@suse.com>' violates pattern constraint of '[-.A-Za-z0-9]+'.
The attribute 'Name' with value 'SUSE WSL Team <wsl@suse.com>' failed to parse.
```

A related unwanted substitution of data also occurs with the `Properties` elements `DisplayName` and `PublisherDisplayName` by `system-container`, as shown in the above diff. These should also be preserved verbatim for proper publishing in the Windows Store.

## Expected behaviour

- kiwi image generation for image type `appx` should not overwrite `AppxManifest.xml` element `Identity` attribute `Name`.
- The `AppxManifest.xml` schema regex for element `Identity` attribute `Name`.  `[-.A-Za-z0-9]+` which cannot use an email address.
- In this case, the overloaded term `Name` refers to a fully qualified published image identity in the Windows Store, not author identity, so these fields are not quite the correct semantic information.
- The above diff in output `AppxManifest.xml` should be eliminated for publishing the the Windows store to work.

Build `kiwi-images-wsl` appx images in OBS using `AppxManifest.xml` prepared by `wsl-appx`. The OBS project configuration to implement this currently lives in ephemeral package project branches. which I can point to in comments if requested.

+ KIWI version: `python3-kiwi-9.21.23` according to `osc buildinfo -d`
+ Operating system: openSUSE Leap 15.3 Beta image target, but OBS is the build environment
+ OBS version: deployed version at build.opensuse.org as of 2021-03-19, I don't know how to check that deployment.

davidcassany:
This commit prevents KIWI from setting Identity Name attribute and
DisplayName and PublisherDisplayName elements.

Fixes #1780

### DIFF
--- a/kiwi/container/setup/appx.py
+++ b/kiwi/container/setup/appx.py
@@ -96,40 +96,15 @@ class ContainerSetupAppx(ContainerSetupBase):
                 identity.set('Version', history.get('package_version'))
             if history.get('created_by'):
                 identity.set('Publisher', history.get('created_by'))
-            if history.get('author'):
-                identity.set('Name', history.get('author'))
 
     def _update_desktop_visual(self):
         history = self.custom_args.get('history') or {}
-        displaynames = self._get_xpath(
-            '//wsl:Properties/wsl:DisplayName'
-        )
-        published_displaynames = self._get_xpath(
-            '//wsl:Properties/wsl:PublisherDisplayName'
-        )
         visuals = self._get_xpath(
             '//wsl:Applications/wsl:Application/uap:VisualElements'
         )
-        titles = self._get_xpath(
-            '//wsl:Applications/wsl:Application/'
-            'uap:VisualElements/uap:DefaultTile'
-        )
-        if self.custom_args.get('container_name'):
-            for displayname in displaynames:
-                displayname.text = \
-                    self.custom_args.get('container_name')
-            for published_displayname in published_displaynames:
-                published_displayname.text = \
-                    self.custom_args.get('container_name')
         for visual in visuals:
-            if self.custom_args.get('container_name'):
-                visual.set(
-                    'DisplayName', self.custom_args.get('container_name')
-                )
             if history.get('comment'):
                 visual.set('Description', history.get('comment'))
-        for titel in titles:
-            titel.set('ShortName', self.custom_args.get('container_name'))
 
     def _update_application_metadata(self):
         history = self.custom_args.get('history') or {}


davidcassany:
@jfkw with this change KIWI won't ever modify `Name` attribute of the `Identity` element neither will set the `DisplayName` and `PoblisherDisplayName`.  Is this what is requested, is it the expected workflow that the user has no change to modify this values from the the KIWI xml description?

Also it would be great if you could give it a test to this PR

schaefi (approved):
The changes looks good, deleting code is always nice :smile: 

However, it should be crystal clear that this is the intention. Therefore we need feedback from the author of the report if this is the desired behavior prior merge

jfkw:
With this change, KIWI appx image generation is working. The existing metadata in `AppxManifest.xml` is no longer overwritten. Thank you very much for the rapid fix.

jfkw:
An addendum to the earlier success report. The `AppxManifest.xml` attribute values for `DisplayName` and `ShortName` are still being overwritten with value `system-container`:

```diff
--- wsl-appx-input-AppxManifest.xml
+++ kiwi-wsl-images-output-AppxManifest.xml
@@ -17,8 +17,8 @@
     </Capabilities>
     <Applications>
         <Application EntryPoint="Windows.FullTrustApplication" Executable="openSUSE-Leap-15.3-Beta.exe" Id="openSUSELeap153Beta">
-            <uap:VisualElements BackgroundColor="#73ba25" Description="openSUSE Leap 15.3 Beta" DisplayName="openSUSE Leap 15.3 Beta" ... >
-                <uap:DefaultTile ShortName="openSUSE Leap 15.3 Beta" ... >
+            <uap:VisualElements BackgroundColor="#73ba25" Description="openSUSE Leap 15.3 Beta" DisplayName="system-container" ... >
+                <uap:DefaultTile ShortName="system-container" ... >
                     <uap:ShowNameOnTiles>
                         <uap:ShowOn Tile="wide310x150Logo"/>
                         <uap:ShowOn Tile="square310x310Logo"/>
```

This allows the `.appx` image to be installed, but will likely prevent uploading to the Microsoft store.

This test was run using `python3-kiwi-9.23.22` (same version currently in factory), plus commit 7eac24301e08670d351f0182f3f012befb1679e2 from this PR in patch form.

Thanks again for the help.

jfkw:
To the point about the changes in this PR matching end user expectations: I presume the [existing KIWI mechanism](https://osinside.github.io/kiwi/building_images/build_wsl_container.html?highlight=wsl#setup-of-the-wsl-distrolauncher) for supplying `AppxManifest.xml` metadata in `<containerconfig><history/></containerconfig>` will continue to work and will override any individual values supplied in the `AppxManifest.xml` file on disk at `metadata_path` ? If so, I think that will be what end users need, i.e. the ability to customize one or more specific attributes from the defaults provided by package `wsl-appx`.

schaefi:
> To the point about the changes in this PR matching end user expectations: I presume the [existing KIWI mechanism](https://osinside.github.io/kiwi/building_images/build_wsl_container.html?highlight=wsl#setup-of-the-wsl-distrolauncher) for supplying `AppxManifest.xml` metadata in `<containerconfig><history/></containerconfig>` will continue to work and will override any individual values supplied in the `AppxManifest.xml` file on disk at `metadata_path` ? If so, I think that will be what end users need, i.e. the ability to customize one or more specific attributes from the defaults provided by package `wsl-appx`.

yes the basic concept of the overriding mechanism will stay. This pull request reduces the context of the manifest which gets touched by kiwi though.

To be more precise the following fields should no longer be touched and stay as they are provided:

```xml
<Package ...
    <Identity Name="out_of_scope"  ...
    <Properties>
        <DisplayName>out_of_scope</DisplayName
        <PublisherDisplayName>out_of_scope</PublisherDisplayName>
    ...
```

>An addendum to the earlier success report. The AppxManifest.xml attribute values for DisplayName and ShortName are still being overwritten with value ```system-container```

yes it seems the PR is not yet complete

Just to put us all right. You intentionally don't want kiwi to change the values for

* Package(Name)
* Identity/Properties/DisplayName
* Identity/Properties/PublisherDisplayName
* Applications/Application/VisualElements/DefaultTile(ShortName)

Ist that correct ?

Regarding the ShortName we set the given container name or ```system-container``` as the default container name.
I thought this would be a good setting but I must admit I don't know the exact context for which ShortName is used on
Windows.

Can you share your thoughts about ShortName and if that should be a setting that should correlate to the kiwi image description or not

Thanks

jfkw:
Thanks for the review and summary. Yes, I think it would be best to not set `ShortName` to be the KIWI image definition `container-name` as `ShortName` is limited to between 1 and 40 characters. In `wsl-appx` we apply some targeted truncation to get a nice looking string during beta/rc phases.

Here are the current `AppxManifest.xml` templated values (`@VALUE@`) that `wsl-appx` takes responsibility for computing, and should only be overridden by user supplied values in `<containerconfig><history/></containerconfig>`:

```
<Identity Name="46932SUSE.@IDENTITYAPPID@" ProcessorArchitecture="@ARCH@" Publisher="@PUBLISHER@" Version="@VERSION@"/>
    <DisplayName>@PRETTY_NAME@</DisplayName>
    <PublisherDisplayName>@PUBLISHER_DISPLAY_NAME@</PublisherDisplayName>
     <Application Executable="@LAUNCHERNAME@" Id="@APPID@">
         <uap:VisualElements Description="@PRETTY_NAME@" DisplayName="@PRETTY_NAME@">
             <uap:DefaultTile ShortName="@SHORT_NAME@">
             <uap3:Extension Executable="@LAUNCHERNAME@">
                     <desktop:ExecutionAlias Alias="@LAUNCHERNAME@"/>
```

The `AppxManifest.xml` schema places restrictions on length and allowed characters that should not restrict other KIWI images, so it seems best to treat `<containerconfig><history/></containerconfig>` as the only interaction with `AppxManifest.xml` values, even if that means some duplication of standard KIWI image description fields.

As we gain more experience publishing to the Microsoft Store we can hopefully supply some of those constraints in a format suitable for potential inclusion in the KIWI RNG schema. Perhaps a future version of the feature would allow embedding of any validated `AppxManifest.xml` elements for image customization, beyond the currently enumerated `created_by`, `author`, `application_id`, `package_version`, and `launcher`. Customizing branding information for example may be easier for end users to accomplish in their KIWI definition than in a bespoke wsl-appx package.

### AppxManifest.xml schema excerpts

[Identity element schema](https://docs.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-identity#syntax):

```
<Identity Name                   = A string between 3 and 50 characters in length that consists of alpha-numeric, period, and dash characters.
          ProcessorArchitecture? = "x86" | "x64" | "arm" | "arm64" | "neutral"
          Publisher              = A string between 1 and 8192 characters in length that fits the regular expression  of a distinguished name : "..."
          Version                = A version string in quad notation, "Major.Minor.Build.Revision" where Major cannot be "0". 
```


[DisplayName element schema](https://docs.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-displayname#syntax):

```
<DisplayName>
  A string between 1 and 256 characters in length. This string is localizable. 
</DisplayName>
```

[PublisherDisplayName element schema](https://docs.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-publisherdisplayname#syntax)

```
<PublisherDisplayName>
  A string between 1 and 256 characters in length. This string is localizable. 
</PublisherDisplayName>
```

The attribute `ShortName` has a length limit that the user might easily exceed the allowed 40 characters:

[uap:DefaultTile element schema](https://docs.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-uap-defaulttile#syntax)

```
  ShortName?         = A string between 1 and 40 characters in length. 
```

schaefi:
Great :+1: thanks for the details.

@davidcassany will you have a look where in the code the DisplayName is still set and to remove the ShortName handling ? or do you want me to kick in, just in case you are busy with other stuff ?

Thanks much

davidcassany:
Hi @schaefi and @jfkw thank you both for all the details and this precise review.



> @davidcassany will you have a look where in the code the DisplayName is still set and to remove the ShortName handling ? or do you want me to kick in, just in case you are busy with other stuff ?

I'll try to find the spot today afternoon to finalize this card. I will let you know if I struggle to find the time, but it should be fine.

Thanks

davidcassany:
@jfkw the updated PR I think now matches the expectations. Would you mind give it a try? I'm sorry I do not have any testing environment at hand by myself and I can't test it.

jfkw:
Thank you @davidcassany , the updated PR is working perfectly. There is zero diff to the `AppxManifest.xml` put on disk by package `wsl-appx`.

Here is a visual example of where it used to display `system-container` and is now fixed:

wsl-displayname-fixed

Since this PR is purely code removal, and in a code path that only affects WSL image generation, would it be possible to push a backported change to SLE-15 tomorrow in time for the RC deadline? This would prevent a the need for branched package in each WSL image generation project.

lkocman:
Hello team, any chance to get this in prior RC (this Wednesday 21st April) ? Thank you

schaefi:
> Hello team, any chance to get this in prior RC (this Wednesday 21st April) ? Thank you

Thanks much David for the additional fix. I will merge the PR and provide a new kiwi in the Staging are from here: https://download.opensuse.org/repositories/Virtualization:/Appliances:/Staging/ If all tests pass a submission to Factory follows. 

Our SLES refresh cycle is every three month and there is currently a submission pending which you can follow in #1717 If you need anything in SLES earlier we need to talk

davidcassany:
@jfkw @lkocman I was just about to point out the same @schaefi did, we just did a submission of KIWI to SLE 15 a couple of weeks ago to match some RC deadline already, I would prefer if we can just wait until the next SLE regular update cycle. If this is required earlier we need to talk about it, as will not fork KIWI across SPs, so whatever is needed in SLE we aim to push it to the SLE 15 code base and this might take some time.

lkocman:
ah oky thanks for explanation, so then for us the importance of having the maintenance request merged is GM/GA image as this is when we'll build the final image.

schaefi:
@lkocman yes at the time when you plan to release this change needs to be available. This means we should know about your release schedule for the WSL containers. Can you share this information ? 

> I can try to help increase priority of the maint request if needed.

Do you have a jira number which we can use as a reference for this release ? I will add that to the history and this makes it also easier to have conversations with maintenance to speedup submissions.

If you can't share this information publicly here I'd like to ask you to send the information by mail

Thanks

PR #1797, opened 2021-04-19
Change zypper download mode to in-advance

### LINKED ISSUE #1789: Use DownloadInAdvance mode for zypper
## Problem description

In relation to https://build.opensuse.org/request/show/883228 and other upcoming zypper changes (to make use of librpm on single transaction operations) there is the possibility that file triggers start being used. In order to ensure zypper behaves consistently even when the underlying rpm is not including the mentioned patch `DownloadInAdvance` mode should be used, this way the transaction happens as a whole and with the upcoming zypper changes zypper will still be capable to handle the file triggers.

Due to an old Yast related hack zypper defaults to `DownloadAsNeeded` when `--root` flag is provided, which in turn it also disables file checks. We have been suggested to be explicit and always use `--download-in-advance` (or `commit.downloadmode` options in `zypp.conf`) mode in KIWI. 

## Expected behaviour

We run the installation of bootstrap packages in a single transaction, so first all packages are downloaded and then installed.

schaefi:
In relation to upcoming zypper changes e.g to make use of librpm on
single transaction operations there is the possibility that file
triggers start being used. To ensure zypper behaves consistently
DownloadInAdvance mode should be used, this way the transaction
happens as a whole and with the upcoming zypper changes zypper
will still be capable to handle the file triggers.
This Fixes #1789

### DIFF
--- a/kiwi/package_manager/zypper.py
+++ b/kiwi/package_manager/zypper.py
@@ -113,7 +113,8 @@ class PackageManagerZypper(PackageManagerBase):
         """
         command = ['zypper'] + self.zypper_args + [
             '--root', self.root_dir,
-            'install', '--auto-agree-with-licenses'
+            'install', '--download', 'in-advance',
+            '--auto-agree-with-licenses'
         ] + self.custom_args + ['--'] + self._install_items()
         return Command.call(
             command, self.command_env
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
Looks good :+1:

PR #1813, opened 2021-05-12
Fixed setup of repository architecture

schaefi:
Unfortunately the architecture reported by uname is not
necessarily the same name as used in the repository metadata.
For example on armv7l the repo arch is armv7hl when using
zypper. Therefore it's required to allow for a translation
map. This commit implements it for use with zypper on arm
and Fixes bsc#1185287

### DIFF
--- a/kiwi/boot/image/base.py
+++ b/kiwi/boot/image/base.py
@@ -45,9 +45,11 @@ class BootImageBase:
     :param string target_dir: target dir to store the initrd
     :param string root_dir: system image root directory
     :param list signing_keys: list of package signing keys
+    :param str target_arch: target architecture
     """
     def __init__(
-        self, xml_state, target_dir, root_dir=None, signing_keys=None
+        self, xml_state, target_dir, root_dir=None,
+        signing_keys=None, target_arch=None
     ):
         self.xml_state = xml_state
         self.target_dir = target_dir
@@ -55,6 +57,7 @@ class BootImageBase:
         self.boot_xml_state = None
         self.setup = None
         self.signing_keys = signing_keys
+        self.target_arch = target_arch
         self.boot_root_directory = root_dir
 
         if not os.path.exists(target_dir):
--- a/kiwi/boot/image/builtin_kiwi.py
+++ b/kiwi/boot/image/builtin_kiwi.py
@@ -77,7 +77,9 @@ class BootImageKiwi(BootImageBase):
             root_dir=self.boot_root_directory,
             allow_existing=True
         )
-        manager = system.setup_repositories(signing_keys=self.signing_keys)
+        manager = system.setup_repositories(
+            signing_keys=self.signing_keys, target_arch=self.target_arch
+        )
         system.install_bootstrap(
             manager
         )
--- a/kiwi/repository/dnf.py
+++ b/kiwi/repository/dnf.py
@@ -51,6 +51,8 @@ class RepositoryDnf(RepositoryBase):
         """
         self.custom_args = custom_args
         self.exclude_docs = False
+        self.locale = None
+        self.target_arch = None
 
         # extract custom arguments not used in dnf call
         if 'exclude_docs' in self.custom_args:
@@ -63,11 +65,16 @@ class RepositoryDnf(RepositoryBase):
         else:
             self.gpg_check = '0'
 
-        self.locale = list(
-            item for item in self.custom_args if '_install_langs' in item
-        )
+        for argument in self.custom_args:
+            if '_install_langs' in argument:
+                self.locale = argument
+            elif '_target_arch' in argument:
+                self.target_arch = argument
         if self.locale:
-            self.custom_args.remove(self.locale[0])
+            self.custom_args.remove(self.locale)
+        if self.target_arch:
+            self.custom_args.remove(self.target_arch)
+            self.target_arch = self.target_arch.split('%')[1]
 
         self.repo_names: List = []
 
@@ -114,7 +121,7 @@ class RepositoryDnf(RepositoryBase):
             self.root_dir, Defaults.get_custom_rpm_image_macro_name()
         )
         if self.locale:
-            rpmdb.set_macro_from_string(self.locale[0])
+            rpmdb.set_macro_from_string(self.locale)
         rpmdb.write_config()
 
         rpmdb = RpmDataBase(self.root_dir)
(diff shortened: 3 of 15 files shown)


schaefi:
Currently the arch setting is done for zypper in zypp.conf only. The other package managers are not affected. However if cross arch building will be a topic for the other package managers and the arch setting becomes required there as well we might need to update the map with respect to the arch names used/supported in this case. The code from this PR imho allows for this evolution

davidcassany (approved):
Looks good. All this architecture naming is a mess, they are all different in different distros and even within the same distro utilities...

I guess this will happen more and more as long arm is becoming a thing for general purpose IT.

kiwi/defaults.py
```diff
@@ -139,6 +139,32 @@ def get_platform_name():
...
+
+        :return: repo architecture name
+
+        :rtype: str
+        """
+        arch = PLATFORM_MACHINE
+        repo_arch_map = {
+            'zypper': {
+                'armv6l': 'armv6hl',
+                'armv7l': 'armv7hl'
```
afaerber, line 161: You may need to add a mapping for `armv8l` (32-bit kernel on 64-bit hardware), but not sure whether to `armv8hl` (don't think we build any special openSUSE packages for it?) or `armv7hl` then.
Not to mention that this may vary by distro - Raspbian/Debian was using `armv5tel` (or similar) on `armv6l`.
Is there some fallback mechanism for either inputs or outputs here? You may also need to consider `i686`/`i586`/etc.

Conan-Kudo (changes requested):
I foresee many headaches if we let this in. For one, the `armv6l` and `armv7l` mapping to the hard-float names is not actually correct, since it's possible to have soft-float builds.

We need a _non-automatic_ way to do this, because there's no generic means of handling this properly.

kiwi/defaults.py
schaefi: > Is there some fallback mechanism for either inputs or outputs here? 

The fallback mechanism is using what comes out of platform.machine() basically ```uname -m```

> Not to mention that this may vary by distro

yes we are aware of it, that's why the map differentiates per repository manager. Zypper is currently the only one for which we define an arch map in cases when platform.machine() is not adequate. Actually Debian distros are the ones were architecture names are almost never match the platform architecture name.

Conan-Kudo: We're going to need a similar map for DNF too. Otherwise this method of cross-arch image builds will fail when DNF is selected.

schaefi:
> I foresee many headaches if we let this in. For one, the `armv6l` and `armv7l` mapping to the hard-float names is not actually correct, since it's possible to have soft-float builds.
> 
> We need a _non-automatic_ way to do this, because there's no generic means of handling this properly.

I agree with you. Actually the non-automatic way is to set the architecture on the commandline via --target-arch <name>

we might want to change the kiwi code in here:

```
self.runtime_zypp_config.set(
    'main', 'arch', ...
)
```

and do this only on an explicit arch setting from the commandline

Conan-Kudo:
@schaefi I vastly prefer that over this. We should also take care to wire up DNF too, because it has this capability.

schaefi:
ok

Conan-Kudo:
For DNF, if `--target-arch` is passed, we need to set in `dnf.conf` both `arch=<arch>` and `ignorearch=True` (that's what DNF's `--forcearch` option does).

afaerber:
Won't expecting explicit command line arguments still break the case of Kiwi in OBS for armv7l that this pull request is trying to fix?

Conan-Kudo:
No, because OBS emulates the arch before loading the image build command, so KIWI thinks it operates in native mode.

kiwi/defaults.py
afaerber: > > Is there some fallback mechanism for either inputs or outputs here?
> 
> The fallback mechanism is using what comes out of platform.machine() basically `uname -m`

Not what I meant. I only see a 1:1 uname to RPM arch name mapping here. What if the output doesn't exist?
If on x86 you get `i686` from `uname -m` and either identity-map it or not, only few packages on openSUSE would actually be available for `i686` (glibc maybe) but most would be `i586`, which i686 and later are backwards-compatible with; you can't generally map `i686` to `i586` either though. Similarly, on an `armv8l` system you could fall back to installing `armv7hl` packages, as I pointed out, but would come up empty with either `armv8l` or `armv8hl` for openSUSE RPMs while not forbidden to multi-build them.
RPM maintains mappings of compatible architectures, allowing it to install `i*86` packages as alternative to `x86_64`. So my implied question was, do we need to a) make that map a 1:n list (which could then handle soft-float for Neal, too) and b) have a fallback code path looping over compatible architectures? I.e., what exactly is the mapped target architecture passed to (zypper? rpm? manual filesystem ops?) and does that take fallback to compatible architectures into account?

x32 and ILP32 are more special cases. MIPS has several ABIs, too.

afaerber:
> No, because OBS emulates the arch before loading the image build command, so KIWI thinks it operates in native mode.

If it would work, there would be no Bugzilla ticket nor pull request...

Conan-Kudo:
@afaerber We are currently unconditionally setting the arch. We don't need to be doing that. @schaefi will make a PR to make it so we don't do that unless the option is set, which fixes ARM-based "native" builds in the OBS cross environment.

kiwi/defaults.py
schaefi: I got your point and I also agree with Neal about the arch mapping. Thus I abandoned the idea to map anything as it just causes headaches.

Regarding the multiarch compatibility you mentioned, this is not something kiwi cares for. In kiwi we will create the opportunity to specify one base architecture which is global to the application. On the level of the repository setup it will cause the arch options to be set as available for the package manger in question. In case of zypp.conf it's just "arch: ..." how zypper now handles compatible other archs is outside of kiwi's control

schaefi:
> For DNF, if `--target-arch` is passed, we need to set in `dnf.conf` both `arch=<arch>` and `ignorearch=True` (that's what DNF's `--forcearch` option does).

ah ok, I just set basearch from the dnf docs, but I will follow your suggestion as it seems you have done it that way

schaefi:
> > No, because OBS emulates the arch before loading the image build command, so KIWI thinks it operates in native mode.
> 
> If it would work, there would be no Bugzilla ticket nor pull request...

correct. There is a kiwi plugin written to support cross building of images in obs: https://github.com/OSInside/kiwi-crossprepare-plugin. obs uses this plugin to prepare the env for running cross-arch binaries. From there it was still needed that kiwi has an option to know about the architecture. This knowledge is currently needed for the repository configuration such that the package manager can work with the repos. The emulation obs chooses is not a complete VM emulation because of performance reasons.

schaefi:
@Conan-Kudo ok I think I have covered the issues brought up so far

Conan-Kudo (approved):

PR #1814, opened 2021-05-12
Fixed live network setup to be generic

### LINKED ISSUE #1802: PXE Boot of Tumbleweed iso fails with network setup failure when booting
## Problem description
I have followed this document to setup a PXE Boot of an exported Tunmbleweed Live CD .iso
https://forums.opensuse.org/showthread.php/521017-Pxe-boot-sever-in-combination-with-Kiwi-Image-System

## Expected behaviour
I expected it to boot normally but it actually fails with a network failure. The startup log shows:
`[ 16.429139] dracut: FATAL: Network setup failed, see tmp/net.info`
and /tmp/net.info is:
`/lib/kiwi-lib.sh: line 78: ifup: command not found`

The full transcript of what I did and some very helpful comments are in an openSUSE forum thread that I started here - the thread is actually very short:

https://forums.opensuse.org/showthread.php/553245-PXE-boot-of-an-ISO-live-distrbution

Please see comment #5 from @arvidjaar as he clearly knows way more than I do on this. 

See above link. I also have a preserved system and happy to test potential fixes,

+ KIWI version: unknown
+ Operating system: openSUSE Tumbleweed iso
https://download.opensuse.org/tumbleweed/iso/openSUSE-Tumbleweed-XFCE-Live-x86_64-Current.iso.mirrorlist
Downloaded on 26-04-2021
+ OBS version: 

I have add a link to issue #1684 as it maybe related.

schaefi:
In dracut the network setup comes with different models
providing a different set of functions. The ifup method as
used in the live iso dracut module is only available with the
network-legacy mode and fails with network-wicked. This commit
uses a dracut conf file in /etc/cmdline.d which uses the dracut
network interface parameters instead of calling module specific
methods. This Fixes #1802

### DIFF
--- a/dracut/modules.d/90kiwi-live/kiwi-live-genrules.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-live-genrules.sh
@@ -15,7 +15,11 @@ case "${root}" in
         wait_for_dev -n "${root#live:}"
         ;;
     live:aoe:/dev/*)
-        /sbin/initqueue \
-            --settled --onetime --unique /sbin/kiwi-live-root "${root}"
+        {
+        printf 'KERNEL=="%s", RUN+="/sbin/initqueue %s %s %s"\n' \
+            "${root#live:aoe:/dev/}" "--settled --onetime --unique" \
+            "/sbin/kiwi-live-root" "${root#live:aoe:}"
+        } >> /etc/udev/rules.d/99-live-aoe-kiwi.rules
+        wait_for_dev -n "${root#live:aoe:}"
         ;;
 esac
--- a/dracut/modules.d/90kiwi-live/parse-kiwi-live.sh
+++ b/dracut/modules.d/90kiwi-live/parse-kiwi-live.sh
@@ -14,6 +14,11 @@ fi
 
 modprobe -q loop
 
+need_network=0
+if getargbool 0 rd.kiwi.live.pxe; then
+    need_network=1
+fi
+
 case "${liveroot}" in
     live:CDLABEL=*|CDLABEL=*) \
         root="${root#live:}"
@@ -21,15 +26,25 @@ case "${liveroot}" in
         root="live:/dev/disk/by-label/${root#CDLABEL=}"
         rootok=1 ;;
     live:AOEINTERFACE=*|AOEINTERFACE=*) \
+        modprobe -q aoe
         root="${root#live:}"
         root="${root//\//\\x2f}"
         root="live:aoe:/dev/etherd/${root#AOEINTERFACE=}"
+        need_network=1
         rootok=1 ;;
 esac
 
 [ "$rootok" = "1" ] || return 1
 
 info "root was ${liveroot}, is now ${root}"
+
+# create network setup if needed
+if [ "${need_network}" = "1" ];then
+    echo "rd.neednet=1" > /etc/cmdline.d/kiwi-generated.conf
+    if ! getarg "ip="; then
+        echo "ip=dhcp" >> /etc/cmdline.d/kiwi-generated.conf
+    fi
+fi
 
 # make sure that init doesn't complain
 [ -z "${root}" ] && root="live"
--- a/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
@@ -70,19 +70,7 @@ function initGlobalDevices {
     if [ -z "$1" ]; then
         die "No root device for operation given"
     fi
-    if getargbool 0 rd.kiwi.live.pxe; then
-        local bootdev
-        read -r bootdev < /tmp/net.bootdev 2>/dev/null
-        if [ -n "${bootdev}" ] && [ -e /tmp/net."${bootdev}".did-setup ]; then
-            : # already set up
-        elif ! ifup lan0 &>/tmp/net.info;then
-            die "Network setup failed, see /tmp/net.info"
-        fi
-        modprobe aoe
-        isodev="${1#live:aoe:}"
-    else
-        isodev="$1"
-    fi
+    isodev="$1"
     local isodisk
     isodisk=$(lookupIsoDiskDevice "${isodev}")
     isodiskdev=$(echo "${isodisk}" | cut -f1 -d%)
--- a/dracut/modules.d/90kiwi-live/kiwi-live-root.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-live-root.sh
@@ -11,14 +11,14 @@ setupDebugMode
 # initialize profile environment
 initialize
 
+# load required kernel modules
+loadKernelModules
+
 # device nodes and types
 initGlobalDevices "$1"
 
 # live options and their default values
 initGlobalOptions
-
-# load required kernel modules
-loadKernelModules
 
 # mount ISO device
 iso_mount_point=$(mountIso)


arvidjaar:
It is not enough. Currently there is no waiting for AOE root device because network is expected to be configured synchronously (although even in this case it is probably wrong as there is no guarantee that AOE is initialized soon enough. initGlobalDevices modprobes aoe and immediately attempts to detect device - could be too fast). To use common network configuration this needs to be changed to generic event driven framework like in standard block device case - wait for device and act when device appears.

arvidjaar:
To illustrate:

```
[   13.714828] localhost kernel: squashfs: version 4.0 (2009/01/31) Phillip Lougher
[   13.732216] localhost kernel: aoe: AoE v85 initialised.
[   13.721042] localhost dracut-initqueue[670]: mount: /run/overlay/live: unknown filesystem type ''.
[   13.797602] localhost dracut: FATAL: Failed to mount live ISO device
[   13.799883] localhost dracut: Refusing to continue
[   13.732249] localhost dracut-initqueue[664]: Warning:
[   13.753627] localhost systemd[1]: Starting Dracut Emergency Shell...
[   13.782340] localhost systemd[1]: Received SIGRTMIN+21 from PID 589 (plymouthd).
[   13.797560] localhost systemd[1]: Received SIGRTMIN+21 from PID 589 (n/a).
```

It tries to mount root right away without any waiting.

schaefi:
yes you are right, the blocking part was formerly done by the ifup. I've pushed a commit which should handle this more dracut like. So when using AOE the block device /dev/etherd... is being waited for in the initqueue now. As the module parser has setup the need for the network beforehand we should now be stuck in the initqueue until the device appears. From there mounting and all other operations should be safe

During testing I also realized that AOE does not work over KVM virtio-net, so just in case you are using virtio-net for testing you will be stuck in the initqueue waiting...

Let me know if this works better now.

Thanks much

dracut/modules.d/90kiwi-live/kiwi-live-genrules.sh
```diff
@@ -15,7 +15,11 @@ case "${root}" in
...
         ;;
     live:aoe:/dev/*)
-        /sbin/initqueue \
-            --settled --onetime --unique /sbin/kiwi-live-root "${root}"
+        {
+        printf 'KERNEL=="%s", RUN+="/sbin/initqueue %s %s %s"\n' \
+            "${root#live:aoe:/dev/}" "--settled --onetime --unique" \
+            "/sbin/kiwi-live-root" "${root#live:aoe:}"
+        } >> /etc/udev/rules.d/99-live-aoe-kiwi.rules
+        wait_for_dev -n "${root#live:aoe:}"
```
davidcassany, line 23: I like this :+1: Synchronizing the execution of the script in initqueue based on udev events looks like a rubost a approach to me.

dracut/modules.d/90kiwi-live/parse-kiwi-live.sh
```diff
@@ -13,6 +13,12 @@ fi
 [ "${liveroot%%:*}" = "live" ] || return 1
 
 modprobe -q loop
+modprobe -q aoe
```
davidcassany, line 16: I am just wondering if this could cause the script to fail if the module can't be loaded. I think this is script is sourced not executed (not fully sure about that), so if it is sourced it might be executed in some context including `set -e` then this script would fail due to aoe even when `rd.kiwi.live.pxe` is not provided.

I know this might be a chain of unlikely circumstances, but in any case I think it might be just safer to `modprobe aoe` only if needed, as it was.

davidcassany (commented):
I just have one comment regarding the load of `aoe` module, beyond that it all looks good to me.

arvidjaar:
> I also realized that AOE does not work over KVM virtio-net

What makes you think so? May be you mean "over QEMU user networking (SLIRP)", but that is clear - AOE needs L2 network.

Anyway, I successfully booted Live TW with your patch. The options were `rd.kiwi.live.pxe root=live:AOEINTERFACE=e1.0 ip=lan0:none`.

Note that `rd.kiwi.live.pxe` attempts to rename **each** interface to lan0, which is unpredictable in case of multiple interfaces. I am not sure what is intended usage.

Of course due to https://bugzilla.opensuse.org/show_bug.cgi?id=1182227 `ip=dhcp` does not work with wicked backend. But that is out of scope for this PR.

AOE does not really require IP configuration, so ideally we'd need something like global `ip=none` that just performs `ip link set up` for each interface to let AOE work. `ip=dhcp` just adds delays. But that is something you need to discuss with dracut folks :)

schaefi:
> What makes you think so? May be you mean "over QEMU user networking (SLIRP)", but that is clear - AOE needs L2 network.

yes that's what I meant. Thanks for making it clear. I'm not an expert on the network layer, so maybe just out of interest you can give me some insights. I thought SLIRP is an implementation in software of PPP, SLIP. Those two I've learned are considered very simple layer two protocols. Thus in Theory AOE could fit in ? I'm sure I'm mistaken here but would be great to understand this better. Thanks

> Anyway, I successfully booted Live TW with your patch. 

Good news, thanks for the feedback :)

> Note that rd.kiwi.live.pxe attempts to rename each interface to lan0, which is unpredictable in case of multiple interfaces. I am not sure what is intended usage.

you are right. There is a net-if udev rule which rewrites the interface and names it lan0. The reason for this was to be able to provide an interface configuration in the image with the expectation that there is always a lan0 interface present. I agree with you on multiple nics this becomes an issue. Maybe it would be a good followup change to delete that rule rewrite and leave it up to users how to manage the interface names. For predictable interface names net.if.names=1 needs to be passed to the kernel and some nic policy for the name should be set. By default predictable interface naming is switched off on suse, so with or without the lan0 rule if there are multiple network interfaces available some strategy to work with the correct one is needed and I think there is no generically applicable solution in this case.

>Of course due to https://bugzilla.opensuse.org/show_bug.cgi?id=1182227 ip=dhcp does not work with wicked backend. But that is out of scope for this PR.

yes. I have hope this will self heal ;)

>AOE does not really require IP configuration, so ideally we'd need something like global ip=none that just performs ip link set up for each interface to let AOE work. ip=dhcp just adds delays. But that is something you need to discuss with dracut folks :)

yes that's a topic for an upstream conversation

Thanks much for testing and the help

arvidjaar:
> I thought SLIRP is an implementation in software of PPP, SLIP. Those two I've learned are considered very simple layer two protocols. Thus in Theory AOE could fit in ?

User network in QEMU is terminated in QEMU process. No packet is forwarded directly between guest and host - connectivity is implemented by NAT inside QEMU, any outgoing connection is on behalf of QEMU process and incoming connection is port forwarded by QEMU. AoE requires direct L2 connectivity, it does not use IP at all (it is separate Ethernet level protocol) so cannot be NAT'ed. It starts with simple fact that vblade needs interface to listen on, and QEMU user networking does not have any interface on host side.

schaefi:
Thanks, got it

dracut/modules.d/90kiwi-live/parse-kiwi-live.sh
schaefi: I agree, will move the loading to conditional place

schaefi: done now

dracut/modules.d/90kiwi-live/kiwi-live-genrules.sh
schaefi: Thanks and seems to be approved by the great people testing :)

PR #1815, opened 2021-05-12
Fix shim path on fedora

### LINKED ISSUE #1806: RHEL/Fedora UEFI (signed EFI) is not set up correctly
## Problem description

KIWI does not detect and install signed EFI binaries for Shim and GRUB2 on Fedora when the legacy `shim.efi` binary path is missing. Some time ago, the EFI binaries changed to be `shim<efiarch>.efi` in `/boot/efi/EFI/<vendor>` in order to support multiarch setup for UEFI. In particular, that change was done for [supporting booting from 32-bit UEFI on x86_64](https://fedoraproject.org/wiki/Changes/32BitUefiSupport).

+ KIWI version: 9.23.22
+ Operating system: Fedora 34
+ OBS version: N/A

schaefi:
Update shim path lookup
    
Distributions like Fedora has changed the EFI binaries location
to be shim<efiarch>.efi in /boot/efi/EFI/<vendor> in order to
support multiarch setup for UEFI. This change requires the
lookup in KIWI to be more global matching. This Fixes #1806

### DIFF
--- a/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
+++ b/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
@@ -53,10 +53,10 @@
         <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root" groups="root"/>
     </users>
     <repository type="rpm-md" alias="kiwi-next-generation" priority="1">
-        <source path="obs://Virtualization:Appliances:Staging/Fedora_33"/>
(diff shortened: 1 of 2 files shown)


schaefi:
@Conan-Kudo, @martinezjavier ok so I think this change should do the trick to not fall apart if the shim backward compatibility will go away

martinezjavier:
@schaefi looks good to me, thanks!

Conan-Kudo (approved):

PR #1816, opened 2021-05-16
Fixed package build

schaefi:
The new version of sphinx puts manual pages into doc/build/man/8
when it was doc/build/man before. This breaks the Makefile target
to install the documentation. This commit updates the Makefile
to follow the change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,10 +38,10 @@ install:
 	${MAKE} -C tools buildroot=${buildroot} install
 	# manual pages
 	install -d -m 755 ${buildroot}usr/share/man/man8
-	for man in doc/build/man/*.8; do \
+	for man in doc/build/man/8/*.8; do \
 		test -e $$man && gzip -f $$man || true ;\
 	done
-	for man in doc/build/man/*.8.gz; do \
+	for man in doc/build/man/8/*.8.gz; do \
 		install -m 644 $$man ${buildroot}usr/share/man/man8 ;\
 	done
 	# completion


Conan-Kudo:
Why are _we_ compressing the man pages ourselves?

schaefi:
> Why are _we_ compressing the man pages ourselves?

The sphinx build of the man pages leaves them uncompressed, so I thought it needs to be done. What else compresses man pages ?

Conan-Kudo:
Well, `man` doesn't require man pages to be compressed, but in a package build, RPM will compress the man pages for you. So I'm not sure why we're doing it.

schaefi:
> Well, `man` doesn't require man pages to be compressed, but in a package build, RPM will compress the man pages for you. So I'm not sure why we're doing it.

ah ok so rpm would compress the uncompressed man pages. As we build from spec also for packman and debian wouldn't this cause an inconsistency if I delete the compression step from the Makefile here ?

schaefi:
I'm going to merge this one now as it blocks update of kiwi in Staging. I think we can handle the compression in a subsequent pull request be there a need to change this. Thanks

Conan-Kudo:
> > Well, `man` doesn't require man pages to be compressed, but in a package build, RPM will compress the man pages for you. So I'm not sure why we're doing it.
> 
> ah ok so rpm would compress the uncompressed man pages. As we build from spec also for packman and debian wouldn't this cause an inconsistency if I delete the compression step from the Makefile here ?

Yeah, can we have a PR to delete the compression steps?

schaefi:
I'm looking into it

PR #1818, opened 2021-05-17
Update kiwi installation documentation

### LINKED ISSUE #1799: Installation documentation has an old key
## Problem description

The installation documentation contains an expired key:

```
rpm -qi gpg-pubkey-74cbe823-* | gpg2
gpg: WARNING: no command supplied.  Trying to guess what you mean ...
pub   dsa1024 2009-05-04 [SC] [expires: 2020-10-09]
      F7E82012C74FD0B85F5334DC994B195474CBE823
```

Copied from: https://osinside.github.io/kiwi/installation.html#installation-from-obs

## Expected behaviour

Installation steps, including validation, work without errors. 

Thank you! :)

schaefi:
The installation chapter contained information about the manual
install of package keys. That information is suspect to be always
outdated because these keys changes. Instead of describing the
manual install of the package key the docs moved to use the
auto-import feature of the package manager. As the instructions
were also rpm specific but we also support install via other
package mangers the complete chapter was a bit reworked and
should be more straight forward now. This Fixes #1799

### DIFF
--- a/doc/source/building_images/build_expandable_disk.rst
+++ b/doc/source/building_images/build_expandable_disk.rst
@@ -99,7 +99,7 @@ The following steps shows how to do it:
 
    .. code:: bash
 
-       $ qemu -hda target_disk -m 4096 -serial stdio
+       $ sudo qemu -hda target_disk -m 4096 -serial stdio
 
 
    At first boot of the target_disk the system is expanded to the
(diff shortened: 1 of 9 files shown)


schaefi:
The best way to review is by reading the rendered results

```
tox -e doc
```

* point your browser to doc/build/html/index.html

PR #1819, opened 2021-05-17
Update integration tests to be non interactive

### LINKED ISSUE #1811: Prepare integration tests without interaction
## Problem description

Some integration tests needs a selection at boot time in the grub menu or a selection of the install device if multiple are present

## Expected behaviour

We should go over all integration tests and make sure that no interaction is needed to let them run.

This is important for the function testing concept we want to implement based on openQA. When running through openQA such manual selections are painful and can be easily avoided

schaefi:
Some integration tests allows for interactive dialogs on the
bootloader menu or in the installation process. As we plan to
use these tests for automated functional testing there should
be no interaction whenever possible. This Fixes #1811

### DIFF
--- a/build-tests/arm/fedora/test-image-live/appliance.kiwi
+++ b/build-tests/arm/fedora/test-image-live/appliance.kiwi
@@ -18,7 +18,7 @@
     </preferences>
     <preferences>
         <type image="iso" flags="overlay" firmware="uefi" hybridpersistent_filesystem="ext4" hybridpersistent="true" kernelcmdline="console=ttyS0">
-            <bootloader name="grub2" console="serial"/>
+            <bootloader name="grub2" console="serial" timeout="2"/>
         </type>
     </preferences>
     <users>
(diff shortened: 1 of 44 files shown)


dcermak (changes requested):
Could you please increase the timeout to 5-10s? It could happen that we actually want to do something in the bootloader and then a 2s timeout might not be enough for openQA to spot the bootloader and react accordingly.

davidcassany (approved):
Looks good :+1:

schaefi:
> Could you please increase the timeout to 5-10s? It could happen that we actually want to do something in the bootloader and then a 2s timeout might not be enough for openQA to spot the bootloader and react accordingly.

ok, are we in agreement that 5sec should be enough to hit some key ? With regards to openQA we are using the sendKey feature through the qemu monitor, a 5sec delay should be ok there too

Thoughts ?

dcermak:
> > Could you please increase the timeout to 5-10s? It could happen that we actually want to do something in the bootloader and then a 2s timeout might not be enough for openQA to spot the bootloader and react accordingly.
> 
> ok, are we in agreement that 5sec should be enough to hit some key ? With regards to openQA we are using the sendKey feature through the qemu monitor, a 5sec delay should be ok there too

I'd rather have 10s, unless that will make certain workflows rather cumbersome?

The issue is not so much about sending keys, but rather looking at the screen and finding out **when** to send the key. This looking (= needle matching) is performed once every 0.5-2 seconds depending on the system load. So if you have to look twice, 5 seconds might not give you enough time.

schaefi:
ok, got it. Will update the descriptions to 10sec. Thanks

PR #1820, opened 2021-05-17
Prevent explicit man page compression

schaefi:
The manual pages are compressed by the packager tooling.
There is no need to do this ourselves


schaefi:
I have tested building rpm and debian distros locally with this change and it still works. I have not tested pacman

Can you guys check if you see and regression or bad after effect on this change ?

Thanks

Conan-Kudo (approved):

PR #1821, opened 2021-05-17
Fix blocking of verification dialog

schaefi:
Fixed blocking of progress dialog
    
When starting a progress dialog it must be sure that the fifo
has data to read PRIOR the progress dialog to start, otherwise
there is a blocking condition which still stop the deployment
and never succeeds

Skip verification process for CentOS8
    
On this distribution we run into a console blocking issue
which I have no good explanation for and only seen on
CentOS8. If the console ist not set to ttyS0 or the
multiple consoles (ttyS0, tty0) are configured the issue
is gone. But with only the serial console enabled the
verification fails because the dialog cannot be displayed
for still unknown reasons.

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-dialog-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-dialog-lib.sh
@@ -34,6 +34,7 @@ function run_progress_dialog {
     # Run the gauge dialog via systemd reading progress
     # status information from stdin
     # """
+    stop_plymouth
     declare fifo=${current_progress_fifo}
     local title=$1
     local backtitle=$2


schaefi:
This is not a real fix and I'm not really happy with it :(

I'm bringing this up mostly for a conversation. The changes I made in the code seems obvious to me and maybe deserves a merge. However, it does not solve the blocking issue on CentOS8.

What's very strange is that I only see this blocking problem on CentOS8 and nowhere else. On all other distros the exact same initrd code and dialog display just works. Thus my current assumption is a change on systemd that affects the behavior on CentOS8

* With the provided change on the image description level the issue is gone by a workaround
* It can also be workaround  via ```console="ttyS0" console="tty0"``` or by not using the serial console

This is all not very appealing. With the change in code the behavior is also not a blocking state but a verification failure because the dialog cannot start up. I think this behavior is better than the blocking state

Thoughts ?

schaefi:
It imho would also be good to have an option that allows for opening a debug session instead of reboot when a failed dialog message appears

schaefi:
> It imho would also be good to have an option that allows for opening a debug session instead of reboot when a failed dialog message appears

This can be achieved by

```
rd.debug rd.shell
```

schaefi:
I found the issue

```
rd.plymouth=0 plymouth.enable=0
```

This fixes it on serial console only workloads

I assume there is an issue on plymouth which does not exists on other distros and gets triggered on CentOS8

schaefi:
two days of debugging and then I found the issue and the most simplest fix ever :smile:

schaefi:
The kiwi dialog lib stops plymouth but not for progress dialogs. This change fixes it. I have tested the CentOS8 build which exposes an issue because of it. @davidcassany you can test the currently built integration check, it worked for me now

Issue #1754, opened 2021-03-05
incorrect lookup of image grub path


gdstewart:
## Problem description
I tried to build an ISO image for Debian 10 and it failed with this error. I tried again with the [example appliance description for Debian 10](https://github.com/OSInside/kiwi-descriptions/tree/master/debian/x86_64/debian-buster) and it failed with the same error.

```
graeme@graeme:~> sudo kiwi-ng --profile=Disk --type oem system build --description kiwi-descriptions/debian/x86_64/debian-buster --target-dir /var/cache/test-image-debian
[ INFO    ]: 16:31:43 | Reading runtime config file: '/etc/kiwi.yml'
[ INFO    ]: 16:31:43 | Loading XML description
[ INFO    ]: 16:31:43 | Support for XML markup available
[ INFO    ]: 16:31:43 | --> loaded kiwi-descriptions/debian/x86_64/debian-buster/config.xml
[ INFO    ]: 16:31:43 | --> Selected build type: oem
[ INFO    ]: 16:31:43 | --> Selected profiles: Disk
[ INFO    ]: 16:31:43 | Preparing new root system
[ INFO    ]: 16:31:43 | Setup root directory: /var/cache/test-image-debian/build/image-root
[ INFO    ]: 16:31:43 | Setting up repository http://ftp.halifax.rwth-aachen.de/debian
[ INFO    ]: 16:31:43 | --> Type: apt-deb
[ INFO    ]: 16:31:43 | --> Translated: http://ftp.halifax.rwth-aachen.de/debian
[ INFO    ]: 16:31:43 | --> Alias: 11d73968d25aec7118b11e0a8762028d
[ INFO    ]: 16:31:43 | Setting up repository obs://Virtualization:Appliances:Builder/Debian_10
[ INFO    ]: 16:31:43 | --> Type: apt-deb
[ INFO    ]: 16:31:44 | --> Translated: http://download.opensuse.org/repositories/Virtualization:/Appliances:/Builder/Debian_10/
[ INFO    ]: 16:31:44 | --> Alias: 80d9781543808931573b34b958c155e3
[ INFO    ]: 16:31:44 | Using package manager backend: apt-get
[ INFO    ]: 16:31:44 | Installing bootstrap packages
[ INFO    ]: 16:31:44 | --> collection type: onlyRequired
[ INFO    ]: 16:31:44 | --> package: apt-get
[ INFO    ]: 16:31:44 | --> package: grub2
[...]

PR #1823, opened 2021-05-20
Add log information on grub search

schaefi:
There is a method in kiwi which searches for grub files.
As grub is packaged differently within the distributions
a dynamic lookup is needed. However, the result and where
kiwi looked it up was not part of the log file. In terms
of issues like the one from Issue #1754 it would be very
handy to know about this information. Thus this commit
adds debug information to the log file regarding what
grub files are searched and where and if found

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -643,6 +643,7 @@ class Defaults:
 
         :rtype: str
         """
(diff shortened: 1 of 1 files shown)


davidcassany (approved):
:+1:

PR #1825, opened 2021-05-20
Do not return default stdout if it is no raising on failure

davidcassany:
This commit prevents the use of a default stdout and stderr in case
return code reports errors and it is not raising an exception.

If we are not raising an exception there is no specific need to
artificially append some stdout and stderr default message, we just
behave as if there was no error.

### DIFF
--- a/kiwi/command.py
+++ b/kiwi/command.py
@@ -120,11 +120,11 @@ class Command:
                 '%s: %s: %s' % (command[0], type(e).__name__, format(e))
             )
         output, error = process.communicate()
-        if process.returncode != 0 and not error:
-            error = bytes(b'(no output on stderr)')
-        if process.returncode != 0 and not output:
-            output = bytes(b'(no output on stdout)')
         if process.returncode != 0 and raise_on_error:
+            if not error:
+                error = bytes(b'(no output on stderr)')
+            if not output:
+                output = bytes(b'(no output on stdout)')
             log.debug(
                 'EXEC: Failed with stderr: {0}, stdout: {1}'.format(
                     Codec.decode(error), Codec.decode(output)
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi I found an issue with c080b92710, we know grub changes come at a price :sweat_smile: 

The specific issue is hard to guess from the changes in this PR. Let me explain the issue. Note that if we run a command that returns a non zero value we append `(no output on stdout)` to stdout, which make sense if we want to print the messages the command might be providing. However if we have `raise_on_error=False` it does not make much sense doing so. In fact is causing issues on EFI partition setup here

https://github.com/OSInside/kiwi/blob/5e1853fe94e7965dda1fc54dfe5766f328b56795/kiwi/bootloader/config/grub2.py#L498-L505

Note that if the bash_command here returns a non zero value and an empty stdout our command interface feeds it with `(no output on stdout)`, meaning that in this case the `efi_boot_path` would be set to `(no output on stdout)`, I realized that by analyzing the EFI setup of a plain simple Leap image in which I had this strange folder `/boot/efi/(no output on stdout)` :sweat_smile: 

The change of this PR fixes it and IMHO provides a better command behavior. However this is a scary change, since it is a behavior change in Command utility which is extensively used within KIWI. I checked if `(no output stdout)` literal is evaluated somewhere in code and I could not find it, thus I believe this a safe move.

schaefi:
Thanks for the detailed explanation. I agree with you that was really a good catch and actually I never hit that issue. I think this command behavior exists for years :) Thanks much for fixing it. Even though it's in the heavily used command interface I don't see big potential for regressions.

There is only one case when there is code that checks the "artificial stdout" as you named it, this code would then be broken.
Let's take that risk

schaefi (approved):

PR #1827, opened 2021-05-21
Add integration tests for functions.sh

dcermak:
This is a simple proposal to add integration tests for the functions in `functions.sh` utilizing containers and the [testinfra](testinfra.readthedocs.io/) module.

* create a new tox target `scripts`
* create a test module to verify a few branches of `suseSetupProduct`

TODO:
- [x] documentation
- [ ] more tests
- [x] CI integration

### DIFF
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -129,6 +129,14 @@ To resume your work, :command:`cd` into your local Git repository and call:
     $ source .tox/3/bin/activate
 
 
+Alternatively, you can launch single commands inside the virtualenv without
+sourcing it directly:
+
+.. code:: shell-session
+
+   $ tox -e devel -- kiwi-ng --version
+
+
 Running the Unit Tests
 ----------------------
 
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -181,6 +181,95 @@ example runs the test cases for the Python 3.6 interpreter only:
 
     $ tox -e unit_py3_6
 
+
+Write Integration Tests for the Scripts
+---------------------------------------
+
+Kiwi ships a set of helper functions that can be used in :file:`config.sh` (see
+also: :ref:`working-with-kiwi-user-defined-scripts`:). These utilize containers
+to run the individual functions and verify that they resulted in the desired
+state.
+
+Ensure that you have either :command:`podman` or :command:`docker` installed and
+configured on your system. The integration tests will use :command:`podman` in
+**rootless mode** by default, if it is installed on your system. You can select
+:command:``docker` instead by setting the environment variable
+``CONTAINER_RUNTIME`` to ``docker``. Then you can run the integration tests via
+tox:
+
+.. code:: shell-session
+
+    $ tox -e scripts -- -n NUMBER_OF_THREADS
+
+
+It is recommended to leverage `testinfra https://testinfra.readthedocs.io/`__
+and the `shared_container` and `container_per_test` fixtures for writing these
+integration tests. The fixtures give your test functions a connection to a
+running container (by default that will be `opensuse/tumbleweed` with python3
+installed in it) :file:`functions.sh` copied to :file:`/bin/functions.sh` inside
+the container. You can then use the connection to perform some setup, tear down
+and the actual tests as follows:
+
+.. code:: python
+
+    def test_RmWorks(shared_container):
+        # create the file /root/foobar
+        shared_container.run_expect([0], "touch /root/foobar")
+        assert shared_container.file("/root/foobar").exists
+
+        # source the functions and execute our function under test
+        shared_container.run_expect([0], ". /bin/functions.sh && Rm /root/foobar")
+
+        # verify the result
+        assert not shared_container.file("/root/foobar").exists
+
+
+In this example we used the `shared_container` fixture: it creates a podman
+container at the start of the test session and gives each function using this
+fixture the same connection. Therefore you must only use it for tests where you
+do not perform any mutation of the container that you are not undoing
+afterwards! If you need to perform extensive mutation that you cannot or do not
+want to undo yourself, then resort to the `container_per_test` fixture. It will
+give you a fresh container for each test function. While this makes writing
+tests simpler, it also increases the runtime significantly, thus only use it
+when necessary.
+
+
+Running Tests for multiple container images
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It is important to test certain functions on multiple operating systems &
+versions, to e.g. ensure that older tools behave the same way that you expect
+them to.
+
+This can be achieved by leveraging pytest's `fixture parametrization
+https://docs.pytest.org/en/stable/parametrize.html`__ as follows:
+
+.. code:: python
+
+    @pytest.mark.parametrize(
+        "shared_container",
+        (
+            None,  # None means default (-> tumbleweed)
+            "Leap-15.2",
+        ),
+        indirect=True,
+    )
+    def test_something(shared_container):
+        pass
+
+
+Where we pass multiple image names to the container images to the
+`shared_container` fixture. Pytest will then look for the image with the given
+name in the predefined list of containers in :file:`conftest.py`.
+
+To add a new container, simply add a new `Container` class to the `CONTAINERS`
+list and give it appropriate values for `name`, `url` and optionally for
+`install_command`. The later will be inserted into a :file:`Dockerfile` to build
+a container image for the actual tests. It can be used to include additional
+packages that are not available in the base image.
+
+
 Create a Branch for each Feature or Bugfix
 ------------------------------------------
 
--- a/tox.ini
+++ b/tox.ini
@@ -18,7 +18,8 @@ envlist =
     check,
     unit_py3_8,
     unit_py3_6,
-    packagedoc
+    packagedoc,
+    scripts,
 
 
 [testenv]
@@ -27,12 +28,12 @@ description =
     devel: Test KIWI
 whitelist_externals = *
 basepython =
-    {check,devel,packagedoc,doc,doc_gh_pages,doc_suse,doc_man}: python3
+    {check,devel,packagedoc,doc,doc_gh_pages,doc_suse,doc_man,scripts,}: python3
     unit_py3_8: python3.8
     unit_py3_6: python3.6
     release: python3.6
 envdir =
-    {check,devel,packagedoc,doc,doc_gh_pages,doc_suse,doc_man}: {toxworkdir}/3
+    {check,devel,packagedoc,doc,doc_gh_pages,doc_suse,doc_man,scripts,}: {toxworkdir}/3
     unit_py3_8: {toxworkdir}/3.8
     unit_py3_6: {toxworkdir}/3.6
     release: {toxworkdir}/3.6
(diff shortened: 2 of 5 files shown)


davidcassany (commented):
I like the implementation even I have to admit I am not sure to what extend this is needed as we try to keep `functions.sh` minimal. Anyway I am fine with the implementation, I am just wondering about the `podman` dependency it introduces with the development environment. I'd say this require some comment or explanation somewhere in [docs](https://osinside.github.io/kiwi/contributing.html).

Conan-Kudo (approved):

Conan-Kudo:
I'm okay with this, but I have the same comment as @davidcassany.

schaefi:
I like the implementation and the design of the test code :+1: 

It would be good though to refactor functions.sh first, obsolete code we know we don't want to have there and apply this testing only to the code we also want to maintain in the scope of kiwi.

I will start the refactor of functions.sh and come up with a PR for review. I suggest we agree on that part first and then add the testing. The proposed tests for setting up the product are great and I definitely want to keep them

Thanks for the effort, let's keep this open until the environment around has settled

Thoughts ?

dcermak:
> I like the implementation and the design of the test code +1
> 
> It would be good though to refactor functions.sh first, obsolete code we know we don't want to have there and apply this testing only to the code we also want to maintain in the scope of kiwi.

I think it makes sense to only test what we want to retain, but I think it's important to **first** implement tests before refactoring (especially for code that is only tested in production). Otherwise we might break the code during when refactoring and won't notice it until much later (e.g. when it trickles down to an older SLE relaase and breaks something there).

schaefi:
> > I like the implementation and the design of the test code +1
> > It would be good though to refactor functions.sh first, obsolete code we know we don't want to have there and apply this testing only to the code we also want to maintain in the scope of kiwi.
> 
> I think it makes sense to only test what we want to retain, but I think it's important to **first** implement tests before refactoring (especially for code that is only tested in production). Otherwise we might break the code during when refactoring and won't notice it until much later (e.g. when it trickles down to an older SLE relaase and breaks something there).

sure, with refactoring I also meant to come up with the deprecation concept. Because it doesn't make sense to write tests for code that is deprecated. Please see the open PR I have created to cover that part. I gave a reason for any method that is deprecated. Customers calling those function will get the deprecation information in the log

dcermak:
> > > I like the implementation and the design of the test code +1
> > > It would be good though to refactor functions.sh first, obsolete code we know we don't want to have there and apply this testing only to the code we also want to maintain in the scope of kiwi.
> > 
> > 
> > I think it makes sense to only test what we want to retain, but I think it's important to **first** implement tests before refactoring (especially for code that is only tested in production). Otherwise we might break the code during when refactoring and won't notice it until much later (e.g. when it trickles down to an older SLE relaase and breaks something there).
> 
> sure, with refactoring I also meant to come up with the deprecation concept. Because it doesn't make sense to write tests for code that is deprecated. Please see the open PR I have created to cover that part. I gave a reason for any method that is deprecated. Customers calling those function will get the deprecation information in the log

Ah, yeah, that definitely makes sense.

schaefi:
@dcermak ok, the refactoring of the functions.sh is done and merged. I think we could start to add the integration testing concept. Are you ok with that ?

dcermak:
Marcus Schäfer ***@***.***> writes:

> @dcermak ok, the refactoring of the functions.sh is done and merged. I think we could start to add the integration testing concept. Are you ok with that ? 

Sure. How should we proceed?

schaefi:
great :+1: can you bring this PR into a state ready for review ? As you have another PR open for changing the suseSetupProduct() method I think the integration testing should at least cover this method

When done, let's merge this PR and then continue the conversation on the suseSetupProduct() change ... I still have some questions on that one ;)

Thanks much

doc/source/contributing.rst
```diff
@@ -129,6 +129,16 @@ To resume your work, :command:`cd` into your local Git repository and call:
...
 
 
+Alternatively, you can launch single commands inside the virtualenv without
+sourcing it directly:
+
+.. code:: shell-session
+
+   $ tox -e devel -- /usr/bin/emacs README.rst
+   $ # or launch kiwi:
+   $ tox -e devel -- kiwi-ng
```
schaefi, line 139: I think we should give users not too many "or's" one example how to directly launch from a venv is imho enough, maybe

```
$ tox -e devel -- kiwi-ng --version
```

Thoughts ?

tox.ini
```diff
@@ -198,3 +199,16 @@ usedevelop = True
 commands =
     python setup.py sdist
     twine upload --skip-existing --non-interactive --config-file ./.pypirc dist/*
+
+[testenv:scripts]
+description = Integration tests of the builtin shell scripts
+deps =
+     {[testenv]deps}
+     pytest-testinfra
```
schaefi, line 207: Do we need to set a version constraint ? or is any version of testinfra working with this conftest ?

schaefi (commented):
Just some comments. Design wise this is nice. We are creating a dependency to testinfra though

doc/source/contributing.rst
dcermak: Sure, why not.

tox.ini
dcermak: I don't think that this is really necessary, testinfra appears to have a rather stable interface (plus tox will pull in the latest version anyway). I must confess that I have not tried other versions besides the latest though, so I can add a version constraint here if you'd like.

dcermak:
> We are creating a dependency to testinfra though

Yes, I have added it for convenience, but it's just a test dependency, so I hope that's not a big problem. I can try to remove it if you don't like further dependencies. It will make the code a bit more complex, but it should be manageable.

tox.ini
schaefi: That's fine. I haven't worked with testinfra and if we don't need a version constraint that's great. Just wanted to check with you. Thanks

schaefi:
> > We are creating a dependency to testinfra though
> 
> Yes, I have added it for convenience, but it's just a test dependency, so I hope that's not a big problem. I can try to remove it if you don't like further dependencies. It will make the code a bit more complex, but it should be manageable.

No problem, as you said it's a tox local test dependency and that's fine. I just want to avoid any package (spec) dependencies for python modules as this always triggers a nightmare of issues on suse. Whereas without a version constraint it shouldn't be a problem for the packaging as well

dcermak:
> > > We are creating a dependency to testinfra though
> > 
> > 
> > Yes, I have added it for convenience, but it's just a test dependency, so I hope that's not a big problem. I can try to remove it if you don't like further dependencies. It will make the code a bit more complex, but it should be manageable.
> 
> No problem, as you said it's a tox local test dependency and that's fine. I just want to avoid any package (spec) dependencies for python modules as this always triggers a nightmare of issues on suse. Whereas without a version constraint it shouldn't be a problem for the packaging as well

This dependency needn't even be part of the spec, because these tests require containers and that will very likely fail to run in OBS/koji anyway. That's also why I did not add them to any of the `*requirements.txt`.

schaefi:
@dcermak I have checked out your fork and switched to the branch and called

```
$ tox -e scripts
```

This results in the following error

```
(functions_integration_tests)> tox -e scripts
scripts create: /home/ms/__foo/kiwi/.tox/3
scripts installdeps: -r.virtualenv.dev-requirements.txt, pytest-testinfra
scripts develop-inst: /home/ms/__foo/kiwi
scripts installed: alabaster==0.7.12,anymarkup==0.8.1,anymarkup-core==0.8.1,appdirs==1.4.4,attrs==21.2.0,Babel==2.9.1,bleach==3.3.0,bump2version==1.0.1,bumpversion==0.6.0,certifi==2021.5.30,cffi==1.14.5,chardet==4.0.0,click==8.0.1,colorama==0.4.4,configobj==5.0.6,coverage==5.5,cryptography==3.4.7,distlib==0.3.2,docopt==0.6.2,docutils==0.16,execnet==1.9.0,filelock==3.0.12,flake8==3.9.2,generateDS==2.29.24,ghp-import==2.0.1,idna==2.10,imagesize==1.2.0,importlib-metadata==4.5.0,importlib-resources==5.1.4,iniconfig==1.1.1,jeepney==0.6.0,Jinja2==3.0.1,json5==0.9.5,keyring==23.0.1,-e git+https://github.com/dcermak/kiwi.git@73b4948fad583300c6656bab4b59a3bdf30f0e42#egg=kiwi,lxml==4.6.3,MarkupSafe==2.0.1,mccabe==0.6.1,mock==4.0.3,mypy==0.902,mypy-extensions==0.4.3,packaging==20.9,pkginfo==1.7.0,pluggy==0.13.1,py==1.10.0,pycodestyle==2.7.0,pycparser==2.20,pyenchant==3.2.0,pyflakes==2.3.1,Pygments==2.9.0,pyparsing==2.4.7,pytest==6.2.4,pytest-cov==2.12.1,pytest-forked==1.3.0,pytest-testinfra==6.3.0,pytest-xdist==2.3.0,python-dateutil==2.8.1,pytz==2021.1,pyxattr==0.7.2,PyYAML==5.4.1,readme-renderer==29.0,requests==2.25.1,requests-toolbelt==0.9.1,rfc3986==1.5.0,rstxml2docbook @ git+https://github.com/openSUSE/rstxml2docbook.git@f000fb64b79011e216a6981592fced2ef3c92c75,SecretStorage==3.3.1,six==1.16.0,snowballstemmer==2.1.0,Sphinx==4.0.2,sphinx-rtd-theme==0.5.2,sphinxcontrib-applehelp==1.0.2,sphinxcontrib-devhelp==1.0.2,sphinxcontrib-htmlhelp==2.0.0,sphinxcontrib-jsmath==1.0.1,sphinxcontrib-qthelp==1.0.3,sphinxcontrib-serializinghtml==1.1.5,sphinxcontrib-spelling==7.2.1,toml==0.10.2,tox==3.23.1,tqdm==4.61.1,travis-sphinx==2.2.1,twine==3.4.1,typed-ast==1.4.3,typing-extensions==3.10.0.0,urllib3==1.26.5,virtualenv==20.4.7,webencodings==0.5.1,xattr==0.9.3,xmltodict==0.12.0,zipp==3.4.1
scripts runtests: PYTHONHASHSEED='660591771'
scripts runtests: commands[0] | pytest -vv
ImportError while loading conftest '/home/ms/__foo/kiwi/test/scripts/conftest.py'.
conftest.py:2: in <module>
    from dataclasses import dataclass
E   ModuleNotFoundError: No module named 'dataclasses'
ERROR: InvocationError: '/home/ms/__foo/kiwi/.tox/3/bin/pytest -vv'
ERROR:   scripts: commands failed
ms@asterix:/home/ms/__foo/kiwi
```

Can you check on this ?

Thanks

dcermak:
Which python version are you running @schaefi ?

dcermak:
@schaefi The code no longer needs buildah and it is now configurable whether you want to use docker or podman. Could you recheck if they now work for you?

FYI, concerning the failure that you saw, it appears to me that you are trying to run podman in rootless mode without having subuids correctly configured as described here: https://github.com/containers/podman/blob/master/docs/tutorials/rootless_tutorial.md#etcsubuid-and-etcsubgid-configuration

schaefi (approved):
I've tested the concept and once you get podman or docker to work in rootless mode on the system ```tox -e scripts``` works nicely. Thanks for all this effort. I'd like to merge this first version and come up with a follow up PR for some smaller modifications :)

Can you turn this from draft to mergable ?

Thanks

dcermak:
See https://github.com/OSInside/kiwi/pull/1851

PR #1829, opened 2021-05-27
config/functions.sh: Avoid non-zero exit status

jgleissner:

* In baseStripDocs and baseStripFirmware avoid non-zero exit status of grep. This allows the functions to be used in a script that sets the exit-on-error flag.


dcermak (approved):
LGTM, thanks!

PR #1830, opened 2021-05-27
Remove grep and find from suseSetupProduct

dcermak:
* Remove the usage of grep and find from `suseSetupProduct`

I have verified the change using the tests introduced in #1827.

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -567,7 +567,7 @@ function suseSetupProduct {
     if [ -f /etc/SuSE-brand ];then
         prod=$(head /etc/SuSE-brand -n 1)
     elif [ -f /etc/os-release ];then
-        prod=$(grep "^NAME" /etc/os-release | cut -d '=' -f 2 | cut -d '"' -f 2)
+        prod=$(while read line; do if [[ $line =~ ^NAME ]]; then echo "$line"; fi; done < /etc/os-release | cut -d '=' -f 2 | cut -d '"' -f 2)
     fi
     if [ -d /etc/products.d ];then
         pushd /etc/products.d || return
@@ -576,7 +576,14 @@ function suseSetupProduct {
         elif [ -f "SUSE_${prod}.prod" ];then
             ln -sf "SUSE_${prod}.prod" baseproduct
         else
-            prod=$(find . -maxdepth 1 -name "*.prod" 2>/dev/null | tail -n 1)
+            find_prod() {
+                for f in $(ls .); do
+                    if [[ $f =~ \.prod$ ]]; then
+                        echo "$f"
+                    fi
+                done
+            }
+            prod=$(find_prod | tail -n 1)
             if [ -f "${prod}" ];then
                 ln -sf "${prod}" baseproduct
             fi


kiwi/config/functions.sh
```diff
@@ -998,11 +998,13 @@ function suseSetupProduct {
     # /etc/SuSE-brand or /etc/os-release or the latest .prod file
     # available in /etc/products.d
     # ----
+    set -e
```
schaefi, line 1001: This changes the behavior on a global level. Any command with exit != 0 will now cause the script to exit. The setting will also persist beyond the scope of the method. We should not do this in a function imho

kiwi/config/functions.sh
```diff
@@ -998,11 +998,13 @@ function suseSetupProduct {
...
     local prod=undef
     if [ -f /etc/SuSE-brand ];then
         prod=$(head /etc/SuSE-brand -n 1)
     elif [ -f /etc/os-release ];then
-        prod=$(grep "^NAME" /etc/os-release | cut -d '=' -f 2 | cut -d '"' -f 2)
+        prod=$(python3 -c 'import re
+with open("/etc/os-release") as f: name = re.search(r"^NAME.*$", f.read(-1), re.M)[0]; print(name.split("=")[1].replace("\"", ""))')
```
schaefi, line 1007: can we be sure python to be around in any case ? I think this has potential to fail. The detection of the product is also one of the rare cases were many people rely on in scripts. A change here will impact many image builds. Thus should be done with care. From the change here I cannot tell for sure if it leads to the same result and to be honest what values python+re+replace over grep+cut ?

I'm just asking for the benefit of the change with respect to some potential for regressions

Thanks

kiwi/config/functions.sh
```diff
@@ -1011,7 +1013,14 @@ function suseSetupProduct {
...
         else
-            prod=$(find . -maxdepth 1 -name "*.prod" 2>/dev/null | tail -n 1)
+            find_prod() {
+                for f in $(ls .); do
+                    if [[ $f =~ \.prod$ ]]; then
+                        echo $f
+                    fi
+                done
+            }
+            prod=$(find_prod | tail -n 1)
```
schaefi, line 1023: Can you explain why find should not be used ?

The tools used in the former implementation are provided by:

* findutils
* coreutils

After the change the tools needed are provided by

* coreutils(ls)
* python

Can you share your thoughts why we should change the code ?

Thanks

schaefi (commented):
code wise this is all good. I just don't understand the motivation or benefit ?

Thanks

kiwi/config/functions.sh
dcermak: Oh, I thought it would be only in this function scope. I'll remove it then.

kiwi/config/functions.sh
dcermak: Well, kiwi itself is implemented in python, so python **must** be present for kiwi to run?

kiwi/config/functions.sh
dcermak: `find` introduces another dependency that needs to be included in the buildroot for very minimal image builds, whereas using python, which is already in the buildroot, does not.

dcermak:
> code wise this is all good. I just don't understand the motivation or benefit ?

The motivation is to be able to build images in a really tiny buildroot where next to nothing besides bash, coreutils and kiwi are installed. I am currently running into all kinds of issues when trying to get rid of packages like `grep`, `sed` or `find` from minimal container images as many, many packages assume that they just exist (including kiwi), but I don't want to include them in the buildroot (as I do not want to have to remove them manually later on).

kiwi/config/functions.sh
schaefi: you overlooked that scripts are called chrooted. The python requirement for kiwi is on the host. The requirement you added with the scripts now forces the image to have python installed. It's questionable if we want that

kiwi/config/functions.sh
schaefi: I did not understand that part. You can build an image without python installed. As soon as scripts requires python you cannot have an image without python installed or do not use our script methods. I can't tell if find is more likely to exist on Linux than python, but one is core linux and the other is a programming language

schaefi:
> The motivation is to be able to build images in a really tiny buildroot

ok I got that, but I think the change will not solve this problem. Today our images that uses functions.sh requires findutils and coreutils to be part of the image. After the change the requirement changes to coreutils and python. I think if we would completely replace the shell scripts by python we can go down to only python as a requirement. But all this are requirements to the image root and not the buildroot because most kiwi scripts are executed in a chroot inside of the new root tree

I'm a bit confused :)

kiwi/config/functions.sh
dcermak: Argh, ok, that's not something that I want to achieve. I was assuming that python would be available from kiwi anyway…

kiwi/config/functions.sh
dcermak: Yes, that's an oversight on my side. I definitely do **not** want the image to require python being installed.

dcermak:
> > The motivation is to be able to build images in a really tiny buildroot
> 
> ok I got that, but I think the change will not solve this problem. Today our images that uses functions.sh requires findutils and coreutils to be part of the image. After the change the requirement changes to coreutils and python. I think if we would completely replace the shell scripts by python we can go down to only python as a requirement. But all this are requirements to the image root and not the buildroot because most kiwi scripts are executed in a chroot inside of the new root tree
> 
> I'm a bit confused :)

So, now python is no longer required as well and you only need coreutils.

schaefi:
> So, now python is no longer required as well and you only need coreutils.

looks good to me, thanks :+1: 

Would you like to rebase the commit with master, it would lead to the scripts testing to run in the gitlab CI and we can see if all still works as before ?

Thanks

schaefi (approved):

Issue #1828, opened 2021-05-26
Remove usage of grep, sed, find from functions.sh


dcermak:
## Problem description

We are currently building very minimal container images and try to exclude as much as possible from them, for instance `sed` and `grep`. Unfortunately, that results in kiwi's `config.sh` failing as it relies on `sed` and `grep` to exist.

It would be nice if kiwi could work without `sed` and `grep` being present in the build root.

## OS and Software information
+ KIWI version: 9.23.20
+ Operating system: (irrelevant)
+ OBS version: build.opensuse.org

dcermak:
I'm wondering whether we could simply leverage python for anything that is not part bash, as kiwi depends on python anyway?

schaefi:
The entire functions.sh needs a refactor. Please note most of the stuff is only there for backward compatibility with custom scripts that might use this methods.

Ideally I would like to get rid of almost everything functions.sh provided. We added these methods to support users with common functions for simple setup tasks. In the end however, it has turned out that this causes a lot of maintenance effort as distributions changes all the time and there is actually no common way to setup a system for feature X.

As it's hard to simply remove published stuff the code is there and gets worse while the distros changes. So from my perspective it would be really good if we can find a strategy to refactor/reduce the code from functions.sh into a state that is useful, stable and reliable.

Thoughts ?

dcermak:
> Ideally I would like to get rid of almost everything functions.sh provided. We added these methods to support users with common functions for simple setup tasks. In the end however, it has turned out that this causes a lot of maintenance effort as distributions changes all the time and there is actually no common way to setup a system for feature X. Looking back I think it was a mistake to offer it. 

I'm afraid this is going to be very hard as many `config.sh` scripts in the buildservice just get copied over from the previous release and no one even knows what these functions do and whether they are necessary (hint: in 99% of the cases, they aren't!). So if you eventually really remove certain functions, you'll get a lot of angry people even if you add huge **THIS WILL BE REMOVED!!!** warnings (because who looks at build logs if they don't fail your build?). Maybe a less brutal way would be to make deprecated functions error out unless the user does a `EXPORT DONT_ERROR_OUT_ON_DEPRECATED_FUNCTIONS_I_KNOW_THIS_WILL_BITE_ME_LATER=1`?

> There are many opportunities to roll out system configurations and way better tools like ansible and friends to do it. Integrating them in a kiwi image is pretty easy and moves the responsibility for the system setup to the right place.

[...]

PR #1833, opened 2021-06-01
Refactor config functions code

schaefi:
Reorganize the code into more readable areas like methods
present as helpers, methods for customers, methods which are
distribution specific and also methods that are deprecated
and give a good reason why they are deprecated when they
get called. This is related to Issue #1828

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -1,51 +1,29 @@
-#================
-# FILE          : functions.sh
-#----------------
-# PROJECT       : openSUSE Build-Service
-# COPYRIGHT     : (c) 2006 SUSE LINUX Products GmbH, Germany
-#               :
-# AUTHOR        : Marcus Schaefer <ms@suse.de>
-#               :
-# BELONGS TO    : Operating System images
-#               :
-# DESCRIPTION   : This module contains common used functions
-#               : for the config.sh and image.sh scripts
-#               :
-#               :
-# STATUS        : Development
-#----------------
-
-#======================================
-#             IMPORTANT
-#======================================
-# If you change *anything* in this file
-# PLEASE also adapt the documentation
-# in doc/source/working_with_kiwi/shell_scripts.rst
-#======================================
-#             IMPORTANT
-#======================================
-
+#!/bin/bash
+# Copyright (c) 2021 SUSE Software Solutions Germany GmbH.  All rights reserved.
+#
+# This file is part of kiwi.
+#
+# kiwi is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# kiwi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kiwi.  If not, see <http://www.gnu.org/licenses/>
+#
 #======================================
 # work in POSIX environment
 #--------------------------------------
 export LANG=C
 export LC_ALL=C
-#======================================
-# check base tools
-#--------------------------------------
-for tool in basename dirname;do
-    if [ -x "/bin/${tool}" ] && [ ! -e "/usr/bin/${tool}" ];then
-        ln -s "/bin/${tool}" "/usr/bin/${tool}"
-    fi
-done
-for tool in setctsid klogconsole;do
-    if [ -x "/usr/bin/${tool}" ] && [ ! -e "/usr/sbin/${tool}" ];then
-        ln -s "/usr/bin/${tool}" "/usr/sbin/${tool}"
-    fi
-done
 
 #======================================
-# baseSystemdServiceInstalled
+# Common Functions
 #--------------------------------------
 function baseSystemdServiceInstalled {
     local service=$1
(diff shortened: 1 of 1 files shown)


schaefi:
@dcermak I know this is hard to review because I moved code blocks into more readable areas of the file. Most interesting is the data in the section about deprecated methods. Please see the reasons I gave with any deprecation and if you think I did something wrong just veto in :)

Other than that this is only moving existing code and fixing some long line issues

If we are ok with that (and I'm willing to take the blame from customers) I think we should apply your test suite concept on the rest of the code.

kiwi/config/functions.sh
```diff
@@ -1,51 +1,29 @@
...
-#======================================
-# If you change *anything* in this file
-# PLEASE also adapt the documentation
-# in doc/source/working_with_kiwi/shell_scripts.rst
-#======================================
-#             IMPORTANT
-#======================================
-
+#!/bin/bash
+# Copyright (c) 2015 SUSE Linux GmbH.  All rights reserved.
```
dcermak, line 2: The "All rights reserved" part is definitely wrong and clashes with the GPL
```suggestion
# Copyright (c) 2021 SUSE Linux GmbH.
```

dcermak (commented):
I must confess that I did not check if anything changed in the code when you moved a lot of the functions around, so I trust you here :wink:.

I am not entirely sure whether a deprecation warning is really enough though, as I think most of these functions are simply kept to rot in legacy `config.sh` scripts on OBS & IBS where **no one** looks at the logs, unless they blow up. So maybe we should add some more drastic measures here like failing the build unless you set a environment variable (or do that in a year or so…).

kiwi/config/functions.sh
schaefi: yes will fix it, thanks

schaefi:
>I am not entirely sure whether a deprecation warning is really enough though, as I think most of these functions are simply kept to rot in legacy config.sh scripts on OBS & IBS where no one looks at the logs, unless they blow up. So maybe we should add some more drastic measures here like failing the build unless you set a environment variable (or do that in a year or so…).

I agree, we should be more "aggressive" and fail the script in these cases. I'll adapt

kiwi/config/functions.sh
```diff
@@ -549,7 +187,8 @@ function baseStripUnusedLibs {
     for libname in $1; do
         for libfile in \
             /lib*/"$libname"* /usr/lib*/"$libname"* \
-            /lib/x86_64-linux-gnu/"$libname"* /usr/lib/x86_64-linux-gnu/"$libname"* \
+            /lib/x86_64-linux-gnu/"$libname"* \
+            /usr/lib/x86_64-linux-gnu/"$libname"* \
```
Conan-Kudo, line 191: This won't work for non-x86_64 machines.

kiwi/config/functions.sh
```diff
@@ -1066,62 +491,138 @@ function baseSetRunlevel {
...
+    # /etc/SuSE-brand or /etc/os-release or the latest .prod file
+    # available in /etc/products.d
     # ----
-    # systemd random seed
-    rm -f /var/lib/systemd/random-seed
+    local prod=undef
+    if [ -f /etc/SuSE-brand ];then
+        prod=$(head /etc/SuSE-brand -n 1)
+    elif [ -f /etc/os-release ];then
+        prod=$(grep "^NAME" /etc/os-release | cut -d '=' -f 2 | cut -d '"' -f 2)
```
Conan-Kudo, line 570: This should be flipped. Also, os-release can be sourced in a shell script, so you can simplify this.

Conan-Kudo (changes requested):

dcermak:
Neal Gompa (ニール・ゴンパ) ***@***.***> writes:

>> +    if [ -f /etc/SuSE-brand ];then
> +        prod=$(head /etc/SuSE-brand -n 1)
> +    elif [ -f /etc/os-release ];then
> +        prod=$(grep "^NAME" /etc/os-release | cut -d '=' -f 2 | cut -d '"' -f 2)
>
> This should be flipped. Also, os-release can be sourced in a shell script, so you can simplify this.

Sadly not on some older distros that omit the '"' after the '='.

Conan-Kudo:
@dcermak when you source them, the quotes are already stripped when you echo the variable

schaefi:
@dcermak @Conan-Kudo thanks for the review. Please keep in mind I haven't changed the code, just re-organized by moving and dropped obsolete code. I'm sure the code has issues too, but can we first finish the refactor and look at the code in detail afterwards and create followup PRs for each issue we find ?

Thanks

dcermak:
Neal Gompa (ニール・ゴンパ) ***@***.***> writes:

> @dcermak when you source them, the quotes are already stripped when you echo the variable

That's not what I meant, on some older systems I've seen /etc/os-release
look like this:
```
NAME=openSUSE Leap
```
and sourcing that doesn't work. It might not be an issue anymore, but
the current code handles this (broken) case better.

schaefi:
I'd like to move forward with this one now. My plan would be

* merge the refactor from here, without substantial code changes from non deprecated functions
* submit to Staging, check integration test results. If deprecated methods are used the build will now fail
* On failed builds check why and update either the description or the methods from here
* Continue improving the code in functions.sh which is left. Taking into account what @Conan-Kudo and @dcermak said here
* Continue adding the changes from the other PRs from @dcermak 
* Continue adding the testsuite as suggested by @dcermak

Unless there are objections I'll go ahead now with this approach

dcermak (approved):
Sounds good to me @schaefi

Conan-Kudo (approved):

PR #1849, opened 2021-06-16
Allow creation of LUKS system with empty key

schaefi:
To support cloud platforms better we should allow the
creation of an initial(insecure) LUKS encrypted image
with an empty passphrase/keyfile

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -686,7 +686,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                     grub_default_entries['GRUB_USE_INITRDEFI'] = 'true'
         if self.xml_state.build_type.get_btrfs_root_is_snapshot():
             grub_default_entries['SUSE_BTRFS_SNAPSHOT_BOOTING'] = 'true'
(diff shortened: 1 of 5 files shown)


schaefi:
This is only part of the story. Even with kiwi now able to create LUKS encrypted images with an empty passphrase you are still blocked on boot of the system because systemd and/or grub are asking for that passphrase. You hit enter and the process continuous, but what people want is a non-interactive way to boot up and then change the keyfile such that this can be workable in cloud environments

So the kiwi part is easy but the rest covers cryptsetup and the distro

schaefi:
luks-grub
luks-at-boot-in-systemd

schaefi:
You just hit enter and continue, but it's still a blocking process

rjschwei (commented):
LGTM

schaefi:
success :) with the last patch and the mandatory change

```xml
<type ... luks="" bootpartition="true"/>
```

the image boots without interaction in LUKS mode (dancing) :)

Please note in full encryption mode without an extra boot partition there is still interaction because the
grub cryptodisk  module always asks even if the keyfile has an empty passphrase. For that part to work a patch in the grub  cryptodisk module will be needed

PR #1850, opened 2021-06-16
Allow the ISO to be booted from grub directly

Mexit:
This PR adds the ability to run an ISO image directly from grub.
This can be done using a simple menu, e.g:  [link](https://github.com/Mexit/MultiOS-USB/blob/master/config/grub_loopback/loopback.cfg).
Loopback compatible ISO images on your USB, HDD, SSD drive are automatically detected and you can boot from the selected ISO image. There is no need to burn an ISO image to disk. This allows you to have many different versions on one disk.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -439,6 +439,12 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 lookup_path=lookup_path, mbrid=mbrid
             )
 
+        log.info('--> Creating loopback config')
+        loopback_file = self.boot_dir + '/' + self.get_boot_path() + '/' + self.boot_directory_name + '/loopback.cfg'
+        self._create_loopback_config(
+            loopback_file
+        )
+
     def setup_live_boot_images(self, mbrid, lookup_path=None):
         """
         Create/Provide grub2 boot images and metadata
@@ -1397,4 +1403,12 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         except Exception:
             raise KiwiDiskGeometryError(
                 'unknown format for disk geometry: %s' % fdasd_output
+            )
+
+    def _create_loopback_config(self, filename):
+        with open(filename, 'w') as loopback_cfg:
+            loopback_cfg.write(
+                'source /boot/{0}/grub.cfg{1}'.format(
+                    self.boot_directory_name, os.linesep
+                )
             )
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -440,7 +440,9 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             )
 
         log.info('--> Creating loopback config')
-        loopback_file = self.boot_dir + '/' + self.get_boot_path() + '/' + self.boot_directory_name + '/loopback.cfg'
+        loopback_file = os.path.join(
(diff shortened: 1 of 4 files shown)


schaefi:
Hey, thanks much for your work :+1: Looking at it today

kiwi/bootloader/config/grub2.py
```diff
@@ -439,6 +439,12 @@ def setup_install_boot_images(self, mbrid, lookup_path=None):
                 lookup_path=lookup_path, mbrid=mbrid
             )
 
+        log.info('--> Creating loopback config')
+        loopback_file = self.boot_dir + '/' + self.get_boot_path() + '/' + self.boot_directory_name + '/loopback.cfg'
```
schaefi, line 443: Can this be done by a join, e.g

```python
loopback_file = os.path.join(
    self.boot_dir, self.get_boot_path(), self.boot_directory_name, '/loopback.cfg'
)
```

kiwi/bootloader/template/grub2.py
```diff
@@ -36,6 +36,9 @@ def __init__(self):
                 menuentry "If OK, run snapper rollback and reboot." { true; }
               }
             fi
+            if [ -n "$${iso_path}" ]; then
+              isoboot="iso-scan/filename=$${iso_path}"
```
schaefi, line 40: nipick on 4 space indentation :)

schaefi (commented):
First review just on python stuff...

schaefi:
Just to put me right and to get things in order. The concept here is as follows:

* For install and live ISO images loopback.cfg is created which sources the grub.cfg
* The loopback.cfg is read in via a custom grub menu from "somewhere" (e.g on stick)
   ```
   menuentry ... {
       iso_path=...
       export iso_path
       loopback
       configfile (loop)/.../loopback.cfg
       loopback ... loop
   }
   ```
* The exported iso_path variable from that custom menu entry causes the setting of the isoboot variable
   ```isoboot="iso-scan/filename=$${iso_path}"``` which makes use of the iso-scan feature
* The final boot a menu entry from the grub.cfg in the image contains iso-scan/... in the kernel cmdline and can
   boot ISO media that supports the iso-scan feature

Did I summarize this correctly ?

schaefi:
If I understood it correctly I think the missing part is in the documentation. It would be imho good to have an at least high level overview about this feature and how it can be used. Thoughts ?

Mexit:
Sorry for my mistakes, these are my first steps with Python.

I agree, you summarised it well.

You can manually create, for example, such a grub menu:
```
menuentry "Boot from openSUSE ISO" {
	iso_path="/path/to/openSUSE.iso"
	export iso_path
	loopback loop "$iso_path"
	root=(loop)
	configfile /boot/grub2/loopback.cfg
	loopback --delete loop
}
```
The final kernel cmdline additionally includes:
`iso-scan/filename=/path/to/openSUSE.iso`

schaefi:
ok, great. From a coding perspective this is good now, thanks :+1: 

Before merge we need to update the documentation and give users some information how to make use of the feature. As you said it requires some pre-work with the grub menu entry to make use of the feature. Could you add this documentation in a chapter here:

```doc/source/working_with_images```

This can be simple and straight forward.

Could you do this ?

Thanks much

Mexit:
> Could you do this ?

I will try.

Some infos about this feature: https://git.kernel.org/pub/scm/boot/dracut/dracut.git/log/?qt=grep&q=iso-scan%2Ffilename

schaefi (approved):
Thanks for the docs, this looks great now :+1:

PR #1851, opened 2021-06-18
Functions integration tests

dcermak:
Supersedes https://github.com/OSInside/kiwi/pull/1827

### DIFF
--- a/doc/source/contributing/scripts_testing.rst
+++ b/doc/source/contributing/scripts_testing.rst
@@ -1,0 +1,83 @@
+Write Integration Tests for the Scripts
+---------------------------------------
+
+Kiwi ships a set of helper functions that can be used in :file:`config.sh` (see
+also: :ref:`working-with-kiwi-user-defined-scripts`). These utilize containers
+to run the individual functions and verify that they resulted in the desired
+state.
+
+Ensure that you have either :command:`podman` or :command:`docker` installed and
+configured on your system. The integration tests will use :command:`podman` in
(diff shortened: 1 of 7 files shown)


schaefi:
@dcermak can you please review #1853 before making further force pushes here which produces conflicts ? It's hard to collaborare/incorporate changes that way ;) Thanks

dcermak:
> @dcermak can you please review #1853 before making further force pushes here which produces conflicts ? It's hard to collaborare/incorporate changes that way ;) Thanks

I'd like to get the CI working first before throwing in more changes, otherwise that all requires manual verification.

doc/source/contributing/scripts_testing.rst
```diff
@@ -0,0 +1,83 @@
...
+This can be achieved by leveraging pytest's `fixture parametrization
+<https://docs.pytest.org/en/stable/parametrize.html>`__ as follows:
+
+.. code:: python
+
+    @pytest.mark.parametrize(
+        "shared_container",
+        (
+            "Tumbleweed",
+            "Leap-15.2",
```
schaefi, line 70: Can we avoid using distro names that will be outdated and some point in the future ? I see two options:

1. Just stick with Tumbleweed
2. Use a variable for Leap-15.2 in ```doc/source/conf.py``` in the ```prolog_replacements``` and reference it here as {...}

dcermak: This should be fixed now. I've reused the variable `exc_os_version` as there's not much to parametrize if you have only one OS :wink:

PR #1853, opened 2021-06-18
Update for functions integration tests

schaefi:
* Move dependencies from tox into the .virtualenv.dev-requirements.txt
  This allows to re-use tox environments for other tasks
* Add debug logging information to conftest.py
* Integrate scripts to linter checks
* Fixup code smells

### DIFF
--- a/.virtualenv.dev-requirements.txt
+++ b/.virtualenv.dev-requirements.txt
@@ -50,3 +50,7 @@ python-dateutil
 
 # for release
 twine
+
+# for scripts testing
+pytest-testinfra
+dataclasses ; python_version < "3.7"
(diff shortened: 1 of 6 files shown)


schaefi:
@dcermak Hmm, any reason you changed your branch prior looking at the changes from here ?

dcermak:
> @dcermak Hmm, any reason you changed your branch prior looking at the changes from here ?

Yes, I would like to get the CI to work first instead of having to manually test every change.

.virtualenv.dev-requirements.txt
```diff
@@ -50,3 +50,7 @@ python-dateutil
 
 # for release
 twine
+
+# for scripts testing
```
dcermak, line 54: This will cause testinfra & dataclasses to get pulled in into every development virtualenv. Unless you want that, I'd keep them separated as they are currently.

PR #1855, opened 2021-06-18
Raise XML related error on syntax error

rjschwei:
At present kiwi raises a KiwiConfigFileFormatNotSupported error when
an XML file is used that contains a syntax error. The message about
needing Python anymarkup is confusing. If the file failing to be
parsed is indeed XML as indicated by <?xml at the beginning of the file
we now raise a syntax error.

### DIFF
--- a/kiwi/markup/base.py
+++ b/kiwi/markup/base.py
@@ -20,7 +20,10 @@ from lxml import etree
 
 # project
 from kiwi.defaults import Defaults
-from kiwi.exceptions import KiwiConfigFileFormatNotSupported
+from kiwi.exceptions import (
+    KiwiConfigFileFormatNotSupported,
+    KiwiConfigSyntaxError
+)
 
 
 class MarkupBase:
@@ -55,6 +58,12 @@ class MarkupBase:
         try:
             parsed_description = etree.parse(description)
         except etree.XMLSyntaxError:
+            with open(description) as image_description:
+                content = image_description.readlines()
+                if content[0].startswith('<?xml '):
+                    raise KiwiConfigSyntaxError(
+                        'The configuration file contains an XML syntax error '
+                        'and the file could not be parsed.')
             raise KiwiConfigFileFormatNotSupported(
                 'Support for non-XML formatted config files requires '
                 'the Python anymarkup module.')
--- a/kiwi/exceptions.py
+++ b/kiwi/exceptions.py
@@ -194,6 +194,13 @@ class KiwiConfigFileFormatNotSupported(KiwiError):
     """
 
 
+
+class KiwiConfigSyntaxError(KiwiError):
+    """
+    Exception raised if kiwi description is in XML but cannot be parsed.
+    """
+
+
 class KiwiContainerSetupError(KiwiError):
     """
     Exception raised if an error in the creation of the
--- a/test/data/example_no_xml_config.xml
+++ b/test/data/example_no_xml_config.xml
@@ -1,0 +1,45 @@
+<image schemaversion="7.4" name="LimeJeOS-openSUSE-13.2">
+    <description type="system">
+        <author>Marcus Schäfer</author>
+        <contact>ms@suse.com</contact>
+        <specification>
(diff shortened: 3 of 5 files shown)


schaefi:
Hi, thanks for looking into this. Before I share my thoughts there are the following ```tox``` issues reported:

```
___________ TestMarkupBase.test_apply_xslt_stylesheets_syntax_error ____________
[gw0] linux -- Python 3.6.13 /home/ms/__rjschwei/kiwi8/.tox/3.6/bin/python

>           with open(description) as image_description:
E           FileNotFoundError: [Errno 2] No such file or directory: '../data/xample_syntaxerror_config.xml'
```

See the typo in the name

and 

```
check run-test: commands[0] | flake8 --statistics -j auto --count /home/ms/__rjschwei/kiwi8/kiwi
kiwi/exceptions.py:198:1: E303 too many blank lines (3)
1     E303 too many blank lines (3)
1
ERROR: InvocationError for command /home/ms/__rjschwei/kiwi8/.tox/3/bin/flake8 --statistics -j auto --count kiwi (exited with code 1)
ERROR:   check: commands failed
```

schaefi:
The CI does not run on forks but you can easily run all checks locally by calling ```tox```

kiwi/markup/base.py
```diff
@@ -55,6 +58,12 @@ def apply_xslt_stylesheets(self, description: str) -> str:
         try:
             parsed_description = etree.parse(description)
         except etree.XMLSyntaxError:
+            with open(description) as image_description:
+                content = image_description.readlines()
+                if content[0].startswith('<?xml '):
```
schaefi, line 63: I see several issues with this "XML input type detection". Please also note we had a relatively long conversation with Fergel at the time when we added the message for this exception. Without anymarkup in place it's not easy to detect the markup type of the given input and we actually wanted to use the filetype magic for the detection. But also the filetype magic has potential to fail.

The way you solved this unfortunately exposes similar problems. Here are the ones I have tested

1. ```<?xml>``` tag is an optional tag and not mandatory for XML input. It's also written that way in the XML documentation. Thus if you simply drop that line it can still be valid or invalid XML or anything else and your check has no opportunity to check that

2. The expectation that the first line is the one with the ```<?xml>``` tag makes this less robust. The header information must be first prior any markup data but there can be spaces/emptylines or any other sort of non XML markup related characters in the input stream. In your case just adding an empty line before the tag breaks the check. I think that part can be fixed easily

I think what is being tried here, to know for sure that the intention of the user was to validate against XML, cannot be solved in a generic way. This is also the reason why we did not add more or less sophisticated ways to try the guessing game.

However, I agree with you that the error message is misleading if the XML input is invalid XML and expected to be XML. Wouldn't it be an acceptable change to just give the error message more scope ?

Something like

```python
message = dedent('''\n
    The configuration file was validated against the XML markup and failed
    with a syntax error. Please rerun kiwi with --debug to get further details
    on the error if your configuration is intended to be in XML. In case your
    configuration is using another markup language please note that this
    requires the installation of the Python anymarkup module.
''')
try:
    parsed_description = etree.parse(description)
except etree.XMLSyntaxError as issue:
    log.debug(issue)
    raise KiwiConfigFileFormatNotSupported(message)
```

If you would consider this as an option we only need to change the code to print the etree.parse syntax error message from the exception as a log.debug message.

From my perspective this is a good compromise to the user friendly aspect and also not misleading

Thoughts ?

schaefi (changes requested):

kiwi/markup/base.py
schaefi: I'm sure you can also come up with a much better message but I hope you get the idea ?

schaefi:
There is also a small style change which I'd like to ask you to update.

The commit message starts with a "-" in the headline. We don't do this in kiwi and it also seems not common practice for the git history format. It does not hurt at all but I like consistency at this point. So if you could get rid of the leading "-" it would be great. Thanks

rjschwei:
Superseded by #1859

Issue #1856, opened 2021-06-20
dnf package list is passed unqouted to a shell


keszybz:
## Problem description
When a package name with shell-special characters is used, dnf invocation fails:

```xml  
<packages type="bootstrap">
        <package name="(glibc-minimal-langpack if glibc)" />
</packages>
```

```console
$ sudo kiwi-ng system build --allow-existing-root --description kiwi-descriptions/fedora/x86_64/fedora-34.0/ …
...
[ ERROR   ]: 09:03:06 | KiwiBootStrapPhaseFailed: Bootstrap package installation failed: bash: -c: line 1: syntax error near unexpected token `('
bash: -c: line 1: `dnf --config /var/tmp/f34/build/image-root/tmpf9zy5lqq -y --installroot /var/tmp/f34/build/image-root --setopt=install_weak_deps=False install (glibc-minimal-langpack if glibc) basesystem dnf filesystem grub2-efi-x64 grub2-efi-x64-modules shim-x64 syslinux syslinux-efi64'
```

Please don't use `shell=True`.

kiwi-cli-9.23.34-1.fc34.noarch

schaefi:
Actually quite interesting why it is done that way, we don't run the other package managers through a shell. I'll change it and send feedback asap

schaefi:
This affects the dnf and microdnf and pacman package managers. zypper and apt are not affected

schaefi:
@keszybz @Conan-Kudo Hi. When looking at how we call dnf I realized why it's done through a shell. Actually installing with dnf can call dnf twice if groups are used. In kiwi we then call

* dnf install ...
* dnf group install ...

Because it's two times dnf and we want to fetch and manage output in one process we call it with

* bash -c cmd_1 && cmd_2

Are you aware of an opportunity how to call dnf in a way that it does both actions in one dnf call with parameters ?

Thanks

Conan-Kudo:
All DNF groups can be referred to as package names too. For example: `dnf install @kde-desktop-environment`.

These names show up with `dnf group list -v`:

```
ngompa@localhost ~> sudo dnf group list -v
Loaded plugins: builddep, changelog, config-manager, copr, debug, debuginfo-install, download, generate_completion_cache, groups-manager, needs-restarting, playground, repoclosure, repodiff, repograph, repomanage, reposync, system-upgrade
DNF version: 4.7.0
cachedir: /var/cache/dnf
[...]

PR #1857, opened 2021-06-21
Prevent calling pacman through a shell

schaefi:
There is no reason to shell out for calling pacman.
In fact it can be counter productive if the shell
evaluates eventually existing package name/instruction
patterns. This is related to Issue #1856

### DIFF
--- a/kiwi/package_manager/pacman.py
+++ b/kiwi/package_manager/pacman.py
@@ -114,7 +114,7 @@ class PackageManagerPacman(PackageManagerBase):
                 '--root', self.root_dir, '-Sy'
             ]
         )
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
This is interesting. I have no idea why we coded it that way... probably related to some stdout parsing limitation that we had at some point? All these are used in `Command.call` context and stdout is then parsed (even we are no doing any proper parsing for pacman).

Anyway I am absolutely fine with this change :+1: we just need to keep an eye to the integration tests afterwards to double check it doesn't fail neither produces some strange output, even I can't see how this could happen.

PR #1858, opened 2021-06-21
Do not shell out for calling microdnf.

schaefi:
In fact it can be counter productive if the shell
evaluates eventually existing package name/instruction
patterns. This is related to Issue #1856


davidcassany (approved):
LGTM

Conan-Kudo (approved):

PR #1864, opened 2021-06-23
Add missing util-linux-systemd dependency to dracut-kiwi-overlay

dcermak:
The script kiwi-overlay-root.sh requires lsblk which is provided by
util-linux-systemd. If that package is missing in the final image, then booting
an overlayroot image hangs with:
```
dracut-pre-mount[480]: //lib/dracut/hooks/pre-mount/30-kiwi-overlay-root.sh: line 46: lsblk: command not found
```

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -530,6 +530,10 @@ BuildRequires:  dracut
 %endif
 Requires:       dracut-kiwi-lib = %{version}-%{release}
 Requires:       util-linux
+# lsblk is part of util-linux-systemd on openSUSE
+%if 0%{?suse_version}
+Requires:       util-linux-systemd
+%endif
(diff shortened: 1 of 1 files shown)


package/python-kiwi-spec-template
```diff
@@ -530,6 +530,7 @@ BuildRequires:  dracut
 %endif
 Requires:       dracut-kiwi-lib = %{version}-%{release}
 Requires:       util-linux
+Requires:       util-linux-systemd
```
Conan-Kudo, line 533: This is a SUSE-only package, please guard it with `%if 0%{?suse_version}`.

dcermak: Done

Conan-Kudo (approved):

PR #1869, opened 2021-06-28
Use namespaced files in /var/tmp for temporary files

### LINKED ISSUE #1866: Kiwi fails to create image because no space on /tmp
## Problem description

While creating an image kiwi fails with 

```
ERROR: 14:04:17 | KiwiCommandError: mksquashfs: stderr: 
Write failed because No space left on device
```
## Expected behaviour
Image builds and I get an ISO
On a bare metal machine just try to build ISO with the following command 
```kiwi-ng system build --description ./iso-build --target-dir ./tmpoutput```

+ KIWI version: 9.23.34
+ Operating system:  Fedora 33
+ OBS version:  N/A

[kiwi-error.log](https://github.com/OSInside/kiwi/files/6718140/kiwi-error.log)

Conan-Kudo:
Previously, kiwi created the squashfs as a plain temporary file
in /tmp, which causes issues on operating systems where /tmp is tmpfs.
Notably, image builds would fail with "no space left on the device"
because the tmpfs was not big enough for everything to exist there.

To fix this, we change to use /var/tmp, and additionally add a prefix
for our temporary files so that the user knows which ones kiwi created.

Fixes #1866.

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -742,7 +742,7 @@ class DiskBuilder:
 
         if self.root_filesystem_is_overlay:
             log.info('--> creating readonly root partition')
-            squashed_root_file = NamedTemporaryFile()
+            squashed_root_file = NamedTemporaryFile(dir='/var/tmp', prefix='kiwi-')
             squashed_root = FileSystemSquashFs(
                 device_provider=DeviceProvider(), root_dir=self.root_dir,
                 custom_args={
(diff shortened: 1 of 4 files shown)


davidcassany (approved):
Looks good to me, I guess there is no way to avoid the bandit warning. I can't think of an alternative that does not imply having to hardcode `/var/tmp`. I don't see a benefit on hardcoding it into some environment variable or any similar approach.

davidcassany:
This works as this specific solution, but a quick search on `NamedTemporaryFile` in kiwi code reveals it is being called in many places. We should review we are not using it for heavy files without forcing the use of some disk directory such as `/var/tmp`. Creating a new card for that.

Conan-Kudo:
@davidcassany I added a couple of other places I found to this ticket, but we should comprehensively review usage of `NamedTemporaryFile()` and come up with a unified handler for this.

PR #1873, opened 2021-07-01
Fixed fedora integration test builds

schaefi:
Maintain the repos in the obs prj config which prevents
the weird "nothing provides kernel-obs-build" error


schaefi:
* https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_x86:fedora

finally building again

davidcassany (approved):
If the build works this way this definitely the best approach in OBS so we get the project configuration inheritence.

A side benefit of it is that is that it actively tracks if Fedora is kept as a first class citizen in OBS world, meaning that Fedora projects are properly configured so that they can be smoothly used.

Issue #1876, opened 2021-07-05
suseImportBuildKey is not a NOOP


dcermak:
## Problem description

The function `suseImportBuildKey` got recently deprecated by https://github.com/OSInside/kiwi/pull/1833 but this function was **not** a NOOP. In fact, removing this function completely breaks images build by kiwi, as they lack any signing keys on the root filesystem and thus every `zypper in` fails as the repository signing keys are not present.

I have removed `suseImportBuildKey` from my vagrant boxes on OBS and now my integration testing pipeline is completely red, because it tries a `zypper in etags` and `zypper up`, failing with:
```ShellSession
$ zypper --non-interactive refresh
Retrieving repository 'openSUSE-Tumbleweed-Non-Oss' metadata [.
.


New repository or package signing key received:

  Repository:       openSUSE-Tumbleweed-Non-Oss
  Key Fingerprint:  22C0 7BA5 3417 8CD0 2EFE 22AA B88B 2FD4 3DBD C284
  Key Name:         openSUSE Project Signing Key <opensuse@opensuse.org>
  Key Algorithm:    RSA 2048
  Key Created:      Mon 05 May 2014 10:37:40 AM CEST
  Key Expires:      Thu 02 May 2024 10:37:40 AM CEST
  Rpm Name:         gpg-pubkey-3dbdc284-53674dd4

    Note: Signing data enables the recipient to verify that no modifications occurred after the data
    were signed. Accepting data with no, wrong or unknown signature can lead to a corrupted system
    and in extreme cases even to a system compromise.
    Note: A GPG pubkey is clearly identified by it's fingerprint. Do not rely the keys name. If you
    are not sure whether the presented key is authentic, ask the repository provider or check his
    web site. Many provider maintain a web page showing the fingerprints of the GPG keys they are
    using.
Do you want to reject the key, trust temporarily, or trust always? [r/t/a/?] (r): r
error]
Repository 'openSUSE-Tumbleweed-Non-Oss' is invalid.
[repo-non-oss|http://download.opensuse.org/tumbleweed/repo/non-oss/] Valid metadata not found at specified URL
History:
 - Signature verification failed for repomd.xml
 - Can't provide /repodata/repomd.xml

Please check if the URIs defined for this repository are pointing to a valid repository.
Skipping repository 'openSUSE-Tumbleweed-Non-Oss' because of the above error.
Retrieving repository 'openSUSE-Tumbleweed-Oss' metadata [.


New repository or package signing key received:

[...]

PR #1877, opened 2021-07-05
Re-add suseImportBuildKey

dcermak:
`suseImportBuildKey` is not required during the image build as kiwi imports the
correct keys by itself. However, the created images lack the repository signing
keys and any `zypper` commands will thus fail.

This fixes https://github.com/OSInside/kiwi/issues/1876

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -123,6 +123,50 @@ function baseService {
     fi
 }
 
+#======================================
+# suseImportBuildKey
+#--------------------------------------
+function suseImportBuildKey {
+    # /.../
+    # Add missing gpg keys to rpm database
+    # ----
+    local KEY
+    local TDIR
+    local KFN
+    local dumpsigs=/usr/lib/rpm/gnupg/dumpsigs
+    TDIR=$(mktemp -d)
+    if [ ! -d "${TDIR}" ]; then
+        echo "suseImportBuildKey: Failed to create temp dir"
+        return
+    fi
+    if [ -d "/usr/lib/rpm/gnupg/keys" ];then
+        pushd "/usr/lib/rpm/gnupg/keys" || return
+    else
+        pushd "${TDIR}" || return
+        if [ -x "${dumpsigs}" ];then
+            ${dumpsigs} /usr/lib/rpm/gnupg/suse-build-key.gpg
+        fi
+    fi
+    for KFN in gpg-pubkey-*.asc; do
+        if [ ! -e "${KFN}" ];then
+            #
+            # check if file exists because if the glob match did
+            # not find files bash will use the glob string as
+            # result and we just continue in this case
+            #
+            continue
+        fi
+        KEY=$(basename "${KFN}" .asc)
+        if rpm -q "${KEY}" >/dev/null; then
+            continue
+        fi
+        echo "Importing ${KEY} to rpm database"
+        rpm --import "${KFN}"
+    done
+    popd || return
+    rm -rf "${TDIR}"
+}
+
 function baseStripLocales {
     local keepLocales="$*"
     find /usr/lib/locale -mindepth 1 -maxdepth 1 -type d 2>/dev/null |\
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
This seems more like we're doing the wrong thing for configuring repos. We shouldn't rely on this for Zypper to work correctly.

dcermak:
> This seems more like we're doing the wrong thing for configuring repos. We shouldn't rely on this for Zypper to work correctly.

That is correct. But once the version of kiwi without suseImportBuildKey lands in Factory, it will break a lot of appliance builds in a subtle way. So I'd propose that we first restore the old behavior and then fix it properly, before we break a lot of appliances on OBS.

Conan-Kudo:
> > This seems more like we're doing the wrong thing for configuring repos. We shouldn't rely on this for Zypper to work correctly.
> 
> That is correct. But once the version of kiwi without suseImportBuildKey lands in Factory, it will break a lot of appliance builds in a subtle way. So I'd propose that we first restore the old behavior and then fix it properly, before we break a lot of appliances on OBS.

Do you think that the appliances will actually get fixed unless they're broken first?

schaefi:
I just don't understand why we need this code when this here:

```
--gpg-auto-import-keys  Automatically trust and import new repository
                            signing keys.
```

is expected to do exactly that, or am I completely wrong with this assumption ? We call zypper with that option and my expectation was that it does what the script here does

dcermak:
> I just don't understand why we need this code when this here:
> 
> ```
> --gpg-auto-import-keys  Automatically trust and import new repository
>                             signing keys.
> ```
> 
> is expected to do exactly that, or am I completely wrong with this assumption ? We call zypper with that option and my expectation was that it does what the script here does

Because `suseImportBuildKey` inserts the rpm signing keys into the final image at build time. So if you invoke zypper inside the final image, it "just works". Without this function, **everyone** will have to add the `--gpg-auto-import-keys` flag later on and they should actually check that zypper fetched the correct keys…

That's imho a pretty bad UX if you ask me.

Vogtinator:
Yeah, `suseImportBuildKey` needs to stay as-is at least for now. `gpg-auto-import-keys` is not an alternative at all, it makes it highly insecure because it just accepts whatever key the repository has without any checks at all!

> This seems more like we're doing the wrong thing for configuring repos. We shouldn't rely on this for Zypper to work correctly.

The `openSUSE-build-key` package contains the GPG keys for repos and packages, but doesn't import them on installation.

Vogtinator:
> I just don't understand why we need this code when this here:
> 
> ```
> --gpg-auto-import-keys  Automatically trust and import new repository
>                             signing keys.
> ```
> 
> is expected to do exactly that, or am I completely wrong with this assumption ? We call zypper with that option and my expectation was that it does what the script here does

If you pass that to zypper in OBS, then it won't actually do anything because the local package repos in the build root are not signed.

Conan-Kudo:
> Yeah, `suseImportBuildKey` needs to stay as-is at least for now. `gpg-auto-import-keys` is not an alternative at all, it makes it highly insecure because it just accepts whatever key the repository has without any checks at all!
> 
> > This seems more like we're doing the wrong thing for configuring repos. We shouldn't rely on this for Zypper to work correctly.
> 
> The `openSUSE-build-key` package contains the GPG keys for repos and packages, but doesn't import them on installation. That's what `suseImportBuildKey` currently does.

Conan-Kudo:
We also don't specify the GPG key file in the manifest, which would be ideal for resolving this properly. I generally don't like the idea of just mass-importing keys...

schaefi:
> Yeah, suseImportBuildKey needs to stay as-is at least for now.

ok, so let's not mess with it, get back to the working state and revisit this topic on the community meeting. I think it's pretty odd if the install of a key package has no effect and is also not effectively used by the package manager unless there is some extra code moving the bits around. Anyway I'll merge this here, thanks much @dcermak and maybe someone can explain the deeper design aspects of the build key provision at the meeting next week

PR #1878, opened 2021-07-05
Fixed grub cmdline setup with custom root

### LINKED ISSUE #1875: kernel commandline is truncated when it contains 'root='
## Problem description
When the `kernelcmdline` attribute contains `root=` to set the root device kiwi will truncat the command line written to /etc/default/grub and any command line arguments given after root= will be omitted from `GRUB_CMDLINE_LINUX_DEFAULT`

## Expected behaviour
No truncation. All specified kernel command line arguments form the XML should be added to `GRUB_CMDLINE_LINUX_DEFAULT`

Build an image with `root=` and some argument after in the `kernelcmdline` attribute. The resulting image will hav ethe settings after `root=` missing in `/etc/default/grub`

+ KIWI version: 
+ Operating system: SLES 15 SP3
+ OBS version:

schaefi:
If the kiwi kernelcmdline attribute contains root=... information
it is extracted from being written to GRUB_CMDLINE_LINUX_DEFAULT.
However, the regexp to extract the root=... information was broken
and deleted more elements of the cmdline information than just
the root device information. This commit fixes the regexp to only
delete the root=... information taking into account that every
kernel parameter is delimited by '\s'
This Fixes #1875

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -641,7 +641,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             'GRUB_TERMINAL': '"{0}"'.format(self.terminal)
         }
         grub_final_cmdline = re.sub(
-            r'root=.* |root=.*$', '', self.cmdline
+            r'(^root=[^\s]+)|( root=[^\s]+)', '', self.cmdline
         ).strip()
         if self.persistency_type != 'by-uuid':
             grub_default_entries['GRUB_DISABLE_LINUX_UUID'] = 'true'


kiwi/bootloader/config/grub2.py
```diff
@@ -641,7 +641,7 @@ def _setup_default_grub(self):
             'GRUB_TERMINAL': '"{0}"'.format(self.terminal)
         }
         grub_final_cmdline = re.sub(
-            r'root=.* |root=.*$', '', self.cmdline
+            r'root=[^\s]+', '', self.cmdline
```
rjschwei, line 644: I think the expression should include the ' ' (space) we expect in front of the `root` attribute. Otherwise if there is ever another attribute that ends with `root` it will also get substituted. For example

```
>>> a = 'bootroot=/dev/foo root=/dev/bar hi'
>>> re.sub(r'root=[^\s]+', '', a)
'boot  hi'
>>> re.sub(r' root=[^\s]+', '', a)
'bootroot=/dev/foo hi'
```

schaefi: yes good point, thanks much :+1:  I'll adapt

schaefi: done

rjschwei (commented):
LGTM

PR #1879, opened 2021-07-06
Fixed secure boot fallback setup

schaefi:
Make sure MokManager gets copied. This Fixes bsc#1187515

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -750,6 +750,11 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                     os.sep.join([self.efi_boot_path, grub_image.binaryname])
                 ]
             )
+            mok_manager = Defaults.get_mok_manager(lookup_path)
+            if mok_manager:
+                Command.run(
+                    ['cp', mok_manager, self.efi_boot_path]
+                )
         else:
             # Without shim a self signed grub image is used that
             # gets loaded by the firmware
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -710,6 +710,31 @@ class Defaults:
                 return shim_file
 
     @staticmethod
+    def get_mok_manager(root_path: str) -> Optional[str]:
+        """
+        Provides Mok Manager file path
+
+        Searches distribution specific locations to find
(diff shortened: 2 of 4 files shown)


schaefi:
@Conan-Kudo I think MokManager.efi is a stable name across distributions and afaik also part of the EFI standard. So do you agree with me that we don't have to differentiate the name between the distros ?

Thanks

Vogtinator:
> @Conan-Kudo I think `MokManager.efi` is a stable name across distributions and afaik also part of the EFI standard. So do you agree with me that we don't have to differentiate the name between the distros ?

The name of `MokManager.efi` is configured during build of `shim`, the default is actually `mm$(ARCH).efi`...
openSUSE uses `MokManager.efi` (without any arch suffix), no idea about others.

Conan-Kudo:
We should probably fix `MokManager.efi` to go back to the default name with the arch in it...

kiwi/bootloader/config/grub2.py
```diff
@@ -750,6 +750,12 @@ def _setup_secure_boot_efi_image(self, lookup_path, uuid=None, mbrid=None):
                     os.sep.join([self.efi_boot_path, grub_image.binaryname])
                 ]
             )
+            mok_manager = os.sep.join(
+                [os.path.dirname(shim_image), 'MokManager.efi']
```
Conan-Kudo, line 754: Both Fedora and Ubuntu use `mm<EFIARCH>.efi` instead of `MokManager.efi`.

Conan-Kudo (changes requested):

kiwi/bootloader/config/grub2.py
schaefi: Uhm, ok yet another translation table...

schaefi: thanks

schaefi: done now

schaefi: @Conan-Kudo this should do the trick. if suse switches the name to use what mm<EFIARCH>.efi the match table will continue to apply

Conan-Kudo (approved):

PR #1881, opened 2021-07-07
Fix vagrant usr etc split

dcermak:
In Tumbleweed we switched from shipping sshd's config file in /etc to /usr/etc, but `baseVagrantSetup` was not respecting that.

Now we try to do "the right thing" depending on where the sshd config file exists. Also, this PR adds test coverage for that function.

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -545,6 +545,7 @@ function baseVagrantSetup {
     # insert the default insecure ssh key from here:
     # https://github.com/hashicorp/vagrant/blob/master/keys/vagrant.pub
     mkdir -p /home/vagrant/.ssh/
+    chmod 0700 /home/vagrant/.ssh/
     echo "ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEA6NF8iallvQVp22WDkTkyrtvp9eWW6A8YVr+kz4TjGYe7gHzIw+niNltGEFHzD8+v1I2YJ6oXevct1YeS0o9HZyN1Q9qgCgzUFtdOKLv6IedplqoPkcmF0aYet2PkEDo3MlTBckFXPITAMzF8dJSIFo9D8HfdOV0IAdx4O7PtixWKn5y2hMNG0zQPyUecp4pzC6kivAIhyfHilFR61RGL+GPXQ2MWZWFYbAGjyiYJnAmCP3NOTd0jMZEnDkbUvxhMmBYSdETk1rRgm+R4LOzFUGaHqHDLKLX+FIPKcF96hrucXzcWyLbIbEgE98OHlnVYCzRdK8jlqm8tehUc9c9WhQ== vagrant insecure public key" > /home/vagrant/.ssh/authorized_keys
     chmod 0600 /home/vagrant/.ssh/authorized_keys
     chown -R vagrant:vagrant /home/vagrant/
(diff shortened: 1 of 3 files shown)


Conan-Kudo (changes requested):
As `/usr/etc` is a SUSE-ism, I don't want it in the `base*` functions.

Conan-Kudo:
Also, you can just write drop-in files to `/etc/ssh/sshd_config.d` instead of editing `sshd_config`.

dcermak:
> Also, you can just write drop-in files to `/etc/ssh/sshd_config.d` instead of editing `sshd_config`.

Good idea, that actually makes the code even simpler!

kiwi/config/functions.sh
```diff
@@ -545,13 +545,16 @@ function baseVagrantSetup {
     # insert the default insecure ssh key from here:
     # https://github.com/hashicorp/vagrant/blob/master/keys/vagrant.pub
     mkdir -p /home/vagrant/.ssh/
+    chmod 0700 /home/vagrant/.ssh/
     echo "ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEA6NF8iallvQVp22WDkTkyrtvp9eWW6A8YVr+kz4TjGYe7gHzIw+niNltGEFHzD8+v1I2YJ6oXevct1YeS0o9HZyN1Q9qgCgzUFtdOKLv6IedplqoPkcmF0aYet2PkEDo3MlTBckFXPITAMzF8dJSIFo9D8HfdOV0IAdx4O7PtixWKn5y2hMNG0zQPyUecp4pzC6kivAIhyfHilFR61RGL+GPXQ2MWZWFYbAGjyiYJnAmCP3NOTd0jMZEnDkbUvxhMmBYSdETk1rRgm+R4LOzFUGaHqHDLKLX+FIPKcF96hrucXzcWyLbIbEgE98OHlnVYCzRdK8jlqm8tehUc9c9WhQ== vagrant insecure public key" > /home/vagrant/.ssh/authorized_keys
     chmod 0600 /home/vagrant/.ssh/authorized_keys
     chown -R vagrant:vagrant /home/vagrant/
 
+    SSHD_CONFIG=/etc/ssh/sshd_config.d/99-vagrant.conf
```
Conan-Kudo, line 553: You'll need to test for the directory, since it doesn't exist in RHEL 7, RHEL 8, SLE 12, or SLE 15. When it doesn't exist, fallback to the old behavor of extending `/etc/ssh/sshd_config`.

Conan-Kudo (changes requested):

kiwi/config/functions.sh
dcermak: Right, that makes sense. I've changed the behavior accordingly and added a test on CentOS7 just to be sure, which surfaced another bug in this function as well (the custom settings need to get prepended, not appended…).

Conan-Kudo (changes requested):

test/scripts/conftest.py
```diff
@@ -74,6 +74,7 @@ def cleanup_container(runner_binary: str, container_id: str) -> None:
     Container(
         name="SLE-12-SP5", url="registry.suse.com/suse/sles12sp5:latest"
     ),
+    Container(name="centos:7", url="quay.io/centos/centos:7"),
```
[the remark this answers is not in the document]
dcermak: Done

Conan-Kudo (approved):

PR #1886, opened 2021-07-14
Increase size of efi fat image

schaefi:
For ISO images an embedded efi fat image is needed to boot.
As consequence of adding the mok manager it can happen that
the size of the efi fat image is too small. With this commit
the size is increased to prevent an out of space issue

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -824,7 +824,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             [self.boot_dir + '/boot/', self.arch, '/efi']
         )
         Command.run(
-            ['qemu-img', 'create', efi_fat_image, '15M']
+            ['qemu-img', 'create', efi_fat_image, '20M']
         )
         Command.run(
             ['mkdosfs', '-n', 'BOOT', efi_fat_image]
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -300,7 +300,7 @@ class TestBootLoaderConfigGrub2:
         assert mock_command.call_args_list == [
             call(
                 [
-                    'qemu-img', 'create', 'root_dir/boot/x86_64/efi', '15M'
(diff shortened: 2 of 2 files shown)


schaefi:
See Virtualization:Appliances:Images:Testing_arm:fedora/test-image-live failing for this reason

kiwi/bootloader/config/grub2.py
```diff
@@ -824,7 +824,7 @@ def _create_embedded_fat_efi_image(self):
             [self.boot_dir + '/boot/', self.arch, '/efi']
         )
         Command.run(
-            ['qemu-img', 'create', efi_fat_image, '15M']
+            ['qemu-img', 'create', efi_fat_image, '20M']
```
Conan-Kudo, line 827: Typically, Fedora makes 100MB filesystems by default (to support multi-arch and multi-boot using one ESP), do we want to consider that?

Cf. https://pagure.io/fedora-kickstarts/blob/611edda1490ebb93d996acb29692805ec9e481d3/f/fedora-cloud-base.ks#_42

Conan-Kudo (changes requested):

kiwi/bootloader/config/grub2.py
schaefi: I think we are talking about two different situations.

The one you refer to is the size of the EFI fat partition on "normal" systems which comes with a writable EFI space. I agree with you here on the size constraints and in kiwi we can even configure that size via the efipartsize attribute

The situation I'm talking about is booting ISO (readonly) images via EFI. As you might know EFI hard requires a fat partition. Therefore it's not enough to create a EFI directory on the ISO filesystem. To make it EFI bootable you have to create a fat filesystem image and crate the EFI data structure inside of it. This is what we call the efi_fat_image. The file lives on the ISO filesystem and is therefore immutable and readonly. Thus the size of this efi_fat_image must only be as big as it is required to hold the data that is placed into it. That's also the reason why this is a fixed value in code.

Hope this makes sense ?

Conan-Kudo: Ah, yes, makes sense!

Conan-Kudo (approved):

PR #1893, opened 2021-07-22
Moving temp data handling to its own namespace

### LINKED ISSUE #1870: Avoid default use of `NamedTemporaryFile()` for big files
## Problem description

`NamedTemporaryFile()` is likely to place files in `/tmp` and this temporary directory is curretly mounted on tmpfs in some distros,  including Fedora, openSUSE, etc. Using `/tmp` for huge temporary files (e.g. filesystem images) is likely to consume all the available space.

## Expected behaviour

In KIWI there should be a criteria about how to make use of `NamedTemporaryFiles` so that we can control big files are not placed over tmpfs filesystems. #1869 fixes it for live images, however temporary files are used in places, the same error is likely to appear for other use cases.

Related to #1866

schaefi:
Moving use of mkdtemp, NamedTemporaryFile and TemporaryDirectory
into its own class called Temporary: By default all temporary
data is created below /var/tmp but can be changed via the
global commandline option --temp-dir. This Fixes #1870

### DIFF
--- a/kiwi/utils/temporary.py
+++ b/kiwi/utils/temporary.py
@@ -1,0 +1,49 @@
+# Copyright (c) 2021 SUSE Linux GmbH.  All rights reserved.
+#
+# This file is part of kiwi.
+#
+# kiwi is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# kiwi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kiwi.  If not, see <http://www.gnu.org/licenses/>
+#
+from typing import IO
+
+from tempfile import (
+    NamedTemporaryFile,
+    TemporaryDirectory
+)
+
+# project
+import kiwi.defaults as defaults
+
+
+class Temporary:
+    """
+    **Provides namespace to handle temporary files and directories**
+    """
+    def __init__(
+        self, dir: str = defaults.TEMP_DIR, prefix: str = '',
+        delete: bool = True
+    ):
+        self.prefix = f'kiwi_{prefix}' if prefix else 'kiwi_'
+        self.delete = delete
+        self.dir = dir
+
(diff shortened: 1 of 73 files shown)


schaefi:
This is a pretty big refactoring. It basically introduces a new class in the kiwi namespace called ```Temporary``` From there two methods new_dir() and new_file() exists which I applied across the entire code base including all the relevant tests. That's why the change is so big. I did not saw a way to make that a smaller PR, let's hope the review is not a too much pita

I will do some builds tests with different --temp-dirs and see if all works as expected

Conan-Kudo (approved):

kiwi/utils/temporary.py
```diff
@@ -0,0 +1,49 @@
...
+import kiwi.defaults as defaults
+
+
+class Temporary:
+    """
+    **Provides namespace to handle temporary files and directories**
+    """
+    def __init__(
+        self, dir: str = defaults.TEMP_DIR, prefix: str = '',
+        delete: bool = True
```
davidcassany, line 35: I am wondering if the delete attribute shouldn't be a parameter in `new_file` method rather part of the Temporary object. My reasoning is simple, this flag does not affect  `new_dir` method, so from an API perspective this might be confusing. Is there any place we are using it with `delete=false`? I hope not 😅 Probably we could omit this parameter and always force the cleanup.

Anyway this is a really minor detail, I can perfectly live with the current implementation :+1:

davidcassany (approved):
I did a quick look to the nature of the changes done in most places and it all looks sane.

I really like this new temporary class and the removal of many cleanup methods we had around :wink:

I have just a nit-picky comment, ignore it if you feel like.

kiwi/utils/temporary.py
Conan-Kudo: Well, we could make it affect directories ourselves in the future, right?

schaefi: I think we don't want the temporary data to stay. Thus I agree with @davidcassany that it might be the best to no expose the delete feature at all. There is one place in the kiwi code that used it, in kiwi/system/setup.py create_recovery_archive() the created tmpfile was created with delete=False. It is done that way because the
generated tmpfile gets moved. If you move the file away prior the NamedTemporaryFile cleanup step it ends in
an exception

```python
import os
from tempfile import NamedTemporaryFile

x = NamedTemporaryFile()

print(x.name)

os.unlink(x.name)
```

I will refactor the code in kiwi as we only have one place where this issue exists

schaefi: @davidcassany pushed two commits to handle this better

PR #1894, opened 2021-07-29
Fixed fstab entry for swap on LVM

schaefi:
If an LVM setup is used together with a swapspace the
swap is created as a volume in the volume group. The
required fstab entry to activate swap was not using
the LVM exposed device but the UUID of the low level
block layer. This low level device is not created
by udev because LVM takes over precedence in this
case.


schaefi:
The PR also updates the lvm test to use a swapspace such that it can be actually tested

PR #1903, opened 2021-08-12
Allow target dir for archive

### LINKED ISSUE #1794: Allow to specify target for archive element
## Problem description

The current ```<archive>``` element does not allow to specify a target directory to unpack the archive
It would be a nice feature if this would be possible

```xml
<archive name="foo.tar.gz" target_dir="etc"/>
```

jesusbv:
- Add the option to specify a target directory
  to unpack the archive

  This Fixes #1794

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -187,9 +187,11 @@ div {
 div {
     k.archive.name.attribute = k.name.attribute
     k.archive.bootinclude.attribute = k.bootinclude.attribute
+    k.archive.target_dir.attribute = k.path.attribute
     k.archive.attlist =
         k.archive.name.attribute &
-        k.archive.bootinclude.attribute?
+        k.archive.bootinclude.attribute? &
+        k.archive.target_dir.attribute?
     k.archive =
         ## Name of an image archive file (tarball)
         element archive {
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -231,6 +231,7 @@ class SystemPrepare:
         bootstrap_collections = self.xml_state.get_bootstrap_collections()
         bootstrap_products = self.xml_state.get_bootstrap_products()
         bootstrap_archives = self.xml_state.get_bootstrap_archives()
+        bootstrap_archives_target_dirs = self.xml_state.get_bootstrap_archives_target_dirs()
         # process package installations
         if collection_type == 'onlyRequired':
             manager.process_only_required()
@@ -265,7 +266,9 @@ class SystemPrepare:
         # process archive installations
         if bootstrap_archives:
             try:
-                self._install_archives(bootstrap_archives)
+                self._install_archives(
+                    bootstrap_archives, bootstrap_archives_target_dirs
+                )
             except Exception as issue:
                 raise KiwiBootStrapPhaseFailed(
                     self.issue_message.format(
@@ -297,6 +300,7 @@ class SystemPrepare:
         system_collections = self.xml_state.get_system_collections()
         system_products = self.xml_state.get_system_products()
         system_archives = self.xml_state.get_system_archives()
+        system_archives_target_dirs = self.xml_state.get_system_archives_target_dirs()
         system_packages_ignored = self.xml_state.get_system_ignore_packages()
         # process package installations
         if collection_type == 'onlyRequired':
@@ -332,7 +336,10 @@ class SystemPrepare:
         # process archive installations
         if system_archives:
             try:
-                self._install_archives(system_archives)
+                self._install_archives(
+                    system_archives,
+                    system_archives_target_dirs
+                )
             except Exception as issue:
                 raise KiwiInstallPhaseFailed(
                     self.issue_message.format(
@@ -496,7 +503,7 @@ class SystemPrepare:
         )
         manager.clean_leftovers()
 
-    def _install_archives(self, archive_list):
+    def _install_archives(self, archive_list, archive_target_dir_dict):
         log.info("Installing archives")
         for archive in archive_list:
             log.info("--> archive: %s", archive)
@@ -517,8 +524,15 @@ class SystemPrepare:
                 archive_file = '/'.join(
                     [derived_description_dir, archive]
                 )
+            target_dir = self.root_bind.root_dir
+            if archive_target_dir_dict.get(archive):
+                target_dir = os.path.join(
+                    target_dir,
+                    archive_target_dir_dict.get(archive)
+                )
+            log.info('--> target dir: %s', target_dir)
             tar = ArchiveTar(archive_file)
-            tar.extract(self.root_bind.root_dir)
+            tar.extract(target_dir)
 
     def _setup_requests(
         self, manager, packages, collections=None, products=None, ignored=None
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -2110,6 +2110,45 @@ class XMLState:
         """
         return self.root_filesystem_uuid
 
+    @staticmethod
+    def get_archives_target_dirs(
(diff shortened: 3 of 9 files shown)


kiwi/xml_state.py
```diff
@@ -2110,6 +2108,37 @@ def get_root_filesystem_uuid(self) -> Optional[str]:
...
+        packages section(s)
+        :return: archive names and its target dir
+        :rtype: dict
+        """
+        result = {}
+        bootstrap_packages_sections = self.get_bootstrap_packages_sections()
+        if bootstrap_packages_sections:
+            for bootstrap_packages_section in bootstrap_packages_sections:
+                for archive in bootstrap_packages_section.get_archive():
+                    result[archive.get_name().strip()] = archive.get_target_dir()
```
jesusbv, line 2123: This could be done in `get_bootstrap_archives` (and same for `get_system_archives`) but
it would change the method, returning a `dict` not a `list` and few extra changes but 
there would be less duplication and also this loop would be done fewer times.

I was not sure about changing those methods but if everybody is ok, more than happy 
to do the update

kiwi/schema/kiwi.rnc
```diff
@@ -187,9 +191,11 @@ div {
 div {
     k.archive.name.attribute = k.name.attribute
     k.archive.bootinclude.attribute = k.bootinclude.attribute
+    k.archive.target_dir.attribute = k.target_dir.attribute
     k.archive.attlist =
         k.archive.name.attribute &
-        k.archive.bootinclude.attribute?
+        k.archive.bootinclude.attribute? &
+	k.archive.target_dir.attribute?
```
schaefi, line 198: A few issues here:

1. There is no need to introduce a new attribute specifier (k.target_dir.attribute). The content for target_dir is a path and for this there is already ```k.path.attribute```
2. small Indentation issue

So the following should be enough:

```
k.archive.target_dir.attribute = k.path.attribute

k.archive.attlist =
        k.archive.name.attribute &
        k.archive.bootinclude.attribute? &
        k.archive.target_dir.attribute?
```

kiwi/system/prepare.py
```diff
@@ -496,7 +503,7 @@ def clean_package_manager_leftovers(self) -> None:
         )
         manager.clean_leftovers()
 
-    def _install_archives(self, archive_list):
+    def _install_archives(self, archive_list, archive_target_dir):
```
schaefi, line 506: The parameter should be imho named ```archive_target_dir_dict```, it's misleading otherwise. Alternatively I'm fine adding annotations even though it's a private method not considered to be used in the API

kiwi/xml_state.py
```diff
@@ -503,10 +503,8 @@ def get_bootstrap_archives(self) -> List:
         bootstrap_packages_sections = self.get_bootstrap_packages_sections()
         if bootstrap_packages_sections:
             for bootstrap_packages_section in bootstrap_packages_sections:
-                archive_list = bootstrap_packages_section.get_archive()
-                if archive_list:
-                    for archive in archive_list:
-                        result.append(archive.get_name().strip())
+                for archive in bootstrap_packages_section.get_archive():
+                    result.append(archive.get_name().strip())
```
schaefi, line 507: I think this change is unrelated to the support for target_dir in ```<archive>``` With the version of generateDS as you used it it's correct that section lists with no value return an empty list [] rather than None. However it's not guaranteed that an older version of generateDS is used to create the XML API and then this change can lead to unexpected failures. This is because if you do this:

```python
x = None

for f in x:
    print(f)
```

you get an error in python. I coded it to be safe in any direction even though that requires one more line of code :) If you don't mind can we drop this change as it's unrelated ?

Thanks

kiwi/xml_state.py
schaefi: yeah I was thinking the same but then realized these are methods of the public API and we cannot change their return value type unless we do a major version upgrade of kiwi. So the way you designed it is great, let's keep that.

However, on the code duplication end the implementation could be improved imho. You can think of something like this:

```python
def get_archives_target_dirs(self, package_section_names: List[str]) -> Dict:
    ...

def get_system_archives_target_dirs(self) -> Dict:
    return self.get_archives_target_dirs(
        self.get_packages_sections(['image', self.get_build_type_name()])
    ) 

def get_bootstrap_archives_target_dirs(self) -> Dict:
    return self.get_archives_target_dirs(
        self.get_packages_sections(['bootstrap'])
    ) 
```

schaefi: Also remember that list types in generateDS created data structures can return None instead of []. This should be taken care in the implementation of ```get_archives_target_dirs()```

schaefi (changes requested):
Thanks much for your effort :+1: 

I like the implementation and commented basically on code duplication and consistency. I've also tested the implementation and it worked for me

Well done :)

jesusbv:
When all ok, I will squash the commit into one

test/unit/system/prepare_test.py
```diff
@@ -299,7 +299,7 @@ def test_install_bootstrap(
         mock_tar.assert_called_once_with(
             '{0}/bootstrap.tgz'.format(self.description_dir)
         )
-        tar.extract.assert_called_once_with('root_dir')
+        tar.extract.assert_called_once_with('root_dir/foo')
```
schaefi, line 302: I think it would be good if the test that extracts an archive without a target_dir continues to stay. That way we can ensure the change does not break the "default" behavior. It would be imho good to just have one additional test to explicitly test the case with target_dir. What do you think ?

schaefi (changes requested):
Thanks for the adaptions, this looks great. I just stumbled over the unit tests which I think should not replace the former test but add a new one.

Finally I also realized there is no documentation about the new attribute. Please add some words here: doc/source/image_description/elements.rst

after that we are good to go :)

schaefi:
There is also a complain from Codacy which makes sense to me

* https://app.codacy.com/gh/OSInside/kiwi/pullRequest?prid=7908064

-> "Method could be a function"

kiwi/xml_state.py
```diff
@@ -2110,6 +2110,44 @@ def get_root_filesystem_uuid(self) -> Optional[str]:
         """
         return self.root_filesystem_uuid
 
+    def get_archives_target_dirs(
+        self, packages_sections_names: Optional[List[xml_parse.packages]]
+    ) -> Dict:
```
schaefi, line 2115: This can be static

```python
@staticmethod
def get_archives_target_dirs(
    packages_sections_names: Optional[List[xml_parse.packages]]
) -> Dict:
```

schaefi: to fix the codacy issue report

jesusbv:
> Finally I also realized there is no documentation about the new attribute. Please add some words here: doc/source/image_description/elements.rst

I updated that documentation

test/unit/system/prepare_test.py
```diff
@@ -304,6 +304,86 @@ def test_install_bootstrap(
...
+        self.manager.request_product.assert_called_once_with(
+            'kiwi'
+        )
+        self.manager.process_install_requests_bootstrap.assert_called_once_with(
+            self.system.root_bind
+        )
+        mock_tar.assert_called_once_with(
+            '{0}/bootstrap.tgz'.format(self.description_dir)
+        )
+        tar.extract.assert_called_once_with('root_dir/foo')
```
schaefi, line 382: This is a 99% copy of the test_install_bootstrap() test. I think it's not needed to have all the other assertions tested again in this test. I suggest to reduce the test to just the relevant line testing for tar.extract and how it is called.

I understand that the copy&paste was the fastest way to make this test but it will strike back if code changes causes adaptions of tests. So please let us avoid the duplication. Makes sense ?

schaefi (approved):

PR #1917, opened 2021-09-08
Added support for building rpm package from bundle

schaefi:
With the new option --package-as-rpm it is possible to
call the kiwi result bundler such that the image build
results gets packaged into an rpm. I think this is a
handy feature to transport image builds via repositories


schaefi:
This is just a little extension which I think is handy to transport image results through package managers. I also plan to use this in my OSInside/cloud-builder project which will support native image building directly from git in a more straight forward way than obs

PR #1924, opened 2021-09-10
Fixed uninstall handling via dnf, microdnf, zypper

schaefi:
The above package managers supports uninstall instructions  like 'iwl*'. In kiwi there was code checking via rpm if the packages given to uninstall actually exists. That code does not work if the given package to uninstall is an instruction that matches a pattern. Therefore if we use the uninstall section in the kiwi image description, just pass the provided information to the package manager and  don't try to be clever in kiwi itself.

### DIFF
--- a/kiwi/package_manager/dnf.py
+++ b/kiwi/package_manager/dnf.py
@@ -154,20 +154,22 @@ class PackageManagerDnf(PackageManagerBase):
 
         :rtype: namedtuple
         """
-        delete_items = []
-        for delete_item in self.package_requests:
-            try:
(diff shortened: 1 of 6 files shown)


schaefi:
@Conan-Kudo found this when looking at the kickstart files and porting the instructions there to kiwi

Conan-Kudo (changes requested):
We probably want this also for `microdnf`, right?

schaefi:
yes

schaefi:
done for all others were this is an issue too

Conan-Kudo (approved):

PR #1925, opened 2021-09-10
Only wipe bundle dir when required

schaefi:
The given result bundle dir must only be wiped if the
request to turn the result files into an rpm was given.
Only in this case the given bundle dir must start empty


schaefi:
The latest change on the bundler in #1917 introduced a regression that breaks the boxbuild command. The unconditional wipe of the bundle dir does not work if that directory is a mount point. This is the case in the boxbuild plugin were the bundle dir is a 9p shared folder. Also there is no reason to unconditionally delete the bundle dir because this is only needed when we package the result files into an rpm to avoid packaging data that does not belong to the bundle

Conan-Kudo (approved):

PR #1932, opened 2021-09-21
Add support for toplevel include directive(s)

### LINKED ISSUE #1929: Add support for equivalent to kickstart `%include` directive
Right now, KIWI descriptions require all stanzas to be in a single description document. This gets extremely cumbersome when the descriptions get very complex, and also makes it difficult to trivially swap out pieces of it through automation.

For [Fedora's kickstarts used to produce images](https://pagure.io/fedora-kickstarts), the kickstarts are broken up into multiple smaller files expressly to ensure common parts are in their own files, and the portions relevant to specific deliverables are able to be maintained by those folks without breaking everyone else. This makes it easier for people to understand which parts they need to work with to maintain their deliverables and also makes it easier for multiple changes by different people to different parts of the kickstart to land in parallel without conflicts. Finally, it also makes it very easy for new deliverables to be constructed, since reuse of the existing framework makes it possible to quickly make a conformant image.

We could add an `include` stanza of our own that would have similar properties to kickstart `%include`, including document merging when KIWI processes them.

schaefi:
On the toplevel of an image description you can now
specify include directive(s) like in the following
example

```xml
<image ...>
    ...
    <include from="filename_a.xml"/>
    ...
    <include from="filename_b.xml"/>
</image>
```

At the place of their occurence the include statement
will be replaced with the contents of the given filename.
The implementation is based on a XSLT stylesheet and
applies very early in the process. The stylesheet reads
the contents of the given file as XML document().
Thus only valid XML documents gets accepted by this
include concept. This Fixes #1929

Fixes # .

*
*


schaefi:
@Conan-Kudo this should allow to do almost the same as with the %kickstart includes

Feedback welcome

Conan-Kudo:
@schaefi I will look at this over the weekend, thanks!

kiwi/markup/base.py
```diff
@@ -88,3 +103,16 @@ def get_yaml_description(self) -> str:
...
+
+class FileResolver(etree.Resolver):
+    def resolve(self, url, pubid, context):
+        if os.path.exists(url):
+            return self.resolve_filename(url, context)
+        else:
+            # In case a document() reference file cannot be found
+            # an <include> statement is created to be consumed by
+            # the check_include_references_unresolvable runtime check
+            return self.resolve_string(
```
davidcassany, line 116: I am fine with this code. I am trying to understand its motivation though. Correct me if wrong, we are not throwing an exception here for two reasons:

* Provide a got XML schema error if any (unrelated with the current include)
* Provide a proper list of missing includes on runtime check once the schema (without the missing include) is already validated.

Just asking because to my eyes throwing an exception here could save us some lines of code at the expense of not having a full schema validation.

davidcassany (approved):
I assume most of the magic happens at XSLT level with the `xslt_transform_parser` . I definitely should learn XSLT some day... 😅 It is hard to follow for me unless I spend quite some time on it.

schaefi:
> I am fine with this code. I am trying to understand its motivation though. 

@davidcassany  The main intention was to provide a collective error information for all missing includes if there are any.
It could be done nicer with an exception but it will raise on the first occurrence of a missing include and if there are more that can't be found you would realized it step by step. But I agree it's probably better to do it exception based.

I pushed a commit to implement it as I think you intended

Thoughts ?

kiwi/markup/base.py
schaefi: I think this will be addressed from the latest commit I did

schaefi:
> I definitely should learn XSLT some day.

@davidcassany God it's so hard for me too ;) functional programming and all this recursions...

davidcassany:
> I pushed a commit to implement it as I think you intended

Yes this is what I had in mind when I read the code. To me it was convoluted to add a special `filenotfound` include to later process it and raise an Exception at a later point. Something else that now I am asking myself is if the `<include>` element is actually needed in schema. I mean, up to my undersanding XLST runs the conversion and then the schema is validated after **all** `<include>` are substituted. Anyway even if the schema change is not required I like having it there since it is also a source of valid documentation.

schaefi:
yes the `<include>` element should be known to the schema, otherwise you are tightly coupling an XSLT processor with the schema. That would be pretty bad design, not only from a documentation perspective but also from a schema perspective.

schaefi:
@davidcassany Hi, I thought again about what you said to not change the schema definition to add the `<include>` statement as known to the schema. From a design perspective I still think it's questionable but it would also provide a broader scope to use the `<include>` statement anywhere in the document, also in sub sections and not only on the toplevel. So I decided to push yet another commit which implements your idea and does not change the schema. From my perspective we should discuss:

* There is now an `<include>` statement used by XSLT only and does reserved but not listed as an element of the schema. This section must not be added to the schema from now on and this is not obvious because it does not appear there. That's my point on the design flaw

* There is now an   `<include>` statement which can be used anywhere in the document and really flexible

I would be ok if we agree that the flexibility weights more than the design issue in this case

Feedback much appreciated. I don't want to make this decision on my own ;)

davidcassany:
>     * There is now an   `<include>` statement which can be used anywhere in the document and really flexible

Good point! I did not realize about it at first.

I believe for the only non top level element that someone might be tempted to use the `include` function is for `<type>` section or anything under it. Is this extra feature relevant enough to omit the include part in schema? Honestly I don't think so. Also I don't think it is a good idea to split `<type>` section across multiple files, sounds like a bad approach to me. So I clearly vote for having the `include` element inside the schema. Note that I never requested such a thing, I just noted it was possible. To my eyes `<include>` should be part of the schema so it is clearly documented and visible to the user. The resulting limited functionality, might also be good as it does not encourage users to do insane includes.

All that being said I thought about two more cases I'd like to discuss 😅 Sorry, it is is not that I want to make this PR stand here forever, but it is good chance to discuss the scope of it.

* What happens on nested `<includes>`? What if an included file add a new top level `<include>`? I don think we should do anything specific about that beyond knowing what actually happens during the XSLT convertion.
* To achieve a great level of flexibility I believe it would be nice to be capable of adding a profile within an `include`. Currently this is not really possible since the profile definition itself is not a top level element, it lives under `<profiles>`. However I believe this might be easy to achieve if we slightly modify our current schema to allow multiple `<profiles>` and simply merging them if there are several (like we do for `<packages>`). So consider these two included files:

include 1:
```xml
<profiles>
  <profile name="extraPackages" description="Additional packages"/>
</profiles>
<packages profiles="extraPackages">
  <package name="emacs"/>
</packages>
```

include 2:
```xml
<profiles>
  <profile name="updated" description="Additional updates repository"/>
</profiles>
<repository type="rpm-md" profiles="updated">
  <source path="updates_repo_url"/>
</repository>
```

So these two when included could result in something equivalent to:
```xml
...
<profiles>
  <profile name="updated" description="Additional updates repository"/>
</profiles>
<profiles> <!-- these two profiles lists are just interpreted as just a single list -->
  <profile name="extraPackages" description="Additional packages"/>
</profiles>
<repository type="rpm-md" profiles="updated">
  <source path="updates_repo_url"/>
</repository>
<packages profiles="extraPackages">
  <package name="emacs"/>
</packages>
...
```

Conan-Kudo:
The idea here is to replicate the model that the [Fedora kickstarts](https://pagure.io/fedora-kickstarts) use for separating aspects of the image build configuration. Such would be used for flavors, disk configurations, repository setup, etc.

schaefi:
> The idea here is to replicate the model that the [Fedora kickstarts](https://pagure.io/fedora-kickstarts) use for separating aspects of the image build configuration. Such would be used for flavors, disk configurations, repository setup, etc.

yes and I think for this purpose the include supported on toplevel elements will be sufficient

schaefi:
@davidcassany Thanks much for your valuable feedback. I know you did not request the include to be available anywhere. With the last commit I just wanted to open that discussion and also show how it would like in code. Ok, I think we agree that we want the include to be a schema element. I will revert the last commit which brings us back to that state. Thanks

Regarding the profile setting I agree with you this would be a really nice opportunity. I'm going to add that next

Last I will test what happens if an include includes an include. In functional programming and XSLT is all about this, there are recursions and I assume this will just work. But I'll check on it

schaefi:
@davidcassany ok, I'm done with the changes. I think this is nice now.

Please note I realized an XML document must have a document root node to be processed properly. This also applies to included files which gets read via the document() method from XSLT. That in the end requires us to embed any include file in a root node. I chose the same root node kiwi uses for its master document. 

```xml
<image>
   <!... data and sections you want to include -->
</image>
```

Only the inner information will be included. This is required but I think it's also not too bad. The example David made would then look like this

```xml
<image>
    <profiles>
        <profile name="extraPackages" description="Additional packages"/>
    </profiles>
    <packages profiles="extraPackages">
        <package name="emacs"/>
    </packages>
</image>

File: kiwi/schema/kiwi.rnc [hunk 2]
```diff
@@ -78,9 +78,10 @@ div {
         ## The root element of the configuration file   
         element image {
             k.image.attlist &
+            k.include* &
             k.description &
             k.preferences+ &
-            k.profiles? &
+            k.profiles* &
```
davidcassany, line 84: Cool

PR #1933, opened 2021-09-21
This patch fix the issue #1911 and makes secure boot images to work a…

### LINKED ISSUE #1911: Secure Boot (w/ MS keys) with Ubuntu ISO
## Problem description
A Ubuntu (20.04) ISO cannot be booted with Secure Boot, error is Access Denied (as reported by TianoCore).
The root cause is that Ubuntu has two shim files for some reason:
- usr/lib/shim/shimx64.efi
- usr/lib/shim/shimx64.efi.signed

We need to use the latter in [`get_shim_loader()`](https://github.com/OSInside/kiwi/blob/master/kiwi/defaults.py#L733). At least, that one works.
It would also be great if you could mention in the documentation that one needs to use `firmware=uefi` and not `firmware=efi` to enable Secure Boot.

## Expected behaviour
It can be loaded w/ Secure Boot.

Build an ISO based on Ubuntu 20.04 and boot it e.g. with qemu using `ovmf-x86_64-ms-code.*`.

+ KIWI version: 9.23.52
+ Operating system: KIWI runs on SLES15 SP2 in this case, and builds an Ubuntu 20.04
+ OBS version: ?

luisgf:
Signed shimx64.efi in Ubuntu 20.04 is distributed with a new name shimx64.efi.signed, so get_shim_loader() doesn't make the copy properly. This patchs fix that adding new patterns to detect this an copy the file correctly.

Ubuntu 20.04

Fixes #1911

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -748,14 +748,14 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 'Signed grub2 efi loader not found'
             )
         shim_image = Defaults.get_shim_loader(lookup_path)
-        if shim_image:
+        if shim_image and shim_image.filename:
             # The shim concept is based on a two step system including a
             # grub image(shim) that got signed by Microsoft followed by
             # a grub image that got signed by the shim. The shim image
             # is the one that gets loaded by the firmware which itself
             # loads the second stage grub image
             Command.run(
-                ['cp', shim_image, self._get_efi_image_name()]
+                ['cp', shim_image.filename, self._get_efi_image_name()]
             )
             Command.run(
                 [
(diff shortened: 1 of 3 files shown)


schaefi (approved):
This looks perfect and I very much like the refactoring bits :+1: Thanks much

schaefi:
I let the workflows for testing run, once they are all good I'm happy to merge

schaefi:
@luisgf some small linter issues that should get our attention. Are you going to address this ? Thanks

luisgf:
> @luisgf some small linter issues that should get our attention. Are you going to address this ? Thanks

Yes, i take care of that.

schaefi:
going to merge this after lunch... Thanks much

PR #1934, opened 2021-09-22
Run scripts via a screen session in debug mode

schaefi:
When creating a custom script it usually takes some iterations of
try and testing until a final stable state is reached. To support
developers with this task kiwi calls scripts associated with a
screen session. The connection to screen is only done if kiwi
is called with the --debug option.

### DIFF
--- a/doc/source/concept_and_workflow/shell_scripts.rst
+++ b/doc/source/concept_and_workflow/shell_scripts.rst
@@ -53,6 +53,53 @@ bit is set (in that case a shebang is mandatory) otherwise they will be
 invoked via the BASH. If a script exits with a non-zero exit code
 then {kiwi} will report the failure and abort the image creation.
 
+Developing/Debugging Scripts
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When creating a custom script it usually takes some iterations of
+try and testing until a final stable state is reached. To support
+developers with this task {kiwi} calls scripts associated with a
+`screen` session. The connection to `screen` is only done if {kiwi}
+is called with the `--debug` option.
+
+In this mode a script can start like the following template:
+
+.. code:: bash
+
+   # The magic bits are still not set
+
+   echo "break"
+   /bin/bash
+
+At call time of the script a `screen` session executes and you get
+access to the break in shell. From this environment the needed script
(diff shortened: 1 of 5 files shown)


davidcassany (approved):
Looks good to me. I am not proficient on `screen` so I just trust you about the screen flags 😉

davidcassany:
In the past few times I used an ugly hack by adding a function such as 

```bash
function debug                                                                  
{                                                                               
    echo "#############|  Entering DEBUG mode  |####################";          
    cmd=""                                                                      
    while [[ $cmd != "exit" ]]; do                                              
        read -p "> " cmd                                                        
        case "$cmd" in                                                          
            exit ) ;;                                                           
            * ) eval "$cmd";;                                                   
        esac                                                                    
    done                                                                        
    echo "#############|  End of DEBUG mode |####################";             
}
```
in `config.sh` and just call it where ever I wanted to stop in interactive mode resulting in messy non-intuitive state 😅 .

schaefi:
> In the past few times I used an ugly hack by adding a function such as
> 
> ```shell
> function debug                                                                  
> {                                                                               
>     echo "#############|  Entering DEBUG mode  |####################";          
>     cmd=""                                                                      
>     while [[ $cmd != "exit" ]]; do                                              
>         read -p "> " cmd                                                        
>         case "$cmd" in                                                          
>             exit ) ;;                                                           
>             * ) eval "$cmd";;                                                   
>         esac                                                                    
>     done                                                                        
>     echo "#############|  End of DEBUG mode |####################";             
> }
> ```
> 
> in `config.sh` and just call it where ever I wanted to stop in interactive mode resulting in a messy non-intuitive state  .

Wuhu that's fun code :) I think it will be a much better experience now with this change

schaefi:
@davidcassany regarding the options to screen I'm also not an expert but I have tested it.

* Obvious -X "Execute <cmd> as a screen command"
* Not so obvious -t , sets a shell-title (none in our case). It's required for screen to come with a title for each window. As we are starting a new session executing a command -t is required in combination with -X. If not specified screen says "No screen session found" and even worse would send the command as screen command to an open session if there is any. Thus "-t" is mandatory to run the command in a new screen session and title

PR #1937, opened 2021-10-04
Missing command is not reported properly

tkopecek:
removal of raise_on_error means that KiwiCommandNotFound will
be never raised.

I'm not sure if there was a bigger intent, so maybe this fix breaks something else. Ping to @schaefi as author of previous removal of this option.

### DIFF
--- a/kiwi/xml_description.py
+++ b/kiwi/xml_description.py
@@ -203,7 +203,8 @@ class XMLDescription:
         """
         try:
             Command.run(
-                ['jing', schema_file, description_file]
+                ['jing', schema_file, description_file],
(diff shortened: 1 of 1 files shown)


schaefi:
I think I don't understand the intention of this change. raise_on_error=True is the default setting for Command.run.
see here: https://github.com/OSInside/kiwi/blob/master/kiwi/command.py#L62

Can you explain the issue you had more detailed ?

Thanks

tkopecek:
Yep, you're right - it was confusion created by mixing kiwi code from f33/f34 rpms and github. Too much experimenting on my side. (dropping)

PR #1941, opened 2021-10-06
Added mdadm to requires for systemdeps

schaefi:
When building raid based images the buildhost needs mdadm

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -234,16 +234,10 @@ Requires:       btrfs-progs
 %endif
 %endif
 %if 0%{?suse_version}
-Requires:       gptfdisk
 Requires:       squashfs
 %else
-Requires:       gdisk
 Requires:       squashfs-tools
 %endif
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

package/python-kiwi-spec-template
```diff
@@ -244,6 +244,7 @@ Requires:       lvm2
 Requires:       parted
 Requires:       kpartx
 Requires:       cryptsetup
+Requires:       mdadm
```
Conan-Kudo, line 247: Shouldn't this be for disk images and not filesystems?

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
Conan-Kudo: Yes, I agree that we should. These tools don't work if you're just creating plain filesystems anyway.

Conan-Kudo (approved):

Issue #1918, opened 2021-09-09
Integrating KIWI with Koji


schaefi:
## Description

Discuss opportunities to use kiwi with Koji

schaefi:
Takeaways from the meeting

* an include directive to manage and allow overrides of top level elements in the description would be good to have: #1929
  we are not blocked in going forward but in the long run this feature should become true

* we can help to convert the kickstart files from https://pagure.io/fedora-kickstarts to kiwi descriptions. Let's use this issue to keep track on that effort

* last for execution of kiwi in Koji a plugin to register a ```kojibuild``` command could be helpful. We can help with setting up that plugin and would be happy to do so

schaefi:
@Conan-Kudo I have pushed an initial example "clone from kickstart to kiwi" here:

* https://github.com/OSInside/fedora-kiwi-descriptions/tree/rawhide/fedora-disk

This is probably far away from being useful in Koji but it should be good as a start. I plan to add the include feature relatively quickly and also shared how I would do it.

At this point I'd like to wait for feedback from the Koji people to see if they still consider the kiwi toolchain as useful to the image targets Koji would like to offer. From there I think I will need a better understanding how Koji calls into the image building tools and then we might end up with the kojibuild plugin. But I think a conversation before the next steps would then be needed as well.

Does this makes sense ?

Thanks

Conan-Kudo:
Sure, we should ask @mikem23 and @tkopecek to take a look and give us feedback.

tkopecek:
Quick update: Now it is on us - I'm preparing some PoC koji plugin, so we can evaluate it in some real-world situations. I'll provide code/update later.

schaefi:
@tkopecek great thanks :+1:

schaefi:
Related to this project : #1932

tkopecek:
@schaefi - question - now I'm looking into kiwi.result file. It is a good start but I would need some additional info about installed rpms - basically their headers (full list of info is: name,version,release,epoch,arch,buildtime,sigmd5/payloadhash,filesize) so we can be 100% sure that we know what is this exact rpm.
I see that for fedora (test-image-microdnf) there remain rpms in build/image-root/var/cache/yum. Can this be relied on or is it more setup without proper cleaning? Is there some other way how e.g rpmdb could be retrieved?

Conan-Kudo:
It should be possible to extend the result file to include what you need, right @schaefi?

schaefi:
[...]

PR #1949, opened 2021-10-12
Add support for portable result data

schaefi:
In addition to the serialized Result instance kiwi.result
file this commit also creates a portable version of this
information in kiwi.result.json. Only the information that
can be expressed as json document is part of the portable
version. This is related to Issue #1918

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,8 @@ config = {
         'lxml',
         'pyxattr',
         'requests',
-        'PyYAML'
+        'PyYAML',
+        'simplejson'
     ],
     'packages': ['kiwi'],
     'cmdclass': {


schaefi:
> python's stdlib json is also ok with tuples. It looks to me (from tox.ini) that kiwi support 3.6+ so it is already there:

Hmm, I see a difference. So I'm running on ```Python 3.6.12``` and had issues with tuples. Here is an example

```python
from typing import NamedTuple
import json
import simplejson

result_file_type = NamedTuple(
    'result_file_type', [
        ('filename', str),
        ('use_for_bundle', bool),
        ('compress', bool),
        ('shasum', bool)
    ]
)


data = {
    'key': result_file_type(
        filename='foo', use_for_bundle=True, compress=True, shasum=True
    )
}

print(json.dumps(data))

print(simplejson.dumps(data))
```

This prints

```
{"key": ["foo", true, true, true]}
{"key": {"filename": "foo", "use_for_bundle": true, "compress": true, "shasum": true}}
```

tkopecek:
> > python's stdlib json is also ok with tuples. It looks to me (from tox.ini) that kiwi support 3.6+ so it is already there:
> 
> Hmm, I see a difference. So I'm running on `Python 3.6.12` and had issues with tuples. Here is an example
Understood - I was thinking only about "classic" tuples.

PR #1950, opened 2021-10-14
Fixed use of LUKS encrypted images with empty pass

schaefi:
For initial provisioning of LUKS encrypted disk images an empty passphrase key is handy to avoid interaction in the deployment process. However, the dracut kiwi modules were lacking the information that the luks keyfile could be an empty passphrase key which must not be opened with the potential risk to get prompted for input. This commit introduces a new profile environment variable evaluated by the dracut kiwi lib code to open the LUKS pool and allows to distinguish the situation on key files with or without a passphrase

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-luks-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-luks-lib.sh
@@ -23,8 +23,22 @@ function resize_luks {
 }
 
 function activate_luks {
+    declare kiwi_luks_empty_passphrase=${kiwi_luks_empty_passphrase}
     local device=$1
-    /usr/lib/systemd/systemd-cryptsetup attach luks "${device}"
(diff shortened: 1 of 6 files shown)


schaefi:
@Jack. If you are reading this and think about how to test this, it could be a bit tricky before merge and the rebuild of new packages. Basically this is touching two components of kiwi:

* The builder code itself, to bring in the new environment variable
* The initrd code for the 90kiwi-repart module which makes use of the new profile variable

For testing without new packages you would need:

1. Patching of the python code locally as done here
2. Overwrite of dracut code inside of your image build. There is only one file affected: ```dracut/modules.d/99kiwi-lib/kiwi-luks-lib.sh``` The most simple way to make this effective in your image build is by creating an overlay file in your image description

    ```bash
    root/usr/lib/dracut/modules.d/99kiwi-lib/kiwi-luks-lib.sh
    ```

Then rebuild

PR #1951, opened 2021-10-19
Raise exception on umount issue

### LINKED ISSUE #1914: MountManager gives up on unmounting, resulting in broken ISO (?)
## Problem description
Sporadically, `MountManager.umount` fails on my machine because of antivirus scanning the new files, like so:
```[ WARNING ]: 20:41:31 | 3 umount of /var/tmp/kiwi_mount_manager.nueuyww_ failed, try again in 1sec
[ WARNING ]: 20:41:32 | /var/tmp/kiwi_mount_manager.nueuyww_ still busy at MountManager
Exception ignored in: <finalize object at 0x7f95110977e0; dead>
  File "/usr/lib64/python3.6/weakref.py", line 548, in __call__
    return info.func(*info.args, **(info.kwargs or {}))
  File "/usr/lib64/python3.6/tempfile.py", line 799, in _cleanup
    _shutil.rmtree(name)
  File "/usr/lib64/python3.6/shutil.py", line 490, in rmtree
    onerror(os.rmdir, path, sys.exc_info())
  File "/usr/lib64/python3.6/shutil.py", line 488, in rmtree
    os.rmdir(path)
OSError: [Errno 16] Device or resource busy: '/var/tmp/kiwi_mount_manager.nueuyww_'```

I've added a little `lsof` diagnostic to `MountManager`, that's why I suspect the AntiVirus.

Whenever the above logs appear, the resulting ISO is broken. Or rather, whenever the resulting ISO is broken, I can see that unmounting failed as displayed above.

During boot of the ISO, you see something like:
```[...]
[   19.454751] squashfs: version 4.0 (2009/01/31) Phillip Lougher
[   20.375893] EXT4-fs (loop1): INFO: recovery required on readonly filesystem
[   20.376308] EXT4-fs (loop1): write access unavailable, cannot proceed (try mou)
[   20.361686] dracut-initqueue[446]: mount: /run/overlay/rootfsbase: cannot moun.
[   20.399798] dracut: FATAL: Failed to mount live ISO root filesystem
[   20.400292] dracut: Refusing to continue
[...]```

My hypothesis is that due to the failure to unmount, something isn't synced properly before Kiwi starts with the next step. The resulting filesystem ends up corrupted, and this cannot be dealt with during boot.

## Expected behaviour
Either failure of the image build, or trying harder to unmount. IMHO, a success with a broken ISO is not a great result.
I know that in the end, it's the fault of the antivirus, in a way. But I'd rather have a build failure with non-zero exit code than a broken ISO.

Open a file inside the mount point and keep it open?

+ KIWI version: 9.23.52
+ Operating system: SLES15 SP2
+ OBS version: ?

schaefi:
Up to now kiwi tried to umount a resource it has mounted for some
time and if it was not possible due to a busy state the process
continued. In any of these situations it was not possible to free
up the nested resources and it could also happen that the image
build result got corrupted because there is still data that can
be consumed but shouldn't be there anymore. The change here makes
kiwi to stop the process and raise with an error also including
the information that there are now resources still active but
busy on the host and a manual cleanup is unfortunately needed.
This Fixes #1914

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -119,6 +119,7 @@ Requires:       kiwi-tools
 Requires:       mtools
 Requires:       rsync
 Requires:       tar >= 1.2.7
+Requires:       lsof
 Requires:       openssl
 
 %description -n kiwi-systemdeps-core


schaefi:
> Btw I've locally hacked mount manager to invoke lsof / fuser when unmounting failed, to print the processes which prevent unmounting. That would be the cherry on top :)

I think calling this tools would create an unnecessary dependency to the code base. I agree with you the information would be valuable. So how about adding the command to the error message to provide users a debugging pointer ?

schaefi:
All right I added an lsof call (or message to install it for more details). From my perspective the fuser data is a subset of what lsof provides, so this should be enough.

How about it now :) ?

Thanks

schaefi:
> @schaefi Is there a reason we don't want to list `lsof` in the core sysdeps subpackage? I think it's pretty reasonable for us to do that...

@Conan-Kudo This is a good idea. I added a commit to this PR to add lsof to the required packages. Can you double check if this is suitable with regards to other distros ? Thanks much

PR #1952, opened 2021-10-20
Allow firmware equals None

isbm:
Allows firmware to be equal None.

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -1148,7 +1148,7 @@ class Defaults:
             'i586': ['bios'],
             'i686': ['bios'],
             'ix86': ['bios'],
-            'aarch64': ['efi', 'uefi'],
+            'aarch64': ['efi', 'uefi', 'none'],
             'arm64': ['efi', 'uefi'],
             'armv5el': ['efi', 'uefi'],
             'armv5tel': ['efi', 'uefi'],
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -1498,7 +1498,7 @@ div {
         ## the disk layout and the partition table type. By default
         ## the standard x86 bios firmware setup is used
         attribute firmware {
-            "bios" | "ec2" | "efi" | "uefi" | "ofw" | "opal"
+            "bios" | "ec2" | "efi" | "uefi" | "ofw" | "opal" | "none"
         }
         >> sch:pattern [ id = "firmware" is-a = "image_type"
             sch:param [ name = "attr" value = "firmware" ]
--- a/kiwi/schema/kiwi.rng
+++ b/kiwi/schema/kiwi.rng
@@ -2202,6 +2202,7 @@ the standard x86 bios firmware setup is used</a:documentation>
           <value>uefi</value>
           <value>ofw</value>
           <value>opal</value>
+          <value>none</value>
         </choice>
       </attribute>
       <sch:pattern id="firmware" is-a="image_type">


isbm:
OK, that explains fair enough. So what actually needs to be done, in order to achieve the desired effect?

What currently comes to my mind is that we could make a custom placeholder for a plugin custom functions, maybe? Like instead of `none`, one would say `custom` and then just call python modules via some established interface, so anything (literally) can be routed? But that sounds complex and long-term solution.

What we can do today?

schaefi:
Regarding the desired effect, I still don't understand how the image should look like in the end and how that is depending on the firmware setting

> We'd like to build a single image with a boot and root partition (and actually some more). In the boot partition, we install u-boot. And since we do not want any efi stuff, but still want the partitioning to happen, we need firmware = none.

* extra boot partition is ```bootpartition="true" bootpartsize="..."```
* (and actually more) this is mostly specific to the board and actually I think we should implement the ```<partitions>``` feature you have mentioned.
* we install u-boot.... this tells me you want to skip any bootloader action done by kiwi and do whatever is needed on your own. You can simply reach this by
  ```xml
      <bootloader name="custom"/>
  ```

So apart from the (and actually more) partitions I think you can reach your goal with todays options. I think also helpful in this regard is the example I created here:

* https://github.com/OSInside/kiwi/blob/master/build-tests/x86/leap/test-image-embedded/appliance.kiwi

It has similar requirements compared to your use case

Regarding the custom partitions I'm very open to add the support for this... maybe in January 2022 :)

Until then extra custom partitions can be created through the scripting engine. An example for such an image can be found here:

* https://github.com/OSInside/kiwi/tree/master/build-tests/x86/tumbleweed/test-image-custom-partitions

If you tell me exactly what you need from a disk layout perspective I can also setup an image that builds that.

What else do we need ?

Thanks

schaefi:
@MofX Hi, after some time of letting this settle I came back reading were we are with this. If I got it right we are in agreement that we want a ```firmware="uboot"``` setting to be able to define certain settings in regard to this bootloader. However, with the discussed existing options and also the new ability to define custom partitions I'm not sure if we still need the extra firmware group setting ?

Please advise and if you think this is still needed it would be great if you can adapt the patch here to follow the proposed solution.

Thanks much

MofX:
> @MofX Hi, after some time of letting this settle I came back reading were we are with this. If I got it right we are in agreement that we want a `firmware="uboot"` setting to be able to define certain settings in regard to this bootloader. However, with the discussed existing options and also the new ability to define custom partitions I'm not sure if we still need the extra firmware group setting ?

It definitely is still needed, but as long as kiwi does not do anything special, if firmware="uboot", I would vote for firmeate="custom", otherwise it implies, that something is done to install uboot.

> Please advise and if you think this is still needed it would be great if you can adapt the patch here to follow the proposed solution.

 I cannot change this PR, because it was created by @isbm.

schaefi:
>It definitely is still needed, but as long as kiwi does not do anything special, if firmware="uboot", I would vote for firmeate="custom", otherwise it implies, that something is done to install uboot.

sounds good to me. I have implemented this now in #2012 

> I cannot change this PR, because it was created by @isbm.

I see, sorry I overlooked this part.

schaefi:
@MofX Hope you don't mind, I added you with read permissions to the project. This will allow to discuss and review stuff. Would be great if you can have a look at the open PR #2012 

If you are interested it would also be great to have you as a collaborator for code :) Just drop me a note.

Thanks much

schaefi:
I'm closing this one on behalf of #2012

PR #1954, opened 2021-10-20
Add support for custom result bundle naming

### LINKED ISSUE #1943: Custom naming for image file
Currently there is no way of customizing the naming of the output image.
For our current practical example on JeOS (that is currently being renamed to Minimal):
The definition would be
`<image  name="SLES15-SP3-JeOS" .../>`
Building the OpenStack profile defined as:
`<profile name="OpenStack-Cloud" description="JeOS for OpenStack Cloud" arch="x86_64"/>`
And having the profile configured with
`<preferences profiles="OpenStack-Cloud">
        <version>15.3</version>`
we would get an image named

`SLES15-SP1-JeOS.x86_64-15.1-OpenStack-Cloud-Build.qcow2`

Canonically meaning IMAGE_NAME.ARCH-PROFILE_VERSION-PROFILE_NAME-BUILD_VERSION.FORMAT

in the rename of JeOS to Minimal we face one specific challenge, we want to separate the Minimal-VM from the Minimal-Image, meaning, the VM <type>image="vmx" from the  image="oem", but profile preferences don't actually help much here.

Ideally we would like to be able to achieve

`SLES15-SP4-Minimal-VM.x86_64-OpenStack-Cloud-GM.qcow2
SLES15-SP4-Minimal-Image.aarch64-RaspberryPi-GM.raw.xz`

meaning canonically something like 
IMAGE_NAME.BUILD_TYPE.ARCH-PROFILE_NAME-BUILD_VERSION.FORMAT

Basically dropping the profile version that is currently not really useful given that the version information is already on the NAME tag (15-SP4) and having a way to differentiate the BUILD_TYPE between vmx and oem

In a more complete solution would be nice to be able to have the actual version (15-SP4) split from the name as well. Maybe even splitting or creating more fields.

Of course I understand is not a simple change, but having an extra mechanism that can define canonical naming based on existing fields and maybe adding a few more fields related to naming specifically might make the naming more robust, including support multiple naming schemes of different distributions.

What are you guys suggestions and ideas regarding that?

schaefi:
Allow to specify a bundle_format per <type> which is used
at the time of the result bundle creation to create image
output names matching the provided pattern.
This Fixes #1943

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-bundle-format/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-bundle-format/appliance.kiwi
@@ -1,0 +1,79 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<image schemaversion="7.4" name="test-image-bundle-format">
+    <description type="system">
+        <author>Marcus Schäfer</author>
+        <contact>ms@suse.com</contact>
+        <specification>Disk test build with custom bundle format</specification>
+    </description>
+    <preferences>
+        <version>1.42.1</version>
+        <packagemanager>zypper</packagemanager>
+        <locale>en_US</locale>
+        <keytable>us</keytable>
+        <timezone>Europe/Berlin</timezone>
+        <rpm-excludedocs>true</rpm-excludedocs>
+        <rpm-check-signatures>false</rpm-check-signatures>
+        <bootsplash-theme>bgrt</bootsplash-theme>
+        <bootloader-theme>openSUSE</bootloader-theme>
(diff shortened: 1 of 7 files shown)


gmoro:
@schaefi  %T is turning into "oem" even when <type image="vmx"/> , might be worth to check that

schaefi:
This is correct. There is no vmx type anymore. the vmx type is auto converted into an oem type by an XSLT transformation. So the resulting image description contains an oem type setting. If you are still using old schema format I suggest to update:

```
kiwi-ng image info --description /path/to/your/description --print-xml
```

If you do that you will also see the conversion that happened to your document. For details on the vmx to oem conversion see: #1425

gmoro:
Oh, I didn't see that, but was my first suspicion, thanks to the pointer, I will update our files appropriately.

PR #1965, opened 2021-11-07
utils/temporary: the --temp-dir does not work

kshtsk:
Using of --temp-dir argument does not make an effect,
because optional 'dir' parameter defaults to the
global TEMP_DIR value before it's changed.

This patch address this issue.

Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@gmail.com>

### DIFF
--- a/kiwi/utils/temporary.py
+++ b/kiwi/utils/temporary.py
@@ -31,10 +31,10 @@ class Temporary:
     **Provides namespace to handle temporary files and directories**
     """
     def __init__(
-        self, dir: str = defaults.TEMP_DIR, prefix: str = ''
+        self, dir: str = '', prefix: str = ''
     ):
         self.prefix = prefix if prefix else 'kiwi_'
-        self.dir = dir
+        self.dir = dir or defaults.TEMP_DIR
 
     def new_file(self) -> IO[bytes]:
         return NamedTemporaryFile(
--- a/kiwi/utils/temporary.py
+++ b/kiwi/utils/temporary.py
@@ -31,17 +31,17 @@ class Temporary:
     **Provides namespace to handle temporary files and directories**
     """
     def __init__(
-        self, dir: str = '', prefix: str = ''
+        self, path: str = '', prefix: str = ''
     ):
         self.prefix = prefix if prefix else 'kiwi_'
-        self.dir = dir or defaults.TEMP_DIR
+        self.path = path or defaults.TEMP_DIR
 
     def new_file(self) -> IO[bytes]:
         return NamedTemporaryFile(
-            dir=self.dir, prefix=self.prefix
+            dir=self.path, prefix=self.prefix
         )
 
     def new_dir(self) -> TemporaryDirectory:
         return TemporaryDirectory(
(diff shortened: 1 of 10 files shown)


schaefi:
You are so right, shame on me, this is a bad one. Thanks much for fixing

kiwi/utils/temporary.py
```diff
@@ -31,10 +31,10 @@ class Temporary:
     **Provides namespace to handle temporary files and directories**
     """
     def __init__(
-        self, dir: str = defaults.TEMP_DIR, prefix: str = ''
+        self, dir: str = '', prefix: str = ''
```
schaefi, line 34: As we are on it, using the built-in name "dir" was not a good idea from me as well. Codacy is also complaining about it. Could we rename the parameter e.g ```dir_name``` or ```tmp_dir_name```

schaefi: I realized changing the parameter name will impact some calls, but I think this would be a good change

kshtsk: It is a `path` so maybe using it instead of `dir_name`?
Like in description `Provides namespace to handle temporary files and directories`.

schaefi: Fine by me

schaefi:
I've approved the workflow run, once you change the parameter from "dir" to "path" there will be some errors.

Just in case you don't know you can run the tests and static type checks locally with:

```
tox -e check,unit_py3_[YourMajorPythonVersion] "-n 5"
```

schaefi:
Thanks much

kshtsk:
> I've approved the workflow run, once you change the parameter from "dir" to "path" there will be some errors.
> 
> Just in case you don't know you can run the tests and static type checks locally with:
> 
> ```
> tox -e check,unit_py3_[YourMajorPythonVersion] "-n 5"
> ```
```
tox -e 'check,unit_py3_[YourMajorPythonVersion]' "-n 5"
ERROR: unknown environment 'unit_py3_[YourMajorPythonVersion]'
```

kshtsk:
ah...stupid me, I should replace with version number, hm... I have only 3.9 at my system :-D no worries, I will sort it out

kiwi/utils/temporary.py
kshtsk: @schaefi I took another look at the file, it looks like a wrapper around `tempfile.NamedTemporaryFile` and `tempfile.TemporaryDirectory`, which is using `dir` arguments in constructor, so it makes sense to leave the older signature too.

schaefi: yes the methods from the tempfile module uses ```dir``` arguments, which we just have to accept as it is

schaefi (approved):
Looks great to me :+1: Let's wait for the tests to pass and we are ready for merging

schaefi:
> ah...stupid me, I should replace with version number, hm... I have only 3.9 at my system :-D no worries, I will sort it out

Adding 3.9 and maybe even 3.10 targets to tox.ini sounds like a good idea too. I'll look into that once we have merged this PR

PR #1969, opened 2021-11-08
Drop usage of `factor` from coreutils

### LINKED ISSUE #1968: Unittests failed due to missing 'factor'.
## Problem description
When developing the library on system where 'factor' dependency is missing it is not possible to pass unittests.

## Expected behaviour
Unittests should be crossplatform without extra dependencies.

Run it any OS where are no 'factor' tool installed, for example, mac os, or bsd.
tox -e 'unit_py3_8' "-n 5"

kshtsk:
Fixes #1968.

### DIFF
--- a/kiwi/utils/checksum.py
+++ b/kiwi/utils/checksum.py
@@ -21,7 +21,6 @@ import hashlib
 import encodings.ascii as encoding
 
 # project
-from kiwi.command import Command
 from kiwi.utils.compress import Compress
 
 from kiwi.exceptions import (
@@ -168,11 +167,7 @@ class Checksum:
 
         :rtype: tuple
         """
-        blocksize = 1
-        for factor in self._prime_factors(file_size):
-            if blocksize * factor > 8192:
-                break
-            blocksize *= factor
+        blocksize = self._block_size(file_size)
         blocks = int(file_size / blocksize)
         block_list = namedtuple(
             'block_list', ['blocksize', 'blocks']
@@ -182,16 +177,63 @@ class Checksum:
             blocks=blocks
         )
 
-    def _prime_factors(self, number):
+    @staticmethod
+    def _block_size(number, max_size=8192):
         """
-        Get prime factors for the given number
+        Compute maximal block size equal to a product of the first
+        prime factors of the given number.
 
         :param int number: number to factorize
+        :param int max_size: maximal value for block size
 
-        :return: prime factors
+        :return: int: block size
+        :rtype: int
+        """
+        n = number
+        bs = 1
+        for i in Checksum.primes(max_size):
+            while n % i == 0:
+                _bs = bs * i
+                if _bs > max_size:
+                    return bs
+                bs = _bs
+                n = n // i
+                if n == 1:
+                    return bs
+        return bs
+
+    _primes = [2, 3, 5, 7, 11, 13]
+
+    @classmethod
+    def _update_primes(cls, n):
+        if n <= cls._primes[-1]:
+            return
+        _n = int(n**0.5) + 1
+        cls._update_primes(_n)
+
+        start = cls._primes[-1] + 1
+        for i in range(start, n + 1):
+            prime = True
+            for p in cls.primes(_n):
+                if i % p == 0:
+                    prime = False
+                    break
+            if prime:
+                cls._primes.append(i)
+
+    @classmethod
+    def primes(cls, number):
+        """
+        Get prime numbers no greater than given number.
+
+        :param int number: highest possible number to return.
 
         :rtype: int generator
         """
-        factor_call = Command.run(['factor', format(number)])
-        for factor in factor_call.output.split(':')[1].lstrip().split(' '):
-            yield int(factor)
+        if number > cls._primes[-1]:
+            cls._update_primes(number)
+        for p in cls._primes:
+            if p <= number:
+                yield p
+            else:
+                break
--- a/kiwi/utils/checksum.py
+++ b/kiwi/utils/checksum.py
@@ -22,6 +22,7 @@ import encodings.ascii as encoding
 
 # project
 from kiwi.utils.compress import Compress
+from kiwi.utils.primes import factors
 
 from kiwi.exceptions import (
     KiwiFileNotFound
@@ -167,7 +168,11 @@ class Checksum:
 
         :rtype: tuple
         """
-        blocksize = self._block_size(file_size)
+        blocksize = 1
+        for factor in factors(file_size, 8192):
+            if blocksize * factor > 8192:
+                break
+            blocksize *= factor
         blocks = int(file_size / blocksize)
         block_list = namedtuple(
             'block_list', ['blocksize', 'blocks']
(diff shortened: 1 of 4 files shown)


kshtsk:
The py 3.9 commit can be dropped once the #1967 merged

Conan-Kudo:
I'm not sure we want this, because we generally expect everything to run on Linux environments, not BSD or macOS environments.

Conan-Kudo:
Also, this dependency is not available for all distributions we target, so landing this will require work to add it to at least Fedora EPEL for CentOS/RHEL. I'm not convinced we should do this yet, though...

kshtsk:
> Also, this dependency is not available for all distributions we target, so landing this will require work to add it to at least Fedora EPEL for CentOS/RHEL. I'm not convinced we should do this yet, though...

which dependency?

Conan-Kudo:
`sympy`

kshtsk:
> sympy

but it is there https://dl.fedoraproject.org/pub/epel/7/x86_64/Packages/s/sympy-0.7.6-2.el7.noarch.rpm

Conan-Kudo:
It's not present in EPEL 8, which is where kiwi is packaged today: https://src.fedoraproject.org/rpms/kiwi

kshtsk:
> It's not present in EPEL 8, which is where kiwi is packaged today: https://src.fedoraproject.org/rpms/kiwi

Sorry, I don't know that we need a dependency in packages, why not using pip for dependencies?
Maybe it is easier to add a simple function for prime number generation? I just think that calling external tool is overkill.

kshtsk:
Thanks guys for the discussion, I understand it totally, it appears for me that installing coreutils to mac or bsd is easier task than support sympy or other non-standard libs. For now I can workaround the issue using *brew*, but I will look into adding pure python implementation for the factor function later.

kshtsk:
@Conan-Kudo @schaefi guys, could you please take another look

schaefi:
@kshtsk Thanks for your effort to write your own primes generator. I was looking at it and thought it might be written a bit shorter. What do you think about the following:

```python
from typing import List

def factor(number: int, primes: List[int]) -> None:
    use_primes = [2, 3, 5, 7, 11, 13]
    quotient = None

    if number == 1 and not primes:
        primes.append(1)
        return

    for prime in use_primes:
        if number % prime == 0:
            quotient = int(number / prime)
            primes.append(prime)
            break

    if quotient:
        factor(quotient, primes)
```

could be called as

```python
prime_factors = []
factor(21, prime_factors)
print(prime_factors)
```

What do you think ? Thanks

schaefi:
Having a factor method instead of calling a factor program would then be a drop in replacement which should also make the diff more straight forward imho

kshtsk:
> @kshtsk Thanks for your effort to write your own primes generator. I was looking at it and thought it might be written a bit shorter. What do you think about the following:

It can be written shorter but it will not be as computational efficient as my version, which is really short enough simplified version of building list of primes. But it is still not the most efficient, it would be more complicated, to get it real fast.
Btw, the reason I've dropped `factor` is that according to previously existing code it wasn't necessary to factor whole file size since the maximum block size is 8192, but the highest possible prime factor is about `sqrt(number to factor)` So the new block_size method is just adoption of simplified factorisation algorithm. So in general, we do not need primes greater than 8192. I was trying to save the functionality. For really big file sizes for example 3GB, which is realistic, like DVD iso, we luckily do not need go through all of primes, which would be way slower than coreutils `factor` which is using more advanced factorisation techniques.

Also I didn't use typing because the rest of the class does not use typing either, but if you like I can rewrite using typing.

> 
> ```python
> from typing import List
> 
> def factor(number: int, primes: List[int]) -> None:
> ```
Looks like you're using procedural approach here, which is not good, in place factor function should return a List instead of None in my opinion.
>```
>     use_primes = [2, 3, 5, 7, 11, 13]
>```
First of all I don't see the `use_primes` list is extended in your function, so the code is not equivalent what 'factor' does.
Even if you add some code later, the list of primes should be global so you can reuse it for further calls.
But when you add use_primes calculation the code lines will grow.

>```
>     quotient = None
> 
>     if number == 1 and not primes:
>         primes.append(1)
>         return
>     for prime in use_primes:
>         if number % prime == 0:
>             quotient = int(number / prime)
>             primes.append(prime)
>             break
> 
>     if quotient:
>         factor(quotient, primes)
> ```
> 
> could be called as
> 
> ```python
> prime_factors = []
> factor(21, prime_factors)
> print(prime_factors)
> ```

Now, instead of 21 try to use some prime number a bit greater, for example, 8191. 

> 
> What do you think ? Thanks

schaefi:
> First of all I don't see the use_primes list is extended in your function
> Now, instead of 21 try to use some prime number a bit greater, for example, 8191.

correct, that part is missing

> Looks like you're using procedural approach here, which is not good
> in place factor function should return a List instead of None in my opinion.

It uses a recursive approach from which it is hard to return a list unless working from a class or global scope.
I was also not happy with that part, but I don't wanted to compete with your code, I just thought it maybe can be
done in a simpler drop in replacement factor shell out vs. factor method

Nevermind it looked like you did a deeper walk through to this so let's go with that.

Just one aspect and you mentioned it already:

> Also I didn't use typing because the rest of the class does not use typing either, but if you like I can rewrite using typing.

We would like to use types and annotations as much as possible. There are code parts which were not yet covered by this.
The reason is it's not so long ago when kiwi was also supporting py2 and from that perspective we started using the py3
types and annotations syntax not so long ago.

So in short, yes please if you can update the code to use the typing that would be great

Thanks

schaefi:
ok just for completeness and because I don't want to die stupid ;) I put this into a class

```python
class Number:
    def __init__(self, number: int) -> None:
        self.number = number

        # init prime numbers for 1/2 of the value to calculate prime factors for
        self.prime_list = self.primes(int(number / 2))
        self.prime_factors = []

    def primes(self, number: int) -> List[int]:
        # This for sure can be optimized...
        primes = [2]
        for i in range(3, number, 2):
            if not any(i % prime == 0 for prime in primes):
                primes.append(i)
        return primes

    def factor(self) -> List[str]:
        self.prime_factors = []
        self._get_prime_factors(self.number)
        return self.prime_factors

    def _get_prime_factors(self, number: int) -> None:
        # Using the division method...
        quotient = None

        if number == 1 and not self.prime_factors:
            self.prime_factors.append(1)
            return

        for prime in self.prime_list:
            if number % prime == 0:
                quotient = int(number / prime)
                self.prime_factors.append(prime)
                break

        if quotient:
            self._get_prime_factors(quotient)
        elif not quotient and number > 1:
            self.prime_factors.append(number)



print(Number(47128).factor())

print(Number(1).factor())

print(Number(8191).factor())

print(Number(21).factor())

print(Number(128).factor())

print(Number(512).factor())
```

@kshtsk your version is for sure faster so I'm fine if we go for it. I think it deserves its own namespace though. What do you think if we put your code into an extra class ?

Thanks

kshtsk:
@schaefi I moved the code to a separate module, however I gave up with classes, since there is absolutely no need in them.

schaefi (approved):
Great I like it :+1: Thanks

schaefi:
will take some sleep and merge it tomorrow ;)

PR #1971, opened 2021-11-09
config/functions: make baseQuoteFile use extended regex

kshtsk:
It is more universal to use extended regex, because they can
be used on wider range of platforms, for example, same
regex works equally on bsd sed and linux distributions.

Signed-off-by: Kyr Shatskyy <kyrylo.shatskyy@gmail.com>


schaefi (approved):
Looks good to me :+1: 

Less quoting, thanks.

PR #1975, opened 2021-11-11
Drop cdrtools support

### LINKED ISSUE #1908: Drop support for cdrtools (mkisofs/genisoimage)
## Problem description

kiwi still supports the use of the old mkisofs/genisoimage toolchain. However everybody is using xorriso and the
distributions without support for xorriso are out of support from kiwi. Thus if there are no objections we should get
rid of this old code part

Changes affects:

* kiwi/kiwi/iso_tools/cdrtools.py
* kiwi/kiwi.yml
* kiwi/doc/source/api/kiwi.iso_tools.rst

schaefi:
Tools like mkisofs, genisoimage and friends are all unmaintained and have found their substitute in xorriso for quite some time.
This commit deletes the code from kiwi which supported creation of ISO images using the obsolete cdrtools collection. In addition all hack code associated with these tools like mbr insertion via isohybrid and things like that got dropped as well.

This Fixes #1908


schaefi:
ok guys let's remove some crap :)

Conan-Kudo (approved):

Issue #1964, opened 2021-11-03
Support for "this://" for repository paths


zohourih:
In original Kiwi (https://doc.opensuse.org/projects/kiwi/doc/, section 4.1.7), a "this://" keyword was provided that allowed settings repository paths based on relative path rather than full path (the latter being done using "dir://"). However, this functionality does not seem to have been ported over to Kiwi NG. Is there a way to achieve the same functionality with Kiwi NG, or are there any plans to add support for this functionality?

schaefi:
You are right, this "this" uri type was not ported.

Just to put me right, the this:// path was basically a pointer to the directory which holds your image description. In the past people could use it in the repository section to point to a repo which they host as part of their image description.

This is what you would like to achieve correct ?

Thanks

zohourih:
@schaefi Yes; specifically, this keyword provides us with the possibility of pointing to a local repository that gets pulled to a random location using a script, by using the path of the repository folder relative to the image description, and not having to keep track of the full path of the random location.

Conan-Kudo:
We also going to be dropping `dir://` handler too, per #1926.

zohourih:
Ah, well, that is not very good news. Will there be any alternative to achieve the same behavior?

schaefi:
> Ah, well, that is not very good news. Will there be any alternative to achieve the same behavior?

Let's add something that brings a smile on your face :) First I have added the this:// type such that you can reference the repo living in your image description. We might consider to allow this uri type for other source locators in the future as well but so far it can only be used in a repo definition like in the following example

```xml
<repository>
    <source path="this://my_repo"/>
</repository>
```

schaefi:
Regarding the information from @Conan-Kudo about the removal of ```rpm-dir``` in the future, this is correct and not a limitation of kiwi but something package managers will not support. Actually this is a good change because metadata for a repo should not be calculated by the package manager.

For you this means you need to create repo metadata but this is not a big issue imho. Assuming you have the directory

```
my_repo/package_a.rpm
my_repo/package_b.rpm
```

you can go in that dir and call

```
createrepo .
```

which will create you repo metadata. Once there is repo metadata you don't need to specify the repo type in the kiwi XML description anymore because package managers will now find out temselves.

Respectively if you have other types of packages, e.g .debs you need to create repo metadata using the tools that exists for this case.

I hope this makes sense to you and helps you to move forward.

Thanks

zohourih:
[...]

PR #1976, opened 2021-11-12
Fixed include processing

schaefi:
This commit fixes several issue connected with the use of
the <include> directive:

First and foremost the XSLT chain was broken in a way that
the include XSLT in combination with the PrettyPrinter XSLT
were called not in the chain of stylesheets but together.
This results in XML descriptions which duplicated the content
and went invalid

Another change is, when the include XSLT is called in the chain.
This commit moves it to become the very first processing
instruction such that the included data is part of all subsequent
XSLT stylesheets. This also allows to use older schema versions
in included XML data and they get automatically converted through
the chain of XSLT stylesheets.

Last change is the evaluation of the from= attribute value. This
value is now interpreted as an URI. Currently only local URIs are
supported. The reason to do this is because XSLT when processing
a document resolves relative paths according to the file path
of the master document. As kiwi does not change the original
content that path with will be a /var/tmp location if one of
the XSLT stylesheets were used. The documentation for this change
was updated as well

### DIFF
--- a/doc/source/image_description.rst
+++ b/doc/source/image_description.rst
@@ -101,7 +101,7 @@ Image Includes
 
 .. code:: xml
 
-   <include from="filename.xml"/>
+   <include from="file://filename.xml"/>
 
 The optional :ref:`sec.include` element allows to drop in the contents
 of the specified :file:`filename.xml` file at the place were the `include`
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -51,7 +51,7 @@ Optional include of XML file content from file
 .. code:: xml
 
    <image schemaversion="{schema_version}" name="{exc_image_base_name}">
-       <include from="description.xml"/>
+       <include from="file://description.xml"/>
    </image> 
 
 with file :file:`description.xml` as follows:
@@ -68,7 +68,9 @@ with file :file:`description.xml` as follows:
 
 This will replace the `include` statement with the contents
 of :file:`description.xml`. The validation of the result happens
-after the inclusion of all `include` references.
+after the inclusion of all `include` references. The value for
+the `from` attribute is interpreted as an URI, as of now only
+local URI types are supported.
 
 .. note::
 
--- a/kiwi/markup/base.py
+++ b/kiwi/markup/base.py
@@ -17,6 +17,7 @@
 #
 import os
 from lxml import etree
+from urllib.parse import urlparse
 
 # project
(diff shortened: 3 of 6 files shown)


schaefi:
@Conan-Kudo with this change I was able to produce a valid document with the data you provided

schaefi:
It produces the following document

[config.final.txt](https://github.com/OSInside/kiwi/files/7527215/config.final.txt)

schaefi:
strange github does not support upload of xml files

doc/source/image_description.rst
```diff
@@ -101,7 +101,7 @@ Image Includes
 
 .. code:: xml
 
-   <include from="filename.xml"/>
+   <include from="file://filename.xml"/>
```
Conan-Kudo, line 104: Crap, does this mean we probably want to re-introduce the `this://` URI handler as asked in #1964? Or are we going to automatically prepend `file://` if not specified?

Conan-Kudo (changes requested):

doc/source/image_description.rst
schaefi: I think it would be a good idea to have this:// supported because it makes things explicit. But this should be done in another PR. So far you can still use the same settings as you had before, Examples:

1. from="file:///some/path/to/file.xml"
2. from="file://../../data/some/relative/path/to/file.xml"
3. from="/some/oath/to/file.xml"
4. from="../some/relative/path/to/file.xml"
 
So these are all accepted but the behavior is not really deterministic for 2. and 4.

What relative means is not clear. It depends on the pwd if no stylesheets worked on it, it depends on if you nest includes, it depends on if a schema version needs to be converted into a newer one because the converted result will come from /var/tmp

So overall I think with includes it should be clear, either by an absolute path name or by this:// which resolves into the image description directory

Thoughts ?

Conan-Kudo: So if I do `foo.xml`, it should resolve to `./foo.xml`, which resolves to `file://./foo.xml`, which resolves to the file relative to the working directory currently?

schaefi: yes, that's what also the unit test does.

schaefi: I realized that relative locations can be weird when the unit test which tested nested includes failed. So with this PR there are no behavior changes to what you saw before, we just have to realize that relative include locations can be weird under certain conditions and these conditions are not obvious to users imho

Conan-Kudo (approved):

PR #1977, opened 2021-11-12
Allow to set sign keyfile in repository elements

### LINKED ISSUE #1883: Allow signing-keys to be specified in the XML description
## Problem description

kiwi allows to specify repo signing keys via the command line but not via the XML description

## Expected behaviour

Extend the XML description with a section to define signing keys to be imported
like it is done by the custom suseImportBuildKey method (#1876)

schaefi:
This commit add a new attribute signing_key to the repository
element. The keyfile specified is placed into the keyring as
used by the selected package manager. Signing keys specified
on the commandline and signing keys specified in the image
description will be combined. This Fixes #1883

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -793,7 +793,9 @@ repository_gpgcheck="true|false"
   to run repository signature validation. If not set, no value is
   appended into the repository configuration file. If set the
   relevant key information needs to be provided on the {kiwi}
-  commandline using the `--signing-key` option
+  commandline using the `--signing-key` option or via the `<signing>`
+  element as part of the `<repository><source>` setting in the
+  image description.
 
 customize="/path/to/custom_script"
   Custom script hook which is invoked with the repo file as parameter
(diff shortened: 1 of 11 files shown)


schaefi:
With this path you can set something like this in a repository definition

```xml
<repository signing_key="key_b">
    <source path="..."/>
</repository>
```

Conan-Kudo:
Does this let you set multiple keys for a repository? For example, the openSUSE Leap repo requires multiple ones: https://code.opensuse.org/package/rpm-repos-openSUSE/blob/master/f/opensuse-leap-oss.repo#_6-8

schaefi:
> Does this let you set multiple keys for a repository? For example, the openSUSE Leap repo requires multiple ones: https://code.opensuse.org/package/rpm-repos-openSUSE/blob/master/f/opensuse-leap-oss.repo#_6-8

not with the current implementation. I thought one repo can only have one signing key. But ok in this case it might be good to change the way how we describe this. I was thinking about the following

```xml
<repository signing_key="key_b">
    <signing key="key_a"/>
    <signing key="key_b"/>
    <source path="..."/>
</repository>
```

Thoughts please before I change the code. Thanks

Conan-Kudo:
@schaefi That looks good to me.

schaefi:
ok, the way to add signing keys into the description has slightly changed according to my last comment. The way to specify
signing keys is as follows:

```xml
<repository>
    <source path="...">
        <signing key="keyfile_a"/>
        <signing key="keyfile_b"/>
    </source>
</repository>
```

The reason why this became a sub element of the ```<source>``` is because the data structure to parse the repository would have changed with an addition element. I don't want to introduce this python API change and added the optional sub element below ```<source>``` which doesn't have sub elements to parse, thus only new python data structure elements gets created and no incompatible change will be introduced.

Conan-Kudo:
LGTM, merged!

PR #1981, opened 2021-11-18
Pass signing keys from the XML to the repositories

davidcassany:
This commits makes sure signing keys are passed to
repositories setup in build task.

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -38,6 +38,7 @@ portnum-type = xsd:token {pattern = "(\d+|\d+/(udp|tcp))"}
 grub_console = xsd:token {pattern = "(console|gfxterm|serial)( (console|gfxterm|serial))*"}
 fs_attributes = xsd:token {pattern = "(no-copy-on-write|synchronous-updates)(,(no-copy-on-write|synchronous-updates))*"}
 package-version-type =  xsd:token {pattern = "(0|[1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])(\.(0|[1-9][0-9]{0,3}|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])){3}"}
+simple-uri-type = xsd:token {pattern = "(file:|https:|http:|ftp:).*"}
 
 #==========================================
 # start with image description
@@ -1032,7 +1033,7 @@ div {
 div {
     k.signing.key.attribute =
         ## Specify path to a signing key for this repo
-        attribute key { text }
+        attribute key { simple-uri-type }
     k.signing.attlist =
         k.signing.key.attribute
     k.signing =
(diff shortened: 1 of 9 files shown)


davidcassany:
Looks like gpg does not work with local URIs (`file://...`)

Conan-Kudo:
> Looks like gpg does not work with local URIs (`file://...`)
> 
> @Conan-Kudo to me it looks like you XML includes the `<signing key="file:///usr/share/keyrings/ubuntu-archive-keyring.gpg"/>`, could you test with `<signing key="/usr/share/keyrings/ubuntu-archive-keyring.gpg"/>`. If URIs are needed because we want to consider remote signing keys we can consider this, IIRC rpm already allows to download keys, it calls curl behind the scenes to download remote paths. I doubt gpg supports it, but I will have a look.

I've tried both ways and it's broken both ways. We need to have XML require local URIs and strip the `file://` prefix when passing to GPG for apt.

davidcassany:
@Conan-Kudo @schaefi right this is what I came out with in order to force signing keys to be passed as URIs in XML. We discussed about it matrix chat. I am not fully convinced by this approach, using URIs for local paths tends to be confusing to me, specially for relative paths.

Anyway, the logic of the current implementation relies on kiwi.system.uri utility and its translate method, which turns local URIs into a simple absolute path (resolves relative paths based on the working directory). The schema allows only `file`, `http`, `https` and `ftp` URIs. I did it this way to pair with supported schemas in `kiwi.system.uri` at the same time we do not allow any of the custom schemas such `obs` or `obsrepositories`. Remote URIs are simply passed through unless there is some special convention, which is not the case for any of the allowed remote URIs.

I am not sure about this implementation I am not sure if it would be just simpler to let user enter free text and do not manipulate it. Also I do not want to code a bunch of new specific logic related to URIs, so if we want to use URIs I want to make use of `kiwi.system.uri` which is a class that might not suit perfectly as include some OBS tricks and it was mostly coded having in mind repository URIs.

Any thoughts?

Conan-Kudo:
@davidcassany I think this makes sense and it makes sense to me.

davidcassany:
> @davidcassany I think this makes sense and it makes sense to me.

Cool, that's great. Now lets see if this is fine for @schaefi, since he introduced the signing feature few days ago I'd like him to also approve it before merging.

Conan-Kudo (approved):
Works for me, and my image builds work now.

schaefi (approved):
This looks great to me :+1:  I read the conversation on the uri reference for signing keys and it makes perfect sense to me. I also realized I forgot to add the signing key reading for the "kiwi system build" tasks when I added the support for it. @davidcassany Thanks much for catching this too

The only part I was thinking a bit about is that we restrict the setting of the key value in the schema. As we pass this along to the Uri class which applies an urlparse on it we have now two places to verify the correctness of the Uri. In addition a simple value not starting with an URL type will not be allowed by the schema

I'm fine if we restrict it and it's probably the better approach, just something that came to my mind.

davidcassany:
> As we pass this along to the Uri class which applies an urlparse on it we have now two places to verify the correctness of the Uri.

That's correct. I know this is not ideal, the problem here is that `Uri` class supports a wider range of what we want here and I preferred to add constraints in schema rather than modifying `Uri` class logic.

> In addition a simple value not starting with an URL type will not be allowed by the schema

Yes, that's the whole purpose of it. That's what we discussed with @Conan-Kudo on matrix. At the end we agreed this is a way to be consistent with repository URLs definition, so enforcing a URL even if it is a local one. Note this only applies for the XML schema. Signing keys applied from the command line are not processed by `Uri` class or any other utility, the string passed there just goes on to whatever KIWI uses behind the scenes. So, the XML input is more strict regarding format.

schaefi:
@davidcassany ok perfect, then just merge as you see fit. Thanks

PR #1982, opened 2021-11-18
Fix Source Install Disk Detection in get_disk_list

TheMisanthropicHumanist:
The original check doesn't always work because PARTITIONS get labels, NOT disks. If you install with a Flash Drive and deploy to a machine with an NVMe drive, the Flash Drive is listed first because SCSIs are shown first by lsblk (At least they were in our scenario). When we run 'blkid "[Install Flash Drive Disk]" -s LABEL -o value' in this scenario, we get nothing returned so this conditional evaluates incorrectly. We should get the ${kiwi_install_volid} Partition Block Name, and search for whether ${disk_device} is contained within it as the Disk Device that corresponds to the Install Partition.

Ex:
     Flash Drive ${disk_device} '/dev/sda' is a substring of ${kiwi_install_disk_part} '/dev/sda1', which has a LABEL/${kiwi_install_volid} of 'INSTALL'. Therefore, we should ignore this ${disk_device}

Because Disk Device + Partition Block Names are standardized and additive for NVMes, SCSIs, and every other Type, this method of detection of the Install Device will work under every scenario. The Disk Name is always the first part of the Partition Name.

issue #1983

### DIFF
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -77,10 +77,15 @@ function get_disk_list {
     elif [ "${kiwi_oemmultipath_scan}" = "true" ];then
         scan_multipath_devices
     fi
+    kiwi_install_disk_part=$(
+        eval lsblk "${blk_opts_plus_label}" | \
+        tr -s ' ' ":" | \
+        grep ":${kiwi_install_volid}$" | \
+        cut -f1 -d:
+    )
     for disk_meta in $(
         eval lsblk "${blk_opts}" | grep -E "disk|raid" | tr ' ' ":"
     );do
-        kiwi_install_disk_part=$(eval lsblk "${blk_opts_plus_label}" | tr ' ' ":" | grep ":${kiwi_install_volid}$" | cut -f1 -d:)
         disk_device="$(echo "${disk_meta}" | cut -f1 -d:)"
         if [[ "${kiwi_install_disk_part}" == "${disk_device}"* ]]; then
             # ignore install source device
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -80,7 +80,7 @@ function get_disk_list {
     for disk_meta in $(
         eval lsblk "${blk_opts}" | grep -E "disk|raid" | tr ' ' ":"
     );do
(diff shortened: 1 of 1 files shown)


TheMisanthropicHumanist (commented):
Oops. The Options passed to `lsblk` need to be unquoted because the command syntax fails when they're grouped together as a single argument. The other instance of `lsblk`/`blkid` in the script + arguments passed as a quoted string of variables utilized an `eval` to double-expand the string of arguments into multiple individual argument strings.

schaefi:
Thanks for your effort and deep dive into the issue. This is dracut code and a sensitive area. Please give us some time to look at it. Thanks

@davidcassany would be great if you can have a look at that change as well, checking the after effects of initrd code changes are always good with some eyes more :) Thanks

schaefi:
@TheMisanthropicHumanist If you can provide us the ```lsblk``` output that appears in your use case it would be great and easier to review. Thanks

dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
```diff
@@ -78,10 +80,9 @@ function get_disk_list {
     for disk_meta in $(
         eval lsblk "${blk_opts}" | grep -E "disk|raid" | tr ' ' ":"
     );do
+        kiwi_install_disk_part=$(eval lsblk "${blk_opts_plus_label}" | tr ' ' ":" | grep ":${kiwi_install_volid}$" | cut -f1 -d:)
```
schaefi, line 83: The change looks good to me, this is a better way to find the partition labeled with the install label. I think the tr call can be changed to

```bash
| tr -s ' ' : |
```
-s, --squeeze-repeats

Just in case there are more spaces playing tricks

schaefi (approved):
Looks good :+1:

dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
schaefi: and it would be nice to break up the line to not exceed the 80characters... just a policy we follow for initrd shell code and because I'm an old man ;)

TheMisanthropicHumanist: Thank you! I'll update the change as indicated, and add the lsblk output and blkid output to the conversation when I get back into the office.

I also just realized, it's probably unnecessary to calculate the install partition on each iteration of the loop. Not that we care too much about performance in this particular loop, but it would technically perform better if it was calculated outside the loop.

schaefi: yep good point, makes perfect sense to me too

schaefi (approved):
Looks great to me, thanks :+1:

PR #1987, opened 2021-11-23
Do not force dracut into a compression setting

schaefi:
So far we called dracut with --xz which forces the initrd
to be xz compressed. There are other compression formats
used by the distributions and they might differe from xz.
The selection for a compression tool is done by a dist
configuration in dracut.conf.d which is provided by the
distributions as they see fit. For us this means not
forcing dracut into a specific compression setting allows
to make use of the distro provided setting and also
allows to change/override this setting by an overlay
file. This Fixes bsc#1192975

### DIFF
--- a/doc/source/building_images/build_expandable_disk.rst
+++ b/doc/source/building_images/build_expandable_disk.rst
@@ -181,7 +181,7 @@ target system:
 
       .. code:: bash
 
-          scp pxeboot.{exc_image_base_name_disk}.x86_64-{exc_image_version}.initrd.xz PXE_SERVER_IP:/srv/tftpboot/boot/initrd
+          scp pxeboot.{exc_image_base_name_disk}.x86_64-{exc_image_version}.initrd PXE_SERVER_IP:/srv/tftpboot/boot/initrd
           scp pxeboot.{exc_image_base_name_disk}.x86_64-{exc_image_version}.kernel PXE_SERVER_IP:/srv/tftpboot/boot/linux
 
 3. Copy the disk image, MD5 file, system kernel, initrd and bootoptions to
--- a/doc/source/building_images/build_kis.rst
+++ b/doc/source/building_images/build_kis.rst
@@ -77,7 +77,7 @@ tested with QEMU as follows:
 
    $ sudo qemu
        -kernel /tmp/myimage/*.kernel \
-       -initrd /tmp/myimage/*.initrd.xz \
+       -initrd /tmp/myimage/*.initrd \
        -append $(cat /tmp/myimage/*.append) \
        -hda /tmp/myimage/{exc_image_base_name_pxe}.*-{exc_image_version} \
        -serial stdio
--- a/doc/source/working_with_images/legacy_netboot_root_filesystem.rst
+++ b/doc/source/working_with_images/legacy_netboot_root_filesystem.rst
@@ -87,7 +87,7 @@ system. As diskless client, a QEMU virtual machine is used.
 
    .. code:: bash
 
-       $ cp *.initrd.xz /srv/tftpboot/boot/initrd
+       $ cp *.initrd /srv/tftpboot/boot/initrd
        $ cp *.kernel /srv/tftpboot/boot/linux
 
 5. Copy the system image and its MD5 sum to :file:`/srv/tftpboot/image`:
--- a/doc/source/working_with_images/network_overlay_boot.rst
+++ b/doc/source/working_with_images/network_overlay_boot.rst
@@ -52,7 +52,7 @@ from the network:
 
    .. code:: bash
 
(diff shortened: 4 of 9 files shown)


schaefi:
@Conan-Kudo @davidcassany your feedback would be appreciated regarding the scope of this change. Thanks much

Conan-Kudo (approved):

dcermak (approved):

davidcassany (commented):
I am wondering the implications this change has to the dracut modules, more specific, to the `kiwi-dump` module and `kiwi-lib` module. I see there is the expectation that compressed files are xz compressed.

Could it be that this change needs to be reflected somehow in dracut modules?

What about `fetch_file` method from `kiwi/dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh`? At a glance this looks like the only place where files are uncompressed and this is used in the scope of `kiwi-dump` module.

I am not sure now if we ever decompress the initrd in those modules, I'd say no. But I see we use this method to fetch initrds in `kiwi-dump`. I believe we should test few deployment methods prior merging this change. I am running out of time today, tomorrow I can give it a deeper look if it is not too late.

schaefi:
network deployment is a very good point. So we have documented this here:

* https://osinside.github.io/kiwi/building_images/build_expandable_disk.html#network-deployment

We expect users to copy the initrd and other data over to the boot server. For that purpose we describe it as

```
scp pxeboot.kiwi-test-image-disk.x86_64-1.15.3.initrd.xz PXE_SERVER_IP:/srv/tftpboot/boot/initrd
```

At this point an initrd that is not using the .xz extension would need to be copied differently. I suggest an update of the docs

Next the initrd is referenced in the boot server setup as:

```
append initrd=boot/initrd ...
```

At this point due to the copy of the initrd into a "common" name this should not become an issue for users.
From a remote deployment point of view I suggest to add a hint to the docs

Looking at the dump module next

schaefi:
In ./99kiwi-lib/kiwi-net-lib.sh we have

```
function _is_compressed {
    local source_url=$1
    [[ ${source_url} =~ .xz$ ]]
}
```

Used only in the scope of

```
if _is_compressed "${source_url}";then
        fetch="${fetch} | xz -d"
    fi
```

So imho not related to the initrd compression. Please confirm

schaefi:
In ./90kiwi-dump/kiwi-dump-image.sh we have

```
function get_remote_image_source_files {
    ...
    image_initrd_uri=$(
        echo "${image_uri}" | awk '{ gsub("\\.xz",".initrd", $1); gsub("dolly:","tftp:", $1); print $1 }'
    )
```

This is an issue !

schaefi:
Hrm, so yes this change is creating issues :(

schaefi:
Next there is boot/image/dracut.py

```
dracut_initrd_basename += '.xz'
```

darn... this change is creating many issues

@davidcassany thanks David for bringing this up

schaefi:
So we call dracut that is creates the initrd file name with the ".xz" extension. If we delete the ```--xz``` from the commandline we would still create a file with the extension ".xz", that avoids issues with the further processing of the files and the later use in the deployment code but it is very misleading

Maybe we should think about creating the initrd with a compression independent suffix and change all places in the dracut module code and elsewhere

Thoughts ?

davidcassany:
> Maybe we should think about creating the initrd with a compression independent suffix and change all places in the dracut module code and elsewhere

I think this could be good option. It could potentially fix some these issues in a straight forward manner.

I am still wondering about

```bash 
image_initrd_uri=$(
        echo "${image_uri}" | awk '{ gsub("\\.xz",".initrd", $1); gsub("dolly:","tftp:", $1); print $1 }'
    )
```
I was thinking about it yesterday. I don't see it as an issue, the initrd URI is created from the image uri and changes the extension `.xz` to `.initrd`, so this prevents the decompression in `fetch_file` method. I had the impression yesterday that this was still functional, but there are details I might miss, I don't know by heart logic and it is not that simple to follow all execution paths by looking at dracut module code.

schaefi:
> I think this could be good option.

ok  have done this now with another commit. @davidcassany could you give this a pedantic review :) Thanks much. I have build an image that also creates an install image and tested the deployment, works so far. I have not tested the installpxe though.

schaefi:
> I am still wondering about

yes this was a false positive from my side. it should be safe before and after the change since there is no change on the image file name. I found a bugs on the docs though and also fixed it here

davidcassany (approved):
Looks good to me. I have not executed the install media out of this PR but the code looks good to me. If the initrd is not including the xz suffix the dracut module code should be safe.

davidcassany:
Just for the records I successfully pxe deployed an OEM image (using `installpxe`). 👍🏽

PR #1988, opened 2021-11-25
Prevent warning message on resize exception

### LINKED ISSUE #1958: Prevent 'scary' warning message in dracut log
## Problem description

When using the ```dracut-kiwi-oem-repart``` dracut module the feature to repart/resize your disk to changing
disk geometry is active. However, if the disk geometry does not change the module provides a "scary" message
saying:

```
Warning: Requested OEM systemsize exceeds free space on the disk:
Warning: Disk won't be re-partitioned !
Warning: Requested size(s): root:(keep)
Warning: ==> Free Space on disk: 1 MB
```

A report from a user describes this as such:

```
Initially you write the image onto a 8GiB microSD and use it in the RPi. On the first boot,
kiwi resizes the filesystem as expected and all the space can be used. At some point you notice
that you need more space, so you switch to a 16GiB card by just copying the data over. On the
next boot, kiwi notices that there's free space and resizes again to make it available to the filesystem.

It's however much more common that the system is booted on the same device than a large
device, but kiwi shows a (scary!) warning on all of those boots, even though this is the expected
situation. That's why I'm suggesting to just not show a warning at all in that case, maybe
log an info message.
```

And I agree we could make the message less "scary" :-)

schaefi:
On systems which are configured to run the oem resize at
every boot (default case) kiwi checks how much space is free
and if that fits the constraints configured as part of the
image description. If the constraints are not met a warning
message is displayed and the boot continuous without any
resize action happening.

This warning message however, always appears after the first
boot when the resize had happened and no rest space on disk
is present unless the disk geometry would have changed.
The situation of the reboot of the system without any disk
geometry change is the standard case and happens way more
often than the reboot with a disk geometry change.

Therefore the warning message displayed is not actually
a real warning and most often considered as an issue
when there is none. To relax this situation, this commit
only shows the warning message if the detected free space
on disk is greater than 2M, which is the condition under
which we assume an intentional (user made) disk geometry
change.

This Fixes #1958

### DIFF
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -183,18 +183,29 @@ function check_repart_possible {
         fi
     fi
     if [ "${min_additional_mbytes}" -gt "${disk_free_mbytes}" ];then
-        # Requested size for root exceeds free space on disk
-        local requested_size
-        if [ -n "${kiwi_oemrootMB}" ];then
-            requested_size="root:($((kiwi_oemrootMB - disk_root_mbytes)) MB)"
+        if [ "${disk_free_mbytes}" -gt 2 ];then
+            # Requested size for root exceeds free space on disk
+            local requested_size
+            if [ -n "${kiwi_oemrootMB}" ];then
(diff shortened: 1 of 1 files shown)


Vogtinator:
Change looks good to me!

Conan-Kudo (approved):

PR #1989, opened 2021-11-25
Care for different snapper template locations

schaefi:
snapper recently changed their config template location
from etc/ to usr/. This commit handles the two locations
and Fixes bsc#1192940

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -470,15 +470,26 @@ class Defaults:
         return 'INSTALL'
 
     @staticmethod
-    def get_snapper_config_template_file():
+    def get_snapper_config_template_file(root: str) -> str:
         """
-        Provides the default configuration template file for snapper
+        Provides the default configuration template file for snapper.
+        The location in etc/ are preferred over files in usr/
 
-        :return: file
+        :return: file path
 
         :rtype: str
         """
-        return '/etc/snapper/config-templates/default'
+        snapper_templates = [
+            'etc/snapper/config-templates/default',
+            'usr/share/snapper/config-templates/default'
+        ]
+        snapper_default_conf = ''
+        for snapper_template in snapper_templates:
+            template_config = os.path.join(root, snapper_template)
+            if os.path.exists(template_config):
+                snapper_default_conf = template_config
+                break
+        return snapper_default_conf
 
     @staticmethod
     def get_default_video_mode():
--- a/kiwi/volume_manager/btrfs.py
+++ b/kiwi/volume_manager/btrfs.py
@@ -387,10 +387,10 @@ class VolumeManagerBtrfs(VolumeManagerBase):
 
     def _create_snapper_quota_configuration(self):
         root_path = os.sep.join([self.mountpoint, '@/.snapshots/1/snapshot'])
-        snapper_default_conf = os.path.normpath(os.sep.join(
-            [root_path, Defaults.get_snapper_config_template_file()]
-        ))
-        if os.path.exists(snapper_default_conf):
+        snapper_default_conf = Defaults.get_snapper_config_template_file(
+            root_path
+        )
+        if snapper_default_conf:
             # snapper requires an extra parent qgroup to operate with quotas
             Command.run(
                 ['btrfs', 'qgroup', 'create', '1/0', self.mountpoint]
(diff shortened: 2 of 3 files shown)


kiwi/defaults.py
```diff
@@ -470,15 +470,24 @@ def get_install_volume_id():
...
+        snapper_templates = [
+            'etc/snapper/config-templates/default',
+            'usr/share/snapper/config-templates/default'
+        ]
+        snapper_default_conf = ''
+        for snapper_template in snapper_templates:
+            snapper_default_conf = os.path.join(root, snapper_template)
+            if os.path.exists(snapper_default_conf):
+                break
+        return snapper_default_conf
```
Vogtinator, line 490: If neither of those files exists, it returns the last entry of the list instead of an empty string

kiwi/volume_manager/btrfs.py
```diff
@@ -387,10 +387,10 @@ def _xml_pretty(self, toplevel_element):
...
     def _create_snapper_quota_configuration(self):
         root_path = os.sep.join([self.mountpoint, '@/.snapshots/1/snapshot'])
-        snapper_default_conf = os.path.normpath(os.sep.join(
-            [root_path, Defaults.get_snapper_config_template_file()]
-        ))
-        if os.path.exists(snapper_default_conf):
+        snapper_default_conf = Defaults.get_snapper_config_template_file(
+            root_path
+        )
+        if snapper_default_conf:
```
Vogtinator, line 393: IMO the `os.path.exists` call can stay

Vogtinator (changes requested):

kiwi/defaults.py
schaefi: oh yes, true my bad, will fix it

kiwi/volume_manager/btrfs.py
schaefi: the idea was that the string is empty when no files could be found... but the logic is broken as you found out

kiwi/defaults.py
schaefi: should be fixed now

kiwi/defaults.py
```diff
@@ -470,15 +470,26 @@ def get_install_volume_id():
...
         """
-        return '/etc/snapper/config-templates/default'
+        snapper_templates = [
+            'etc/snapper/config-templates/default',
+            'usr/share/snapper/config-templates/default'
+        ]
+        snapper_default_conf = ''
+        for snapper_template in snapper_templates:
+            template_config = os.path.join(root, snapper_template)
+            if os.path.exists(snapper_default_conf):
```
Vogtinator, line 489: Checks for the wrong file, the unit test elegantly works around that bug...

```suggestion
            if os.path.exists(template_config):
```

Vogtinator (changes requested):

kiwi/defaults.py
```diff
@@ -470,15 +470,26 @@ def get_install_volume_id():
...
+        :return: file path
 
         :rtype: str
         """
-        return '/etc/snapper/config-templates/default'
+        snapper_templates = [
+            'etc/snapper/config-templates/default',
+            'usr/share/snapper/config-templates/default'
+        ]
+        snapper_default_conf = ''
```
Vogtinator, line 486: I suggest to just return early:

```
        for snapper_template in snapper_templates:
            template_config = os.path.join(root, snapper_template)
            if os.path.exists(template_config):
                return template_config

        return None
```

kiwi/defaults.py
schaefi: I need vacation, sure you are right

kiwi/defaults.py
schaefi: This introduces a typing issue which means the method can either return a "str" or a None type. This could be expressed with Optional[str] but I don't like this much and prefer a static type in any case.

I should just stop making stupid mistakes :/ thanks for reading all this

Vogtinator: In that case just:

```
        for snapper_template in snapper_templates:
            template_config = os.path.join(root, snapper_template)
            if os.path.exists(template_config):
                return template_config

        return ''
```

Vogtinator (approved):

Conan-Kudo (approved):

PR #1990, opened 2021-11-25
Allow repo path relative to the image description

### LINKED ISSUE #1964: Support for "this://" for repository paths

schaefi:
This commit adds a new URI type called this://... The
this:// part will be resolved into the absolute path to
the image description. A source path like the following:

```xml
    <source path="this://my_repo"/>
```

is resolved to

```xml
    <source path="dir:///path/to/image/description/my_repo"/>
```

This change provides the requested opportunity to reference
repos provided as part of the image description and
Fixes #1964

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -70,7 +70,9 @@ This will replace the `include` statement with the contents
 of :file:`description.xml`. The validation of the result happens
 after the inclusion of all `include` references. The value for
 the `from` attribute is interpreted as an URI, as of now only
-local URI types are supported.
+local URI types are supported as well as the `this://` resource
+locator which translates into the path to the KIWI image
+description.
 
 .. note::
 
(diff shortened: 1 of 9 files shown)


Conan-Kudo:
Would it make sense for `include` directives to support this too?

schaefi:
> Would it make sense for `include` directives to support this too?

yes we could add that

Conan-Kudo:
This is relative to the description directory passed into `--description`, right?

schaefi:
> This is relative to the description directory passed into `--description`, right?

yes, the information is turned into an abspath internally though

Conan-Kudo:
That means that the symlinks should be resolved, correct?

schaefi:
> That means that the symlinks should be resolved, correct?

nope, abspath doesn't, realpath() does. I will change this

PR #1993, opened 2021-11-29
Allow to specify kiwi file

### LINKED ISSUE #1973: RFE: Add ability to specify exact *.xml/*.kiwi file to use for an image build
## Problem description

I have a kiwi description repo that has multiple `*.xml` or `*.kiwi` files that I want to be able to build images of. However, kiwi currently only works with a single xml/kiwi file in a directory. It'd be great to have an option to specify the filename of the description in the directory to look up as a way to override the current auto-selection behavior.

schaefi:
Add global --kiwi-file option
    
When building with kiwi a search on the kiwi main config file is made inside of the given --description directory. The search looks up for the file ```config.xml``` or ```*.kiwi```.  So far there was no opportunity to specify another name.  This commit adds an option in the global area named:
    
```bash
--kiwi-file name
```
    
which will make kiwi to lookup this file inside of the given ```--description``` directory and fail if it does not exist.

This Fixes #1973

### DIFF
--- a/doc/source/commands/kiwi.rst
+++ b/doc/source/commands/kiwi.rst
@@ -18,6 +18,7 @@ SYNOPSIS
            [--debug]
            [--color-output]
            [--config=<configfile>]
+           [--kiwi-file=<kiwifile>]
        image <command> [<args>...]
    kiwi-ng [--debug]
            [--color-output]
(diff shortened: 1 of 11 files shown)


schaefi:
Example use:

```
kiwi-ng --kiwi-file my_kiwi.xml system build --description /path/to/image/description --target-dir /tmp/myimage
```

* expects: ```/path/to/image/description/my_kiwi.xml```

will fails if not present

Conan-Kudo (approved):

PR #1994, opened 2021-11-29
Added support for setting up release version

### LINKED ISSUE #1927: Add `$releasever` support
Currently, we set `--releasever 0` for DNF and Micro DNF in order to enable bootstrapping. However, in order to support leveraging the same description across different Fedora releases with templated repository URLs, we need to be able to pass a real value in from the command-line or the XML description to set this value. If not set, we should continue to set it to zero when the package manager supports `--releasever`.

schaefi:
Currently the release version is not set or set to '0' for package managers which requires a value to operate. However, in order to support leveraging the same description across different releases it is important to have the opportunity to specify a setting for the release version. This commit adds a new optional attribute to the preferences section which allows to specify a custom value which serves as the release version:

```xml
<preferences>
    <release_version>TEXT</release_version>
</preferences>
```

If not specified the default setting as before applies. Please note the release version information is currently used in dnf and microdnf package managers only. It might happen that it gets applied to the other package manager backends as well. This will happen on demand though.

Related to Issue #1918. This Fixes #1927

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -289,6 +289,22 @@ bootloader has theme support.
 Along with the version and the packagemanager at least one image type
 element must be specified to indicate which image type should be build.
 
+<preferences><release-version>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Specifies the distribution global release version as consumed
+by package managers. Currently the release version is not set or
+set to `0` for package managers which requires a value to operate.
+With the optional `release-version` section, users have an
+opportunity to specify a custom value which is passed along the package
+manager to define the distribution release.
+
+.. note::
+
+   The release version information is currently
+   used in dnf and microdnf package managers only. It might
+   happen that it gets applied to the other package manager
+   backends as well. This will happen on demand though.
+
 <preferences><type>
 ~~~~~~~~~~~~~~~~~~~
 At least one type element must be configured. It is possible to
--- a/kiwi/package_manager/__init__.py
+++ b/kiwi/package_manager/__init__.py
@@ -50,7 +50,7 @@ class PackageManager(metaclass=ABCMeta):
     @staticmethod
     def new(
         repository: object, package_manager_name: str,
-        custom_args: List=None  # noqa: E252
+        custom_args: List = [], release_version: str = ''
     ):
         name_map = {
             'zypper': ['zypper', 'Zypper'],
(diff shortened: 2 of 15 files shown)


schaefi:
@Conan-Kudo if you look at this, please take a special look on the change in the ```dnf.py``` code. We only explicitly set --releasever in microdnf and this change here also applies the setting to dnf. I think I did that correctly but I haven't tested it. The integration tests will outline issues but maybe we can double check before

Conan-Kudo:
I will test this later today and give the appropriate feedback. 😉

schaefi:
@Conan-Kudo any news on this one ?

Conan-Kudo:
I get the following error with this patch:

```
[root@a75839102358 kiwi-desc]# kiwi-ng --type=oem --profile=OpenStack system build --description ./centos9 --target-dir $PWD/tmpoutput
[ INFO    ]: 14:59:24 | Reading runtime config file: '/etc/kiwi.yml'
[ INFO    ]: 14:59:24 | Loading XML description
[ INFO    ]: 14:59:24 | Support for multiple markup descriptions available
[ WARNING ]: 14:59:24 | Command "jing" not found in the environment
[ WARNING ]: 14:59:24 | For detailed schema validation report, install: jing
[ INFO    ]: 14:59:24 | Showing only raw library error log:
[ INFO    ]: 14:59:24 | --> /var/tmp/kiwi_xslt-6al3mple:14:0:ERROR:RELAXNGV:RELAXNG_ERR_EXTRACONTENT: Element preferences has extra content: release_version
[ ERROR   ]: 14:59:24 | KiwiDescriptionInvalid: Failed to validate schema and/or schematron rules. Use --debug for more details
```

schaefi:
```
div {
    k.release-version.content = text
```

The element is named ```<release-version>``` not ```<release_version>```

I check if I did a mistake in the docs

schaefi:
... make me realize I forgot to add the docs, shame on me.

schaefi:
@Conan-Kudo, I've updated the PR and added the missing doc hint. Also fixed the commit message which uses the wrong section name. If you adapt your example and change '_' to '-' your XML will be valid. As a side note, if you have jing installed you get way better error messages as the ones from libXML

Conan-Kudo:
@schaefi It's not being passed in for bootstrap, which causes it to fail.

Conan-Kudo (approved):
It works now!

schaefi:
@Conan-Kudo great, thanks much for testing. I'll merge and make a Staging update to see if all tests pass

PR #1997, opened 2021-11-30
Compress container images in builder class

### LINKED ISSUE #1996: Do not compress OCI images within the OCI class
## Problem description

Docker and OCI images are compressed within the OCI class while in all other images types the compression is applied at builder class as part of the result.

## Expected behaviour

The compression of OCI images is managed the same way as the other image types, by setting the result to be compressed or not.

davidcassany:
This commit changes the stage at which container images are compressed
to match the criteria applied to other image types. Instead of
compressing the image in OCI devoted classes now it is happening
in builder class by setting Result instance properties.

Fixes #1996

### DIFF
--- a/kiwi.yml
+++ b/kiwi.yml
@@ -39,9 +39,9 @@
 
 # Setup process parameters for container image creation
 #container:
-#  # Specify compression algorithm for container images
-#  # Possible values are xz or none
-#  - compress: xz
+#  # Specify compression for container images
+#  # Possible values are true, false, xz or none.
+#  - compress: true
 
 
 # Setup process parameters for ISO image creation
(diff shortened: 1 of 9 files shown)


davidcassany:
This PR includes a couple of small behavior changes, both acceptable IMHO:

* `kiwi-ng system build` no longer produces compressed docker images, the result is `*.tar` (still loadable with `docker load`) image, since the compression, if any, happens on `kiwi-ng result bundle`. This means the outcome of `kiwi-ng system build` plus `kiwi-ng result bundle` is the **same** as it was before. So from OBS perspective the outcome doesn't change.

* There are no choices for container images compression, they are only compressed with `xz`. In fact, before that there were no choices too, but the code was ready to include further compression algorithms. Now you can only choose to compress or not, you are not passing the compression algorithm anymore. The `xz` option in config.yml is accepted just to keep backward compatibility.

schaefi:
Nice changes :+1: Thanks much

I think there is one file

```
kiwi.yml
```

which needs to be adapted as well

PR #2004, opened 2021-12-16
index.rst: Change title (bsc#1189294#c2)

taroth21:
Reference: bsc#1189294#c2 

* title change: 'KIWI NG 9: KIWI NG Documentation' -> 'Building Linux System Appliances with KIWI Next Generation (KIWI NG <VERSION>)
* suggested in  for more clarity

The change has been discussed with and approved by main author (@schaefi) upfront.

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -1,7 +1,7 @@
 .. kiwi documentation master file
 
 Building Linux System Appliances with {kiwi-product} |version|
-=================================================================
+==============================================================================
 
 .. note::
    This documentation covers {kiwi-product} |version|- the command line
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -1,7 +1,7 @@
 .. kiwi documentation master file
 
-Building Linux System Appliances with {kiwi-product} |version|)
-===============================================================
+Building Linux System Appliances with {kiwi-product} |version|
+=================================================================
 
 .. note::
    This documentation covers {kiwi-product} |version|- the command line
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -1,7 +1,7 @@
 .. kiwi documentation master file
 
-{kiwi} Documentation
-=====================
+Building Linux System Appliances with {kiwi-product} |version|)
+===============================================================
 
 .. note::
    This documentation covers {kiwi-product} |version|- the command line


schaefi (approved):
I like it, thanks :+1: 

Let's wait for the doc build tests to pass and once succeeded I'll merge

schaefi:
@taroth21 could you have a short look on the doc build error:

* https://github.com/OSInside/kiwi/runs/4558330078?check_suite_focus=true

You see it complains about the underline "===" to be too short for the new title. The reason is that you use {kiwi-product} which gets replaced to "KIWI Next Generation (KIWI NG)". For the result title the "===" underline is too short. In ReST I think this is mandatory.

Thanks

taroth21:
@schaefi : Thanks, should be fixed now (length of underlines looked OK in the Web editor yesterday but obviously it wasn't - but today I spotted another issue to fix anyway). ;)

schaefi:
> @schaefi : Thanks, should be fixed now (length of underlines looked OK in the Web editor yesterday but obviously it wasn't - but today I spotted another issue to fix anyway). ;)

Still too short :) I think you cannot prove this via the web-editor because the {kiwi-product} macro gets replaced at build time of the docs. You can always build the docs locally via ```tox -e doc``` which will also expose the error if the underline is too short after macro expansion.

For me the following worked:

```diff
diff --git a/doc/source/index.rst b/doc/source/index.rst
index 98ed763c4..056abd30d 100644
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -1,7 +1,7 @@
 .. kiwi documentation master file
 
-{kiwi} Documentation
-=====================
+Building Linux System Appliances with {kiwi-product} |version|
+==============================================================================
 
 .. note::
    This documentation covers {kiwi-product} |version|- the command line
```

taroth21:
> > @schaefi : Thanks, should be fixed now (length of underlines looked OK in the Web editor yesterday but obviously it wasn't - but today I spotted another issue to fix anyway). ;)
> 
> Still too short :) I think you cannot prove this via the web-editor because the {kiwi-product} macro gets replaced at build time of the docs. 

Ok, that is a bit of a problem then, because I was not aware that this is is only evaluated at build time. I did not check out the repo, so I cannot build the docs locally. Might I ask yout to fix the length of the underline then?  


> 
> For me the following worked:
> 
> ```diff
> diff --git a/doc/source/index.rst b/doc/source/index.rst
> index 98ed763c4..056abd30d 100644
> --- a/doc/source/index.rst
> +++ b/doc/source/index.rst
> @@ -1,7 +1,7 @@
>  .. kiwi documentation master file
>  
> -{kiwi} Documentation
> -=====================
> +Building Linux System Appliances with {kiwi-product} |version|
> +==============================================================================
>  
>  .. note::
>     This documentation covers {kiwi-product} |version|- the command line
> ```

schaefi:
>Ok, that is a bit of a problem then, because I was not aware that this is is only evaluated at build time. I did not check out the repo, so I cannot build the docs locally. Might I ask yout to fix the length of the underline then?

@taroth21 You are working from a fork and not from a branch, that's why I added the diff which provides you the fix. I think you can just take the diff and apply it in your fork.

If you want I can add you to the team which will allow you to create a pull request from a branch next time. Collaborating from branches is easier than from forks and I would be very happy to add you to the team list ;)

PR #2006, opened 2021-12-22
Pass custom boostrap packages to builtin_kiwi initrd

aaannz:

PR https://github.com/OSInside/kiwi/pull/1186 added option to specify extra packages on command line to be included in the bootstrap. This works well for dracut initrd. However when users use netboot or oemboot kiwi initrd, these specified packages are not added as those initrd systems use separate build root during create phase.

This PR adds this option also for these initrd systems.

### DIFF
--- a/kiwi/boot/image/base.py
+++ b/kiwi/boot/image/base.py
@@ -218,7 +218,7 @@ class BootImageBase:
             )
         )
 
-    def prepare(self) -> None:
+    def prepare(self, plus_packages=None) -> None:
         """
         Prepare new root system to create initrd from. Implementation
         is only needed if there is no other root system available
--- a/kiwi/boot/image/builtin_kiwi.py
+++ b/kiwi/boot/image/builtin_kiwi.py
@@ -68,7 +68,7 @@ class BootImageKiwi(BootImageBase):
         self.temp_directories: List[str] = []
         self.load_boot_xml_description()
 
-    def prepare(self) -> None:
+    def prepare(self, plus_packages=None) -> None:
         """
         Prepare new root system suitable to create a kiwi initrd from it
         """
@@ -94,7 +94,7 @@ class BootImageKiwi(BootImageBase):
                 signing_keys=self.signing_keys, target_arch=self.target_arch
             )
             system.install_bootstrap(
-                manager
+                manager, plus_packages
             )
             system.install_system(
                 manager
--- a/kiwi/boot/image/dracut.py
+++ b/kiwi/boot/image/dracut.py
@@ -151,7 +151,7 @@ class BootImageDracut(BootImageBase):
             with open(config_file, 'w') as config_handle:
                 config_handle.writelines(dracut_config)
 
-    def prepare(self) -> None:
+    def prepare(self, plus_packages=None) -> None:
         """
         Prepare dracut caller environment
 
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -129,6 +129,10 @@ class DiskBuilder:
         if custom_args and 'signing_keys' in custom_args:
             self.signing_keys = custom_args['signing_keys']
 
+        self.add_bootstrap_package = None
+        if custom_args and 'add_bootstrap_package' in custom_args:
+            self.add_bootstrap_package = custom_args['add_bootstrap_package']
+
         self.boot_image = BootImage.new(
             xml_state, target_dir, root_dir, signing_keys=self.signing_keys
         )
@@ -232,7 +236,7 @@ class DiskBuilder:
         # prepare initrd
         if self.boot_image.has_initrd_support():
             log.info('Preparing boot system')
-            self.boot_image.prepare()
+            self.boot_image.prepare(self.add_bootstrap_package)
 
         # precalculate needed disk size
         disksize_mbytes = self.disk_setup.get_disksize_mbytes()
--- a/kiwi/builder/kis.py
+++ b/kiwi/builder/kis.py
@@ -72,6 +72,9 @@ class KisBuilder:
         self.xz_options = custom_args['xz_options'] if custom_args \
             and 'xz_options' in custom_args else None
 
+        self.add_bootstrap_package = custom_args['add_bootstrap_package'] if custom_args \
+            and 'add_bootstrap_package' in custom_args else None
+
         self.boot_image_task = BootImage.new(
             xml_state, target_dir, root_dir,
             signing_keys=self.boot_signing_keys
@@ -134,7 +137,7 @@ class KisBuilder:
         # prepare initrd
         if self.boot_image_task.has_initrd_support():
             log.info('Creating boot image')
-            self.boot_image_task.prepare()
+            self.boot_image_task.prepare(self.add_bootstrap_package)
 
         # export modprobe configuration to boot image
         self.system_setup.export_modprobe_setup(
--- a/kiwi/tasks/system_build.py
+++ b/kiwi/tasks/system_build.py
@@ -286,7 +286,8 @@ class SystemBuildTask(CliTask):
                 'signing_keys': self.command_args[
                     '--signing-key'
                 ] + self.xml_state.get_repositories_signing_keys(),
-                'xz_options': self.runtime_config.get_xz_options()
+                'xz_options': self.runtime_config.get_xz_options(),
+                'add_bootstrap_package': self.command_args['--add-bootstrap-package']
             }
         )
         result = image_builder.create()
--- a/test/unit/boot/image/builtin_kiwi_test.py
+++ b/test/unit/boot/image/builtin_kiwi_test.py
@@ -67,7 +67,7 @@ class TestBootImageKiwi:
             signing_keys=None, target_arch=None
         )
         self.system_prepare.install_bootstrap.assert_called_once_with(
-            self.manager
+            self.manager, None
         )
         self.system_prepare.install_system.assert_called_once_with(
             self.manager
--- a/test/unit/builder/disk_test.py
+++ b/test/unit/builder/disk_test.py
@@ -350,7 +350,9 @@ class TestDiskBuilder:
             '/dev/boot-device'
         )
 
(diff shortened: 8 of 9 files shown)


aaannz:
Ah, I forgot I was doing these changes against 9.23.20. I will redo them against master.

schaefi:
@aaannz sorry for the late response. I found the time to look at your pull requests. Thanks for your work :+1: From a coding perspective this looks good. However, I was thinking if it is really needed and I guess it depends on your use case. The way you add additional packages to a custom kiwi boot image is via the commandline. You changed the scope of the ```--add-bootstrap-package``` option to also add the specified packages to the custom kiwi initrd package.

This brings up the question if this coupling belongs together ? From my perspective explicitly adding a package to be installed as part of the system image bootstrap does not automatically imply that this package will also be installed to a custom boot image (initrd). If we compare this with dracut, dracut works from the system image root and includes components by some default settings or by dependencies to other modules. Components that does not match but should be part of the dracut initrd must also be added explicitly by some setting in a dracut.conf.d/... file.

From this perspective I thought it might be rather unexpected that the specified bootstrap-packages have an effect on the used boot image description. What do you think ?

Furthermore in kiwi we support adding packages to the boot image via the ```bootinclude=true|false"``` attribute.
Thus, if setting this information on the commandline is not mandatory for your use case you can also specify this as part of the kiwi XML description e.g

```xml
<package name="some-package" bootinclude="true"/>
```

This tells kiwi to install the package to the system image and also to the boot image. From my perspective this makes it explicit and a nice read such that people understand what is going to happen

If we need a commandline option I suggest to add a new option which explicitly says that this package is installed as part of the custom kiwi boot image creation. e.g ```--add-kiwi-initrd-bootstrap-package``` or similar.

The feature of kiwi to build its own initrd through a kiwi description for this initrd is only rarely used these days.
The reason is that dracut and the way to add modules and custom code to a dracut initrd is considered the standard
at least on Linux distributions which adopted to dracut. Since I didn't saw a reason to drop the kiwi initrd feature because
it does not hurt :) it still exists today. But we do not promote it very much as most people use dracut.

Therefore the question is also, does your use case really require the use of a custom kiwi initrd, or would it maybe possible to make use of dracut as well ? And if you need the custom initrd built by kiwi, would the already exsting ```bootinclude``` feature also work for you ?

Thanks much for your feedback

aaannz:
> This brings up the question if this coupling belongs together ? From my perspective explicitly adding a package to be installed as part of the system image bootstrap does not automatically imply that this package will also be installed to a custom boot image (initrd). If we compare this with dracut, dracut works from the system image root and includes components by some default settings or by dependencies to other modules. Components that does not match but should be part of the dracut initrd must also be added explicitly by some setting in a dracut.conf.d/... file.
> 
> From this perspective I thought it might be rather unexpected that the specified bootstrap-packages have an effect on the used boot image description. What do you think ?

The problem here is different. This comes from users of SUSE Manager that for some reason do not use dracut initrd yet.
The problem I am trying to solve by this is that when kiwi prepares kiwi initrd root it omits package with SUSE Manager CA certificate which is required to access repositories from SUSE Manager. We pass this package name on command line and it works perfectly for system image. But because kiwi initrd uses different root zypper will fail during build on untrusted connection.

schaefi:
Hmm, ok thanks for the explanation. I understand that the proposed solution fixes your use case but I struggle a bit to add this as generic change to the upstream code. I see the following issues:

* It's not obvious to me that ```--add-bootstrap-package``` also adds these package to a custom boot image if used
* It's an implicit behavior change which imho smells a bit
* It's inconsistent compared to dracut vs. custom kiwi initrd. This one is not so troublesome as there are already inconsistencies because the tooling is different on its own. However, I would prefer to prevent adding further inconsistencies in the process of creating an initrd. For dracut all this is a noop
* It changes the public interface for BootImage::prepare with an optional ```plus_packages``` argument. It could be considered confusing because it's only effective for custom kiwi built boot images which is a feature that will gain less and lesser attraction in the future.

I fully understand that it makes sense to fix your issue but at the moment I'm not convinced to make this an upstream change.

One question. In the process of calling kiwi as part of SUMA you need to construct the commandline which among others seems to contain some logic to add the SUMA CA package. Would it be too intrusive to add the definition

```xml
<package name="suma-ca-package" bootinclude="true"/>
```

into the provided image description instead of passing this as a commandline argument ? I think anyone who wants to build the image using SUMA repos would require this change, which means it seems mandatory and not an option on the commandline

As I said earlier I think a clean fix that can go upstream would be to expose the existing ```bootinclude``` attribute
also as a commandline option ```--bootinclude```. That way the logic to incorporate packages into a custom kiwi boot image stays the same, no inconsistency will be introduced and also the documentation stays consistent with how packages not part of the custom kiwi boot image can be added at build time.

There is the method in kiwi/xml_state.py

```python
def copy_bootincluded_packages(self, target_state: Any) -> None:
```

which I think can be extended to

```python
def copy_bootincluded_packages(self, target_state: Any, plus_packages: List[str] = []) -> None:
```

In combination with a new option

```--botinclude ...```

you can just pass the list of extra packages from the option to this method

What do you think ?

Thanks for your understanding

**Comment by schaefi** [comment]
your kiwi commandline would then use this new option like this

```
kiwi-ng ... --add-bootstrap-package XXX --bootinclude XXX
```

It's a bit more repetition but it's imho more obvious and explicit

**Comment by aaannz** [comment]
Seems to me we are talking about different things (for the record, this package is in bootstrap section of the kiwi xml, see at the end)

I do not need this package to be in initrd itself. I need this package to be in the root from which initrd is generated.
IIUC kiwi does these steps for kiwi initrd:

1) call prepare phase - this is running on build host machine but using specified repositories
2) call create phase - this is running in chroot build it prepare stage and still can install packages

This `create phase` then fails because zypper cannot download packages from SUMA because of missing trusted CA.

To answer your points:

> It's not obvious to me that --add-bootstrap-package also adds these package to a custom boot image if used

It should not add it to boot image, just to boot image temporary root as part of the bootstrap from which initrd is then generated.

> It's an implicit behavior change which imho smells a bit

I don't understand, what do you mean?

> It's inconsistent compared to dracut vs. custom kiwi initrd. This one is not so troublesome as there are already inconsistencies because the tooling is different on its own. However, I would prefer to prevent adding further inconsistencies in the process of creating an initrd. For dracut all this is a noop

True that this is noop for dracut as dracut does not use separate chroot to build initrd image. But dracut uses image chroot for which we already have this. So IMO this is making it consistent.

> It changes the public interface for BootImage::prepare with an optional plus_packages argument. It could be considered confusing because it's only effective for custom kiwi built boot images which is a feature that will gain less and lesser attraction in the future.

True, I was not aware that this is part of public API.

For record bootstrap part of the kiwi xml:
```xml
<packages type="bootstrap">
  <package name="udev"/>
  <package name="filesystem"/>
  <package name="sles-release"/>
  <package name="rhn-org-trusted-ssl-cert-osimage" bootinclude="true"/>
</packages>
```

schaefi:
>Seems to me we are talking about different things (for the record, this package is in bootstrap section of the kiwi xml, see at the end)

Ok, thanks I was not aware of that.

>I do not need this package to be in initrd itself. I need this package to be in the root from which initrd is generated.

Yes I understood the reasons. We are on the same page here

> IIUC kiwi does these steps for kiwi initrd:
>
>call prepare phase - this is running on build host machine but using specified repositories
>call create phase - this is running in chroot build it prepare stage and still can install packages
>This create phase then fails because zypper cannot download packages from SUMA because of missing trusted CA.

Yes you are absolutely correct about the process details. From what you have said and with the information
that the image description contains ```<package name="rhn-org-trusted-ssl-cert-osimage" bootinclude="true"/>```
I think I understood why it did not work for you.

If we look into ```kiwi/xml_state.py``` and there into the ```copy_bootincluded_packages``` you see the
following code

```python
target_packages_sections = \
    target_state.get_image_packages_sections()
    if not target_packages_sections:
        # no packages type=image section was found, add to bootstrap
        target_packages_sections = \
            target_state.get_bootstrap_packages_sections()
```

This means if the boot image description used contains a ```<packages type="image"/>``` section the
```rhn-org-trusted-ssl-cert-osimage``` is added to this section and not to the type="bootstrap" section.
In your case this is too late because you need the certs to be installed in bootstrap such that the
chroot based create step can work

I think if we change the logic to bootinclude packages to the bootstrap section this will fix it

In the image description and in the ```<type>``` section there is an attribute set: ```boot="..."```
this references the path to the used boot image description. To prove the theory could you check the
packages settings from that boot image description ? I assume it defines both a type="image" and
a type="bootstrap" section. If so the above issue would apply

I think changing the kiwi code to add bootincludes to bootstrap would be a good change

What do you think ?

schaefi:
>It's an implicit behavior change which imho smells a bit
>>I don't understand, what do you mean?

I meant we would be changing the behavior of ```--add-bootstrap-package``` in an implicit way. right now it acts only on the system image description. After the change it would act on system image and boot image descriptions.

schaefi:
I tried to code what I think here #2018 

@aaannz Would you mind to double check if the proposed changes makes sense ?

Thanks much

aaannz:
I will check it out and report back. By the way, this commit looks like revert of https://github.com/OSInside/kiwi/commit/9500232ba1e81523c9bcc763eabf8c6025407609 Will this not break some other behavior?

schaefi:
Thanks for the pointer, I will check if this opens another issue

schaefi:
@aaannz the commit in https://github.com/OSInside/kiwi/commit/9500232ba1e81523c9bcc763eabf8c6025407609 is rather old from 2017 and unfortunately I did not give a reason why this change was done at that time (shame on me). In 2017 some images still used custom kiwi boot images. From today's perspective dracut is the successor of all this and the custom kiwi boot image feature is still there but only rarely used.

Out of my mind I could not think of a regression when we put the package in the bootstrap section rather than in the image section. The host package manager should be able to install the package like it has to be able to install all other bootstrap packages.

So I think I would be willing to take the risk ;)

aaannz:
Your PR worker like a charm. So I will close this one and let's go ahead with yours.

PR #2007, opened 2021-12-23
Switch to stdlib json package

dirkmueller:
simplejson is a backport of the stdlib json package for older
python versions. It went into stdlib with 3.4, and it is being developed
there since then. the comment was that simplejson supports tuples, but
I see no difference with python 3.6+:

>>> myvar=(1,2,3)
>>> import json
>>> json.dumps(myvar)
'[1, 2, 3]'
>>> import simplejson
>>> simplejson.dumps(myvar)
'[1, 2, 3]'

On the other hand, json is simply faster than simplejson, and it is one
less dependency to worry about. here's a comparison (x86_64 AMD Renoir,
1 million invocations)

      json dumps 0.918306345003657 seconds
      simplejson dumps 1.505740466993302 seconds
      json loads 0.793854812043719 seconds
      simplejson loads 0.879414324008394 seconds

Fixes # .

*
*

### DIFF
--- a/.virtualenv.dev-requirements.txt
+++ b/.virtualenv.dev-requirements.txt
@@ -31,7 +31,6 @@ types-pkg_resources
 types-requests
 types-PyYAML
 types-mock
-types-simplejson
 
 # Generate Data Structures from XML Schema
 # http://pythonhosted.org/generateDS
--- a/.virtualenv.requirements.txt
+++ b/.virtualenv.requirements.txt
@@ -19,6 +19,3 @@ requests
 
 # markup support
 anymarkup
-
-# json support that also knows about tuples
-simplejson
--- a/kiwi/system/result.py
+++ b/kiwi/system/result.py
@@ -16,7 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import logging
-import simplejson
+import json
 import pickle
 import os
 from typing import (
@@ -152,7 +152,7 @@ class Result:
             if portable:
                 with open(filename + '.json', 'w') as result_portable:
                     result_portable.write(
-                        simplejson.dumps(
+                        json.dumps(
                             self.result_files, sort_keys=True, indent=4
                         ) + os.linesep
                     )
--- a/package/python-kiwi-pkgbuild-template
+++ b/package/python-kiwi-pkgbuild-template
@@ -21,7 +21,7 @@ build() {
 }
 
 package_python-kiwi(){
-  depends=(python-docopt python-simplejson python-future python-lxml python-requests python-setuptools python-six python-pyxattr python-yaml grub qemu squashfs-tools gptfdisk pacman e2fsprogs xfsprogs btrfs-progs libisoburn lvm2 mtools parted multipath-tools rsync tar shadow screen kiwi-man-pages)
+  depends=(python-docopt python-future python-lxml python-requests python-setuptools python-six python-pyxattr python-yaml grub qemu squashfs-tools gptfdisk pacman e2fsprogs xfsprogs btrfs-progs libisoburn lvm2 mtools parted multipath-tools rsync tar shadow screen kiwi-man-pages)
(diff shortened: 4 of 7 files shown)


ggardet:
This should fix errors seen at: https://build.opensuse.org/package/live_build_log/openSUSE:Leap:15.4:ARM:Images/opensuse-leap-dnf-image:lxc-microdnf/images/armv7l

```
[   83s] Traceback (most recent call last):
[   83s]   File "/usr/bin/kiwi", line 6, in <module>
[   83s]     from pkg_resources import load_entry_point
[   83s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3123, in <module>
[   83s]     @_call_aside
[   83s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3107, in _call_aside
[   83s]     f(*args, **kwargs)
[   83s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 3136, in _initialize_master_working_set
[   83s]     working_set = WorkingSet._build_master()
[   83s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 578, in _build_master
[   83s]     ws.require(__requires__)
[   83s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 895, in require
[   83s]     needed = self.resolve(parse_requirements(requirements))
[   83s]   File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 781, in resolve
[   83s]     raise DistributionNotFound(req, requirers)
[   83s] pkg_resources.DistributionNotFound: The 'simplejson' distribution was not found and is required by kiwi
```

schaefi:
I don't think the switch to the stdlib json can be done. The reasons are explained here:

* https://github.com/OSInside/kiwi/pull/1949

At the end of the conversation you can see an example where I found a significant difference between the two

Thanks

dirkmueller:
@schaefi ah I see, so it is about namedtuple not being marshalled correctly. is that frequently used in kiwi? whats the advantage of namedtuple over dict, which should marshall properly?

what actually is using the json output? I could only see it exported, but no related import? what is the full workflow that I can use for testing?

I can see no option to marshall namedtuple "properly" with the std json lib, so if there is no way around that then this PR should be closed. :/

schaefi:
>ah I see, so it is about namedtuple not being marshalled correctly. is that frequently used in kiwi?

kiwi is integrated into Koji and they requested a portable version of the result information. Until then kiwi provided
a binary object dump (pickle.dump) of the Result class. The implementation of the Result class uses NamedTuples.
If we would have made the decision to move from tuples to dict we would have changed the public interface of the
Result class and since the object dump also expose the binary representation of it, we would have created a major
incompatibility. The footprint to introduce a json module which can handle tuples was smaller than the other
solution.

I agree with you that exposing the binary representation of a class as result information was a bad idea, but I can't
turn back the wheel, unfortunately. With the next major release (kiwi10) we will correct some of the mistakes of the
past and also document them.

It seems the dependency to simplejson is creating an issue for you ? Could you explain what the problem is ? maybe we find another solution ?

Thanks and sorry

schaefi:
> It seems the dependency to simplejson is creating an issue for you ? 

@dirkmueller with regards to the last comments, is there anything we can help to solve this issue ?

dirkmueller:
@schaefi there was a problem with simplejson module but it seems fixed now, so there is no remaining issue that I'm aware of. I'm still a bit confused on why it would be an abi breakage with koji because kiwi merely does an export, and it shouldn't be an issue to hack up two or three lines of python code to convert nametuples into dicts before running json export. 

I wanted to take a look at that and refine my PR but I didn't get around to. Its fine to close this until I find the time.

Conan-Kudo:
@dirkmueller The problem is that the Pickle data structure is not stable across Python versions. The Koji hub could be running on Python 2.6 on RHEL 6 (!!!) while the build environment that kiwi ran used Python 3.11 on Fedora Linux 38. It's just too risky.

PR #2011, opened 2022-01-02
Add util-linux dep for -systemdeps-disk-images subpackage

Conan-Kudo:
Without this dependency, kiwi fails to work properly in minimal image
build environments, like in a mock chroot where util-linux is not installed.


schaefi (approved):
Looks good :+1:

PR #2012, opened 2022-01-02
Allow firmware="custom" setting

schaefi:
The firmware attribute in kiwi is used to indicate for
which boot firmware the image should be build. Specifying
the target firmware is helpful to create for example the
correct disk layout. If no firmware is specified KIWI
decides for a default according to the image architecture.
This selection is not 100% accurate and as we don't know
the later target system. Especially for embedded devices
the correct disk layout and other settings can be
board specific and KIWI's default settings regarding the
firmware could be invalid. For compatibility reasons we
cannot switch off the default selection case and therefore
a new attribute value "custom" is introduced with this
commit. If set KIWI does not select any firmware and
consequently all settings caused by a firmware setup will
be skipped. On the other hand this means all needed
settings for the target to boot and not done by KIWI
needs to be specified explicitly and as needed.

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -1195,6 +1195,7 @@ class Defaults:
             'ppc64': 'ofw',
             'ppc64le': 'ofw',
             'arm64': 'efi',
+            'aarch64': 'efi',
             'armv5el': 'efi',
             'armv5tel': 'efi',
             'armv6hl': 'efi',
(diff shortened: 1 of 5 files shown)


schaefi:
@MofX let me know if this meets what we have discussed. Thanks much

MofX:
Looks good, I have to check if everything is working with these changes next week

MofX:
> Looks good, I have to check if everything is working with these changes next week

Just checked with our use case and this is working as expected.

davidcassany (approved):

PR #2021, opened 2022-01-09
Fixed regression in compression detection

### LINKED ISSUE #2015: kiwi-dump-image writes compressed disk image during PXE installation
## Problem description
When installing images using the PXE method, everything works as expected until the image itself (_foobar.x86_64-20220106.xz_) is dumped to disk. Due to changes in the [kiwi-net-lib.sh](https://github.com/OSInside/kiwi/commit/282529de8f612dee32d54ee868c2365dcd829220#diff-737d484eaf4d87a23c3e04b3b3aeece6b7bf3d71f79b9589c3f1a66457a2b344), the `_is_compressed` function can't return the correct state of any external file.

[`fetch_file`](https://github.com/OSInside/kiwi/blob/282529de8f612dee32d54ee868c2365dcd829220/dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh#L30) tries to detect the file type / compression but fails, because the file is not stored locally:
```bash
# xz -l http://example.com/foobar.x86_64-20220106.xz; echo $?
xz: http://example.com/foobar.x86_64-20220106.xz: No such file or directory
1
```

As we did not alter our PXE installation process and this issue came up after updating kiwi, I'm curious why this worked out in the first place:

> Just for the records I successfully pxe deployed an OEM image (using `installpxe`). 👍🏽
_Originally posted by @davidcassany in https://github.com/OSInside/kiwi/issues/1987#issuecomment-977749776_

## Expected behaviour
Image stream is decompressed first before dumped to disk.

- create PXE image using the `installpxe="true"` and `<oem-unattended>true</oem-unattended>` option
- try to install the PXE image using the kernel parameter for kiwi: `rd.kiwi.install.pxe rd.kiwi.install.image=http://example.com/foobar.x86_64-20220106.xz`

+ KIWI version: 9.24.14-1.1
+ Operating system: Leap 15.2
+ OBS version: 2.10.11-lp152.78.1

schaefi:
The change from  282529de8f612dee32d54ee868c2365dcd829220
Introduced a bad regression. The assumption was made that the
xz tool could be used to detect if a file is compressed or not.
However, this requires the file to be locally present. In the
scope of the method call is_compressed() and within a remote
deployment e.g PXE this is not the case. Therefore the former
way to "detect" the compression according to the .xz postfix
of the source filename was restored. In addition the function
name was changed to is_xz_compressed() because that's what the
method can do and not more. This Fixes #2015

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-net-lib.sh
@@ -27,7 +27,7 @@ function fetch_file {
     if _is_dolly "${source_url}";then
         fetch="dolly"
     fi
-    if _is_compressed "${source_url}";then
+    if _is_xz_compressed "${source_url}";then
         fetch="${fetch} | xz -d"
     fi
     if [ -z "${source_uncompressed_bytes}" ];then
@@ -78,9 +78,9 @@ function uri_fragment {
 #======================================
 # Methods considered private
 #--------------------------------------
-function _is_compressed {
+function _is_xz_compressed {
     local source_url=$1
-    xz -l "${source_url}" &>/dev/null
+    [[ ${source_url} =~ .xz$ ]]
 }
 
 function _is_dolly {


schaefi:
@davidcassany It would be great if you can take a moment and review this change.

I'm aware that this also changes the function ```is_compressed``` to ```is_xz_compressed```. I did that on purpose because the method can only check for xz compression and with this change even this is not safe as it does it only by the name of the file which is not a safe check.

I looked through the code and there is no other use of this method, so I thought it's ok to adapt the name. By convention methods starting with "_" was set as private. This concept does not really exist in shell coding but by convention there should be no other dracut code using this method.

It would be great if you can have a look and check if you see other issues caused by the change. Thanks much

davidcassany:
Sure, it is on my TODO list. Gonna have look, I should have noticed this regression... I was testing this... 😔

qby-wenzel:
LGTM... That's kind of what our quick hack/workaround script in our dev env does - patching the detection line in the initrd back to its origin.

davidcassany (approved):
Looks good to me

Conan-Kudo (approved):

PR #2027, opened 2022-01-13
Fix booting GRUB submenu entries with hybrid images (linux/linuxefi)

Vogtinator:
Variables assigned with "set" are not visible in submenus for some reason.
Export $linux and $initrd, so that they also work in submenu entries.

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1192523

### DIFF
--- a/kiwi/bootloader/template/grub2.py
+++ b/kiwi/bootloader/template/grub2.py
@@ -52,12 +52,13 @@ class BootLoaderTemplateGrub2:
         self.header_hybrid = dedent('''
             set linux=linux
             set initrd=initrd
-            if [ "$${grub_cpu}" = "x86_64" -o "$${grub_cpu}" = "i386" ];then
+            if [ "$${grub_cpu}" = "x86_64" -o "$${grub_cpu}" = "i386" ]; then
                 if [ "$${grub_platform}" = "efi" ]; then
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
Looks good, thanks for the bugfix

davidcassany:
@Vogtinator just one small detail. I am not sure the commit message will be properly parsed once in changes file. To ensure bugzilla traces bugfixes I use `bsc#1192523` notation in commit messages instead of URLs. Does the bugzilla parser in OBS track full URLs?

schaefi:
@Vogtinator looks good, thanks for the fix.

I second the question David has. I've never used the full URL to reference the bugzilla entry. Since we are creating the .changes file from the git history this might be an issue for the suse bug tracker. I'll keep the PR open until we have clarity on this part and set a label such that it does not get accidentally merged

Vogtinator:
> @Vogtinator just one small detail. I am not sure the commit message will be properly parsed once in changes file. To ensure bugzilla traces bugfixes I use `bsc#1192523` notation in commit messages instead of URLs. Does the bugzilla parser in OBS track full URLs?

No, I'll change it. I specifically used the long URL because I couldn't find any recent use of `boo#` in the git log. Looks like you're using `bsc#` only.

PR #2028, opened 2022-01-13
Allow "uninitialized" in /etc/machine-id instead of emptying it

Vogtinator:
According to machine-id(5), an empty file does not signal that this is the
first boot of the system. Write "uninitialized" into the file instead, as
documented in the man page.

Improve the documentation of the method, e.g. dracut is not involved.

Fixes https://bugzilla.suse.com/show_bug.cgi?id=1194484.

I'm not sure whether the `if os.path.exists(machine_id):` should be deleted as well.
On systems without systemd, the file shouldn't be created, but on systems with systemd, it's basically always wrong to not have it. It should either be empty (to not trigger `ConditionFirstBoot`) or contain "uninitialized" (to trigger `ConditionFirstBoot`).

### DIFF
--- a/doc/source/concept_and_workflow/shell_scripts.rst
+++ b/doc/source/concept_and_workflow/shell_scripts.rst
@@ -281,68 +281,68 @@ Configuration Tips
 
 #. **Locale configuration:**
 
-  KIWI in order to set the locale relies on :command:`systemd-firstboot`,
-  which in turn writes the locale configuration file :file:`/etc/locale.conf`.
-  The values for the locale settings are taken from the description XML
-  file in the `<locale>` element under `<preferences>`.
+   KIWI in order to set the locale relies on :command:`systemd-firstboot`,
+   which in turn writes the locale configuration file :file:`/etc/locale.conf`.
+   The values for the locale settings are taken from the description XML
+   file in the `<locale>` element under `<preferences>`.
 
-  KIWI assumes systemd adoption to handle these locale settings, in case the
-  build distribution does not honor `/etc/locale.conf` this is likely to not
-  produce any effect on the locale settings. As an example, in SLE12
-  distribution the locale configuration is already possible by using the
-  systemd toolchain, however this approach overlaps with SUSE specific
-  managers such as YaST. In that case using :command:`systemd-firstboot`
-  is only effective if locales in :file:`/etc/sysconfig/language` are
-  not set or if the file does not exist at all. In SLE12
-  :file:`/etc/sysconfig/language` has precendence over
-  :file:`/etc/locale.conf` for compatibility reasons and management tools
-  could still relay on `sysconfig` files for locale settings.
+   KIWI assumes systemd adoption to handle these locale settings, in case the
+   build distribution does not honor `/etc/locale.conf` this is likely to not
+   produce any effect on the locale settings. As an example, in SLE12
+   distribution the locale configuration is already possible by using the
+   systemd toolchain, however this approach overlaps with SUSE specific
+   managers such as YaST. In that case using :command:`systemd-firstboot`
+   is only effective if locales in :file:`/etc/sysconfig/language` are
+   not set or if the file does not exist at all. In SLE12
+   :file:`/etc/sysconfig/language` has precendence over
+   :file:`/etc/locale.conf` for compatibility reasons and management tools
+   could still relay on `sysconfig` files for locale settings.
 
-  In any case the configuration is still possible in KIWI by using
-  any distribution specific way to configure the locale setting inside the
-  :file:`config.sh` script or by adding any additional configuration file
-  as part of the overlay root-tree.
+   In any case the configuration is still possible in KIWI by using
+   any distribution specific way to configure the locale setting inside the
+   :file:`config.sh` script or by adding any additional configuration file
+   as part of the overlay root-tree.
 
 #. **Stateless systemd UUIDs:**
 
-  Machine ID files are created and set (:file:`/etc/machine-id`,
-  :file:`/var/lib/dbus/machine-id`) during the image package installation
-  when *systemd* and/or *dbus* are installed. Those UUIDs are intended to
-  be unique and set only once in each deployment. {kiwi} follows the `systemd
-  recommendations
-  <https://www.freedesktop.org/software/systemd/man/machine-id.html>`_ and
-  wipes any :file:`/etc/machine-id` content, leaving it as an empty file.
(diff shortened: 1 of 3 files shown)


schaefi:
This is interesting. Are you aware of a behavior change at this point ? The reason why I'm asking is because I remember we intentionally created the file as empty file to indicate that machine-id should be re-created. 

* https://www.freedesktop.org/software/systemd/man/machine-id.html

```
For operating system images which are created once and used on multiple machines, for example for containers or in the cloud, /etc/machine-id should be either missing or an empty file in the generic file system image
```

Is this information wrong ? 

Thanks

davidcassany:
Indeed we should be careful with this change as I remember in the past we started triggering the systemd-firstboot.service because of changes in that area causing an interactive firstboot.

I just did a quick test with this branch and the ConditionFirstBoot was not triggered, hence systemd-firstboot.service not started which is the expected behavior for the example appliance from https://github.com/OSInside/kiwi-descriptions/tree/master/suse/x86_64/suse-leap-15.3

I loop mounted the build image and I could verify machine-id was set to `uninitalized`, after first boot was recreated with some generated ID.

This behavior does not seam to match with:
> On systems without systemd, the file shouldn't be created, but on systems with systemd, it's basically always wrong to not have it. It should either be empty (to not trigger ConditionFirstBoot) or contain "uninitialized" (to trigger ConditionFirstBoot)

Could it be that this behavior is only for TW at the moment?

Conan-Kudo:
Does this behavior change exist in Fedora Rawhide too? I wonder now, because we truncate the file to nothing in Fedora images to trigger firstboot behavior.

Vogtinator:
Initially I thought that an empty `/etc/machine-id` was basically equivalent to having no `/etc/machine-id` at all, just that booting without the `rw` parameter works. (see also https://github.com/OSInside/kiwi/issues/1650#issuecomment-945665656). When I submitted that change to JeOS in SLE, @gmoro told me that it's in fact not equivalent.

It looks like the `uninitialized` value is indeed new in v247 and mostly implemented for a more robust firstboot detection: https://github.com/systemd/systemd/pull/16939. I'd even say that it's an implementation detail, but it's documented too well for that to be the case.

So always writing `uninitialized` is not an option, as it would break with older systemd versions. There might also be a case where a kiwi built image should intentionally have an empty `machine-id`, to not trigger firstboot behaviour. I can't come up with a valid scenario myself though.

Question is how to achieve that booting without the `rw` option works (i.e. `/etc/machine-id` needs to exist, so that systemd can use a mount), but does also trigger firstboot behaviour? Except for writing `uninitialized` I don't see one. So we might have to add `rw` as kernel option (back), and fix https://github.com/OSInside/kiwi/pull/1939 differently.

> Does this behavior change exist in Fedora Rawhide too? I wonder now, because we truncate the file to nothing in Fedora images to trigger firstboot behavior.

That an empty `/etc/machine-id` does not trigger `ConditionFirstBoot` does not seem new behaviour.
FWICT that was added in 2014: https://github.com/systemd/systemd/commit/baa1bdf70f21848fbe01d3f383ae0f59d86a9bf3

Conan-Kudo:
Good grief. That means we're doing it wrong basically everywhere... 🤦🏾‍♂️

davidcassany:
> Good grief. That means we're doing it wrong basically everywhere...

I don't think so, @Vogtinator findings are in line with what we experienced so far. We intentionally wrote an empty `/etc/machine-id` to prevent triggering ConditionFirstBoot. So we are good in that regard. To me it looks like what we are finding now is that some further control might be needed to fulfill some use cases.

What about making an empty /etc/machine-id only if this is not set to `uninitialized`. IIRC the `/etc/machine-id` gets generated at systemd install time to a random UUID. Then in `config.sh` we could overwrite it manually  to `uninitialized` with a simple `echo "uninitialized" > /etc/machine-id`. Then the prepare step of the initrd would wipe `machine-id` content only if this is not set to `uninitialized`.

Could an strategy like this provide enough flexibility to allow ConditionFirstBoot to be triggered at users wish and also be backward compatible?

It is a bit convoluted, but systemd behavior in that regard is pretty obscure too, I believe it might already the third or fourth time in few years we discuss about the default value `/etc/machine-id` should have.

Vogtinator:
>Good grief. That means we're doing it wrong basically everywhere...

By "we", do you mean kiwi or specific images? Do you expect `ConditionFirstBoot` to trigger or not?

> I don't think so, @Vogtinator findings are in line with what we experienced so far. We intentionally wrote an empty /etc/machine-id to prevent triggering ConditionFirstBoot.

Question is whether that is correct - why not trigger `ConditionFirstBoot`?

> What about making an empty /etc/machine-id only if this is not set to uninitialized. IIRC the /etc/machine-id gets generated at systemd install time to a random UUID.

Currently the `systemd` package (in openSUSE) creates an empty `/etc/machine-id` on initial installation, the UUID is written on the next boot.

>  Then in config.sh we could overwrite it manually to uninitialized with a simple echo "uninitialized" > /etc/machine-id. Then the prepare step of the initrd would wipe machine-id content only if this is not set to uninitialized.
>
> Could an strategy like this provide enough flexibility to allow ConditionFirstBoot to be triggered at users wish and also be backward compatible?

Yeah, that could work. It would move the responsibility for the systemd version compatibility to the image.

Conan-Kudo:
> By "we", do you mean kiwi or specific images? Do you expect `ConditionFirstBoot` to trigger or not?
> 

Specific images. The guidance around ensuring firstboot gets triggered is wrong...

schaefi:
From what I understood the current solution to deal with all possible systemd behavior is to write the initialization content for ```/etc/machine-id``` as part of a custom ```config.sh``` script ?

Is that correct ?

If so I'm fine with this solution. This means it's the user responsibility to do the "right" thing if the empty
file is not the correct way to do it ? It's maybe not so user friendly though because I think not many people
know about this behavior.

We have documented this setup in kiwi here:

* https://osinside.github.io/kiwi/concept_and_workflow/shell_scripts.html

> 1. Stateless systemd UUIDs:

I suggest we add some information to this section to point out the changed behavior and the required config.sh bits

Thoughts ?

schaefi:
@Vogtinator looks good to me. From my perspective nothing holds us back to turn this from draft to review

Vogtinator:
> @Vogtinator looks good to me. From my perspective nothing holds us back to turn this from draft to review

Done. I just finished testing the three different possibilities in `config.sh` (not existing, invalid content and `uninitialized`).

schaefi (approved):
This is nice and thanks for the doc update :+1:

Issue #1750, opened 2021-03-04
Live isos contain duplicate/unnecessary files


Vogtinator:
Currently I'm trying to optimize the size of the Tumbleweed Rescue-CD as it went above the 650MiB limit again.

I noticed that there are redundant/unneeded files in the .iso's filesystem:

`/boot/grub2/x86_64-efi` is not needed because `grub.efi` contains everything necessary already and with EFI secure boot can't load modules anyway.

`/boot/x86_64/loader` contains gfxboot, even though it's not used outside of 32bit x86

`./boot/x86_64/efi` is a 15MiB file which contains a copy of `/EFI/BOOT`. kiwi calls `xorriso` with `-append_partition 2 0xef /usr/src/packages/KIWI-iso/live-media.u4fmfxae/boot/x86_64/efi`, which IIUC copies that file as partition into the .iso image, so it can be deleted afterwards.

Vogtinator:
The TW rescue CD ran over its max size again several times, if this gets fixed there would be some more headroom for the future.

PR #2029, opened 2022-01-13
Don't copy GRUB modules for EFI with secure boot enabled

Vogtinator:
When booting grub.efi with secure boot enabled, modules can't be loaded
and thus the grub.efi image needs to be complete. Save some space in live
images by not copying them into the ISO filesystem.

Fixes part of #1750

I built livecd-tumbleweed-x11 and was able to boot it on x86_64 with and without EFI just like before, but it's now 6MiB (>1%!) smaller.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -435,7 +435,6 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             self._setup_efi_image(mbrid=mbrid, lookup_path=lookup_path)
             self._copy_efi_modules_to_boot_directory(lookup_path)
         elif self.firmware.efi_mode() == 'uefi':
-            self._copy_efi_modules_to_boot_directory(lookup_path)
             self._setup_secure_boot_efi_image(
                 lookup_path=lookup_path, mbrid=mbrid
             )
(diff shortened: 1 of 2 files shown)


Conan-Kudo (approved):

Conan-Kudo:
One of these days, we need to consolidate `efi` and `uefi` into a single option with an attribute...

schaefi:
Hmm, @Conan-Kudo you merged this and I'm not sure if it will cause some unwanted side effects. I think this change might break installation ISO's and maybe secure boot on non suse distros. We should have an eye on that. As it is merged now I will place an updated kiwi in Staging today evening. This will rebuild all integration tests and it would be good to test them

my gut feeling tells me this is causing trouble because that copy action was done for a reason... let's hope it is no longer required

Vogtinator:
> I think this change might break installation ISO's and maybe secure boot on non suse distros

GRUB with secure boot can't load modules, so unless those modules were somehow used by something other than GRUB, I don't expect any breakage.

Conan-Kudo:
Since the UEFI mode pulls the secure boot stuff, there's no way this could break stuff more than it already would have, since as @Vogtinator, they can't even be used.

PR #2030, opened 2022-01-13
Don't include the ESP in the ISO9660 partition as well (#1750)

Vogtinator:
I'm not very familiar with kiwi code, so I'm not sure what the best way to implement this is.
Currently kiwi puts everything generated for the final `xorriso` call into the media directory, which is written to the ISO9660 filesystem. Some arguments are written into the image though, which means that those just waste space in the filesystem.
Is there some other place where stuff like the `efi` loader and `eltorito.img` could go?

For now I'm using `Temporary().new_file()`.

Fixes the biggest part of #1750.

Note: Somewhat untested and unit tests haven't been adjusted.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -172,10 +172,6 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             with open(config_file, 'w') as config:
                 config.write(self.config)
 
-            if self.firmware.efi_mode():
-                if self.iso_boot:
-                    self._create_embedded_fat_efi_image()
-
     def write_meta_data(self, root_device=None, boot_options=''):
         """
         Write bootloader setup meta data files
@@ -833,20 +829,20 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             ['bash', '-c', bash_command]
         )
 
-    def _create_embedded_fat_efi_image(self):
-        Path.create(self.boot_dir + '/boot/' + self.arch)
-        efi_fat_image = ''.join(
-            [self.boot_dir + '/boot/', self.arch, '/efi']
+    def _create_embedded_fat_efi_image(self, path):
+        """
+        Creates a EFI system partition image at the given path.
+        Must be called after setup_install_boot_images and write.
+        """
+        Command.run(
+            ['qemu-img', 'create', path, '20M']
         )
         Command.run(
-            ['qemu-img', 'create', efi_fat_image, '20M']
-        )
-        Command.run(
-            ['mkdosfs', '-n', 'BOOT', efi_fat_image]
+            ['mkdosfs', '-n', 'BOOT', path]
         )
         Command.run(
             [
-                'mcopy', '-Do', '-s', '-i', efi_fat_image,
+                'mcopy', '-Do', '-s', '-i', path,
                 self.boot_dir + '/EFI', '::'
             ]
         )
--- a/kiwi/iso_tools/xorriso.py
+++ b/kiwi/iso_tools/xorriso.py
@@ -136,7 +136,7 @@ class IsoToolsXorrIso(IsoToolsBase):
             '-boot_image', 'any', 'load_size=2048'
         ]
 
-    def add_efi_loader_parameters(self) -> None:
+    def add_efi_loader_parameters(self, loader_file: str) -> None:
         """
         Add ISO creation parameters to embed the EFI loader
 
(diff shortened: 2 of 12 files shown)


Vogtinator:
It works now, the resulting iso is 20MiB smaller and still works as before. But before I spent time fixing the unit tests, please tell whether this approach is ok.

kiwi/bootloader/config/grub2.py
```diff
@@ -827,20 +823,16 @@ def _setup_bios_image(self, mbrid=None, lookup_path=None):
             ['bash', '-c', bash_command]
         )
 
-    def _create_embedded_fat_efi_image(self):
-        Path.create(self.boot_dir + '/boot/' + self.arch)
-        efi_fat_image = ''.join(
-            [self.boot_dir + '/boot/', self.arch, '/efi']
-        )
+    def create_efi_loader_image(self, path):
```
schaefi, line 826: I like your changes and this makes sense to me. Thanks for starting the refactor.

Regarding the proposed change of the function name I would not do it for the following reasons:

* by convention a method starting with "_" is a private method for which we don't provide a stable API. The new name indicates this is now a public method of the bootloader API. I don't think we should do this
* the new name is misleading to me. ```create_efi_loader_image``` indicates you are creating an EFI image. This is not what the method does. The former name ```_create_embedded_fat_efi_image``` describes what it actually does. When booting ISO images in EFI mode the ISO has to provide a fat filesystem because the EFI standard requires this. The method here wraps an existing EFI image into the standard required FAT filesystem structure. It does not create an EFI image.

So if you don't mind, can we keep the name of the method ?

kiwi/bootloader/config/grub2.py
```diff
@@ -827,20 +823,16 @@ def _setup_bios_image(self, mbrid=None, lookup_path=None):
...
         Command.run(
-            ['qemu-img', 'create', efi_fat_image, '20M']
+            ['qemu-img', 'create', path, '20M']
```
schaefi, line 828: I like the argument change and would make it even more explicit, e.g ```path_efi_fat_image``` or similar

kiwi/iso_tools/xorriso.py
```diff
@@ -147,16 +147,14 @@ def add_efi_loader_parameters(self) -> None:
         file refer to _create_embedded_fat_efi_image() from
         bootloader/config/grub2.py
         """
-        loader_file = os.sep.join([self.source_dir, self.boot_path, 'efi'])
-        if os.path.exists(loader_file):
```
schaefi, line 151: Any reason why the check for the existence of the given loader_file had to be deleted ?

schaefi (commented):
Thanks much for your work. I agree it would be good to create the embedded fat loader file outside of the ISO structure to safe space. Just some comments on the implementation

:+1:

kiwi/iso_tools/xorriso.py
Vogtinator: If `add_efi_loader_parameters` is called, the file has to exist. The check would only hide bugs.

kiwi/bootloader/config/grub2.py
Vogtinator: > by convention a method starting with "_" is a private method for which we don't provide a stable API. The new name indicates this is now a public method of the bootloader API. I don't think we should do this

I thought `_` was to indicate private methods which are only called from inside the class itself (and child classes). I'll change it.

> So if you don't mind, can we keep the name of the method ?

Will do.

kiwi/bootloader/config/grub2.py
schaefi: ok

kiwi/iso_tools/xorriso.py
schaefi: ok makes sense

Vogtinator:
Method got renamed and I added a docstring as well. Tests should be adjusted now.

I haven't built a test ISO yet, once that's done I'll report back.

schaefi (approved):
perfect :+1:

schaefi:
This is a nice refactor thanks much. Before merge I'd like to have @davidcassany a look because all this bootloader code is better reviewed by more eyes :)

Thanks

schaefi:
@Vogtinator btw, you know you have fill write permissions to the project. So you can go with branches if you like and skip the repo fork

Vogtinator:
> I haven't built a test ISO yet, once that's done I'll report back.

Done, works! I tried both livecd-tumbleweed-x11 and openSUSE-MicroOS:ContainerHost-SelfInstall on Legacy and EFI VMs.

davidcassany (approved):
Agree, nice refactor :+1:

Conan-Kudo (approved):
I don't see any reason this will break Fedora ISOs, so let's get this in!

Vogtinator:
Yay!

> Agree, nice refactor +1

I wish. If I knew more of the kiwi code, I would've changed it such that the other files (`eltorito.img`, `boot_hybrid.img`) would be read from a different location as well. Those are pretty tiny though and so more of a theoretical than a practical issue.

PR #2031, opened 2022-01-13
Add support for custom GRUB templates

### LINKED ISSUE #1970: Feature request: add support for custom bootloader template classes
## Problem description
I'm building a live ISO and would like to fully customize the grub menu's behavior, changing the full script body (including all menu entries).

## Expected behaviour
The user is able to write a custom Python class that receives [parameters](https://github.com/OSInside/kiwi/blob/master/kiwi/bootloader/config/grub2.py#L353-L374) as input. A way of specifying the class name/path should be added to `config.xml`.

iammattcoleman:
This allows the user to specify a template file to customize
the bootloader menu. This only applies to oem and iso image
types: other image types use the grub mkconfig toolchain.

Fixes #1970

### DIFF
--- a/doc/source/building_images/build_simple_disk.rst
+++ b/doc/source/building_images/build_simple_disk.rst
@@ -102,48 +102,15 @@ Setting up the Bootloader of the Image
 .. code:: xml
 
    <preferences>
-     <bootloader name="grub2"/>
+     <type>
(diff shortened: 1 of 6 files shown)


iammattcoleman:
I forgot to update the unit test.

PR #2033, opened 2022-01-17
Fix requirements for kiwi-systemdeps-filesystems

schaefi:
Extend btrfs condition applying for EL8 to apply for EL9 too

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -215,7 +215,7 @@ Obsoletes:      kiwi-filesystem-requires < %{version}-%{release}
 %if "%{_vendor}" != "debbuild"
 Provides:       kiwi-image:pxe
 Provides:       kiwi-image:kis
-%if ! 0%{?el8}
+%if ! 0%{?el8} && ! 0%{?el9}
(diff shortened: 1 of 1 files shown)


package/python-kiwi-spec-template
```diff
@@ -215,7 +215,7 @@ Obsoletes:      kiwi-filesystem-requires < %{version}-%{release}
 %if "%{_vendor}" != "debbuild"
 Provides:       kiwi-image:pxe
 Provides:       kiwi-image:kis
-%if ! 0%{?el8}
+%if ! 0%{?el8} && ! 0%{?el9}
```
Conan-Kudo, line 218: How about `%if ! (0%{?rhel} >= 8)` instead? That's what I use in the Fedora kiwi spec.

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: sounds good, how about it now ?

Conan-Kudo (approved):

PR #2035, opened 2022-01-19
Allow calling podman within chroot

schaefi:
Added helper functions and env preparation code
to allow calling podman from within a chroot. This
allows to run podman from e.g config.sh and also
inside of OBS workers

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -25,6 +25,36 @@ export LC_ALL=C
 #======================================
 # Common Functions
 #--------------------------------------
+function mountCgroups {
+    mount -t tmpfs cgroup_root /sys/fs/cgroup
+    for group in devices blkio cpuset pids; do
+        mkdir -p /sys/fs/cgroup/"${group}"
+        mount -t cgroup "${group}" -o "${group}" /sys/fs/cgroup/"${group}"
+    done
+}
+
+function umountCgroups {
+    for group in devices blkio cpuset pids; do
+        umount /sys/fs/cgroup/"${group}"
+    done
+    umount /sys/fs/cgroup
+}
+
+function setupContainerRuntime {
+    cat >/etc/containers/containers.conf <<- EOF
+	[engine]
+	no_pivot_root = true
+	cgroups = "disabled"
+	events_logger = "none"
+	cgroup_manager = "cgroupfs"
+	EOF
+
+    cat >/etc/containers/storage.conf <<- EOF
+	[storage]
+	driver = "vfs"
+	EOF
+}
+
 function baseSystemdServiceInstalled {
     local service=$1
     local sd_dirs="
--- a/kiwi/system/root_bind.py
+++ b/kiwi/system/root_bind.py
@@ -69,7 +69,8 @@ class RootBind:
             '/proc',
             '/dev',
             '/var/run/dbus',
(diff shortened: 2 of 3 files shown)


schaefi:
with this simple change I could run containers via podman in e.g config.sh

kiwi/system/root_bind.py
```diff
@@ -82,6 +83,11 @@ def mount_kernel_file_systems(self) -> None:
             fails to mount
         """
         try:
+            shared_mount = MountManager(
+                device=self.root_dir, mountpoint=self.root_dir
```
davidcassany, line 87: I don't understand that...

kiwi/config/functions.sh
```diff
@@ -25,6 +25,36 @@ export LC_ALL=C
 #======================================
 # Common Functions
 #--------------------------------------
+function mountCgroups {
+    mount -t tmpfs cgroup_root /sys/fs/cgroup
+    for group in devices blkio cpuset pids; do
+        mkdir -p /sys/fs/cgroup/"${group}"
+        mount -t cgroup "${group}" -o "${group}" /sys/fs/cgroup/"${group}"
+    done
+}
```
davidcassany, line 34: I trust you on all these functions, I never done something like this, so I have no experience. Looks good without knowing the details.

kiwi/system/root_bind.py
schaefi: yes sorry my fault, this cannot be understood without an explanation. The way this mount is done is a trick to overcome an issue in docker and podman when called from within a chroot. There is some information in this thread

* https://github.com/moby/moby/issues/34817

I think a comment with that issue link is a must have

kiwi/config/functions.sh
schaefi: This is just a simple cgroup setup in a new tmpfs. I found out that podman cannot be used with cgroups disabled, even though there is a configuration option for it available it complains at "podman run", as there is no systemd in an obs env I choose to use cgroupfs and a new cgroup's mount.

I did not add this information to the method as it just creates a new cgroup mount list which is a common task

Do you think we need more information ?

schaefi:
@davidcassany thanks for taking the time to look at this code. I will definitely add some comments

kiwi/config/functions.sh
Conan-Kudo: Did you try with OBS' systemd-nspawn mode for builds instead of using simple chroot?

schaefi: I was not aware that you can choose this in any way ? are there docs ? And even if it can be selected I have the requirement that it has to work from within a chroot.

Conan-Kudo: Yes, it's selectable in sysconfig files.

kiwi/system/root_bind.py
schaefi: done

kiwi/config/functions.sh
schaefi: Maybe you can give me a pointer into that and how it hooks into obs, on the matrix channel ? Thanks much

Other than that if nobody objects I would go forward with the implementation for now ? The way it's provided in this PR I have tested and works fine in obs as well as in the private instance at EB for which it relates to the job

Thanks

PR #2039, opened 2022-01-21
Set /.snapshots subvolume to mode 0700 (bsc#1194992)

Vogtinator:
Avoid that users other than root can enter or even change the content.
This is what snapper does as well.


schaefi (approved):
Looks good thanks :+1:

PR #2040, opened 2022-01-21
Fixed grub loader/entries setup

### LINKED ISSUE #2038: CentOS 9 Stream disk images are broken
## Problem description

The disk images that are produced by kiwi for Centos 9 stream are currently all broken, when you boot them, they will simply stall at "booting from hard drive". The serial console shows the actual issue:
```
error: ../../grub-core/fs/fshelp.c:257:file
`/usr/src/packages/KIWI-oem/build/image-root
/boot/vmlinuz-5.14.0-43.el9.x86_64' not found.
error: ../../grub-core/loader/i386/pc/linux.c:422:you need to load the kernel
first.

Press any key to continue...
```

Affected images:
- kiwi-test-image-live-disk-v9.x86_64-1.3.0-Disk-Build7.1.raw
- kiwi-test-image-live-disk-v9.x86_64-1.3.0-Virtual-Build7.1.qcow2
- the disk image that is created after booting from kiwi-test-image-live-disk-v9.x86_64-1.3.0-Disk-Build7.1.install.iso

## Expected behavior

The disk images should boot.

+ KIWI version: latest staging
+ Operating system: n/a
+ OBS version: whatever is currently deployed on build.opensuse.org

schaefi:
If called in non standard environments like an OBS worker
the grub tooling does not work correctly and produces invalid
results. For these cases kiwi provides a collection of fix_
methods to change the produced results. This commit covers
the invalid path in loader/entries/*.conf files pointing to
the kernel and the initrd as they exist in the image-root
directory. This results for example in settings like:

* linux /usr/src/packages/KIWI-oem/build/image-root/boot/vmlinuz-5.14.0-43.el9.x86_64

when it should be:

* linux /boot/vmlinuz-5.14.0-43.el9.x86_64

This Fixes #2038

### DIFF
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -777,7 +777,8 @@ class TestBootLoaderConfigGrub2:
                 'root=rootdev nomodeset console=ttyS0 console=tty0\n' \
                 'root=PARTUUID=xx'
             file_handle_grubenv.read.return_value = 'root=rootdev'
-            file_handle_menu.read.return_value = 'options foo bar'
+            file_handle_menu.read.return_value = \
+                'options foo\nlinux unexpected/boot/vmlinuz\ninitrd /boot/initrd'
 
             self.bootloader.setup_disk_image_config(
                 boot_options={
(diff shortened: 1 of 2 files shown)


test/unit/bootloader/config/grub2_test.py
```diff
@@ -777,7 +777,8 @@ def open_file(filename, mode=None):
                 'root=rootdev nomodeset console=ttyS0 console=tty0\n' \
                 'root=PARTUUID=xx'
             file_handle_grubenv.read.return_value = 'root=rootdev'
-            file_handle_menu.read.return_value = 'options foo bar'
+            file_handle_menu.read.return_value = \
+                'options foo\nlinux bullshit/boot/vmlinuz\ninitrd /boot/initrd'
```
Conan-Kudo, line 781: We probably shouldn't put profanity into the codebase...

Conan-Kudo (changes requested):

test/unit/bootloader/config/grub2_test.py
davidcassany: yes, however I have to admit it is hard not to get emotional when coding this :)

Conan-Kudo: Yeah, I agree. It's all terrible. :)

davidcassany (approved):
Looks good to me. But I also agree @Conan-Kudo got a point, probably the test literals should be changed.

test/unit/bootloader/config/grub2_test.py
schaefi: ok, guys I agree, sorry I was a little stressed when I coded this. I will change it

schaefi: it's gone, hope that's better

Conan-Kudo (approved):

PR #2043, opened 2022-01-24
Fix ramdisk deployments from PXE

davidcassany:
This commit fixes PXE deployments on ramdisk. In such cases the
former fix from df4e62a4 is not sufficient as there is no `root=`
parameter within the kernel cmd line and hence this logic is never
executed.


ignatenkobrain (approved):

Conan-Kudo (approved):

schaefi (approved):
great catch thanks :+1:

PR #2044, opened 2022-01-24
Fix the uninstall(force) on Debian based dists

schaefi:
Packages marked for uninstall via <package name="delete|uninstall"/>
failed to become removed for several reasons. The way this was done
in kiwi did not work because dpkg needs to be called differently
and with some nasty pre-processing in order to allow for force
deletion. In force mode we also allow to remove packages marked as
essential. In gracefull uninstall mode this commit makes sure the
environment is prepared and does not fail for false-positive
reasons.

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -270,9 +270,50 @@ class PackageManagerApt(PackageManagerBase):
                 'None of the requested packages to delete are installed'
             )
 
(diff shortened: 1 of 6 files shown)


Conan-Kudo (approved):
Gross, but sadly makes sense...

davidcassany (approved):
Oh God, the more I know about debian package tool chain (dpkg, apt...) the less I like it...

PR #2046, opened 2022-01-25
Ensure SELinux labels are set based on the policy

Conan-Kudo:
When running kiwi from a filesystem tree that has custom labels applied
(such as when using kiwi from within a container on an SELinux-enabled
host), the filesystem labeling doesn't correctly apply on some files
and folders with a warning about the location being customized by
the administrator. This causes all kinds of strange results with
the built images and makes them unbootable.

To resolve this, tell setfiles to forcibly set files and folders
with the default context from the installed policy.

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -476,7 +476,7 @@ class SystemSetup:
         Command.run(
             [
                 'chroot', self.root_dir,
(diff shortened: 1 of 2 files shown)


ignatenkobrain (approved):

schaefi (approved):
Great catch, thanks :+1:

PR #2047, opened 2022-01-25
Fix use of xattrs for container sync

### LINKED ISSUE #2009: Include specific set of xattrs for OCI images
## Problem description

```xml

<image schemaversion="7.4" name="c9s">
    <description type="system">
        <author>Igor Raits</author>
        <contact>igor.raits@gmail.com</contact>
        <specification>CentOS Stream 9</specification>
    </description>
    <preferences>
        <version>1.0.0</version>
        <!-- <release-version>9</release-version> -->
        <packagemanager>dnf</packagemanager>
        <locale>en_US</locale>
        <keytable>us</keytable>
        <timezone>Europe/Prague</timezone>
        <type image="docker">
            <containerconfig name="c9s-base"/>
        </type>
        <rpm-check-signatures>true</rpm-check-signatures>
    </preferences>
    <repository type="rpm-md" alias="baseos">
        <source path="http://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/"/>
    </repository>
    <repository type="rpm-md" alias="appstream">
        <source path="http://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/"/>
    </repository>
    <packages type="image" patternType="onlyRequired">
        <package name="bash"/>
    </packages>
    <packages type="bootstrap">
        <package name="centos-stream-release"/>
    </packages>
</image>

```

Building such image fails on:

```
[ DEBUG   ]: 00:05:06 | EXEC: [rsync -a -H -X -A --one-file-system --inplace --delete --exclude /image --exclude /.profile --exclude /.kconfig --exclude /run/* --exclude /tmp/* --exclude /.buildenv --exclude /var/cache/kiwi --exclude /dev/* --exclude /sys/* --exclude /proc/* /home/brain/tmp/tmp/out/build/image-root/ /var/lib/containers/storage/overlay/83aacfbaf9fc572a5850a7513c9619076514c7d15f931146b0e92eb7235c8ff4/merged]
[ DEBUG   ]: 00:05:06 | EXEC: Failed with stderr: rsync: [receiver] rsync_xal_set: lsetxattr("/var/lib/containers/storage/overlay/83aacfbaf9fc572a5850a7513c9619076514c7d15f931146b0e92eb7235c8ff4/merged/.","security.selinux") failed: Operation not supported (95)
… (bunch more messages about selinux)
```

## Expected behaviour

Most likely rsync needs to skip extended attributes as those are not useful in the containers .

Get simple config.xml and try to build an image.

+ KIWI version: 9.24.13 (+ `buildah-1.23.1-1.fc36.x86_64` + `skopeo-1.5.2-1.fc36.x86_64`)
+ Operating system: Fedora 36 (host), CentOS Stream 9 (target)
+ OBS version: N/A

schaefi:
when syncing data for containers only a subset of xattr
attributes can be applied. This Fixes #2009

### DIFF
--- a/kiwi/oci_tools/buildah.py
+++ b/kiwi/oci_tools/buildah.py
@@ -161,7 +161,12 @@ class OCIBuildah(OCIBase):
         self._sync_data(
             ''.join([root_dir, os.sep]), self.oci_root_dir,
             exclude_list=exclude_list,
-            options=Defaults.get_sync_options() + ['--delete']
+            options=Defaults.get_sync_options() + [
+                '--filter', '-x! user.*',
+                '--filter', '-x! security.ima*',
+                '--filter', '-x! security.capability*',
+                '--delete'
+            ]
         )
 
     def import_rootfs(self, root_dir, exclude_list=None):
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -314,7 +314,10 @@ class Defaults:
 
         :rtype: list
         """
-        return ['-a', '-H', '-X', '-A', '--one-file-system', '--inplace']
+        return [
+            '--archive', '--hard-links', '--xattrs', '--acls',
+            '--one-file-system', '--inplace'
+        ]
 
     @staticmethod
     def get_exclude_list_for_root_data_sync():
--- a/kiwi/oci_tools/umoci.py
+++ b/kiwi/oci_tools/umoci.py
@@ -123,7 +123,12 @@ class OCIUmoci(OCIBase):
             ''.join([root_dir, os.sep]),
             os.sep.join([self.oci_root_dir, 'rootfs']),
             exclude_list=exclude_list,
-            options=Defaults.get_sync_options() + ['--delete']
+            options=Defaults.get_sync_options() + [
+                '--filter', '-x! user.*',
+                '--filter', '-x! security.ima*',
+                '--filter', '-x! security.capability*',
+                '--delete'
+            ]
         )
 
     def import_rootfs(self, root_dir, exclude_list=None):
--- a/kiwi/utils/sync.py
+++ b/kiwi/utils/sync.py
@@ -89,11 +89,11 @@ class DataSync:
             rsync_options = options
         if not self.target_supports_extended_attributes():
             warn_me = False
-            if '-X' in rsync_options:
-                rsync_options.remove('-X')
+            if '--xattrs' in rsync_options:
+                rsync_options.remove('--xattrs')
                 warn_me = True
-            if '-A' in rsync_options:
-                rsync_options.remove('-A')
+            if '--acls' in rsync_options:
+                rsync_options.remove('--acls')
                 warn_me = True
             if warn_me:
                 log.warning(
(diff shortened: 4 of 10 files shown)


schaefi:
I realized rsync filters differently. The option does not work the way I though it would... this needs to change

schaefi:
I think we need

```
--filter=’-x! user.*’
--filter=’-x! security.ima*’
--filter=’-x! security.capability*’
```

kiwi/oci_tools/buildah.py
```diff
@@ -161,7 +161,12 @@ def sync_rootfs(self, root_dir, exclude_list=None):
         self._sync_data(
             ''.join([root_dir, os.sep]), self.oci_root_dir,
             exclude_list=exclude_list,
-            options=Defaults.get_sync_options() + ['--delete']
+            options=Defaults.get_sync_options() + [
+                '--filter="-x! user.*"',
+                '--filter="-x! security.ima*"',
+                '--filter="-x! security.capability*"',
```
Conan-Kudo, line 167: If we don't sync caps, then things like unprivileged ping wouldn't work, right?

Conan-Kudo (changes requested):

kiwi/oci_tools/buildah.py
schaefi: I think I can't follow you. The "security.capability*" matches what you described as "caps" or is this a different namespace ?

actually the filter definition of "-x! namespace" says filter out everything but namespace (negated match). I hope that multiple of these filters can be combined. And yes this needs testing, I haven't done this before with rsync

So which namespace are you referring to exactly ?

Thanks

Conan-Kudo: I think I didn't understand what this option does. Are you saying we're syncing this attribute in?

schaefi: yes this is at least how I understood it when reading man rsync

```
 To skip all namespaces except  the  user  namespace,  you  could
              specify a negated-user match:

              --filter=’-x! user.*’

```

Conan-Kudo: Sounds good to me, then.

Conan-Kudo (approved):

schaefi:
Thanks, let's keep it as "don't merge" because I wanted to test this. Which brings me to a good question. How can I create a file with extended attributes such that I can test the filters. I would take one with attributes such that it gets filtered out and one with attributes that matches the filter criteria. To be honest I never have worked with files and extended attributes 

Thanks for any hint

kiwi/oci_tools/buildah.py
```diff
@@ -161,7 +161,12 @@ def sync_rootfs(self, root_dir, exclude_list=None):
         self._sync_data(
             ''.join([root_dir, os.sep]), self.oci_root_dir,
             exclude_list=exclude_list,
-            options=Defaults.get_sync_options() + ['--delete']
+            options=Defaults.get_sync_options() + [
+                '--filter="-x! user.*"',
```
ignatenkobrain, line 165:
```suggestion
                '--filter', '-x! user.*',
```

… and so on.

Otherwise I get

```
1[ ERROR   ]: 23:08:52 | KiwiCommandError: rsync: stderr: Unknown filter rule: `"-x! user.*"'
```

schaefi: Hey, thanks for giving this a shot. I adapted the PR as you suggested

ignatenkobrain: Thanks! it fixes my issue :)

ignatenkobrain (approved):

Conan-Kudo (approved):

PR #2051, opened 2022-01-31
Ensure backward compatibility on deprecated methods

davidcassany:
This commit ensures backward compatibility for deprecated
config bash script utilities.

Fixes bsc#1195229

### DIFF
--- a/kiwi/config/functions.sh
+++ b/kiwi/config/functions.sh
@@ -877,7 +877,7 @@ function deprecated {
         cat
         echo "]"
     } >&2
-    exit 1
+    exit 0
 }
 
 


Conan-Kudo:
What is this BZ about?

davidcassany:
Some images and templates SUMA exposes to users can't use an up to date KIWI because of this deprecated method causing error builds. Note that this kind of manual changes (simply removing deprecated methods) might not be that obvious for automated builds where the user has no real direct contact to KIWI.

Conan-Kudo:
That sounds like a problem with SUMA rather than us?

davidcassany:
Kind of, lets put it this way: SUMA did not follow our deprecation advises and we (SUSE employees) could have done better to ensure they were aware of it.

Anyway I am starting to think that probably assuming a patch at SLE side is the way to go, even that puts some maintenance overhead on my plate...

davidcassany:
I leave it opened just to give it a chance to @schaefi to go through it. I am preparing a backport on SLE side.

schaefi:
Hrm, this is a really small change and a big impact :)

To be honest I'm not so much in favor to to turn this button on the upstream code. The reason is, I think we had an agreement that we don't want people to continue using shell functions which we considered deprecated. We gave them a six month period from obsolete (warning only) to deprecated (error), send a mail to the mailing list and also documented the still present methods. From my experience people only move when they have to and the process before is to make sure we tried hard to not offend anybody by a fast and not announced change.

So now there is this PR which changes the exit code from 1 to 0 which means we are turning back and allow the stuff we wanted to get rid of to stay for how long ? ... probably forever. This is the part I think we should stay strict, at least upstream

On the other hand I can understand that SLES cannot move because customers with older descriptions must be able to build their images within the same major SLES code stream. But exactly for that reason the git fork (kiwi_sle15) exists which allows to maintain changes like the proposed one.

So unless people object and overrule my thinking I suggest to not merge this one into master

Thoughts ?

davidcassany:
Yes completely agree, in fact, the patch on SLE is already in the pipeline.

PR #2052, opened 2022-01-31
Fixed UUID used in grub early boot script

schaefi:
In case the system is luks encrypted the UUID of the root
partition was used in the grub early boot script. However,
this condition is only correct if in addition to the luks
encryption the boot area is on crypto too. If boot is not
on crypto the UUID must be the boot partition and not root.
Only if root AND boot is on crypto the kiwi created early
boot script includes the grub cryptomount calls.


schaefi:
Jack Sedita verified the this change to fix the desktop image build based on luks+lvm(with extra bootpart)

Going to merge

PR #2053, opened 2022-02-03
treat armv8l as armv7hl

dirkmueller:
openSUSE switched from armv7hl buildworkers to aarch64 in 32bit mode
which identifies itself as "armv8l" in uname -m.

Fixes # .

*
*


schaefi (approved):
Thanks :+1:

PR #2055, opened 2022-02-04
Fixed handling of oem reboot settings

schaefi:
There are oem settings called oem-reboot, oem-reboot-interactive
as well as oem-shutdown and oem-shutdown-interactive. When used
the information is passed along to the profile but not evaluated
by any initrd code. I don't know where on the way we lost the
code that actually works with these settings but this commit
makes them effective.


schaefi:
@davidcassany would be great if you find the time to look at this change. It's initrd code you know it's mean and always strikes back :) Thanks much.

I also set the WIP label as I wanted to test it extensively before merge

davidcassany (approved):
Looks good! It is incredible we lost that at some point... 😓 Good this is all back. All these portions from XML to dracut modules are hard to track.

schaefi:
@davidcassany Thanks for looking at it. I have done some tests and it worked in the desired way. I will merge this as soon as the tests passes

PR #2059, opened 2022-02-08
Make use of container name in OCI images

### LINKED ISSUE #2050: buildah: tag is not really a tag but name
## Problem description

```xml
<type image="oci">
  <containerconfig name="gdc-c9s-base" tag="latest"/>
</type>
```

^ with buildah results into a container that has name `latest` which suggests that the `name` attribute is not used at all.

## Expected behaviour

`gdc-c9s-base:latest`

1. Build image with above settings
2. `podman load -i …`
3. `podman images`

+ KIWI version: 9.24.19
+ Operating system: Fedora 35
+ OBS version: N/A

davidcassany:
Fixes #2050

### DIFF
--- a/kiwi/container/oci.py
+++ b/kiwi/container/oci.py
@@ -136,21 +136,18 @@ class ContainerImageOCI:
         oci.set_config(self.oci_config)
         oci.post_process()
 
+        image_ref = '{0}:{1}'.format(
+            self.oci_config['container_name'],
+            self.oci_config['container_tag']
+        )
+        additional_refs = []
         if self.archive_transport == 'docker-archive':
-            image_ref = '{0}:{1}'.format(
-                self.oci_config['container_name'],
-                self.oci_config['container_tag']
-            )
-            additional_refs = []
             if 'additional_tags' in self.oci_config:
                 additional_refs = []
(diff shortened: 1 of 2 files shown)


schaefi (approved):
Looks good to me. We set name now in any case and independent of the transport

I assume this is wanted. The former implementation did it only for the docker transport. So I guess the name tag was not supported in OCI formerly ?

I'm fine with this change, it simplifies the code a lot

Thanks

davidcassany:
> I assume this is wanted. The former implementation did it only for the docker transport. So I guess the name tag was not supported in OCI formerly ?

Yes. In former days there was no `oci-archive` transport supported in skopeo and we were simply doing a tarball of the oci image layout created by umoci. OCI image layouts do not have repository or name concept. As far as I am aware this is a transport only concept, `umoci` does not know about the `<repo>:<tag>` image references, `umoci` only knows about tags. I am not completely sure this is/was equivalent for `buildah`, however since we use skopeo to export images regardless of using  `buildah` or `umoci` toolchain we are safe here, the names are applied on the `skopeo copy <internal-image> <final-image>` command.

schaefi:
perfect, thanks much for the explanation. Just merge as you see fit :)

PR #2060, opened 2022-02-08
Fixup inplace podman storage and container conf

schaefi:
Newer versions of podman requires runroot and graphroot
to be explicitly set in storage.conf.

Newer versions of podman no longer reads the engine.cgroups
setting on containers.conf and prints a 'Failed to decode the
keys [\"engine.cgroups\"]' warning message

This commit fixes storage.conf and containers.conf written
by kiwi if the setupContainerRuntime method is used in
scripts.


davidcassany (approved):
Looks good

PR #2062, opened 2022-02-10
Added overlayroot_write_partition attribute

schaefi:
For the oem type only, allows to specify if the extra read-write
partition in an overlayroot setup should be created or not.
By default the partition is created and the kiwi-overlay dracut
module also expect it to be present. However, the overlayroot
feature can also be used without an initrd and under certain
circumstances it is handy to configure if the partition table
should contain the read-write partition or not.

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -572,6 +572,15 @@ overlayroot="true|false"
   will not have any effect because the write partition will be
   resized on first boot to the available disk space.
 
+overlayroot_write_partition="true|false"
+  For the `oem` type only, allows to specify if the extra read-write
+  partition in an `overlayroot` setup should be created or not.
+  By default the partition is created and the kiwi-overlay dracut
+  module also expect it to be present. However, the overlayroot
+  feature can also be used without an initrd and under certain
+  circumstances it is handy to configure if the partition table
+  should contain the read-write partition or not.
+
 bootfilesystem="ext2|ext3|ext4|fat32|fat16":
   If an extra boot partition is required this attribute
   specify which filesystem should be used for it. The
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -577,9 +577,9 @@ overlayroot_write_partition="true|false"
   partition in an `overlayroot` setup should be created or not.
   By default the partition is created and the kiwi-overlay dracut
   module also expect it to be present. However, the overlayroot
-  feature can also be used without an initrd and under certain
-  circumstances it is handy to configure if the partition table
-  should contain the read-write partition or not.
+  feature can also be used without dracut (`initrd_system="none"`)
+  and under certain circumstances it is handy to configure if the
+  partition table should contain the read-write partition or not.
 
 bootfilesystem="ext2|ext3|ext4|fat32|fat16":
   If an extra boot partition is required this attribute
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -89,6 +89,8 @@ class DiskBuilder:
             xml_state.build_type.get_spare_part_mountpoint()
         self.persistency_type = xml_state.build_type.get_devicepersistency()
         self.root_filesystem_is_overlay = xml_state.build_type.get_overlayroot()
+        self.root_filesystem_has_write_partition = \
+            xml_state.build_type.get_overlayroot_write_partition()
         self.custom_root_mount_args = xml_state.get_fs_mount_option_list()
         self.custom_root_creation_args = xml_state.get_fs_create_option_list()
         self.build_type_name = xml_state.get_build_type_name()
@@ -268,6 +270,12 @@ class DiskBuilder:
         # create disk partitions and instance device map
         device_map = self._build_and_map_disk_partitions(disk, disksize_mbytes)
 
+        if self.root_filesystem_is_overlay and \
+           self.root_filesystem_has_write_partition is False:
+            device_map['root'] = device_map['readonly']
+            disk.public_partition_id_map['kiwi_RootPart'] = \
+                disk.public_partition_id_map['kiwi_ROPart']
+
         # create raid on current root device if requested
         raid_root = None
         if self.mdraid:
@@ -318,6 +326,9 @@ class DiskBuilder:
                 )
             device_map['luks_root'] = device_map['root']
             device_map['root'] = luks_root.get_device()
+            if self.root_filesystem_is_overlay and \
+               self.root_filesystem_has_write_partition is False:
+                device_map['readonly'] = device_map['root']
 
         # create spare filesystem on spare partition if present
         system_spare = self._build_spare_filesystem(device_map)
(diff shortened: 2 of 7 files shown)


schaefi:
Not yet complete. There is a device map issue when using this attribute in combination with luks encryption

kiwi/builder/disk.py
```diff
@@ -823,17 +834,21 @@ def _build_and_map_disk_partitions(
...
-
+        if self.root_filesystem_is_overlay and \
+           self.root_filesystem_has_write_partition is False:
+            log.warning(
+                '--> overlayroot explicitly requested no write partition'
+            )
         else:
-            log.info('--> creating root partition')
-            disk.create_root_partition(rootfs_mbsize)
+            if self.volume_manager_name and self.volume_manager_name == 'lvm':
```
davidcassany, line 843: Why are you embedding this part as an `else`? Isn't it validated before, is it possible having `mraid` with overlay and write partition? This change of having this part under an `else` for the new use case feels odd or unneeded.

doc/source/image_description/elements.rst
```diff
@@ -572,6 +572,15 @@ overlayroot="true|false"
   will not have any effect because the write partition will be
   resized on first boot to the available disk space.
 
+overlayroot_write_partition="true|false"
+  For the `oem` type only, allows to specify if the extra read-write
+  partition in an `overlayroot` setup should be created or not.
+  By default the partition is created and the kiwi-overlay dracut
+  module also expect it to be present. However, the overlayroot
+  feature can also be used without an initrd and under certain
```
davidcassany, line 580: I don't understand this explanation 😓 How is an OEM image wihout an initrd? And where is the relation between having or not an initrd and the use of a persistent partitions for overlay deployments?

I am failing to understand potential use cases.

davidcassany (commented):
A couple of minor comments, mostly because I don't understand what you are trying to solve.

A question about the behavior, IIRC this partitions are created at boot time, so this change just instructs the dracut modules to create the partition or not, is that correct?

doc/source/image_description/elements.rst
schaefi: ok sorry, let me explain the use case first. See the following type definition

```xml
<type image="oem" filesystem="ext2" firmware="custom" initrd_system="none" overlayroot="true" overlayroot_write_partition="false" bootpartition="true" bootpartsize="20">
    <bootloader name="grub2"/>
    <size unit="M">128</size>
    <partitions>
        <partition name="app" size="10" mountpoint="/app" filesystem="ext2"/>
    </partitions>
</type>
```

This is the type setup of an image with a read-only rootfs based on squashfs. The only persistent write area is a special partition for the application. In kiwi we support squashfs root via the ```overlayroot``` attribute. Along with this overlayroot feature there is always an extra partition that can be used to make the system writable through overlayfs. This extra partition is unwanted on production. The new attribute ```overlayroot_write_partition="true|false"``` allows to control if
this extra write partition for the overlay exists or not.

This is the use case.

If the changelog comment doesn't make this clear it would be great if you can rephrase it such that it makes sense for you. I'm starring at this for tool long and maybe couldn't find the right words to explain it :)

Thanks

schaefi: Also to mention the system boots up with a custom kernel. No initrd is required to bootup as bootup times matter and all drivers are compiled in. Think embedded ;)

kiwi/builder/disk.py
schaefi: Hmm, I'm not sure what you mean by that and maybe it is a bit hard to read how diff structured the change. If you take the code in question as a block it looks like this

```python
if self.root_filesystem_is_overlay and \
   self.root_filesystem_has_write_partition is False:
    log.warning(
                '--> overlayroot explicitly requested no write partition'
            )
   else:
       if self.volume_manager_name and self.volume_manager_name == 'lvm':
           log.info('--> creating LVM root partition')
           disk.create_root_lvm_partition(rootfs_mbsize)
       elif self.mdraid:
           log.info('--> creating mdraid root partition')
           disk.create_root_raid_partition(rootfs_mbsize)
       else:
           log.info('--> creating root partition')
           disk.create_root_partition(rootfs_mbsize)
```

from that perspective I think it makes sense. In overlayroot and without write partition we don't want an additional partition. Also note in overlayroot mode the ro partition is named rootro and the read write partition is named ```root```
In fact in an overlay setup the write partition is treated as the defacto root partition. The condition for creating this
partition or not is formulated in code like above

And back to your question, yes we also support the root overlay on top of a raid. It imho doesn't make sense but it is possible.

Does this makes more sense now ?

davidcassany: Yes it does,  I got confused by the early check on volume_manager_name against root_filesystem_is_overlay.

doc/source/image_description/elements.rst
davidcassany: I see that sounds cool, relatively similar to what I am doing nowadays in terms of setting an overlayfs over an immutable rootfs. 

Out of curiosity how this partition setup works together with  `overlayroot="true"`. I'd say the result is one partition of 10Mb and then a squashfs image dd'ed to a second partition including an fstab line to mount `/app` from the previous partition. I am getting it right? That's fun.

davidcassany (approved):

doc/source/image_description/elements.rst
schaefi: yes on  a block layer this looks like this

```
BusyBox v1.30.1 (Ubuntu 1:1.30.1-7ubuntu2) built-in shell (ash)
Enter 'help' for a list of built-in commands.

# lsblk 
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
vda    254:0    0  128M  0 disk 
|-vda1 254:1    0   20M  0 part
|-vda2 254:2    0   10M  0 part /app
`-vda3 254:3    0   93M  0 part /
```

Only /app is a writable area and the application is prepared for this condition. I'd say it's not possible in a generic way because most linux applications assumes they can write. Fortunately in this world and where it runs this is not an issue. So all on / is read-only and immutable. That also means you can switch off/on the hardware, fast reboot/crash-reboot
cycles were also important. The extra 20MB is the boot partition but it never gets mounted. The system boots up via uboot and that's the only place the kernel gets read from the disk. Unfortunately uboot does fatload so you have to have a fat partition which is that 20M extra space. During development there is an additional partition which allows to boot up the system with overlayroot in RW mode. This is very handy to test/debug and so on. On Production I just add the attribute from this PR and the extra write space is gone

kiwi/schema/kiwi.rnc
```diff
@@ -1537,6 +1537,22 @@ div {
             sch:param [ name = "attr" value = "overlayroot" ]
             sch:param [ name = "types" value = "oem" ]
         ]
+    k.type.overlayroot_write_partition.attribute =
+        ## Specifies to create an extra read-write partition
+        ## in combination with the overlayroot attribute on the
+        ## target disk. By default the partition is created
+        ## and the kiwi-overlay dracut module also expect it
+        ## to be present. However, the overlayroot feature could
+        ## also be used without an initrd and under certain
```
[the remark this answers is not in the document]
schaefi: ok thanks, I have updated the schema comment and also the docs to be more specific here

PR #2065, opened 2022-02-12
Systemd bootloader spec interface

### LINKED ISSUE #1935: Add support for systemd-boot
## Description

For booting EFI images systemd provides a bootloader called systemd-boot. It might be good to evaluate
supporting this bootloader in kiwi as well. Details here:

* https://wiki.archlinux.org/title/systemd-boot

schaefi:
Added API for BLS and systemd-boot support
    
Bootloaders implementing the Boot Loader Spec (BLS) are not directly compatible with the original Bootloader Class design in kiwi. Because of that an interface class which translates the original API into calls following BLS was added. This allows us to keep the implementations in the Builder classes and the public BootLoader interface untouched. In addition to the BLS API an implementation to support the systemd-boot loader is part of this commit too.
    
An example type definition to use systemd-boot with an EFI disk image type looks like the following:
    
```xml
<type image="oem" filesystem="xfs" firmware="efi" bootloader="systemd_boot" efipartsize="200"/>
```

The implementation uses ```bootctl``` and ```kernel-install``` tools provided from systemd and expects a proper integration of systemd-boot by the distribution maintainers

This Fixes #1935


Conan-Kudo:
I think we should not call this `systemd_spec`, since there are multiple implementations (GRUB2 and sd-boot).

Let's have this module call itself `bootloader_spec` instead.

cc: @martinezjavier who can probably review thie spec implementation better...

schaefi:
> Let's have this module call itself `bootloader_spec` instead.

ok, I'm fine with that. The reason why I named it that way was because it seems the bootloader spec originates from systemd: https://systemd.io/BOOT_LOADER_SPECIFICATION/

Conan-Kudo:
@keszybz, could you take a look at this and see if we're on the right path here?

kiwi/bootloader/config/systemd_spec_base.py
```diff
@@ -0,0 +1,211 @@
...
+        ('disk', str),
+        ('install', str),
+        ('live', str)
+    ]
+)
+
+
+class BootLoaderSystemdSpecBase(BootLoaderConfigBase):
+    """
+    **Base class for Bootloader Spec configuration**
```
keszybz: I think it'd be better to spell this "Boot Loader Specification" to match the spec docs.

FWIW, the specification was very much designed to be independent of any implementation. So "systemd" is **not** part of the name. We nowadays say add "`Type #1`" or "`Type #2`" to distinguish unified kernels where appropriate.

Confusion was created by grub2 implementing something incompatible and using the same name. But those patches are only downstream in Fedora/RHEL and maybe won't even be merged anywhere. So I think it'd be better to distinguish that special variant by some special name, "GRUB2 BLS-based template language" or something, if it is ever implemented here.

keszybz:
> @keszybz, could you take a look at this and see if we're on the right path here?

Hmm, this patch doesn't do anything real, right? I don't think I can comment usefully on the code structure, I don't know kiwi internals…

kiwi/bootloader/config/systemd_spec_base.py
Conan-Kudo: Strictly speaking, that template stuff are supposed to be mostly gone now...

keszybz: I think that apart from the syntax changes, the other differences are not important. Grub2 used some different locations for files, but that kind of thing is better handled through a configuration parameter, not a separate class.

schaefi:
> > @keszybz, could you take a look at this and see if we're on the right path here?
> 
> Hmm, this patch doesn't do anything real, right? I don't think I can comment usefully on the code structure, I don't know kiwi internals…

yes true. Your feedback is in any case valuable as you have more background on how the bootloaders evolve. in kiwi there is a generic BootLoaderConfig class which is used in all its builders. So from an api perspective you just create an instance of this class and the same api can be used for all bootloaders we support. Actually a similar approach to what the "Boot Loader Specification" is doing today. At the time it was done for the kiwi and appliance use case no such bootloader specification existed and when it existed it was of course not compatible with the class interface in kiwi.

Since we cannot afford to change the BootLoaderConfig interface in kiwi right now there is this idea to introduce a new meta class which translates one api into another. This is what you can see here as ```BootLoaderSpecBase```. The actual implementation for e.g systemd_boot is then a class that inherits from it ```class BootLoaderSystemdBoot(BootLoaderSpecBase):``` and only implements the api as we see fit to match the "Boot Loader Specification". Once we do a major shift in kiwi we can also delete the meta class without harming the BLS implementations.

You are right all this is not something you care for. It would still be interesting if you think we are missing a fundamental configuration stage when looking at ```bootloader/config/systemd_boot.py```. From my perspective a bootloader compatible to BLS might need some custom setup code (setup_loader) has to add loader entries (set_loader_entry) and might need to create bootloader binary images, things like efi, shim, core... usually all provided by the distro itself but who knows (create_loader_image)

So the api for BLS that I had in mind just implements these three methods. And for that part whether it's all bullshit or kind a ok I would appreciate your and other people's feedback.

The situation around grub (not BLS) is so messy that it is hard to maintain all this for the different distributions. If possible I'd like to avoid some of the mistakes for the next incarnation of bootloaders :-)

Thanks much

keszybz:
> how the bootloaders evolve

Current thinking (and it's just that, nothing concrete so far) is to make things more EFI-like, even if a particular platform doesn't do EFI. So basically you have a small partition that is identified by a fixed UUID, and all systems drop kernel images there, and BLS is used to list them. I think there's general agreement from various sides that this is the way to go, provide interoperability and abstract differences between systems. We hope to standarize even the boot counting mechanism, so that you can switch boot loaders at any point and get compatible logic.

The second direction of development is unified kernel images. If we move towards generating and signing initrds in distro infrastracture, which we need to do to support SecureBoot more sanely, installing of the kernel+initrd combo would just mean copying a single file to the partition, instead of a few separate files. But even in this scheme we'll still need to do some local steps, for example sign the locally-configured kernel command line using TPM2 on the machine. Though this particular step probably doesn't matter for kiwi if it's building generic images.

The boot loader would generally pick the newest entry (with the highest version number), no explicit config necessary.

I don't know if the above is helpful. Starting from another angle: to install the boot loader and everything else, at various points access is needed to: create ESP and/or XBOOTLDR partitions, put files on ESP and/or XBOOTLDR, save `/etc/kernel/entry-token`, for installations on the final destination system, also access local configuration like user credentials and other secrets, TPM2 for encryption, and ability to save stuff on ESP and/or XBOOTLDR. Right now we also need to build initrds, i.e. have access to everything that `dracut` requires, in particular the installed system. Hopefully this will not be needed in the future, so it'll be possible to create the boot loader installation before or after the main system is installed.

Conan-Kudo:
From my point of view, I think the structure makes sense, but I'd like to see the systemd-boot interface implemented first to see if it is actually going to work.

kiwi/bootloader/config/systemd_spec_base.py
schaefi: Thanks, I applied the proposed changes on the naming.

grub2 BLS will be a matter of a new pull request. With this one the main interest is to get the API right to handle BLS compatible loaders and systemd-boot will be the, let's say, first implementation :)

Thanks

schaefi:
Thanks for all your input, PR is now open for review

kiwi/bootloader/config/bootloader_spec_base.py
```diff
@@ -0,0 +1,215 @@
...
+    ]
+)
+
+
+class BootLoaderSpecBase(BootLoaderConfigBase):
+    """
+    **Base class for Boot Loader Specification**
+
+    This base class follows the bootloader specifications as
+    described in https://systemd.io/BOOT_LOADER_SPECIFICATION
```
DaanDeMeyer, line 39: You might want to link to https://uapi-group.org/specifications/specs/boot_loader_specification/ instead which is the new location for the boot loader specification

kiwi/bootloader/config/systemd_boot.py
```diff
@@ -0,0 +1,110 @@
...
+            boot_options.get('system_volumes')
+        )
+        BootLoaderSystemdBoot._write_config_file(
+            BootLoaderTemplateSystemdBoot().get_loader_template(),
+            self.root_mount.mountpoint + '/boot/efi/loader/loader.conf',
+            self._get_template_parameters()
+        )
+        Command.run(
+            ['chroot', self.root_mount.mountpoint, 'systemd-machine-id-setup']
+        )
```
DaanDeMeyer, line 57: So in newer versions of systemd, we reworked systemd-boot a bit to not require a machine ID to be present in the final image which you want to avoid since each instance of the image should have its own machine ID (which is usually generated on first boot). So depending on which version of bootctl you want to support, you might want to make use of the new --entry-token option. If you want to support older versions, you should definitely remove the machine ID again after installing systemd-boot so you don't end up shipping the image with a machine ID.

kiwi/bootloader/config/systemd_boot.py
```diff
@@ -0,0 +1,110 @@
...
+                'chroot', self.root_mount.mountpoint,
+                'cp', '-L', '/boot/vmlinuz', '/boot/efi/EFI/Linux'
+            ]
+        )
+        Command.run(
+            [
+                'chroot', self.root_mount.mountpoint,
+                'cp', '-L', '/boot/initrd', '/boot/efi/EFI/Linux'
+            ]
+        )
```
Conan-Kudo, line 79: As we discussed in the meeting today, we should drop this and verify that things work on Fedora (which is known to have working sd-boot support). @DaanDeMeyer is willing to help us out with questions and feedback as we iterate through this.

kiwi/bootloader/config/bootloader_spec_base.py
schaefi: done

kiwi/bootloader/config/systemd_boot.py
schaefi: yep thanks for the clarification. I did all tests with Fedora (37) now as it has proper systemd-boot integration and made use of the --entry-token option which makes a lot sense to me

kiwi/bootloader/config/systemd_boot.py
schaefi: yep, I deleted that code part and replaced it by the use of ```kernel-install```. With Fedora systems we now have a proper  support of systemd-boot.

schaefi:
@Conan-Kudo @DaanDeMeyer ok I have reworked the code and changed according to your suggestions and our conversations. I think it looks much more generic now and also works out of the box with Fedora (37). I'm also using the --entry-token option which makes this implementation to require a fairly new version of systemd-boot. Leap 15.3 as one example does not provide it and would fail. As proper integration of systemd-boot is also not done in Leap I think it should be ok for us to make Fedora the reference system for the support of systemd-boot

Please give it another review shot and I hope we are close to merge :)

Thanks

kiwi/bootloader/config/systemd_boot.py
```diff
@@ -0,0 +1,159 @@
...
+        if not kernel_info:
+            raise KiwiKernelLookupError(
+                'No kernel in image tree found'
+            )
+        Command.run(
+            [
+                'chroot', self.root_mount.mountpoint, 'bootctl', 'install',
+                '--esp-path=/boot/efi',
+                '--no-variables',
+                '--entry-token', 'os-id'
```
DaanDeMeyer, line 89: This should probably be something more descriptive than `os-id` no? Ideally the entry-token name is configurable by the user.

kiwi/bootloader/config/systemd_boot.py
```diff
@@ -0,0 +1,159 @@
...
+            ]
+        )
+        Path.wipe(f'{self.root_mount.mountpoint}/boot/loader')
+        Command.run(
+            [
+                'chroot', self.root_mount.mountpoint,
+                'kernel-install', 'add', kernel_info.version,
+                f'/boot/{kernel_info.name}',
+                f'/boot/initramfs-{kernel_info.version}.img'
```
DaanDeMeyer, line 98: This assumes the kernel and initrd can be found at these locations. Not sure if that's the case for kiwi. In mkosi, we have logic to find the vmlinuz/initrd locations that need to be passed to kernel-install (see https://github.com/systemd/mkosi/blob/1d0492ee5e01bd86916a336c1a2b556c56de0277/mkosi/__init__.py#L3941)

kiwi/bootloader/config/systemd_boot.py
```diff
@@ -0,0 +1,159 @@
...
+            self._get_template_parameters()
+        )
+
+    def set_loader_entry(self, target: str) -> None:
+        """
+        Setup/update loader entries
+
+        loader entries for systemd-boot are expected to be managed
+        through kernel-install. However, custom kiwi settings for the
+        kernel cmdline are not taken into account. Thus this method
```
DaanDeMeyer, line 113: Do you write these custom settings to /etc/cmdline? They'll be taken into account by kernel-install (dracut) if you do that.

DaanDeMeyer (commented):
A few more comments

kiwi/bootloader/config/systemd_boot.py
Conan-Kudo: this should be the case for all bls-based things, so this might make sense in the base class?

kiwi/bootloader/config/systemd_boot.py
Conan-Kudo: I'm pretty sure we have similar logic _somewhere_ else because of Debian and SUSE distributions hotpatching dracut to spit out weird filenames for the images, so we probably should rationalize this with that and provide a utility method to use across everything...

kiwi/bootloader/config/systemd_boot.py
schaefi: according to the documentation ```entry-token os-id``` takes the information from ```/etc/os-release``` which I think is a nice idea. In my tests with Fedora this was also the case and the naming looked good to me. Also as ```/etc/os-release``` is distro specific provided data it felt like a good concept

kiwi/bootloader/config/systemd_boot.py
schaefi: the kiwi process makes sure this data exists. some lines up you see an instance of the ```Kernel``` class from which this information is taken. But there is one point that is correct, the name of the initrd varies and I forgot to take the generic interface for it. kiwi also has code to lookup kernels and initrds ;) I'm going to address the initrd part soon

kiwi/bootloader/config/systemd_boot.py
schaefi: I'm sorry to say that but kernel-install doesn't take them into account correctly. kernel-install at least on Fedora 37 took the host kernel cmdline which is of course wrong. In addition kiwi supports custom kernel options via the ```kernelcmdline``` attribute and those should find its way to the config. So yes we are making sure the passed kernel cmdline is correct and expected

martinezjavier: @schaefi `kernel-install` should take `/etc/kernel/cmdline` into account and that should take precedence over `/proc/cmdline` if is present. That's what the `kerne-install` man page documents and if that's not the case, it's a bug.

In fact, the order should be `/etc/kenel/cmdline`, `/usr/lib/kernel/cmdline`, `/proc/cmdline`.

kiwi/bootloader/config/systemd_boot.py
DaanDeMeyer: Ahh, I forgot we added something like that, makes sense

schaefi: :)

kiwi/bootloader/config/systemd_boot.py
Conan-Kudo: Unless the distribution replaces or otherwise breaks `kernel-install`, `/etc/kernel/cmdline` should work everywhere.

schaefi: :+1: great, I will adapt to this concept and update the PR soon

martinezjavier: yes, what @Conan-Kudo said :)

kiwi/bootloader/config/systemd_boot.py
schaefi: done, using the get_boot_names API

kiwi/bootloader/config/systemd_boot.py
schaefi: ok done, prior kernel-install ```etc/kernel/cmdline``` is written to the system. I can confirm this worked. Thanks

DaanDeMeyer (approved):
Only looked at the systemd-boot/kernel-install stuff, which looks good to me

Conan-Kudo:
@schaefi The only other thing we should do is document that either filesystem drivers need to be installed (e.g. `efifs` or similar) or the ESP needs to be bigger for sd-boot in our guidance.

schaefi:
> @schaefi The only other thing we should do is document that either filesystem drivers need to be installed (e.g. `efifs` or similar) or the ESP needs to be bigger for sd-boot in our guidance.

yes good point, I added some bits to the documentation

I also think we missed an integration image build test, which I now also added to this PR

schaefi:
@Conan-Kudo funny I'm running out on pipeline minutes on gitlab, that's why the test here now fails. What's that ? did I miss a change on the free offer from gitlab or did we hit some limit ? See https://gitlab.com/kiwi3/kiwi-ci/-/pipelines/692260448

Conan-Kudo:
> @Conan-Kudo funny I'm running out on pipeline minutes on gitlab, that's why the test here now fails. What's that ? did I miss a change on the free offer from gitlab or did we hit some limit ? See https://gitlab.com/kiwi3/kiwi-ci/-/pipelines/692260448

We missed a change from GitLab. They now require open source projects to apply for more: https://about.gitlab.com/blog/2021/11/11/public-project-minute-limits/

build-tests/x86/fedora/test-image-systemd-boot/appliance.kiwi
```diff
@@ -0,0 +1,54 @@
...
+    </preferences>
+    <users>
+        <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root" groups="root"/>
+    </users>
+    <repository type="rpm-md">
+        <source path="obsrepositories:/"/>
+    </repository>
+    <packages type="image">
+        <package name="grub2"/>
+        <package name="grubby"/>
```
Conan-Kudo, line 35: These need to be dropped. Obviously we no longer need a grub theme when we're not using grub, either.

Conan-Kudo (changes requested):

build-tests/x86/fedora/test-image-systemd-boot/appliance.kiwi
```diff
@@ -0,0 +1,54 @@
...
+        <package name="glibc-all-langpacks"/>
+        <package name="vim"/>
+        <package name="tzdata"/>
+        <package name="NetworkManager"/>
+    </packages>
+    <packages type="bootstrap">
+        <package name="filesystem"/>
+        <package name="basesystem"/>
+        <package name="grub2-efi-x64-modules"/>
+        <package name="grub2-efi-x64"/>
```
Conan-Kudo, line 50: Ditto here.

Conan-Kudo (changes requested):

Conan-Kudo:
Also note that _currently_ shim doesn't work with sd-boot anywhere, so that needs to be documented.

schaefi:
> > @Conan-Kudo funny I'm running out on pipeline minutes on gitlab, that's why the test here now fails. What's that ? did I miss a change on the free offer from gitlab or did we hit some limit ? See https://gitlab.com/kiwi3/kiwi-ci/-/pipelines/692260448
> 
> We missed a change from GitLab. They now require open source projects to apply for more: https://about.gitlab.com/blog/2021/11/11/public-project-minute-limits/

Thanks for the doc link. Very unfortunate. I disabled gitlab CI now and will move the tests to github actions in followup pull requests (low prio though). We used the gitlab pipeline for fedora mock testing of the package build and also for building the package docs (latex based).

build-tests/x86/fedora/test-image-systemd-boot/appliance.kiwi
schaefi: yeah my bad, also the grub theme is no longer needed. I updated the test description. Thanks

build-tests/x86/fedora/test-image-systemd-boot/appliance.kiwi
schaefi: done

schaefi:
> Also note that _currently_ shim doesn't work with sd-boot anywhere, so that needs to be documented.

yes , added a hint in the docs. I think the way how I described this could be done better. A short review on the doc adaptions would be much appreciated. Thanks

schaefi:
@Conan-Kudo Integration testing of systemd-boot in OBS is hard. I stumbled over the following problems so far:

* bootctl --entry-token option is available with Fedora >= 37, but OBS has not Fedora 37 imported
* On Rawhide which is Fedora 38 atm ```kernel-install``` failed with exit 1 not providing any error message
* On TW bootctl failed with ```Failed to open file system "/dev/block/254:1": No such file or directory``` I assume an issue with OBS workers (no udev there), because local build just works

:-1:

Conan-Kudo:
What, Fedora 37 is imported in OBS: https://build.opensuse.org/project/show/Fedora:37

Conan-Kudo (approved):

PR #2069, opened 2022-02-14
Follow up fix on force deleting debs

schaefi:
Also remove eventual post scripting prior force removal
of deb packages. Similar inconsistencies as with the pre
scripts can occur on force removal. We want the operation
to be successful in force mode even if that means to
leave a dirty state.


Conan-Kudo (approved):
Looks fine to me... 🤷🏾‍♂️

PR #2071, opened 2022-02-18
Use DEB822-formatted .sources files instead .list files for APT

### LINKED ISSUE #2023: switch to the DEB822 format for APT sources.list.d files
## Problem description
Debian has deprecated APT's original sources.list file format. As of Debian 11 (and Ubuntu 20.10), APT uses [the newer "DEB822" format](https://manpages.debian.org/unstable/apt/sources.list.5.en.html#DEB822-STYLE_FORMAT) by default. This format has been supported since APT 1.1, so it is compatible with all versions of Debian (and Debian-derived distributions) that Kiwi supports.

## Expected behaviour
Kiwi creates `.sources` files using the DEB822 format instead of the deprecated `.list` format.

1. build a Debian (or Debian-derived) image with Kiwi
2. examine the files within /etc/apt/sources.list.d in the image
3. observe that they're all `.list` files using the deprecated format

+ KIWI version: 9.24.14
+ Operating system: Ubuntu 20.04
+ OBS version: n/a

iammattcoleman:
Fixes #2023 

Debian repository configuration files will be created in the DEB822 .sources file format instead of the now-deprecated .list format.

### DIFF
--- a/kiwi/repository/apt.py
+++ b/kiwi/repository/apt.py
@@ -159,13 +159,13 @@ class RepositoryApt(RepositoryBase):
         :param str customization_script:
             custom script called after the repo file was created
         """
-        list_file = '/'.join(
-            [self.shared_apt_get_dir['sources-dir'], name + '.list']
+        sources_file = '/'.join(
+            [self.shared_apt_get_dir['sources-dir'], name + '.sources']
         )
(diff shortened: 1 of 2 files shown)


schaefi (approved):
Thanks Matt, this is great :+1:

PR #2078, opened 2022-03-01
Fix disk script caller environment

schaefi:
This patch has three parts

**Fixed disk.sh caller environment**

The documentation explains the disk.sh script to be called from inside of the image root as it exists on the block layer. The disk.sh script is therefore also called after the sync of the unpacked image root tree to the block layer. The implementation however, was only partially calling disk.sh from such an environment. In fact the environment was only the mountpoint of the root partition but this is not the complete system regarding layouts that uses extra partitions and/or volumes. This commit introduces the use of the new class ImageSystem and calls disk.sh in the way it was designed and documented.

**Added ImageSystem class**
    
The class responsibility is to provide access to the image root system from the block layer of the image scope

**Prevent superfluous filesystem creation***
    
In case of an overlayroot setup and the request for no extra write partition, it is not needed to create a filesystem for the write space which never gets synced to the image. I found this bug when writing the ImageSystem class and as it has a dependency to the unit test code I put it as part of the PR but as an extra commit

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -25,6 +25,7 @@ from typing import (
 import kiwi.defaults as defaults
 
 from kiwi.utils.temporary import Temporary
+from kiwi.system.mount import ImageSystem
 from kiwi.storage.disk import ptable_entry_type
 from kiwi.defaults import Defaults
 from kiwi.filesystem.base import FileSystemBase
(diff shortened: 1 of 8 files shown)


kiwi/builder/disk.py
```diff
@@ -479,12 +482,18 @@ def create_disk(self) -> Result:
         if self.system_setup.script_exists(
             defaults.POST_DISK_SYNC_SCRIPT
         ):
+            image_system = ImageSystem(
+                device_map, self.root_dir,
+                system.get_volumes() if self.volume_manager_name else {}
+            )
+            image_system.mount()
```
davidcassany, line 489: I see this mounts the root device. I am wondering where it was mounted before this change.

davidcassany (approved):
Looks good to me.

I just could not easily see how the ImageSystem is mounting the root device and where it was mounted before, I expected to see some mounts _moved_ into ImageSystem, however they all look like new mounts to me. I guess I am missing some parts of the implementation. I did not really tried to completely follow former implementation of the disk.sh call.

kiwi/builder/disk.py
schaefi: yeah this was one of the reasons why it's now imho nicer. Before this change it was this part from the diff

```
             disk_system = SystemSetup(
-                self.xml_state, system.get_mountpoint()
+                self.xml_state, image_system.mountpoint()
             )
```

so the mountpoint was only ```system.get_mountpoint()``` which only points to what the root partition is and not more. At call time this partition was mounted due to the sync action which is encoded in the ```_sync_system_to_image``` method that runs before

PR #2079, opened 2022-03-01
Added overlayroot_verity_blocks attribute

schaefi:
Given the following type setup

```xml
<type image="oem" overlayroot="true" overlayroot_verity_blocks="10" .../>
```

Setting the ```overlayroot_verity_blocks``` attribute to a number (or ```alll```) in an overlayroot configuration will create a dm verity hash from the number of given blocks (or all) placed at the end of the squashfs compressed read-only root filesystem. For later verification of the device, and without further image description settings, the credentials information produced by veritysetup from the cryptsetup tools, is created as a file in ```/boot/overlayroot.verity``` and is stored as such into the image by default.

The default behavior is considered insecure as it places the veritysetup credentials information to verify the device as
human readable file to /boot as shown above. How the root hash gets secured is in the responsibility of the author of the image description. One way to deal with the credentials is as part of a ```disk.sh``` script like in the following example:

```bash
if [ -e /boot/overlayroot.verity ];then
    rm -f /boot/overlayroot.verity
fi
```

Of course this is also not recommended as it will delete the credentials information forever. So this should just give an idea which hook in the image creation process can be used to handle the file

For completeness verification of the device can be done in the following way:

```bash
kpartx -a image-file.raw
veritysetup verify /dev/mapper/loop0pX /dev/mapper/loop0pX "root_hash_from_credentials_file" \
    --debug \
    --no-superblock \
    --salt "salt_hash_from_credentials_file" \
    --data-blocks 10 \
    --hash-offset "offset_from_credentials_file"
```

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -572,6 +572,51 @@ overlayroot="true|false"
   will not have any effect because the write partition will be
   resized on first boot to the available disk space.
 
+overlayroot_verity_blocks="number|all"
+  For the `oem` type only, specifies to create a dm verity hash
+  from the number of given blocks (or all) placed at the end of the squashfs
+  compressed read-only root filesystem. For later verification of the device,
+  the credentials information produced by `veritysetup` from the
+  cryptsetup tools, is created as a file in `/boot/overlayroot.verity`
+  and is stored as such into the image by default. The verification
+  file contents looks like the following example:
+
+  .. code:: bash
+
+     Hash type:         1
+     Data blocks:       10
+     Data block size:   4096
+     Hash block size:   4096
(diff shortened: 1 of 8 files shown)


schaefi:
Maybe an open discussion point. As you can see the number of blocks to verify is currently a fixed value in the kiwi defaults area. At the moment 10 blocks from the data area are considered enough to verify a read-only squashfs filesystem. One question that came to my mind was if this value should be configurable. On the other hand I believe if the integrity of the first 10 blocks is violated I would say the entire squashfs is broken. So checking on the first 10 blocks is imho enough. The footprint to allow verification of all blocks of the filesystem will also almost double the size

Feedback welcome

Thanks

PR #2080, opened 2022-03-03
Fixed unconditional grub package requirement

schaefi:
The grub2 package should only be required on 32/64bit
x86 architectures. Building on arm for Debian based
distros causes a dependency issue because there is no
grub package available

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -187,7 +187,9 @@ Recommends:     gfxboot
 Requires:       grub2-efi-x64
 %endif
 %endif
+%if ! (0%{?debian} || 0%{?ubuntu})
 Requires:       grub2
+%endif
 %ifarch %arm aarch64
 %if 0%{?fedora} || 0%{?rhel}
 Requires:       uboot-tools


package/python-kiwi-spec-template
```diff
@@ -187,7 +187,9 @@ Recommends:     gfxboot
 Requires:       grub2-efi-x64
 %endif
 %endif
+%ifarch %{ix86} x86_64
 Requires:       grub2
+%endif
```
Conan-Kudo, line 192: This breaks it for non-x86 RPM distributions, who have their non-x86 variants of GRUB packages provide the `grub2` name.

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: ok, so I changed it to have this requirement only for the distros that uses this concept. How about it now ?

package/python-kiwi-spec-template
```diff
@@ -187,7 +187,9 @@ Recommends:     gfxboot
 Requires:       grub2-efi-x64
 %endif
 %endif
+%if 0%{?fedora} || 0%{?rhel} || 0%{?suse_version}
```
Conan-Kudo, line 190:
```suggestion
%if ! (0%{?debian} || 0%{?ubuntu})
```

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: yes that's a better read, thanks I adapted to it

Conan-Kudo (approved):

PR #2081, opened 2022-03-05
Support extended partitions

schaefi:
Add support for extended layout to msdos table
    
This commit adds the following new type attribute
    
```xml
<type ... dosparttable_extended_layout="true|false"/>
```
    
If set it specifies to make use of logical partitions inside of an extended one. Effective only on type configurations which uses the msdos table type, it will cause the fourth partition to be an extended partition and all following partitions will be placed as logical partitions inside of that extended partition. This setting is useful if more than 4 partitions needs to be created in a msdos table. In addition to the support for extended/logical partitions the the attributes ```mountpoint``` and ```filesystem``` in the ```<partitions>```  section becomes optional. This also allows to place partitions as placeholders not mounted into the system

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -394,6 +394,15 @@ efiparttable="msdos|gpt":
   table type to use. If not set defaults to the GPT partition
   table type
 
+dosparttable_extended_layout="true|false":
+  For oem disk images, specifies to make use of logical partitions
+  inside of an extended one. If set to true and if the msdos table type
+  is active, this will cause the fourth partition to be an
+  extended partition and all following partitions will be
+  placed as logical partitions inside of that extended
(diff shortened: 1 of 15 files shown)


schaefi:
Example type setup:

```xml
<type image="oem" filesystem="ext4" firmware="bios" dosparttable_extended_layout="true">
    <partitions>
        <partition name="data1" size="100" mountpoint="/data1" filesystem="ext3"/>
        <partition name="data2" size="100" mountpoint="/data2" filesystem="ext3"/>
        <partition name="data3" size="100"/>
        <partition name="data4" size="100" mountpoint="/data4" filesystem="ext3"/>
        <partition name="data5" size="100" mountpoint="/data5" filesystem="ext3"/>
    </partitions>          
</type>
```

Conan-Kudo (approved):

PR #2082, opened 2022-03-06
Support mount by PARTUUID

schaefi:
In addition to by-label and by-uuid also support mounting
by PARTUUID. Please note kiwi also makes sure that the grub
generated config file uses the root=PARTUUID= notation and it's
not clear if grub-mkconfig will persist making use of it.
Nevertheless there are also systems which uses different
methods to boot and it makes sense to support partuuid
mappings as well

### DIFF
--- a/build-tests/x86/leap/kiwi-settings/kiwi-settings.spec
+++ b/build-tests/x86/leap/kiwi-settings/kiwi-settings.spec
@@ -1,0 +1,28 @@
+Name:           kiwi-settings
+Version:        1.1.1
+Release:        0
+License:        GPL-3.0-or-later
+%if "%{_vendor}" == "debbuild"
+Packager:       Marcus Schaefer <marcus.schaefer@gmail.com>
+%endif
+Summary:        KIWI - runtime config file
+Group:          System/Management
+Source:         %{name}-%{version}.tar.gz
+BuildRoot:      %{_tmppath}/%{name}-%{version}-build
+BuildArch:      noarch
+
+%description
+Provides a KIWI runtime config file suitable building
+the leap based integrations test
+
+%prep
+%setup -q
+
+%install
+install -D -m 644 kiwi-settings/kiwi.yml %{buildroot}/etc/kiwi.yml
+
+%files
+%defattr(-,root,root)
+%config /etc/kiwi.yml
+
+%changelog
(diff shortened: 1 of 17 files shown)


dcermak:
According to the grub docs:
```texi
@item GRUB_DISABLE_LINUX_PARTUUID
If @command{grub-mkconfig} cannot identify the root filesystem via its
universally-unique indentifier (UUID), @command{grub-mkconfig} can use the UUID
of the partition containing the filesystem to identify the root filesystem to
the Linux kernel via a @samp{root=PARTUUID=...} kernel parameter.  This is not
as reliable as using the filesystem UUID, but is more reliable than using the
Linux device names. When @samp{GRUB_DISABLE_LINUX_PARTUUID} is set to
@samp{false}, the Linux kernel version must be 2.6.37 (3.10 for systems using
the MSDOS partition scheme) or newer.  This option defaults to @samp{true}.  To
enable the use of partition UUIDs, set this option to @samp{false}.
```

So it looks like you probably want to set `GRUB_DISABLE_LINUX_PARTUUID` to false and then it *should* work :crossed_fingers:. But that sounds like something which we should probably create an integration test for :thinking: ?

schaefi:
@dcermak very good point, thanks. I added a commit to address this

schaefi:
+1 for the integration test. I'll add, not sure when I get to it, thus keeping this one open as a WIP

davidcassany (approved):
Looks good to me, I agree there should be a simple integration test for that. Gurb features/changes are hard to track, so if we have a tests there are greater chances that it does not silently break over time.

Conan-Kudo (approved):
LGTM, but we need an integration test for this.

schaefi:
> LGTM, but we need an integration test for this.

https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:leap/test-image-embedded

schaefi:
I placed the test as a profile into the existing embedded testing. I plan to have more features more or less only needed in that space to be tested there

schaefi:
I added a change needed on the dracut module code to support partuuid references. Going to test the integration test and respond back once it all works. Until then, please don't merge, label is set

schaefi:
During testing I found that by-partuuid should be limited to partition tables that actually supports UUIDs natively. I don't want to trust the UUIDs created by Linux from the disk signature. This value can change if you just re-install MBR code (grub2-install) and we don't want to base the boot of a system on a value that can change without even having touched the partition table itself :) Thus I added a runtime check for it

schaefi:
@dcermak The integration is ready and should also be usable within openQA. I tested it locally in QEMU, would be great if you can give it a try too:

* https://build.opensuse.org/package/binaries/Virtualization:Appliances:Images:Testing_x86:leap/test-image-embedded:SystemFeatures/images

after booting you should see by-partuuid device names used in e.g ```/etc/fstab```

Thanks

PR #2084, opened 2022-03-07
Support additional names for docker containers

### LINKED ISSUE #2045: additionaltags for container should allow different container name
Currently, when building container, we can only add different tag with the same containername, which prevent easy double publishing of a container.

For instance:
```xml
<preferences>
    <type image="docker">
      <containerconfig
        name="foo"
        tag="V1"
        additionaltags="V2"
```
will returns in the additional tags added to the container to be:
`foo:V2`

but if we want to ship foo container as "bar" container, we can't attach all the tags specified in the additionaltags.

We are relying on `<!-- OBS-AddTag: -->` but it isn't enough as it only propagate all additionaltags to it.

davidcassany:
Docker containers used to support the attribute `additionaltags` which was used to provide multiple tags for the same image. Since only tags were supported this commit renames the attribute to `additionalnames` and now supports tags and names witht he following syntax:

* `<name>:<tag>` -> adds a full docker image reference including name and tag
* `:<tag>` -> adds an additional tag while reusing the former name
* `<name>` -> adds an additional name while reusing the former tag

Fixes #2045

### DIFF
--- a/build-tests/arm/fedora/test-image-live/appliance.kiwi
+++ b/build-tests/arm/fedora/test-image-live/appliance.kiwi
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<image schemaversion="7.4" name="kiwi-test-image-live">
+<image schemaversion="7.5" name="kiwi-test-image-live">
     <description type="system">
         <author>Marcus Schaefer</author>
         <contact>ms@suse.com</contact>
--- a/build-tests/arm/rawhide/test-image-live-disk/appliance.kiwi
+++ b/build-tests/arm/rawhide/test-image-live-disk/appliance.kiwi
@@ -2,7 +2,7 @@
 
 <!-- OBS-Profiles: @BUILD_FLAVOR@ -->
 
-<image schemaversion="7.4" name="kiwi-test-image-live-disk">
+<image schemaversion="7.5" name="kiwi-test-image-live-disk">
     <description type="system">
         <author>Marcus Schaefer</author>
         <contact>ms@suse.com</contact>
--- a/build-tests/arm/tumbleweed/test-image-live-disk/appliance.kiwi
+++ b/build-tests/arm/tumbleweed/test-image-live-disk/appliance.kiwi
@@ -2,7 +2,7 @@
 
 <!-- OBS-Profiles: @BUILD_FLAVOR@ -->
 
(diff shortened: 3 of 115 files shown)


davidcassany:
Marked as a draft as I want to revalidate the schema changes. I am new to XSLT and the schema change introduced here is a bit tricky.

I fact this is why it took me a month to run this PR, the python code I had since few weeks but it took me a while to find the time to try to learn/understand the required XSLT converter.

schaefi:
@davidcassany Hi David I hope you don't mind I applied the new schema to the XML descriptions we have in the kiwi project. Looks good to me. Also the integration test was updated which uses additionaltags and now additionalnames

I also added a commit to update the provides tag in the spec file

From my perspective it looks good to become a PR for review. Thanks much for your effort. I can imagine that XSLT kungfu was not easy :)

schaefi (approved):

davidcassany:
@schaefi @dcermak I'd appreciate if you can have a look to the updated docs

doc/source/building_images/build_docker_container.rst
```diff
@@ -33,6 +33,12 @@ specified:
   `subcommand`. Equivalent of the `USER` directive of a Docker file.
 * ``workingdir``: Sets the working directory to be used when running `cmd` and
   `entrypoint`. Equivalent of the `WORKDIR` directive of a Docker file.
+* ``additionalnames``: Allows to specifiy additional image references.
+  References are passed as comma separated values and follow the 
+  common docker registry format `<repository>:<tag>`. Either repository or
+  tag can be missing (e.g. `myalternaterepo:` or `:sometag`), in that case the
+  values from `name` and `tag` attributes will used to complete the image
+  reference.
```
dcermak, line 41:
```suggestion
* ``additionalnames``: Allows add additional image references (names and tags).
  References are passed as comma separated values and follow the 
  common docker registry format `<repository>:<tag>`. Either the repository or
  the tag can be missing (e.g. `myalternaterepo:` or `:sometag`), in that case the
  values from `name` and `tag` attributes will used to complete the image
  reference.
  For instance ``<containerconfig name="python" tag="3.10" additionalnames="python:3,:latest,py:"/>``
  would yield the references: ``python:3.10``, ``python:3``, ``python:latest`` and ``py:3.10``.
```

dcermak (approved):
LGTM, thanks!

PR #2085, opened 2022-03-09
Style changes in container docs

dcermak:
A few style changes that I would like to apply to the container guide after taking a look at it while reviewing #2084

### DIFF
--- a/doc/source/building_images/build_container_image.rst
+++ b/doc/source/building_images/build_container_image.rst
@@ -66,7 +66,7 @@ openSUSE Leap:
 
       $ zypper addrepo http://download.opensuse.org/repositories/Virtualization:/containers/<DIST> container-tools
 
-   where the placeholder `<DIST>` is the preferred distribution. 
+   where the placeholder `<DIST>` is the preferred distribution.
 
 #. Install :command:`umoci` and :command:`skopeo` tools
 
--- a/doc/source/building_images/build_container_image.rst
+++ b/doc/source/building_images/build_container_image.rst
@@ -28,31 +28,31 @@ specified:
 * ``tag``: Sets the tag of the Container Image.
 * ``user``: Sets the user name or user id (UID) to be used when
   running `entrypoint` and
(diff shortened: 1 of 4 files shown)


doc/source/building_images/build_container_image.rst
```diff
@@ -0,0 +1,103 @@
...
+
+* ``subcommand``: Provides the default execution parameters of the
+  container. Equivalent of the `CMD` directive in a :file:`Dockerfile`.
+* ``labels``: Adds custom metadata to an image using key-value pairs.
+  Equivalent to one or more `LABEL` directives in a :file:`Dockerfile`.
+* ``expose``: Define which ports can be exposed to the outside when
+  running this container image.  Equivalent to one or more `EXPOSE`
+  directives in a :file:`Dockerfile`.
+* ``environment``: Sets environment variables using key-value pairs.
+  Equivalent to one or more the `ENV` directives of a
```
schaefi, line 47: I think: "one or more of the ..." was meant

doc/source/building_images/build_container_image.rst
```diff
@@ -0,0 +1,103 @@
...
+  :file:`Dockerfile`.
+* ``entrypoint``: Sets the binary via which all commands inside the
```
schaefi, line 49: via which -> from which ?

schaefi (approved):
Thanks looks good :+1:

doc/source/building_images/build_container_image.rst
dcermak: Not really, because whatever is written in `entrypoint` calls your supplied parameters, so I don't think that "from which" is a good fit here.

doc/source/building_images/build_container_image.rst
dcermak: Good catch! Should be fixed now

doc/source/building_images/build_container_image.rst
schaefi: ok

PR #2086, opened 2022-03-09
Move scope of veritysetup

schaefi:
veritysetup was called as part of the disk builder.
However the veritysetup should be a responsibility of
the classes which implements the sync_data method.
This allows to use the creation of a verity hash right
after sync as a feature to these classes and in a broader
scope


schaefi:
@davidcassany this PR is just moving the verity feature to exist at a better place in the API. If you can take a real quick look for obvious mistakes it would be great. Thanks much

schaefi:
For the background, dm_verity is a method often used on embedded devices to run block level integrity checks. It actually makes more sense in combinaton with dm_verity on top of dm_crypt. As we support dm_crypt (through luks) the integrity layer is a new feature we are currently missing

PR #2089, opened 2022-03-10
Add support for custom size in filesystem classes

schaefi:
Allow to create filesystems with an optional size parameter.
If no size is provided the filesystem gets as big as the device
which is the default and unchanged behavior. In addition a
size counting from the beginning (>0) as well as a size
counting from the end (<=0) can be provided.

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -35,7 +35,32 @@ from kiwi.version import (
 
 from kiwi.exceptions import KiwiBootLoaderGrubDataError
 
+shim_loader_type = NamedTuple(
+    'shim_loader_type', [
+        ('filename', str),
+        ('binaryname', str)
+    ]
+)
+
+grub_loader_type = NamedTuple(
+    'grub_loader_type', [
+        ('filename', str),
+        ('binaryname', str)
+    ]
+)
+
+unit_type = NamedTuple(
+    'unit_type', [
+        ('byte', str),
+        ('kb', str),
+        ('mb', str),
+        ('gb', str)
+    ]
+)
+
+
 # Default module variables
+UNIT = unit_type(byte='b', kb='k', mb='m', gb='g')
 POST_DISK_SYNC_SCRIPT = 'disk.sh'
 PRE_DISK_SYNC_SCRIPT = 'pre_disk_sync.sh'
 POST_BOOTSTRAP_SCRIPT = 'post_bootstrap.sh'
@@ -51,20 +76,6 @@ CUSTOM_RUNTIME_CONFIG_FILE = None
 PLATFORM_MACHINE = platform.machine()
 
 log = logging.getLogger('kiwi')
-
-shim_loader_type = NamedTuple(
-    'shim_loader_type', [
-        ('filename', str),
-        ('binaryname', str)
-    ]
-)
-
-grub_loader_type = NamedTuple(
-    'grub_loader_type', [
-        ('filename', str),
-        ('binaryname', str)
-    ]
-)
 
 
 class Defaults:
--- a/kiwi/filesystem/base.py
+++ b/kiwi/filesystem/base.py
@@ -23,6 +23,8 @@ from typing import (
 )
 
 # project
+import kiwi.defaults as defaults
+
 from kiwi.defaults import Defaults
 from kiwi.utils.sync import DataSync
 from kiwi.mount_manager import MountManager
@@ -107,14 +109,21 @@ class FileSystemBase:
         if not self.custom_args.get('fs_attributes'):
             self.custom_args['fs_attributes'] = []
 
-    def create_on_device(self, label: str = None):
+    def create_on_device(
+        self, label: str = None, size: int = 0, unit: str = defaults.UNIT.kb
+    ):
         """
         Create filesystem on block device
 
         Implement in specialized filesystem class for filesystems which
         requires a block device for creation, e.g ext4.
 
-        :param string label: label name
+        :param str label: label name
+        :param int size:
+            size value, can also be counted from the end via -X
+            The value is interpreted in units of: unit
+        :param str unit:
+            unit name. Default unit is set to: defaults.UNIT.kb
         """
         raise NotImplementedError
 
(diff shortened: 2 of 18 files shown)


kiwi/filesystem/ext3.py
```diff
@@ -24,16 +24,20 @@ class FileSystemExt3(FileSystemBase):
...
+        :param int kb_size:
+            size in KB, can also be counted from the end via -KB
         """
-        device = self.device_provider.get_device()
+        device = [self.device_provider.get_device()]
         if label:
             self.custom_args['create_options'].append('-L')
             self.custom_args['create_options'].append(label)
+        if kb_size:
+            device.append(self._fs_size(kb_size))
```
davidcassany, line 40: nit, feel free to ignore it. I am not sure `device` is the best variable name here, that forced me to read and check  `mkfs.ext*` man pages 😅  feels like having multiple devices :)

davidcassany (approved):
Looks good to me :+1:
Curious about how or why this is needed or useful, though. Feels weird formatting a device with a different size of the partition. Probably make sense for non partition devices? In any case I like this extra little flexibility, from an API PoV this is a good move.

schaefi:
> Curious about how or why this is needed or useful...

yeah I was asking this myself. In the world of embedded things like geometry changes or dynamic size allocations doesn't exist. The OS and the app are flashed on a storage device of a fixed size and this never going to be changed. Because of this reason it is common practice to append raw bytes of a certain structure at the end of a filesystem to e.g store verification keys, signatures and other customer specific information. This means you often see a layout like this

```
|filesystem|0xff|binary_blob|
|-----------partition-size-------|
```

In kiwi there is currently no way to construct something like this. Having the opportunity to create filesystems smaller than
the actual partition is a first step. Also there is some proprietary bull/zensored/ which I will never touch but the appliance builder should be flexible enough to at least create the foundation to perform all needed actions

I also learned that images for the automotive are not delivered as plug&play because the individual car companies add changes on top of the images. So they modify stuff and that requires the layout to be prepared for the modifications. So my plan with all changes that can go into kiwi is to make it a good player in the field of embedded. Usually you see yocto there everywhere and the wic image builder which is more a python library and each individual group or team writes additional (they call it plugin) code to serve the use cases. This process has several advantages (you can do whatever you want in code) but it also has several disadvantages regarding maintenance, reproducability, testing. I think for those companies who likes to deliver a distribution based on packages and not "compile-to-image" kiwi could also be useful.

So my personal goal is a peaceful co-existence with the "compile-to-image" frameworks and the request for "deliver-an-embedded-distro". But I also think the kiwi use case in embedded is only relevant if there is a distro used as foundation. I think this is a very rare case because most of those companies just takes source code, compile it, massage it and that's what they consider a delivery. Packages, dependencies, organizing software in intermediate formats (packages) is a very new thing :) I can imagine that this is also one of the reasons why you mostly see Debian in this world. The rules and regulations of distributors seems to not really fit to how embedded companies function. But there are also projects e.g on the Fedora side to solve this issues.

Let's see what the future brings

kiwi/filesystem/ext3.py
schaefi: yeah I think part of the problem is that mkfs.ext expects a custom size as a positional parameter. The tooling does not provide an option name or something:

```
mke2fs ... device [ fs-size ]
```

This is how the man page describes it. From the above code it imho becomes clear if one knows how mkfs.extX expects the options.

Do you have an idea how to change it to become a better read ? For me it was so obvious that I'm just blind and can't think of a better alternative :-))

davidcassany:
> Usually you see yocto there everywhere and the wic image builder which is more a python library and each individual group or team writes additional (they call it plugin) code to serve the use cases. This process has several advantages (you can do whatever you want in code) but it also has several disadvantages regarding maintenance, reproducability, testing.

Yes, I have been there (at 2012 it was already like this), pulling alpine packages with a damn big mixture of code compiled from the source including kernel from the board vendor plus some kernel modules form the SOC vendor and then putting everything together was a pain.

I also do believe distros have place in embedded world. However I can see the complexity of fitting the old enterprise distro workflow (such as SLE or RedHat) into the customization of low level and sensitive bits that are required in embedded world. Also, one of the problems of embedded is the lack of good development infra, cross-compiling is horrible and makes everything way more complicated.

> Packages, dependencies, organizing software in intermediate formats (packages) is a very new thing :)

Well yes, it is hard to find good tools to manage packaging at low scale. Last time I downloaded an SRPM (years ago) and tried to build it directly on my shell by running `rpmbuild` I miserably failed 😅, too many things where missing and it was hard to create the build env from scratch for a package that was meant to be built in OBS like services.

kiwi/filesystem/ext3.py
davidcassany: I don't know honestly, but probably `device_args` or even simply `args`. My point is that I had to reread `_fs_size` method to check if it was returning some sort of device object, also the fact that `device` is a list did not help. I got confused by reading these lines as I did not notice at a glance it was just collecting a list of command arguments.

schaefi: ok, thanks yes let me change that

schaefi: @davidcassany I pushed a commit to address your issue. I agree it really was not obvious how the device variable was used. Thanks

kiwi/filesystem/base.py
```diff
@@ -106,14 +106,16 @@ def post_init(self, custom_args: Dict):
         if not self.custom_args.get('fs_attributes'):
             self.custom_args['fs_attributes'] = []
 
-    def create_on_device(self, label: str = None):
+    def create_on_device(self, label: str = None, kb_size: int = 0):
```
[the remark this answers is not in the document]
schaefi: ok let's add a bit more comfort, done now

Conan-Kudo (changes requested):

kiwi/defaults.py
```diff
@@ -35,7 +35,31 @@
...
+        ('filename', str),
+        ('binaryname', str)
+    ]
+)
+
+unit_type = NamedTuple(
+    'unit_type', [
+        ('byte', str),
+        ('kb', str),
+        ('mb', str)
```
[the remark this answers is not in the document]
schaefi: yes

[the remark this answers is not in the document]
schaefi: done

Issue #2020, opened 2022-01-09
Building container images from within Mock(chroot) does not work


ignatenkobrain:
## Problem description

As part of the Koji adoption, I've been trying also to build container images, but that did not go smooth.

```
DEBUG util.py:444:  [ ERROR   ]: 01:31:29 | KiwiCommandError: buildah: stderr: Getting image source signatures
DEBUG util.py:444:  Copying blob sha256:27c3f988a847f29d7b93105f07994c4cc9d73b304047aadc98b628d2c7b3948f
DEBUG util.py:444:  error committing container "kiwi-container-hwwgr4" to "kiwi-image-8hwgf6:tag-o35idx": error copying layers and metadata for container "d44cd072c57ee0b0685995b7e39bdd29681fd5bf2c8481450fbdd994cc642ed7": writing blob: adding layer with blob "sha256:27c3f988a847f29d7b93105f07994c4cc9d73b304047aadc98b628d2c7b3948f": Error processing tar file(exit status 1): remount /, flags: 0x44000: invalid argument
DEBUG util.py:444:  , stdout: (no output on stdout)
DEBUG util.py:444:  Exception ignored in: <function OCIBuildah.__del__ at 0x7fecbb9d74c0>
DEBUG util.py:444:  Traceback (most recent call last):
DEBUG util.py:444:    File "/usr/lib/python3.9/site-packages/kiwi/oci_tools/buildah.py", line 327, in __del__
DEBUG util.py:444:      Command.run(['buildah', 'rmi', self.working_image])
DEBUG util.py:444:    File "/usr/lib/python3.9/site-packages/kiwi/command.py", line 133, in run
DEBUG util.py:444:      raise KiwiCommandError(
DEBUG util.py:444:  kiwi.exceptions.KiwiCommandError: buildah: stderr: 1 error occurred:
DEBUG util.py:444:  	* kiwi-image-8hwgf6:tag-o35idx: image not known
DEBUG util.py:444:  , stdout: (no output on stdout)
DEBUG util.py:598:  Child return code was: 1
```

## Expected behaviour

Build works.

## Steps to reproduce the behaviour

Build some simple container image (of CentOS Stream 9) within `mock` (which will setup chroot) on CentOS Stream 8.

## OS and Software information
+ KIWI version: 9.24.15
+ Operating system: CentOS Stream 8 (host), CentOS Stream 9 (target, in chroot)
+ OBS version: N/A

ignatenkobrain:
[...]

PR #2090, opened 2022-03-12
Added container build test for Fedora

schaefi:
Fedora systems uses buildah to create containers. There is
no integration test for kiwi which tests building containers
with buildah. This commit adds a build test to cover this
path. Related to Issue #2020

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,7 +52,6 @@ var/
 #  Usually these files are written by a python script from a template
 #  before PyInstaller builds the exe, so as to inject date/other infos into it.
 *.manifest
-*.spec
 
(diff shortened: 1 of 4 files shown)


schaefi:
See the integration test for this PR running here

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:fedora/test-image-docker

schaefi:
The buildah based build succeeded for me, so probably we have an issue when Mock comes into play

looks like @davidcassany was correct with his assumption that the Mock created environment is not suitable for buildah to work

Conan-Kudo (approved):

PR #2094, opened 2022-03-17
Allow Btrfs and XFS as options for the boot partition filesystem

Conan-Kudo:
We already do this implicitly when we do not define this attribute
and request a boot partition, so let us explicitly offer these as
options too.

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -614,7 +614,7 @@ overlayroot_readonly_partsize="mbsize"
   If the value can be provided beforehand this also speeds
   up the build process significantly
 
-bootfilesystem="ext2|ext3|ext4|fat32|fat16":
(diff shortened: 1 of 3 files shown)


schaefi:
I think xfs and btrfs were excluded because bootloaders back in time were not able to read from it. This was actually the reason for people to create an extra boot partition of another fs understood by the bootloader. From todays perspective at least grub can boot from xfs and btrfs so yes let's add them to the matrix

Thanks

schaefi:
@Conan-Kudo can you please also update the documentation:

* https://osinside.github.io/kiwi/image_description/elements.html?highlight=bootfilesystem

Thanks

schaefi (dismissed):
Looks good to me, but is missing a doc update

Conan-Kudo:
@schaefi done.

schaefi (approved):
:+1:

Issue #2093, opened 2022-03-17
Excluding /run from images breaks Ceph deployment in Rook-based k8s environments


tserong:
## Problem description

Prior to #1759 being merged, `/run` and `/tmp` were sync'd from the build root to the image. This notably meant that `/run/*` _subdirectories_ created by RPM packages as part of their postinstall scripts were also included in the built image. This is important for Ceph when deployed via Rook in Kubernetes environments (https://github.com/rook/rook/blob/master/design/ceph/ceph-config-updates.md#minimal-command-line-options points out that "...it is [...] desirable to use the /var/run/ceph path in containers"). TL;DR: without `/run/ceph` existing inside Ceph container images, those container images are unusable in a k8s/rook environment.

I do understand that we don't want weird build-time artifacts landing in container images, but given that removing everything under `/run` _also_ removes directories that might be required by software installed in the container, can this blanket exclusion be reverted? Or can we somehow add an override for directories we know we want to keep?

## Expected behaviour

Subdirectories of `/run` that were created as a result of `systemd-tmpfiles --create` when packages were installed are present in container images built with kiwi.

## Steps to reproduce the behaviour

Pull a ceph container image built with kiwi >= v9.23.21 and observe that the `/run` directory is practically empty:

```
> podman pull registry.opensuse.org/filesystems/ceph/images/opensuse/ceph/ceph
> podman run -it registry.opensuse.org/filesystems/ceph/images/opensuse/ceph/ceph /bin/bash
# ls /run
.containerenv
```

Prior to the exclusion of `/run`, you'd see something more like this:
```
# ls /run
.containerenv ceph cryptsetup faillock lock log lsm lvm motd.d nfs regenerate-initrd systemd user zabbix zypp.pid
```

Obviously having `zypp.pid` and `regenerate-initrd` in there aren't great, and I'm not sure where `nfs` came from, but all the rest (`ceph`, `cryptsetup`, `faillock`, `lock`, `log`, `lsm`, `lvm`, `motd.d`, `systemd`, `user`, and `zabbix`) are put there as a result of packages being installed in the image, so presumably should remain.

## OS and Software information
+ KIWI version: 9.24.16
+ Operating system: SLE 15 SP2
+ OBS version: Whatever is currently running on build.suse.de.

Vogtinator:
Packages shouldn't put any content into `/run` or `/tmp` themselves, it wouldn't be visible in booting systems anyway.

[...]

Issue #2067, opened 2022-02-13
build stops with "useradd: group 'foo' does not exist"


nexryai:
## Problem description
Build aborted with error "KiwiCommandError: chroot: stderr: useradd: group 'hoge' does not exist". Here is the log.

```
INFO: 20:50:13 | Copying user defined files to image tree
DEBUG: 20:50:13 | EXEC: [rsync -r -p -t -D -H -X -A --one-file-system --links /run/media/suser/bb3e71a7-9e71-450b-9b72-9056e61d78f4/alexandrite/tmp/config/root/ /run/media/suser/bb3e71a7-9e71-450b-9b72-9056e61d78f4/alexandrite/out/build/image-root]
DEBUG: 20:50:13 | EXEC: [chroot /run/media/suser/bb3e71a7-9e71-450b-9b72-9056e61d78f4/alexandrite/out/build/image-root grep -q ^root: /etc/group]
DEBUG: 20:50:13 | EXEC: [chroot /run/media/suser/bb3e71a7-9e71-450b-9b72-9056e61d78f4/alexandrite/out/build/image-root grep -q ^live: /etc/group]
DEBUG: 20:50:13 | EXEC: Failed with stderr: (no output on stderr), stdout: (no output on stdout)
INFO: 20:50:13 | Adding group live
DEBUG: 20:50:13 | EXEC: [chroot /run/media/suser/bb3e71a7-9e71-450b-9b72-9056e61d78f4/alexandrite/out/build/image-root groupadd live]
INFO: 20:50:13 | Setting up user root
DEBUG: 20:50:13 | EXEC: [chroot /run/media/suser/bb3e71a7-9e71-450b-9b72-9056e61d78f4/alexandrite/out/build/image-root grep -q ^root: /etc/passwd]
DEBUG: 20:50:13 | EXEC: [openssl passwd -1 -salt xyz root7]
INFO: 20:50:13 | --> Modifying user: root
INFO: 20:50:13 | --> Primary group for user root: root
DEBUG: 20:50:13 | EXEC: [chroot /run/media/suser/bb3e71a7-9e71-450b-9b72-9056e61d78f4/alexandrite/out/build/image-root usermod -p $1$xyz$0QQY0hyW88RFzAkb82Lrz1 -g root root]
INFO: 20:50:13 | Setting up user live
[...]

PR #2095, opened 2022-03-18
Include /run subdirectories in container images

tserong:
Rather than excluding *everything* in /run, this commit ensures that the directory structure inside /run in the image root is preserved inside container images.  Files remain excluded (so we won't pick up transient pidfiles and such).  This change _only_ applies for container image builds.  All other build types remain unaffected (i.e. all other build types will still have an empty /run directory).

Fixes: https://github.com/OSInside/kiwi/issues/2093
Signed-off-by: Tim Serong <tserong@suse.com>

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -787,6 +787,11 @@ derived_from="string":
   The image created by {kiwi} will use the specified container as the
   base root to work on.
 
+ensure_empty_tmpdirs="true|false":
+  For OCI container images, specifies whether to ensure /run and /tmp
+  directories are empty in the container image created by Kiwi.
+  Default is true.
+
 publisher="string":
   For ISO images, specifies the publisher name of the ISO.
 
--- a/kiwi/builder/container.py
+++ b/kiwi/builder/container.py
@@ -55,6 +55,7 @@ class ContainerBuilder:
         self.requested_container_type = xml_state.get_build_type_name()
         self.base_image = None
         self.base_image_md5 = None
+        self.ensure_empty_tmpdirs = True
 
         self.container_config['xz_options'] = \
             self.custom_args.get('xz_options')
@@ -83,6 +84,9 @@ class ContainerBuilder:
                         self.base_image_md5
                     )
                 )
+
+        if xml_state.build_type.get_ensure_empty_tmpdirs() is False:
+            self.ensure_empty_tmpdirs = False
 
         self.system_setup = SystemSetup(
             xml_state=xml_state, root_dir=self.root_dir
@@ -140,7 +144,7 @@ class ContainerBuilder:
             self.requested_container_type, self.root_dir, self.container_config
         )
         self.filename = container_image.create(
-            self.filename, self.base_image
+            self.filename, self.base_image, self.ensure_empty_tmpdirs
         )
         Result.verify_image_size(
             self.runtime_config.get_max_size_constraint(),
--- a/kiwi/container/appx.py
+++ b/kiwi/container/appx.py
@@ -68,12 +68,13 @@ class ContainerImageAppx:
                 )
             )
 
-    def create(self, filename, base_image=None):
+    def create(self, filename, base_image=None, ensure_empty_tmpdirs=None):
         """
         Create WSL/Appx archive
 
         :param string filename: archive file name
         :param string base_image: not-supported
+        :param string ensure_empty_tmpdirs: not-supported
         """
         exclude_list = Defaults.\
             get_exclude_list_for_root_data_sync() + Defaults.\
--- a/kiwi/container/oci.py
+++ b/kiwi/container/oci.py
@@ -97,7 +97,7 @@ class ContainerImageOCI:
             self.oci_config['history']['created_by'] = \
                 Defaults.get_default_container_created_by()
 
-    def create(self, filename, base_image):
+    def create(self, filename, base_image, ensure_empty_tmpdirs=True):
         """
         Create compressed oci system container tar archive
 
@@ -105,7 +105,7 @@ class ContainerImageOCI:
         :param string base_image: archive used as a base image
         """
         exclude_list = Defaults.\
-            get_exclude_list_for_root_data_sync() + Defaults.\
+            get_exclude_list_for_root_data_sync(ensure_empty_tmpdirs) + Defaults.\
             get_exclude_list_from_custom_exclude_files(self.root_dir)
         exclude_list.append('dev/*')
         exclude_list.append('sys/*')
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -336,7 +336,7 @@ class Defaults:
         ]
 
     @staticmethod
-    def get_exclude_list_for_root_data_sync():
+    def get_exclude_list_for_root_data_sync(no_tmpdirs: bool = True):
         """
         Provides the list of files or folders that are created
         by KIWI for its own purposes. Those files should be not
@@ -347,7 +347,11 @@ class Defaults:
         :rtype: list
         """
         exclude_list = [
-            'image', '.profile', '.kconfig', 'run/*', 'tmp/*',
+            'image', '.profile', '.kconfig'
+        ]
+        if no_tmpdirs:
+            exclude_list += ['run/*', 'tmp/*']
+        exclude_list += [
             Defaults.get_buildservice_env_name(),
             Defaults.get_shared_cache_location()
         ]
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -1967,6 +1967,14 @@ div {
             sch:param [ name = "attr" value = "derived_from" ]
             sch:param [ name = "types" value = "docker oci" ]
         ]
+    k.type.ensure_empty_tmpdirs.attribute =
+        ## Whether to ensure /run and /tmp directories are empty in the
+        ## container image created by Kiwi.  Default is true.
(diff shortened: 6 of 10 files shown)


schaefi:
Thanks much, from my perspective this looks good but I'd like to have @davidcassany a look as well since he is also the expert on containers.

Conan-Kudo:
This makes me very leery, since `/run` is supposed to be a tmpfs and thus completely empty.

davidcassany:
Hi Tim, thanks much for the PR. Your time and efforts are really appreciated. To be honest I am not fully convinced with this PR, I have a couple of concerns:

* To me it feels that there is not good reason to apply this filter under `/run/` but not on `/tmp`. I mean the former change that triggered this issue applied for both `/tmp` and  `/run`, ceph/rook could have used `/tmp` for some other transient data and we would have the same issue. So to make this PR complete we are missing the filter for `/tmp` IMHO.

* On the other side I have been thinking that probably the best for containers is simply do not exclude `/run` and `/tmp` at all and assume it is at users hands to clean it up if wanted, in fact this would provide maximum flexibility. Recently we have been discussing around `/run` usage during the image build here #2067, as a result of that I think we are clear we won't bind mount host's `/run` into chroot used for the root tree creation. So after the image creation the contents of `/run` should only be whatever packages installation included in there and almost nothing else, since we are not running services in that chroot, mostly the package manager, few linux common utilities and the user defined config.sh.

On full OS systems keeping stuff inside the image in `/run` and `/tmp` is really bad since these paths are usually mounted as tmpfs at boot time, hence any data in there becomes hidden behind a mountpoint, so it is far from obvious that there was something there. For containers this is not happening hence unwanted data, if any, should be clearly visible.

I believe it is a bad practice having containers with preloaded data in `/run` or `/tmp` as this is likely to point to some erratic usage of these paths, but I guess from a KIWI POV there are no good reasons to be that opinionated about directory hierarchy for containers and leave it at users hands.

Do you have any other thoughts?

Conan-Kudo:
> I believe it is a bad practice having containers with preloaded data in /run or /tmp as this is likely to point to some erratic usage of these paths, but I guess from a KIWI POV there are no good reasons to be that opinionated about directory hierarchy for containers and leave it at users hands.

From a kiwi point of view, we should only provide _empty_ `/run` and `/tmp`, because our principal focus is ensuring we make valid OS images.

davidcassany:
> > I believe it is a bad practice having containers with preloaded data in /run or /tmp as this is likely to point to some erratic usage of these paths, but I guess from a KIWI POV there are no good reasons to be that opinionated about directory hierarchy for containers and leave it at users hands.
> 
> From a kiwi point of view, we should only provide _empty_ `/run` and `/tmp`, because our principal focus is ensuring we make valid OS images.

I fully agree on that, but containers are not really OS images and my suggestion is to only relax this constraints for containers.

Conan-Kudo:
> > > I believe it is a bad practice having containers with preloaded data in /run or /tmp as this is likely to point to some erratic usage of these paths, but I guess from a KIWI POV there are no good reasons to be that opinionated about directory hierarchy for containers and leave it at users hands.
> > 
> > 
> > From a kiwi point of view, we should only provide _empty_ `/run` and `/tmp`, because our principal focus is ensuring we make valid OS images.
> 
> I fully agree on that, but containers are not really OS images and my suggestion is to only relax this constraints for containers.

The issue with that is people could be creating OS containers (I do this for something at work) and not having that guarantee  in kiwi introduces a _huge_ footgun. Honestly, this sounds like something is broken with Rook/Ceph because this isn't required for Ceph on a VM or bare metal (I've run Ceph on Fedora before, and this is definitely unnecessary for it).

davidcassany:
> The issue with that is people could be creating OS containers (I do this for something at work) and not having that guarantee in kiwi introduces a _huge_ footgun.

This is indeed a good point and I also use containers as OS base, which makes wonder what happens on `buildah commit` or `docker commit` if I have added stuff in `/run`. Gonna try even if it is just to know how they behave. Also this reveals that this sort of change does no make sense at all for WSL, ns-spawn, lxc and any OS like container approach. The discussion should be limited to OCI only.

> Honestly, this sounds like something is broken with Rook/Ceph because this isn't required for Ceph on a VM or bare metal (I've run Ceph on Fedora before, and this is definitely unnecessary for it).

This is also clear to me and changes, if any, should be motivated in favor of providing more flexibility. So the question is if KIWI should allow data under `/run` or `/tmp` for OCI images. Given that out there is plenty of containers that do no follow any OS convention (executables at `/`, single binary containers, etc.), which I think is bad practice, but it happens and people seams to expect that level of freedom for OCI/docker containers. I think the discussion should around this idea, just trying to find pros and cons.

Conan-Kudo:
If want to make a custom image similar to the `ubi-init` container (`registry.access.redhat.com/ubi8-minimal`), contents in `/run` and `/tmp` would absolutely break it. And that's an OCI container.

Conan-Kudo:
If want to make a custom image similar to the `ubi-init` container (`registry.access.redhat.com/ubi8-init`), contents in `/run` and `/tmp` would absolutely break it. And that's an OCI container.

tserong:
> This is also clear to me and changes, if any, should be motivated in favor of providing more flexibility. So the question is if KIWI should allow data under `/run` or `/tmp` for OCI images. Given that out there are plenty of containers that do no follow any OS convention (executables at `/`, single binary containers, etc.), which I think is bad practice, but it happens and people seams to expect that level of freedom for OCI/docker containers, I think the discussion should be around this idea. I am just trying to find pros and cons.

I think Kiwi _should_ allow data under `/run` (or `/tmp`), because being an imaging tool, it makes sense to allow the tool to create images including anything the image creator requires, even if doing so might be considered non-standard/undesirable in the general OS case.

Would you be open to making this optional? Can we add a new option to the kiwi XML schema somewhere (call it "purge_run" or "purge_tmpfses" or something), default it to _true_, so we keep the current behaviour and avoid potential footguns, while still allowing container creators to turn that option off if they need to ensure a directory structure is preserved in `/run`? I think that'll give everyone what they need, and I'm happy to work on the implementation if we can get general agreement on the right approach.

> Honestly, this sounds like something is broken with Rook/Ceph because this isn't required for Ceph on a VM or bare metal (I've run Ceph on Fedora before, and this is definitely unnecessary for it).

It's a deliberate design decision by the Rook folks to rely on `/run/ceph` and `/var/run/ceph` existing in container images, so it's not a bug, even if it might be "wrong" in some sense ;-)  The reason you've never had this problem running Ceph on bare metal is because in the bare metal case, systemd's tmpfiles.d mechanism is used to create the required subdirectories of `/run`. This was discussed at more length in the comments on https://github.com/OSInside/kiwi/issues/2093.

Conan-Kudo:
> Would you be open to making this optional? Can we add a new option to the kiwi XML schema somewhere (call it "purge_run" or "purge_tmpfses" or something), default it to true, so we keep the current behaviour and avoid potential footguns, while still allowing container creators to turn that option off if they need to ensure a directory structure is preserved in `/run`? I think that'll give everyone what they need, and I'm happy to work on the implementation if we can get general agreement on the right approach.

As icky as this whole thing is, yeah, I'd probably be fine with that. We have other options like that...

Conan-Kudo:
> Would you be open to making this optional? Can we add a new option to the kiwi XML schema somewhere (call it "purge_run" or "purge_tmpfses" or something), default it to true, so we keep the current behaviour and avoid potential footguns, while still allowing container creators to turn that option off if they need to ensure a directory structure is preserved in `/run`? I think that'll give everyone what they need, and I'm happy to work on the implementation if we can get general agreement on the right approach.

As icky as this whole thing is, yeah, I'd probably be fine with that. We have other options like that. I'd suggest `ensure_empty_tmpdirs` and default it to True.

davidcassany:
@tserong I think this is a good suggestion and a reasonable option. It should be simple to implement and easy to understand and use if required under some uncommon use case or need.

I think a reasonable way to code that could be by adding a default `no_tmpdirs=true` value into the `defaults.get_exclude_list_for_root_data_sync` method. So we would exclude or not `/tmp` and `/run` according to this value.

schaefi:
@davidcassany I like the idea, are you going to implement this part ? Thanks

tserong:
I'm working on it ATM

schaefi:
Thanks much

tserong:
OK, how's it look now? :-)

schaefi (approved):
The implementation is great, thanks :+1: 

In the beginning I struggled a bit that this can only be set for containers but I think if we need it for other image types as well the scope can be easily increased by adapting the schematron rule, the docs and the respective code parts for other types

So from my point of view this is good

Thanks much

schaefi:
@tserong sorry due to the merge of #2101 there is now a conflict on the auto generated (generateDS) data. I think you will be able to simply resolve this by deleting xml_parse.py in your branch, rebase to master and run make kiwi/schema/kiwi.rng to create the api code again. Let me know if this is causing you trouble. Thanks

tserong:
No problem @schaefi, rebased.

davidcassany (approved):
Looks good, thanks @tserong 

As a side note if you want it to be released in a SLE stream any time relatively soon I'd suggest you to add a bugzilla ticket reference (`bsc#<somenumber>`) in the commit message as we generate RPM change log based on github history.

It is unclear to me under which scope this was an issue as in the github issue you mention SLE15 SP2 however looks like you were using opensuse based images on the tests. Anyway, just warning you that without a proper bug reference and prioritized bug ticket this is likely to take a while until it hits SLE 15.

tserong:
Bug ref added

PR #2096, opened 2022-03-18
Allow luks format options

### LINKED ISSUE #1898: Support for LUKS2
Some time ago, cryptsetup introduced the new LUKS2 format which is very similar to LUKS1. It's the upstream default, but some distros (like openSUSE) use LUKS1 as default. For Tumbleweed this is about to change.

GRUB 2.06 supports booting from LUKS2 (though there are a few bugs, I'm working on fixes currently), but the `luks2` module needs to be included. FWICT, this isn't listed in `get_grub_basic_modules` yet.

It would be nice if kiwi also allows specifying whether to use LUKS1 or LUKS2.

schaefi:
Added new luksformat element which allows to pass
along options to the luksFormat call. This allows users to
switch between LUKS and LUKS2 via e.g

```xml
<luksformat>
    <option name="--type" value="luks2"/>
</luksformat>
```

It also allows to pass along a set of options only available
to LUKS2, e.g

```xml
<luksformat>
    <option name="--cipher" value="aes-gcm-random"/>
    <option name="--integrity" value="aead"/>
</luksformat>
```

In addition to the new attribute the existing luks attribute
can also be specified to read credentials from a keyfile by
using the file:// source locator, e.g

```xml
<type ... luks="file:///path/to/a/keyfile"/>
```

This Fixes #1898

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -485,11 +485,15 @@ kernelcmdline="string":
   Additional kernel parameters passed to the kernel by the
   bootloader.
 
-luks="passphrase":
-  Supplying a value will trigger the encryption of the partitions
-  using the LUKS extension and using the provided string as the
-  password. Note that the password must be entered when booting the
-  appliance!
+luks="passphrase|file:///path/to/keyfile":
+  Supplying a value will trigger the encryption of the partition
+  serving the root filesystem using the LUKS extension. The supplied
+  value represents either the passphrase string or the location of
+  a key file if specified as `file://...` resource. When using
+  a key file it is in the responsibility of the user how
+  this key file is actually being used. By default any
+  distribution will just open an interactive dialog asking
+  for the credentials at boot time !
 
 target_blocksize="number":
   Specifies the image blocksize in bytes which has to
@@ -750,6 +754,33 @@ publisher="string":
 
 The following sections shows the supported child elements of the `type`
 element including references to their usage in a detailed type setup:
+
+.. _preferences-type-luksformat:
+
+<preferences><type><luksformat>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The `luksformat` element is used to specify additional luks options
+passed on to the `cryptsetup luksFormat` call. The element requires
+the attribute `luks` to be set in the `<type>` section referring to
+`luksformat`. Several custom settings related to the LUKS and LUKS2
+format features can be setup. For example, making the switch to `LUKS2`:
+
+.. code:: xml
+
+   <luksformat>
+     <option name="--type" value="luks2"/>
+   </luksformat>
+
+Or to enable additional device mapper layers and features,
+for example `dm_integrity`:
+
+.. code:: xml
+
+   <luksformat>
+     <option name="--type" value="luks2"/>
+     <option name="--cipher" value="aes-gcm-random"/>
+     <option name="--integrity" value="aead"/>
+   </luksformat>
 
 .. _preferences-type-bootloader:
 
--- a/kiwi/system/profile.py
+++ b/kiwi/system/profile.py
@@ -335,7 +335,7 @@ class Profile:
         self.dot_profile['kiwi_startsector'] = \
             self.xml_state.get_disk_start_sector()
         self.dot_profile['kiwi_luks_empty_passphrase'] = \
-            True if type_section.get_luks() == '' else False
+            self.xml_state.get_luks_credentials() == ''
 
     def _profile_names_to_profile(self):
         # kiwi_profiles
--- a/kiwi/xml_parse.py
+++ b/kiwi/xml_parse.py
@@ -1762,6 +1762,86 @@ class product(GeneratedsSuper):
 # end class product
 
 
+class option(GeneratedsSuper):
+    """A commandline option specification"""
+    subclass = None
+    superclass = None
+    def __init__(self, name=None, value=None):
+        self.original_tagname_ = None
+        self.name = _cast(None, name)
+        self.value = _cast(None, value)
+    def factory(*args_, **kwargs_):
+        if CurrentSubclassModule_ is not None:
+            subclass = getSubclassFromModule_(
+                CurrentSubclassModule_, option)
+            if subclass is not None:
+                return subclass(*args_, **kwargs_)
+        if option.subclass:
+            return option.subclass(*args_, **kwargs_)
+        else:
+            return option(*args_, **kwargs_)
+    factory = staticmethod(factory)
+    def get_name(self): return self.name
+    def set_name(self, name): self.name = name
+    def get_value(self): return self.value
+    def set_value(self, value): self.value = value
+    def hasContent_(self):
+        if (
+
+        ):
+            return True
+        else:
+            return False
+    def export(self, outfile, level, namespaceprefix_='', name_='option', namespacedef_='', pretty_print=True):
+        imported_ns_def_ = GenerateDSNamespaceDefs_.get('option')
+        if imported_ns_def_ is not None:
+            namespacedef_ = imported_ns_def_
+        if pretty_print:
+            eol_ = '\n'
+        else:
+            eol_ = ''
+        if self.original_tagname_ is not None:
+            name_ = self.original_tagname_
+        showIndent(outfile, level, pretty_print)
+        outfile.write('<%s%s%s' % (namespaceprefix_, name_, namespacedef_ and ' ' + namespacedef_ or '', ))
+        already_processed = set()
+        self.exportAttributes(outfile, level, already_processed, namespaceprefix_, name_='option')
+        if self.hasContent_():
+            outfile.write('>%s' % (eol_, ))
+            self.exportChildren(outfile, level + 1, namespaceprefix_='', name_='option', pretty_print=pretty_print)
+            outfile.write('</%s%s>%s' % (namespaceprefix_, name_, eol_))
+        else:
+            outfile.write('/>%s' % (eol_, ))
+    def exportAttributes(self, outfile, level, already_processed, namespaceprefix_='', name_='option'):
+        if self.name is not None and 'name' not in already_processed:
+            already_processed.add('name')
+            outfile.write(' name=%s' % (self.gds_encode(self.gds_format_string(quote_attrib(self.name), input_name='name')), ))
+        if self.value is not None and 'value' not in already_processed:
+            already_processed.add('value')
+            outfile.write(' value=%s' % (self.gds_encode(self.gds_format_string(quote_attrib(self.value), input_name='value')), ))
+    def exportChildren(self, outfile, level, namespaceprefix_='', name_='option', fromsubclass_=False, pretty_print=True):
+        pass
+    def build(self, node):
+        already_processed = set()
+        self.buildAttributes(node, node.attrib, already_processed)
+        for child in node:
+            nodeName_ = Tag_pattern_.match(child.tag).groups()[-1]
+            self.buildChildren(child, node, nodeName_)
+        return self
+    def buildAttributes(self, node, attrs, already_processed):
+        value = find_attr_value_('name', node)
+        if value is not None and 'name' not in already_processed:
+            already_processed.add('name')
+            self.name = value
+        value = find_attr_value_('value', node)
+        if value is not None and 'value' not in already_processed:
+            already_processed.add('value')
+            self.value = value
+    def buildChildren(self, child_, node, nodeName_, fromsubclass_=False):
+        pass
+# end class option
+
+
 class package(GeneratedsSuper):
     """Name of an image Package"""
     subclass = None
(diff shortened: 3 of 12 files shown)


Conan-Kudo:
I don't think I like setting the LUKS type this way. We try to avoid freeform options we can't validate, right? This seems antithetical to that goal.

isbm (changes requested):
I am not sure about the design, but allowing the entire options as one command line with whatever possible is something that we did in a past in Salt and it was a very bad idea, which we had to rework afterwards. But it is up to you (I would vote against that though, if my voice is any interesting).

kiwi/xml_parse.py
```diff
@@ -2718,7 +2718,7 @@ class type_(GeneratedsSuper):
...
...e, editbootinstall=None, filesystem=None, flags=None, format=None, formatoptions=None, fsmountoptions=None, fscreateoptions=None, squashfscompression=None, gcelicense=None, hybridpersistent=None, hybridpersistent_filesystem=None, gpt_hybrid_mbr=None, force_mbr=None, initrd_system=None, image=None, metadata_path=None, installboot=None, install_continue_on_timeout=None, installprovidefailsafe=None, installiso=None, installstick=None, installpxe=None, mediacheck=None, kernelcmdline=None, luks=None, luks_format_options=None, luksOS=None, mdraid=None, overlayroot=None, overlayroot_write_partition=None, overlayroot_readonly_partsize=None, overlayroot_verity_blocks=None, primary=None, ramonly=None, rootfs_label=None, spare_part=None, spare_part_mountpoint=None, spare_part_fs=None, spare_part_fs_attributes=None, spare_part_is_last=None, target_blocksize=None, target_removable=None, vga=None, vhdfixedtag=None, volid=None, wwid_wait_timeout=None, derived_from=None, xen_server=None, publisher=None, disk_start_sector=None, bundle_format=None, bootloader=None, containerconfig=None, machine=None, oemconfig=None, size=None, systemdisk=None, partitions=None, vagrantconfig=None, installmedia=None):
```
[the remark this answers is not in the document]
schaefi: autogenerated code ;)

kiwi/xml_state.py
```diff
@@ -2212,6 +2213,39 @@ def get_fs_create_option_list(self) -> List:
...
+    def get_luks_credentials(self) -> Optional[str]:
+        """
+        Return key or passphrase credentials to open the luks pool
+
+        :return: data
+
+        :rtype: str
+        """
+        data = self.build_type.get_luks()
+        if data and data.startswith('file://'):
```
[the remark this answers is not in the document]
schaefi: yeah we even have an Uri class. I did not use it because this imply we allow keyfiles from other locations like http, ftp, etc etc. As this is the path to a private key file it will most likely never be exposed on anything else than the local system. That was the reason why I chose only file://

I'm open to change that if people thinks it makes sense ?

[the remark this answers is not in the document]
Conan-Kudo: I think we should handle it properly. Maybe extend the Uri class to have a flag to restrict to local only as desired?

schaefi:
> I don't think I like setting the LUKS type this way. We try to avoid freeform options we can't validate, right? This seems antithetical to that goal.

done, I have changed the way to describe options to be clean now

schaefi:
> I am not sure about the design, but allowing the entire options as one command line with whatever possible is something that we did in a past in Salt and it was a very bad idea, which we had to rework afterwards. But it is up to you (I would vote against that though, if my voice is any interesting).

yep your vote is always welcome, I have changed it. Thanks

kiwi/system/profile.py
```diff
@@ -335,7 +335,7 @@ def _type_to_profile(self):
         self.dot_profile['kiwi_startsector'] = \
             self.xml_state.get_disk_start_sector()
         self.dot_profile['kiwi_luks_empty_passphrase'] = \
-            True if type_section.get_luks() == '' else False
+            True if self.xml_state.get_luks_credentials() == '' else False
```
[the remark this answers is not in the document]
isbm: Well. Cmon. :)
```python
self.dot_profile['kiwi_luks_empty_passphrase'] = self.xml_state.get_luks_credentials() == ''
```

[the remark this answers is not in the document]
schaefi: yeah true, thanks. I changed it

kiwi/xml_state.py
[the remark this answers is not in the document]
isbm: `/me` wish to have such hadouken powers of Conan-Kudo over @schaefi sometimes... :wink:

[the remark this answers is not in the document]
schaefi: This are just the simple rules of democracy :smile:

doc/source/image_description/elements.rst
```diff
@@ -751,6 +755,33 @@ publisher="string":
...
+Or to enable additional device mapper layers and features,
+for example `dm_integrity`:
+
+.. code:: xml
+
+   <luksformat>
+     <option name="--type" value="luks2"/>
+     <option name="--cipher" value="aes-gcm-random"/>
+     <option name="--integrity" value="aead"/>
+   </luksformat>
```
Conan-Kudo, line 783: This is only somewhat better in my view, in that each option is there now, but are we able to validate what options are being set and what are the acceptable values? If we can't do that, I'm not sure this is enough.

Conan-Kudo (changes requested):

kiwi/xml_state.py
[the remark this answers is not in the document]
schaefi: changes done, this is now hopefully more acceptable :) Thanks

doc/source/image_description/elements.rst
isbm: Oh, I like your idea! But not a lot,

Because why `--` is needed for long options? There will be more stuff like that in a future, so why not add the whole args aggregate instead and then reuse it in a future elsewhere as a standard way of integrating things? Something like this:

```xml
<whatever-stuff>
  <args>
    <!-- Default. But in some crazy software like e.g. Java,
         long opts with just one "-" so we can redefine -->
    <longopt value="--" /> 
    <flag name="d" /> <!-- just a whatever boolean -->
    <flag name="v" />
    <flag name="v" />
    <flag name="v" /> <!-- verbooooose! -->
    <option name="type" value="foo" />
    <cmd value="bar" /> <!-- that could be a filename or whatever that is not a flag or an option -->
  </args>
</whatever-stuff>
```

That would be parsed to:

`blah -d -v -v -v --type=foo bar`

So OK, I don't like your idea. :wink:

kiwi/xml_parse.py
```diff
@@ -1762,6 +1762,86 @@ def buildChildren(self, child_, node, nodeName_, fromsubclass_=False):
...
+    def set_name(self, name): self.name = name
+    def get_value(self): return self.value
+    def set_value(self, value): self.value = value
+    def hasContent_(self):
+        if (
+
+        ):
+            return True
+        else:
+            return False
```
isbm, line 1794: I don't get it.

kiwi/xml_parse.py
```diff
@@ -1762,6 +1762,86 @@ def buildChildren(self, child_, node, nodeName_, fromsubclass_=False):
...
+    def export(self, outfile, level, namespaceprefix_='', name_='option', namespacedef_='', pretty_print=True):
+        imported_ns_def_ = GenerateDSNamespaceDefs_.get('option')
+        if imported_ns_def_ is not None:
+            namespacedef_ = imported_ns_def_
+        if pretty_print:
+            eol_ = '\n'
```
isbm, line 1800: Actually `os.linesep` if to be proper.

doc/source/image_description/elements.rst
schaefi: I'm sorry but a validation for correctness of options to LUKS is imho out of scope. If you look at ```man cryptsetup``` you find many many options that you can set. As we are also talking about cryptographic settings many of the option values hooks into capabilities of the kernel. I don't see a good way to pre-condition these settings on the level of kiwi. To be honest I would like to offer users the flexibility to set any option possible. If the desired cryptographic output can be generated or not will be the return of the cryptsetup call issued by kiwi including these options.

From that information users imho gets all information they need to find out e.g

* option only available to luks2
* kernel has no support for cipher xyz
* ...and-a-ton-of-other-possible-issues

So you are saying from a documentation perspective it could be done better. Yes I agree but I have no clue about all possible cryptographic settings, which kernel versions on what distro supports which setting and what options can be safely combined with other features of luks.

That's why I just made a simple example for the use of some optional crypto settings.

What's your suggestion how to change this ?

kiwi/xml_parse.py
```diff
@@ -1762,6 +1762,86 @@ def buildChildren(self, child_, node, nodeName_, fromsubclass_=False):
...
+        if value is not None and 'name' not in already_processed:
+            already_processed.add('name')
+            self.name = value
+        value = find_attr_value_('value', node)
+        if value is not None and 'value' not in already_processed:
+            already_processed.add('value')
+            self.value = value
+    def buildChildren(self, child_, node, nodeName_, fromsubclass_=False):
+        pass
+# end class option
```
isbm, line 1842: Is this is some sort of automatically generated code or? Then you could use [`autopep8`](https://github.com/hhatto/autopep8) or [`yapf`](https://github.com/google/yapf). Still would be more human-compatible. Just ensure it is using always the same rules.

isbm (changes requested):

doc/source/image_description/elements.rst
schaefi: > Because why `--` is needed for long options?

There is no distinction in how an option is set and it's not a responsibility of kiwi to specify a schema for it.
I have implemented the ```<option>``` element in the most generic variant to be be re-usable in other sub-elements
of the description when needed. A command has options and options can have optional values. That's all you need
to know and specify from a schema perspective in kiwi. This allows to set and handle every caller aspect in my eyes

```xml
<option name="-h"/>

<option name="--help"/>

<option name="fetch" value="--all"/>

<option name="ec2"/>
<option name="describe-security-groups"/>

<option name="kafka"/>
<option name="create-cluster"/>
<option name="--cluster-name" value="cloud-builder"/>
<option name="--broker-node-group-info" value="file://cb-msk-setup.json"/>
```

you can do everything without setting scope on option syntax

> So OK, I don't like your idea. wink

That's fair and I don't like yours ;) mostly because it seems you are thinking in terms of an option parser and that's not what we need here. All what we need is an element container to specify options such that they can be passed along to a command no matter if that command uses long-options, short-options, positional parameters or some strange combination of them

So please let us keep this to the simplest possible element description. Well the simplest one was the one string for all and I agreed with all of you that it was a bad idea, but there is no need for any more complicated schema syntax than a command has options and optional values to them.

kiwi/xml_parse.py
schaefi: can you please stop reviewing auto-generated code ? All this is from generateDS. Thanks

kiwi/xml_parse.py
schaefi: autogenerated

kiwi/xml_parse.py
schaefi: autogenerated

doc/source/image_description/elements.rst
Conan-Kudo: I think we can at least make some flags, like the luks version, something we can validate. We could at least verify that the options being passed are correct for a given LUKS version, even if we can't validate the values. That's not different than what we do for other things.

isbm: If you in favour of a half-baked one-size-fits-all options, then what's wrong with just this:
```xml
<options>
-h --help fetch --all ec2 describe-security-groups 
kafka create-cluster
--cluster-name cloud-builder
--broker-node-group-info file://cb-msk-setup.json
</options>
```
...? At least one will not full himself that the `<option/>` means something validated.

schaefi: > If you in favour of a half-baked one-size-fits-all options, then what's wrong with just this:

This approach would then be no different to allow a custom option string. There would be not difference in ```attribute="text"``` vs ```<element>text</element>```. I thought we agreed on to not allow this flexibility. I think the biggest disadvantage in free form text is that you have to specify a delimiter that separates one option from the next
and this always causes issues. So I still think we want each option in its own namespace

> I think we can at least make some flags, like the luks version, something we can validate. We could at least verify that the options being passed are correct for a given LUKS version, even if we can't validate the values. That's not different than what we do for other things.

I agree that we do this in other parts of the schema. I would also agree that it might be good to have an extra attribute
to set the luks version e.g ```luks_version="1|2"```

Regarding the options per version these are quite some values:

generic options that we can use in this scope:

--hash,
--cipher
--verify-passphrase
--use-random |--use-urandom
--uuid
--iter-time
--header
--pbkdf-force-iterations
--force-pass-word
--disable-locks

additional to luks2

--integrity
--integrity-no-wipe
--sector-size
--label
--subsystem
--pbkdf
--pbkdf-memory
--pbkdf-parallel
--disable-locks
--disable-keyring
--luks2-metadata-size
--luks2-keyslots-size
--keyslot-cipher
--keyslot-key-size

I still don't understand why we want to validate them in the scope of kiwi ? I believe this validation does not bring much value to the user. Adding them to the schema forces us to maintain cryptsetup options in the scope of the kiwi schema. I think we can only loose here because what if a new version of cryptsetup deprecates caller options or add new ones. We would be constantly have to maintain this list to be valid and correct.

I'm not aware where in the kiwi schema we would do this for any other tool used when building images ? We don't to this for lvcreate, pvcreate, vgcreate, nor for mdadm, mkfs.xxx or any other tooling.

I really don't understand why we want to add that burden ?

Thanks

Conan-Kudo: Do we let people pass options for those? I didn't know we let people pass arbitrary mkfs options.

isbm: @Conan-Kudo that was also my impression. So I firmly bullt my opinion (probably a wrong one) that if we can let people pass arbitrary opts, then why not make it then somewhat "interface-ish"?..

isbm: > I think we can only loose here because what if a new version of cryptsetup deprecates caller options or add new ones. We would be constantly have to maintain this list to be valid and correct.

Actually that what we did in Salt to proxy-pass any kinf of command. In version of Solaris it was even more funny how did we "supported" ZFS's tooling (zpool and such): by simply call a `zpool --help` and then parse the STDERR (to which the help is printed) into the options and make a list of valid opts. :wink: 

Probably @Conan-Kudo wants to prevent from that happen if a user will write something like:

```xml
<option key="help" value=";rm -rf /" />
```

On the other hand, I would not care much of this, because Kiwi asks for a root anyways — selber Schuld. :wink: As of my side, if I would need to pass call options to an external program, I bet this will pop out in a future again. So instead of making built-in schema into a particular command, I would simply abstract from it and make a _reusable_ aggregate of options, that could be appliccable again and again elsewhere.

But sure, format this particular case is entirely up to you.

schaefi: > Do we let people pass options for those? I didn't know we let people pass arbitrary mkfs options

we don't allow passing arbitrary options to any of the tools I mentioned. ok I agree this would be the first time we would allow users to send custom options even invalid ones to a tool we use in kiwi.

So what do we want to achieve ?

* The risk @isbm mentioned with ```<option key="help" value=";rm -rf /" />``` does not exist, we don't shell evaluate anything that is considered arbitrary data
 
* The risk to pass along options to cryptsetup that doesn't make any sense, yes this risk exists and will lead the image build to fail at call time of cryptsetup. What is wrong with that ? Obviously you have more concerns about this than I have, but why ? I would like to understand that

In kiwi we have the concept of runtime checks and we also have a class that supports: ```CommandCapabilities.has_option_in_help(...)```. So are you saying we should create a runtime check for the options passed and see if they are valid for cryptsetup ? Is that what you are after ? This can be done and it will be good for the "fail fast" behavior. I'm open to add that runtime validation

I'm against adding a set of options as fixed values to the schema, sorry I'm not going to do this. I don't see how this would be helpful in a world of constant changes to linux tools.

> then I would not reinvent the whell all over again adding another complex schema with options

If you look at how the ```<option>``` element is encoded in the .rnc file you see that it can be generically used for any sub-element. Currently it applies only to ```<luksformat>```. Thus the no-reinvent-flag is imho already taking into account. But again you will not see me adding any specific tool option to the schema. Unless somebody can explain to me why this would be useful

Thanks

schaefi: And then there is still the question if we want to have an extra attribute to differentiate between luks version 1 and 2, something like

```xml
<type ... luks_version="2"/>
```

this information would then also help with the runtime validation of options to cryptsetup.

But let us first come all on the same page regarding the options topic in general.

I would have never thought that it could be that much of an issue. Thanks

Conan-Kudo: > In kiwi we have the concept of runtime checks and we also have a class that supports: CommandCapabilities.has_option_in_help(...). So are you saying we should create a runtime check for the options passed and see if they are valid for cryptsetup ? Is that what you are after ? This can be done and it will be good for the "fail fast" behavior. I'm open to add that runtime validation

That would be great to have. I don't want this to be just _nothing at all_.

Conan-Kudo: > And then there is still the question if we want to have an extra attribute to differentiate between luks version 1 and 2, 

I would prefer to have one, if for nothing else so external parsing to determine basic setup of descriptions aren't so bad.

schaefi: > I would prefer to have one, if for nothing else so external parsing to determine basic setup of descriptions aren't so bad.

yes makes sense. I added that part now

```xml
<type ... luks_version="luks|luks2"
```

LUKS (version 1) stays to be the default

schaefi: Looking into the validation of options against the help message as suggested next

schaefi: as suggested options passed will no be validated against a runtime check which looks up the option in the help output from cryptsetup. Please also note all runtime checks can be individually switched on/off just in case the checking delivers false positive. But the help output from cryptsetup is nice and unless they change that in the future to something unmatchable I don't expect issues

schaefi: @Conan-Kudo @isbm I hope addressed the concerns, how about it now ?

Thanks

isbm: From my side it looks to me still odd. Why I would have in schema `luks_version`, if there are also other stuff out there like eCryptfs, CipherShed, EncFS etc? How to use these then, if I don't want LUKS at all? But OK. Probably not my business. :wink:

schaefi: @isbm I see and understand your point. It's true the information to activate and use the feature of encrypted images is targeted to luks exclusively in kiwi and that you can also see on the description level. To make this independent of an actual implementation the schema description needs to change in something like

```xml
<disk_encryption engine="luks|X|Y|Z">
    ... more stuff to describe encryption
</disk_encryption>
```

From the scope of this PR I'm afraid this cannot be a part of it because it's not a small change. However, it's possible to make this change as a refactoring task including an XSLT schema transform such that no breakage with older descriptions will happen.

I will add a new issue describing this idea and so we can have a conversation about further details there.
Thanks

Conan-Kudo (approved):

PR #2098, opened 2022-03-21
Add support for verity setup on standard rootfs

schaefi:
So far the verity support was only available with the overlayroot layout and the read-only squashfs root. This commit adds a new attribute: ```verity_blocks="number|all"``` which allows to create the verity setup also on the standard root partition

In addition to the change it was needed to extend the Filesystem API with an additional optional paramter to allow setup of the filesystem UUID. Having the opportunity to set the UUID at filesystem creation is generally useful and with regards to this particular change it became also required

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -554,6 +554,16 @@ devicepersistency="by-uuid|by-label":
 squashfscompression="uncompressed|gzip|lzo|lz4|xz|zstd":
   Specifies the compression type for mksquashfs
 
+verity_blocks="number|all"
+  For the `oem` type only, specifies to create a dm verity hash
+  from the number of given blocks (or all) placed at the end of the
+  root filesystem For later verification of the device,
+  the credentials information produced by `veritysetup` from the
+  cryptsetup tools are needed. This data as of now is only printed
+  as debugging information to the build log file. A concept to
+  persistently store the verification metadata as part of the
+  partition(s) will be a next step.
+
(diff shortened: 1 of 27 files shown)


doc/source/image_description/elements.rst
```diff
@@ -582,14 +592,9 @@ overlayroot="true|false"
...
 overlayroot_verity_blocks="number|all"
-  For the `oem` type only, specifies to create a dm verity hash
-  from the number of given blocks (or all) placed at the end of the squashfs
-  compressed read-only root filesystem. For later verification of the device,
-  the credentials information produced by `veritysetup` from the
-  cryptsetup tools are needed. This data as of now is only printed
-  as debugging information to the build log file. A concept to
-  persistently store the verification metadata as part of the
-  partition(s) will be a next step.
+  Same as the `verity_blocks` attribute but the verity hash is
```
davidcassany, line 595: What happens if `overlayroot_verity_blocks` and `verity_blocks` are both used at the same time? I am asking because I wonder if we could get rid of `overlyroot_verity_blocks` in favor of only having `verity_blocks` and just behave as it used to do in case of overlayroot.

I am not verify familiar with this feature hence I am not sure I miss something.

davidcassany (approved):
Looks good to me even I am not familiar with the feature so I do not completely follow the implementation.

doc/source/image_description/elements.rst

Conan-Kudo (approved):

PR #2102, opened 2022-03-25
Fixed parsing of veritysetup output

schaefi:
veritysetup uses tabs to align values. The way kiwi parsed
the values did not strip out the tabs and later on keeps
them in the verification metadata block. The unit test
did not catch this because the mock output used for
veritysetup did not contain tabs. This commit fixes the
test to catch this condition and also fixes the code to
handle all space characters (tabs, space, newlines) in
a safe way


raimue (approved):
Yes, removing all whitespace before parsing the output into key/value pairs looks good.

PR #2103, opened 2022-03-29
Ensure BootloaderConfig resources are cleaned first

davidcassany:
This commit wraps the manual BootloaderConfig instance cleanup in
disk builder into a try/finally scope. This way if KIWI is aborted
or fails within this scope the BootloaderConfig is cleaned up first.

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -1322,14 +1322,15 @@ class DiskBuilder:
 
         if self.bootloader != 'custom':
             # create bootloader config prior bootloader installation
-            self.bootloader_config.setup_disk_image_config(
-                boot_options=custom_install_arguments
-            )
-            if 's390' in self.arch:
-                self.bootloader_config.write()
-
-            # cleanup bootloader config resources taken prior to next steps
-            del self.bootloader_config
+            try:
+                self.bootloader_config.setup_disk_image_config(
+                    boot_options=custom_install_arguments
+                )
+                if 's390' in self.arch:
+                    self.bootloader_config.write()
+            finally:
+                # cleanup bootloader config resources taken prior to next steps
+                del self.bootloader_config
 
             log.debug(
                 "custom arguments for bootloader installation %s",


kiwi/builder/disk.py
```diff
@@ -1322,14 +1322,15 @@ def _install_bootloader(
...
-
-            # cleanup bootloader config resources taken prior to next steps
-            del self.bootloader_config
+            try:
+                self.bootloader_config.setup_disk_image_config(
+                    boot_options=custom_install_arguments
+                )
+                if 's390' in self.arch:
+                    self.bootloader_config.write()
+            finally:
```
davidcassany, line 1331: Note that without this try/finally wrap of `bootloader_config` calls if KIWI is aborted during `bootloader_config.setup_disk_image_config` or `bootloader_config.write` the bootloader_config instance is not immediately cleaned up as it would happen in a successful execution.

davidcassany:
To be honest this possibly unveils a couple of design flaws in our code:

* If we require manually calling `del` over some object to force the destructor execution we should also consider the exception path and probably force it's execution under certain scope.
* If we require manually calling `del` probably there is something in our classes that could have been coded in a different way to be explicit about this requirement, aka there are interactions between objects that do not hold a reference of each other (BootloaderConfing mounts a luks device but does not know about the instance of LuksDevice that represents it)

Probably there are more like this one in KIWI code. Not a big deal, but makes me thing about good practices and how to prevent this sort of issues from a design PoV.

kiwi/builder/disk.py
schaefi: Thanks I understand and makes a lot sense.

I think the entire disk builder should be refactored. This is imho the most complicated builder we have and it's no longer a straight forward read as it was when it was implemented back in time. It has grown to a complexity level where I think a refactor would make sense also with regards to the resource cleanup. Maybe we can think of a kiwi10 branch from where we also don't need to stay backward compatible and introduce a new disk builder. Just an idea

schaefi: Thanks I understand and makes a lot sense.

I think the entire disk builder should be refactored. This is imho the most complicated builder we have and it's no longer a straight forward read as it was when it was implemented back in time. It has grown to a complexity level where I think a refactor would make sense also with regards to the resource cleanup. Maybe we can think of a kiwi10 branch from where we also don't need to stay backward compatible and introduce a new disk builder. Just an idea

schaefi (approved):

PR #2105, opened 2022-03-31
Fixed overlay root dracut module

schaefi:
make sure there is always only one selection for the
readonly root partition

### DIFF
--- a/dracut/modules.d/90kiwi-overlay/kiwi-overlay-root.sh
+++ b/dracut/modules.d/90kiwi-overlay/kiwi-overlay-root.sh
@@ -40,12 +40,12 @@ function initGlobalDevices {
     else
         write_partition="$1"
         root_disk=$(
-            lsblk -p -n -r -s -o NAME,TYPE "${write_partition}" |\
-            grep disk | cut -f1 -d ' '
+            lsblk -p -n -r -s -o NAME,TYPE "${write_partition}" \
+                | grep disk | cut -f1 -d ' '
         )
         read_only_partition=$(
-            lsblk -p -r --fs -o NAME,FSTYPE "${root_disk}" |\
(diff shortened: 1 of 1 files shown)


schaefi:
A clean detection based on an ID as we do it anywhere else would be preferred but so far I don't think this identification exists. Using the PARTUUID would only work for partition tables that supports it and thus I don't want to make use of this concept

iammattcoleman (approved):

dracut/modules.d/90kiwi-overlay/kiwi-overlay-root.sh
```diff
@@ -45,7 +45,7 @@ function initGlobalDevices {
         )
         read_only_partition=$(
             lsblk -p -r --fs -o NAME,FSTYPE "${root_disk}" |\
-            grep squashfs | cut -f1 -d ' '
+            grep squashfs | head -n 1 | cut -f1 -d ' '
```
iammattcoleman, line 48: For easier readability, I usually do two things slightly differently on lines like this:

1. I put the pipe character on the second line to clearly indicate that the line isn't starting a new command.
2. I indent the second line to reinforce that it's not a new command.

It's 100% just personal preference, but I feel like it leads to better maintainability because it makes newcomers less likely to misread those two lines as separate commands because they missed the trailing `\` on the first line.

```
lsblk -p -r --fs -o NAME,FSTYPE "${root_disk}" \
    | grep squashfs | head -n 1 | cut -f1 -d ' '
```

schaefi: I completely agree, thanks for the hint. I adapted the code part in this block to follow your idea

However, I feel there might be other shell code parts where you will find my style ;) let's improve that part as we stumble over it. Thanks

PR #2107, opened 2022-04-01
Don't compress .appx containers (#2106)

### LINKED ISSUE #2106: Appx containers are XZ compressed
It looks like appx images are now XZ compressed, which is inconvenient for users (Windows doesn't have tools for that) and not of much use as the main file inside is already compressed (`install.tar.gz`).

Vogtinator:
The container is actually inside and already compressed.

Fixes #2106


schaefi (approved):
Thanks, looks good :+1:

Conan-Kudo (approved):

PR #2108, opened 2022-04-01
Preserve LABEL setting

rjschwei:
+ Preserve the LABEL= setting when the grub config file is re-generated.
  the GRUB_ENABLE_LINUX_LABEL setting does not exists upstream and
  not in any SUSE distribution. Set the grub setting such that LABEL
  is preserved on SUSE distros. (bsc#1197616)

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -635,11 +635,13 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         * GRUB_TIMEOUT
         * GRUB_TIMEOUT_STYLE
         * SUSE_BTRFS_SNAPSHOT_BOOTING
+        * SUSE_REMOVE_LINUX_ROOT_PARAM
         * GRUB_BACKGROUND
         * GRUB_THEME
         * GRUB_USE_LINUXEFI
         * GRUB_USE_INITRDEFI
         * GRUB_SERIAL_COMMAND
+        * GRUB_CMDLINE_LINUX
         * GRUB_CMDLINE_LINUX_DEFAULT
         * GRUB_GFXMODE
         * GRUB_TERMINAL
@@ -657,6 +659,14 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             r'(^root=[^\s]+)|( root=[^\s]+)', '', self.cmdline
         ).strip()
         if self.persistency_type == 'by-label':
+            label = re.search(r'(^root=[^\s]+)|( root=[^\s]+)', self.cmdline)
+            if label and label.groups()[-1]:
+                label_definition = label.groups()[-1].strip()
+                if 'LABEL' in label_definition:
+                    grub_default_entries['GRUB_CMDLINE_LINUX'] = \
+                        '"{0}"'.format(label_definition)
+                    grub_default_entries['SUSE_REMOVE_LINUX_ROOT_PARAM'] = \
+                        'true'
             grub_default_entries['GRUB_ENABLE_LINUX_LABEL'] = 'true'
             grub_default_entries['GRUB_DISABLE_LINUX_UUID'] = 'true'
         elif self.persistency_type == 'by-partuuid':
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -681,6 +681,50 @@ class TestBootLoaderConfigGrub2:
 
     @patch('os.path.exists')
     @patch('kiwi.bootloader.config.grub2.SysConfig')
+    @patch('kiwi.bootloader.config.grub2.Command.run')
+    def test_setup_default_grub_use_of_by_label(
+        self, mock_Command_run, mock_sysconfig, mock_exists
+    ):
+        grep_grub_option = Mock()
+        grep_grub_option.returncode = 0
+        mock_Command_run.return_value = grep_grub_option
+        grub_default = MagicMock()
+        mock_sysconfig.return_value = grub_default
+        mock_exists.return_value = True
+        self.bootloader.terminal = 'serial'
+        self.bootloader.theme = 'openSUSE'
+        self.bootloader.displayname = 'Bob'
+        self.bootloader.cmdline = 'abcd root=LABEL=foo console=tty0'
+        self.bootloader.persistency_type = 'by-label'
+
+        self.bootloader._setup_default_grub()
+
+        assert grub_default.__setitem__.call_args_list == [
+            call(
+                'GRUB_BACKGROUND',
+                '/boot/grub2/themes/openSUSE/background.png'
+            ),
+            call('GRUB_CMDLINE_LINUX', '"root=LABEL=foo"'),
+            call('GRUB_CMDLINE_LINUX_DEFAULT', '"abcd console=tty0"'),
+            call('GRUB_DISABLE_LINUX_UUID', 'true'),
+            call('GRUB_DISTRIBUTOR', '"Bob"'),
+            call('GRUB_ENABLE_BLSCFG', 'true'),
+            call('GRUB_ENABLE_CRYPTODISK', 'y'),
+            call('GRUB_ENABLE_LINUX_LABEL', 'true'),
+            call('GRUB_GFXMODE', '800x600'),
+            call(
+                'GRUB_SERIAL_COMMAND', '"serial --speed=38400"'
+            ),
+            call('GRUB_TERMINAL', '"serial"'),
+            call('GRUB_THEME', '/boot/grub2/themes/openSUSE/theme.txt'),
+            call('GRUB_TIMEOUT', 10),
+            call('GRUB_TIMEOUT_STYLE', 'countdown'),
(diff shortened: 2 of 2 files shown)


kiwi/bootloader/config/grub2.py
```diff
@@ -657,6 +659,14 @@ def _setup_default_grub(self):
             r'(^root=[^\s]+)|( root=[^\s]+)', '', self.cmdline
         ).strip()
         if self.persistency_type == 'by-label':
+            label = re.search(r'(^root=[^\s]+)|( root=[^\s]+)', self.cmdline)
+            if label and label.groups()[-1]:
+                label_definition = label.groups()[-1].strip()
+                if 'LABEL' in label_definition:
+                    grub_default_entries['GRUB_CMDLINE_LINUX'] = \
+                        '"{0}"'.format(label_definition)
+                    grub_default_entries['SUSE_REMOVE_LINUX_ROOT_PARAM'] = \
```
schaefi, line 668: I assume adding SUSE specific value does not harm (os is just ignored by grub tooling) in case of other distributions ?

kiwi/bootloader/config/grub2.py
```diff
@@ -657,6 +659,14 @@ def _setup_default_grub(self):
             r'(^root=[^\s]+)|( root=[^\s]+)', '', self.cmdline
         ).strip()
         if self.persistency_type == 'by-label':
+            label = re.search(r'(^root=[^\s]+)|( root=[^\s]+)', self.cmdline)
+            if label and label.groups()[-1]:
+                label_definition = label.groups()[-1].strip()
+                if 'LABEL' in label_definition:
+                    grub_default_entries['GRUB_CMDLINE_LINUX'] = \
+                        '"{0}"'.format(label_definition)
```
schaefi, line 667: Do you have a specific reason why the ```root=LABEL=foo``` is added to the GRUB_CMDLINE_LINUX variable instead of the GRUB_CMDLINE_LINUX_DEFAULT variable ?

Am I correct that writing the ```root=``` information will disable the grub tooling to operate on the root device and simply trust the value given here ?

If so this would invalidate the setting GRUB_ENABLE_LINUX_LABEL to be effective.

Would it be possible that your code changes only apply if GRUB_ENABLE_LINUX_LABEL is not available ?
I think the scope should be limited. If you look in ```bootloader/config/grub2.py``` you see examples for which we only handle the grub option if there is an implementation for it available. e.g ```enable_blscfg_implemented = ...```

Thanks

schaefi, line 667: Do you have a specific reason why the ```root=LABEL=foo``` is added to the GRUB_CMDLINE_LINUX variable instead of the GRUB_CMDLINE_LINUX_DEFAULT variable ?

Am I correct that writing the ```root=``` information will disable the grub tooling to operate on the root device and simply trust the value given here ?

If so this would invalidate the setting GRUB_ENABLE_LINUX_LABEL to be effective.

Would it be possible that your code changes only apply if GRUB_ENABLE_LINUX_LABEL is not available ?
I think the scope should be limited. If you look in ```bootloader/config/grub2.py``` you see examples for which we only handle the grub option if there is an implementation for it available. e.g ```enable_blscfg_implemented = ...```

Thanks

schaefi (changes requested):
Hi Robert, thanks much for pull request. You touched the most fragile code with all that bootloader weirdness :) Please bear with us that we need to be very careful what we add there. Thus I have a few questions on the possible impact of the change and I suggest to limit its scope a bit more

schaefi:
@davidcassany Would you mind to look at this one too. bootloader stuff deserves more eyes to look at :) Thanks much

kiwi/bootloader/config/grub2.py
rjschwei: @schaefi I was unable to find any evidence of `GRUB_ENABLE_LINUX_LABEL` upstream looking at even the latest (version 2.6 doc) or within the SUSE version of grub (2.4). As such I was unable to trace where this is coming from and what it does. https://www.gnu.org/software/grub/manual/grub/grub.html

Based on the name it probably should do what we are after but as I couldn't find it I am at a loss.

`root=LABEL=foo` is added since `SUSE_REMOVE_LINUX_ROOT_PARAM` will remove any `root=` entry that is generated. We still need to tell the kernel how to find root.

kiwi/bootloader/config/grub2.py
rjschwei: Yes, these setting are used in the various scripts that the grub package of a distro delivers in `/etc/grub.d/` and other distros would not have any code looking for `SUSE_*` settings

rjschwei:
And as a general comment, I think other tests that set `self.bootloader.persistency_type = 'by-label'` are incorrect as they set `root=UUID` which is not the behavior we get from kiwi when `by-label` is specified. However, I wasn't certain about the intention and as such I just left those tests alone. IMHO the extra if condition I inserted as `if 'LABEL' in label_definition` should not be necessary.

schaefi:
>And as a general comment, I think other tests that set self.bootloader.persistency_type = 'by-label' are incorrect as they set root=UUID

I agree, I overlooked the mocked data for some by-label tests which still defines ```self.bootloader.cmdline = 'root=UUID=foo'``` I'll create a followup PR fixing the tests

schaefi:
@rjschwei so after all for me there is only one question open and that's why ```root=LABEL=foo``` is put to the ```GRUB_CMDLINE_LINUX``` and not to the ```GRUB_CMDLINE_LINUX_DEFAULT``` variable. Did you had a specific reason for this separation ? if not it imho would be good to stick with less variables for grub :) Thoughts ?

schaefi:
@rjschwei also noted that you are for some reason no longer in the contributor list. I added you with write permissions

rjschwei:
> ...and something is strange with the github review workflow, conversations appears twice ??

Yes, I noticed that as well and was confused at first to as in "I just answered this, didn't I?" when I saw the same questions again as I was going down the comment blocks.

rjschwei:
> @rjschwei also noted that you are for some reason no longer in the contributor list. I added you with write permissions

Thanks

rjschwei:
> > And as a general comment, I think other tests that set self.bootloader.persistency_type = 'by-label' are incorrect as they set root=UUID
> 
> I agree, I overlooked the mocked data for some by-label tests which still defines `self.bootloader.cmdline = 'root=UUID=foo'` I'll create a followup PR fixing the tests

If you plan on changing this, then the new test is not really necessary, I just added it because I was not sure of the intent of the existing `by-label` tests

rjschwei:
> @rjschwei so after all for me there is only one question open and that's why `root=LABEL=foo` is put to the `GRUB_CMDLINE_LINUX` and not to the `GRUB_CMDLINE_LINUX_DEFAULT` variable. Did you had a specific reason for this separation ? if not it imho would be good to stick with less variables for grub :) Thoughts ?

Using `GRUB_CMDLINE_LINUX` was recommended in conjunction with the use of `SUSE_REMOVE_LINUX_ROOT_PARAM`. It should also work as part of `GRUB_CMDLINE_LINUX_DEFAULT` as grub2-mkconfig is supposed to put those things together, and as far as I can tell looking at the grub script that handles this that is the case.

In any event the `root=` part that is pulled out on line 684 needs to be found and added again. Or we have to do a bit more restructuring such that 684 does not happen in all cases, only for the not `by-label` cases.

schaefi (approved):
I'm good, thanks :+1:

schaefi:
@rjschwei I'd like David to have a look at this one too and he already agreed, so let's wait for him prior merge. Thanks

schaefi:
> > > And as a general comment, I think other tests that set self.bootloader.persistency_type = 'by-label' are incorrect as they set root=UUID
> > 
> > 
> > I agree, I overlooked the mocked data for some by-label tests which still defines `self.bootloader.cmdline = 'root=UUID=foo'` I'll create a followup PR fixing the tests
> 
> If you plan on changing this, then the new test is not really necessary, I just added it because I was not sure of the intent of the existing `by-label` tests

it's ok to have an extra test for this particular case. The open PR #2110 is actually a test to check the behavior of GRUB_CMDLINE_LINUX_DEFAULT were we don't expect any root= definition. With your change merged we will allow this setting in GRUB_CMDLINE_LINUX but still don't want it in GRUB_CMDLINE_LINUX_DEFAULT. So it's good to have this behavior check as part of the tests. Once your PR is merged I have to adapt mine because new variables will appear
but GRUB_CMDLINE_LINUX_DEFAULT must not be affected.

So all good in my eyes

davidcassany (approved):
Looks good to me :+1:

Indeed `GRUB_ENABLE_LINUX_LABEL` is puzzling as it looks like it sets to force LABELs use, but does not define any label, hence it is not obvious how it should guess the root device. 

Anyway, since this affects only the defaults file and sets the `root=LABEL=foo` into the  `GRUB_CMDLINE_DEFAULT` only if this is also included as part of the KIWI's computed cmdline I believe this change is already pretty constrainted.

davidcassany:
> ...and something is strange with the github review workflow, conversations appears twice ??

Indeed, it took me a while to figure out how to read all your comments 😅

PR #2111, opened 2022-04-05
Added new CloneDevice class

schaefi:
Added CloneDevice class to the storage interface. The class allows to create clone(s) from a given source block device into a list of target block devices. The target block devices are clones of the source but prevents device naming conflicts for unique identifiers like the UUID. This is required to still allow to boot from images containing device clones and needs to be
handled by tools that might work on top of the cloned devices.


Conan-Kudo:
Is there a reason we couldn't do a btrfs subvolume clone too?

PR #2112, opened 2022-04-05
Fix LABEL detection

rjschwei:
+ When only "root=" is specified on the kernel command line the match is
  located in the first group. Loop through the groups upon mtach to find
  what we are looking for.


schaefi (approved):
Thanks looks good :+1:

PR #2120, opened 2022-04-26
Fix vagrant sshd config file

dcermak:
* Add Leap 15.4, SLE 15 SP4 & CentOS Stream 9 to the scripts tests
* Change the custom vagrant config file to `00-vagrant.conf`, If it is called `99-vagrant.conf`, then anything "before" that, like
`50-redhat.conf` takes precedence and overrides our custom settings.


schaefi:
0 = highest prio, good to know. Thanks

Issue #2097, opened 2022-03-18
Broken link in kiwi/doc/source/concept_and_workflow/customize_the_boot_process.rst


janajaeger:
## Problem description

A link has become broken in the above mentioned file: http://people.redhat.com/harald/dracut.html 
Suggested fix: Kill the link and the entire sentence pointing to it?

This also affects translated versions of the doc.

PR #2121, opened 2022-04-26
Correct the URL to the dracut home page

dcermak:
Fixes https://github.com/OSInside/kiwi/issues/2097


schaefi (approved):
Good catch, thanks :+1:

PR #2126, opened 2022-04-27
Changed LUKS type 'luks' to 'luks1'

ozboss:
The default cryptsetup `--type luks` apparently has a different meaning on different distributions. On Debian creating a volume with `--type luks` creates a luks2 volume. With this change the type is defined more explicitly and allows creating a LUKS1 volume on Debian. This way /boot can also be encrypted and unlocked with GRUB.

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -497,10 +497,11 @@ luks="passphrase|file:///path/to/keyfile":
 
 luks_version="luks|luks1|luks2":
   Specify which `LUKS` version should be used. If not set and by
-  default `luks` is used. The specification of the `LUKS` version
-  allows using a different set of `luksformat` options. To
-  investigate the differences between the two please consult the
-  `cryptsetup` manual page.
+  default `luks` is used. The interpretation of the default depends
+  on the distribution and could result in either 'luks1' or 'luks2'.
+  The specification of the `LUKS` version allows using a different
+  set of `luksformat` options. To investigate the differences between
+  the two please consult the `cryptsetup` manual page.
 
 target_blocksize="number":
   Specifies the image blocksize in bytes which has to
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -495,9 +495,9 @@ luks="passphrase|file:///path/to/keyfile":
   distribution will just open an interactive dialog asking
   for the credentials at boot time !
 
-luks_version="luks|luks2":
+luks_version="luks1|luks2":
   Specify which `LUKS` version should be used. If not set and by
-  default `luks` is used. The specification of the `LUKS` version
+  default `luks1` is used. The specification of the `LUKS` version
   allows using a different set of `luksformat` options. To
   investigate the differences between the two please consult the
   `cryptsetup` manual page.
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -495,9 +495,9 @@ luks="passphrase|file:///path/to/keyfile":
   distribution will just open an interactive dialog asking
   for the credentials at boot time !
 
-luks_version="luks1|luks2":
+luks_version="luks|luks1|luks2":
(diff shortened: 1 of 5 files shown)


ozboss:
I added back the default option. So basically the PR now extends the options by 'luks1'.

doc/source/image_description/elements.rst
```diff
@@ -495,7 +495,7 @@ luks="passphrase|file:///path/to/keyfile":
   distribution will just open an interactive dialog asking
   for the credentials at boot time !
 
-luks_version="luks|luks2":
+luks_version="luks|luks1|luks2":
```
rjschwei, line 498: IMHO the description should be updated to more explicitly state that `luks` make the image build subject to the default on the distribution.

ozboss: I extended the documentation. Is it sufficient?

schaefi: yep, thanks this looks good to me :+1:

schaefi: There is just one outstanding task. In your changes you touched the ```kiwi.rnc``` file. This is the compact RelaxNG file from which the ```kiwi.rng``` and ```xsi.xsd``` schema files will be created. Also the xml_parse.py API code is auto created from the schema. In order to make your changes effective you need to run the following in your local kiwi git checkout:

```
tox -e devel
. .tox/3/bin/activate
make kiwi/schema/kiwi.rng
```

* you can fetch python3-tox either from your distro or via pip
* once done please update the PR with the new modified files

Let me know if you run into issues

Thanks much

schaefi: @ozboss in case this extra steps bothers you too much, just let me know. We can also merge the current state and I run the API auto creation afterwards as a followup commit. Thanks for your effort

ozboss: Hey sorry I just came back from a trip. I'm actually on Manjaro right now. I found a trang package from 2018 in the AUR with outdated PGP keys. So I'll just spin up a VM and commit the changes this way. I can probably do it tomorrow but if you want to include the changes earlier go ahead :)

ozboss: Ok I managed to generate kiwi.rng :) this was quite the journey actually.

I took the chance to try out Fedora 36 as it is supported by `helper/install_devel_packages.sh`. Unfortunately the script fails as the package `texlive-updmap-map` does not exist in the repository anymore. I checked and it seems like it is also missing from the Fedora 35 repositories. When removing the package from the script it is able to installs all other packages. I don't know if the missing package will be a problem down the line, so just a heads up for you.
Also `trang` is not included in the script but it also does not exist in the repository...

So this gave me the chance to also try out the new Ubuntu release where everything work fine and also `trang` exists.
At some point I will also give an OS distro a chance, I promise ;)

If anything else is missing please let me know. Sorry for keeping you waiting.

schaefi: > Ok I managed to generate kiwi.rng :) this was quite the journey actually.

great and looks good :+1: Thanks

> helper/install_devel_packages.sh

Thanks I update the contents according to your findings. The script is really not much of a helper as distros quickly change and the information on the helper becomes outdated again. It would be nice to reduce the list and maybe find a different concept which is more stable. Actually most of the stuff there is to compile the python sphinx ReST docs into a pdf document which we place as local docs in the package. This is done via the sphinx latex target and from there to pdf via pdflatex. Unfortunately this introduces the long list of latex packages. Maybe there is a better method to provide the in-package documentation ?

Regarding trang I could find it here:

* https://packages.fedoraproject.org/pkgs/jing-trang/

looks pretty unmaintained and seems to be not part of any repo in Fedora. @Conan-Kudo Do you know more about this package ? we use trang in kiwi to convert schema formats. This works flawlessly but if there is also a better or more maintained tooling across distros I'm willing to change it. Thanks

ok, so we are a bit off topic :) @ozboss thanks a ton for the change it looks all good to me. I'm going to merge after the test run succeeds

schaefi (approved):

PR #2129, opened 2022-04-28
Fixed UUID setup for XFS

schaefi:
Make sure the log got replayed prior generating a new UUID


schaefi:
when setting a new UUID on an XFS filesystem it is required to do this on a clean xfs log. In order to ensure the log got replayed and is clean an xfs_repair call is required. In case of an unclean log the xfs_admin call just skips the UUID update but does not report this as an error

Conan-Kudo (approved):

PR #2130, opened 2022-04-28
Update get_disksize_mbytes to support clones

schaefi:
When using partition clones the pre-calculation of the
disk size needs to take this into account.


schaefi:
In preparation to the final pull request supporting partition clones this commit adds the needed code to pre-calculate disk sizes if partition clones are used

Conan-Kudo (approved):

PR #2131, opened 2022-04-29
Add support for partition cloning

schaefi:
Support creating block level clones of certain partitions
used in the image. Clones can be created from the root, boot
and any partition listed in the ```<partitions>``` element.

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -485,6 +485,20 @@ kernelcmdline="string":
   Additional kernel parameters passed to the kernel by the
   bootloader.
 
+root_clone="number"
+  For oem disk images, this attribute allows to create `number`
+  clone(s) of the root partition, with `number` >= 1. A clone partition
+  is content wise an exact byte for byte copy of the origin root partition.
+  However, to avoid conflicts at boot time the UUID of any
+  cloned partition will be made unique. In the sequence of partitions,
+  the clone(s) will always be created first followed by the
+  partition considered the origin. The origin partition is the
+  one that will be referenced and used by the system.
+  Also see :ref:`clone_partitions`
+
+boot_clone="number"
+  Same as `root_clone` but applied to the boot partition if present
+
 luks="passphrase|file:///path/to/keyfile":
   Supplying a value will trigger the encryption of the partition
   serving the root filesystem using the LUKS extension. The supplied
--- a/doc/source/working_with_images/clone_partitions.rst
+++ b/doc/source/working_with_images/clone_partitions.rst
@@ -1,0 +1,128 @@
+.. _clone_partitions:
+
+Partition Clones
+================
+
+.. sidebar:: Abstract
+
+   This page provides details about the partition clone feature
+   and its use cases
+
+{kiwi} allows to create block level clones of certain partitions
+used in the image. Clones can be created from the `root`, `boot`
+and any other partition listed in the `<partitions>` element.
+
+A partition clone is a simple byte dump from one
+block storage device to another. However, this would cause conflicts
+during boot of the system because all unique identifiers like
+the UUID of a filesystem will no longer be unique. The clone
+feature of {kiwi} takes care of this part and re-creates the
+relevant unique identifiers per cloned partition. {kiwi} allows
+this also for complex partitions like LVM, LUKS or RAID.
+
+The partition clone(s) will always appear first in the partition table,
+followed by the origin partition. The origin partition is the one
+whose identifier will be referenced and used by the system. By default
+no cloned partition will be mounted or used by the system at boot time.
+
+Let's take a look at the following example:
+
+.. code:: xml
+
+   <type image="oem" root_clone="1" boot_clone="1" firmware="uefi" filesystem="xfs" bootpartition="true" bootfilesystem="ext4">
+       <partitions>
+           <partition name="home" size="10" mountpoint="/home" filesystem="ext3" clone="2"/>
+       </partitions>
+   </type>
+
+With the above setup {kiwi} will create a disk image that
+contains the following partition table:
+
+.. code::
+
+   Number  Start (sector)    End (sector)  Size       Code  Name
+      1            2048            6143   2.0 MiB     EF02  p.legacy
+      2            6144           47103   20.0 MiB    EF00  p.UEFI
+      3           47104          661503   300.0 MiB   8300  p.lxbootclone1
+      4          661504         1275903   300.0 MiB   8300  p.lxboot
+      5         1275904         1296383   10.0 MiB    8300  p.lxhomeclone1
+      6         1296384         1316863   10.0 MiB    8300  p.lxhomeclone2
+      7         1316864         1337343   10.0 MiB    8300  p.lxhome
+      8         1337344         3864575   1.2 GiB     8300  p.lxrootclone1
+      9         3864576         6287326   1.2 GiB     8300  p.lxroot
+
+When booting the system only the origin partitions `p.lxboot`, `p.lxroot`
+and `p.lxhome` will be mounted and visible in e.g. :file:`/etc/fstab`,
+the bootloader or the initrd. Thus partition clones are present as a data
+source but are not relevant for the operating system from a functional
+perspective.
+
+As shown in the above example there is one clone request for root and boot
+and a two clone requests for the home partition. {kiwi} does not sanity-
+check the provided number of clones (e.g. whether your partition table
+can hold that many partitions).
+
+.. warning::
+
+   There is a limit how many partitions a partition table can hold.
+   This also limits how many clones can be created.
+
+Use Case
+--------
+
+Potential use cases for which a clone of one or more partitions
+is useful include among others:
+
+Factory Resets:
+  Creating an image with the option to rollback to the
+  state of the system at deployment time can be very helpful
+  for disaster recovery
+
+System Updates with Rollbacks e.g A/B:
+  Creating an image which holds extra space allowing to rollback
+  modified data can make a system more robust. For example
+  in a simple A/B update concept, partition A would get updated
+  but would flip to B if A is considered broken after applying the
+  update.
+
+.. note::
+
+   Most probably any use case based on partition clones requires
+   additional software to manage them. {kiwi} provides the
+   option to create the clone layout but it does not provide
+   the software to implement the actual use case for which the
+   partition clones are needed.
+
+Developers writing applications based on a clone layout created
+with {kiwi} can leverage the metadata file :file:`/config.partids`.
+This file is created at build time and contains the mapping between
+the partition `name` and the actual partition number in the partition
+table. For partition clones, the following naming convention applies:
+
+.. code::
+
+   kiwi_(name)PartClone(id)="(partition_number)"
+
+The `(name)` is either taken from the `name` attribute
+of the `<partition>` element or it is a fixed name assigned by {kiwi}.
+There are the following reserved partition names for which cloning
+is supported:
+
+* root
+* readonly
+* boot
+
+For the mentioned example this will result in the
+following :file:`/config.partids`:
+
+.. code::
+
+   kiwi_BiosGrub="1"
+   kiwi_EfiPart="2"
+   kiwi_bootPartClone1="3"
+   kiwi_BootPart="4"
+   kiwi_homePartClone1="5"
+   kiwi_homePartClone2="6"
+   kiwi_HomePart="7"
+   kiwi_rootPartClone1="8"
+   kiwi_RootPart="9"
--- a/doc/source/working_with_images/custom_partitions.rst
+++ b/doc/source/working_with_images/custom_partitions.rst
@@ -71,6 +71,16 @@ filesystem="btrfs|ext2|ext3|ext4|squashfs|xfs
   Mandatory filesystem configuration to create one of the supported
   filesystems on the partition.
 
+clone="number"
+  Optional setting to indicate that this partition should be
+  cloned `number` of times. A clone partition is content wise an
+  exact byte for byte copy of the origin. However, to avoid conflicts at boot
+  time the UUID of any cloned partition will be made unique. In the
+  sequence of partitions, the clone(s) will always be created first
+  followed by the partition considered the origin. The origin
+  partition is the one that will be referenced and used by the
+  system
+
 Despite the customization options of the partition table shown above
 there are the following limitations:
 
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -36,6 +36,7 @@ from kiwi.system.identifier import SystemIdentifier
 from kiwi.boot.image import BootImage
 from kiwi.storage.setup import DiskSetup
 from kiwi.storage.loop_device import LoopDevice
+from kiwi.storage.clone_device import CloneDevice
 from kiwi.firmware import FirmWare
 from kiwi.storage.disk import Disk
 from kiwi.storage.raid_device import RaidDevice
@@ -105,6 +106,10 @@ class DiskBuilder:
             xml_state.build_type.get_embed_verity_metadata()
         self.dosparttable_extended_layout = \
             xml_state.build_type.get_dosparttable_extended_layout()
+        self.boot_clone_count = int(xml_state.build_type.get_boot_clone()) \
+            if xml_state.build_type.get_boot_clone() else 0
+        self.root_clone_count = int(xml_state.build_type.get_root_clone()) \
+            if xml_state.build_type.get_root_clone() else 0
         self.custom_root_mount_args = xml_state.get_fs_mount_option_list()
         self.custom_root_creation_args = xml_state.get_fs_create_option_list()
         self.build_type_name = xml_state.get_build_type_name()
@@ -235,7 +240,7 @@ class DiskBuilder:
 
         # a list of instances with the sync_data capability
         # representing the custom partitions area of the disk
-        system_custom_parts: List[FileSystemBase] = []
+        system_custom_parts: Dict[str, FileSystemBase] = {}
 
         if self.install_media and self.build_type_name != 'oem':
             raise KiwiInstallMediaError(
@@ -258,7 +263,9 @@ class DiskBuilder:
             self.boot_image.prepare()
 
         # precalculate needed disk size
-        disksize_mbytes = self.disk_setup.get_disksize_mbytes()
+        disksize_mbytes = self.disk_setup.get_disksize_mbytes(
+            root_clone=self.root_clone_count, boot_clone=self.boot_clone_count
+        )
 
         # create the disk
         log.info('Creating raw disk image %s', self.diskname)
@@ -736,8 +743,8 @@ class DiskBuilder:
     def _build_custom_parts_filesystem(
         self, device_map: Dict,
         custom_partitions: Dict['str', ptable_entry_type]
-    ) -> List[FileSystemBase]:
-        filesystem_list = []
+    ) -> Dict[str, FileSystemBase]:
+        filesystem_dict = {}
         if custom_partitions:
             for map_name in sorted(custom_partitions.keys()):
                 if map_name in device_map:
@@ -751,8 +758,8 @@ class DiskBuilder:
                         filesystem.create_on_device(
                             label=map_name.upper()
                         )
-                        filesystem_list.append(filesystem)
-        return filesystem_list
+                        filesystem_dict[map_name] = filesystem
+        return filesystem_dict
 
     def _build_spare_filesystem(self, device_map: Dict) -> Optional[FileSystemBase]:
         if 'spare' in device_map and self.spare_part_fs:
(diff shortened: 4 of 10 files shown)


doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+  Creating an image which holds an extra space to allow rollback
+  of modified data can make a system more robust. For example
+  in a simple A/B update concept, the partition A can change
+  and could flip to B if A is considerd broken after the update.
+
+.. note::
+
+   Most probably any use case based on partition clones requires
+   additional software to manage them. {kiwi} provides the
+   opportunity to create the clone layout but it does not provide
```
dcermak, line 87:
```suggestion
   option to create the clone layout but it does not provide
```

doc/source/working_with_images/custom_partitions.rst
```diff
@@ -71,6 +71,16 @@ filesystem="btrfs|ext2|ext3|ext4|squashfs|xfs
   Mandatory filesystem configuration to create one of the supported
   filesystems on the partition.
 
+clone="number"
+  Optional setting to indicate that this partition should be
+  cloned `number` times. A clone partition is content wise an
```
dcermak, line 76:
```suggestion
  cloned `number` of times. A clone partition is content wise an
```

doc/source/working_with_images/custom_partitions.rst
```diff
@@ -71,6 +71,16 @@ filesystem="btrfs|ext2|ext3|ext4|squashfs|xfs
...
+  exact byte copy of the origin. However, to avoid conflicts at boot
```
dcermak, line 77:
```suggestion
  exact byte for byte copy of the origin. However, to avoid conflicts at boot
```

doc/source/working_with_images/custom_partitions.rst
```diff
@@ -71,6 +71,16 @@ filesystem="btrfs|ext2|ext3|ext4|squashfs|xfs
...
+  time the UUID of any cloned partition will be unique. In the
```
dcermak, line 78:
```suggestion
  time the UUID of any cloned partition will be made unique. In the
```

doc/source/image_description/elements.rst
```diff
@@ -485,6 +485,20 @@ kernelcmdline="string":
   Additional kernel parameters passed to the kernel by the
   bootloader.
 
+root_clone="number"
+  For oem disk images, specifies to create `number` clone(s) of
```
dcermak, line 489:
```suggestion
  For oem disk images, this element allows to create `number` clone(s) of
```

doc/source/image_description/elements.rst
```diff
@@ -485,6 +485,20 @@ kernelcmdline="string":
...
+  the root partition, with `number` >= 1. A clone partition is
+  content wise an exact byte copy of the origin root partition.
```
dcermak, line 491:
```suggestion
  content wise an exact byte for byte copy of the origin root partition.
```

doc/source/image_description/elements.rst
```diff
@@ -485,6 +485,20 @@ kernelcmdline="string":
...
+  However, to avoid conflicts at boot time the UUID of any
+  cloned partition will be unique. In the sequence of partitions,
```
dcermak, line 493:
```suggestion
  cloned partition will be made unique. In the sequence of partitions,
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+.. sidebar:: Abstract
+
+   This page provides details about the partition clone feature
+   and its use cases
+
+{kiwi} allows to create block level clones of certain partitions
+used in the image. Clones can be created from the `root`, `boot`
+and any partition listed in the `<partitions>` element.
+
+Technically a partition clone is a simple byte dump from one
```
dcermak, line 15:
```suggestion
A partition clone is a simple byte dump from one
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+block storage device to another. However, this is causing conflicts
```
dcermak, line 16:
```suggestion
block storage device to another. However, this would cause conflicts
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+at boot time of the system because all unique identifiers like
```
dcermak, line 17:
```suggestion
during boot of the system because all unique identifiers like
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+the UUID of a filesystem will be no longer unique. The clone
```
dcermak, line 18:
```suggestion
the UUID of a filesystem will no longer be unique. The clone
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+
+.. code:: xml
+
+   <type image="oem" root_clone="1" boot_clone="1" firmware="uefi" filesystem="xfs" bootpartition="true" bootfilesystem="ext4">
+       <partitions>
+           <partition name="home" size="10" mountpoint="/home" filesystem="ext3" clone="2"/>
+       </partitions>
+   </type>
+
+With the above type setup {kiwi} will create a disk image that
```
dcermak, line 39:
```suggestion
With the above setup {kiwi} will create a disk image that
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+      6         1296384         1316863   10.0 MiB    8300  p.lxhomeclone2
+      7         1316864         1337343   10.0 MiB    8300  p.lxhome
+      8         1337344         3864575   1.2 GiB     8300  p.lxrootclone1
+      9         3864576         6287326   1.2 GiB     8300  p.lxroot
+
+On boot of the system only the origin partitions `p.lxboot`, `p.lxroot`
+and `p.lxhome` will be mounted and visible in for example fstab,
+bootloader, initrd, etc. Thus partition clones are present as a data
+source but not relevant for the operating system from a functional
+perspective.
```
dcermak, line 59:
```suggestion
When booting the system only the origin partitions `p.lxboot`, `p.lxroot`
and `p.lxhome` will be mounted and visible in e.g. `/etc/fstab`,
the bootloader or the initrd. Thus partition clones are present as a data
source but are not relevant for the operating system from a functional
perspective.
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+
+As shown in the example there is one clone request for root and boot
+and a two clones request for the home partition. {kiwi} does not
+check on the given number of clones but there is a limitation how many
+partitions a partition table can hold. This limitation also limits
+how many clones can be created.
```
dcermak, line 65:
```suggestion
As shown in the above example there is one clone request for root and boot
and a two clone requests for the home partition. {kiwi} does not
check on the given number of clones.

.. warning:: 
    There is a limit how many partitions a partition table can hold. This also 
    limits how many clones can be created.
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Use Case
+--------
+
+Possible use cases for which a clone of one or more partitions
+is useful could be for example:
```
dcermak, line 71:
```suggestion
Potential use cases for which a clone of one or more partitions
is useful include among others:
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Factory Reset:
+  Creating an image with the opportunity to rollback to the
+  state of the system at deployment time can be quite handy
```
dcermak, line 75:
```suggestion
Factory Resets:
  Creating an image with the option to rollback to the
  state of the system at deployment time can be very helpful
  for disaster recovery
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Update Concepts with Rollback e.g A/B:
+  Creating an image which holds an extra space to allow rollback
+  of modified data can make a system more robust. For example
+  in a simple A/B update concept, the partition A can change
+  and could flip to B if A is considerd broken after the update.
```
dcermak, line 81:
```suggestion
System Updates with Rollbacks e.g A/B:
  Creating an image which holds extra space allow to rollback
  modified data can make a system more robust. For example
  in a simple A/B update concept, partition A would get updated
  but would flip to B if A is considered broken after applying the
  update.
```

dcermak (commented):
The docs are very good (that's also all that I checked), I have added some suggestions where the wording could be improved imho.

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+
+Technically a partition clone is a simple byte dump from one
+block storage device to another. However, this is causing conflicts
+at boot time of the system because all unique identifiers like
+the UUID of a filesystem will be no longer unique. The clone
+feature in {kiwi} takes care for this part and re-creates the
+relevant unique identifiers per clone partition. {kiwi} allows
+this also for complex partitions like LVM, crypto or Raid members.
+
+In the sequence of partitions the clone(s) will always be created
```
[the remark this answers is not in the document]
schaefi: I wanted to express the following. If you request two clones you have a sequence of three partitions: cloneA followed by cloneB followed by origin. I wanted to express that in this chain the clones comes first and origin is last

schaefi:
> The docs are very good (that's also all that I checked), I have added some suggestions where the wording could be improved imho.

Thanks I had hoped you would review that part. I updated the docs according to your suggestions, is a much better read now. Thanks

doc/source/working_with_images/clone_partitions.rst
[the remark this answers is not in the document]
dcermak: Yeah, I got that the clones come first. But where do they come first, in the partition table or just some arbitrary list?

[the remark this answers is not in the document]
schaefi: in the partition table

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,89 @@
...
+first followed by the partition considered the origin. The origin
```
dcermak, line 24: How about this?
```suggestion
The partition clone(s) will always appear first in the partition table,
followed by the origin partition. The origin
```

schaefi: ok that should make it clear, done

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,93 @@
...
+================
+
+.. sidebar:: Abstract
+
+   This page provides details about the partition clone feature
+   and its use cases
+
+{kiwi} allows to create block level clones of certain partitions
+used in the image. Clones can be created from the `root`, `boot`
+and any partition listed in the `<partitions>` element.
```
dcermak, line 13:
```suggestion
and any other partition listed in the `<partitions>` element.
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,93 @@
...
+
+A partition clone is a simple byte dump from one
+block storage device to another. However, this would cause conflicts
+during boot of the system because all unique identifiers like
+the UUID of a filesystem will no longer be unique. The clone
+feature in {kiwi} takes care of this part and re-creates the
+relevant unique identifiers per cloned partition. {kiwi} allows
+this also for complex partitions like LVM, LUKS or RAID members.
```
dcermak, line 21:
```suggestion
feature of {kiwi} takes care of this part and re-creates the
relevant unique identifiers per cloned partition. {kiwi} allows
this also for complex partitions like LVM, LUKS or RAID.
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,93 @@
...
+      3           47104          661503   300.0 MiB   8300  p.lxbootclone1
+      4          661504         1275903   300.0 MiB   8300  p.lxboot
+      5         1275904         1296383   10.0 MiB    8300  p.lxhomeclone1
+      6         1296384         1316863   10.0 MiB    8300  p.lxhomeclone2
+      7         1316864         1337343   10.0 MiB    8300  p.lxhome
+      8         1337344         3864575   1.2 GiB     8300  p.lxrootclone1
+      9         3864576         6287326   1.2 GiB     8300  p.lxroot
+
+When booting the system only the origin partitions `p.lxboot`, `p.lxroot`
+and `p.lxhome` will be mounted and visible in e.g. `/etc/fstab`,
```
dcermak, line 55:
```suggestion
and `p.lxhome` will be mounted and visible in e.g. :file:`/etc/fstab`,
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,93 @@
...
+the bootloader or the initrd. Thus partition clones are present as a data
+source but are not relevant for the operating system from a functional
+perspective.
+
+As shown in the above example there is one clone request for root and boot
+and a two clone requests for the home partition. {kiwi} does not
+check on the given number of clones.
```
dcermak, line 62:
```suggestion
and a two clone requests for the home partition. {kiwi} does not sanity-
check the provided number of clones (e.g. whether your partition table
can hold that many partitions).
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -0,0 +1,93 @@
...
+Potential use cases for which a clone of one or more partitions
+is useful include among others:
+
+Factory Resets:
+  Creating an image with the option to rollback to the
+  state of the system at deployment time can be very helpful
+  for disaster recovery
+
+System Updates with Rollbacks e.g A/B:
+  Creating an image which holds extra space allow to rollback
```
dcermak, line 81:
```suggestion
  Creating an image which holds extra space allowing to rollback
```

dcermak (changes requested):
The docs are pretty much good to go, I've just flagged a few nitpicks that I missed in the first pass.

I have one question about this feature: are the partition names somehow predictable upfront? I guess this could be useful to pre-configure any software leveraging the clone partitions.

schaefi:
> I have one question about this feature: are the partition names somehow predictable upfront? I guess this could be useful to pre-configure any software leveraging the clone partitions.

It depends on the type of partition table. In GPT the partition table support GUID and kiwi uses this to name the partitions. You can see this in the docs for example ```p.lxboot``` kiwi applies a naming convention on the partition names in this case
which is "p.[lx][name]"  lx for linux partition (there are no others atm) and name follows a list of static names for certain partitions like boot, root, lvm ... currently only documented by source code ;)

On the clone partitions the same naming convention was used and I just appended cloneNr at the end of the partition name.
So a ```p.lxboot``` is always ```p.lxbootclone1```, ```p.lxbootclone2``` and so on

But I would not mention this as something really predictable because it depends on the partition table type and currently only GPT has this opportunity.

schaefi:
The probably more generic and predictable way to know which partition serves which purpose is by the kiwi written ```config.partids``` file. It contains a mapping from name to partition number and this is independent of the capabilities of the actual partition table. The custom tooling around such a partition layout would have an opportunity to make use of this information.

schaefi:
> > The probably more generic and predictable way to know which partition serves which purpose is by the kiwi written `config.partids` file. It contains a mapping from name to partition number and this is independent of the capabilities of the actual partition table. The custom tooling around such a partition layout would have an opportunity to make use of this information.
> 
> `configs.partids` sounds like something that I'd recommend mentioning in the docs as well, as it could be pretty useful in such a situation.

I added information about that part

kiwi/builder/disk.py
```diff
@@ -229,7 +234,7 @@ def create_disk(self) -> Result:
 
         # a list of instances with the sync_data capability
         # representing the custom partitions area of the disk
-        system_custom_parts: List[FileSystemBase] = []
+        system_custom_parts: Dict[str, FileSystemBase] = {}
```
davidcassany, line 237: Was this list expected to be in some order? I saw in some parts a `sort` is applied over keys, however keys are strings, hence I assume after sorting keys the results is not in the same order as it was the former `List[FileSystemBase]`.

I am just wondering there could be an error caused by looping through partitions in a different order to what was formerly expected.

davidcassany (approved):
Looks good to me :+1: I think I'd struggle to find an code a use case for these though.

Just to brainstorm a little. I imagined the use case of cloning partitions (or creating partition images) for a factory reset from a KIWI PoV a bit different. I imagined having the capability of defining dual bootable root-trees in KIWI so like composing two roots from  a single set of preferences (e.g. having two bootstraps and two image packages list), in that way we could imagine one system to be a really streamed down JeOS  including a clone of the main (or default) root partition.

Another idea I had at a time was having some sort of squashedfs partition clones, so you could eventually dump them into a partition and boot it as a live system. So something like what is here in this PR but the cloned partition image embedded and compressed as squashfs and then squashedfs image dumped a disk partition. This way there could be a relevant size reduction of the redundant data. 

Just some ideas around the recovery or factory reset story

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -92,3 +92,38 @@ System Updates with Rollbacks e.g A/B:
...
+the following naming convention applies:
+
+.. code::
+
+   kiwi_(name)PartClone(id)="(partition_number)"
+
+The `(name)` information is either taken from the `name` attribute
+of the `<partition>` element or it is a fixed name assigned by {kiwi}.
+There are the following reserved partition names from which cloning
+is supported:
```
dcermak, line 110:
```suggestion
The `(name)` is either taken from the `name` attribute
of the `<partition>` element or it is a fixed name assigned by {kiwi}.
There are the following reserved partition names for which cloning
is supported:
```

doc/source/working_with_images/clone_partitions.rst
```diff
@@ -92,3 +92,38 @@ System Updates with Rollbacks e.g A/B:
    option to create the clone layout but it does not provide
    the software to implement the actual use case for which the
    partition clones are needed.
+
+Developers writing applications based on a clone layout created
+with {kiwi} can consider to take the metadata file :file:`/config.partids`
+into account. At build time of the image this file exists and
+creates a mapping between the partition `name` and the actual
+partition number in the partition table. For partition clones,
+the following naming convention applies:
```
dcermak, line 101:
```suggestion
Developers writing applications based on a clone layout created
with {kiwi} can leverage the metadata file :file:`/config.partids`.
This file is created at build time and contains the mapping between
the partition `name` and the actual partition number in the partition
table. For partition clones, the following naming convention applies:
```

dcermak (approved):

kiwi/builder/disk.py
schaefi: > Was this list expected to be in some order?

This is a good question. Actually the former list and now the dict contains the FileSystem objects matching each ```<partition>``` element from ```<partitions>```. When syncing the data into the partition device the order does not
matter. However, I agree with you the loop should be sorted such that the iteration is deterministic

While checking on your concern I found there can be another problem caused when using a nested partition
setup. Assuming the following setup:

```xml
<partitions>
    <partition name="usr" size="100" mountpoint="/usr" filesystem="ext3"/>
    <partition name="usr_lib" size="100" mountpoint="/usr/lib" filesystem="ext3"/>
</partitions>
```

It doesn't matter in which order ```/usr``` is synced to ```/dev/some-device-a``` and ```/usr/lib``` to ```/dev/some-device-b``` but the ```usr``` should exclude ```usr/lib``` which is currently not implemented for custom partitions. We have that implemented for volumes (lvm and btrfs) but not for the custom partitions.

This is not related to the clone feature imho but an issue on the custom partitions support when using nested partitions

On behalf of your concern I will make sure the loop runs sorted

schaefi:
> Another idea I had at a time was having some sort of squashedfs partition clones, so you could eventually dump them into a partition and boot it as a live system. So something like what is here in this PR but the cloned partition image embedded and compressed as squashfs and then squashedfs image dumped to a disk partition. This way there could be a relevant size reduction of the redundant data.
> 
> Just some ideas around the recovery or factory reset story

Yeah valuable ideas :+1: 

Regarding the use case there is one which was the reason why I started the implementation. The image I'm providing to the car manufacturer is using a squashfs compressed root filesystem. The partition layout offers an extra data partition which is the only space you can write. The concept of an update of the root filesystem is based on the A/B concept. This implies you need the same partition you are currently using for the rootfs twice. In the initial drop of the image both partitions must be able to be used to run the system, and that's where cloning provides both requirements at the same time. There is the space to drive the update and A and B can do the same. In contrast to the server world the storage capacity on a car system is a fixed value and also not considered cheap. This means an update of the OS must fit into the same size constraints which were set at the time the image for this customer was created. Assuming this is a 128MB flash disk the image will built for this size.

Now if there is a, let's say security update to the kernel, we get a new kernel package, the image rebuilds and you could extract the rootfs partition binary dump from the new image and feed that into the upgrade process that exists in the car. The upgrade process will feed partition B, does some sanity checks and flips the system. The data partition stays untouched.

If everything went well we get the success state back and mark the new image build as sort of security release version X. Over time several image builds will be produced and we can keep them including the history of changes, which is as a supplier also a requirement.

So this is the use case I'm going to really apply this feature to with reference to the automotive processes

Regarding factory reset I liked the ease of use. I mean you just say ```root_clone=1``` and you get a copy of your root filesystem data at the time you have created the image. This opens some different ways to reset the system to its factory
state. It can be either done on a binary base (directly working with the partitions) but it can also be done on a filesystem
base by mounting the clone partition and run sort of a file based restore. Even if it's just used to diff with factory the data to do it would be present.

Since I could imagine at least a few more use cases than just the one for which I need it, I thought it was worth introducing this as a generic feature.

> I imagined having the capability of defining dual bootable root-trees in KIWI so like composing two roots from a single set of preferences (e.g. having two bootstraps and two image packages list), in that way we could imagine one system to be a really streamed down JeOS including a clone of the main (or default) root partition.

I was thinking into this direction first, so basically a level up by describing content multiple times. It would allow for more flexibility on describing what content should be in the tree. In the end the reason why I did not follow that path was basically
simplicity and some laziness ;) I can imagine having the opportunity to build multiple root-trees from the same preferences could be yet another cool feature

PR #2132, opened 2022-05-02
Add support for dm integrity with secret key

schaefi:
Add support for dm integrity with secret key
    
Allow to protect the opening of the integrity data map and journal through a keyfile. For setting the key file two new optional type attributes were added:
    
* integrity_keyfile
* integrity_metadata_key_description
    
The key file format must be correct according to the selected integrity algorithm. As of now the kiwi default ```hmac-sha256``` algorithm is used with the selected keyfile
    
The optional integrity_metadata_key_description attribute allows to specify a custom description of an integrity key as it is expected to be present in the kernel keyring. The information is placed in the integrity metadata block. If not specified kiwi creates a key argument string instead which is based on the given integrity_keyfile filename. The format of this key argument is:
    
```:BASENAME_OF_integrity_keyfile_WITHOUT_FILE_EXTENSION```

### CODE REVIEW

File: doc/source/image_description/elements.rst [hunk 1]
```diff
@@ -565,11 +565,22 @@ devicepersistency="by-uuid|by-label":
...
+integrity_keyfile="filepath":
+  For the `oem` type only and in combination with the `standalone_integrity`
+  attribute, protects access to the integrity map using the given keyfile.
+  The attribute `integrity_keytyp` allows to specify the format of the key
+  stored in the keyfile. If not specified the default format is set to `hmac`.
+
+integrity_keytyp="hmac|jmac|jcrypt":
+  For the `oem` type only and in combination with the `standalone_integrity`
+  attribute and a `integrity_keyfile`, specifies the format type of the
+  key in the keyfile.
```
raimue, line 581: `integrity_keytype` is misspelled twice as `integrity_keytyp`.

kiwi/storage/integrity_device.py
```diff
@@ -136,10 +160,19 @@ def create_integrity_metadata(self) -> None:
...
             )
-
-            parameters = [
-                f'internal_hash:{self.integrity_algorithm}'
-            ] + list(integrity_superblock['flags'])
+            if self.key:
+                secrettype = \
+                    f'{self.key.secrettype}({defaults.INTEGRITY_ALGORITHM})'
+                keyplaceholder = \
+                    f':dm-integrity-{self.key.secrettype}-secret'
```
raimue, line 167: This string is supposed to be configurable as a key description in the kernel keyring. It should be an input string and it does not always have to be formatted like this which was just an example.

raimue (changes requested):
Unfortunately this key type does not make sense... There are [three options for dm-integrity](https://www.kernel.org/doc/html/latest/admin-guide/device-mapper/dm-integrity.html) that optionally take keys: `internal_hash:`, `journal_crypt:`, `journal_mac:`. They are not of different type, but for different purpose.

For standalone operation, we only want to pass a key for the `internal_hash:` and can forget about the others. The algorithm given to `internal_hash:` must always be like `hmac(sha256)` for cryptographic authentication.

schaefi:
> Unfortunately this key type does not make sense...

yes I realized I did something stupid and will rework the implementation. Thanks for pointing this out

schaefi:
@raimue ok, I reworked the implementation and added operation modes as I see them tested here:
https://fossies.org/linux/cryptsetup/tests/integrity-compat-test

* There is the simple case, integritysetup without any key... we already had that using sha256 hashing
* There is the case, integritysetup with an hmac key referenced by integrity_keyfile="..." we support hmac-sha256 currently a fixed setting in the kiwi defaults.
* There is the case, integritysetup with an hmac and a jmac key referenced by integrity_keyfile="..." and integrity_journal_keyfile="..." we support hmac-sha256 for both currently a fixed setting in the kiwi defaults

I was able to build with integritysetup for these cases. But there are several questions and issues I found:

1. I tried to create an encrypted journal and all I did failed, which is also the reason why I currently spared out any implementation for it. The way the cryptsetup test page tested this did not work for me and I also spent some time on different parameters, none were successful.

2. Opening an integrity map with a key for the journal (jmac) requires this key to be referenced for for the open call. As you can see in the current implementation uses ```--integrity-key-file``` plus ```--journal-integrity-key-file``` if both keys are specified. The cryptsetup test page does not combine them and I thought it might be an unsupported mode, but it isn't, meaning integritysetup works with that specification. However ```/etc/integritytab``` only allows to specify one key and it also seems that exclusive operation modes. So my question here is, does this make sense at all ? I assume you either want integrity or journal-integrity ? I was confused about that part and at least for systemd it cannot not be configured.

Can we clarify the expectation on that part first ? Thanks

schaefi:
@raimue regarding the ```internal_hash:``` specification if a key is in use I'm confused. The current implementation constructs a cominit specific format so to say. I think this is ok because currently there is only cominit as the tool reading the metadata header. Thus we define the standard so to say. 

From that perspective the key description is:

```
internal_hash:hmac(sha256):KEY_FILE_NAME_POLICY
```

The first part ```hmac(sha256)``` is afaik the way how the kernel requires the information about the cipher and in the implementation this is constructed out of the algorithm name ```hmac-sha256```. As we currently have this setting fixed as a defaults it was relatively easy to reformat but I can imagine this is not a general rule ?

The second part ```KEY_FILE_NAME_POLICY``` is a naming policy for the keyfile as cominit expects it. Currently this is set to the basename of the given keyfile stripped by the ```.bin``` extensions and prefixed with a ```:```

I'd like to understand if you think this concept is generic enough. As you said you wanted to have this key description as an input string I was not sure what exactly you had in mind. I think it should be constructed according to the provided information and keyfiles. Given I'm only testing with that one algorithm type ```hmac-sha256``` I'm not sure if we are generic enough. However, looking at the cryptsetup test code there is only one cipher with key.

So from my perspective should be ok

Thoughts ?

kiwi/storage/integrity_device.py
schaefi: I started a followup conversation about this part below

raimue:
>     1. I tried to create an encrypted journal and all I did failed, which is also the reason why I currently spared out any implementation for it. The way the cryptsetup test page tested this did not work for me and I also spent some time on different parameters, none were successful.

Sorry, I also did not attempt to set up anything with encrypted journal... With `internal_hash:` and `journal_crypt:`, we would encrypt the journal, but the data itself will be unencrypted on the block device. As I read the documentation, `journal_crypt:` is supposed to protect against attacks on a stacked dm-crypt. But in standalone mode, this does not seem useful to me.
 
>     2. Opening an integrity map with a key for the journal (jmac) requires this key to be referenced for for the open call. As you can see in the current implementation uses `--integrity-key-file` plus `--journal-integrity-key-file` if both keys are specified. The cryptsetup test page does not combine them and I thought it might be an unsupported mode, but it isn't, meaning integritysetup works with that specification. However `/etc/integritytab` only allows to specify one key and it also seems that you want exclusive operation modes. So my question here is, does this make sense at all ? I assume you either want integrity or journal-integrity ? I was confused about that part and at least for systemd it cannot not be configured.

As far as I understand it, it is possible to specify a separate key for `journal_mac:`, but it is not needed when also using `internal_hash:` as the same key will be used for both.

From the [kernel device-mapper documentation](https://www.kernel.org/doc/html/latest/admin-guide/device-mapper/dm-integrity.html):

> journal_mac:algorithm(:key) (the key is optional)
>
>    [...]
>
>    This option is not needed when using internal-hash because in this mode, the integrity of journal entries is checked when replaying the journal. Thus, modified sector number would be detected at this stage.

Therefore I do not think we need to specify `journal_mac:` together with `internal_hash:`, as the integrity of the journal is already protected. Using only `journal_mac:` also does not seem useful to me. Then only the journal would be protected, but not the actual data? I think a separate `journal_mac:` also only makes sense together with a stacked dm-crypt.

raimue:
> The first part `hmac(sha256)` is afaik the way how the kernel requires the information about the cipher and in the implementation this is constructed out of the algorithm name `hmac-sha256`. As we currently have this setting fixed as a defaults it was relatively easy to reformat but I can imagine this is not a general rule ?

I guess cryptsetup/integritysetup added this mapping to make the command line interface more shell-friendly by leaving out any special characters. The mapping is defined here in the code:
https://gitlab.com/cryptsetup/cryptsetup/-/blob/ede2a8a45fbdf7f8aedb7047e68d78bb915648bb/lib/utils_crypt.c#L97-130

> The second part `KEY_FILE_NAME_POLICY` is a naming policy for the keyfile as cominit expects it. Currently this is set to the basename of the given keyfile stripped by the `.bin` extensions and prefixed with a `:`
> 
> I'd like to understand if you think this concept is generic enough. As you said you wanted to have this key description as an input string I was not sure what exactly you had in mind. I think it should be constructed according to the provided information and keyfiles. Given I'm only testing with that one algorithm type `hmac-sha256` I'm not sure if we are generic enough. However, looking at the cryptsetup test code there is only one cipher with key.

The main thing is that this is not meant to refer to any file, but it is a key description in the kernel keyring. We chose the leading colon character as this is also what [dm-crypt](https://www.kernel.org/doc/html/latest/admin-guide/device-mapper/dm-crypt.html) already uses to make a `<key>` argument refer to the kernel keyring.

There is no strict relation between the name of the key file used to format the dm-integrity block device at image generation and this key description used at runtime. In the end we will expect this key description to be found in the kernel keyring after boot, but we do not yet know how much control we will have over this string.

Therefore I think it would make sense to allow specification of this key description in the input. Maybe make that something like `integrity_metadata_keydesc="..."` because it is only used with the embedded metadata? If this key description is not specified, it could still be deduced from the `integrity_keyfile` filename by stripping the extension.

schaefi:
> Therefore I think it would make sense to allow specification of this key description in the input. Maybe make that something like integrity_metadata_keydesc="..." because it is only used with the embedded metadata? If this key description is not specified, it could still be deduced from the integrity_keyfile filename by stripping the extension.

ok I like this suggestion

schaefi:
> Therefore I do not think we need to specify journal_mac: together with internal_hash:, as the integrity of the journal is already protected. Using only journal_mac: also does not seem useful to me. Then only the journal would be protected, but not the actual data? I think a separate journal_mac: also only makes sense together with a stacked dm-crypt.

I agree with you. That in the end brings us currently to the use of one key, ok thanks. I will adapt the code accordingly

> Sorry, I also did not attempt to set up anything with encrypted journal... With internal_hash: and journal_crypt:, we would encrypt the journal, but the data itself will be unencrypted on the block device. As I read the documentation, journal_crypt: is supposed to protect against attacks on a stacked dm-crypt. But in standalone mode, this does not seem useful to me.

ok got it. I will leave this aspect out of scope now for this pull request.

raimue (approved):
Thank you for the rework. Looks good to me now!

PR #2134, opened 2022-05-05
Follow up fix for isolinux-config

schaefi:
isolinux-config is called to update the search path inside
of the isolinux binary. isolinux/syslinux is exclusive to
the ix86 architecture and to BIOS firmware. Therefore the
condition to actually call it should reflect this.


schaefi:
The former implementation failed on any non efi firmware which includes e.g PPC with ofw firmware. Actually all this isolinux stuff is only valid for the old bios firmware

PR #2138, opened 2022-05-13
Subformats should also not be compressed when encryption is enabled

rjschwei:
This is a follow on change to bdba953. When the filesystem is encrypted the
resulting image should not be compressed.

Fixes # .

*
* Do not compress encrypted formats

### DIFF
--- a/kiwi/builder/filesystem.py
+++ b/kiwi/builder/filesystem.py
@@ -91,6 +91,7 @@ class FileSystemBuilder:
         self.filesystems_no_device_node = [
             'squashfs'
         ]
+        self.luks = xml_state.get_luks_credentials()
         self.result = Result(xml_state)
         self.runtime_config = RuntimeConfig()
 
@@ -128,13 +129,14 @@ class FileSystemBuilder:
         )
         if self.bundle_format:
             self.result.add_bundle_format(self.bundle_format)
+        compression = self.runtime_config.get_bundle_compression(default=True)
+        if self.luks is not None:
+            compression = False
         self.result.add(
             key='filesystem_image',
             filename=self.filename,
             use_for_bundle=True,
-            compress=self.runtime_config.get_bundle_compression(
-                default=True
-            ),
+            compress=compression,
             shasum=True
         )
         self.result.add(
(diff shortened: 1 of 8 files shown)


kiwi/builder/filesystem.py
```diff
@@ -128,13 +129,14 @@ def create(self) -> Result:
         )
         if self.bundle_format:
             self.result.add_bundle_format(self.bundle_format)
+        compression = self.runtime_config.get_bundle_compression(default=True)
+        if self.luks is not None:
+            compression = False
```
schaefi, line 134: shouldn't the setting of an eventual runtime config not always have the highest priority ? iirc if luks is activated the compression is switched off even if the runtime config would configure something different. I think this would be unexpected ?

kiwi/builder/filesystem.py
```diff
@@ -128,13 +129,14 @@ def create(self) -> Result:
...
         self.result.add(
             key='filesystem_image',
             filename=self.filename,
             use_for_bundle=True,
-            compress=self.runtime_config.get_bundle_compression(
-                default=True
-            ),
```
schaefi, line 137: If we say:

```python
compress=self.runtime_config.get_bundle_compression(
    default=True if self.luks is None else False
),
```

we would still respect any setting of the runtime config if it exists

Thoughts ?

schaefi: I see the same issue at the other changed places

schaefi: ok, but I think this should be explained to the user like you did to me. Any objection that you add this explanation to the ```kiwi/kiwi.yml``` file which we ship ? In this file we explain all configuration parameters. This one would be the first were the user gets overruled under a certain condition. I don't like black magic as I suffer from obs every day and I usually go for user requests no matter how stupid they are unless they are not possible to fulfill. In this case we act as "we know better" and I think you and I made many jokes about it in the past ;) Anyway I would go with your implementation if there are comments which we can point the angry user to :)

Thanks much

schaefi (approved):
Thanks looks good :+1:

PR #2142, opened 2022-05-16
use -s for kexec to allow secure boot

msmeissn:
Fixes that a OEM ISO could not boot on an secure boot enabled system.

* use -s as kexec option to use the kexec_file_load system call ( available since kernel 3.17)

### DIFF
--- a/dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
+++ b/dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
@@ -7,6 +7,10 @@ type run_dialog >/dev/null 2>&1 || . /lib/kiwi-dialog-lib.sh
 #======================================
 # Functions
 #--------------------------------------
+function get_kernel_version_majorminor {
+	printf "%d%02d" "$(uname -r | cut -f1 -d.)" "$(uname -r | cut -f2 -d.)"
+}
+
 function boot_installed_system {
     declare kiwi_oemreboot=${kiwi_oemreboot}
     declare kiwi_oemrebootinteractive=${kiwi_oemrebootinteractive}
@@ -50,7 +54,12 @@ function boot_installed_system {
     if getargbool 0 rd.kiwi.debug; then
         boot_options="${boot_options} rd.kiwi.debug"
     fi
-    kexec -l /run/install/boot/*/loader/linux \
+    options=""
+    if [ "$(get_kernel_version_majorminor)" -ge "317" ];then
+	# this variant is secure boot capable, but only available starting with Kernel 3.17+
+	options="--kexec-file-syscall"
+    fi
+    kexec $options -l /run/install/boot/*/loader/linux \
         --initrd /run/install/initrd.system_image \
         --command-line "${boot_options}"
     if ! kexec -e; then


schaefi:
Thanks for the pull request, I like this change.

I struggled a bit if we can just do the move without harming older systems. 

```
-s, --kexec-file-syscall Use file based syscall for kexec operation
-c, --kexec-syscall  Use the kexec_load syscall for for compatibility
                      with systems that don't support -s (default)
```

maybe you can do a simple usage check on kexec and then decide for either -s or -c. Since "-c" is still the default we imho should not unconditionally use "-s"

Thoughts ?

schaefi:
@davidcassany @dcermak @Conan-Kudo your feedback would be welcome. Thanks

Conan-Kudo:
We still support RHEL 7 (3.10 kernel). Do we know if this feature was backported there?

schaefi:
> i am not familar with RHEL.
> 
> Also do not know if you can fallback , e.g. if kexec -s would fail with error if not available? then we could use kexec -s ... || kexec construct

nope that will not work, if you issue kexec "wrong" it will hang the kernel, you can't go back.

davidcassany:
I have no experience on that area, I am not really aware of kexec differences across different and older distros. I guess we should essentially check the oldest distros we don't want to risk to break. Trying to list major ones, please correct me if wrong (there is no SLE 12 since kiwi is forked there and Leap 15 and SLE 15 are paired):

* openSUSE Leap 15.2 (soon Leap 15.3 as 15.4 is pretty close)
* Ubuntu 20.04
* Fedora 34
* Debian 10
* CentOS 7

I believe we only need to ensure we are not breaking CentOS 7, all others are pretty recent compared to CentOS 7. I am also wondering until when we should support CentOS 7, not that I want to drop its support but the gap between CentOS 7 and other is starting to be huge, so this is likely to be a recurring question now. I think EOF for CentOS is up to 2024, would it be reasonable to drop it before that?

schaefi:
I do agree with what David said. I also know at least two users who run into a kernel crash with the default kexec_load syscall and I could not create a crash with "--kexec-file-syscall" so far. My vote is to go with the proposed change

Conan-Kudo:
I want us to drop a fair bit of legacy cruft for KIWI 10, and I think dropping EL7 support at the same time would be reasonable.

schaefi:
> I want us to drop a fair bit of legacy cruft for KIWI 10, and I think dropping EL7 support at the same time would be reasonable. That is, we won't account for it anymore for future development.

ok in this case try to add the switch depending on the kernel version. I suggest the following in ```dracut/modules.d/99kiwi-lib/kiwi-lib.sh```

```bash
function get_kernel_version_majorminor {
    printf "%d%02d" "$(uname -r | cut -f1 -d.)" "$(uname -r | cut -f2 -d.)"
}
```

and then a simple

if [ "$(get_kernel_version_majorminor)" -ge "317" ];then
    # add me the -s switch....
fi

Thoughts ?

@msmeissn if acceptable, would you be willing to update your PR with regards to that ?

Thanks

msmeissn:
I put it in as you suggested (untested).

dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
```diff
@@ -50,7 +54,12 @@ function boot_installed_system {
     if getargbool 0 rd.kiwi.debug; then
         boot_options="${boot_options} rd.kiwi.debug"
     fi
-    kexec -l /run/install/boot/*/loader/linux \
+    switchs=""
```
schaefi, line 57: Thanks for the update. Sorry really nitpick but can we name this

```
options=""
```

dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
```diff
@@ -50,7 +54,12 @@ function boot_installed_system {
...
+    if [ "$(get_kernel_version_majorminor)" -ge "317" ];then
+	# this variant is secure boot capable, but only available with Kernel 3.17+
+	switchs="-s"
```
schaefi, line 60: and be more clear on what -s is with the long form

```
options="--kexec-file-syscall"
```

who know what other call types we will have to handle in the future and I think the code is then a bit better prepared for future changes

Thoughts ?

msmeissn:
i made that change and squashed it in 1 commit

schaefi (approved):
Looks good to me

@davidcassany can you please double check, initrd code you know should have at least two people to approve ;) Thanks

dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
```diff
@@ -7,6 +7,10 @@ type run_dialog >/dev/null 2>&1 || . /lib/kiwi-dialog-lib.sh
 #======================================
 # Functions
 #--------------------------------------
+function get_kernel_version_majorminor {
+	printf "%d%02d" "$(uname -r | cut -f1 -d.)" "$(uname -r | cut -f2 -d.)"
```
davidcassany, line 11: I guess having two digits for the minor version is enough, let's hope something like kernel `v5.99` or `v5.125` won't ever exist or, at least, is not about to happen at mid term :)

davidcassany (approved):
Looks good.

I am just wondering how should we track/mark the code that is about to become obsolete. I believe this kernel version constraint makes sense only as long as we keep support to CentOS/RHEL 7. In any case it should not hurt having the version check even if we are always over version 3.17

dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
schaefi: You are right this code has some potential to return numbers which doesn't serve the purpose for a comparison as number. Maybe it should not be done. How about this information:

```
 -a (--kexec-syscall-auto)
              Try the new KEXEC_FILE_LOAD syscall first and  when  it  is  not
              supported  or  the kernel does not understand the supplied image
              fall back to the old KEXEC_LOAD interface.

              There is no one single interface that always works.

              KEXEC_FILE_LOAD is required  on  systems  that  use  locked-down
              secure  boot  to verify the kernel signature.  KEXEC_LOAD may be
              also disabled in the kernel configuration.

              KEXEC_LOAD is required for some  kernel  image  formats  and  on
              architectures that do not implement KEXEC_FILE_LOAD.

```

@Conan-Kudo @davidcassany Thoughts ?

Conan-Kudo: This is probably the best we could do...

msmeissn: hmm, Linus did never do more than 2 digit minor version number so far. 

i fear this -a option was only added in later kexec binaries perhaps and might not be available in older ones? Otherwise it would be nice to have that.

schaefi: sounds good to me

@msmeissn would you like to update the change in a way suggested by David or would you rather like me to take the current code and move it into a branch from where I have permissions to provide a new PR ? Either way is fine for me and I don't want to stress you with even more effort on this one. Just let us know what you prefer ?

Thanks

schaefi: Before entering vacation I'd like to open the alternative way suggested by @davidcassany. I hope you don't mind that I created #2155 as another possible solution. Let's decide what we prefer.

Thanks

schaefi:
Closing on behalf of #2155

PR #2143, opened 2022-05-17
Add support for prebuilt bootstrap package for apt

schaefi:
When using the apt packagemanager kiwi required the use of debootstrap to create the initial rootfs. This works as long as there is always a main distribution repository available which follows the structure of the official debian mirrors. However if such a main distribution is not present or an alternative layout like e.g OBS repos is used, debootstrap will refuse to work. To allow for an alternative and without the dependency to debootstrap kiwi supports using a prebuilt bootstrap package providing the mini rootfs to serve as the bootstrap result. As all other package managers properly supports installation into an empty new root, this feature was only added when using the apt packagemanager

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -420,6 +420,15 @@ btrfs_root_is_readonly_snapshot="true|false":
   effective if `btrfs_root_is_snapshot` is also set to true. By default the
   root filesystem snapshot is writable.
 
+bootstrap_package="package_name":
+  For use with the `apt` packagemanager only. Specifies the name
+  of a bootstrap package which provides a bootstrap tarball
+  in :file:`/var/lib/bootstrap/PACKAGE_NAME.ARCH.tar.xz`.
+  The tarball will be unpacked and used as the bootstrap
+  rootfs to begin with. This allows for an alternative bootstrap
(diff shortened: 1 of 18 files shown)


Conan-Kudo:
Why is this restricted to the apt package manager? Wouldn't this be generically useful as an alternative to the bootstrap phase entirely?

schaefi:
> Why is this restricted to the apt package manager? Wouldn't this be generically useful as an alternative to the bootstrap phase entirely?

it's generally useful yes but it also offers pain when done wrong. All package managers except apt properly supports bootstrapping into an empty root. For those I intentionally don't want to introduce an alternative because I consider it as the preferred and supported way to bootstrap a new system. Only the Debian and derivatives still uses this more or less outdated packagemanager stack and as long as "apt --root ... install something" is not working I wanted to provide an alternative solution.

It would be very easy to extend the concept to all other packagemanager as well but I see it more like a weakness of apt for which kiwi can provide an alternative rather than a feature that I wanted people to use in larger scope. If it turns out that there is demand by our users for a prebuilt bootstrap as a general concept and solution we can fulfill the wish but I would more like to see a useful extension to apt and a good dependency solver (sat) to finally land there

Conan-Kudo (approved):

PR #2146, opened 2022-05-23
Added --target-arch for image info

schaefi:
Allow cross arch dependency solving

For example

```
user@host_some_arch:
> kiwi-ng --target-arch aarch64 image info --description /path/to/image_description --resolve-package-list
```


schaefi:
failing gitlab pipeline can be ignored for now, it's a mock issue investigated by the Fedora team

PR #2148, opened 2022-05-24
Setup SELinux on every system prepare / build

### LINKED ISSUE #2147: selinux security context only created for disk builder
## Problem description

In the disk builder code there is

```python
def _setup_selinux_file_contexts(self) -> None:
        security_context = '/etc/selinux/targeted/contexts/files/file_contexts'
        if os.path.exists(self.root_dir + security_context):
            self.system_setup.set_selinux_file_contexts(
                security_context
            )
```

However other builders like kis or simple filesystem builder does not create any security context

## Expected behaviour

All builders for which the creation of the selinux context is needed should create it

zzzoom:
Fixes #2147. Tested on a bunch of image types.

I'm not entirely convinced about setting file contexts before `config.sh`, though.

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -513,9 +513,6 @@ class DiskBuilder:
             disk.storage_provider
         )
 
-        # set SELinux file security contexts if context exists
-        self._setup_selinux_file_contexts()
-
         # run pre sync script hook
         if self.system_setup.script_exists(
             defaults.PRE_DISK_SYNC_SCRIPT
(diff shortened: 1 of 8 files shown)


schaefi:
@zzzoom let me know if you need help regarding the tests, but I think you will manage it :)

zzzoom:
@schaefi it should be fixed now.

schaefi:
@zzzoom 

Thanks for the update on the unit tests. As you have increased the scope of the code there is
now code that is not covered by a test. Thus the result in the unit fails due to missing coverage:

```
/home/runner/work/kiwi/kiwi/kiwi/system/setup.py                                   520      3    99%   491-493
```

Can you look again on the tests and cover this part ?

If you feel this stands in your way just let me know then I'll complete on the test area

I'm a friend of 100% test coverage for the reason to have full freedom to do changes.
I know there are mixed feelings about this part in test driven development. Thus I respect any opinion
developers have in this regard :)

Just let me know your thoughts

Thanks

zzzoom:
@schaefi ok, it should have full coverage now.

schaefi:
Thanks much, going to merge it :+1:

PR #2152, opened 2022-06-08
Correct setfiles relabeling

schaefi:
This change was inspired by a change done on Fedora's livecd-tools
from here: livecd-tools/livecd-tools#236. The patch corrects issues
with the setfiles SELinux relabel command. The issues become apparent
when the host and guest policies differ. Thus it becomes required
to explicitly set the policy to decouple from eventual unwanted
host settings.

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -544,6 +544,11 @@ target_removable="true|false":
   of the target machine. By default the target disk is
   expected to be non-removable
 
+selinux_policy.attribute="targeted|mls|minimum":
+  The `selinux_policy` attribute sets the SELinux policy to use.
+  `targeted` policy is the default policy. Only change this option
+  if you want to use the `mls` or `minimum` policy.
+
 spare_part="number":
   Request a spare partition right before the root partition
   of the requested size. The attribute takes a size value
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -1419,6 +1419,12 @@ div {
             sch:param [ name = "attr" value = "target_removable" ]
             sch:param [ name = "types" value = "oem" ]
         ]
+    k.type.selinux_policy.attribute =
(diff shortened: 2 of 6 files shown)


schaefi:
@Conan-Kudo it would be great if you can give this one a closer look. I haven't tested it for real on a selinux enabled system and I wanted to make sure that I did not mix up with the policy paths and such. If time permits I will do an image build in the cloud on some Fedora host system to see if it actually does work as expected.

Thanks

doc/source/image_description/elements.rst
```diff
@@ -544,6 +544,11 @@ target_removable="true|false":
   of the target machine. By default the target disk is
   expected to be non-removable
 
+selinux_policy.attribute="targeted|mls":
```
Conan-Kudo, line 547: There are three types: `targeted`, `mls`, and `minimum`. The default should be `targeted`.

Conan-Kudo (changes requested):

Conan-Kudo:
A preliminary test package of this patch (https://copr.fedorainfracloud.org/coprs/ngompa/kiwi-testselinuxfix/) seemed to make it so that I could successfully build a CentOS 7 machine on CentOS 9 properly with SELinux set up properly.

I haven't tested building a CentOS machine on Fedora or vice versa though.

schaefi:
> A preliminary test package of this patch (https://copr.fedorainfracloud.org/coprs/ngompa/kiwi-testselinuxfix/) seemed to make it so that I could successfully build a CentOS 7 machine on CentOS 9 properly with SELinux set up properly.

Thanks a ton for the effort and the feedback. That's great. I'm going to update the schema according to your first comment and then I think we are good to go.

Will do this today

doc/source/image_description/elements.rst
schaefi: Thanks, added it to the schema

Conan-Kudo (approved):

PR #2155, opened 2022-06-13
Prefer file based syscall in kexec when possible

schaefi:
Use file based syscall in kexec if available. This is needed to
support boot on an secure boot enabled system and is in general
more reliable to boot into the system on real hardware platforms


schaefi:
On behalf of #2142 I implemented the suggestion from @davidcassany. Please let us decide which solution we would like to prefer or consider more maintainable.

Thanks much

msmeissn:
looks also good to me.

PR #2156, opened 2022-06-13
Fix btrfs volume mounting

lnussel:
If the second argument of os.path.join is an absolute directory, the
result would be that directory. The intention is to produce a
subdirectory of the mountpoint though. So pass a relative path.

Without the fix, kiwi would try to e.g mount the /var subvolume of
the image over the /var of the host, screwing everthing up of course
:-)

### DIFF
--- a/kiwi/system/mount.py
+++ b/kiwi/system/mount.py
@@ -175,7 +175,7 @@ class ImageSystem:
             volume_mount = MountManager(
                 device=self.volumes[volume_path]['volume_device'],
                 mountpoint=os.path.join(
-                    mountpoint, volume_path
+                    mountpoint, os.path.relpath(volume_path, '/')
                 )
             )
             self.mount_list.append(volume_mount)


lnussel:
I'm not a kiwi developer, so please don't ask me for a test case. I can't provide it :-) The way I found it was trying to add a disk.sh script to openSUSE-MicroOS.kiwi

davidcassany:
Damn! good catch! I can assist on providing/adapting a unit test if needed.

schaefi:
good catch thanks much for the fix :+1: 

The unit tests needs to be adapted though. Let us know if you need assistance on that part as David suggested.
Thanks

lnussel:
would be good if you could take care of the test cases. Looks like the openSUSE package doesn't run them.

davidcassany:
> would be good if you could take care of the test cases. Looks like the openSUSE package doesn't run them.

Sure I will,

Just let us know if you are still working on the PR, I saw you added a new commit since I last viewed it

lnussel:
my use case works now, so that's it I think

davidcassany:
@lnussel closing this PR on behalf of #2161. There where I coded the unit tests

PR #2159, opened 2022-06-14
Handle older versions of setfiles correctly

schaefi:
In older versions of setfiles we need a two pass setup
First set the policy, second apply the security context.
This commit checks in the usage message of setfiles which
invocation syntax is required

### DIFF
--- a/kiwi/utils/command_capabilities.py
+++ b/kiwi/utils/command_capabilities.py
@@ -17,6 +17,7 @@
 #
 import re
 import logging
+from typing import List
 
 # project
 from kiwi.command import Command
@@ -35,8 +36,8 @@ class CommandCapabilities:
     """
     @staticmethod
     def has_option_in_help(
-        call, flag, help_flags=None,
-        root=None, raise_on_error=True
+        call: str, flag: str, help_flags: List[str] = [],
+        root: str = '', raise_on_error: bool = True
     ):
         """
         Checks if the given flag is present in the help output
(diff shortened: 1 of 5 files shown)


schaefi:
As the answer on the RHEL bugzilla regarding the upstream patch to be included in EL7 was a decline I continued testing the patch from here:

@Conan-Kudo ok I have tested this patch with older (centos7) and newer (fedora, centos9) builds and it worked for me

@davidcassany I had to add a small change to the CommandCapabilities class which I think makes sense in general. Would you mind to take a look at that part of the patch ?

Thanks

kiwi/utils/command_capabilities.py
```diff
@@ -58,7 +58,7 @@ def has_option_in_help(
         else:
             arguments = [call] + help_args
         try:
-            command = Command.run(arguments)
+            command = Command.run(arguments, raise_on_error=False)
```
davidcassany, line 61: mm I am not so sure about that... this method has its own raise_on_error to swallow errors, with the arguable benefit that at least throws a devoted warning message. If one uses this without the raise_on_error it also returns  `KIWICommandCapabilitiesError`. I do not think error filtering is something we will ever need in that context, but who knows...

Moreover I saw you are already calling `has_option_in_help` with `raise_on_error=False`, so I can't see the behavior difference of having this change in this particular case.

Am I missing something?

schaefi: I think you are missing the following

```python
command = Command.run(arguments, raise_on_error=False)
for line in command.output.splitlines():
    if flag in line:
        return True
for line in command.error.splitlines():
    if flag in line:
        return True
```

As you can see we catch information from the error channel of the called command. However this path will never be reached except for commands that prints on stderr but exit with an exit code == 0. Usually commands that prints on stderr also exits with something != 0. In this case Command.run raises with an exception and so all of the second for loop will not be reached. This is also what happens if you call ```setfiles``` it prints its usage and extis with 255. Thus our class cannot be used to fetch data from stderr.

In my opinion this limits the scope a lot and we should not care for the exit code (raise_on_error=False) and allow the class to always see both stdout and stderr data.

So far we used it only in combination with "program --help" which prints on stdout and exits with == 0

Makes sense ?

schaefi: yes you are right, I will adapt the code. Thanks

PR #2161, opened 2022-06-15
Fix btrfs volume mounting

davidcassany:
If the second argument of os.path.join is an absolute directory, the
result would be that directory. The intention is to produce a
subdirectory of the mountpoint though. So pass a relative path.

### DIFF
--- a/kiwi/system/mount.py
+++ b/kiwi/system/mount.py
@@ -167,7 +167,7 @@ class ImageSystem:
             volume_mount = MountManager(
                 device=self.volumes[volume_path]['volume_device'],
                 mountpoint=os.path.join(
-                    mountpoint, os.path.relpath(volume_path, '/')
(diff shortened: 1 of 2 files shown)


davidcassany:
This PR supersedes #2156

davidcassany:
@lnussel Note I changed the `os.path.relpath(volume_path, "/")` to `volume_path.lstrip(os.sep)`. I found some unexpected behavior with `os.path.relpath`:

```
$ ptyhon3
>>> import os
>>> os.path.relpath("/var", "/")
'var'
>>> os.path.relpath("var", "/")
'home/david/work/kiwi/var'
```

So if the volume_path was already a relative path it appends the current directory starting from `/`. So I don't think this is a good option.

schaefi (approved):
looks good :+1: Thanks

Conan-Kudo (approved):

PR #2167, opened 2022-07-06
Fixed custom disk start sector setup

schaefi:
The attribute

```xml
<type ... disk_start_sector="..."/>
```

allows to specify a custom start sector for the first partition of the disk. On GPT tables everything works nicely, on DOS tables the used tools fdisk/sfdisk are not able to manage the start/end values of subsequent partitions if the first partition doesn't start with the tooling default. This patch allows to set the start sector after the partition table has been created


schaefi:
I need this one to go in as I have to continue with a task where this change is required. The patch is actually working on my target board here on my desk but needs to become effective upstream for a release. @davidcassany if you don't mind I'm going to merge this with pending review.

If there is an issue found by after the merge please drop me a short note. Thanks and sorry but otherwise this is blocking me from work

PR #2168, opened 2022-07-07
Umount device before cloning

schaefi:
In case a clone should be created from a partition we need
to make sure to umount the device after sync and prior cloning.
Otherwise the clone operation is not safe because the rsynced
data might still be in memory and not synced out to the block
device.


davidcassany (approved):
Looks good to me :+1:

PR #2175, opened 2022-07-21
Fix mocking of grub2-install (fix #2124)

### LINKED ISSUE #2124: grub2-install mocking of /bin/true is copying host machine binary
Copy command is not run inside the chroot, and /bin/true is a symlink so you end copying the host machine file.
Noticed the problem because my /bin/true binary is linked against a newer version of glibc, what caused the binary to not being able to run.

Will provide a PR soon to run the command inside the chroot instead to guarantee we copy the right file

gmoro:
/bin/true was being used from the host machine instead of
the chroot

Fixes #2124  .

### DIFF
--- a/kiwi/bootloader/install/grub2.py
+++ b/kiwi/bootloader/install/grub2.py
@@ -335,28 +335,32 @@ class BootLoaderInstallGrub2(BootLoaderInstallBase):
         if os.access(root_path, os.W_OK):
             grub2_install = ''.join(
                 [
-                    root_path, '/usr/sbin/',
+                    '/usr/sbin/',
                     self._get_grub2_install_tool_name(root_path)
                 ]
             )
             grub2_install_backup = ''.join(
                 [grub2_install, '.orig']
             )
-            grub2_install_noop = ''.join(
-                [root_path, '/bin/true']
+            grub2_install_noop = '/bin/true'
+            Command.run(
+                [
+                    'chroot', root_path,
+                    'cp', '-p', grub2_install, grub2_install_backup
+                ]
             )
             Command.run(
-                ['cp', '-p', grub2_install, grub2_install_backup]
-            )
-            Command.run(
-                ['cp', grub2_install_noop, grub2_install]
+                [
+                    'chroot', root_path,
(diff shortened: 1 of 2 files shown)


kiwi/bootloader/install/grub2.py
```diff
@@ -335,28 +335,32 @@ def _disable_grub2_install(self, root_path):
...
                 [grub2_install, '.orig']
             )
-            grub2_install_noop = ''.join(
-                [root_path, '/bin/true']
-            )
+            grub2_install_noop = '/bin/true'
             Command.run(
-                ['cp', '-p', grub2_install, grub2_install_backup]
+                [
+                    'chroot', self.root_mount.mountpoint,
```
schaefi, line 348: For consistency I think this should be

'chroot', root_path,

kiwi/bootloader/install/grub2.py
```diff
@@ -335,28 +335,32 @@ def _disable_grub2_install(self, root_path):
...
+                    'cp', '-p', grub2_install, grub2_install_backup
+                ]
             )
             Command.run(
-                ['cp', grub2_install_noop, grub2_install]
+                [
+                    'chroot', self.root_mount.mountpoint,
```
schaefi, line 354: For consistency I think this should be

'chroot', root_path,

kiwi/bootloader/install/grub2.py
```diff
@@ -365,7 +369,10 @@ def _enable_grub2_install(self, root_path):
             )
             if os.path.exists(grub2_install_backup):
                 Command.run(
-                    ['mv', grub2_install_backup, grub2_install]
+                    [
+                        'chroot', self.root_mount.mountpoint,
```
schaefi, line 373: For consistency I think this should be

'chroot', root_path,

schaefi (changes requested):
Thanks for your effort, looks good to me just some consistency suggestion

See my comments

schaefi:
@gmoro any feedback from your side regarding the review ? Thanks

schaefi:
@gmoro any news ?

PR #2182, opened 2022-08-06
Support custom size for embedded EFI FAT image

schaefi:
For ISO images (live and install) the EFI boot requires an embedded
FAT image. As of now a fixed size of 20M was used which leads to a
problem if the EFI image or the initrd or the kernel is bigger than
20M. With the new attribute:

```
efifatimagesize="nonNegativeInteger"
```

we can now set a different value for the FAT image. Please note the
value must be aligned to the also customizable efipartsize value
which allows to configure the size of the EFI partition


schaefi:
With this patch the failed integration test from here can be fixed

* Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-disk-legacy

PR #2186, opened 2022-08-11
Further Prevent Race Conditions for re-part/fsck

### LINKED ISSUE #2185: Dracut Re-Part/fsck Race Conditions Still Exist
## Problem description
Some hardware simply takes an unusual amount of time to detach/re-attach disks during execution of this area of the code on post-deployment reboots. Even with the prior fix implemented, we still see situations where the Root Device isn't available in the expected location when the `fsck` SystemD Service starts.

## Expected behaviour
Root Device is reachable before `fsck` Service starts after repartitioning

Deploy a PXE Bootable OEM Expandable Disk Image of SLES 12 SP5 (Or probably anything, really.) on certain older hardware and wait until the post-deployment reboot (Not counting the reboot triggered at the end of the deployment.)

SLES 12 SP5 +-
Older Hardware

* KIWI version: 9.24 +-
* Operating system host version: SLES 15 SP2
* Operating system target version: SLES 12 SP5

TheMisanthropicHumanist:
Some hardware simply takes an unusual amount of time to detach/re-attach disks during execution of this area of the code on post-deployment reboots. This `until` loop guarantees that the `/dev/disk/by-uuid/[Root Disk Device]` Node exists before we enable the `fsck` Service regardless of what happens during repartitioning. Without this, even with the prior fix implemented, we still see situations where the Root Device isn't available in the expected location when the `fsck` SystemD Service starts.

Fixes #2185

### DIFF
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -240,7 +240,11 @@ setup_debug
 # More details: https://github.com/SUSE/kiwi/issues/1034
 
 # make sure we unmask the fsck service
-trap unmask_fsck_root_service EXIT
+trap "
+until [ -e \"${root_device}\" ]; do
+     sleep .5
+done
+unmask_fsck_root_service" EXIT
 
 mask_fsck_root_service
 
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -240,11 +240,11 @@ setup_debug
 # More details: https://github.com/SUSE/kiwi/issues/1034
 
 # make sure we unmask the fsck service
-trap "
+trap '
 until [ -e ${root_device} ]; do
      sleep .5
 done
-unmask_fsck_root_service" EXIT
+unmask_fsck_root_service' EXIT
 
 mask_fsck_root_service
 
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -241,7 +241,7 @@ setup_debug
 
 # make sure we unmask the fsck service
 trap '
-until [ -e ${root_device} ]; do
+until [ -e "${root_device}" ]; do
      sleep .5
 done
 unmask_fsck_root_service' EXIT
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -241,7 +241,7 @@ setup_debug
 
 # make sure we unmask the fsck service
 trap "
-until [ -e '${root_device}' ]; do
+until [ -e ${root_device} ]; do
      sleep .5
 done
 unmask_fsck_root_service" EXIT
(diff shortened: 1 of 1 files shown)


TheMisanthropicHumanist:
This fix has been tested successfully on our end. Let me know if you need some logs from either a pre-fix or post-fix scenario.

dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
```diff
@@ -240,7 +240,11 @@ setup_debug
 # More details: https://github.com/SUSE/kiwi/issues/1034
 
 # make sure we unmask the fsck service
-trap unmask_fsck_root_service EXIT
+trap '
+until [ -e "${root_device}" ]; do
+     sleep .5
+done
+unmask_fsck_root_service' EXIT
```
schaefi, line 247: Thanks for your effort

From the description of the PR I understand the motivation, but from a code perspective it might be confusing for readers. Actually the trap now changes its scope from a simple "unmask" to a "wait_for_dev before we continue". 
Thus keeping the name ```unmask_fsck_root_service``` is misleading now

The implementation itself waits for the root_device to appear. We already have a method to do this
in ```99kiwi-lib/kiwi-partitions-lib.sh``` called ```wait_for_storage_device```. I recommend using this method
and I see an endless loop option in the code proposed here for the case that $root_device never appears

What I was wondering most is that the extended trap solves your issue ? If you look in the module code you
will already see the following line:

```bash
# wait for the root device to appear
wait_for_storage_device "${root_device}"
```

I believe your extra trap code is effective if the module returns earlier. There is one condition when this happens
and that is here:

```bash
# check if repart/resize is wanted
if ! resize_wanted "${root_device}" "${disk}"; then
    return
fi
```

So my suggestion would be that we do the following:

1. Rename the trap method to actually explain what it really does (maybe just a simple "wait_and_clean") ... naming is so hard :)
2. Change your implementation to something like this

```bash
function wait_and_clean {
    # wait for the root device to appear
    wait_for_storage_device "${root_device}"

    info "enabling systemd-fsck-root.service"
    systemctl unmask systemd-fsck-root.service
}

# wait for the root_device to appear and unmask the fsck service prio exit
trap wait_and_clean EXIT
```

Thoughts ?

schaefi: Regarding the "device does not appear" issues you still have. When using ```wait_for_storage_device``` there is also a variable ```DEVICE_TIMEOUT``` which you can set at the kernel cmdline or in the ```kernel_cmdline="..."``` attribute in your XML description.

The device timeout if not specified is set to 30sec. You said you are using hardware that might exceed this timeout. So you can play with the timeout value here. Please note the timeout is passed along to the ```udevadm settle --timeout=${limit}``` call

schaefi (changes requested):
Thanks for your effort. I understand the motivation, please see my comments

dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
TheMisanthropicHumanist: Ahh, I see your points and I agree, this does have broader implications and should leverage existing code where possible with more accurate naming.

I'm happy with your suggestion, and will give that a test when I get a chance, although it may take some time as I'm pretty swamped the rest of this month. Feel free to close this or mark it pending or whatever for now if you need and I can update this after I've had a chance to test.

Thanks!

schaefi: yep no worries, I set the WIP label. If you will not be able to continue on this one for let's say multiple months, it would be great if you can drop us a note. I'd like to prevent bitrot ;) Thanks

schaefi:
@TheMisanthropicHumanist Hi, after the long time this draft PR is open now I assume it will not be continued. Feel free to re-open a new PR in case you get back to this topic. Thanks much

PR #2189, opened 2022-08-31
Limit repo alias names to be a safe POSIX name

### LINKED ISSUE #2170: Ubuntu build fails with spaces in repo alias
## Problem description

When I define a ubuntu/debian repository in the xml , if the alias name contains spaces, the install step fails.

For example, this works:

```
    <repository type="apt-deb" alias="focalupdates" distribution="focal-updates" components="main restricted universe multiverse" repository_gpgcheck="false" profiles="upstream">
        <source path="http://archive.ubuntu.com/ubuntu/"/>
    </repository>
```

This fails:

```
    <repository type="apt-deb" alias="focal updates" distribution="focal-updates" components="main restricted universe multiverse" repository_gpgcheck="false" profiles="upstream">
        <source path="http://archive.ubuntu.com/ubuntu/"/>
    </repository>
```

with errors when installing the system packages, e.g.:

```
[ DEBUG   ]: 13:40:25 | system: Package dpkg-dev is not available, but is referred to by another package.
[ DEBUG   ]: 13:40:25 | system: This may mean that the package is missing, has been obsoleted, or
[ DEBUG   ]: 13:40:25 | system: is only available from another source
[ DEBUG   ]: 13:40:25 | system: Package gnupg2 is not available, but is referred to by another package.
[ DEBUG   ]: 13:40:25 | system: This may mean that the package is missing, has been obsoleted, or
[ DEBUG   ]: 13:40:25 | system: is only available from another source
[ DEBUG   ]: 13:40:25 | system: However the following packages replace it:
[ DEBUG   ]: 13:40:25 | system:   gpgv
[ DEBUG   ]: 13:40:25 | system: Package libpam0g-dev is not available, but is referred to by another package.
[ DEBUG   ]: 13:40:25 | system: This may mean that the package is missing, has been obsoleted, or
[ DEBUG   ]: 13:40:25 | system: is only available from another source
[ DEBUG   ]: 13:40:25 | system: However the following packages replace it:
[ DEBUG   ]: 13:40:25 | system:   libpam-runtime
[ DEBUG   ]: 13:40:25 | system: Package bzip2 is not available, but is referred to by another package.
[ DEBUG   ]: 13:40:25 | system: This may mean that the package is missing, has been obsoleted, or
[ DEBUG   ]: 13:40:25 | system: is only available from another source
[ DEBUG   ]: 13:40:25 | system: Package xz-utils is not available, but is referred to by another package.
[ DEBUG   ]: 13:40:25 | system: This may mean that the package is missing, has been obsoleted, or
[ DEBUG   ]: 13:40:25 | system: is only available from another source
```

This is probably because the generated file under `/var/cache/kiwi` has spaces in the name.

## Expected behaviour

Spaces in alias gives no problems or spaces in alias are explicitly not allowed, giving a more clear error earlier on.

Try to build Ubuntu image with spaces in alias

* KIWI version: 9.24.44
* Operating system host version: Ubuntu 20.04 LTS 
* Operating system target version: Ubuntu 20.04 LTS 

schaefi:
Characters like spaces or other symbols used in repo alias names
can cause the package manager to fail setting up the repo. Thus
this patch changes the schema to only allow for safe POSIX names
matching:

```
{pattern = "[a-zA-Z0-9_\-\.]+"}.
```

This Fixes #2170

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -1002,12 +1002,13 @@ div {
         }
     k.repository.alias.attribute =
         ## Alias name to be used for this repository. This is an
-        ## optional free form text. If not set the source attribute
+        ## optional free-form text restricted to characters from the
+        ## POSIX standard. If not set the source attribute
         ## value is used and builds the alias name by running a md5 digest
         ## of the defined URI of the repository. An alias name should be
         ## set if the source argument doesn't really explain what this
         ## repository contains.
-        attribute alias { text }
+        attribute alias { safe-posix-name }
     k.repository.components.attribute =
         ## Distribution components, used for deb repositories. If
         ## not set it defaults to main
--- a/kiwi/schema/kiwi.rng
+++ b/kiwi/schema/kiwi.rng
@@ -1521,11 +1521,13 @@ definition can be composed by other existing profiles.</a:documentation>
     <define name="k.repository.alias.attribute">
       <attribute name="alias">
         <a:documentation>Alias name to be used for this repository. This is an
-optional free form text. If not set the source attribute
+optional free-form text restricted to characters from the
+POSIX standard. If not set the source attribute
 value is used and builds the alias name by running a md5 digest
 of the defined URI of the repository. An alias name should be
 set if the source argument doesn't really explain what this
 repository contains.</a:documentation>
+        <ref name="safe-posix-name"/>
       </attribute>
     </define>
     <define name="k.repository.components.attribute">
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -1090,7 +1090,8 @@ the following optional attributes:
 alias="name"
   Specifies an alternative name for the configured repository. If the
   attribute is not specified {kiwi} will generate a random alias name
-  for the repository.
+  for the repository. The specified name must match the pattern:
+  `[a-zA-Z0-9_\-\.]+`
 
 components="name"
   Used for Debian (apt) based repositories only. Specifies the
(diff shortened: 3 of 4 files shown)


schaefi:
@davidcassany @Conan-Kudo I was struggling a bit to limit the alias names but I think it makes sense, also other package managers could be affected.

* If no alias is set kiwi calculates an md5 digest from the repo source spec, that is a safe as it was before
* Users who uses alias names that were accepted but contains characters not matching the pattern would now see a failed build and are "forced" to change the alias name. Is this something we are willing to accept the blame for ? ;)
* Going forward I think the pattern allows good enough space to create useful alias names

Thoughts ?

Conan-Kudo:
I haven't tested this, but I think DNF accepts pretty much anything for repo IDs except spaces. For somewhat absurd reasons, you can now do variable substitution in repo IDs too.

I'm fine with being more restrictive though. Being POSIX safe is probably a good idea since APT has us writing the IDs as filenames.

kiwi/schema/kiwi.rng
```diff
@@ -1521,11 +1521,12 @@ definition can be composed by other existing profiles.</a:documentation>
     <define name="k.repository.alias.attribute">
       <attribute name="alias">
         <a:documentation>Alias name to be used for this repository. This is an
-optional free form text. If not set the source attribute
+optional POSIX form text. If not set the source attribute
```
Conan-Kudo, line 1524: This is still free-form text, it's just restricted on what characters are allowed in the field.

kiwi/schema/kiwi.rnc
```diff
@@ -1002,12 +1002,12 @@ div {
         }
     k.repository.alias.attribute =
         ## Alias name to be used for this repository. This is an
-        ## optional free form text. If not set the source attribute
+        ## optional POSIX form text. If not set the source attribute
```
Conan-Kudo, line 1005: This is still free-form text, it's just restricted on what characters are allowed in the field.

Conan-Kudo (changes requested):

kiwi/schema/kiwi.rng
schaefi: ok, done. Fixed the schema docstring here and also updated the online documentation which I had forgotten

schaefi:
@Conan-Kudo update the PR, can you have a look again ? thanks

Conan-Kudo (approved):

PR #2190, opened 2022-08-31
ci(lint): Add Shell linter - Differential Shellcheck

jamacku:
Differential ShellCheck is a GitHub action that performs differential ShellCheck scans on shell scripts changed via PR and reports results directly in PR. The beauty of differential scans is that you will get reports only about newly added defects. 

Since this repository has few shell scripts, I think you might find some value in having a shell linter. Differential ShellCheck action is able to produce reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

### DIFF
--- a/.github/workflows/ci-differential-shellcheck.yml
+++ b/.github/workflows/ci-differential-shellcheck.yml
@@ -1,0 +1,27 @@
+name: CI-Differential-ShellCheck
+on:
+  pull_request:
+    branches: [master]
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    name: Linter - Differential ShellCheck
+    runs-on: ubuntu-latest
+
+    permissions:
+      security-events: write
+      pull-requests: write
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}


schaefi:
Hi, thanks this is really interesting. I wasn't aware the shell scorecard. I had enabled the security scorecard but checking the shellcode is good too.

btw as part of our unit testing we use shellcheck to check relevant shell codings of the project. You see this here:
https://github.com/OSInside/kiwi/blob/master/tox.ini#L204

and it gets called as part of:
https://github.com/OSInside/kiwi/blob/master/.github/workflows/ci-code-style.yml

I have no objections to do more checking on the shell code :)
 
Just one little thing, could you rename the workflow file to start with "ci-" it's a little policy we followed on all workflow files

Thanks much

jamacku:
> btw as part of our unit testing we use shellcheck to check relevant shell codings of the project. You see this here: https://github.com/OSInside/kiwi/blob/master/tox.ini#L204
>
> and it gets called as part of: https://github.com/OSInside/kiwi/blob/master/.github/workflows/ci-code-style.yml

I noticed it just after submitting PR, but I still thought you might be interested.

> Just one little thing, could you rename the workflow file to start with "ci-" it's a little policy we followed on all workflow files

I have updated the PR. @schaefi, please have a look.

schaefi:
Thanks, really nice :+1: , let's go for it. I'll merge once the workflow runs are done

PR #2191, opened 2022-09-03
Support DM integrity legacy HMAC

schaefi:
Add a new attribute integrity_legacy_hmac="true|false" which
allows to use old flawed HMAC calculation (does not protect superblock)
Do not use this attribute until compatibility with a specific
old kernel is required!

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -612,15 +612,6 @@ integrity_legacy_hmac="true|false":
      Do not use this attribute until compatibility with
      a specific old kernel is required!
 
-integrity_legacy_padding="true|false":
-  For the `oem` type only and in combination with the `standalone_integrity`
-  attribute, allow to use inefficient legacy padding.
-
-  .. warning::
-
-     Do not use this attribute until compatibility with
-     a specific old kernel is required!
-
 integrity_keyfile="filepath":
   For the `oem` type only and in combination with the `standalone_integrity`
   attribute, protects access to the integrity map using the given keyfile.
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -609,7 +609,7 @@ integrity_legacy_hmac="true|false":
 
   .. warning::
 
-     Do not use this attribute until compatibility with
+     Do not use this attribute unless compatibility with
      a specific old kernel is required!
 
 integrity_keyfile="filepath":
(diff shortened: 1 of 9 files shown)


doc/source/image_description/elements.rst
```diff
@@ -603,6 +603,15 @@ standalone_integrity="true|false":
   For the `oem` type only, specifies to create a standalone
   `dm_integrity` layer on top of the root filesystem
 
+integrity_legacy_hmac="true|false":
+  For the `oem` type only and in combination with the `standalone_integrity`
+  attribute, Allow to use old flawed HMAC calculation (does not protect superblock).
+
+  .. warning::
+
+     Do not use this attribute until compatibility with
```
Conan-Kudo, line 612: `until` -> `unless`

schaefi: thanks, changed

schaefi:
> But also, do we _really_ want to add this?

I was also thinking a lot about this change. The thing is fix_hmac is a flag for dm_integrity which was added in the kernel after kernel version 5.10. There are still some distributions that we support which are using even an older kernel e.g Leap 15.3 is 5.3.x. Thus if you are building a dm_integrity image on a host that supports fix_hmac but your target kernel does not you produce something that cannot work.

The solution to this I have seen from other people is to make sure the running kernel at image creation time is "old"
enough to support the target. However, this implicitly creates an integrity superblock with an hmac that populates a
security issue. You have to inspect the superblock to check on the flags to see this.

Now if we support in kiwi a legacy option it becomes clear and explicit that this image build uses an hmac implementation
which has issues and the customer can decide for go, no go. Since "man integritysetup" supports these legacy options I also thought it would be good make it explicit on the builder side and prevent people from being forced to look at workarounds
that hides them.

Because of this I think we should allow the new attribute. Once we are getting mature enough it would be no problem to
get rid of the legacy attribute by a schema transformation.

I agree with you we should avoid supporting legacy stuff but in this case I rather want to see the setting in the kiwi description instead of the surprise at boot time

Thoughts ?

Conan-Kudo:
My main concern is what we're going to do at the KIWI 10 timeframe. We need to start figuring out deprecations and retirement of features that don't make sense once we drop some older distributions.

With the context you have stated, I think it makes sense to add, but I wonder if we should mark it as a deprecated flag right from the get go and establish when we'll remove it.

Conan-Kudo:
On the other hand, it's explicitly documented in the man pages, so, exposing it isn't any worse.

Conan-Kudo (approved):

PR #2194, opened 2022-10-11
List riscv64 as a valid architecture in the schemas

dirkmueller:
This is needed so that architecture filters on riscv64 specifics can be defined.

Fixes # .

*
*


dirkmueller:
I don't understand the CI error

schaefi:
> I don't understand the CI error

me neither, this is not related to your change but due to a newer version of mypy being picky on the type check which worked before

schaefi (approved):
Looks good, thanks :+1:

PR #2195, opened 2022-10-12
Bind-mount /dev into the OCI container instead of rsync-ing it

ignatenkobrain:
In containers (nspawn) where part of the /dev filesystem is bind-mounted from outside system, kiwi fails to do the rsync (in creation of the nodes).

There is no reason to actually copy whole tree inside so let's just bind-mount it in.

Fixes https://github.com/OSInside/kiwi/issues/2020

### DIFF
--- a/kiwi/container/setup/base.py
+++ b/kiwi/container/setup/base.py
@@ -20,7 +20,6 @@ import os
 
 # project
 from kiwi.command import Command
-from kiwi.utils.sync import DataSync
 
 from kiwi.exceptions import (
     KiwiContainerSetupError
@@ -144,28 +143,6 @@ class ContainerSetupBase:
                 'console': 'console'
             }
         )
-
-    def setup_static_device_nodes(self):
-        """
-        Container device node setup
-
-        Without subsystems like udev running in a container it is
-        required to provide a set of device nodes to let the
-        system in the container function correctly. This is
-        done by syncing the host system nodes to the container.
-        That this will also create device nodes which are not
-        necessarily present in the container later is a know
-        limitation of this method and considered harmless
-        """
-        try:
-            data = DataSync('/dev/', self.root_dir + '/dev/')
-            data.sync_data(
-                options=['-a', '-x', '--devices', '--specials']
-            )
-        except Exception as e:
-            raise KiwiContainerSetupError(
-                'Failed to create static container nodes %s' % format(e)
-            )
 
     def get_container_name(self):
         """
--- a/kiwi/container/setup/oci.py
+++ b/kiwi/container/setup/oci.py
@@ -50,7 +50,6 @@ class ContainerSetupOCI(ContainerSetupBase):
 
         self.deactivate_bootloader_setup()
         self.deactivate_root_filesystem_check()
-        self.setup_static_device_nodes()
         self.setup_root_console()
 
         for service in services_to_deactivate:
--- a/test/unit/container/setup/base_test.py
+++ b/test/unit/container/setup/base_test.py
@@ -2,7 +2,6 @@ from mock import (
     patch, call, mock_open
 )
 from pytest import raises
-import mock
 
 from kiwi.container.setup.base import ContainerSetupBase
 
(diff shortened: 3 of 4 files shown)


davidcassany (commented):
I do not have any reason in favor of attempting to use some static devices for now. So the change looks good to me generally speaking.

Most of the code of ContainerSetupBase class is coming from 2016, containers technologies significantly evolved meanwhile. I wouldn't be surprised if the reason being of some portions of the code in ContainerSetupBase is no longer valid.

I don't know why it was coded with rsync at first. Also, I am wondering to what extend the comment in `setup_device_nodes` is still valid after the change.

What I am missing is the umount step. I understand that if we have `setup_device_nodes` setting  a mountpoint we should also have some logic to umount such a mountpoint.

kiwi/mount_manager.py
```diff
@@ -62,6 +62,15 @@ def bind_mount(self) -> None:
                 ['mount', '-n', '--bind', self.device, self.mountpoint]
             )
 
+    def rbind_mount(self) -> None:
```
schaefi, line 65: Thanks nice extension to the mount manager

kiwi/container/setup/base.py
```diff
@@ -145,23 +145,21 @@ def setup_root_console(self):
...
         necessarily present in the container later is a know
-        limitation of this method and considered harmless
+        limitation of this method and considered harmless.
         """
         try:
-            data = DataSync('/dev/', self.root_dir + '/dev/')
-            data.sync_data(
-                options=['-a', '-x', '--devices', '--specials']
-            )
+            mm = MountManager('/dev', self.root_dir + '/dev')
```
schaefi, line 161: I think the reason for the sync of the device nodes was that in former times a container had to have some device nodes statically available as part of the container archive.

>Without subsystems like udev running in a container it is  required to provide a set of device nodes to let the
 system in the container function correctly.

This is why we rsynced them. With this change the device nodes are available at build time but inside of the container ```/dev``` will stay empty

If this is no longer an issue your change is just perfect. However if container engines still requires some /dev/
nodes e.g console, tty or similar to be present to be able to create an instance we should not change it

If we stick to the mounting part an umount is missing. I suggest to add a destructor to the ContainerSetupBase class and keep the instance of MountManager in class scope (self.mm)

schaefi (changes requested):
Good catch, thanks for the changes. Please see my comments

kiwi/container/setup/base.py
ignatenkobrain: > This is why we rsynced them. With this change the device nodes are available at build time but inside of the container /dev will stay empty

I don't think it is needed on my system, but I don't know about others…

schaefi: I think @davidcassany has more insights on that part. I tested with podman on non privileged containers which just worked. I tested with privileged ones and also with containers that runs systemd inside and they worked for me if I did ```-v /dev:/dev``` but the workload also used loops and created dev nodes, thus /dev had to exist.

I cannot foresee if this breaks existing workloads, really no idea but I would be willing to take the risk since a static copy from the host into the container is also not really a nice solution

The last word is on David though :)

schaefi:
btw the failing tests are not related to your changes. See #2196

schaefi:
@ignatenkobrain ```git rebase master``` will fix the unit test failures now. Thanks

kiwi/container/setup/base.py
davidcassany: > I cannot foresee if this breaks existing workloads, really no idea but I would be willing to take the risk since a static copy from the host into the container is also not really a nice solution

I second this, I am also not capable to see if this will cause issues in some very specific cases. IMHO the expectation from contianer engines is that no /dev is required,  in fact note the podman and docker both include a `--tty` to allocate a pseudo terminal driver on the fly. Also binding hosts /dev with containers /dev should be a valid solution if there is a weird case in which the proposed change causes an issue.

If at some point there is a container requiring some device that was present in builder host that is not present in runtime host I'd consider it a bug on the container design and unrelated to KIWI.

davidcassany (approved):

kiwi/container/setup/base.py
ignatenkobrain: so I will remove the static_dev setup completely and that's it?

schaefi: > so I will remove the static_dev setup completely and that's it?

That would be ok for me yes. But please let us make sure that the rootfs tree contains a ```/dev``` directory such that ```-v /dev/:/dev``` stays possible. Thanks

ignatenkobrain: when you are doing mounts with docker/podman, it does not matter if directory exists or not. the only case where directory would need to exist is when you do manual mount which I believe is not the case..

schaefi (approved):
looks good :+1:

PR #2196, opened 2022-10-15
Weaken type check on Result.load() return type

schaefi:
With an update of mypy the bound TypeVar is no longer allowed. To be honest I did not get a clue on the error message and opened this pull request which basically removes the type check for the return type of the load() method. I'm happy if we find a better solution during review. Thanks

### DIFF
--- a/kiwi/system/result.py
+++ b/kiwi/system/result.py
@@ -162,7 +162,7 @@ class Result:
             )
 
     @staticmethod
-    def load(filename: str) -> result_type:
+    def load(filename: str) -> result_type:  # type: ignore
         """
         Load pickle dumped filename into a Result instance
 


kiwi/system/result.py
```diff
@@ -162,7 +160,7 @@ def dump(self, filename: str, portable: bool = True) -> None:
             )
 
     @staticmethod
-    def load(filename: str) -> result_type:
```
jesusbv, line 165: I get `error: A function returning TypeVar should receive at least one argument containing the same Typevar` 
This issue AFAIU comes from [this check](https://github.com/python/mypy/pull/13166), hope that clarifies the error message
IMO, without debugging more, a possible solution could be to ignore it, such as
```python
def load(filename: str) -> result_type  # type: ignore[type-var]
```

schaefi: That I tried which results in

```
check run-test: commands[0] | flake8 --statistics -j auto --count /home/ms/Project/kiwi/kiwi
kiwi/system/result.py:165:32: F821 undefined name 'Result'
1     F821 undefined name 'Result'
1
```

which was the reason for using

```
result_type = TypeVar('result_type', bound='Result')
```

jesusbv: Yes, but that is fixable adding
```python
from __future__ import annotations
```
at the beggining of the file ([more info](https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations) notice _It will become the default in Python 3.10._ in that doc)
or 
```python
def load(filename: str) -> Result:  # noqa: F821
```
Now, it is a decision of 
- do we want to keep `Result` (type method returns) ? My answer is yes
- do we want to add  `#noqa` ? My opinion is no, because that would hide possible errors but no strong opinion
- do we want to remove the type ? I think we should keep it

my 2 cents

Conan-Kudo: we could just make SUSE build this against a newer Python in SLE 15. They have Python 3.6, 3.8, and 3.10 now.

schaefi:
Maybe the last change did the trick...

kiwi/system/result.py
```diff
@@ -19,9 +19,14 @@
 import simplejson
 import pickle
 import os
+import sys
 from typing import (
     Dict, NamedTuple, TypeVar, Any
 )
+if sys.version_info[:2] >= (3, 7):  # pragma: no cover
+    from typing import Self
```
jesusbv, line 27: AFAIK, `Self` will be [available on Python 3.11](https://peps.python.org/pep-0673/)

another solution, following that approach could be

```python
if sys.version_info[:2] >= (3, 7):  # pragma: no cover
    from __future__ import annotations
else:
    Result = TypeVar('Result', bound='Result')
...
...
        @staticmethod
        def load(filename: str) -> Result
```

schaefi: unfortunately not as you can not redefine Result

schaefi: Also

```SyntaxError: from __future__ imports must occur at the beginning of the file```

seems like no if clause is allowed to do this conditionally

Comment (schaefi) [reply kiwi/system/result.py:27]: and yes Self exists with python 3.11 not earlier... so that try also did not work

Comment (schaefi) [reply kiwi/system/result.py]: I hate to say that, but after several testing of ideas shared here I still think the less intrusive change to keep type annotations working with python 3.6 still supported is the proposed "weaken" change

I agree SLE15 provides alternative interpreters but plain Leap15 without backports doesn't and I think 3.6 might still be in use elsewhere.

Thoughts ?

Comment (Conan-Kudo) [reply kiwi/system/result.py]: I'm fine with weakening...

Comment (jesusbv) [reply kiwi/system/result.py]: With the original code, adding ` # type: ignore[type-var]` to the method signature works, no errors in mypy no errors in flake8

```python
def load(filename: str) -> result_type:  # type: ignore[type-var]
```
```bash
> mypy kiwi/system/result.py
Success: no issues found in 1 source file
> flake8 --statistics -j auto --count kiwi/system/result.py
0
```

If decided to remove the type, that is fine by me as well

schaefi: > With the original code, adding ` # type: ignore[type-var]` to the method signature works, no errors in mypy no errors in flake8

yes this seems to be the most efficient way to ignore it for the moment. Thanks much @jesusbv

jesusbv: TL;DR
Solution is not to use 
```python
def load(filename: str) -> result_type:  # type: ignore[type-var]
```
but
```python
def load(filename: str) -> result_type:  # type: ignore
```


That's weird
```bash
> mypy --version
mypy 0.981 (compiled: no)
> mypy kiwi/system/result.py
Success: no issues found in 1 source file
> python3.8 -m mypy --version
mypy 0.790
> python3.8 -m mypy kiwi/system/result.py
Success: no issues found in 1 source file
> tox -e unit_py3_8
.....
unit_py3_8 run-test: commands[1] | bash -c 'cd ../../ && mypy kiwi'
kiwi/system/result.py:118: note: Consider using the upper bound "Result" instead
kiwi/system/result.py:118: note: Error code "misc" not covered by "type: ignore" comment
Success: no issues found in 183 source files
ERROR: InvocationError for command /bin/bash -c 'cd ../../ && mypy kiwi' (exited with code 1)
_________________________________________________________ summary __________________________________________________________
ERROR:   unit_py3_8: commands failed
> mypy kiwi
Success: no issues found in 183 source files
> bash -c 'mypy kiwi'
Success: no issues found in 183 source files
> cd test/unit/
> bash -c 'cd ../../ && mypy kiwi'
Success: no issues found in 183 source files
```
From that output I understand `mypy` is OK but the way to call it inside `kiwi` or something else related to the way it is invoked is borked/has an issue

Taking a look at more verbose output, mypy version is 0.982 which also says no errors
```bash
> tox -rvve unit_py3_8
...
Collecting mypy
  Using cached mypy-0.982-cp38-cp38-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (17.4 MB)
  ...
[notice] A new release of pip available: 22.2.2 -> 22.3
[notice] To update, run: pip install --upgrade pip
unit_py3_8 finish: getenv path/to/kiwi/.tox/3.8 after 40.05 seconds
....
Success: no issues found in 183 source files
ERROR: InvocationError for command /bin/bash -c 'cd ../../ && mypy kiwi' (exited with code 1)
unit_py3_8 finish: run-test  after 3.09 seconds
unit_py3_8 start: run-test-post 
unit_py3_8 finish: run-test-post  after 0.00 seconds
```
So, I can confirm something is borked, just not in `kiwi`
errors shown when running `tox` with `# type: ignore]type-var]` however, if the code is
```python
    def load(filename: str) -> result_type:  # type: ignore
```
outputs are:
```bash
> tox -e unit_py3_8
unit_py3_8 run-test: commands[1] | bash -c 'cd ../../ && mypy kiwi'
Success: no issues found in 188 source files
_______________________________________________ summary _______________________________________________
  unit_py3_8: commands succeeded
  congratulations :)
> tox -e unit_py3_10
Finished processing dependencies for kiwi==9.24.48
unit_py3_10 run-test: commands[1] | bash -c 'cd ../../ && mypy kiwi'
Success: no issues found in 188 source files
_______________________________________________ summary _______________________________________________
  unit_py3_10: commands succeeded
  congratulations :)
> 
```
Not sure if there is much interest in debugging why it is happening so, I may debug a bit to see the difference between `type: ignore[type-var]` vs `type: ignore` but I understand we are OK with that comment to ignore the error

jesusbv (approved):
Thanks for the changes

LGTM

schaefi:
> Thanks for the changes
> 
> LGTM

Thanks to you for all the debugging and input, much appreciated. The final change is super simple now :)

PR #2197, opened 2022-10-25
Add full URL to the Sources line

dirkmueller:
This is needed for being able to validate the origin of the sources

Fixes # .

*
*


Conan-Kudo (changes requested):
This is not valid. The tarballs submitted to OBS do not match our PyPI nor our Git tarballs.

dirkmueller:
okay, I see it contains generated files. well then maybe those tarballs should be uploaded as artifacts when released?

PR #2199, opened 2022-10-29
Added support for --logsocket

schaefi:
Like with --logfile this commit adds support for using an existing Unix Domain Socket for logging. It's required that there is a listener on the given socket otherwise kiwi exits with an appropriate error message from the socket layer. A simple listener could look like the following:

```python
sock_file = '/tmp/log_socket'
buffer = 1024

if os.path.exists(sock_file):
    os.unlink(sock_file)

sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
sock.bind(sock_file)
sock.listen(1)

while True:
    connection, client_address = sock.accept()
    try:
        while True:
            data = connection.recv(buffer)
            if not data:
                break
            print(data.decode())
    finally:
        connection.close()
```

With the listener in place kiwi can be called as follows:

```
kiwi-ng --logsocket /tmp/log_socket ...
```

### DIFF
--- a/kiwi/cli.py
+++ b/kiwi/cli.py
@@ -22,6 +22,7 @@ usage: kiwi-ng -h | --help
                [--target-arch=<name>]
                [--type=<build_type>]
                [--logfile=<filename>]
+               [--logsocket=<socketfile>]
                [--debug]
                [--debug-run-scripts-in-screen]
                [--color-output]
@@ -29,6 +30,7 @@ usage: kiwi-ng -h | --help
                [--kiwi-file=<kiwifile>]
            image <command> [<args>...]
        kiwi-ng [--logfile=<filename>]
+               [--logsocket=<socketfile>]
                [--debug]
                [--debug-run-scripts-in-screen]
                [--color-output]
@@ -40,6 +42,7 @@ usage: kiwi-ng -h | --help
                [--target-arch=<name>]
                [--type=<build_type>]
                [--logfile=<filename>]
+               [--logsocket=<socketfile>]
                [--debug]
                [--debug-run-scripts-in-screen]
                [--color-output]
@@ -63,6 +66,9 @@ global options:
         debug information even if this is was not requested by the
         debug switch. The special call: '--logfile stdout' sends all
         information to standard out instead of writing to a file
+    --logsocket=<socketfile>
+        send log data to the given Unix Domain socket in the same
+        format as with --logfile
     --debug
         print debug information
     --debug-run-scripts-in-screen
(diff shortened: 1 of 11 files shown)


isbm:
Can we control level here? Like `--log-level=error` and you see only failures. That would be interesting to capture only failures and not necessarily dump to a _log stream_ elsewhere, but put some popups etc. And also would be good to know how to distinguish that it is a log. E.g. if receiver socket is accepting various things, how to say "this is a log message and this is some other command"?

schaefi:
> Can we control level here? Like `--log-level=error` and you see only failures. That would be interesting to capture only failures and not necessarily dump to a _log stream_ elsewhere, but put some popups etc. And also would be good to know how to distinguish that it is a log. E.g. if receiver socket is accepting various things, how to say "this is a log message and this is some other command"?

we can but I like to do this in a followup pull request because --log-level would have a general effect on all logging code and I don't want to mix implementations. If you can add this as an issue it would be good, so we don't forget. Thanks

isbm:
OK, that makes perfect sense. Seems there is only one level for now :wink: And in this case yes, it requires another PR.

isbm:
Can we also control the content? Say `--log-prefix="BLERGH:"` so then all the messages are prefixed with it like so:

    [PREFIX]<MESSAGE>

In this case no ad-hoc solutions needed. Say, the receiver will use that prefix for filtering-out messages. And if you want to simply push that to a socket receiver as-is, no need to define `--log-prefix`.

kiwi/cli.py
```diff
@@ -22,13 +22,15 @@
                [--target-arch=<name>]
                [--type=<build_type>]
                [--logfile=<filename>]
+               [--logsocket=<socketfile>]
```
jesusbv, line 25: so AFAIU `--logsocket` and `--logfile` are not mutually exclusive and can be used together,
what would happen if Kiwi is run with both option pointing at the same path ? Like
`kiwi-ng --logsocket /tmp/log_socket --logfile /tmp/log_socket`

schaefi: > so AFAIU `--logsocket` and `--logfile` are not mutually exclusive and can be used together, what would happen if Kiwi is run with both option pointing at the same path ? Like `kiwi-ng --logsocket /tmp/log_socket --logfile /tmp/log_socket`
> 
> I understand that case should not be an issue, right ?

yes not an issue. You can use them together. As I wrote in the initial comment you need a listener for the specified logsocket. The python logging facility connects to the specified Unix Doman Socket. If there is no listener (aka server) that connect will fail and kiwi stops with an error message produced by that connection attempt. In the initial commit I posted a very simple example of such a listener such that people could try it out. You could also use ```--logsocket /dev/log``` which usually exists on Linux and points to syslog. In this case the kiwi logs appears on the syslog interface

jesusbv (approved):
LGTM

PR #2200, opened 2022-11-03
Logging improvements

schaefi:
This change is two fold

**Added --loglevel option**
    
specify logging level as number. Details about the available log levels can be found at:   https://docs.python.org/3/library/logging.html#logging-levels. Setting a log level causes all
message >= level to be displayed.

**Consolidate and cleanup logging**
    
Make sure all loggers; stream handlers, file and socket handler uses the same logging format. Also make sure that there is only one place for setLogLevel when kiwi is used as application

### DIFF
--- a/kiwi/tasks/base.py
+++ b/kiwi/tasks/base.py
@@ -28,7 +28,6 @@ from kiwi.cli import Cli
 from kiwi.xml_state import XMLState
 from kiwi.xml_description import XMLDescription
 from kiwi.runtime_checker import RuntimeChecker
-from kiwi.runtime_config import RuntimeConfig
 
 from kiwi.exceptions import (
     KiwiConfigFileNotFound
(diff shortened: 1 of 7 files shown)


kiwi/tasks/base.py
```diff
@@ -127,9 +113,35 @@ def __init__(self, should_perform_task_setup: bool = True) -> None:
...
+                try:
+                    log.setLogLevel(int(self.global_args['--loglevel']))
+                except ValueError:
+                    # Not a numeric log level, stick with the default
+                    # which is DEBUG
+                    pass
+            elif self.global_args['--debug']:
+                log.setLogLevel(logging.DEBUG)
+            else:
+                log.setLogLevel(logging.INFO)
```
jesusbv, line 127: is this where the default logging level is set ? If so, shouldn't it be `DEBUG` as mentioned in line 122,
or update line 122 ?

schaefi: good point, line 122 is wrong :)

jesusbv (approved):
Thank you for the changes

LGTM

PR #2204, opened 2022-11-12
Add GRUB_DEFAULT=saved if not set

### LINKED ISSUE #2198: RHEL based distro (CentOS 7 and AlmaLinux 8) images don't enable freshly installed kernels automatically
First of all I would like to say thank you for this great project (I've donated some money recently so you can at least have 1 coffee on me). I rely on Kiwi to automatically build my images for Hoster (https://github.com/yaroslav-gwit/HosterRed-HyperVisor) which is just a PoC at this point but will be soon used in production.

## Problem description
RHEL based images, namely CentOS 7 and AlmaLinux 8, don't enable freshly installed kernels for some reason.
Every time there is a new kernel installed I have to re-generate the grub config, like so, in order to activate a new kernel:
```
grub2-set-default 0
grub2-mkconfig -o /boot/grub2/grub.cfg
```

* KIWI version: 9.24.36
* Operating system host version: OpenSuSe Leap 15.4
* Operating system target version: CentOS 7/AlmaLinux 8

schaefi:
If there is no GRUB_DEFAULT setting in etc/default/grub we add GRUB_DEFAULT=saved as the default setting.
This Fixes #2198

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -634,6 +634,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         Create or update etc/default/grub by parameters required
         according to the root filesystem setup
 
+        * GRUB_DEFAULT
         * GRUB_TIMEOUT
         * GRUB_TIMEOUT_STYLE
         * SUSE_BTRFS_SNAPSHOT_BOOTING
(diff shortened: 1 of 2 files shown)


schaefi:
ignore the gitlab CI issue, we will stop using it as it has turned into a non free service

Conan-Kudo (approved):

PR #2205, opened 2022-11-13
Added systemd_boot integration test for Rawhide

schaefi:
The integration test, tests the systemd_bootloader. The test builds here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:rawhide/test-image-systemd-boot

and can run in QEMU as follows:

```
kvm -bios /usr/share/qemu/ovmf-x86_64-ms.bin -hda kiwi-test-image-systemd-boot.x86_64-2.0.0-Build2.1.qcow2
```


build-tests/x86/rawhide/test-image-systemd-boot/appliance.kiwi
```diff
@@ -0,0 +1,67 @@
...
+    <description type="system">
+        <author>Marcus Schaefer</author>
+        <contact>marcus.schaefer@gmail.com</contact>
+        <specification>Fedora Rawhide Appliance</specification>
+    </description>
+    <preferences>
+        <version>2.0.0</version>
+        <packagemanager>dnf</packagemanager>
+        <bootsplash-theme>charge</bootsplash-theme>
+        <bootloader-theme>breeze</bootloader-theme>
```
Conan-Kudo, line 13: This is useless here.

schaefi:
@Conan-Kudo as mentioned on matrix there is a patch which should go upstream to systemd.

I did not yet investigate where and how to contribute to systemd upstream. I guess it will be https://github.com/systemd/systemd

build-tests/x86/rawhide/test-image-systemd-boot/appliance.kiwi
```diff
@@ -0,0 +1,67 @@
...
+    </repository>
+-->
+<!--
+    <repository type="rpm-md" alias="Fedora38" sourcetype="metalink">
+        <source path="https://mirrors.fedoraproject.org/metalink?repo=fedora-38&amp;arch=x86_64"/>
+    </repository>
+    <repository type="rpm-md" alias="Fedora38-updates" sourcetype="metalink">
+        <source path="https://mirrors.fedoraproject.org/metalink?repo=updates-released-f38&amp;arch=x86_64"/>
+    </repository>
+-->
```
Conan-Kudo, line 48: Drop this it makes no sense.

build-tests/x86/rawhide/test-image-systemd-boot/appliance.kiwi
```diff
@@ -0,0 +1,67 @@
...
+    <packages type="image">
+        <!-- without grubby there are no kernels in /boot for some reason -->
+        <package name="grubby"/>
```
Conan-Kudo, line 51: We don't need this.

Conan-Kudo (changes requested):

build-tests/x86/rawhide/test-image-systemd-boot/appliance.kiwi
schaefi: Problem is if I don't install grubby, for some reason, all kernel files vanishes from /boot

Conan-Kudo:
> @Conan-Kudo as mentioned on matrix there is a patch which should go upstream to systemd.
> 
> I did not yet investigate where and how to contribute to systemd upstream. I guess it will be https://github.com/systemd/systemd

@schaefi Yes, that's the correct place.

build-tests/x86/rawhide/test-image-systemd-boot/appliance.kiwi
Conan-Kudo: But if they're in `/boot`, sd-boot can't load them either, no? They need to be in the ESP.

build-tests/x86/rawhide/test-image-systemd-boot/appliance.kiwi
schaefi: yes, fixed

build-tests/x86/rawhide/test-image-systemd-boot/appliance.kiwi
schaefi: yes, just a left over, saw it too late ;) done

build-tests/x86/rawhide/test-image-systemd-boot/config.sh
```diff
@@ -0,0 +1,30 @@
...
+#======================================
+# Activate services
+#--------------------------------------
+baseInsertService dbus-broker
+baseInsertService NetworkManager
+
+#======================================
+# Setup default target, multi-user
+#--------------------------------------
+baseSetRunlevel 3
```
Conan-Kudo, line 30: we really should have a `baseSetTarget` function instead, because we can't guarantee that the runlevel symlinks for targets exist.

build-tests/x86/rawhide/test-image-systemd-boot/appliance.kiwi
schaefi: what I meant is this. Usually the /boot partition on Fedora is similar to other /boot locations. On my tests it looks like this
```
/boot/System.map-6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64
/boot/config-6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64
/boot/efi
/boot/initramfs-6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64.img
/boot/symvers-6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64.gz -> /lib/modules/6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64/symvers.gz
/boot/vmlinuz-6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64
```
according to the package database the kernel (vmlinuz) is provided by ```kernel-core```. However, if you don't install ```grubby``` and not other component of grub the ```/boot``` partition does not contain the kernel anymore. I did not investigate why this is the case and just added grubby to the package list plus the comment

> But if they're in /boot, sd-boot can't load them either, no? They need to be in the ESP.

since kiwi checks if there is any kernel /boot we don't reach the point when sd-boot tooling comes into play

Conan-Kudo: So we have a broken assumption in kiwi then? Because when we're using sd-boot, we need a flag for whether we're using `/boot` or `/boot/efi` for that. To be clear, we can use `/boot` just fine if we install an EFI filesystem driver for the `/boot` partition. Fedora has the `efifs` package which includes various filesystem EFI drivers that we could install.

build-tests/x86/rawhide/test-image-systemd-boot/config.sh
schaefi: We did not change the name of the method for compat reasons. However, I think the code behind it does things correctly for systemd. Please double check: https://github.com/OSInside/kiwi/blob/master/kiwi/config/functions.sh#L530 if you disagree.

Regarding the naming I have no objections to add a baseSetTarget, the times of sysV init are mostly over and it could be added in addition to avoid the runlevel terminology. If this should be needed it can be done in a followup request but not here please

Conan-Kudo: Sure, I'm just noting we should probably have an alias or something here added at some point.

build-tests/x86/rawhide/test-image-systemd-boot/appliance.kiwi
schaefi: sorry I can't follow you. If fedora has a package called ```kernel-core``` and it provides ```/boot/vmlinuz-xxxx``` then I expect this file to be there when I install this package. This is not the case here. There is some magic that runs in package scripts or whatever I don't know but the kernel is definitely missing in /boot if I don't install grubby and that feels pretty much like a bug. If kiwi would not run the kernel lookup check before bootctl or kernel-install the process would fail when these tools are called.

I do not understand how this is related to EFI or bootloader at all... there is no kernel anymore if I don't install grubby

Conan-Kudo: How are we doing the kernel lookup? Because kernel-core only provides those paths as ghost files...

I think I don't understand what's going on, maybe @DaanDeMeyer or @martinezjavier might understand this better.

schaefi: > Because kernel-core only provides those paths as ghost files

interesting. So what creates ```/boot/vmlinuz-6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64``` and where is its origin ? If ```/boot``` cannot be trusted to contain a kernel I agree with you then the kiwi kernel lookup needs to be adapted to look elsewhere, wherever that is ? :)

schaefi: Also it looks like this is now going beyond this PR, if we need to lookup kernels in other places on Fedora that sounds like a different issue. I would also like to understand why under certain circumstances there are kernels in /boot and in other not. It feels very odd to me at the moment

schaefi: to what package belongs the kernel in Fedora, it's really strange. ```kernel``` contains nothing, ```kernel-core``` is a ghost, so what ??

build-tests/x86/rawhide/test-image-systemd-boot/config.sh
schaefi: yes agreed, I put it on my list :+1:

schaefi: https://github.com/OSInside/kiwi/issues/2206

build-tests/x86/rawhide/test-image-systemd-boot/appliance.kiwi
Conan-Kudo: `kernel` is a metapackage to pull in everything else.

DaanDeMeyer: In mkosi, when generating a bootable fedora image, we install the following packages:

- systemd
- util-linux
- dnf
- kernel-core
- kernel-modules
- dracut
- systemd-udev

That should be sufficient to get kernel-install to pick up the fedora kernels

schaefi: @DaanDeMeyer yep also using these packages. The issue I'm facing is that ```/boot``` does not provide the kernel. I was confused about the following:

```
rpm -ql kernel-core | grep vmlinuz
/boot/.vmlinuz-6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64.hmac
/boot/vmlinuz-6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64
/lib/modules/6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64/.vmlinuz.hmac
/lib/modules/6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64/vmlinuz
```

From here I thought ok there is still the kernel in /boot. But as @Conan-Kudo mentioned that's just a ```%ghost```
so it does not really exist. And that leads in kiwi to

```
[ ERROR   ]: 10:45:34 | KiwiKernelLookupError: No kernel found in ...
```

If you install any package related to grub e.g ```grubby``` this problem "resolves" because for some reason
now kernel files are also in ```/boot```

As I like to support systemd-boot without any grub components I was thinking where is the kernel now ?

So I assume Fedora packages the kernel in

```
/lib/modules/6.1.0-0.rc4.20221111git4bbf3422df78.39.fc38.x86_64/vmlinuz
```

Is this correct ?

If so I need to add another search path in the kiwi code that looks up kernels

Can you confirm/advise ?

Thanks much

DaanDeMeyer: Ah, yes, if you're running `kernel-install` manually, then you'll need to look up kernels in /usr/lib/modules. Same goes for Arch Linux. Gentoo puts them in a weird location. Debian/Ubuntu put them in /boot.

schaefi: yep I finally understood the concept, now it's clear. I pushed a commit here to extend the kernel lookup of kiwi to
* /boot
* /lib/modules
* /usr/lib/modules

With this change the integration builds and works without any grub components

Thanks

Conan-Kudo:
@schaefi can you at least pull out the lookup commit and make a separate PR so we can merge that without this?

schaefi:
> @schaefi can you at least pull out the lookup commit and make a separate PR so we can merge that without this?

yes I will

schaefi:
@DaanDeMeyer @Conan-Kudo 

I created an upstream patch suggestion for systemd and kernel-install here:

https://github.com/systemd/systemd/pull/25388

Hope this makes sense. Thanks

Conan-Kudo:
> @Conan-Kudo my patches to systemd got accepted and merged. I'd like to keep the custom patch in the integration test here and get it merged. Once the target distro has updated systemd we will see the test to fail because the patch will no longer apply and then we can create a follow up PR deleting the custom patch again.
> 
> Thoughts ?

This was fixed in Rawhide yesterday and should be making its way out to mirrors today.

build-tests/x86/rawhide/test-image-systemd-boot/config.sh
```diff
@@ -0,0 +1,35 @@
...
+echo "Configure image: [$kiwi_iname]..."
+
+#======================================
+# Fix kernel-install
+#--------------------------------------
+# patch sent upstream and merged. This patching here can
+# be deleted as soon as the main distro has an update for
+# systemd available
+patch -p0 < kernel-install.patch
+rm -f kernel-install.patch
```
Conan-Kudo, line 24: This patch should no longer be needed. The upstreamed version is now in Rawhide.

Conan-Kudo (changes requested):

build-tests/x86/rawhide/test-image-systemd-boot/config.sh
schaefi: great, thanks. I'll give it a test after work today

schaefi: done, checking if the build test completes

schaefi:
@Conan-Kudo ok the integration test builds and works, no extra patches needed anymore ;) I think we are good to go now

PR #2211, opened 2022-11-28
Swap from DM kpartx to util-linux partx for creating loop devices

### LINKED ISSUE #2208: Switch from using kpartx (DM-based) to partx (simple loop based)
## Problem description
KIWI currently uses device-mapper backed loop devices for creating disks:

https://github.com/OSInside/kiwi/blob/7fdf127e89ac97b45e140b99112b059b9df5bb84/kiwi/storage/disk.py#L394-L409

https://github.com/OSInside/kiwi/blob/7fdf127e89ac97b45e140b99112b059b9df5bb84/kiwi/storage/disk.py#L479-L499

https://github.com/OSInside/kiwi/blob/7fdf127e89ac97b45e140b99112b059b9df5bb84/kiwi/storage/disk.py#L501-L514

This causes all kinds of problems when trying to use it in various build environments.

From what I can tell, there doesn't seem to be a reason for us to use this over the plain `partx` provided by util-linux. The only real restriction is that if you're trying to create disk images with LVM in them, you'll hit the same problems we have now, but that drastically reduces the scope of issues that we have now when trying to run kiwi in build environments or container environments.

The basic flow for using plain `partx` is like follows (snipped from some other shell code):

```shell
# Create a 120GB disk image file in the temp directory
disk_size_in_gb=120

fallocate -l "${disk_size_in_gb}G" "${work_dir}/${image_name}.img"

# Instantiate a loop device backed by the image file
loopdev=$(losetup --show -f "${work_dir}/${image_name}.img")

# [...]

# Remove loop partition devices
partx -d "${loopdev}"

# Remove loop device
losetup -d "${loopdev}"

```

* KIWI version: 9.24.48
* Operating system host version: CentOS Stream 9
* Operating system target version: CentOS Stream 9
* Koji version (N/A if not using Koji): 1.30.1

Conan-Kudo:
The device-mapper (DM) based loop devices that kiwi has historically generated creates issues when trying to run kiwi in confined build environments. The DM tools prefer to have a tighter coupling between userspace and kernel interfaces, and we cannot necessarily guarantee that in the variety of build environments that kiwi can be run in. In particular, Koji uses either nspawn containers or chroots through Mock to run kiwi and that is where this fails.

However, we do not need to use DM for this purpose when util-linux provides a perfectly serviceable alternative. This commit changes kiwi's loop device setup to use `partx(8)` from util-linux instead.

As long as the appliance image being created doesn't use a disk scheme that requires DM, it becomes entirely possible to produce an image without needing DM at runtime at all.

Fixes #2208.

### DIFF
--- a/kiwi/storage/disk.py
+++ b/kiwi/storage/disk.py
@@ -400,7 +400,7 @@ class Disk(DeviceProvider):
         """
         if self.storage_provider.is_loop():
             Command.run(
-                ['kpartx', '-s', '-a', self.storage_provider.get_device()]
+                ['partx', '--add', self.storage_provider.get_device()]
             )
             self.is_mapped = True
         else:
@@ -482,7 +482,7 @@ class Disk(DeviceProvider):
         if self.storage_provider.is_loop():
             device_base = os.path.basename(self.storage_provider.get_device())
             device_node = ''.join(
-                ['/dev/mapper/', device_base, 'p', partition_number]
+                ['/dev/', device_base, 'p', partition_number]
             )
         else:
             device = self.storage_provider.get_device()
@@ -503,9 +503,9 @@ class Disk(DeviceProvider):
             log.info('Cleaning up %s instance', type(self).__name__)
             try:
                 for device_node in self.partition_map.values():
-                    Command.run(['dmsetup', 'remove', device_node])
+                    Command.run(['partx', '--delete', device_node])
                 Command.run(
-                    ['kpartx', '-d', self.storage_provider.get_device()]
+                    ['partx', '--delete', self.storage_provider.get_device()]
                 )
             except Exception:
                 log.warning(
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
cc: @dwminer, @iammattcoleman, @ignatenkobrain

Conan-Kudo:
cc: @davide125, @michel-slm

Conan-Kudo:
cc: @tkopecek

schaefi:
Thanks @Conan-Kudo :+1: much appreciated. The way this change is done looks clean and good to me. As you already noted there are other subsystems for which the device mapper will be used. LVM is one of them but there are also others like ```raid```, ```luks```, ```integritysetup``` for which the device mapper is still in use.

So for those features we will still have to deal with the DM ABI compatibility between host and build environment.
Reducing the set using partx over kpartx is perfectly fine.

I was just questioning myself; Is the change backward compatible to the older distros we still support ? RHEL, SLES ?

Thanks

Conan-Kudo:
I reworded the commit change from mentioning LVM to say "disk scheme" instead, since we have multiple disk setups that require DM.

schaefi:
Did several image build tests and works for me now :+1:

schaefi (approved):

PR #2212, opened 2022-11-30
Support repo URI's with credentials on cmdline

schaefi:
Specifying a repository as part of the image description allows for credentials via the username and password attributes. Howver, repositories can also be specified on the commandline via the --set-repo / --add-repo options. The options on the commandline did not allow to specify credentials so far. This commit adds the commandline options --set-repo-credentials and --add-repo-credentials to support them

### DIFF
--- a/kiwi/tasks/base.py
+++ b/kiwi/tasks/base.py
@@ -280,7 +280,7 @@ class CliTask:
 
         :rtype: list
         """
-        tokens = option.split(',', tuple_count - 1)
+        tokens = option.split(',', tuple_count - 1) if option else []
         return [
             self._pop_token(tokens) if len(tokens) else None for _ in range(
                 0, tuple_count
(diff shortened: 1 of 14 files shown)


kiwi/tasks/base.py
```diff
@@ -280,7 +280,7 @@ def _ntuple_token(
 
         :rtype: list
         """
-        tokens = option.split(',', tuple_count - 1)
+        tokens = option.split(',', tuple_count - 1) if option else []
```
jesusbv, line 283: This seems a fix for the commandline options in general not specifically for 'adding credentials to commandline'  feature ?

jesusbv (approved):
LGTM

kiwi/tasks/base.py
schaefi: correct, stumbled over it when testing and thought it can't hurt to fix it in the same slot

PR #2213, opened 2022-11-30
Eliminate redundant partx calls for cleaning up loop devices

Conan-Kudo:
partx can clean up all partition loops associated with a loop device when deleting the main loop device. Apparently, sometimes it goes and does this even when only deleting the partition loop, so to avoid this problem, we will just eliminate the redundant call.

Fixes: 8f2b8fda82a37d7a7ebbf8a6557870cc63b0ae91


schaefi (approved):
Thanks :+1:

PR #2214, opened 2022-12-01
Allow to select partition mapper tool

schaefi:
The recent change from kpartx to partx caused some appliance builds to break depending on their configuration. I spotted issues when building disks with veritysetup or integritysetup root devices. There are also issues with grub-install on other architectures e.g s390. It seems partx cannot be used as a drop in replacement and so I suggest to make this a runtime configuration option with partx as the new default but also allow for the old method

### DIFF
--- a/kiwi/storage/disk.py
+++ b/kiwi/storage/disk.py
@@ -28,6 +28,7 @@ from kiwi.command import Command
 from kiwi.storage.device_provider import DeviceProvider
 from kiwi.storage.mapped_device import MappedDevice
 from kiwi.partitioner import Partitioner
+from kiwi.runtime_config import RuntimeConfig
 from kiwi.exceptions import KiwiCustomPartitionConflictError
 
 ptable_entry_type = NamedTuple(
@@ -66,6 +67,7 @@ class Disk(DeviceProvider):
             partitions will be placed as logical partitions inside
             of that extended partition
         """
+        self.partition_mapper = RuntimeConfig().get_mapper_tool()
         # bind the underlaying block device providing class instance
         # to this object (e.g loop) if present. This is done to guarantee
         # the correct destructor order when the device should be released.
@@ -399,9 +401,14 @@ class Disk(DeviceProvider):
         required to map them if the storage provider is loop based
         """
         if self.storage_provider.is_loop():
-            Command.run(
-                ['partx', '--add', self.storage_provider.get_device()]
-            )
+            if self.partition_mapper == 'kpartx':
+                Command.run(
+                    ['kpartx', '-s', '-a', self.storage_provider.get_device()]
+                )
+            else:
+                Command.run(
+                    ['partx', '--add', self.storage_provider.get_device()]
+                )
             self.is_mapped = True
         else:
             Command.run(
@@ -481,9 +488,14 @@ class Disk(DeviceProvider):
         partition_number = format(self.partitioner.get_id())
         if self.storage_provider.is_loop():
             device_base = os.path.basename(self.storage_provider.get_device())
-            device_node = ''.join(
-                ['/dev/', device_base, 'p', partition_number]
-            )
+            if self.partition_mapper == 'kpartx':
+                device_node = ''.join(
+                    ['/dev/mapper/', device_base, 'p', partition_number]
+                )
+            else:
+                device_node = ''.join(
+                    ['/dev/', device_base, 'p', partition_number]
+                )
         else:
             device = self.storage_provider.get_device()
             if device[-1].isdigit():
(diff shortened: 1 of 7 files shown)


Conan-Kudo:
If this is actually the case, shouldn't we also force one mode over the other when using specific features then?

Conan-Kudo:
I'm also suspicious of this dependency on `kpartx` for veritysetup and such, because `partx` loop devices are in the same class as regular disk devices, and you can do veritysetup and stack dm devices on top of regular disk devices. I'm pretty sure that also works with regular loop devices too.

schaefi:
> If this is actually the case, shouldn't we also force one mode over the other when using specific features then?

The problem here is that this question can not be easily answered. I mean it should work with partx in any case, shouldn't it ? so if we say we use kpartx if integritysetup is used it feels wrong because it should work but it does not.
It also doesn't fail on integritysetup but later when you want to install a bootloader

Also the behavior is very strange. If you use ```loop -> partx -> integritysetup -> grub``` it fails on the grub level with a grub error ```[ 1074s] Error: Could not get disk geometry``` very strange. If you use ```loop -> partx -> lvm -> grub``` it works.

So obviously there seems to be a difference between a device mapper mapped device and a geometry offset based
device and sending this information to the kernel.

Conan-Kudo:
> 
> Also the behavior is very strange. If you use `loop -> partx -> integritysetup -> grub` it fails on the grub level with a grub error `[ 1074s] Error: Could not get disk geometry` very strange. If you use `loop -> partx -> lvm -> grub` it works.
> 

This might actually be a bug in GRUB. Where did you observe this?

cc: @martinezjavier

schaefi:
> I'm also suspicious of this dependency on `kpartx` for veritysetup and such, because `partx` loop devices are in the same class as regular disk devices, and you can do veritysetup and stack dm devices on top of regular disk devices. I'm pretty sure that also works with regular loop devices too.

yes this is correct and they also take the device and work on them but if you are using veritysetup and/or integritysetup and write the metadata into the same device which I do and is also officially supported, the later grub cannot be installed anymore.

It is as you said "partx will work if not too weird things are done". However, I need this to stay functional also for the weird, or let's say more complex, configurations.

schaefi:
> > Also the behavior is very strange. If you use `loop -> partx -> integritysetup -> grub` it fails on the grub level with a grub error `[ 1074s] Error: Could not get disk geometry` very strange. If you use `loop -> partx -> lvm -> grub` it works.
> 
> This might actually be a bug in GRUB. Where did you observe this?
> 
> cc: @martinezjavier

at the job, and it's hard to share this here. But you can also see it on all s390 tests which uses zipl-to-grub. That would be here: https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_s390:tumbleweed

schaefi:
I'm afraid I won't have time to hunt done maybe bugs in other components when partx is in use. I have to have a way to get back to the working state. So I see your concerns and I agree to them, which is also the reason why I'm fine with partx as the default but I have to have a way to say use kpartx if that fixes certain layout configurations

Conan-Kudo:
Okay, fine. But we need to get bug reports and reproducers for these things, so we can later remove this knobs when they don't apply anymore.

kiwi/storage/disk.py
```diff
@@ -399,9 +401,14 @@ def map_partitions(self):
...
-                ['partx', '--add', self.storage_provider.get_device()]
-            )
+            if self.partition_mapper == 'partx':
+                Command.run(
+                    ['partx', '--add', self.storage_provider.get_device()]
+                )
+            else:
+                Command.run(
+                    ['kpartx', '-s', '-a', self.storage_provider.get_device()]
+                )
```
Conan-Kudo, line 411: This condition should be flipped. Only when `kpartx` is used should we use `kpartx`. That makes it obvious that `kpartx` is the special case.

kiwi/storage/disk.py
```diff
@@ -481,9 +488,14 @@ def _add_to_map(self, name):
...
-                ['/dev/', device_base, 'p', partition_number]
-            )
+            if self.partition_mapper == 'partx':
+                device_node = ''.join(
+                    ['/dev/', device_base, 'p', partition_number]
+                )
+            else:
+                device_node = ''.join(
+                    ['/dev/mapper/', device_base, 'p', partition_number]
+                )
```
Conan-Kudo, line 498: This condition should be flipped. Only when `kpartx` is used should we use `/dev/mapper` paths. That makes it obvious that `kpartx` is the special case.

kiwi/storage/disk.py
```diff
@@ -502,9 +514,16 @@ def __del__(self):
...
+                if self.partition_mapper == 'partx':
+                    Command.run(
+                        ['partx', '--delete', self.storage_provider.get_device()]
+                    )
+                else:
+                    for device_node in self.partition_map.values():
+                        Command.run(['dmsetup', 'remove', device_node])
+                    Command.run(
+                        ['kpartx', '-d', self.storage_provider.get_device()]
+                    )
```
Conan-Kudo, line 526: This condition should be flipped. Only when `kpartx` is used should we use `kpartx`/`dmsetup`. That makes it obvious that `kpartx` is the special case.

Conan-Kudo (changes requested):

kiwi/storage/disk.py
schaefi: ok good point

schaefi: done

kiwi/storage/disk.py
schaefi: done

kiwi/storage/disk.py
schaefi: done

Conan-Kudo (approved):
LGTM. I'll merge this after tests pass.

Conan-Kudo:
> > Okay, fine. But we need to get bug reports and reproducers for these things, so we can later remove this knobs when they don't apply anymore.
> 
> I'm going to file bug reports on the s390 side. I will also create a dm-integrity / dm-verity integration build that I expect to fail and we can point the grub people to it for further debugging

Excellent!

PR #2219, opened 2022-12-09
Fix unit test run for BLS spec base class

schaefi:
Do not really call blkid as used in get_boot_cmdline

### DIFF
--- a/.github/workflows/ci-code-style.yml
+++ b/.github/workflows/ci-code-style.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   unit_tests:
     name: Linter checks for KIWI python and Shell code
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         python-version: ["3.6", "3.8", "3.9", "3.10"]
--- a/.github/workflows/ci-publish-pages.yml
+++ b/.github/workflows/ci-publish-pages.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6]
+        python-version: [3.8]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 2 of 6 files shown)


schaefi:
Hmm, 

```
Run actions/setup-python@v2
Version 3.6 was not found in the local cache
Error: Version 3.6 with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

really ?

schaefi:
The test fails are not related to the actual change here, but if it's true we need to deactivate 3.6 checking from actions

Conan-Kudo:
We're going to need a way to make sure that we don't break things on Python 3.6 though. RHEL 8 and SLE 15 both use that Python version.

Conan-Kudo:
Maybe we could use this image? https://github.com/fedora-python/fedora-python-tox

schaefi:
> We're going to need a way to make sure that we don't break things on Python 3.6 though. RHEL 8 and SLE 15 both use that Python version.

yes I also don't want to delete this target. I think I found a way to deal with it. Reason is that runs-on: ubuntu-latest targets Ubuntu 22.04 now which does not longer have py3.6. I switched those workflows to runs-on: ubuntu-20.04. See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#choosing-github-hosted-runners for details

I switched those targets where the python version doesn't play a role to a newer python version

That should do the trick for now

schaefi:
Due to an update of tox I also had to adapt tox.ini to the new regulations by tox

schaefi:
> Maybe we could use this image? https://github.com/fedora-python/fedora-python-tox

self hosted runners. I have never set that up myself. It might be interesting to do this to have more control on the test env. Have you done that in one of your projects ?

Conan-Kudo:
No, I was thinking more of just switching to using the container image rather than what the runners use by default.

schaefi:
> No, I was thinking more of just switching to using the container image rather than what the runners use by default.

But you select the container image through the ```runs-on: name``` setting. And if that name is not a github hosted container you can setup a self hosted systemd and select from it. From my understanding this is the only way to select a different container as a runner instance for actions.

Anyway we are off topic, this one here is good to go imho. Can you give it a final look.

Thanks in advance

Conan-Kudo (approved):

PR #2226, opened 2023-01-02
Pass along tmpdir to skopeo

schaefi:
When using the --temp-dir parameter, it was not passed to the skopeo calls when building a container image with kiwi.

### DIFF
--- a/kiwi/oci_tools/base.py
+++ b/kiwi/oci_tools/base.py
@@ -15,9 +15,12 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+import os
 from datetime import datetime
 
 # project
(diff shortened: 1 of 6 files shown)


schaefi:
A report from the mailing list showed that the way we call skopeo does not take into account an alternative tmpdir setting. This commit should fix it. The problem, I can't find a ```--tmpdir``` option in my skopeo version (skopeo version 0.1.41). @davidcassany any thoughts about it ?

Thanks

schaefi:
skopeo in v0.2 has support for --tmpdir, needs to be checked in code

schaefi:
@davidcassany finally found the time to complete this pull request. Can you please double check if the version check and parameter setup looks correct. Thanks much

davidcassany (approved):
Looks good to me :+1:

PR #2227, opened 2023-01-05
Allow to specify fixed size for the root partition

### LINKED ISSUE #2203: RFE: add an option to set the size of the system partition
I'd like to be able to set the exact size of the root partition (via a parameter under `<type>` I guess). I've tried using `<oem-systemsize>` to achieve the same result, but it doesn't seem to have any effect (likely because I have `<oem-resize>` to false as we're using systemd native filesystem growth instead).

schaefi:
So far the last partition (typically root) in the partition table takes all the rest space of the partition table in the image file. At deployment/boot time users had several options to let that partition grow to a custom size. However, during build time of the image it was not possible to specify a specific fixed size for the root partition as we don't wanted to produce gaps of unpartitioned space in the image file. It has turned out that there is hardware available which requires a partition to be an exact multiple of some blocksize. As kiwi supports size constraints for all other partitions but not for root this commit now allows for it. The oem-systemsize element is now also taken into account at build time of the image if the oem-resize which would do that at deployment/boot time is switched off like the following example shows:

```xml
<oemconfig>
    <!-- set root partition to 2048MB -->
    <oem-systemsize>2048</oem-systemsize>
</oemconfig>
```

This Fixes #2203

### DIFF
--- a/doc/source/working_with_images/custom_partitions.rst
+++ b/doc/source/working_with_images/custom_partitions.rst
@@ -84,7 +84,7 @@ clone="number"
 Despite the customization options of the partition table shown above
 there are the following limitations:
 
-1. The root partition is always the last one
+1. By default the root partition is always the last one
 
    Disk imags build with {kiwi} are designed to be expandable.
    For this feature to work the partition containing the system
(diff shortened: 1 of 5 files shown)


davidcassany:
Nice :+1: This is an interesting corner case.

Conan-Kudo (approved):

PR #2238, opened 2023-01-18
Use rsync defaults to sync the initrd root-tree

davidcassany:
This commits makes use of rsync default options to sync the root-tree of the boot image for custom initrds.

Fixes bsc#1207128 where it was noted hardlinks were not preserved inside the initrd.

### DIFF
--- a/kiwi/boot/image/builtin_kiwi.py
+++ b/kiwi/boot/image/builtin_kiwi.py
@@ -162,7 +162,7 @@ class BootImageKiwi(BootImageBase):
                 temp_boot_root_directory
             )
             data.sync_data(
-                options=['-a']
+                options=Defaults.get_sync_options()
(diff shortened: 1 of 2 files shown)


davidcassany:
@schaefi another custom boot image issue, apparently in former KIWI hardlinks were preserved in initrds, however they are not preserved in KIWI-NG. I do not see any reason to not to use the same rsync default options in custom boot images (which already include `--hard-links` flag) as we do in sync_data in filesystems or in volume managers.

schaefi (approved):
Looks good to me :+1: 

I also can't come up with a good reason why we not use the default sync options for the kiwi builtin copy

schaefi:
@davidcassany I'm merging because the broken tests are unrelated and fixed in master

PR #2248, opened 2023-02-22
Support new bootloadersettings section

### LINKED ISSUE #2224: tpm disk unlock
## Problem description

There is a set of upstream patches that enables grub to decrypt a disc based on secrets obtained from tpm. Since the patches have not been merged into upstream grub SUSE has merged the patches into the SUSE grub package and created a module to facilitate the functionality. Therefore this is at present SUSE specific. The code is activated by

`grub2-install --suse-enable-tpm`

or

`shim-install --suse-enable-tpm`

in addition 

`TRUSTED_BOOT="yes"`

needs to be set in `/etc/sysconfig/bootloader`

## Expected behaviour

The desire is to trigger kiwi to add the additional argument to `grub2-install` and set the variable in the config.

## Proposal

Introduce `suse-tpm-unlock` attribute to the `bootloader` element as a bool `true|false` and let this configuration in the XML trigger the process of adding the additional argument to grub2-install.

I realize this is very SUSE specific and it would be much nicer to have a non distribution specific implementation. Given the path chosen for grub I am not certain this is possible.

@schaefi thoughts?

When we agree on a direction I can deliver the implementation.

Thanks,
Robert

schaefi:
Allow to specify an optional <bootloadersettings> element inside of the <bootloader> section. The information is used to specify custom arguments for the tools called in a bootloader setup procedure, e.g shim-install, grub-install or grub-mkconfig. kiwi does not judge on the given parameters and if the provided data is effectively used depends on the individual bootloader implementation. In this commit the data will be effectively handled if bootloader="grub2" is configured. More precisely the custom additions to support SUSE's TPM model with grub2 can be configured as follows:

```xml
<bootloadersettings>
    <shimoption name="--suse-enable-tpm"/>
    <installoption name="--suse-enable-tpm"/>
    <configoption name="--set-trusted-boot"/>
</bootloadersettings>
```

This Fixes #2224

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -1083,6 +1083,27 @@ For details see: :ref:`custom_volumes`
    or mount points. As a result, whatever is written in `<partitions>`
    cannot be expressed in the same way in `<volumes>`.
 
+<preferences><type><bootloader><bootloadersettings>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Used to specify custom arguments for the tools called to setup
+secure boot e.g `shiminstall`, installation of the bootloader
+e.g `grub-install` or configuration of the bootloader e.g `grub-mkconfig`.
+
+.. code:: xml
+
+   <bootloadersettings>
+       <shimoption name="--suse-enable-tpm"/>
+       <shimoption name="--bootloader-id" value="some-id"/>
+       <installoption name="--suse-enable-tpm"/>
+       <configoption name="--debug"/>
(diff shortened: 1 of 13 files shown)


davidcassany (approved):
the implementation looks clean and simple to me :+1: 
A not intrusive patch, nice.

I haven't tested it myself

rjschwei:
Building a test image

rjschwei (approved):

rjschwei:
Well, kiwi does what is expected:

....grub2-install --suse-enable-tpm -....

....shim-install --removable --suse-enable-tpm ....

AS such this is good to merge and release from my perspective.

PR #2254, opened 2023-03-09
Fix permissions of live ISO system files

### LINKED ISSUE #2246: Keep all files world-readable on Live ISOs
## Problem description

Initrd and Live image are only root-readable. Not a big issue, just a tad annoying when dealing with Live ISOs as non-root user.
I guess these file were just created that way.

```sh
# ls -l LiveOS/squashfs.img boot/x86_64/loader/initrd 
-rw------- 1 root root 879591424 20. Feb 15:59 LiveOS/squashfs.img
-rw------- 1 root root  93005187 20. Feb 15:58 boot/x86_64/loader/initrd
```

## Expected behaviour

Make all files readable by all users (`chmod a+u`).

## Bonus argument

Fedora gets it right. :-)

schaefi:
Make sure initrd and squashfs.img takes permissions o644 This Fixes #2246

### DIFF
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -278,6 +278,7 @@ class LiveImageBuilder:
             container_image.name
         )
         Path.create(self.media_dir.name + '/LiveOS')
+        os.chmod(container_image.name, 0o644)
         shutil.copy(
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
Looks good to me

Conan-Kudo (approved):

PR #2256, opened 2023-03-10
Use module anymarkup-core directly instead of anymarkup

jgleissner:
Module `anymarkup` is just a wrapper for `anymarkup-core` and was recently dropped from openSUSE.

### DIFF
--- a/.virtualenv.requirements.txt
+++ b/.virtualenv.requirements.txt
@@ -18,7 +18,8 @@ xattr==0.9.3
 requests
 
 # markup support
-anymarkup
+anymarkup-core
+xmltodict
 
 # json support that also knows about tuples
 simplejson
--- a/kiwi/markup/any.py
+++ b/kiwi/markup/any.py
@@ -17,6 +17,7 @@
 #
 from typing import Any
 import importlib
+import importlib.util
 
 # project
 from kiwi.utils.temporary import Temporary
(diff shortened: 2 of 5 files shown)


schaefi:
Hi Joe, thanks the change looks good to me. I haven't looked into side effects of this change regarding older suse or other distributions e.g Fedora. Just to put me right, the change of the Recommends in the spec file would also be working backward compatible, right ?

Thanks

jgleissner:
> Hi Joe, thanks the change looks good to me. I haven't looked into side effects of this change regarding older suse or other distributions e.g Fedora. Just to put me right, the change of the Recommends in the spec file would also be working backward compatible, right ?

anymarkup has been a wrapper for anymarkup-core since version 0.5.0 (released Oct 2015), so anything half-way recent should have anymarkup-core, since that's the actual implementation. AFAICS RHEL/Centos 7 has anymarkup 0.4.3, but since the Recommends for anymarkup is disabled for RHEL 7 in the spec file I assumed that version isn't supported by kiwi anyway.

Conan-Kudo:
Yes, please. 👍🏾

schaefi:
ignore the failed test for the moment, there was a timeout on testing the shell code in the containers. It happens every now and then. I restart the tests which should fix it

Conan-Kudo (approved):
Neat!

PR #2257, opened 2023-03-14
Fix Arch Linux EFI install ISO integration test

### LINKED ISSUE #2232: archlinux kiwi-test-image-live-disk-kis install iso fails to boot in UEFI mode
## Problem description

The Archlinux [test-image-live-disk-kis](https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:archlinux/test-image-live-disk-kis) install iso fails to boot in UEFI mode with the following error:
image

## Expected behavior

It should boot just like it does with BIOS

- Download https://download.opensuse.org/repositories/Virtualization:/Appliances:/Images:/Testing_x86:/archlinux/images/iso/kiwi-test-image-live-disk-kis.x86_64-1.0.0-Disk-Build54.11.install.iso
- boot it in UEFI mode
- wait for it to fail to boot

schaefi:
kiwi has code in bootloader/template/grub2.py which uses linuxefi/initrdefi commands according to the configured grub_platform. These commands does not exist on Arch and therefore the integration test provides its own ISO template. This Fixes #2232

### DIFF
--- a/build-tests/x86/archlinux/test-image-live-disk-kis/appliance.kiwi
+++ b/build-tests/x86/archlinux/test-image-live-disk-kis/appliance.kiwi
@@ -46,12 +46,11 @@
     <preferences profiles="Disk">
         <!--
              KIWI does not natively provide EFI support for Arch Linux,
-             because of that the editbootinstall script is used to hot pacth
-             grub.cfg file. This hotpatch only applies for disk images
-             therefore the installation ISO does not support EFI
+             because of that the editbootinstall script and iso_boot.template
+             is used to hot patch grub.cfg file.
         -->
         <type image="oem" filesystem="ext4" installiso="true" installboot="install" kernelcmdline="console=ttyS0" editbootinstall="editbootinstall_arch.sh" firmware="efi">
-            <bootloader name="grub2" console="serial" timeout="10"/>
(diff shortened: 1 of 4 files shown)


schaefi:
Building the integration test from here with that branch

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:archlinux/test-image-live-disk-kis

schaefi:
> Interesting, I was under the impression we were always applying to patch to transform `linux` into `linuxefi` for `efi` firmware. smiley

you still could be right on this but for grub booting an ISO we use our own templates and if that template does not even contain the fixing characters the fix_ method becomes a noop. I'm still testing though

schaefi:
The part from our (kiwi) default template that bites us is imho this one

```
set linux=linux
set initrd=initrd
if [ "$${grub_cpu}" = "x86_64" -o "$${grub_cpu}" = "i386" ]; then
    if [ "$${grub_platform}" = "efi" ]; then
        set linux=linuxefi
        set initrd=initrdefi
    fi
fi
export linux initrd
```

And I just droped that if condition for the Arch builds with a custom version of a template.
The fix method should now also be a noop

but again ... I'm still testing

davidcassany:
Awesome! @schaefi you rock man 😉

davidcassany (approved):
I am loving it, thanks Marcus. This looks awesome.

PR #2258, opened 2023-03-15
Add dracut-kiwi-verity dracut module package

schaefi:
When building an image with an embedded verity hash via the embed_verity_metadata="true" type attribute it is required to setup the device map at boot time accordingly. The dracut module in this commit provides the needed tooling as a rust implementation which is able to read the kiwi embedded verithash from the offset of the partition. Kudos goes to Isaac True <isaac.true@canonical.com>

### DIFF
--- a/dracut/modules.d/80kiwi-verity/Makefile
+++ b/dracut/modules.d/80kiwi-verity/Makefile
@@ -1,0 +1,23 @@
+buildroot ?= /
+
+CROSS_COMPILE ?=
+CC ?= $(CROSS_COMPILE)gcc
+CFLAGS += -Werror -Wall -Wextra -std=c17
+
+BINARY := kiwi-parse-verity
+
+.PHONY: all
+all: build
+
+.PHONY: install
+install: build
+	install -Dm0755 kiwi-verity-setup.sh module-setup.sh \
+		-t ${buildroot}usr/lib/dracut/modules.d/80kiwi-verity
+	install -Dm0755 $(BINARY) ${buildroot}usr/bin/$(BINARY)
+
+.PHONY: build
+build: $(BINARY)
+
+.PHONY: clean
+clean:
+	rm -f $(BINARY)
--- a/dracut/modules.d/80kiwi-verity/README.rst
+++ b/dracut/modules.d/80kiwi-verity/README.rst
@@ -1,0 +1,67 @@
+kiwi-parse-verity
+=================
+
+Overview
+--------
+
+This tool is used to parse the information stored by
+`Kiwi <https://osinside.github.io/kiwi/>`__ inside the unused space of a
+disk partition when using a SquashFS root filesystem backed by
+``dm-verity``. This data is stored at a known offset (generally 4096
+bytes, or 1 block, from the end of the partition) and follows the layout
+as described in the `Kiwi
+documentation <https://osinside.github.io/kiwi/image_description/elements.html#preferences-type>`__
+for the attribute ``embed_verity_metadata``.
+
+Usage
+-----
+
+The following parameters are required in the Kiwi appliance description:
+
+-  ``verity_blocks="all"``
+-  ``embed_verity_metadata="true"``
+
+The tool is used to pull a specific entry from the metadata and will
+return that entry as plain text. It can be called using the following
+syntax:
+
+::
+
+   kiwi-parse-verity -p <path to partition/device> -o <data type>
+
+The data type is one of the entries stored by Kiwi inside the metadata,
+such as ``root-hash`` or ``algorithm``. More information about the
+parameters can be found by calling ``kiwi-parse-verity --help``.
+
+You will probably need all of the metadata to successfully map the
+device, as Kiwi sets up the ``dm-verity`` device with the
+``--no-superblock`` option, which does not store any data inside the
+partition, meaning all parameters need to be passed to it on the command
+line.
+
+Example
+~~~~~~~
+
+Detailed example of pulling all of the metadata from the partition and
+mapping the device using ``veritysetup``:
+
+::
+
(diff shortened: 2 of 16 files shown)


schaefi:
@glaubitz Hi Adrian, I added you as a reviewer here mostly to make you aware once this is merged there will be a new dracut module package which needs adaptions on the Debian packaging side of things as well. As you take care for the official debian builds it would be great if you can adapt the once this is merged

schaefi:
@IsaacJT I opened the PR but the show is yours :)

dracut/modules.d/80kiwi-verity/kiwi-verity-setup.sh
```diff
@@ -0,0 +1,52 @@
...
+# kiwi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kiwi.  If not, see <http://www.gnu.org/licenses/>
+#
+
+. /lib/dracut-lib.sh
```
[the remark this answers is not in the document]
schaefi: The include path only exists inside of the initrd env, shellcheck needs to ignore this. I fixed that

Conan-Kudo:
This adds a Rust dependency? 😒

dracut/modules.d/80kiwi-verity/README.rst
```diff
@@ -0,0 +1,75 @@
...
+documentation <https://osinside.github.io/kiwi/image_description/elements.html#preferences-type>`__
+for the attribute ``embed_verity_metadata``.
+
+Building
+--------
+
+Rust and Cargo are required to build the tool. Simply call
+``cargo build`` and the tool and its dependencies will be built.
+
+Alternatively, a .deb can be built using the ``debuild`` command.
```
Conan-Kudo, line 22: `debbuild`

Conan-Kudo:
Adding this Rust module _completely_ breaks RHEL 7 and SLE 12 support. We don't have a new enough Rust stack for either of those.

Conan-Kudo (changes requested):

dracut/modules.d/80kiwi-verity/README.rst
IsaacJT: This README was taken from a standalone Debian package that I had created, so `debuild` was correct here. It's however not really relevant anymore, so I'll just remove that line.

IsaacJT:
> Adding this Rust module _completely_ breaks RHEL 7 and SLE 12 support. We don't have a new enough Rust stack for either of those.

I'm not familiar enough with the Rust support for either of those distros, so I can't really comment here. @schaefi?

IsaacJT:
The `kiwi-parse-verity` tool is straightforward enough that it could be relatively easily rewritten in C. I had originally written the tool to be standalone and so it made sense to do it in Rust, but if introducing Rust tools into the Kiwi package breaks compatibility with supported platforms, then I can redo it in C.

Conan-Kudo:
@IsaacJT That would be great if it could be written in C. We don't really have Rust experience either. 😅

IsaacJT:
Hi @schaefi and @Conan-Kudo I've rewritten `kiwi-parse-verity` in C, squashed the commits, and updated the Makefile to build as a static C binary with GCC rather than Rust/Cargo.

dracut/modules.d/80kiwi-verity/Makefile
```diff
@@ -0,0 +1,26 @@
+buildroot ?= /
+
+CROSS_COMPILE ?=
+CC ?= $(CROSS_COMPILE)gcc
+STRIP ?= $(CROSS_COMPILE)strip
```
Conan-Kudo, line 5: We should probably not have this here, since stripping is controlled by the build environment.

dracut/modules.d/80kiwi-verity/Makefile
```diff
@@ -0,0 +1,26 @@
...
+install: build
+	install -Dm0755 kiwi-verity-setup.sh module-setup.sh \
+		-t ${buildroot}usr/lib/dracut/modules.d/80kiwi-verity
+	install -Dm0755 $(BINARY) ${buildroot}usr/bin/$(BINARY)
+
+.PHONY: build
+build: $(BINARY)
+
+$(BINARY): kiwi-parse-verity.c Makefile
+	$(CC) -Werror -Wall -Wextra -std=c17 $(CFLAGS) -static -o $(@) $(<)
```
Conan-Kudo, line 20: Why statically linked? Most initramfs setups these days have libc and other things...

Conan-Kudo (changes requested):

dracut/modules.d/80kiwi-verity/Makefile
IsaacJT: I was thinking of what Marcus said with statically compiled tools reducing the dependencies inside the initrd when I did this, but yeah, it doesn't make much sense here when it just uses libc... I'll change that

IsaacJT:
Pushed a new revision removing the call to strip, removing the static compilation flag, and removing upx (as the binary is now very small)

schaefi:
Looks good, just the minor issues reported from codacy here:

* https://app.codacy.com/gh/OSInside/kiwi/pullRequest?prid=11350905#newIssuesView

report makes sense imho

IsaacJT:
Codacy is very picky about the strncmp usage :D it should be impossible to over-read in this case

Conan-Kudo (approved):

Conan-Kudo:
Looks great, thanks for your contribution! Merged.

PR #2259, opened 2023-03-15
Add selinux integration test for Leap

### LINKED ISSUE #2244: setfiles command fails
https://github.com/OSInside/kiwi/blob/4d6b8412431e6b0d27321921bd1cb7caa5ec847f/kiwi/system/setup.py#L542

The above line has an extra "/" at the end due to which setfiles command fails during kiwi build:

[2023-02-14T20:35:53.786Z] [ DEBUG   ]: 20:35:53 | EXEC: [chroot image-root setfiles -F -p -c /etc/selinux/targeted/policy/policy.32 -e /proc -e /sys -e /dev /etc/selinux/targeted/contexts/files/file_contexts /]
[2023-02-14T20:35:54.049Z] [ DEBUG   ]: 20:35:53 | EXEC: Failed with stderr: usage:  setfiles [-diIDlmnpqvFW] [-e excludedir] [-r alt_root_path] spec_file pathname...
[2023-02-14T20:35:54.049Z] usage:  setfiles [-diIDlmnpqvFW] [-e excludedir] [-r alt_root_path] spec_file -f filename
[2023-02-14T20:35:54.049Z] usage:  setfiles -s [-diIDlmnpqvFW] spec_file
[2023-02-14T20:35:54.049Z] usage:  setfiles -c policyfile spec_file
[2023-02-14T20:35:54.049Z] , stdout: (no output on stdout)
[2023-02-14T20:35:54.049Z] [ ERROR   ]: 20:35:53 | KiwiCommandError: chroot: stderr: usage:  setfiles [-diIDlmnpqvFW] [-e excludedir] [-r alt_root_path] spec_file pathname...

I tried to run on a native SUSE OS it fails as below:

[root@tmp]# setfiles -F -p -c /etc/selinux/targeted/policy/policy.31 -e /proc -e /sys -e /dev /etc/selinux/targeted/contexts/files/file_contexts /
usage:  setfiles [-dilnpqvFW] [-e excludedir] [-r alt_root_path] spec_file pathname...
usage:  setfiles [-dilnpqvFW] [-e excludedir] [-r alt_root_path] spec_file -f filename
usage:  setfiles -s [-dilnpqvFW] spec_file
usage:  setfiles -c policyfile spec_file

OS: openSUSE 15.3
Kiwi version: 9.24.43

schaefi:
SUSE systems supports both, apparmor and selinux, whereas apparmor is the default. As selinux requires to create security labels during build time of an image, this integration test checks if the labeling works for SUSE based OS'es. This Fixes #2244

### DIFF
--- a/build-tests/x86/leap/test-image-disk-selinux/appliance.kiwi
+++ b/build-tests/x86/leap/test-image-disk-selinux/appliance.kiwi
@@ -1,0 +1,75 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<image schemaversion="7.5" name="kiwi-test-image-disk-selinux" displayname="Selinux">
+    <description type="system">
+        <author>Marcus Schäfer</author>
+        <contact>marcus.schaefer@gmail.com</contact>
+        <specification>Simple selinux disk, devel policy</specification>
+    </description>
+    <preferences>
+        <version>1.15.4</version>
+        <packagemanager>zypper</packagemanager>
+        <locale>en_US</locale>
+        <keytable>us</keytable>
+        <timezone>Europe/Berlin</timezone>
+        <rpm-excludedocs>true</rpm-excludedocs>
+        <rpm-check-signatures>false</rpm-check-signatures>
+        <bootsplash-theme>bgrt</bootsplash-theme>
+        <bootloader-theme>openSUSE</bootloader-theme>
+        <type image="oem" filesystem="xfs" kernelcmdline="console=ttyS0 security=selinux selinux=1 enforcing=1" firmware="uefi" format="qcow2">
+            <oemconfig>
+                <oem-resize>false</oem-resize>
+            </oemconfig>
+            <bootloader name="grub2" console="serial" timeout="10"/>
+        </type>
+    </preferences>
+    <users>
+        <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root" groups="root"/>
+    </users>
+    <repository type="rpm-md">
+        <source path="obsrepositories:/"/>
+    </repository>
+    <packages type="image">
+        <package name="patterns-base-minimal_base"/>
+        <package name="restorecond"/>
+        <package name="policycoreutils"/>
+        <package name="setools-console"/>
+        <package name="selinux-policy-targeted"/>
+        <package name="selinux-policy-devel"/>
+        <package name="selinux-autorelabel"/>
+        <package name="bind-utils"/>
+        <package name="systemd"/>
+        <package name="plymouth-theme-bgrt"/>
+        <package name="grub2-branding-openSUSE"/>
+        <package name="iputils"/>
+        <package name="vim"/>
+        <package name="grub2"/>
+        <package name="grub2-x86_64-efi" arch="x86_64"/>
+        <package name="grub2-i386-pc"/>
+        <package name="syslinux"/>
+        <package name="lvm2"/>
+        <package name="plymouth"/>
+        <package name="fontconfig"/>
+        <package name="fonts-config"/>
+        <package name="tar"/>
+        <package name="parted"/>
+        <package name="openssh"/>
+        <package name="iproute2"/>
+        <package name="less"/>
+        <package name="bash-completion"/>
+        <package name="dhcp-client"/>
+        <package name="which"/>
+        <package name="kernel-default"/>
+        <package name="shim"/>
+        <package name="timezone"/>
+    </packages>
+    <packages type="bootstrap">
+        <package name="udev"/>
+        <package name="filesystem"/>
+        <package name="glibc-locale"/>
+        <package name="cracklib-dict-full"/>
+        <package name="ca-certificates"/>
+        <package name="ca-certificates-mozilla"/>
+        <package name="openSUSE-release"/>
+    </packages>
+</image>
--- a/build-tests/x86/leap/test-image-disk-selinux/config.sh
+++ b/build-tests/x86/leap/test-image-disk-selinux/config.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+# shellcheck disable=SC1091
+test -f /.kconfig && . /.kconfig
+
+set -ex
+
+#======================================
+# Setup baseproduct link
+#--------------------------------------
+suseSetupProduct
+
+#======================================
+# Delete auto relabel trigger
+#--------------------------------------
+rm -f /.autorelabel
+
+#======================================
+# Activate services
+#--------------------------------------
+suseInsertService sshd
+
+#======================================
+# Setup default target, multi-user
+#--------------------------------------
+baseSetRunlevel 3


schaefi:
Please find the integration test building here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:leap/test-image-disk-selinux

The behavior of the boot of this image is that on first boot it tells the user that a relabeling is required, it does it and reboots. After that sestatus says

```
Have a lot of fun...
localhost:~ # sestatus -v
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   permissive
Mode from config file:          permissive
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Memory protection checking:     requested (insecure)
Max kernel policy version:      33

Process contexts:
Current context:                unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023
Init context:                   system_u:system_r:init_t:s0
/sbin/agetty                    system_u:system_r:getty_t:s0-s0:c0.c1023
/usr/sbin/sshd                  system_u:system_r:sshd_t:s0-s0:c0.c1023

File contexts:
Controlling terminal:           unconfined_u:object_r:user_tty_device_t:s0
/etc/passwd                     system_u:object_r:passwd_file_t:s0
/etc/shadow                     system_u:object_r:shadow_t:s0
/bin/bash                       system_u:object_r:bin_t:s0 -> system_u:object_r:shell_exec_t:s0
/bin/login                      system_u:object_r:login_exec_t:s0
/bin/sh                         system_u:object_r:bin_t:s0 -> system_u:object_r:shell_exec_t:s0
/sbin/agetty                    system_u:object_r:bin_t:s0 -> system_u:object_r:getty_exec_t:s0
/sbin/init                      system_u:object_r:bin_t:s0 -> system_u:object_r:init_exec_t:s0
/usr/sbin/sshd                  system_u:object_r:sshd_exec_t:s0
```

Question is if the setfiles during build time of the image did a correct job or not. To my understanding a relabeling on first boot should not be needed. But I don't know enough about the inner works of selinux.

Thus let's now use this integration test and the results to have a conversation if anything needs to change in kiwi

Thanks much

build-tests/x86/leap/test-image-disk-selinux/config.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+test -f /.kconfig && . /.kconfig
```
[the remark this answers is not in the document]
schaefi: fixed

build-tests/x86/leap/test-image-disk-selinux/appliance.kiwi
```diff
@@ -0,0 +1,75 @@
...
+        <version>1.15.4</version>
+        <packagemanager>zypper</packagemanager>
+        <locale>en_US</locale>
+        <keytable>us</keytable>
+        <timezone>Europe/Berlin</timezone>
+        <rpm-excludedocs>true</rpm-excludedocs>
+        <rpm-check-signatures>false</rpm-check-signatures>
+        <bootsplash-theme>bgrt</bootsplash-theme>
+        <bootloader-theme>openSUSE</bootloader-theme>
+        <type image="oem" filesystem="xfs" kernelcmdline="console=ttyS0 security=selinux selinux=1 enforcing=0" firmware="uefi" format="qcow2">
```
Conan-Kudo, line 19: Is this any good if we're not testing enforcing mode?

Conan-Kudo (changes requested):

build-tests/x86/leap/test-image-disk-selinux/appliance.kiwi
schaefi: good point, I change it

schaefi: I blindly took over the recommendations from the openSUSE documentation

Conan-Kudo (approved):

Conan-Kudo:
> Question is if the setfiles during build time of the image did a correct job or not. To my understanding a relabeling on first boot should not be needed. But I don't know enough about the inner works of selinux.

You are right that this is wrong. It should not require relabeling on first boot.

Conan-Kudo:
Wait, does Leap even have an SELinux policy package?

schaefi:
> Wait, does Leap even have an SELinux policy package?

yes in the integration test I install the following

```
<package name="selinux-policy-targeted"/>
<package name="selinux-policy-devel"/>
```

and the ```targeted``` policy is used by default unless specified with ```selinux_policy="..."``` in the kiwi config

schaefi:
> > Question is if the setfiles during build time of the image did a correct job or not. To my understanding a relabeling on first boot should not be needed. But I don't know enough about the inner works of selinux.
> 
> You are right that this is wrong. It should not require relabeling on first boot.

Is there a way for me to find out the reason for the relabeling ? I did not dig deeper but I guess this is some systemd service triggerd by a reason

Conan-Kudo:
> selinux-policy-targeted

I'm surprised because they definitely don't publish such a package to the repos: https://code.opensuse.org/leap/features/issue/36

Conan-Kudo:
> > > Question is if the setfiles during build time of the image did a correct job or not. To my understanding a relabeling on first boot should not be needed. But I don't know enough about the inner works of selinux.
> > 
> > 
> > You are right that this is wrong. It should not require relabeling on first boot.
> 
> @Conan-Kudo Is there a way for me to find out the reason for the relabeling ? I did not dig deeper but I guess this is some systemd service triggerd by a reason

@schaefi Hmm, it seems that the selinux-policy package sets the `/.autorelabel` file on initial install itself: https://code.opensuse.org/package/selinux-policy/blob/5c43856f5e0a793f562f955973e84f0c1bdb09c1/f/selinux-policy.spec#_248-275

schaefi:
> > selinux-policy-targeted
> 
> I'm surprised because they definitely don't publish such a package to the repos: https://code.opensuse.org/leap/features/issue/36

yes I needed to add:

```
<path project="security:SELinux" repository="15.4"/>
```

To build the integration test. It then resolves to:

```
<bdep name="selinux-policy-targeted" noinstall="1" version="20230321" release="150400.188.2" arch="noarch" hdrmd5="2022d376034d3262b4ea936e57212062" project="security:SELinux" repository="15.4"/>
```

schaefi:
@Conan-Kudo 

> Hmm, it seems that the selinux-policy package sets the /.autorelabel file on initial install itself:

ah that's the trigger, thanks for pointing me to it. Well in this case it seems to be intentional ?
I can simply remove the file in the integration test, but it seems not a correct solution.

All this is why I actually wanted the responsible people at SUSE to take a look and start a conversation what parts
falls into kiwi's responsibility and what other parts might need a change elsewhere

@dirkmueller with the current findings could you provide some insights on

- policy packages not published ?
- relabeling activated intentionally at install time of the policy bug or feature ?

Thanks

Conan-Kudo:
> @Conan-Kudo
> 
> > Hmm, it seems that the selinux-policy package sets the /.autorelabel file on initial install itself:
> 
> ah that's the trigger, thanks for pointing me to it. Well in this case it seems to be intentional ? I can simply remove the file in the integration test, but it seems not a correct solution.
> 

For our test, we should remove it, because otherwise we're not correctly testing _our_ functionality.

schaefi:
> For our test, we should remove it, because otherwise we're not correctly testing our functionality.

agreed, done

schaefi:
ok with this change the relabeling on first boot does not happen. After boot ```sestatus``` says

```
localhost:~ # sestatus
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   enforcing
Mode from config file:          permissive
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Memory protection checking:     requested (insecure)
Max kernel policy version:      33
```

From my perspective this looks positive.

@Conan-Kudo @dirkmueller Thoughts ?

Thanks

Conan-Kudo:
If it boots up and works, then yeah, :shipit:

PR #2260, opened 2023-03-16
Fixed build of dracut-kiwi-verity

schaefi:
The spec file was missing the package definition to actually create the dracut-kiwi-verity package. In addition the fixed setting to the C standard 17 caused build errors for me on distributions that uses C standard 11/12.

### DIFF
--- a/dracut/modules.d/80kiwi-verity/Makefile
+++ b/dracut/modules.d/80kiwi-verity/Makefile
@@ -2,7 +2,7 @@ buildroot ?= /
 
 CROSS_COMPILE ?=
 CC ?= $(CROSS_COMPILE)gcc
-CFLAGS += -Werror -Wall -Wextra -std=c17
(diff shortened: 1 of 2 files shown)


schaefi:
@IsaacJT looking at the C code I could not see that C17 is a requirement. As I was not able to build for all targets with this setting:

```
cc: error: unrecognized command line option '-std=c17'; did you mean '-std=c11'?
```

I decided to delete this option. Please advise if this is ok

The package build provides the following data

```
/usr/bin/kiwi-parse-verity
/usr/lib/dracut/modules.d/80kiwi-verity
/usr/lib/dracut/modules.d/80kiwi-verity/kiwi-verity-setup.sh
/usr/lib/dracut/modules.d/80kiwi-verity/module-setup.sh
```

This looks good and complete, but also please double check on the list above.

Thanks

PR #2263, opened 2023-03-22
Make a proper use of `btrfs_relative_path` grub2 variable

davidcassany:
This commit gates the inclusion of `btrfs_relative_path` variable in early boot grub2 configuration based on the `btrfs_root_is_snapshot` attribute. This SUSE specific grub variable is added by grub2-mkconfig according to the variable `SUSE_BTRFS_SNAPSHOT_BOOTING` in `/etc/default/grub`.

This commit alignes early boot grub2 setup use of `btrfs_relative_path` with the flag `SUSE_BTRFS_SNAPSHOT_BOOTING` included in `/etc/default/grub`.

Fixes bsc#1209165

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -1018,9 +1018,10 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
 
     def _create_early_boot_script_for_uuid_search(self, filename, uuid):
         with open(filename, 'w') as early_boot:
-            early_boot.write(
-                'set btrfs_relative_path="yes"{0}'.format(os.linesep)
-            )
+            if self.xml_state.build_type.get_btrfs_root_is_snapshot():
+                early_boot.write(
+                    'set btrfs_relative_path="yes"{0}'.format(os.linesep)
+                )
             if self.custom_args.get('boot_is_crypto'):
                 early_boot.write(
                     'insmod cryptodisk{0}'.format(os.linesep)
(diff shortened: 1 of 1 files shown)


davidcassany:
To test it I build a couple of Leap 15.4 JeOS  images, one including the `btrfs_root_is_snapshot` attribute set to `true` and another one having it to `false`. Both images were bootable and I verified that  `btrfs_relative_path=yes` grub2 variable was set in both config files (`/boot/grub2/grub.cfg` and `/boot/efi/EFI/BOOT/grub.cfg`) when using snapshots for root. On the contrary non of the grub configuration files was including `btrfs_relative_path=yes` when root is not set to be an snapshot.

Conan-Kudo:
The SUSE Btrfs patches are part of the Red Hat/Fedora grub tree too, so be careful assuming it's "transparent".

davidcassany:
> The SUSE Btrfs patches are part of the Red Hat/Fedora grub tree too, so be careful assuming it's "transparent".

Indeed this is one of my fears... but then how the `SUSE_BTRFS_SNAPSHOT_BOOTING` translates to Fedora? On SUSE this causes  `btrfs_relative_path=yes` to be included in grub config. If I got it right this setting essentially causes to inclusion of `/@` prefix in path to find kernel and initrd, assuming snapshots are stored there and that given paths are relative to that.

Conan-Kudo:
We use exactly the same options: we didn't modify them so that snapper would remain compatible.

PR #2266, opened 2023-03-25
Support EFI boot from squashfs

schaefi:
grub supports reading from ```squashfs```. If no extra bootloader installation is required this allows to boot from a squashfs rootfs without extra boot partition. 

The prebuilt grub images usually does not contain the ```squash4``` module which is the reason why kiwi creates the EFI module itself if the system is fully immutable.

Please note: The grub configuration must be provided manually as part of the image description in this case because ```grub-mkconfig``` at the time of its invocation has no write permissions anymore.

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -626,6 +626,7 @@ class Defaults:
             'all_video',
             'xfs',
             'btrfs',
+            'squash4',
             'lvm',
             'luks',
             'gcry_rijndael',
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -276,10 +276,11 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         # More details can be found in the individual methods.
         # One fine day the following fix methods can hopefully
         # be deleted...
-        self._fix_grub_to_support_dynamic_efi_and_bios_boot(config_file)
-        self._fix_grub_root_device_reference(config_file, boot_options)
-        self._fix_grub_loader_entries_boot_cmdline()
-        self._fix_grub_loader_entries_linux_and_initrd_paths()
+        if self.xml_state.build_type.get_overlayroot_write_partition() is not False:
+            self._fix_grub_to_support_dynamic_efi_and_bios_boot(config_file)
+            self._fix_grub_root_device_reference(config_file, boot_options)
+            self._fix_grub_loader_entries_boot_cmdline()
+            self._fix_grub_loader_entries_linux_and_initrd_paths()
 
         if self.firmware.efi_mode() and self.early_boot_script_efi:
             self._copy_grub_config_to_efi_path(
@@ -828,7 +829,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         if not lookup_path:
             lookup_path = self.boot_dir
         grub_image = Defaults.get_unsigned_grub_loader(lookup_path)
-        if grub_image:
+        if grub_image and self.xml_state.build_type.get_overlayroot_write_partition() is not False:
             log.info('--> Using prebuilt unsigned efi image')
             Command.run(
                 ['cp', grub_image, self._get_efi_image_name()]
@@ -1063,8 +1064,8 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 'set btrfs_relative_path="yes"{0}'.format(os.linesep)
             )
             early_boot.write(
-                'search --file --set=root /boot/{0}{1}'.format(
-                    mbrid.get_id(), os.linesep
+                'search --file --set=root /boot/mbrid{0}'.format(
+                    os.linesep
                 )
             )
             early_boot.write(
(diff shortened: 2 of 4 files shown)


kiwi/defaults.py
```diff
@@ -626,6 +626,7 @@ def get_grub_basic_modules(multiboot):
             'all_video',
             'xfs',
             'btrfs',
+            'squash4',
```
davidcassany, line 629: @schaefi I doubt it makes a relevant difference to KIWI in this particular case but note that since SLE 15 SP4 grub also includes loopback module and IIRC squash4 too (not 100% sure about squash).
On my team we asked for loopback module and squash that and loopback got accepted. However I do not fully remember about squash4, there was some back and forth as squash4 module also loads other compression modules (xzio and friends).
Long story short this change might not be required for SUSE and simply adding squash4 might be enough for compressed squashfs.

davidcassany (approved):
Looks good to me, I am just wondering if we shouldn't explicitly add some compression module or not.

Have you tested this change with a compressed squashfs?

schaefi:
> Have you tested this change with a compressed squashfs?

yes I tested it with the image design that I have in mind when testing delta containers for ALP. See here: https://build.opensuse.org/package/show/home:marcus.schaefer:delta_containers/test-os-automotive-rpi-small
fully immutable OS, small, easy to update and a container registry that allows for using applications like native programs through the oci-pilot launcher. You need a raspberry PI to test this image though, but looking at the appliance.kiwi and you should get the idea

schaefi:
> grub also includes loopback module and IIRC squash4 too (not 100% sure about squash).

at least leap15.4 doesn't, so I assumed SLE 15 will also not include it. At the moment I force kiwi into self mkimage if the attribute ```overlayroot_write_partition="false"``` is set. I like to change this in the future to have an explicit attribute that allows us to say "build_your_own_efi_image". But that's for another PR :)

davidcassany:
> > Have you tested this change with a compressed squashfs?
> 
> yes I tested it with the image design that I have in mind when testing delta containers for ALP. See here: https://build.opensuse.org/package/show/home:marcus.schaefer:delta_containers/test-os-automotive-rpi-small fully immutable OS, small, easy to update and a container registry that allows for using applications like native programs through the oci-pilot launcher. You need a raspberry PI to test this image though, but looking at the appliance.kiwi and you should get the idea

Not sure I got the whole idea right, do you imagine delta updates based on squashfs? If I got it right this is a read only OS image (A/B partitions for full OS upgrades), another read only image holding preloaded containers (a container storage) and finally another image or container storage with read-write capabilities, is that right? I struggle to see where delta updates apply, stacking deltas in OS image?

schaefi:
> Not sure I got the whole idea right, do you imagine delta updates based on squashfs? 

no. We are a bit off topic now as delta container support is not related to creating a fully squashfs based system with /boot also on squashfs. These are two different topics. I implemented delta container support in kiwi which allows you to create app containers as a delta from a base container. The app container becomes then small and of course you need a provision tool to bring the two together when you start a container instance from a delta. That part is implemented in a new project called oci-pilot. Some weeks ago I had opened a PR which implements the creation of such delta containers in kiwi and I asked for a review from your side but I guess you were busy or did not had time to look at it.

Again two completely different things. However when it comes to embedded those two features play well together. For embedded systems there is often the requirement to have a 100% immutable OS which even doesn't implement write(). That's also the reason why I did not use MicroOS even though I wanted. For that part you need the feature used here. But then you end up in a system that cannot do much, so how to run applications ? without having tons of duplicates with containers ? That's where app containers (deltas) comes into the game.

All this combined in the image I pointed you to. But I only pointed you there to see that I have tested the full squashfs system and it works. If you are interested in the other parts or concepts I'm happy to show or demo it on the next community meeting

Hope this clarifies it a bit and sorts out the confusion ?

Are we good to merge then ?

Thanks

davidcassany:
> > Not sure I got the whole idea right, do you imagine delta updates based on squashfs?
> 
> no. We are a bit off topic now as delta container support is not related to creating a fully squashfs based system with /boot also on squashfs.

Sure, my bad, I was just so curious.

> Some weeks ago I had opened a PR which implements the creation of such delta containers in kiwi and I asked for a review from your side but I guess you were busy or did not had time to look at it.

Damn it... my bad... I did not see it. I am really running out of time these days and I struggle to follow all the development.

> Again two completely different things. However when it comes to embedded those two features play well together. For embedded systems there is often the requirement to have a 100% immutable OS which even doesn't implement write(). 

Interesting, on the Elemental project I am currently related I resisted to drop squashfs support just for this reason, I am happy I managed to keep squashfs into the picture.

> All this combined in the image I pointed you to. But I only pointed you there to see that I have tested the full squashfs system and it works. If you are interested in the other parts or concepts I'm happy to show or demo it on the next community meeting

That would be nice indeed. I'd be happy to see all this new KIWI features in action targeting embedded world.

> Hope this clarifies it a bit and sorts out the confusion ?

Sure, really appreciated, thanks.

PR #2268, opened 2023-03-30
Bring back create_parted_partitions method

davidcassany:
This commit brings back the create_parted_partitions method as it was before commit 1c48f33. Within the mentioned commit the partitioner logic of the dracut modules dropped the use of `parted` in favor of `sgdisk` and `sfdisk` depending on the partition table type.

For s390 platforms it was used `fdasd`, however its seams that nowadays support for DASD is better for `parted` compared with `fdasd`. More specific, `fdasd` does not support resizing the partition table and expanding partitions, but parted does.

For s390 parted should delete and recreate the partition to be expanded, `resizepart` is not functional in that context.

According to SUSE s390 experts `parted` has superseded `fdasd` and it should be the preferred tool for s390 whenever possible.

Within the same issue it has also been noted that `blkid` does not properly detect dasd type, hence an heuristic approach has been implemented based on the device name.

Fixes bsc#1209247

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -17,7 +17,7 @@ function create_partitions {
             create_gpt_partitions "${disk_device}" "${partition_setup}"
         ;;
         "dasd")
-            create_dasd_partitions "${disk_device}" "${partition_setup}"
+            create_parted_partitions "${disk_device}" "${partition_setup}"
         ;;
     esac
     partprobe "${disk_device}"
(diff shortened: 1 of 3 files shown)


davidcassany:
mm not so sure about the Shellcheck complaint, I don't get it locally in my env and I doubt I can actually quote the offending variable expansion without altering the loop behavior.... Anyway, gonna leave it for a later discussion for the time being.

@schaefi I am not pleased with this PR, but according to a reported issue the partition expansion does not work on s390, and according to the experts fdasd is not capable for this job, but parted is... This PR basically imports the parted wrappers before they were actually removed in favor of `sgdisk` and `sfdisk`. I spent yesterday afternoon trying to recode the parted wrapper as I thought it could be done in a simpler way as it used to be, however I quickly ended up having issues to script the detection of the start point of a new partition, setting partition type vs flags and I realized parted is ugly wrap with bash. At that point I noted that I was somehow rewriting what was already there, hence I just brought it back as is.

I'd say we should first try to get some positive feedback to verify the solution and then discuss and iterate if there is something to improve, like shellcheck tests,  count sectors instead of cylinders or review partition type logic.

@Vogtinator I'd appreciate if you could give it a run on this PR. Let me know if you want a project in OBS including this code and some specific image, I can quickly arrange it.

davidcassany:
@Vogtinator I just noted  forgot to include back the parted dependency in dracut module and the related RPMs...  So the image will fail as parted will not be there. You should update the package with the current code, sorry.

Vogtinator:
> Maybe it's possible to use `parted` to expand the partition table and use `fdasd` afterwards?

It is! By doing `echo -e "resizepart 1\n\nq\n" | parted /dev/dasdb` (which resizes the first partition to the max size, effectively a noop), parted rewrites the partition table to span the whole device.
Just doing this before using `fdasd` should do the job.

davidcassany:
> It is! By doing `echo -e "resizepart 1\n\nq\n" | parted /dev/dasdb` (which resizes the first partition to the max size, effectively a noop), parted rewrites the partition table to span the whole device. Just doing this before using `fdasd` should do the job.

Mmm interesting, hence we could eventually find a workaround that is not dropping the current fdasd logic and just adding a dummy `parted` call to ensure the partition table matches disk geometry. Given this is only targeting s390 and that it should take few lines of code I'd say we can consider it. I'll update the PR with such information.

I's say there is not need to run parted interactive so I'd translate your command to something like `parted --script --machine /dev/dasdb resizepart 1`. I am wondering if we always have more than one partition for s390, as far as I understood if the partition is actually a candidate to be resized this will raise a parted bug. AFAIU this is why you run it for the first partition. @Vogtinator Did I got it right?

mfriesenegger:
I tested https://download.suse.de/ibs/home:/favogt:/bsc1209247/images/SLE-Micro.s390x-5.4.0-Default-dasd-Build7.6.raw.xz and had the same problem that I reported in https://bugzilla.suse.com/show_bug.cgi?id=1209247#c14.  

I booted into rescue mode to see what happened to the partitions.

The root partition is deleted by kiwi-partitions-lib.sh
```shell
0:ins005c8:~ # parted /dev/dasda print
Model: IBM S390 DASD drive (dasd)
Disk /dev/dasda: 32.2GB
Sector size (logical/physical): 512B/4096B
Partition Table: dasd
Disk Flags: 

Number  Start   End    Size   File system  Flags
 1      98.3kB  315MB  315MB  ext2
``` 

I manually retrieved get the last sector of partition 

```shell
0:ins005c8:~ # parted /dev/dasda  unit s print
Model: IBM S390 DASD drive (dasd)
Disk /dev/dasda: 62915040s
Sector size (logical/physical): 512B/4096B
Partition Table: dasd
Disk Flags: 

Number  Start  End      Size     File system  Flags
 1      192s   614591s  614400s  ext2
```
 
I manually recreate the last partition

```shell
0:ins005c8:~ # parted /dev/dasda unit s mkpart btrfs 614592s 100%
Information: You may need to update /etc/fstab.

0:ins005c8:~ # parted /dev/dasda  unit s print
Model: IBM S390 DASD drive (dasd)
Disk /dev/dasda: 62915040s
Sector size (logical/physical): 512B/4096B
Partition Table: dasd
Disk Flags: 

Number  Start    End        Size       File system  Flags
 1      192s     614591s    614400s    ext2
 2      614592s  62915039s  62300448s  btrfs
```

The shutdown and boot of sle micro is successful!!

```shell
Welcome to SUSE Linux Enterprise Micro 5.4  (s390x) - Kernel 5.14.21-150400.24.4
6-default (ttyS0).  
  
SSH host key: SHA256:N9QY7+tWPdovF3zHHQPsSmhHn51HF39fJBh1o6iK500 (RSA)  
SSH host key: SHA256:HEkZ+glert1m2Xqw1zxNmOymWrq5/l7ZuxlWkhlf+Oo (DSA)  
SSH host key: SHA256:QSIMDcuSCsfuf1UVK62J22zk3EfCYDGLdrHyaDht0Oc (ECDSA)  
SSH host key: SHA256:nlWnT5IO3A8WuNDf+O7YwlWjZt6A+HPPtsWWcZzgajY (ED25519)  
eth0: 10.161.129.63   
  
  
Configured with Combustion at Fri Mar 31 18:57:58 UTC 2023  
Web console: https://k3s01:9090/ or https://10.161.129.63:9090/  
  
k3s01 login: 
```

The root partition (dasdb2) takes the remainder of the disk

```shell
k3s01:~ # lsblk
NAME     MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
dasda     94:0    0   30G  0 disk 
├─dasda1  94:1    0  300M  0 part /boot/zipl
└─dasda2  94:2    0 29.7G  0 part /home
                                  /opt
                                  /.snapshots
                                  /srv
                                  /usr/local
                                  /boot/grub2/s390x-emu
                                  /boot/writable
                                  /var
                                  /root
                                  /
```
                                  
I believe https://github.com/OSInside/kiwi/pull/2268/files#diff-01eb8371d03fa769864ef87789444c6bb40e3a4b2d1b77df099e1afd602d7026 line 219 is the cause for the `parted: invalid token: primary` error.  

The parted command that worked for me manually is `parted /dev/dasda unit s mkpart btrfs 614592s 100%`.

schaefi:
@davidcassany 

>Hmm not so sure about the Shellcheck complaint, I don't get it locally in my env and I doubt I can actually quote the >offending variable expansion without altering the loop behavior.... 

Well the check is correct and the code should change. See the following example

```bash
x=(a b c)

# critical, whitespace would be a field delimiter
for element in ${x[*]};do
    echo "${element}"
done

# clean
for element in "${x[@]}";do
    echo "${element}"
done
```

my shellcheck provides the same complain and I also stumbled over that issue when using shell arrays.
It's a very simple change. Just use ```for part_uuid in "${part_uuids[@]}";do``` as suggested by the checker. Thanks

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -278,7 +403,16 @@ function get_free_disk_bytes {
 }
 
 function get_partition_table_type {
-    blkid -s PTTYPE -o value "$1"
+    case $(basename "$1") in
+        # Assume DASD if the device starts with "dasd" blkid does not
+        # properly recognize dasd as PTTYPE bsc#1209247
+        dasd*)
```
schaefi, line 409: Uhm, this really smells. There is no programatical way to detect the partition table type on a DASD ?

mfriesenegger:
```
k3s01:~ # parted /dev/dasda print
Model: IBM S390 DASD drive (dasd)
Disk /dev/dasda: 32.2GB
Sector size (logical/physical): 512B/4096B
Partition Table: dasd
Disk Flags: 

Number  Start   End     Size    File system  Flags
 1      98.3kB  315MB   315MB   ext2
 2      315MB   32.2GB  31.9GB  btrfs
```

schaefi:
@davidcassany 
> I am not pleased with this PR, but according to a reported issue the partition expansion does not work on s390, and according to the experts fdasd is not capable for this job, but parted is

If the combination fdasd + parted works I have no objections to bring back parted to the initrd code. Actually from the description here there is also no other chance in order to fix things, right :)

@Vogtinator 
> Yes. Ultimately I'd like to see parted being used for all plaforms again

We had this before and it was a nightmare. parted aimed to be the one partitioner to handle them all but it had so many issues and needed so many nasty code paths that it was more straight forward to add a partitioner interface and allow individual implementations selecting the "best" working tools. Whatever "best" mean can change in the future but unless we have real issues with the current use of the partition tools (like in this case) I'm not open to replace any working code paths with parted. Sorry I can still feel the pain from the past with all that :)

The proposed change here is acceptable, except for the selection by the name of a device node. I can imagine some sort of udev rule, persistent device name or similar which turns this selection method into an unpredictable nightmare. So if we have to keep this then it should be only valid together with an arch check for s390/s390x

Just my2cents

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
schaefi: imho if we have to keep this, the code should be selected exclusively by architecture. DASD only exists on s390/s390x.

schaefi:
> > It is! By doing `echo -e "resizepart 1\n\nq\n" | parted /dev/dasdb` (which resizes the first partition to the max size, effectively a noop), parted rewrites the partition table to span the whole device. Just doing this before using `fdasd` should do the job.
> 
> Mmm interesting, hence we could eventually find a workaround that is not dropping the current fdasd logic and just adding a dummy `parted` call to ensure the partition table matches disk geometry. Given this is only targeting s390 and that it should take few lines of code I'd say we can consider it. I'll update the PR with such information.
> 
> I's say there is not need to run parted interactive so I'd translate your command to something like `parted --script --machine /dev/dasdb resizepart 1`. I am wondering if we always have more than one partition for s390, as far as I understood if the partition is actually a candidate to be resized this will raise a parted bug. AFAIU this is why you run it for the first partition. @Vogtinator Did I got it right?

If this method can be implemented I would also vote for it

> I am wondering if we always have more than one partition for s390

There can be images with just one partition. iirc the boot sequence on s390 was zipl, loading kernel and an initrd from which a userspace grub was called. If zipl is able to read the kernel and initrd from the same device that is the root you can build the image with just one partition.

davidcassany:
> If the combination fdasd + parted works I have no objections to bring back parted to the initrd code. Actually from the description here there is also no other chance in order to fix things, correct ?

Yes agree, I am preparing a new PR for that, just not to mess with this one. I hope we will simply end up closing this one without merging.

> We had this before and it was a nightmare.

Agree, parted is really bad to script, it aims to support everything but then the UX is not properly designed to work for all and there is kind of a mix with flags/partitiontypes/filesystems. As we also saw here in this PR `mkpart` command works different in DOS, GPT and s390... and this is not properly documented in man pages.

davidcassany:
Closing as this is superseded by #2272

PR #2272, opened 2023-04-03
Expand partition table with parted on s390

davidcassany:
Make use of parted to detect partition table type and to expand dasd partition table types.

Fixes bsc#1209247

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -137,6 +137,13 @@ function create_dasd_partitions {
     local ignore_cmd=0
     local ignore_cmd_once=0
     local cmd
+
+    # We resize first partition to enfore rewrite of the current
+    # partition table updated to the actual disk geometry. This is
+    # to circumvent the fdasd limitation of not being capable to
+    # expand the partition table up to the disk size bsc#1209247
+    parted --script --machine "${disk_device}" resizepart 1
+
     for cmd in ${partition_setup};do
         if [ "${ignore_cmd}" = 1 ] && echo "${cmd}" | grep -qE '[dntwq]';then
             ignore_cmd=0
@@ -278,7 +285,9 @@ function get_free_disk_bytes {
 }
 
 function get_partition_table_type {
-    blkid -s PTTYPE -o value "$1"
+    local disk_device="$1"
+    
+    parted --script --machine "${disk_device}" print | grep "^${disk_device}" | cut -d":" -f6
 }
 
 function get_partition_uuid {
--- a/dracut/modules.d/99kiwi-lib/module-setup.sh
+++ b/dracut/modules.d/99kiwi-lib/module-setup.sh
@@ -17,7 +17,7 @@ install() {
     inst_multiple \
         blkid blockdev dd mkdir rmdir \
         grep cut tail head tr bc true false mountpoint \
-        basename partprobe sfdisk sgdisk mkswap readlink lsblk \
+        basename partprobe parted sfdisk sgdisk mkswap readlink lsblk \
         btrfs xfs_growfs resize2fs \
         e2fsck btrfsck xfs_repair \
         vgs vgchange lvextend lvcreate lvresize pvresize \


davidcassany:
Alternate approach of what is attempted in #2268

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -278,7 +285,9 @@ function get_free_disk_bytes {
 }
 
 function get_partition_table_type {
-    blkid -s PTTYPE -o value "$1"
+    local disk_device="$1"
+    
+    parted --script --machine "${disk_device}" print | grep "^${disk_device}" | cut -d":" -f6
```
davidcassany, line 290: This change affects all archs and types. We could consider using parted only for `s390`, but I dislike the idea of start adding arch cuts here and there if there is a generic approach.

davidcassany:
@Vogtinator @schaefi @mfriesenegger here is the other commented approach to workaround fdasd limitations using parted. If this works I'd say this is the less invasive approach and I'd vote for this approach. Parted is hard to script and adding parted back in a transparent manner implies adding quite some additional logic which is also hard to maintain (see #2268).

davidcassany:
@mfriesenegger actually successfully tested it. I consider the PR ready to be reviewed. Also closing #2268 as it was not enough and I believe the current PR is better.

@schaefi any thoughts and concerns regarding this change? specially regarding the partition type detection

schaefi:
@davidcassany 

> any thoughts and concerns regarding this change?

First of all, thanks much for digging into this. The alternative solution provided looks way better and less intrusive.
We can certainly go that way but there is one aspect I think we should change. The proposed change here has
exactly one use case and that's the s390/fdasd issue for which this change here still can be considered a workaround.
The way it's done however, creates a global code change that will affect all architectures and also introduces a
generic dependency to parted again which is of no benefit in this scope but imho could introduce issues. You are most probably right that the change on the detection of the partition table type with parted vs. blkid will not cause any regressions would you put your hand on fire about it ? Also the other parted call that does a resize is called unconditionally in any case and as documented is only useful to the particular s390/fdasd case.

So what I would change is that these code paths should be called only on s390 also the dependency to parted
in the dracut module should be arch specific. I know you would like to prevent this extra conditions but I think we should not run code paths that only exists to make s390 happy in any other scope.

Thoughts ?

schaefi (changes requested):
As I tried to explained I believe the changes are too dangerous with regards to regressions that I think they deserve it's own arch scope for which they were produced

davidcassany:
> @davidcassany
> 
> > any thoughts and concerns regarding this change?
> 
> First of all, thanks much for digging into this. 

Sure, in any case kudos belong to @Vogtinator and @mfriesenegger.

> The alternative solution provided looks way better and less intrusive. We can certainly go that way but there is one aspect I think we should change. The proposed change here has exactly one use case and that's the s390/fdasd issue for which this change here still can be considered a workaround. The way it's done however, creates a global code change that will affect all architectures and also introduces a generic dependency to parted again which is of no benefit in this scope but imho could introduce issues. You are most probably right that the change on the detection of the partition table type with parted vs. blkid will not cause any regressions but would you put your hand on fire about it ? Also the other parted call that does a resize is called unconditionally in any case and as documented is only useful to the particular s390/fdasd case.

Yes, I know and I get your point, that's also why I asked feedback for this specific part on my comments. I am also not fully convinced as this is a change that affects any image using this library. Later today I will add here an arch switch and try to limit the change to s390. I am not a fan of the solution but I agree it is likely to be the safe path.

In addition, regarding parted dependency I noted the spec was already pulling parted as a requirement for all archs but it was not used in any... I guess this was a leftover. In case we want to address that I'd do it in a separate PR.

schaefi:
> In addition, regarding parted dependency I noted the spec was already pulling parted as a requirement for all archs but it was not used in any... I guess this was a leftover. In case we want to address that I'd do it in a separate PR.

yeah you are right, would be great if you can fix that too. Thanks much :+1:

schaefi:
> ... Later today I will add here an arch switch and try to limit the change to s390.

great thanks, I'm going to set the work in progress label until then.

davidcassany:
> > ... Later today I will add here an arch switch and try to limit the change to s390.
> 
> great thanks, I'm going to set the work in progress label until then.

Sure, just adapted the PR with the discussed changes.

@Vogtinator @mfriesenegger Would you mind rechecking the current PR fixes the issue? Changes are trivial since last try, but still would be good giving it a run just in case there is some stupid leftover.

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -137,6 +137,13 @@ function create_dasd_partitions {
     local ignore_cmd=0
     local ignore_cmd_once=0
     local cmd
+
+    # We resize first partition to enfore rewrite of the current
+    # partition table updated to the actual disk geometry. This is
+    # to circumvent the fdasd limitation of not being capable to
+    # expand the partition table up to the disk size bsc#1209247
+    parted --script --machine "${disk_device}" resizepart 1
```
schaefi, line 145: shouldn't this code part also be s390 specific ?

davidcassany: this is only used for dasd partitions (is dasd existing beyond s390?). I agree this is not explicitly called only for s390, but this method is only for dasd partitions which are only detected by `parted` in `get_partition_table_type`.

We can add an arch specific path here, but feels redundant to me. Also the fdasd tool called down the line is also only added for s390 arch (see the `module-setup.sh`) so this method was already not functional for any arch other than s390. Here there is (and there was already) the assumption that partition type dasd matches s390 arch, I think this is a reasonable assumption. I am not familiar with s390 specifics though.

mfriesenegger: > this is only used for dasd partitions (is dasd existing beyond s390?).

dasd is only exists as a disk type for s390x.

schaefi: Sorry I overlooked from reading the patch that this code is in the ```create_dasd_partitions``` which is already exclusive to s390 and the DASD handling. I agree all good, sorry for the noise

schaefi (approved):
From my perspective all good now :+1: Thanks for your effort

schaefi:
Let's wait for Mike and/or Fabian to give their thumbs up

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -278,7 +285,13 @@ function get_free_disk_bytes {
 }
 
 function get_partition_table_type {
-    blkid -s PTTYPE -o value "$1"
+    local disk_device="$1"
+    
+    if [[ "$(uname -m)" =~ s390 ]];then
```
Vogtinator, line 290: IMO needs a comment that blkid doesn't work for dasd

Vogtinator:
> Let's wait for Mike and/or Fabian to give their thumbs up

Code looks good. I built an image with the changes and asked Mike to give it a try.

Vogtinator (approved):
Test was a success

schaefi:
@davidcassany sure, I'll create a release today and put it on Staging

PR #2280, opened 2023-04-13
Add zstd and xzio to grub image

schaefi:
If kiwi builds its own grub image, make sure the compression algorithms zstd and xz are supported


schaefi:
Exactly that's my driver for this change :) In kiwi we can say ```squashfscompression="zstd"``` but it's pointless if the later grub cannot read it :)

Conan-Kudo:
We should probably also ask Red Hat/Fedora to incorporate this into their grub images if they already haven't done so...

PR #2283, opened 2023-04-18
Fixed Fedora prebuild EFI binary lookup

### LINKED ISSUE #2270: kiwi doesn't install the correct grub2-efi binary for ISOs for CentOS/Fedora images
## Problem description
When creating CentOS/Fedora live media, kiwi does not install the right signed grub2 EFI binary.
As a consequence, the resulting image is not bootable.

This is caused by kiwi not being aware of the `gcdx64.efi` (x64), `gcdaa64.efi` (aa64), and `gcdia32.efi` (ia32) binaries to install in place of the normal `grubx64.efi`, `grubaa64.efi`, and `grubia32.efi` binaries for CD/ISO boot.

You can see the install pattern for these in Lorax: https://github.com/weldr/lorax/commit/7ba88931ac38116d1cd0fd9c70b38e5fcc2475a0

## Expected behavior
Built ISO boots in a VM in UEFI mode

1. Build a CentOS Hyperscale GNOME or KDE live image
2. Boot it in GNOME Boxes in UEFI mode

Sample KDE build: https://cbs.centos.org/koji/taskinfo?taskID=3303471
Sample GNOME build: https://cbs.centos.org/koji/taskinfo?taskID=3303474

* KIWI version: 9.24.52
* Operating system host version: CentOS Stream 9
* Operating system target version: CentOS Stream 9
* Koji version (N/A if not using Koji): 1.31

schaefi:
When creating CentOS/Fedora live media, kiwi does not install the right signed grub2 EFI binary. This was caused by kiwi not being aware of the gcdx64.efi(x86_64), gcdaa64.efi(aarch64) binaries for CD/ISO boot. This Fixes #2270

### DIFF
--- a/.github/workflows/ci-code-style.yml
+++ b/.github/workflows/ci-code-style.yml
@@ -22,7 +22,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install Tox
         run: |
-          sudo apt update && sudo apt install tox
+          python -m pip install --upgrade pip
+          python -m pip install tox==3.28.0
       - name: Run code checks
         run: |
           tox -e check
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -789,7 +789,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         code should act as the fallback solution
         """
         log.warning(
-            '--> Running fallback setup for shim secure boot efi image'
+            'Running fallback setup for shim secure boot efi image'
         )
         if not lookup_path:
             lookup_path = self.boot_dir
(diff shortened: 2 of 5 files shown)


schaefi:
> Can we get some non-debug output printed when this happens so that it shows up in the logs?

The output I pasted is from a log excerpt. In particular the setup of the EFI binaries can be found in the log by looking for ```/EFI/BOOT```. Depending on secure boot vs. not secure boot, or shim via shiminstall or fallback, or EFI boot for ISO or disk or grub vs sd-boot, or or or :) there are several different log entries for the different processing that happens behind the scene. What's common is the path the data lands. I think this is also the only thing that is really a common standard .

So I'm not sure what log information you are particularly interested or see missing ?

Thanks

Conan-Kudo:
If you're not using `--debug` it won't be printed in the output. It'll just say the following:

```
[ WARNING ]: 02:24:34 | --> Running fallback setup for shim secure boot efi image
```

I would prefer this to be more specific.

schaefi:
> If you're not using `--debug` it won't be printed in the output. It'll just say the following:
> 
> ```
> [ WARNING ]: 02:24:34 | --> Running fallback setup for shim secure boot efi image
> ```
> 
> I would prefer this to be more specific.

agreed, let's provide more details here

schaefi:
@Conan-Kudo ok I added which grub image(s) are used as info on the console. Can you take a look again ?

Thanks

Conan-Kudo (approved):

PR #2285, opened 2023-04-20
Encode remote URLs

schaefi:
Special characters in a URL e.g the @ sign needs to be encoded as part of a remote URL.


schaefi:
This originates from an issue reported on the mailing list. As I can't reproduce it, the fix was done by best guess and I'd like to wait for some feedback from the user prior merge

schaefi:
I got positive feedback on this patch

@Conan-Kudo could you have a look ? Thanks much

Conan-Kudo (commented):
LGTM.

PR #2286, opened 2023-04-20
Update kiwi-systemdeps-containers requires list

### LINKED ISSUE #2284: kiwi-systemdeps-containers pulls in fb-util-for-appx for WSL
## Problem description

The subpackage `kiwi-systemdeps-containers` pulls in `fb-util-for-appx` tooling for WSL. This requires us to throw `appx` into ALP just to build containers.

## Expected behavior

No WSL tools should be required to build containers.

## Proposal

Introduce a `kiwi-systemdeps-wsl` subpackage that requires `fb-util-for-appx`.

schaefi:
Do not strictly require fb-util-for-appx. In order to create WSL containers fb-util-for-appx is a requirement but usually users create OCI containers and WSL containers are still a niche case. Thus it's ok to reduce the requirement into a recommends. This Fixes #2284

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -127,19 +127,15 @@ This metapackage installs the necessary system dependencies
 to run KIWI.
 
 %package -n kiwi-systemdeps-containers
-Summary:        KIWI - host requirements for container images
+Summary:        KIWI - host requirements for OCI container images
 Group:          %{sysgroup}
 Provides:       kiwi-image-docker-requires = %{version}-%{release}
 Obsoletes:      kiwi-image-docker-requires < %{version}-%{release}
-Provides:       kiwi-image-wsl-requires = %{version}-%{release}
-Obsoletes:      kiwi-image-wsl-requires < %{version}-%{release}
 %if "%{_vendor}" != "debbuild"
 Provides:       kiwi-image:docker
-Provides:       kiwi-image:appx
 %endif
 %if 0%{?suse_version}
 Requires:       umoci
-Recommends:     fb-util-for-appx
 %else
 Requires:       buildah
 %endif
@@ -147,7 +143,23 @@ Requires:       skopeo
 
 %description -n kiwi-systemdeps-containers
 Host setup helper to pull in all packages required/useful on
-the build host to build container images e.g docker, wsl
+the build host to build OCI container images
+
+%package -n kiwi-systemdeps-containers-wsl
+Summary:        KIWI - host requirements for WSL container images
+Group:          %{sysgroup}
+Provides:       kiwi-image-wsl-requires = %{version}-%{release}
+Obsoletes:      kiwi-image-wsl-requires < %{version}-%{release}
+%if "%{_vendor}" != "debbuild"
+Provides:       kiwi-image:appx
+%endif
+%if 0%{?suse_version}
+Requires:       fb-util-for-appx
+%endif
+
+%description -n kiwi-systemdeps-containers-wsl
+Host setup helper to pull in all packages required/useful on
+the build host to build WSL container images
 
 %package -n kiwi-systemdeps-iso-media
 Summary:        KIWI - host requirements for live and install iso images
@@ -317,6 +329,7 @@ Requires:       kiwi-systemdeps-bootloaders = %{version}-%{release}
 %if 0%{?fedora} || 0%{?rhel} || 0%{?suse_version}
 # None of the container build tools are available in Debian/Ubuntu
 Requires:       kiwi-systemdeps-containers = %{version}-%{release}
+Requires:       kiwi-systemdeps-containers-wsl = %{version}-%{release}
 %endif
 Requires:       kiwi-systemdeps-filesystems = %{version}-%{release}
 Requires:       kiwi-systemdeps-disk-images = %{version}-%{release}
@@ -656,6 +669,9 @@ fi
 # Empty metapackage
 
 %files -n kiwi-systemdeps-containers
+# Empty metapackage
+
+%files -n kiwi-systemdeps-containers-wsl
 # Empty metapackage
 
 %files -n kiwi-systemdeps-iso-media
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -139,7 +139,7 @@ Provides:       kiwi-image:appx
 %endif
 %if 0%{?suse_version}
 Requires:       umoci
-Requires:       fb-util-for-appx
+Recommends:     fb-util-for-appx
 %else
 Requires:       buildah
 %endif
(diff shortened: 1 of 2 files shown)


dcermak:
Wouldn't another subpackage not be a better solution? The purpose of these subpackages is mostly to pull in the right dependencies on OBS, but with the WSL dep being weak now, make it effectively useless for the purpose of building on OBS.

Conan-Kudo:
Is this really a change worth doing? `appx-util` is tiny and simple. There's not a lot going on there and it's part of kiwi's container functionality.

dcermak:
> Given we add another sub-package e.g named `systemdeps-containers-wsl`, the prjconf would change to
> 
> ```
> support: systemdeps-containers
> support: systemdeps-containers-wsl
> ```
> 
> of if done as intended obs pulls them in ... same as above
> 
> In any case you need to specify that you want more from the container toolchain and as long as it's only one package there is not much benefit in maintaining an extra systemdeps for it, or am I mistaken ?

If `systemdeps-containers-wsl` requires `systemdeps-containers`, then that should get pulled in as well and you'd save yourself one line.

But my point is: converting the requires to recommends makes this effectively useless and we might just as well drop it, if the user has to add it into the prjconf anyway…

Conan-Kudo:
I'm going to make the argument that we shouldn't do this at all and ask why SUSE has a problem with this.

Vogtinator:
We prefer to avoid having stuff we don't need in our distro(s). Currently without shipping `fb-util-for-appx` you cannot build containers with kiwi.

dcermak:
Neal Gompa ***@***.***> writes:

> I'm going to make the argument that we shouldn't do this at all and ask why SUSE has a problem with this.

Because it requires us to force ignore the dependency on appx-util or
manually maintain the list of build tools for containers. Neither of the
two is really elegant nor imho user-friendly

schaefi:
@dcermak I have no problem in splitting the container systemdeps into OCI and WSL. How about it now ?

schaefi:
@dcermak 

> If systemdeps-containers-wsl requires systemdeps-containers, then that should get pulled in as well and you'd save yourself one line.

you should also have in mind that a user in theory doesn't have to specify any of the systemdeps packages in the prjconf because the OBS build script code implicitly pulls them in according to the selected image type in the kiwi XML config. Thus if we agree to this change here somebody has to inform the OBS team that ````<type image="appx"``` has to pull in ```systemdeps-containers-wsl``` and ````<type image="docker|oci"``` continues to pull in ```systemdeps-containers```

Conan-Kudo:
We should spread the pain if we're going to do this: move the OCI container stuff to its own subpackage too. That way the OBS team has to fix it for everything.

Vogtinator:
pain? It's a single line in the prjconf.

Conan-Kudo:
> pain? It's a single line in the prjconf.

This change affects every user of kiwi, not just OBS.

Vogtinator:
> > pain? It's a single line in the prjconf.
> 
> This change affects every user of kiwi, not just OBS.

I doubt that there are (m)any users of kiwi using it to build appx images.

schaefi:
ok let's collect it:

* The number of users using kiwi for building WSL containers is probably really small, maybe nobody except SUSE with OBS. So @Vogtinator might have a point that the impact of not requiring ```fb-util-for-appx``` in ```kiwi-systemdeps-containers``` is small
* The ```kiwi-systemdeps-*``` concept to pull in required software for building images also comes with the global ```kiwi-systemdeps``` package which itself pulls in all other ```kiwi-systemdeps-*``` packages. That global ```kiwi-systemdeps```, as you can see in the pull request, would also pull in the new ```-wsl``` systemdeps package. Thus there is no behavior change for users of ```kiwi-systemdeps```. I think the most people just use this package to get *everything* and maybe also koji implemented it that way, not sure though ?
* On the support for building WSL containers we might be facing changes on the toolchain. afaik ```fb-util-for-appx``` has some sort of successor (don't remember the name) and there is also WSL2, so maybe a split of the systemdeps makes sense ?
* Only in case the building host pulls in required software according to the requested image type, the change from here needs to be taken into account. Usually this sort of host setup is connected with some infrastructure code. As I said OBS is affected and needs an update of the build script. I don't know about koji but maybe @Conan-Kudo knows. I'm not aware of other build host systems which would be affected and for manual setups I'm pretty sure used just pull in kiwi-systemdeps and do not differentiate.

From my perspective I think the pain is small maybe not even existing and handling WSL requirements in an extra package might be beneficial in the long run.

Just my2cents

Conan-Kudo:
> On the support for building WSL containers we might be facing changes on the toolchain. afaik `fb-util-for-appx` has some sort of successor (don't remember the name) and there is also WSL2, so maybe a split of the systemdeps makes sense ?

The successor is `appx-util`, which is part of this GitHub org: https://github.com/OSInside/appx-util

dcermak (approved):
LGTM, thanks!

package/python-kiwi-spec-template
```diff
@@ -127,27 +127,39 @@ This metapackage installs the necessary system dependencies
...
+Summary:        KIWI - host requirements for WSL container images
+Group:          %{sysgroup}
+Provides:       kiwi-image-wsl-requires = %{version}-%{release}
+Obsoletes:      kiwi-image-wsl-requires < %{version}-%{release}
+%if "%{_vendor}" != "debbuild"
+Provides:       kiwi-image:appx
+%endif
+%if 0%{?suse_version}
+Requires:       fb-util-for-appx
+%endif
```
Conan-Kudo, line 158: This is bad. It makes this package pointless on non-SUSE systems.

schaefi: @Conan-Kudo I also wondered about this part, but this is what we had in the spec before. Here is the former ```kiwi-systemdeps-containers``` package spec

```
%package -n kiwi-systemdeps-containers
Summary:        KIWI - host requirements for container images
Group:          %{sysgroup}
Provides:       kiwi-image-docker-requires = %{version}-%{release}
Obsoletes:      kiwi-image-docker-requires < %{version}-%{release}
Provides:       kiwi-image-wsl-requires = %{version}-%{release}
Obsoletes:      kiwi-image-wsl-requires < %{version}-%{release}
Provides:       kiwi-image:docker
Provides:       kiwi-image:appx
%endif
%if 0%{?suse_version}
Requires:       umoci
Requires:       fb-util-for-appx
%else
Requires:       buildah
%endif
Requires:       skopeo
```

So you see ```fb-util-for-appx``` was only used in combination with suse systems. I guess we had a reason to do it that way ? Probably because it does not exist for others ? We also have WSL integration tests for suse systems building. Which brings me to the question if and how we support WSL containers for other OS'es ?

Conan-Kudo: We probably failed to update this when `appx-util` got added to Fedora and EPEL.

Issue #2172, opened 2022-07-14
AlmaLinux/CentOS UEFI Live images not booting in qemu/VirtualBox/(most) pyhsical PCs


TheMasterFX:
## Problem description
I build a Live ISO with this kiwi config: https://github.com/OSInside/kiwi/tree/master/build-tests/x86/centos/test-image-live-disk-v8
When I try to boot the ISO in qemu in uefi mode (virt-manager => mount ISO) it doesn't  boot properly (no boot device found). I found the same problem on many physical PCs.
If I manually add a boot entry which points to /EFI/almalinux/shimx64.efi or /EFI/centos/shimx64.efi it works properly. 
Also on many physical PCs I get a message to repair the boot entrys (seems to be part of the uefi firmware)
In my opionion there is something wrong in the bootfiles structure as mentioned here: [Issue #2008](https://github.com/OSInside/kiwi/issues/2008), so maybe my issue is kind of a duplicate.
As mentioned by _ignatenkobrain_ in his issue the structure of a "official ISO" seems to be different.
I tried to modify the boot structure by using the editbootinstall script, but this only works on OEM images, not (Live)ISO, even this don't might have worked since the efi-partition for the ISO is build in a different loop-device/folder as the appliance itself. 
I even have the problem with CentOS 7.

This is what Kiwi generates in the EFI folder
```
$ sudo tree -s /mnt/EFI/
/mnt/EFI/
├── [       2048]  almalinux
│   ├── [        122]  BOOTX64.CSV
│   ├── [       2048]  fonts
│   │   └── [    2560080]  unicode.pf2
│   ├── [        141]  grub.cfg
│   ├── [    2219032]  grubx64.efi
│   ├── [     850232]  mmx64.efi
│   ├── [     922832]  shimx64-almalinux.efi
│   └── [     925424]  shimx64.efi
└── [       2048]  BOOT
    ├── [     922832]  bootx64.efi
    ├── [     925424]  BOOTX64.EFI
    ├── [      90296]  fbx64.efi
    ├── [        141]  grub.cfg
    ├── [    2219032]  grubx64.efi
    └── [     850232]  mmx64.efi

3 directories, 13 files
```
This is what we found on the official AlmaLinux Install ISO
```
$ sudo tree -s /mnt/EFI/
/mnt/EFI/
└── [       2048]  BOOT
    ├── [       1351]  BOOT.conf
    ├── [     737696]  BOOTIA32.EFI
[...]

PR #2288, opened 2023-04-25
Look for specific `shimx64.efi` binary first

jloeser:
In CentOS, `/boot/efi/EFI/*/shim*.efi` extends to
`/boot/efi/EFI/centos/shimx64-centos.efi` which is not signed by M$ but CentOS itself. This in turn does not boot on SecureBoot enabled systems.

---

CentOS 8 Stream ISO does not boot on SecureBoot enabled systems.

Problem is the selected `shimx64-centos.efi` binary which is signed by CentOS, not M$:
```
[ DEBUG   ]: 14:39:06 | EXEC: [cp /tmp/centos/build/image-root/boot/efi/EFI/centos/shimx64-centos.efi /tmp/centos/live-media.s0dscud5/EFI/BOOT/bootx64.efi]
```
```
# pesign -S -i ./mnt/EFI/BOOT/bootx64.efi | grep common
The signer's common name is CentOS Secure Boot Signing 201
``` 

This might fix https://github.com/OSInside/kiwi/issues/2172

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -811,12 +811,12 @@ class Defaults:
             shim_pattern_type('/usr/lib/shim/shim*.efi.signed', 'shimx64.efi'),
             shim_pattern_type('/usr/share/efi/*/shim.efi', None),
             shim_pattern_type('/usr/lib64/efi/shim.efi', None),
-            shim_pattern_type('/boot/efi/EFI/*/shimx64.efi', None),
-            shim_pattern_type('/boot/efi/EFI/*/shim*.efi', None),
+            shim_pattern_type('/boot/efi/EFI/*/shim[a-z]*.efi', None),
+            shim_pattern_type('/boot/efi/EFI/*/shim.efi', None),
             shim_pattern_type('/usr/lib/shim/shim*.efi', None)
         ]
         for shim_file_pattern in shim_file_patterns:
-            for shim_file in glob.iglob(root_path + shim_file_pattern.pattern):
+            for shim_file in sorted(glob.iglob(root_path + shim_file_pattern.pattern), key=len):
                 if not shim_file_pattern.binaryname:
                     binaryname = os.path.basename(shim_file)
                 else:
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -811,6 +811,7 @@ class Defaults:
             shim_pattern_type('/usr/lib/shim/shim*.efi.signed', 'shimx64.efi'),
             shim_pattern_type('/usr/share/efi/*/shim.efi', None),
             shim_pattern_type('/usr/lib64/efi/shim.efi', None),
+            shim_pattern_type('/boot/efi/EFI/*/shimx64.efi', None),
             shim_pattern_type('/boot/efi/EFI/*/shim*.efi', None),
             shim_pattern_type('/usr/lib/shim/shim*.efi', None)
         ]


schaefi (approved):
Looks good to me

I can hardly estimate if there are any unwanted after effects though

@Conan-Kudo @davidcassany can you please double check ? Thanks

Conan-Kudo:
All of them should be copied. Not just one of them. The `shimx64-centos.efi` binary is initial one compiled and signed with the distro key. It is then resigned again with the Microsoft CA (which is the `shimx64.efi` binary).

Cf. https://gitlab.com/redhat/centos-stream/rpms/shim/-/blob/0dd668655814327d46f5d8328dd6b773fc226961/shim.rpmmacros#L96-121

@martinezjavier can confirm my understanding here.

kiwi/defaults.py
```diff
@@ -811,6 +811,7 @@ def get_shim_loader(root_path: str) -> Optional[shim_loader_type]:
             shim_pattern_type('/usr/lib/shim/shim*.efi.signed', 'shimx64.efi'),
             shim_pattern_type('/usr/share/efi/*/shim.efi', None),
             shim_pattern_type('/usr/lib64/efi/shim.efi', None),
+            shim_pattern_type('/boot/efi/EFI/*/shimx64.efi', None),
```
Conan-Kudo, line 814: This will break for non-x64 arches for CentOS.

Conan-Kudo (changes requested):

kiwi/defaults.py
jloeser: You mean it will work only for x64 correctly?

What is about adding concrete filenames for all arches (`shimia32.efi`, `shimaa64.efi`, etc.) before the wildcard line matches?

jloeser:
What's interesting, during my local testing - I just isolated the `get_shim_loader()` function from master - it seems to deliver correct paths:
```
$ ls ~/Downloads/shim-centos-x64/boot/efi/EFI/centos 
BOOTX64.CSV  mmx64.efi  shimx64-centos.efi  shimx64.efi
$ python3 ~/devel/git/github/kiwi/find-shim.py /home/jloeser/Downloads/shim-centos-x64
shim_loader_type(filename='/home/jloeser/Downloads/shim-centos-x64/boot/efi/EFI/centos/shimx64.efi', binaryname='shimx64.efi')
$ ls ~/Downloads/shim-centos-aarch64/boot/efi/EFI/centos 
BOOTAA64.CSV  mmaa64.efi  shimaa64-centos.efi  shimaa64.efi  shim.efi
$ python3 ~/devel/git/github/kiwi/find-shim.py /home/jloeser/Downloads/shim-centos-aarch64 
shim_loader_type(filename='/home/jloeser/Downloads/shim-centos-aarch64/boot/efi/EFI/centos/shimaa64.efi', binaryname='shimaa64.efi')
$ ls ~/Downloads/shim-centos-ia32/boot/efi/EFI/centos 
BOOTIA32.CSV  mmia32.efi  shim.efi  shimia32-centos.efi  shimia32.efi
$ python3 ~/devel/git/github/kiwi/find-shim.py /home/jloeser/Downloads/shim-centos-ia32 
shim_loader_type(filename='/home/jloeser/Downloads/shim-centos-ia32/boot/efi/EFI/centos/shimia32.efi', binaryname='shimia32.efi')
```

jloeser:
Looks like returned glob list order is arbitrary. `glob.py` uses `os.listdir()` and this returns a "list in arbitrary order" (https://docs.python.org/3.6/library/os.html#os.listdir).

First of all I would propose to add `sorted()` to have more deterministic behavior (sort by filename):
```
 47     for shim_file_pattern in shim_file_patterns:                                
 48         for shim_file in sorted(glob.iglob(root_path + shim_file_pattern.pattern)):
```
This at least provides same result on my local and test machine:
```
localhost:~/kiwi/kiwi-descriptions # python3 ~/find-shim.py /tmp/centos/build/image-root/
/tmp/centos/build/image-root//boot/efi/EFI/centos/shimx64-centos.efi
/tmp/centos/build/image-root//boot/efi/EFI/centos/shimx64.efi
```
```
$ python3 ~/devel/git/github/kiwi/find-shim.py /home/jloeser/Downloads/shim-centos-x64
/home/jloeser/Downloads/shim-centos-x64/boot/efi/EFI/centos/shimx64-centos.efi
/home/jloeser/Downloads/shim-centos-x64/boot/efi/EFI/centos/shimx64.efi
```

Does not solve our issue but makes debugging easier, but this needs to be tested of course.

jloeser:
Well, with the sorted globbing result all distribution should be tested, not only EL derivates.

schaefi:
ok from my point of view we can give this a try. I suggest to merge it and then put a version to Staging which triggers all integration test builds. Once they all built we need to test at least the major distro EFI test builds if they work in OVMF qemu.

Thoughts ?

@Conan-Kudo can you look at this again, I'm approving

schaefi (approved):
approved but we need to check the test results after merge

Conan-Kudo (approved):

Conan-Kudo (changes requested):
In the commit message, `derivates` -> `derivatives`

PR #2289, opened 2023-04-28
Fix restoration of grub2-install (bsc#1210948)

Vogtinator:
It checked for grub2-install.orig in the host, not the buildroot. This meant that it left /usr/sbin/grub2-install as a noop.

~Draft because not verified yet.~


Vogtinator:
This bug was introduced by 0aa69bdfc3b83def259e7d76498ebf9bb5ccdc41

CC @gmoro

Vogtinator:
I'm a bit surprised that the CI didn't catch this.

schaefi (approved):
Thanks, good catch :+1:

schaefi:
> I'm a bit surprised that the CI didn't catch this.

me too, I'm looking into it why this did not trigger an error

PR #2290, opened 2023-04-28
Use cross arch macros to interpret uint64_t

davidcassany:
This commit adds inttypes.h to make use of PRIu64 style macros in order to properly translate uint64_t across different architectures. More specific this allows the code to compile for ix86.

### DIFF
--- a/dracut/modules.d/80kiwi-verity/kiwi-parse-verity.c
+++ b/dracut/modules.d/80kiwi-verity/kiwi-parse-verity.c
@@ -22,6 +22,7 @@
 #define _POSIX_C_SOURCE (200809L)
 #define _LARGEFILE64_SOURCE
 
+#include <inttypes.h>
 #include <sys/types.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


davidcassany:
@schaefi my last submission to factory got rejected because dracut-kiwi-verity was not building for i586. I have no experience with such a module (I just quickly checked this feature when you added it, but never played with it). So I am not even sure if this is actually possible to be used within a i586 context, I assume this mostly having in mind RO filesystems like squash like the ones that could be flashed in embedded. So I am not sure there might be something else missing to actually be functional on 32bits archs 🤷🏽‍♂️ Since the fix (to compile at least) was pretty straight forward I just did a PR. 

If we consider this feature makes no sense in ix86 id say we completely skip the package in spec and makefiles on ix86.
Any thoughts?

Code wise I'd say this PR is safe harmless, it just tweaks a couple of printf literals.

schaefi (approved):
Looks good to me, should not harm

PR #2291, opened 2023-05-04
Allow to create plain cpio archives

schaefi:
Allow to create plain cpio archives to serve the idea of the mkosi initrd concept


Conan-Kudo:
We need to add `cpio` as a dependency, don't we?

schaefi:
> We need to add `cpio` as a dependency, don't we?

yes good point,  I added it to the kiwi-systemdeps-core requires. Thanks

Conan-Kudo (approved):

PR #2295, opened 2023-05-16
Added appx-util as WSL requires for Fedora/RHEL

schaefi:
Make sure the kiwi-systemdeps-containers-wsl meta package pulls in the required tools for Fedora/RHEL when building WSL containers. This is a followup to #2286


Conan-Kudo (approved):

Conan-Kudo:
Yeah, Debian/Ubuntu doesn't have it packaged at all.

PR #2298, opened 2023-05-22
Do not depend on dnf for SUSE ALP

jsrain:
SUSE ALP does not include dnf package manager, behave the same as for SLE15. For Tumbleweed nothing changes as suse_version got bumped to 1699.


davidcassany (approved):
Looks good to me, if DNF was not required already by SLE15 and ALP is not currently including it this makes complete sense to me.

jsrain:
We are trying to avoid dnf in ALP, so far successfully

PR #2299, opened 2023-05-23
dracut: modules.d: 99kiwi-lib: add bash shebangs and dependency

IsaacJT:
The scripts in the kiwi-lib module use bash-specific syntax like `function`, which causes the script to fail if another shell (like dash or busybox) is used to interpret the scripts. Specifically set the shebang to use bash as the shell interpreter and add bash as a dependency to the dracut module to fix this.

This resolves the following errors which are printed at boot when busybox tries to interpret the scripts:

    /bin/dracut-pre-mount: 3: /lib/kiwi-filesystem-lib.sh: function: not found
    /bin/dracut-pre-mount: 15: /lib/kiwi-filesystem-lib.sh: Syntax error: "}" unexpected


schaefi (approved):
Looks good, thanks :+1:

PR #2301, opened 2023-05-24
dracut: modules.d: 90kiwi-overlay: add support for device mapper targets

IsaacJT:
Allow specifying a device mapper target for the root of the overlayfs. This adds support for features like adding a dm-verity or dm-crypt backend to the overlayfs.

A device mapper target can be specified using the following syntax on the kernel command line:

    root=overlay:MAPPER=verityRoot

This translates to using the path `/dev/mapper/verityRoot` as the base of the overlayfs.


schaefi (approved):
Looks good :+1:

PR #2302, opened 2023-05-24
Add support for dnf5 package manager

### LINKED ISSUE #2262: Add support for dnf5
## Problem description

dnf5 is the new version of dnf which comes with a different interface than dnf4 uses.
Thus it makes sense to create a new package manager class to support this package
manager. Selecting dnf5 needs its own name. My suggestion would be

```xml
<packagemanager>dnf5</packagemanager>
```

### LINKED ISSUE #2300: Prepare for DNF 5, don't depend on `dnf`
(I'm filing issues with all the packages that currently depend on `dnf`)

DNF 5 is a new package manager that will replace DNF 4 in Fedora 39+. Starting in Fedora 39, the `dnf` command will be owned by the `dnf5` package rather than the `dnf` package, and `dnf5` will obsolete `dnf`. Since KIWI currently depends on DNF 4, it should choose one of the following strategies to avoid breaking the approaching Fedora upgrade and maintain compatibility with future versions of Fedora:
- Add support for DNF 5, and depend on the `dnf5` package in Fedora 39+ instead of dnf. The command-line interface isn't changing a whole lot between DNF 4 and DNF 5, so in KIWI's case, this change may be straightforward. Builds of DNF 5 are available in this COPR repository: https://copr.fedorainfracloud.org/coprs/rpmsoftwaremanagement/dnf-nightly/, and documentation is available here: https://dnf5.readthedocs.io/en/latest/. You can call DNF 5 explicitly at `/usr/bin/dnf5`. I see there is already https://github.com/OSInside/kiwi/issues/2262. I agree with @schaefi that the best course of action is to treat DNF 5 as a separate, new package manager, since its command-line interface is not identical to DNF 4's.
- Alternatively, or in the meantime, change the kiwi-systemdeps-core package to depend on `python3-dnf` instead of `dnf`, and call the `dnf-3` binary instead of `dnf`. The old DNF 4 command will still be available in the distribution, but only as `dnf-3`. (The binary is called `dnf-3`  rather than `dnf4` for historical reasons; it was originally the Python 3 version of DNF). The first option is preferred to this one; it is not recommended to modify installed software using both DNF 4 and DNF 5 on the same system.

We are planning to replace DNF with DNF5 in Fedora Rawhide very soon, in the next couple weeks, and the `kiwi-systemdeps-core` package will break as long as it still depends on the `dnf` package.

For more information about the switch to DNF 5, see https://fedoraproject.org/wiki/Changes/ReplaceDnfWithDnf5.

schaefi:
dnf5 is the next generation of the dnf package manager. dnf5 is not a drop in replacement to dnf which is the reason why this commit adds a new interface class for package and repository management under a new name: dnf5. Beginning with Fedora >= 39 dnf5 will be the standard. Thus the requirements on the packaging level were adapted too.

This Fixes #2300
This Fixes #2262


schaefi:
@Conan-Kudo this is a first and untested version of dnf5 support. The only effective code change is that we call ```dnf5``` instead of ```dnf``` if the ```<packagemanager>dnf5</packagemanager>``` is configured. Along with this code change I will create an integration test as soon as we have dnf5 available in some repo.

As you have packaged it I will see if I can reference it but it would be imho better to test on Rawhide directly as soon as it has landed there. Thus I also marked the PR as WIP but I appreciate a review as soon as possible for obvious "this doesn't work with dnf5 anymore" statements.

Thanks

schaefi:
@Conan-Kudo please find integration tests building with dnf5 here:

* https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_x86:rawhide

package/python-kiwi-spec-template
```diff
@@ -91,17 +91,26 @@ Recommends:     dpkg
 %endif
 %if 0%{?suse_version}
 Recommends:     gpg2
-Recommends:     dnf
 Recommends:     debootstrap
 Recommends:     dpkg
+%if 0%{?suse_version} >= 1650
+Recommends:     dnf
+%endif
```
Conan-Kudo, line 98: Why is this changing?

package/python-kiwi-spec-template
```diff
@@ -91,17 +91,26 @@ Recommends:     dpkg
...
 %endif
 # package managers required by distro
 %if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version} >= 1650
 Provides:       kiwi-packagemanager:microdnf
 Requires:       microdnf
 %endif
+%if 0%{?fedora} >= 39
+Requires:       dnf5
```
Conan-Kudo, line 106: Shouldn't we have `kiwi-packagemanager:dnf5` Provides too?

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: because of #2298

package/python-kiwi-spec-template
schaefi: yes good point, will fix it. Thanks

schaefi: done

Conan-Kudo:
Are we dropping the microdnf test then? microdnf is still sticking around for RHEL and F37...

schaefi:
> Are we dropping the microdnf test then? microdnf is still sticking around for RHEL and F37...

The test for microdnf is still present here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:fedora/test-image-microdnf

This project setup tests against Fedora_37

In the tests for Rawhide I cannot run microdnf tests because microdnf is a link to dnf5. Thus it's the same

Conan-Kudo:
Can you squash the commits? Then I can approve and merge it.

Conan-Kudo:
Sorry, I mean squash the code related changes into the same commit, so we have a code commit and a test commit.

schaefi:
> Sorry, I mean squash the code related changes into the same commit, so we have a code commit and a test commit.

done

PR #2304, opened 2023-05-25
Documentation: clearify dnf based collection selectors

frispete:
With 89b5a6f5269, the selection of collections was changed to allow group IDs only for `dnf` based package installers.

Complements #1860.


schaefi (approved):
Cool :+1: I've learned something. Thanks

Conan-Kudo (approved):

PR #2307, opened 2023-06-01
The actual padding is 0. And switch to sha256.

wfeldt:
## Issues

1. The actual padding used is 0. See https://github.com/OSInside/kiwi/blob/master/kiwi/iso_tools/xorriso.py#L95. Adjust the padding in `tagmedia`. Otherwise media checks will not verify the last 150 kiB.
2. While we change the `tagmedia` call anyway, move to sha256 as checksum.

Note: I haven't looked at what the 'mediacheck' boot option on Live media actually does. Does it rely on md5?

### DIFF
--- a/kiwi/iso_tools/iso.py
+++ b/kiwi/iso_tools/iso.py
@@ -48,9 +48,9 @@ class Iso:
         Command.run(
             [
                 'tagmedia',
-                '--md5',
+                '--digest', 'sha256',
                 '--check',
-                '--pad', '150',
+                '--pad', '0',
                 isofile
             ]
         )
--- a/test/unit/iso_tools/iso_test.py
+++ b/test/unit/iso_tools/iso_test.py
@@ -60,5 +60,5 @@ class TestIso:
     def test_set_media_tag(self, mock_command):
         Iso.set_media_tag('foo')
         mock_command.assert_called_once_with(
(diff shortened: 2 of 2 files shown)


schaefi (approved):
Thanks :+1: looks good

Conan-Kudo:
The actual commit message is pretty useless. ☹️

schaefi:
> The actual commit message is pretty useless. frowning_face

Hmm, yes it's missing the topic, sorry I didn't thought about that part because reading the code change made it obvious for me. So this is about caller parameters to the tagmedia tool for adding an ISO tag. iirc and @wfeldt correct me when wrong, this comes from the checkmedia package and is a suse specific tool, not entirely sure though

schaefi:
In this case we expect ```tagmedia``` to exist in the runtime environment

schaefi:
@wfeldt 

> Note: I haven't looked at what the 'mediacheck' boot option on Live media actually does. Does it rely on md5?

No it doesn't. We just call ```checkmedia $device``` For details see:

* https://github.com/OSInside/kiwi/blob/master/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh#L225

Should be ok, but would be great if you can take a look and give your thumbs up/down

Thanks

PR #2312, opened 2023-06-12
Change package manager dnf to dnf4

schaefi:
With dnf5 there is a successor for dnf but there will also be a transition period where there will be both, the former dnf and the new dnf5 available. For a clear distinction between the two we got the recommendation from the RedHat team to support both in different namespaces. This commit now implements a backward compatible change for kiwi which includes the following modifications:

* XSL stylesheet for automatic schema transformation from

  <packagemanager>dnf</packagemanager> to <packagemanager>dnf4</packagemanager>

* Code copy of dnf API interface from

  PackageManagerDnf -> PackageManagerDnf4 RepositoryDnf -> RepositoryDnf4

* Deprecation of former Dnf API interface

The code change here will force developers to adapt their code if they used RepositoryDnf / PackageManagerDnf classes in their python code. After this change developers will be dropped into a raise condition which exits kiwi at the time of the call. Related to Issue #2300
and Issue #2262

### DIFF
--- a/doc/source/api/kiwi.package_manager.rst
+++ b/doc/source/api/kiwi.package_manager.rst
@@ -14,10 +14,10 @@ Submodules
     :undoc-members:
     :show-inheritance:
 
-`kiwi.package_manager.dnf` Module
----------------------------------
+`kiwi.package_manager.dnf4` Module
+----------------------------------
 
-.. automodule:: kiwi.package_manager.dnf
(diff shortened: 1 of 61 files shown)


schaefi:
This one is blocked by #2302 and will also create an expected spec file conflict once #2302 is merged

schaefi:
@Conan-Kudo ok open for a review now

PR #2313, opened 2023-06-14
Fixed image install of dnf package manager

schaefi:
If the package manager is set to 'dnf4', this name is taken also as package name to install this package manager into the image. Problem is 'dnf4' is not a valid package name. The correct name for 'dnf4' is 'dnf'. This commit fixes the name bug introduced with the changes for Issue #2262

### DIFF
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -498,7 +498,12 @@ class XMLState:
             for package in package_list:
                 result.append(package.package_section.get_name().strip())
             if self.get_system_packages():
-                result.append(self.get_package_manager())
+                package_manager_name = self.get_package_manager()
+                if package_manager_name == 'dnf4':
+                    # The package name for dnf4 is just dnf. Thus
+                    # the name must be adapted in this case
+                    package_manager_name = 'dnf'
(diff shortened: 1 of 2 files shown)


schaefi:
@Conan-Kudo I overlooked that one but the integration tests exposed the problem. Any image builds using ```dnf``` get the packagemanager name translated to ```dnf4``` which also results in this name to be added to the package list. But there is no ```dnf4``` package which causes the build to fail with

```
[  100s] [ INFO    ]: Processing: [                                        ] 0%[ DEBUG   ]: 17:10:07 | bootstrap: No match for argument: dnf4
```

So we did the name cleanup correctly on the code side of things but I forgot that on the packaging side the old dnf package stays as just dnf.

Could you review this one quickly please, because I want to turn the integration test to a working state asap

Thanks

schaefi:
@Conan-Kudo I'm going forward with this one now as I'd like to see the integration tests working again to be sure we did not introduce regressions. If you disagree on the change here we can revert/re-design it later but I think it's inevitable unless Fedora changes on the packaging side for the "old" dnf which I assume is not going to happen.

PR #2317, opened 2023-07-03
Alp disk encryption

schaefi:
Add improvements for fully encrypted disk images
    
This commit adds more flexibility when using LUKS encrypted images. In addition support for key derivation on user provided passwords has been added. The following new attributes were added to the schema:
  
* type.luks_randomize
   allow to enable/disable the data randomizing if luks  encryption is used
    
* type.luks_pbkdf
    When LUKS unlocks a key slot using a user provided  password, it uses a so-called key derivation function to derive a symmetric encryption key from the password. Not all boot loaders support all KDF algorithms, hence  this attribute can be used to select a specific algorithm
    
* bootloader.use_disk_password
   When /boot is encrypted, make the boot loader store the  password in its configuration file (in cleartext). This is useful for full disk encryption images

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -124,6 +124,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             Defaults.get_volume_id()
         self.install_volid = self.xml_state.build_type.get_volid() or \
             Defaults.get_install_volume_id()
+        self.use_disk_password = self.xml_state.get_build_type_bootloader_use_disk_password()
 
         self.live_boot_options = [
             'root=live:CDLABEL={0}'.format(self.volume_id),
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -124,7 +124,8 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             Defaults.get_volume_id()
         self.install_volid = self.xml_state.build_type.get_volid() or \
             Defaults.get_install_volume_id()
-        self.use_disk_password = self.xml_state.get_build_type_bootloader_use_disk_password()
+        self.use_disk_password = \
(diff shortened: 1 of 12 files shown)


dracut/modules.d/99kiwi-lib/kiwi-luks-lib.sh
```diff
@@ -37,6 +44,10 @@ function activate_luks {
             --key-file /dev/zero \
             --keyfile-size 32 \
         luksOpen "${device}" luks
+    elif luks_has_generated_systemd_unit; then
+        systemctl start systemd-cryptsetup@luks
+    elif [ -f /.root.keyfile ]; then
+        cryptsetup --key-file /.root.keyfile luksOpen "${device}" luks
```
schaefi, line 50: As mentioned in the PR before, it would be good if we can get rid of that part

schaefi:
@Olaf / @Jiri we also need to bring back your names to the commits. At the moment this looks like an effort done by me but that would be very much wrong. So don't let us forget to put the right names for the Kudos in the end :)

okirch:
I pushed the updated patch set to the alp_disk_encryption branch. Not all CI checks have completed yet, so I cannot call it "green" yet. However, from my end, I'd be fine with merging these changes now as they are.

schaefi:
Looks really good now :+1: 

I'll ask @davidcassany for a short review so we get another quick look at the code and then I'll merge.

@okirch Thanks much for your effort, much appreciated

jiri:
@schaefi Please avoid mentioning people not related to an issue in the future.

schaefi:
> @schaefi Please avoid mentioning people not related to an issue in the future.

@jir Sorry that was my mistake. I wanted to mention another "Jiri" who is related to the issue but mistype the github name. Just ignore me and sorry for any inconvenience

okirch:
Thank you for your support and help in getting these merged!

Olaf

PR #2318, opened 2023-07-04
Add alternative re-reading of the partition table

schaefi:
To inform the kernel about disk geometry changes, kiwi uses partprobe as a primary tool. However it is provided by parted and not necessarily available due to the package requirements on the dracut module. A second attempt via blockdev which is expected to exist is therefore made by this commit

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -22,7 +22,11 @@ function create_partitions {
             create_dasd_partitions "${disk_device}" "${partition_setup}"
         ;;
     esac
-    partprobe "${disk_device}"
+    if type partprobe &> /dev/null;then
+        partprobe "${disk_device}"
+    else
+        blockdev --rereadpt "${disk_device}"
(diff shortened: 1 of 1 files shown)


the-luddite:
Thanks, looks good but can we guarantee that blockdev is always available? Would it be better to fail with clear error message when both blockdev and parted not available for some reason?

schaefi:
> Thanks, looks good but can we guarantee that blockdev is always available? Would it be better to fail with clear error message when both blockdev and parted not available for some reason?

blockdev is provided by util-linux and that is a required package for this dracut module. So yes the expectation is that it exists as a safe fallback.

Regarding the error message, I think the problem is "some reason". There would be no good explanation we can give why the tool is missing right ? So in the end a self written error message wouldn't be any different than the error from the calling shell that the tool is not there.

That's the reason why I did not invest into more code to log an error message

Thoughts ?

the-luddite:
@schaefi I agree that I might be too paranoid.  I was thinking util-linux can be replaced with busybox but even in this case I also can't foresee what could happen to break it.

Conan-Kudo (approved):

PR #2322, opened 2023-07-10
Load isoschematron module on demand

### LINKED ISSUE #2314: isoschematron got removed from lxml on Rawhide
## Problem description

https://src.fedoraproject.org/rpms/python-lxml/c/9d95f5a04edc386313fa854541971b3af07bcae1?branch=rawhide
https://gitlab.com/fedora/legal/fedora-license-data/-/issues/154

As seen above reasons are a license issue. However, this breaks kiwi completely

We need to find a solution. One solution is to turn the schematron validation into a dynamic loadable code point.
However, I wonder because we are using schematron rules since years and if possible I'd like to get the module back instead of hacking around it

schaefi:
Due to a license issue the isoschematron code might not be available in the lxml package from the distributor. This commit uses isoschematron on demand. If present on the system it gets loaded and used, if not a warning that the schematron validation will be skipped is provided. This Fixes #2314

### DIFF
--- a/kiwi/xml_description.py
+++ b/kiwi/xml_description.py
@@ -15,16 +15,14 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+import importlib
 from typing import (
     Dict, Any
 )
 import os
 import logging
 from xml.dom import minidom
-from lxml import (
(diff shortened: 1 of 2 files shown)


Conan-Kudo (approved):
I really hope this gets fixed upstream, but for now... LGTM.

Issue #2316, opened 2023-07-03
btrfs subvolume handling inconsistency


davide125:
## Problem description
In the Fedora Asahi Remix we use

```
<systemdisk name="fedora">
  <volume name="home"/>
  <volume name="root"/>
</systemdisk>
```

expecting to get two top-level subvolumes, `root` and `home`. Instead, kiwi creates a `@` root subvolume and nests `root` and `home` under it. Compare the layout we get:

```
ID 256 gen 494 top level 5 path @
ID 257 gen 492 top level 256 path home
ID 258 gen 487 top level 256 path root
ID 259 gen 487 top level 256 path var/lib/machines
```

with the one on a stock Anaconda-installed Fedora system

```
ID 256 gen 519138 top level 5 path home
ID 257 gen 519129 top level 5 path root
ID 262 gen 519071 top level 257 path var/lib/machines
```

## Expected behaviour
`home` and `root` are created as top-level subvolumes

## Steps to reproduce the behaviour
https://pagure.io/fedora-asahi/kiwi-descriptions/blob/rawhide/f/platforms/workstation.xml is the full config snippet we're using

## OS and Software information

* KIWI version: 9.24.59
* Operating system host version: Fedora Linux 36
* Operating system target version: Fedora Linux 38 and Rawhide
* Open Build Service version (N/A if not using OBS): N/A
* Koji version (N/A if not using Koji): N/A

Conan-Kudo:
We should not assume SUSE-style btrfs hierarchies. Does adding `mountpoint="/"` (for `root`) and `mountpoint="/home"` (for `home`) fix this?

schaefi:
> We should not assume SUSE-style btrfs hierarchies. Does adding `mountpoint="/"` (for `root`) and `mountpoint="/home"` (for `home`) fix this?

Hmm, I think you are right. Looking at ```kiwi/volume_manager/btrfs.py``` you can see

```python
def setup(self, name=None):
        """
        Setup btrfs volume management

        In case of btrfs a toplevel(@) subvolume is created and marked
        as default volume. 

        ...

        root_volume = self.mountpoint + '/@'        
```

So far there is no config option to influence this.

I think I remember at the time of the implementation for the volume layout that this was a recommendation from the btrfs people and not so much a SUSE style hierarchy. In any case to fix this without regressions we would need a new attribute

davide125:
[...]

PR #2324, opened 2023-07-12
Evaluate the @root volume name also for btrfs

schaefi:
In a volume setup the special volume declaration
```<volume name="@root=identifier"/>``` was only evaluated for the LVM volume manager. In case of btrfs a hardcoded root volume name '@' was used. This commit allows to specify a custom name for the root volume for btrfs as well and also allows to specify that there should be no such root volume. Example:

```xml
<volume name="@root=@"/>
```

Name the root volume '@'. If not specified this stays as the default to stay compatible

```xml
<volume name="@root=/"/>
```

Indicate no root volume is wanted. All subvolumes resides below root (/)

```xml
<volume name="@root=foo"/>
```

Name the root volume 'foo'

This is related to Issue #2316 and a first patch to address the requested changes

### DIFF
--- a/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
+++ b/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
@@ -30,7 +30,7 @@
         </type>
     </preferences>
     <preferences profiles="Virtual">
-        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0 rootflags=subvol=root" firmware="uefi" format="qcow2" btrfs_root_is_subvolume="true" btrfs_set_default_volume="false" rootfs_label="fedora" bootpartition="true" bootfilesystem="ext4">
+        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2" btrfs_root_is_subvolume="true" btrfs_set_default_volume="false" rootfs_label="fedora" bootpartition="true" bootfilesystem="ext4">
             <systemdisk name="fedora">
                 <volume name="@root=root"/>
                 <volume name="home" parent="/"/>
--- a/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
+++ b/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
@@ -30,7 +30,7 @@
         </type>
     </preferences>
     <preferences profiles="Virtual">
-        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2" btrfs_create_toplevel_subvolume="true" rootfs_label="fedora" bootpartition="true" bootfilesystem="ext4">
+        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0 rootflags=subvol=root" firmware="uefi" format="qcow2" btrfs_root_is_subvolume="true" btrfs_set_default_volume="false" rootfs_label="fedora" bootpartition="true" bootfilesystem="ext4">
             <systemdisk name="fedora">
                 <volume name="@root=root"/>
                 <volume name="home" parent="/"/>
--- a/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
+++ b/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
@@ -30,7 +30,11 @@
         </type>
     </preferences>
     <preferences profiles="Virtual">
-        <type image="oem" filesystem="ext4" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2">
+        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2" btrfs_create_toplevel_subvolume="true" rootfs_label="fedora">
+            <systemdisk name="fedora">
+                <volume name="@root=root"/>
+                <volume name="home" parent="/"/>
+            </systemdisk>
             <oemconfig>
                 <oem-resize>false</oem-resize>
             </oemconfig>
@@ -67,6 +71,7 @@
         <package name="vim"/>
         <package name="tzdata"/>
         <package name="NetworkManager"/>
+        <package name="btrfs-progs"/>
     </packages>
     <packages type="image" profiles="Live,Disk">
         <package name="syslinux"/>
--- a/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
+++ b/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
@@ -30,7 +30,7 @@
         </type>
     </preferences>
     <preferences profiles="Virtual">
-        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2" btrfs_create_toplevel_subvolume="true" rootfs_label="fedora">
+        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2" btrfs_create_toplevel_subvolume="true" rootfs_label="fedora" bootpartition="true" bootfilesystem="ext4">
             <systemdisk name="fedora">
                 <volume name="@root=root"/>
                 <volume name="home" parent="/"/>
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -412,6 +412,20 @@ btrfs_quota_groups="true|false":
   Boolean parameter to activate filesystem quotas if
   the filesystem is `btrfs`. By default quotas are inactive.
 
+btrfs_create_toplevel_subvolume="true|false":
+  Tell kiwi to nest all subvolumes below a toplevel volume.
+  The name of this toplevel volume is by default set to: `@`
+  The name of the toplevel volume can be changed via
+  a volume entry of the form
+
+  .. code:: xml
+
+     <systemdisk>
+       <volume name="@root=TOPLEVEL_NAME"/>
+     </systemdisk>
+
+  By default the creation of a toplevel volume is set to: `true`
+
 btrfs_root_is_snapshot="true|false":
   Boolean parameter that tells {kiwi} to install
   the system into a btrfs snapshot. The snapshot layout is compatible with
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -412,11 +412,19 @@ btrfs_quota_groups="true|false":
   Boolean parameter to activate filesystem quotas if
   the filesystem is `btrfs`. By default quotas are inactive.
 
-btrfs_create_toplevel_subvolume="true|false":
-  Tell kiwi to nest all subvolumes below a toplevel volume.
-  The name of this toplevel volume is by default set to: `@`
-  The name of the toplevel volume can be changed via
-  a volume entry of the form
+btrfs_set_default_volume="true|false":
+  Tell kiwi to explicitly make a volume the default volume
+  This can be either `/` or the root subvolume or the root
+  snapshot depending on the specified btrfs configuration
+  attributes. By default btrfs_set_default_volume is set to: true
+  If no default volume should be set, this attribute can be
+  used to turn it off
+
+btrfs_root_is_subvolume="true|false":
+  Tell kiwi to create a root volume to host (/) inside.
+  The name of this subvolume is by default set to: `@`.
+  The name of the subvolume can be changed via a volume entry
+  of the form:
 
   .. code:: xml
 
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -501,7 +501,8 @@ class BootLoaderConfigBase:
             return gfxmode
 
     def _mount_system(
-        self, root_device, boot_device, efi_device=None, volumes=None
+        self, root_device, boot_device, efi_device=None,
+        volumes=None, root_volume_name=None
     ):
         self.root_mount = MountManager(
             device=root_device
@@ -522,7 +523,10 @@ class BootLoaderConfigBase:
                 mountpoint=self.root_mount.mountpoint + '/boot/efi'
             )
 
-        self.root_mount.mount()
+        custom_root_mount_args = []
+        if root_volume_name and root_volume_name != '/':
+            custom_root_mount_args += [f'subvol={root_volume_name}']
+        self.root_mount.mount(options=custom_root_mount_args)
 
         if not self.root_mount.device == self.boot_mount.device:
             self.boot_mount.mount()
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -809,35 +809,47 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             # a grub image that got signed by the shim. The shim image
             # is the one that gets loaded by the firmware which itself
             # loads the second stage grub image
-            log.info(
-                f'--> Using shim image: {shim_image.filename}'
+            target_efi_image_name = self._get_efi_image_name()
+            target_grub_image_name = os.sep.join(
+                [self.efi_boot_path, grub_image.binaryname]
             )
-            log.info(
-                f'--> Using grub image: {grub_image.filename}'
-            )
-            Command.run(
-                ['cp', shim_image.filename, self._get_efi_image_name()]
-            )
-            Command.run(
-                [
-                    'cp', grub_image.filename,
-                    os.sep.join([self.efi_boot_path, grub_image.binaryname])
-                ]
-            )
+            if not os.path.isfile(target_efi_image_name):
+                log.info(
+                    f'--> Using shim image: {shim_image.filename}'
+                )
+                Command.run(
+                    ['cp', shim_image.filename, target_efi_image_name]
+                )
+            if not os.path.isfile(target_grub_image_name):
+                log.info(
+                    f'--> Using grub image: {grub_image.filename}'
+                )
+                Command.run(
+                    ['cp', grub_image.filename, target_grub_image_name]
+                )
             mok_manager = Defaults.get_mok_manager(lookup_path)
             if mok_manager:
-                Command.run(
-                    ['cp', mok_manager, self.efi_boot_path]
+                target_mok_manager = os.sep.join(
+                    [self.efi_boot_path, os.path.basename(mok_manager)]
                 )
+                if not os.path.isfile(target_mok_manager):
+                    log.info(
+                        f'--> Using mok image: {mok_manager}'
+                    )
+                    Command.run(
+                        ['cp', mok_manager, self.efi_boot_path]
+                    )
         else:
             # Without shim a self signed grub image is used that
             # gets loaded by the firmware
-            log.info(
-                f'--> No shim image, using grub image: {grub_image.filename}'
-            )
-            Command.run(
-                ['cp', grub_image.filename, self._get_efi_image_name()]
-            )
+            target_efi_image_name = self._get_efi_image_name()
+            if not os.path.isfile(target_efi_image_name):
+                log.info(
+                    f'--> No shim image, using grub image: {grub_image.filename}'
+                )
+                Command.run(
+                    ['cp', grub_image.filename, target_efi_image_name]
+                )
         self._create_efi_config_search(uuid, mbrid)
 
     def _setup_efi_image(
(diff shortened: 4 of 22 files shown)


Conan-Kudo:
Can you add a test case for the behavior we want in Fedora so that we can verify this does what we expect?

schaefi:
> Can you add a test case for the behavior we want in Fedora so that we can verify this does what we expect?

This is not the complete patch series we need to achieve the end result you want in Fedora. But I will add a test case that shows the volume creation without a root volume

Conan-Kudo:
Once there's a test for it, I'll be happy to approve it.

schaefi:
> Once there's a test for it, I'll be happy to approve it.

There is now a unit test available. I did some real test builds and stumbled over some weird things but it's late here and my brain doesn't work anymore... don't merge it :)

schaefi:
ok finished testing and worked for me. The unit test here is imho not enough, we also want an integration test. I'm going to setup an integration test build next

Conan-Kudo:
I'm not sure this is a good approach. Magic data can be unpredictable.

Here's my suggestion of what we should do: https://github.com/OSInside/kiwi/issues/2316#issuecomment-1636883916

Conan-Kudo:
Once we have an integration test, I'll approve and merge it.

schaefi:
> I'm not sure this is a good approach. Magic data can be unpredictable.
> 
> Here's my suggestion of what we should do: [#2316 (comment)](https://github.com/OSInside/kiwi/issues/2316#issuecomment-1636883916)

I added the mentioned attribute and deprecation message. I think it's a good idea to make it explicitly configured.
The optional parent setup I will add in a new pull request and we need an integration test for this code first.

I'm on it, but might take a bit since there are some tasks I have to complete next week first

schaefi:
Integration test setup done here:

Watch out for the ```Virtual``` profile build. Currently it's all block in obs and we need to wait.

build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
```diff
@@ -30,7 +30,12 @@
         </type>
     </preferences>
     <preferences profiles="Virtual">
-        <type image="oem" filesystem="ext4" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2">
+        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2" btrfs_create_toplevel_subvolume="false">
+            <systemdisk name="fedora">
+                <volume name="home"/>
+                <volume name="root"/>
+                <volume name="var/lib/machines"/>
+            </systemdisk>
```
Conan-Kudo, line 38: This is not correct for the test, it should be:

```suggestion
            <systemdisk name="fedora" rootfs_label="fedora">
                <volume name="home" mountpoint="/home"/>
                <volume name="root" mountpoint="/"/>
                <volume name="nspawn_machines" mountpoint="/var/lib/machines"/>
            </systemdisk>
```

Conan-Kudo (changes requested):

build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
```diff
@@ -30,7 +30,12 @@
         </type>
     </preferences>
     <preferences profiles="Virtual">
-        <type image="oem" filesystem="ext4" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2">
+        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2" btrfs_create_toplevel_subvolume="true">
+            <systemdisk name="fedora" rootfs_label="fedora">
+                <volume name="@root=root"/>
+                <volume name="home" mountpoint="/home" parent="/"/>
```
Conan-Kudo, line 36: Will this ensure that `home` is a separate top-level subvolume rather than being nested on the root subvolume?

schaefi:
@Conan-Kudo @iammattcoleman I added the parent setup and build as I hope the layout that is wanted here. Keeping all this consistent and regression free with how btrfs is used in its "default" aka suse way is really not that easy. We should have had an eye on the general implementation when it was started, but as we can't turn back the time this option does not exist. One part that will stay for the moment is how to specify the root volume. You suggested

```xml
<volume name="root" mountpoint="/"/>
```

and it's a nicer read yes but I have to leave this for backward compatibility to look like:

```xml
<volume name="@root=root"/>
```

This part I will not change.

The other suggestions were implemented the way we discussed them. I will now take a very close look to all the integration tests that we built for btrfs and see how the results look like

build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
```diff
@@ -30,7 +30,12 @@
...
+                <volume name="nspawn_machines" mountpoint="/var/lib/machines" parent="/"/>
```
Conan-Kudo, line 37:
```suggestion
                <volume name="machines" mountpoint="/var/lib/machines" parent="/var/lib"/>
```

Conan-Kudo (changes requested):

build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
schaefi: yes

schaefi: parent is '/' for nesting under the root subvolume you wold have to say parent="root" or not specify the parent

Conan-Kudo:
We should put it on the list of things to evaluate for KIWI 10 then.

build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
```diff
@@ -30,7 +30,12 @@
         </type>
     </preferences>
     <preferences profiles="Virtual">
-        <type image="oem" filesystem="ext4" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2">
+        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2" btrfs_create_toplevel_subvolume="true" rootfs_label="fedora">
+            <systemdisk name="fedora">
+                <volume name="@root=root"/>
+                <volume name="home" mountpoint="/home" parent="/"/>
+                <volume name="machines" mountpoint="/var/lib"/>
```
Conan-Kudo, line 37: I'm confused by this. Shouldn't this be `<volume name="machines" parent="/var/lib"/>`. As a nested volume, it'll automount anyway.

Conan-Kudo (changes requested):

build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
schaefi: You confuse me, you suggested that change in your review and I just adapted. So what do you want to achieve ?

schaefi: done, I kept the mountpoint though, it's easier to read and all this is confusing to no end anyway because fedora does it completely different than suse and all this should not break. No fun to be honest

schaefi:
@Conan-Kudo @iammattcoleman ok I think we are good now. The proposed changes required some refactoring of my btrfs code which was not all that good and imho still needs some love. Review by code might be not so easy but I also setup the integration test for Fedora with the desired layout (also see my last commit which I consider not a kiwi issue)
Please checkout the image from here and give it a test ```etc/fstab``` , ```btrfs subvol list /``` etc...

   Use the ```Virtual``` profile

All other integration test builds are building fine with the changes proposed here and I'm confident that it will not create regressions. Last famous words ;)

Conan-Kudo (changes requested):
The output is still wrong.

The fstab from the image is wrong:

```
UUID=67fa8e31-9d8a-4e05-9bc8-5452c3982b7f / btrfs defaults 0 1
UUID=524f35e0-ae95-4257-b8c4-a7f8bfbca55c /boot ext4 defaults 0 0
UUID=67fa8e31-9d8a-4e05-9bc8-5452c3982b7f /home btrfs defaults,subvol=home 0 0
UUID=05FA-49BA /boot/efi vfat defaults 0 0
UUID=67fa8e31-9d8a-4e05-9bc8-5452c3982b7f /var/lib/machines btrfs defaults,subvol=var/lib/machines 0 0
```

The `/` mountpoint should have a `subvol=root`, and I'm not even sure what the `/var/lib/machines` mountpoint is supposed to be.

schaefi:
> The / mountpoint should have a subvol=root

why ? the root subvolume is marked as the default (see the build log btrfs subvolume set-default. There is no need for an extra entry. Also the fstab root information is useless anyway since the mounting happens through systemd inside of the initrd

> and I'm not even sure what the /var/lib/machines mountpoint is supposed to be.

Didn't you request that as part the history of this pull request as a test for a subvolume ?

schaefi:
Before we continue further with this story. There are naturally completely different design patterns how the distros wants their volume tree to be organized, while things are easy and straight forward with just LVM they are weird in my eyes with btrfs. As you know suse has its own special layout and I think Fedora is also aiming for something different. 

I'm afraid I have to keep backward compatibility in some way for what we have. If that is even possible I'm no longer convinced according to the history of this pull request. Since I already spent days to wrap my head around this I'd like to see a sequence of btrfs commands (not in some code that I have to read first, but here in this PR) that produces the layout which is considered "correct"

From that information it will be easier to decide how to move forward. Thanks

Conan-Kudo:
> > The / mountpoint should have a subvol=root
> 
> why ? the root subvolume is marked as the default (see the build log btrfs subvolume set-default. There is no need for an extra entry. Also the fstab root information is useless anyway since the mounting happens through systemd inside of the initrd
> 

It matters because if it's not actually a subvolume of its own, then everything is nested in it. The whole point is to have separate hierarchies for system and user data so that if you take a snapshot and do system rollbacks, you're not taking user data along for the ride. More importantly, it also allows you to wipe the system data and reset it to factory by re-applying the system volume via btrfs send or resetting snapshot or whatever.

> > and I'm not even sure what the /var/lib/machines mountpoint is supposed to be.
> 
> Didn't you request that as part the history of this pull request as a test for a subvolume ?

What I was trying to accomplish with the test was to have two top level subvolumes and one nested one to ensure we tested that behavior correctly. The nested one doesn't require a mountpoint because by virtue of being nested, it'll be auto-mounted.

Conan-Kudo:
> Before we continue further with this story. There are naturally completely different design patterns how the distros wants their volume tree to be organized, while things are easy and straight forward with just LVM they are weird in my eyes with btrfs. As you know suse has its own special layout and I think Fedora is also aiming for something different.
> 

SUSE's layout was based off of how people typically set up an OS with ZFS. It worked as well as it did because for the longest time, `/home` wasn't a Btrfs subvolume, so the nesting problem wasn't an issue when it came to doing system snapshots and rollbacks.

The Fedora layout is a literal translation of the LVM layout to Btrfs, but I chose to not change it because it had advantages of its own: by using multiple top level subvolumes, you had independent hierarchies for system and user data. You could have independent snapshotting regimes for them, and you could place the snapshot directory outside of the system or user data hierarchy (they don't have to be mounted all the time).

It also had the advantage of letting you erase the system data without erasing the user data if you wanted to do reinstalls or do factory reset type behavior.

> I'm afraid I have to keep backward compatibility in some way for what we have. If that is even possible I'm no longer convinced according to the history of this pull request. Since I already spent days to wrap my head around this I'd like to see a sequence of btrfs commands (not in some code that I have to read first, but here in this PR) that produces the layout which is considered "correct"
> 
> From that information it will be easier to decide how to move forward. Thanks

I can absolutely provide you a sequence of commands that creates the Fedora layout by hand:

1. `mkfs.btrfs -L fedora /dev/loop0p4`
2. `mount /dev/loop0p4 /tmp/fedorasetup`
3. `btrfs subvolume create /tmp/fedorasetup/root`
4. `btrfs subvolume create /tmp/fedorasetup/home`

The key omission is `btrfs subvolume set-default`. By not changing the default subvolume, the hierarchy doesn't get switched around.

This means that the fstab looks like this (forgive not using UUID, I just don't have that as I write this out, in the real world we'd use the UUID):

```
/dev/loop0p4          /                       btrfs   subvol=root    0 0
/dev/loop0p4          /home                   btrfs   subvol=home    0 0
```

For the production images, I'd also set fsoptions to add `compress=zstd:1` so that everything uses compressed I/O with zstd, and that should show up when mounting the filesystem to create the image and in the fstab. But for the purposes of me writing this out, that does not matter.

schaefi:
>  The nested one doesn't require a mountpoint because by virtue of being nested, it'll be auto-mounted.

True but for tools like snapper it's needed to have them mentioned. So this for example is a part I can't change but also wouldn't hurt right ?

schaefi:
Thanks for the details, I will look into it in the next free slot, which will be some time next week. I'm traveling Fri-Sun
stay tuned :)

Conan-Kudo:
> > The nested one doesn't require a mountpoint because by virtue of being nested, it'll be auto-mounted.
> 
> True but for tools like snapper it's needed to have them mentioned. So this for example is a part I can't change but also wouldn't hurt right ?

My understanding is that snapper uses the fstab mounts to figure out how to exclude stuff for snapshotting (to work around the way SUSE setup the subvolume hierarchy), so it only hurts if you don't want snapper to exclude it for the snapshot tree. There are other reasons to create subvolumes where the automatic exclusion might hurt.

It's not my biggest concern right now, but it's something to think about in the future.

Conan-Kudo:
Do we need the flag to also control `btrfs subvolume set-default`? My understanding is that this is required for snapper (along with having a toplevel subvolume), so maybe tying that to the `btrfs_create_toplevel_subvolume` flag makes sense?

schaefi:
> Do we need the flag to also control `btrfs subvolume set-default`? My understanding is that this is required for snapper (along with having a toplevel subvolume), so maybe tying that to the `btrfs_create_toplevel_subvolume` flag makes sense?

I think you always want a default volume (btrfs does that in any case if you set it or not). We just make this an explicit step by either saying the default volume is / (btrfs's default) or it's the configured root volume or it's the snapshot if configured by the respective attribute. That sequence feels correct to me

schaefi:
@Conan-Kudo I'm done with the changes and the local built looks good to me. The obs build now fails because of the xattr changes :( I'm starting to get angry about it, one little change here and again some distro or some stupid niche case fails *grrr* Hope we can finally come to a stable result with all this

Conan-Kudo:
@schaefi Well, I've merged #2336, so we should be good for rebasing this and trying it again.

schaefi:
> @schaefi Well, I've merged #2336, so we should be good for rebasing this and trying it again.

yep I'm on it

schaefi:
@Conan-Kudo ok image are building and the one setup as the integration test for the changes introduced here is done too. From my perspective it looks good now, please double check

```
osc getbinaries Virtualization:Appliances:Images:Testing_x86:fedora test-image-live-disk images x86_64 -M Virtual
kvm-cpu Broadwell-v2 -m 4096 -serial stdio kiwi-test-image-live-disk.x86_64-2.0.0-Virtual-Build77.2.qcow2

root: linux

btrfs subvol list /
mount | grep btrfs
```

Thanks

schaefi:
> The fstab looks right, but can we make it possible so that `btrfs subvolume set-default` doesn't get used during image setup? It's tricky to mount the actual parent volume to the two subvolumes because we do that. My suggestion is that we don't run that command if `btrfs_create_toplevel_subvolume` is set to `false`.

Hmm, I think we shouldn't connect that implicitly. Looking into it

schaefi:
@Conan-Kudo ok I have added an attribute that allows you to configure if kiwi should set a default volume or not. The integration test has this switched off as you wanted.

With the last commit I fixed the places in kiwi which assumed there is a default volume set. With that patch the integration will also start to build again

However, there is now another problem at boot of the test image (I did built it here locally). Not unexpected the image fails to boot and you land in a rescue shell inside of the initrd. As far as I can see the initrd simply tries to mount the root filesystem. Without any default volume configured this of course is not the correct volume. Because btrfs uses (/)
As the requested layout of the system is this:

```
UUID=860cc59c-936f-4c11-ab4a-c4dffdb88fb4 / btrfs defaults,subvol=root 0 1
UUID=860cc59c-936f-4c11-ab4a-c4dffdb88fb4 /home btrfs defaults,subvol=home 0 0
```

you have the ```root``` volume and the ```home``` volume and none of them are default when mounting.
I know at suse they have patched grub to deal with this situation and there is something like
````set btrfs_relative_path="yes"``` in the grub early boot script and the grub package at suse.

Brings me to the question if you don't want any default, how can I make this Fedora37 based system to boot again ?
I don't consider this a kiwi thing and my gut feeling tells me that it will not be easy to find a proper and generic solution to it.

Please advise

**Comment by Conan-Kudo** [comment]
Now we need to propagate `rootflags=subvol=root` to the bootloader config.

For example, this is what one of my installed Fedora systems has:

```
[root@Belldandy-Slimbook ~]# cat /boot/loader/entries/1c0b343ff0414c459df6e3b4aafe72f2-6.3.9-200.fc38.x86_64.conf
title Fedora Linux (6.3.9-200.fc38.x86_64) 38 (KDE Plasma)
version 6.3.9-200.fc38.x86_64
linux /vmlinuz-6.3.9-200.fc38.x86_64
initrd /initramfs-6.3.9-200.fc38.x86_64.img
options root=UUID=e69f9fa5-1641-4369-be7f-4a56a86ca597 ro rootflags=subvol=root resume=UUID=13169969-3d8d-4980-ba78-8e8f79a0138b rhgb quiet
grub_users $grub_users
grub_arg --unrestricted
grub_class fedora
```

**Comment by schaefi** [comment]
Ah ok, that should be easy we already have a section to specify custom bootloader config settings.... I'll jump on it later as I have to do another task now

stay tuned... again :)

**Comment by schaefi** [comment]
ok works, we are getting closer :)

**Comment by Conan-Kudo** [comment]
Are we in a reviewable/testable state now?

**Comment by schaefi** [comment]
> Are we in a reviewable/testable state now?

yes we are, the integration test is currently building. We should be good now. All other builds still succeed and we are backward compatible

File: build-tests/x86/fedora/test-image-live-disk/appliance.kiwi [hunk 5]
```diff
@@ -30,7 +30,11 @@
         </type>
     </preferences>
     <preferences profiles="Virtual">
-        <type image="oem" filesystem="ext4" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2">
+        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0 rootflags=subvol=root" firmware="uefi" format="qcow2" btrfs_root_is_subvolume="true" btrfs_set_default_volume="false" rootfs_label="fedora" bootpartition="true" bootfilesystem="ext4">
```
Conan-Kudo, line 33: `rootflags=subvol=root` should probably be automatic as a consequence of `btrfs_root_is_subvolume="true"`. Requiring that it be specified manually is probably going bite people down the road.

Conan-Kudo (changes requested):

build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
schaefi: Hmm, but it feels distro specific. Setting ```rootflags=subvol=root``` on suse as an example would be useless information. Or is this somewhat standardized ?

schaefi: I'd like to prevent implicit magic unless it would be generally valid.

Conan-Kudo: That is absolutely standard. You would be propagating the `subvol=<rootsubvolname>` thing you already are doing for fstab to the kernel command line, just with the prefix `rootflags=`. By definition, this is not implicit magic like `btrfs subvolume set-default`, since you're explicitly telling it where to look.

Conan-Kudo: Basically, if you don't do `btrfs subvolume set-default` to reorient the filesystem for the root tree, you need to tell the kernel where the root tree is.

schaefi: ok that makes sense to me. I will adapt and it should also not hurt.

schaefi: added a commit to solve this, we need to check on the integration test results though

schaefi:
@Conan-Kudo waiting for all tests to pass...

schaefi:
> @Conan-Kudo waiting for all tests to pass...

looks good

PR #2326, opened 2023-07-18
Add SECURE_BOOT no when the firmware is efi

jesusbv:
This Fixes bsc#1211102


schaefi (approved):
Looks good to me :+1: Thanks

Conan-Kudo (approved):
Fortunately Fedora doesn't use `/etc/sysconfig/bootloader`, so I think we're fine.

PR #2329, opened 2023-07-20
Fix failure when booting KIS via NBD

Glidos:
In the NBD booting case, the function initGlobalDevices in kiwi-overlay-root.sh runs the command ndb-client to connect the remote NBD share to the device /dev/nbd0. The script parse-kiwi-overlay.sh, on the other hand, incorrectly parses overlay:nbd=ip:export to /dev/ip/export, leading to the system waiting for /dev/ip/export to appear. This commit corrects the problem.

With this change NDB booting of KIS images now succeeds, although only if the ndb package is added to the appliance description.


schaefi:
yeah good catch :+1: 

as we explicitly set ```read_only_partition="/dev/nbd0"``` the direct device assignment is good. Thanks much

PR #2330, opened 2023-07-20
Delete (py)xattr module requirement

schaefi:
The required functionality is provided by the core os library


Conan-Kudo (approved):

Conan-Kudo:
We already require at least Python 3.6, so we're good. Merged!

PR #2332, opened 2023-07-24
Fixed writing keyfile path to etc/crypttab

schaefi:
The keyfile path was not correctly set in etc/crypttab which caused systemd not being able to read the keyfile, consequently asking for a passphrase. This commit fixes the writing of the crypttab and also fixes a python name clash with the "os" namespace.


Conan-Kudo (approved):

ignatenkobrain (approved):
fixes my problem

PR #2337, opened 2023-07-27
Use 'source' instead of 'configfile' in generated grub.cfg

### LINKED ISSUE #2320: Grub menu problem with live isos in EFI mode
## Problem description

When I press Esc in the top level grub menu of leap/tumbleweed live isos booted in EFI mode, grub shell screen is shown without an evident way to get back to the menu

## Expected behaviour

Esc in the top level grub menu in EFI mode should do nothing, like when the same iso is booted in BIOS mode.

- create a new EFI vm
- attach a leap/tumbleweed live iso to it. I tried leap 15.5 and tumbleweed 20230705
- power the vm up, wait for grub menu to appear
- press Esc
- try to get back to the menu like press Esc once more

Repeat it in a BIOS vm and notice there is no such problem.

* KIWI version: the problem supposedly dates back to this commit https://github.com/OSInside/kiwi/commit/3255264d4c57678c2f6cd5b1243de340e76230ed
* Operating system host/target version: irrelevant

## Research

The origin of the problem are these lines:

https://github.com/OSInside/kiwi/blob/16a703cd98b8c543bb5e2848e7b5203275472a09/kiwi/bootloader/config/grub2.py#L1078

https://github.com/OSInside/kiwi/blob/16a703cd98b8c543bb5e2848e7b5203275472a09/kiwi/bootloader/config/grub2.py#L1099

Was there any particular reason to use `configfile` instead of `source` there? With `source` the problem goes away: the top level menu is not nested into some scope, there is nowhere to return with Esc, hence it does nothing.

shim-install also creates such transit grub config and it uses `source`:

```
428 cat <<EOF
429 search --fs-uuid --set=root ${cfg_fs_uuid}
430 set prefix=(\${root})`${grub_mkrelpath} ${grub_cfg_dirname}`
431 source "\${prefix}/${grub_cfg_basename}"
432 EOF
```

slowpeek:
Fixes #2320

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -1077,7 +1077,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 )
             )
             early_boot.write(
-                'configfile ($root){0}/{1}/grub.cfg{2}'.format(
+                'source ($root){0}/{1}/grub.cfg{2}'.format(
                     self.get_boot_path(), self.boot_directory_name, os.linesep
                 )
             )
@@ -1098,7 +1098,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 )
             )
             early_boot.write(
-                'configfile ($root)/boot/{0}/grub.cfg{1}'.format(
+                'source ($root)/boot/{0}/grub.cfg{1}'.format(
                     self.boot_directory_name, os.linesep
                 )
             )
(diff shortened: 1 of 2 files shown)


slowpeek:
I see the problems with the tests. Fixing it ..

slowpeek:
The checks should pass this time but I rebased one extra commit, will fix it

slowpeek:
Looks good now

schaefi:
yep looks good, I'd like to have @davidcassany and @Conan-Kudo to take a look at this one too. Thanks

Conan-Kudo (approved):
Seems fine to me.

marcan:
FYI, this creates a timebomb with Fedora/RH installs that explodes on the next GRUB update: https://bugzilla.redhat.com/show_bug.cgi?id=2235692

I don't consider this a Kiwi bug, but you should be aware since I expect more than just Asahi Linux users are affected.

PR #2341, opened 2023-08-01
add GRUB_TIMEOUT_STYLE=menu option

leifliddy:
According to the grub standard
https://www.gnu.org/software/grub/manual/grub/html_node/Simple-configuration.html

`GRUB_TIMEOUT_STYLE`
If this option is unset or set to ‘menu’, then GRUB will display the menu and then wait for the timeout set by ‘GRUB_TIMEOUT’ to expire before booting the default entry. Pressing a key interrupts the timeout.
  

So `GRUB_TIMEOUT_STYLE='menu'` is literally the default option that grub uses, but for some (inexplicable) reason it's not an option that's supported by this project ; )

**# Fixes**
Add **menu** as a supported `GRUB_TIMEOUT_STYLE` option

Note:
I always remove trailing spaces before committing -- and there seemed to be a lot trailing spaces present in the files that were modified.


Conan-Kudo:
Can you please separate your trailing space change into another commit?

leifliddy:
Sure

kiwi/schema/kiwi.rnc
```diff
@@ -2794,7 +2794,7 @@ div {
         ## expired. In countdown mode an indication of the remaining time
         ## is displayed. The attribute is available for the grub loader.
         attribute timeout_style {
-            "countdown" | "hidden"
+            "countdown" | "hidden" | "menu"
```
ericcurtin, line 2797: One of the non-whitespace changes is here (for reviewers)

leifliddy: I separated the original commit into two commits. One for the actual (non-whitespace) change and another commit for only the whitespace changes (which admittedly isn't that important)
This commit doesn't contain whitespace changes.
https://github.com/OSInside/kiwi/pull/2341/commits/7627f2c677052dfe7ab00b3a213dc99c125a120b

ericcurtin (approved):

schaefi (approved):
The kiwi.rng file is an autogenerated one, so I did not look at it. Thanks for splitting the space changes into an extra commit. I wasn't aware that the .rnc schema file has so many white space issues. Thanks for fixing it and the actual change to the schema looks good to me :+1:

PR #2342, opened 2023-08-01
RPM spec: add typing extension

dcermak:


package/python-kiwi-spec-template
```diff
@@ -374,6 +374,7 @@ Requires:       python%{python3_pkgversion}-docopt
 Requires:       python%{python3_pkgversion}-lxml
 Requires:       python%{python3_pkgversion}-requests
 Requires:       python%{python3_pkgversion}-setuptools
+Requires:       (python%{python3_pkgversion}-typing_extensions if python%{python3_pkgversion} < 3.8)
```
Conan-Kudo, line 377: The package name is `python3-typing-extensions` in RH/Fedora and Debian/Ubuntu.

And conditional dependencies don't work on Debian/Ubuntu at all, so this needs to be handled correctly for that too.

Conan-Kudo (changes requested):

Conan-Kudo (approved):

dcermak:
> @dcermak still in draft mode ?

To prevent a premature merge :sweat_smile:

Is it good on OBS?

schaefi:
> To prevent a premature merge 

:) we usually set the "do-not-merge" flag to indicate this

> Is it good on OBS?

still running, stay tuned I will respond back

Thanks

schaefi:
All green again, thanks a ton for fixing this

Issue #2261, opened 2023-03-18
Not setting `console=` still sets `GRUB_TERMINAL=` when it is unwanted


Conan-Kudo:
## Problem description

I need to be able to prepare images without _any_ `GRUB_TERMINAL` setting, but currently if I don't set `console=`, it gets set to `gfxterm`. Instead, I want GRUB's own defaults to kick in because using `console` causes garbage rendering, and `gfxterm` doesn't exist on RH/Fedora apparently.

## OS and Software information

* KIWI version: 9.24.52
* Operating system host version: Fedora Linux 37
* Operating system target version: Fedora Linux 38
* Open Build Service version (N/A if not using OBS): N/A 
* Koji version (N/A if not using Koji): N/A

leifliddy:
To be clear, the garbage rendering issue exists on Apple Silicon systems. 
hmmm, what if we just allowed console to contains an empty string ie `console=""`
And then modified the patterns to allow an empty string as a value ie 
```
validate_grub_console_patterns_ = [['^(\s*|console|gfxterm|serial)( (console|gfxterm|serial))*$']]
...
<param name="pattern">(\s*|console|gfxterm|serial)( (console|gfxterm|serial))*</param>
```
The end result would be that `/etc/default/grub` contained `GRUB_TERMINAL=""`
Would that be an acceptable solution. If so, I could sort that out no problem. 
Another option would be to devise a mechanism in which grub directives could be omitted entirely. 
But that would likely take quite a bit of work as we'd probably need to devise a special keyword like 
`console="omit"` and then match and act on that logic.

Conan-Kudo:
> hmmm....what if we just allowed console to contains an empty string ie `console=""`

That is reasonable.

PR #2346, opened 2023-08-07
allow value of GRUB_TERMINAL to be empty

leifliddy:
Fixes # 
https://github.com/OSInside/kiwi/issues/2261

Allow value of GRUB_TERMINAL to be an empty string
(As of this moment) setting GRUB_TERMINAL to console causes garbage rendering on Apple Silicon systems. So GRUB_TERMINAL needs to contain a blank value (or not be set at all) if you want a properly rendered grub boot menu on Apple Silicon systems.

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -37,7 +37,7 @@ vhd-tag-type = xsd:token {pattern = "[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}"}
 groups-list = xsd:token {pattern = "[a-zA-Z0-9_\-\.:]+(,[a-zA-Z0-9_\-\.:]+)*"}
 arch-name = xsd:token {pattern = "(x86_64|i586|i686|ix86|aarch64|arm64|armv5el|armv5tel|armv6hl|armv6l|armv7hl|armv7l|ppc|ppc64|ppc64le|s390|s390x|riscv64)(,(x86_64|i586|i686|ix86|aarch64|arm64|armv5el|armv5tel|armv6hl|armv6l|armv7hl|armv7l|ppc|ppc64|ppc64le|s390|s390x|riscv64))*"}
 portnum-type = xsd:token {pattern = "(\d+|\d+/(udp|tcp))"}
-grub_console = xsd:token {pattern = "(\s*|console|gfxterm|serial)( (console|gfxterm|serial))*"}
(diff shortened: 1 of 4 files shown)


kiwi/schema/kiwi.rnc
```diff
@@ -37,7 +37,7 @@ vhd-tag-type = xsd:token {pattern = "[0-9a-f]{8}-([0-9a-f]{4}-){3}[0-9a-f]{12}"}
 groups-list = xsd:token {pattern = "[a-zA-Z0-9_\-\.:]+(,[a-zA-Z0-9_\-\.:]+)*"}
 arch-name = xsd:token {pattern = "(x86_64|i586|i686|ix86|aarch64|arm64|armv5el|armv5tel|armv6hl|armv6l|armv7hl|armv7l|ppc|ppc64|ppc64le|s390|s390x|riscv64)(,(x86_64|i586|i686|ix86|aarch64|arm64|armv5el|armv5tel|armv6hl|armv6l|armv7hl|armv7l|ppc|ppc64|ppc64le|s390|s390x|riscv64))*"}
 portnum-type = xsd:token {pattern = "(\d+|\d+/(udp|tcp))"}
-grub_console = xsd:token {pattern = "(console|gfxterm|serial)( (console|gfxterm|serial))*"}
+grub_console = xsd:token {pattern = "(\s*|console|gfxterm|serial)( (console|gfxterm|serial))*"}
```
schaefi, line 40: Could we make this more explicit by providing a ```none``` console type ? Something like this

```
grub_console = xsd:token {pattern = "(none|console|gfxterm|serial)( (console|gfxterm|serial))*"}
```

It imho makes it more clear what is expected to happen with the console setting and could also be
found better in the documentation. 

Talking about the docs, could you add some words about this in
```doc/source/image_description/elements.rst```

Thanks much

schaefi (changes requested):
Thanks, I'm good with the change. I would just propose that we come up with a name for setting no console and some doc hint about it.

kiwi/schema/kiwi.rnc
schaefi: @leifliddy any thoughts ? Thanks

leifliddy: Yup, that makes perfect sense. Sorry, I was on a motorcycle trip over the last several days. I'll try and make these changes tonight.

schaefi: excellent thanks much :+1:

schaefi (approved):
looks good to me :+1: Thanks

PR #2348, opened 2023-08-10
Fix bsc#1209302: add some hints about KIWI extensions and dependencies for SLES

tomschr:
SUSE Linux Enterprise Server (SLES) has KIWI as a package. However, some KIWI extensions and their dependencies can be unavailable.

This commit introduces a new section which deals with this aspect.

Fixes bsc#1209302.

* Add new section about "Installation for SUSE Linux Enterprise Server"
* Describe the general issue when using KIWI extensions on SLES.
  Keep in mind, this was deliberately kept general to fit all extensions.

### DIFF
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -96,10 +96,10 @@ Fedora/Rawhide:
 
 
 
-Installation for SUSE Linux Enterprise Server
----------------------------------------------
+Installation for SUSE Linux Enterprise
+--------------------------------------
 
-{kiwi} is available and supported for SUSE Linux Enterprise Server (SLES).
+{kiwi} is available and supported for SUSE Linux Enterprise (SLE).
 The recommended and supported way is to install {kiwi} by using zypper.
 
 However, if you rely on some plugins for {kiwi}, either the plugin
@@ -107,13 +107,13 @@ itself or any dependencies might not be available for your service pack.
 
 If you want to proceed anyway, keep these things in mind:
 
-* Plugins that are not provided by SLES are not supported.
+* Plugins that are not provided by SLE are not supported.
 * You probably need to install dependencies via :command:`pip`.
   The :command:`pip` command installs these dependencies from PyPI
   (the Python Package Index).
   However, this approach will not update the RPM database.
 * Depending on your security concerns, installing Python packages
-  outside the secured SLES installation may not be desirable.
+  outside the secured SLE installation may not be desirable.
 * Python packages installed from PyPI won't contain any SUSE
   customizations.
 * Depending on your extension and its dependencies, you might even need
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -94,6 +94,32 @@ Fedora/Rawhide:
 
      $ sudo dnf install kiwi-cli
 
+
+
+Installation for SUSE Linux Enterprise Server
+---------------------------------------------
+
+{kiwi} is available and supported for SUSE Linux Enterprise Server (SLES).
+The recommended and supported way is to install {kiwi} by using zypper.
+
+However, if your rely on some extensions for {kiwi}, either the extension
+itself or any dependencies might not be available for your service pack.
+
+If you want to proceed anyway, keep these things in mind:
+
+* Extensions that are not provided by SLES are not supported.
+* You probably need to install dependencies via :command:`pip`.
+  The :command:`pip` command installs these dependencies from PyPI
+  (the Python Package Index).
+  However, this approach will not update the RPM database.
+* Depending on your security concerns, installing Python packages
+  outside the secured SLES installation may not be desirable.
+* Python packages installed from PyPI won't contain any SUSE
+  customizations.
+* Depending on your extension and its dependencies, you might even need
+  a more recent Python version.
+
+
 .. _example-descriptions:
 
 Example Appliance Descriptions
--- a/doc/source/installation.rst
+++ b/doc/source/installation.rst
@@ -102,12 +102,12 @@ Installation for SUSE Linux Enterprise Server
 {kiwi} is available and supported for SUSE Linux Enterprise Server (SLES).
 The recommended and supported way is to install {kiwi} by using zypper.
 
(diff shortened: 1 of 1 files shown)


mairda:
I believe that "However, if your rely on some extensions for..." is not valid English. due to the way "your" is used in-place. I assume a more equitable version is of a form similar to: "However, if your attempted installation shows reliance on some extensions for...". I'm sorry I can't find a way to say it as a reviewer.

tomschr:
> I believe that "However, if your rely on some extensions for..." is not valid English. due to the way "your" is used in-place.

Yes, thanks @mairda. That was a typo. Should be "you" instead of "you":

    However, if you rely on some extensions for {kiwi}, either the extension
    itself or any dependencies might not be available for your service pack.

doc/source/installation.rst
```diff
@@ -94,6 +94,32 @@ Fedora/Rawhide:
...
+
+
+Installation for SUSE Linux Enterprise Server
+---------------------------------------------
+
+{kiwi} is available and supported for SUSE Linux Enterprise Server (SLES).
+The recommended and supported way is to install {kiwi} by using zypper.
+
+However, if your rely on some extensions for {kiwi}, either the extension
+itself or any dependencies might not be available for your service pack.
```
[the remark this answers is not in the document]
tomschr: Ahh, good catch! I used the term from the Bugzilla entry. It makes absolutely sense to use the term "plugin" if you have used in other parts of the documentation.

I'll adapt it.

doc/source/installation.rst
```diff
@@ -94,6 +94,32 @@ Fedora/Rawhide:
...
+* Extensions that are not provided by SLES are not supported.
+* You probably need to install dependencies via :command:`pip`.
+  The :command:`pip` command installs these dependencies from PyPI
+  (the Python Package Index).
+  However, this approach will not update the RPM database.
+* Depending on your security concerns, installing Python packages
+  outside the secured SLES installation may not be desirable.
+* Python packages installed from PyPI won't contain any SUSE
+  customizations.
+* Depending on your extension and its dependencies, you might even need
```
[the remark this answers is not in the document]
tomschr: Yes, I was torn whether to make it general enough or based on a specific example.

I've decided to go for the more general approach because I got the impression from the Buzilla entry that these plugins/dependencies might be added/supported in the future. Maybe I got the wrong impression. Are we sure enough that this specific example is still correct in the future? That's why I took a more general approach and didn't base it on a specific example.

Of course we can add it if you think, this would be a safe bet.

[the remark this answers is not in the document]
schaefi: ok that makes sense to me, so let's leave it more generic. I also agree with you regarding the conversation on slack that with a specific example it might be broken if the plugin changes. As the plugin is in another git project than kiwi and this docu it might happen that there is a disconnect. Even though I think the boxbuild plugin is pretty mature and is not expected to change a lot other than bugfixes and box descriptions.

schaefi (approved):
looks good to me

tomschr:
If anything else needs to be corrected, let me know.

mairda:
> If anything else needs to be corrected, let me know.

Ultra pedantic and not things I demand, but...

* I believe python.org calls them "modules" rather than "extensions"
* pip is python version specific, there are probably multiple versions of python and pip on a given host mostly from the python 3.8 through python 3.11 version range. On Tumbleweed /bin/pip is a link to alts that selects the highest installed version to run. It may require more care for a specific purpose, e.g. pip-<version> if KIWI is configured to operate on another version of python than the latest installed.

doc/source/installation.rst
```diff
@@ -94,6 +94,32 @@ Fedora/Rawhide:
 
      $ sudo dnf install kiwi-cli
 
+
+
+Installation for SUSE Linux Enterprise Server
```
Conan-Kudo, line 99: `SUSE Linux Enterprise Server` -> `SUSE Linux Enterprise`

(this applies to the whole SLE family)

doc/source/installation.rst
```diff
@@ -94,6 +94,32 @@ Fedora/Rawhide:
...
+---------------------------------------------
+
+{kiwi} is available and supported for SUSE Linux Enterprise Server (SLES).
```
Conan-Kudo, line 102: `SUSE Linux Enterprise Server (SLES)` -> `SUSE Linux Enterprise (SLE)`

doc/source/installation.rst
```diff
@@ -94,6 +94,32 @@ Fedora/Rawhide:
...
+The recommended and supported way is to install {kiwi} by using zypper.
+
+However, if you rely on some plugins for {kiwi}, either the plugin
+itself or any dependencies might not be available for your service pack.
+
+If you want to proceed anyway, keep these things in mind:
+
+* Plugins that are not provided by SLES are not supported.
```
Conan-Kudo, line 110: `SLES` -> `SLE`

doc/source/installation.rst
```diff
@@ -94,6 +94,32 @@ Fedora/Rawhide:
...
+* You probably need to install dependencies via :command:`pip`.
+  The :command:`pip` command installs these dependencies from PyPI
+  (the Python Package Index).
+  However, this approach will not update the RPM database.
+* Depending on your security concerns, installing Python packages
+  outside the secured SLES installation may not be desirable.
```
Conan-Kudo, line 116: `SLES` -> `SLE`

Conan-Kudo (changes requested):

doc/source/installation.rst
tomschr: Fixed in commit 31b599c71.

doc/source/installation.rst
tomschr: Fixed in commit 31b599c71.

doc/source/installation.rst
tomschr: Fixed in commit 31b599c71.

doc/source/installation.rst
tomschr: Fixed in commit 31b599c71.

Conan-Kudo (approved):

PR #2349, opened 2023-08-10
Update GitHub Actions "actions/*" part

tomschr:
Update to the latest versions:

* [actions/checkout@v3](https://github.com/actions/checkout)
* [action/setup-python@v4](https://github.com/actions/setup-python
* [action/cache@v3](https://github.com/actions/cache)


schaefi (approved):
Thanks good catch and looks good to me :+1: 

Any reason why it's still in draft mode ?

tomschr:
> Any reason why it's still in draft mode ?

Just in case you had any objections, comments, or any other wishes.

PR #2352, opened 2023-08-15
Add kiwi_live_volid variable to profile

schaefi:
As part of the profile environment only the kiwi_install_volid variable existed with the default value 'INSTALL'. This information is correct when building an oem image with installiso set to true. The resulting .install.iso image has the configured volid or the mentioned default 'INSTALL'. However, when building a live ISO image the default volid value is set to 'CDROM' but the profile environment contains kiwi_install_volid which is misleading and in case no volid is provided in the configuration file, the default value would also be a wrong information. This commit cleans up the mess and provides kiwi_install_volid for install ISO media and kiwi_live_volid for Live ISO media with their respective default values

### DIFF
--- a/kiwi/system/profile.py
+++ b/kiwi/system/profile.py
@@ -184,10 +184,14 @@ class Profile:
         # kiwi_install_volid
         if self.xml_state.is_xen_server():
             self.dot_profile['kiwi_xendomain'] = 'dom0'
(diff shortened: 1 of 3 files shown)


davidcassany (approved):
Wow, good catch! :+1:

Conan-Kudo (approved):

PR #2357, opened 2023-08-22
Make verity metadata always readonly

MofX:
If a partition is verity protected, it can never be writable, since verity is by-definition read-only.


schaefi (approved):
Thanks good catch, looks good to me :+1:

Conan-Kudo (approved):

PR #2359, opened 2023-08-23
Fixed mounting of subvolumes for data sync

### LINKED ISSUE #2356: Incorrect subvolume configuration with btrfs
## Problem description

In the Fedora Asahi Remix we're seeing an issue with `/var` where the `var` subvolume itself is empty and the contents are installed under `/var` in the `root` subvolume itself.

See `root.img` in https://fedora-asahi-remix.org/os/fedora-38-server-202308191708.zip (or https://fedora-asahi-remix.org/os/fedora-38-server-202308191708.raw.zst if you prefer the full raw image). The build logs for that are at https://fedora-asahi-remix.org/os/fedora-38-server-202308191708.logs.zip . You can see the config we used at https://pagure.io/fedora-asahi/kiwi-descriptions/blob/a37f7758c2d9ecc108935456520632f97fdac587/f/platforms/server.xml which specifically does

```
<systemdisk name="fedora">
    <volume name="@root=root"/>
    <volume name="home" parent="/"/>
    <volume name="var" parent="/"/>
</systemdisk>
```

## Expected behaviour
The `/var` subvolume should be populated correctly.

Doing a build with the same config should repro this.

* KIWI version: 9.25.12
* Operating system host version: Fedora Cloud 38
* Operating system target version: Fedora Linux 38

schaefi:
The subvolumes must be mounted below the root of the filesystem tree such that the sync mechanism correctly shuffles the data to the correct volumes.
This Fixes #2356

### DIFF
--- a/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
+++ b/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
@@ -34,6 +34,7 @@
             <systemdisk name="fedora">
                 <volume name="@root=root"/>
                 <volume name="home" parent="/"/>
+                <volume name="var" parent="/"/>
             </systemdisk>
             <oemconfig>
                 <oem-resize>false</oem-resize>
--- a/kiwi/volume_manager/btrfs.py
+++ b/kiwi/volume_manager/btrfs.py
@@ -229,6 +229,7 @@ class VolumeManagerBtrfs(VolumeManagerBase):
                 )
 
                 volume_mountpoint = toplevel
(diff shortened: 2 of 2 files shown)


schaefi:
Integration test runs here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:fedora/test-image-live-disk use the Virtual profile results for testing

Conan-Kudo (approved):
The output looks good from the integration test and comparison of logs from before and after.

PR #2362, opened 2023-08-31
PKGBUILD: Drop unused Python dependencies

### LINKED ISSUE #2361: Useless python-future in dependencies?
Hello. I see that the project depends on future but I cannot find a single usage of it in the codebase.

https://github.com/OSInside/kiwi/blob/9c2c10b3de8d007dbd8190e70ae5610d71526ad4/package/python-kiwi-pkgbuild-template#L24

Do you think it's still useful? I'm asking because future is broken with the latest Python 3.12 and will be even more with 3.13 so it might be a good idea to get rid of it.

Conan-Kudo:
We haven't used some of these dependencies in years...

Fixes #2361.

### DIFF
--- a/package/python-kiwi-pkgbuild-template
+++ b/package/python-kiwi-pkgbuild-template
@@ -21,7 +21,7 @@ build() {
 }
 
 package_python-kiwi(){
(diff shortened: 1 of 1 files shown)


frenzymadness (approved):

schaefi:
looks good to me too, but @davidcassany did the pacman package and I'd like to wait for his approval

davidcassany (approved):
Looks good to me too, thanks!

PR #2363, opened 2023-08-31
Drop hybrid boot snippets from the GRUB 2 configuration template

### LINKED ISSUE #2354: Kiwi's override of the Grub commands when booting with EFI mode don't work in Grub in Ubuntu 23.10
## Problem description

Kiwi adds the following snippet to the beginning of `grub.cfg`:

```
set linux=linux
set initrd=initrd
if [ "${grub_cpu}" = "x86_64" -o "${grub_cpu}" = "i386" ]; then
    if [ "${grub_platform}" = "efi" ]; then
        set linux=linuxefi
        set initrd=initrdefi
    fi
fi
export linux initrd
```

This overrides the commands used for loading the initrd and linux binaries when booting in EFI mode. This worked in previous versions of Ubuntu, but since the upgrade to Grub 2.12 this command has been dropped in Ubuntu (compare the list of files: https://packages.ubuntu.com/jammy/amd64/grub-efi-amd64-bin/filelist lists `linuxefi.mod`, while https://packages.ubuntu.com/mantic/amd64/grub-efi-amd64-bin/filelist does not have this file).

This snippet is generated by the following file: https://github.com/OSInside/kiwi/blob/master/kiwi/bootloader/template/grub2.py#L52

This causes the system to fail to boot with the following error:

```
error: can't find command `linuxefi'.
error: can't find command `initrdefi'.
```

As a workaround, the following script can be added to `root/etc/grub.d/01_fix_linuxefi` to override this behaviour and set the `linux` and `initrd` variables:

```
#!/bin/sh

echo "set linux=linux"
echo "set initrd=initrd"
```

After adding this to the image, the system boots without any issues in EFI mode.

## Expected behaviour

The system should boot in EFI mode using Grub as the bootloader.

Generate an EFI-based image using Ubuntu 23.10 and Grub 2.12.

* KIWI version:  9.25.11
* Operating system host version: Ubuntu 22.04
* Operating system target version: Ubuntu 23.10

Conan-Kudo:
Sometime between GRUB 2.04 and GRUB 2.06, it became no longer necessary to use `linuxefi`+`initrdefi` for UEFI boot. The standard `linux`+`initrd` stanzas work for both legacy BIOS boot and modern UEFI boot.

Some distributions no longer support `linuxefi`+`initrdefi` at all anymore, so let's just use `linux`+`initrd` for everything now.

Fixes #2354.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -678,8 +678,6 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         * SUSE_REMOVE_LINUX_ROOT_PARAM
         * GRUB_BACKGROUND
         * GRUB_THEME
-        * GRUB_USE_LINUXEFI
-        * GRUB_USE_INITRDEFI
         * GRUB_SERIAL_COMMAND
         * GRUB_CMDLINE_LINUX
         * GRUB_CMDLINE_LINUX_DEFAULT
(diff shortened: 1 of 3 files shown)


schaefi:
yeah let's get rid of that linux vs linuxefi stuff in kiwi, it caused more harm than cure

The tests however need some love. Let me know if you want me to take a closer look

Thanks

schaefi:
@Conan-Kudo I'm closing this one as it has conflicts and needs to be re-opened against main.

PR #2364, opened 2023-09-01
use os.path.join to create path

leifliddy:
`self.get_boot_path()` can potentially return a '/'

so when these functions run 
```
            early_boot.write(
                'set prefix=($root){0}{1}'.format(
                    os.path.join(self.get_boot_path(), self.boot_directory_name), os.linesep
                )
            )
            early_boot.write(
                '{0} ($root){1}/grub.cfg{2}'.format(self.grub_load,
                    os.path.join(self.get_boot_path(), self.boot_directory_name),
                    os.linesep
                )
```
....they could potentially return paths that contains double slashes `//`
```
set prefix=($root)//grub2
source ($root)//grub2/grub.cfg
```

So this PR utilizes `os.path.join` to prevent this from happening. 
I understand that `//` in paths are still functional -- but it's just not aesthetically pleasing to look at. Feel free to drop this PR if you feel it's too trivial.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -984,7 +984,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 '-O', Defaults.get_efi_module_directory_name(self.arch),
                 '-o', root_efi_image,
                 '-c', root_efi_path + 'earlyboot.cfg',
-                '-p', self.get_boot_path() + '/' + self.boot_directory_name,
+                '-p', os.path.join(self.get_boot_path(), self.boot_directory_name),
                 '-d', module_path.replace(self.root_dir, '')
             ] + module_list
         )
@@ -1097,15 +1097,14 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 'search --fs-uuid --set=root {0}{1}'.format(uuid, os.linesep)
             )
             early_boot.write(
-                'set prefix=($root){0}/{1}{2}'.format(
-                    self.get_boot_path(), self.boot_directory_name, os.linesep
+                'set prefix=($root){0}{1}'.format(
+                    os.path.join(self.get_boot_path(), self.boot_directory_name), os.linesep
                 )
             )
             early_boot.write(
-                '{0} ($root){1}/{2}/grub.cfg{3}'.format(
-                    self.grub_load, self.get_boot_path(),
-                    self.boot_directory_name, os.linesep
-                )
+                '{0} ($root){1}/grub.cfg{2}'.format(self.grub_load, os.path.join(
+                    self.get_boot_path(), self.boot_directory_name
+                ), os.linesep)
             )
 
     def _create_early_boot_script_for_mbrid_search(self, filename, mbrid):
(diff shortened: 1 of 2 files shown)


leifliddy:
I did, but I'm not sure exactly how it wants me to ident those lines.

ericcurtin:
If we suffer from leaning toothpick syndrome, one leading '/' is what's normally there. But we probably want at least one leading slash.

leifliddy:
So with change in place you'll see 
```
set btrfs_relative_path="yes"
search --fs-uuid --set=root be96b30d-d153-42c1-8225-6d8236e8e767
set prefix=($root)/grub2
configfile ($root)/grub2/grub.cfg
```
vs
```
set btrfs_relative_path="yes"
search --fs-uuid --set=root be96b30d-d153-42c1-8225-6d8236e8e767
set prefix=($root)//grub2
configfile ($root)//grub2/grub.cfg
```
in `/boot/efi/EFI/BOOT/grub.cfg` (or `/boot/efi/EFI/fedora/grub.cfg` on asahi linux)

schaefi:
The change looks good to me, thanks much. Actually the test checks for the use of ```//``` and I can't remember why it was done that way. Looks wrong to me in code and test. So could you also fix the unit test in ```test/unit/bootloader/config/grub2_test.py``` and there the test case for ```test_setup_disk_boot_images_bios_plus_efi``` searching for the the double slash points you directly to the code in question.

For running the unit tests locally you can also call

```
tox -e unit_py3_11
```

if your host has python 3.11, choose another interpreter that matches your host if not 3.11

Thanks much

leifliddy:
Should be good now.

Conan-Kudo:
Style check failed. Can you look to resolve those?

leifliddy:
The line appears to be  indented correctly. I'm not sure where that error is coming from -- or what I need to do to resolve it.

schaefi:
> The line appears to be indented correctly. I'm not sure where that error is coming from -- or what I need to do to resolve it.

This was caused by calling os.path.join() inside of a format() without proper indentation. You can fix this as follows

```diff
diff --git a/kiwi/bootloader/config/grub2.py b/kiwi/bootloader/config/grub2.py
index baffe92f..d7b8bb6d 100644
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -1102,9 +1102,9 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 )
             )
             early_boot.write(
-                '{0} ($root){1}/grub.cfg{2}'.format(self.grub_load,
-                    os.path.join(self.get_boot_path(), self.boot_directory_name), os.linesep
-                )
+                '{0} ($root){1}/grub.cfg{2}'.format(self.grub_load, os.path.join(
+                    self.get_boot_path(), self.boot_directory_name
+                ), os.linesep)
             )
 
     def _create_early_boot_script_for_mbrid_search(self, filename, mbrid):
```

schaefi:
@leifliddy there are more ways to proper indent, I just suggested one, feel free to adapt as it becomes a good read for you. Thanks much

Conan-Kudo:
Can you squash everything into a single commit with a detailed commit message?

leifliddy:
Is everyone happy with this? All requested changes were made. There's another PR I'd like to submit after this one.

Conan-Kudo (approved):

PR #2367, opened 2023-09-11
Fixed root entry in fstab for btrfs layout

### LINKED ISSUE #2366: Transactional system images have wrong subvol= set in /etc/fstab
## Problem description
The subvol for `/` as indicated by `/etc/fstab` is wrong:

```
localhost:~ # cat /etc/fstab 
UUID=879f672c-585d-4742-84ab-8f94bba1980c / btrfs ro,defaults,subvol=@/.snapshots/1/snapshot 0 0
                                                                     ^^^^^^^^^^^^^^^^^^^^^^^
UUID=879f672c-585d-4742-84ab-8f94bba1980c /.snapshots btrfs defaults,subvol=@/.snapshots 0 0
UUID=879f672c-585d-4742-84ab-8f94bba1980c /home btrfs defaults,subvol=@/home 0 0
...
localhost:~ # findmnt /
TARGET
  SOURCE                              FSTYPE OPTIONS
/ /dev/vda3[/@/.snapshots/3/snapshot] btrfs  ro,relatime,seclabel,space_cache=v2,subvolid=270,subvol=/@/.snapshots/3/snapshot
```

This is probably caused by f8f1c0056e711f9d7ceef7b6057768a29d63af91.

## Expected behaviour

There should not be any `subvol=` set for the `/` mount. The default subvol indicates what `/` is.
(This isn't actually fully true if a "Boot into read-only snapshot" option is used, but hardcoding the first snapshot is almost always wrong).

* KIWI version: 9.25.12
* Operating system host version: openSUSE Tumbleweed
* Operating system target version: openSUSE Tumbleweed
* Open Build Service version (N/A if not using OBS): build.opensuse.org

schaefi:
A root (/) entry in fstab for btrfs is only required if no default volume is configured. This commit adapts the code path which adds an fstab entry for (/) to be called only if btrfs_set_default_volume is set to false.
This Fixes #2366

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -413,12 +413,16 @@ btrfs_quota_groups="true|false":
   the filesystem is `btrfs`. By default quotas are inactive.
 
 btrfs_set_default_volume="true|false":
-  Tell kiwi to explicitly make a volume the default volume
-  This can be either `/` or the root subvolume or the root
-  snapshot depending on the specified btrfs configuration
-  attributes. By default btrfs_set_default_volume is set to: true
-  If no default volume should be set, this attribute can be
-  used to turn it off
+  For oem disk images using the btrfs filesystem, requests to
+  set a default volume for the rootfs which is used when the
+  filesystem gets mounted. In case a `true` value is provided or
+  the attribute is not specified at all, kiwi will make a volume
+  the default volume. This can be either `/` or the configured
+  root subvolume or the configured root snapshot. In addition
+  no entry for the rootfs will be added to the `/etc/fstab` file.
+  In case a `false` value is provided, kiwi will not set any
+  default volume which also means an entry for the rootfs is
+  required and will be placed to the `/etc/fstab` file.
 
 btrfs_root_is_subvolume="true|false":
   Tell kiwi to create a root volume to host (/) inside.
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -148,8 +148,8 @@ class DiskBuilder:
         self.target_removable = xml_state.build_type.get_target_removable()
         self.root_filesystem_is_multipath = \
             xml_state.get_oemconfig_oem_multipath_scan()
-        self.btrfs_set_default_volume = \
-            xml_state.build_type.get_btrfs_set_default_volume()
+        self.btrfs_default_volume_requested = \
+            xml_state.btrfs_default_volume_requested()
         self.oem_systemsize = xml_state.get_oemconfig_oem_systemsize()
         self.oem_resize = xml_state.get_oemconfig_oem_resize()
         self.disk_resize_requested = \
@@ -422,8 +422,8 @@ class DiskBuilder:
                 'root_is_subvolume':
                     self.xml_state.build_type.
                     get_btrfs_root_is_subvolume(),
-                'set_default_volume':
-                    self.btrfs_set_default_volume,
+                'btrfs_default_volume_requested':
+                    self.btrfs_default_volume_requested,
                 'quota_groups':
                     self.xml_state.build_type.get_btrfs_quota_groups(),
                 'resize_on_boot':
@@ -1125,7 +1125,8 @@ class DiskBuilder:
             custom_root_mount_args += ['ro']
             fs_check_interval = '0 0'
 
-        if self.volume_manager_name and self.volume_manager_name == 'btrfs':
+        if self.volume_manager_name and self.volume_manager_name == 'btrfs' \
+           and not self.btrfs_default_volume_requested:
             root_volume_name = system.get_root_volume_name()
             if root_volume_name != '/':
                 custom_root_mount_args += [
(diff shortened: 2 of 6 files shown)


Conan-Kudo (approved):

Vogtinator:
What's the default value of `btrfs_set_default_volume`? According to the documentation it's "true" but according to the code it's `None`, i.e. effectively true in this context?

Conan-Kudo:
> What's the default value of `btrfs_set_default_volume`? According to the documentation it's "true" but according to the code it's `None`, i.e. effectively true in this context?

Yes. If it is unset, it's considered true for now.

Vogtinator:
Can that be changed? The combination `btrfs_root_is_readonly_snapshot="true"` + `btrfs_set_default_volume="true"` currently doesn't work.

Otherwise this breaks images which don't have `btrfs_set_default_volume="false"` explicitly set, but also can't set it without explicitly depending on a newer schemaversion and thus kiwi version.

Conan-Kudo:
> Can that be changed? The combination `btrfs_root_is_readonly_snapshot="true"` + `btrfs_set_default_volume="true"` currently doesn't work.
> 
> Otherwise this breaks images which don't have `btrfs_set_default_volume="false"` explicitly set, but also can't set it without explicitly depending on a newer schemaversion and thus kiwi version.

This was the behavior before the introduction of the `btrfs_set_default_volume` flag, so I'm not sure how this could break things more than it already was.

That said, I may have misspoke about this. `None` is a _third_ state, so we need to handle it too.

kiwi/builder/disk.py
```diff
@@ -1125,7 +1125,8 @@ def _write_generic_fstab(
             custom_root_mount_args += ['ro']
             fs_check_interval = '0 0'
 
-        if self.volume_manager_name and self.volume_manager_name == 'btrfs':
+        if self.volume_manager_name and self.volume_manager_name == 'btrfs' \
+           and self.btrfs_set_default_volume is False:
```
Conan-Kudo, line 1129: We need to handle the unset case too.

Conan-Kudo (changes requested):

Vogtinator:
> > Can that be changed? The combination `btrfs_root_is_readonly_snapshot="true"` + `btrfs_set_default_volume="true"` currently doesn't work.
> > Otherwise this breaks images which don't have `btrfs_set_default_volume="false"` explicitly set, but also can't set it without explicitly depending on a newer schemaversion and thus kiwi version.
> 
> This was the behavior before the introduction of the `btrfs_set_default_volume` flag, so I'm not sure how this could break things more than it already was.

Before https://github.com/OSInside/kiwi/commit/f8f1c0056e711f9d7ceef7b6057768a29d63af91, `/` never had any `subvol=` option in fstab.

> That said, I may have misspoke about this. `None` is a _third_ state, so we need to handle it too.

Yeah. Currently both `is True` and `is False` tests are false, which is a bit confusing.

schaefi:
Hmm, I'm confused:

* btrfs_set_default_volume="false": is the False case
* btrfs_set_default_volume="true": is the True case
* no attribute set at all: is the (None) case handled as True case to stay compatible with the state before the attribute existed

So an actual action only happens in case of the explicit setting of "false"

Thus where do you see an error ?

kiwi/builder/disk.py
schaefi: why ? None is handled as True -> default volume is wanted and set and should not appear in fstab

Vogtinator:
I think I got a bit confused about the `btrfs_set_default_volume` option.

> Tell kiwi to explicitly make a volume the default volume This can be either / or the root subvolume or the root snapshot depending on the specified btrfs configuration attributes.

IIUC, with `true`, it uses `btrfs subvolume set-default` and does not set `subvol=` but with `false` it sets `subvol=` instead? i.e. a more suitable name for that option would be `btrfs_volume_for_root_is_default`. In that case this PR looks ok.

schaefi:
> IIUC, with `true`, it uses `btrfs subvolume set-default` and does not set `subvol=` but with `false` it sets `subvol=` instead? i.e. a more suitable name for that option would be `btrfs_volume_for_root_is_default`. In that case this PR looks ok.

ok so the confusion seems to be related to the name of the attribute. Changing the name would be an expensive move as we would need a schema change and a transformation stylesheet and also a schema version change because it would be an incompatible change. Could we agree or improve on the documentation side of things ? The attribute is documented here:

https://github.com/OSInside/kiwi/blob/master/doc/source/image_description/elements.rst

search for "set_default"

I think we can improve the description here to make it more clear ?

schaefi:
ok, I'll adapt :)

schaefi:
I adapted the code to be hopefully a better read and also adapted the documentation. How about it now ?

Conan-Kudo (approved):

doc/source/image_description/elements.rst
```diff
@@ -413,12 +413,16 @@ btrfs_quota_groups="true|false":
...
-  attributes. By default btrfs_set_default_volume is set to: true
-  If no default volume should be set, this attribute can be
-  used to turn it off
+  For oem disk images using the btrfs filesystem, requests to
+  set a default volume for the rootfs which is used when the
+  filesystem gets mounted. In case a `true` value is provided or
+  the attribute is not specified at all, kiwi will make a volume
+  the default volume. This can be either `/` or the configured
+  root subvolume or the configured root snapshot. In addition
+  no entry for the rootfs will be added to the `/etc/fstab` file.
```
Vogtinator, line 422: This is wrong: An entry will be added, just without `subvol=`

schaefi: grmbl, yes my fault, sorry. Going to create a followup PR

PR #2373, opened 2023-09-13
Fix typo in 'image results' doc section

radolin:
I seems to me there is a typo and it should be installpxe instead of installiso in the second instance.


schaefi:
yes good catch, thanks :+1:

PR #2375, opened 2023-09-13
Support TOML markup

### LINKED ISSUE #2372: Consider adding TOML as a supported description format
Currently, kiwi supports JSON, XML, and YAML for the description format. There are various trade-offs with the different formats, but one thing all three suffer from is various issues with being human editable. YAML has "interesting" quirks and JSON and XML are painful due to strictness of structure (and lack of comment support in JSON).

The new description format coming into vogue is [TOML](https://toml.io/). It's an INI-style format that in Python is often internally represented as dictionaries (similar to JSON and YAML). It is considerably easier to read and write, and the way it handles data types is way less messy than YAML.

Starting with Python 3.11, a TOML reader is part of the standard library. That said, a perfectly good module for reading and modifying TOML is [`tomlkit`](https://pypi.org/project/tomlkit/) (which was created by the Poetry developers and used by Poetry itself).

schaefi:
Allow to use TOML markup for the kiwi image description This Fixes #2372

### DIFF
--- a/.virtualenv.requirements.txt
+++ b/.virtualenv.requirements.txt
@@ -11,6 +11,9 @@ lxml
 # YAML file parsing
 PyYAML
 
+# TOML file parsing
+toml
+
 # Python wrapper for extended filesystem attributes
 xattr==0.9.3
 
--- a/kiwi/markup/any.py
+++ b/kiwi/markup/any.py
@@ -70,6 +70,25 @@ class MarkupAny(MarkupBase):
             self.description_markup_processed.name
         )
 
+    def get_toml_description(self) -> str:
+        """
+        Return TOML description file name
+
+        :return: file path name
+
+        :rtype: str
+        """
+        xml_description_xslt_transformed = self.apply_xslt_stylesheets(
+            self.description_markup_processed.name
+        )
+        markup = self.anymarkup.parse_file(
+            xml_description_xslt_transformed, force_types=None
+        )
+        self.anymarkup.serialize_file(
+            markup, xml_description_xslt_transformed, format='toml'
+        )
+        return xml_description_xslt_transformed
+
     def get_yaml_description(self) -> str:
         """
         Return YAML description file name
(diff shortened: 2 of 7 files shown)


schaefi:
To convert your image description to TOML call

```
kiwi-ng image info --description path/to/description --print-toml
```

schaefi:
I did not add support for TOML so far because it adds a new python module requirement as you know. 
I don't think this change here is backward compatible, but supporting more markups is easy :)

.virtualenv.requirements.txt
```diff
@@ -11,6 +11,9 @@ lxml
 # YAML file parsing
 PyYAML
 
+# TOML file parsing
+toml
```
Conan-Kudo, line 15: This module is dead. It was split into [`tomllib` in the standard library](https://docs.python.org/3/library/tomllib.html) for reading, and `tomli-w` for writing.

If we want a singular module for this, `tomlkit` is a better choice, and it's style preserving too.

Conan-Kudo (changes requested):

.virtualenv.requirements.txt
schaefi: It's there because anymarkup uses it as its backend.

schaefi: kiwi depends on anymarkup because it can be feed with any markup and can translate both sides. So if you want tomlkit rather than toml it should be added to anymarkup not kiwi

Conan-Kudo: That's a good idea.

schaefi: I opened an issue upstream, let's see how they feel about it

schaefi:
@Conan-Kudo I haven't heard back from the anymarkup people and guess they are not interested. I would continue adding the toml support but implement it as a plugin such that we don't create a hard requirement to the toml module. Thoughts ?

schaefi:
> I'm concerned that `anymarkup` might be abandoned. If that's the case, we're going to have to take ownership of that module.

It's a different question. From my perspective it's not an orphan module but there is not much development happening anymore and it just does it's job in a good way. If need be we can certainly take ownership. From today's perspective and dev power in the kiwi team I would say it will not get more attention that it gets now ;)

Brings me back to the question if we should support the TOML markup in plugin style to prevent kiwi from another requirement or if we should not add TOML markup support at all ? I don't have any pressing requests for TOML but as we have YAML and coding that as a plugin is easy why not do it ?

Conan-Kudo (approved):

Issue #2370, opened 2023-09-12
Replace setuptools with poetry


Conan-Kudo:
Setuptools is long in the tooth and our workflows with it are rather painful. Let's move to [poetry](https://python-poetry.org/) so that we can simplify our dev, test, and packaging workflows.

Poetry also has the added advantage of providing backwards compatibility for distributions that don't support it (Debian/Ubuntu mainly) in the built sdist tarballs by generating a `setup.py` for use. So we get the best of both worlds here!

Conan-Kudo:
This is actually something I'm already kind of working on and experimenting with.

Issue #2369, opened 2023-09-12
Reduce supported distributions to 2024-era options


Conan-Kudo:
Our supported distribution matrix is so wide that it's becoming difficult for us to reason the code around a lot of core functions.

In the community meeting today, we discussed supporting the latest distributions available in 2024. That basically means:

- CentOS Stream 9+
- Fedora Linux 40+
- SUSE ALP
- Debian 12+
- Ubuntu 24.04+

schaefi:
I also like to add the software components to match with the distros:

* bootloader
   - isolinux (should be imho dropped)
   - grub2 (should be supported only in latest version, get rid of all the linuxefi stuff for example)
   - EFI vs UEFI (there should be only one EFI path, secure boot preferred over non secure boot)
   - MBR boot (should this still be supported)

* python version
   - decide on the min python version

* selinux version and tools
   - what is the right way to use selinux and in which version

* container toolkit, umoci, skopeo, buildah
   - min version for these tools ?

* get rid of tool calls which differentiate on parameter capabilities, search for CommandCapabilities and delete no longer needed checks according to the above mentioned distro list

.... there is more

Conan-Kudo:
For the Python version, I think setting our minimum version to Python 3.9 should be fine.

schaefi:
Status update:

- [ ] grub2 (should be supported only in latest version, get rid of all the linuxefi stuff for example) [done in #2424]
- [ ] MBR boot (should this still be supported) [yes stays supported]
- [ ] For the Python version, I think setting our minimum version to Python 3.9 should be fine, done [3.9+ in workflows on main]

Conan-Kudo:
I will be doing this final transition as part of #2370.

schaefi:
> Well, actually, there's still also dropping syslinux/isolinux support...

done in #2439

PR #2376, opened 2023-09-13
Switch to pyproject.toml

schaefi:
setup.py is marked obsolete for a long time and with this commit pyproject.toml is introduced and used for the development environment setup.

To be more clear, calling setup.py as a tool is deprecated:

* https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#what-commands-should-be-used-instead

The changes here refactors kiwi to no longer call setup.py in any way

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -60,6 +60,11 @@ BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildRequires:  gcc
 BuildRequires:  python%{python3_pkgversion}-%{develsuffix} >= 3.6
 BuildRequires:  python%{python3_pkgversion}-setuptools
+%if 0%{?fedora} || 0%{?suse_version} >= 1650 || 0%{?debian} || 0%{?ubuntu}
+BuildRequires:  python%{python3_pkgversion}-installer
+BuildRequires:  python%{python3_pkgversion}-build
+BuildRequires:  python%{python3_pkgversion}-wheel
+%endif
 %if 0%{?fedora} || 0%{?suse_version}
 BuildRequires:  fdupes
 %endif
@@ -68,6 +73,23 @@ BuildRequires:  shadow
 %endif
 %if 0%{?debian} || 0%{?ubuntu}
 BuildRequires:  passwd
+%endif
+%if 0%{?ubuntu} || 0%{?debian}
+BuildRequires:       python%{python3_pkgversion}-yaml
+BuildRequires:       python%{python3_pkgversion}-typing-extensions
+%else
+BuildRequires:       python%{python3_pkgversion}-PyYAML
+%endif
+BuildRequires:       python%{python3_pkgversion}-simplejson
+BuildRequires:       python%{python3_pkgversion}-docopt
+BuildRequires:       python%{python3_pkgversion}-lxml
+BuildRequires:       python%{python3_pkgversion}-requests
+BuildRequires:       python%{python3_pkgversion}-setuptools
+%if 0%{?rhel} || 0%{?fedora}
+BuildRequires:       (python%{python3_pkgversion}-typing-extensions if python%{python3_pkgversion} < 3.8)
+%endif
+%if 0%{?suse_version}
+BuildRequires:       (python%{python3_pkgversion}-typing_extensions if python%{python3_pkgversion} < 3.8)
 %endif
 
 %description
@@ -625,9 +647,13 @@ sed -e "s|#!/usr/bin/env python||" -i kiwi/xml_parse.py
 # Build C-Tools
 make CFLAGS="${RPM_OPT_FLAGS}" tools
 
+# Build wheel, skip on failure
+python3 -m build --no-isolation --wheel || true
+
 %install
-# Install Python 3 version
-python3 setup.py install --prefix=%{_prefix} --root=%{buildroot} %{?is_deb:--install-layout=deb}
+# Install Python 3 version from wheel, use setup.py as fallback
+python3 -m installer --destdir %{buildroot} dist/*.whl || \
+    python3 setup.py install --prefix=%{_prefix} --root=%{buildroot} %{?is_deb:--install-layout=deb}
 
 # Install C-Tools, man-pages, completion and kiwi default configuration
 make buildroot=%{buildroot}/ install
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -6,3 +6,5 @@ tag = True
 [bumpversion:file:kiwi/version.py]
 
 [bumpversion:file:doc/source/conf.py]
+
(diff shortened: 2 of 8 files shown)


schaefi:
@Conan-Kudo I'm pro switching to poetry. I just opened this PR to show a bit of my experiments with moving away from setup.py and the issues I had and still have with it

schaefi:
Related to #2370

schaefi:
Added more cleanup tasks:

* Makefile is free from setup.py usage now. For building the sdist tarball for the package build we now use ```python -m build``` which of course requires the python ```build``` module to be present on the host. ```pip install build```
* The upload to pypi happens now as a wheel and no longer sdist, this is done by tox calling ```python -m build``` for the release target which defaults to wheel

schaefi:
setup.py now only holds settings that cannot (or I did not understand how to add them) be used in pyproject.toml. These settings are mainly scripts and entry-points and the license. See the dynamic section

```toml
dynamic = [
    "license",
    "entry-points",
    "scripts"
]
```

ideas welcome

schaefi:
setup.py is currently only actively called in the .spec file for installing

```
python3 setup.py install --prefix=%{_prefix} --root=%{buildroot} %{?is_deb:--install-layout=deb}
```

If we want to change this it would mean:

* switch to wheel and call ```python installer the-wheel```
* This creates a dependency to the installer module which I'm not sure exists for all distros
* We could delete the requirement for setuptools consequently
* Our source tarball would be a wheel instead of a tarball. This causes the problem that we add other data e.g the man pages to the source tarball and I don't know how to modify a wheel

schaefi:
@Conan-Kudo I made some progress here, if you have a free minute would be nice if you can take a look at the changeset.

schaefi:
@Conan-Kudo I finished my changes regarding setup.py. I think it is in a review-able state now. My main concern is the change on the .spec file level. I'm not sure if ```BuildRequires:  python%{python3_pkgversion}-installer``` resolves correctly on all distros we support. Thus I still have the no merge label active

package/python-kiwi-spec-template
```diff
@@ -56,10 +56,11 @@ Packager:       Marcus Schaefer <marcus.schaefer@suse.com>
 Group:          %{pygroup}
 Source:         %{name}.tar.gz
 Source1:        %{name}-rpmlintrc
+Source2:        kiwi-%{version}-py3-none-any.whl
```
Conan-Kudo, line 59: We definitely should not be including wheels as sources. We should be constructing them in the spec file.

package/python-kiwi-spec-template
```diff
@@ -56,10 +56,11 @@ Packager:       Marcus Schaefer <marcus.schaefer@suse.com>
...
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 BuildRequires:  gcc
 BuildRequires:  python%{python3_pkgversion}-%{develsuffix} >= 3.6
-BuildRequires:  python%{python3_pkgversion}-setuptools
+BuildRequires:  python%{python3_pkgversion}-installer
```
Conan-Kudo, line 63: This is not right, we want building from source, not installing a binary blob. So we would need both.

package/python-kiwi-spec-template
```diff
@@ -617,17 +618,13 @@ Provides manual pages to describe the kiwi commands
 %prep
 %setup -q -n kiwi-%{version}
 
-# Drop shebang for kiwi/xml_parse.py, as we don't intend to use it
-# as an independent script
-sed -e "s|#!/usr/bin/env python||" -i kiwi/xml_parse.py
```
Conan-Kudo, line 622: This should remain.

package/python-kiwi-spec-template
```diff
@@ -617,17 +618,13 @@ Provides manual pages to describe the kiwi commands
...
-
 %build
 # Build C-Tools
 make CFLAGS="${RPM_OPT_FLAGS}" tools
 
 %install
-# Install Python 3 version
-python3 setup.py install --prefix=%{_prefix} --root=%{buildroot} %{?is_deb:--install-layout=deb}
+# Install Python 3 version from wheel
+python3 -m installer --destdir %{buildroot} %{SOURCE2}
```
Conan-Kudo, line 627: What we're missing here is a build step from source. In the Fedora spec, I have the following:

```
%build
# Because there are some compiled stuff
%set_build_flags

%py3_build

# Build man pages
make -C doc man

# Build C-Tools
make CFLAGS="%{build_cflags}" tools

%install
%py3_install

# Install C-Tools, man-pages, completion and kiwi default configuration (yes, the slash is needed!)
make buildroot=%{buildroot}/ install

# Install dracut modules (yes, the slash is needed!)
make buildroot=%{buildroot}/ install_dracut

[...]
```

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: ok I see... more changes needed. Thanks

package/python-kiwi-spec-template

package/python-kiwi-spec-template
schaefi: yes was removed during my testing, should have never been dropped

package/python-kiwi-spec-template
schaefi: Thanks I did so without using the macros as they differ between the distros

Conan-Kudo:
We cannot merge this until we drop support for old distributions as stated in #2369. Because otherwise this will be broken for half of our supported platforms.

schaefi:
> We cannot merge this until we drop support for old distributions as stated in #2369. Because otherwise this will be broken for half of our supported platforms.

I thought with the last commit there should be a safe fallback for these cases, no ?

Conan-Kudo:
You ripped all the data out of `setup.py`, so now we need something to process `pyproject.toml` to build properly.

Conan-Kudo:
You could try moving the data into a setup.cfg file and tell pyproject.toml that we use setuptools to build. I think that's backward compatible.

schaefi:
> You ripped all the data out of `setup.py`, so now we need something to process `pyproject.toml` to build properly.

I might be missing all details but to my understanding only the direct call of setup.py as a command is obsolete. Meaning setuptools and the use of pyproject.toml in combination with setup.py is still a supported combination. With the current changeset I tried building for a distro that has build and installer python modules available and I also tried building against SLE15 which is py3.6 and there are not out of the box build and installer modules available. The spec tries to create the wheel and fails and as fallback continues with "setup.py install" as we had it before and that worked without any issues. Thus I wonder where the changes here breaks the backward compatibility. I admit that I haven't looked into all details and after effects.

I will close this PR and re-open it against the kiwi-10 branch. I think there we definitely need to move away from setup.py

PR #2378, opened 2023-09-18
Add missing dependency to simplejson

adrianschroeter:
kiwi crashed on start without it in factory builds


schaefi:
Thanks looks good to me :+1:

schaefi:
@Conan-Kudo do you see any negative side effects ?

schaefi:
Hmm, on a second look, there is already a requirement set to ```python%{python3_pkgversion}-simplejson```

Conan-Kudo (changes requested):
This is already present in the spec file.

adrianschroeter:
ah, sorry, a fellow co-maintainer messed this up in OBS:Server:Unstable

PR #2379, opened 2023-09-22
Fix incorrect prefix setting in BLS grub menu entries

schaefi:
The generated /boot/loader/entries/*.conf files contain a prefixed /boot dir of the form:
```
    linux /boot/...
    initrd /boot/...
```

However, this is incorrect if an extra /boot partition is specified. This PR resolves that -- and will create

```
    linux /...
    initrd /...
```

if ```bootpartition="true"``` partition is specified.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -1459,11 +1459,18 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         for menu_entry_file in glob.iglob(loader_entries_pattern):
             with open(menu_entry_file) as grub_menu_entry_file:
                 menu_entry = grub_menu_entry_file.read()
-                menu_entry = re.sub(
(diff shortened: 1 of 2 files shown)


schaefi:
@leifliddy could you please double check if I made any mistake ?

Also please let me know if you prefer a reference to your name in the commit message, because the actual code change is yours not mine ;) 

Thanks much

Conan-Kudo:
@schaefi I would prefer that you'd preserve @leifliddy's attribution.

schaefi:
> @schaefi I would prefer that you'd preserve @leifliddy's attribution.

yeah me too that's why I asked, what would be the best way to do this ? @leifliddy already closed the PR that he opened

schaefi:
Ok I think I correctly preserved the work done by @leifliddy and added my changes on top

Conan-Kudo (approved):

PR #2383, opened 2023-10-16
Fix ramdisk install

schaefi:
Fixed argument processing of config.bootoptions
    
 Instead of adding an extra space to make the subsequent reading to work, use an awk script that does it without nasty workarounds. This Fixes the ramdisk deployment

### DIFF
--- a/dracut/modules.d/90kiwi-dump/kiwi-ramdisk-deployment-generator.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-ramdisk-deployment-generator.sh
@@ -26,16 +26,12 @@ GENERATOR_DIR="$1"
 
 [ -e /config.bootoptions ] || exit 1
 
-# Add a space to /config.bootoptions to make sure the
-# following token based read captures all entries
-echo -n ' ' >> /config.bootoptions
-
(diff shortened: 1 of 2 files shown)


dracut/modules.d/90kiwi-dump/kiwi-ramdisk-deployment-generator.sh
```diff
@@ -26,16 +26,12 @@ GENERATOR_DIR="$1"
...
-echo -n ' ' >> /config.bootoptions
-
 root_uuid=$(
-    while read -r -d ' ' opt; do echo "${opt}";done < /config.bootoptions |\
+    awk '{ for(i=1; i <= NF; i++) {print $i } }' /config.bootoptions |\
     grep root= | cut -f2- -d=
 )
 
-[ -z "${root_uuid}" ] && exit 1
+[ -z "${root_uuid}" ] && exit 2
```
davidcassany, line 34: Yes, I like it, this is helpful for debugging :+1:

davidcassany (approved):
Yes I was also considering changing the `while read` approach to awk or similar tools. I think this is cleaner.

PR #2384, opened 2023-10-23
Fix spec file to require parted in dracut modules

davidcassany:
This commit fixes a regression introduced in commit 0947fce33a. Parted package must be required to all dracut modules using 'parted' tool and/or using 'partprobe' tool. Regardless if 'parted' is used as the partitioning tool, if 'partprobe' is called 'parted' package is still required.

Fixes bsc#1216465

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -466,6 +466,7 @@ Requires:       e2fsprogs
 Requires:       grep
 Requires:       lvm2
 Requires:       mdadm
+Requires:       parted
 Requires:       util-linux
 # lsblk is part of util-linux-systemd on openSUSE
 %if 0%{?suse_version}
@@ -484,7 +485,6 @@ Requires:       device-mapper
 %endif
 %ifarch s390 s390x
 Requires:       s390-tools
-Requires:       parted
 %endif
 License:        GPL-3.0-or-later
 Group:          %{sysgroup}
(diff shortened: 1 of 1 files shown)


davidcassany:
My bad when I updated the spec to align with s390 arch use of parted I did not realize `partprobe` is also provided by `parted`.

schaefi:
Hmm, yes partprobe is provided by parted, my bad that I overlooked it at the time of the review, but it's not easy to spot, so don't worry too much.

Actually not having the parted(and tools) requirement was a good move. With the change here we sort of revert that and get dependent on parted again. I was looking into the code and maybe there is a way to keep us free from the parted requirement. parted and partprobe use is found in

```
./90kiwi-live/kiwi-live-lib.sh
./99kiwi-lib/kiwi-partitions-lib.sh
```

In the partitions lib we already have

```
if type partprobe &> /dev/null;then
    partprobe "${disk_device}"
else
    blockdev --rereadpt "${disk_device}"
fi
```

The only place in code which is not covered with an alternative to partprobe is ./90kiwi-live/kiwi-live-lib.sh

```
if ! partprobe "${isodiskdev}"; then
    return 1
fi
```

We could consider to change this code path and leave the spec untouched ?

partprobe is still mentioned as a tool to include in the ```module-setup.sh```, but dracut does not fail if the tool is not there for inclusion, which we probably can live with

Thoughts ?

davidcassany:
Closing in favor of #2385

PR #2385, opened 2023-10-25
Fallback to partx if partprobe is not available

davidcassany:
This commit makes use of partx instead of partprobe if latter is not present in initrd.

This is to not force require parted in kiwi-live module. Partx is provided by util-linux.

### DIFF
--- a/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
@@ -201,11 +201,17 @@ function preparePersistentOverlayDiskBoot {
     # used as presented
     local overlay_mount_point=$1
     local partitions_before_cow_part
+    local read_part_cmd
     mkdir -m 0755 -p "${overlay_mount_point}"
     if ! mount -L cow "${overlay_mount_point}"; then
         partitions_before_cow_part=$(_partition_count)
         echo -e "n\np\n\n\n\nw\nq" | fdisk "${isodiskdev}"
-        if ! partprobe "${isodiskdev}"; then
+        if type partprobe &> /dev/null;then
+            read_part_cmd="partprobe ${isodiskdev}"
+        else
+            read_part_cmd="partx -u ${isodiskdev}"
+        fi
+        if ! eval "${read_part_cmd}"; then
             return 1
         fi
         if [ "$(_partition_count)" -le "${partitions_before_cow_part}" ];then
--- a/dracut/modules.d/90kiwi-live/module-setup.sh
+++ b/dracut/modules.d/90kiwi-live/module-setup.sh
@@ -29,7 +29,7 @@ install() {
     inst_multiple \
         umount dmsetup blockdev blkid lsblk dd losetup \
         grep cut partprobe find wc fdisk tail mkfs.ext4 mkfs.xfs \
-        dialog cat mountpoint
+        dialog cat mountpoint partx
 
     dmsquashdir=$(find "${dracutbasedir}/modules.d" -name "*dmsquash-live")
     if [ -n "${dmsquashdir}" ] && \


davidcassany:
@schaefi I think this is a better approach to handle the live image partprobe usage. For now I did a conservative PR with just changing the behavior when partprobe is missing. 
Probably I would completely remove partprobe from kiwi-live and relay on partx as we already do for KIWI itself.

I have not used `blockdev --rereadpt` because on my TW tests it just fails, it complains about a busy device, could it be another issue? 🤷🏽‍♂️ 

The tests I did for this PR are based on the example TW iso from kiwi-descriptions repository. If we remove parted package from the list and build an image including this changes the ISO boots with persistency, without parted and without this patch it fails to set the persistent storage.

dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
```diff
@@ -201,11 +201,17 @@ function preparePersistentOverlayDiskBoot {
...
+    local read_part_cmd
     mkdir -m 0755 -p "${overlay_mount_point}"
     if ! mount -L cow "${overlay_mount_point}"; then
         partitions_before_cow_part=$(_partition_count)
         echo -e "n\np\n\n\n\nw\nq" | fdisk "${isodiskdev}"
-        if ! partprobe "${isodiskdev}"; then
+        if type partprobe &> /dev/null;then
+            read_part_cmd="partprobe ${isodiskdev}"
+        else
+            read_part_cmd="partx -u ${isodiskdev}"
```
schaefi, line 212: I wonder why we use yet another way to re-read the partition table. In ```kiwi-partitions-lib.sh``` we use either partprobe or ```blockdev --rereadpt```. Both tools blockdev and partx belongs to util-linux. For consistency reasons I suggest that our fallback from partprobe is the same for 99kiwi-lib and 90kiwi-live. ```blockdev --rereadpt``` has proven to do the job, so maybe we stick to it, unless you have reasons that favors one over the other ?

The reasons why we have a bit of code duplication between 99kiwi-lib and 90kiwi-live is because we had planned to get rid of 90kiwi-live if all live ISO features went into dracut upstream. As long as we have the two code paths they should be consistent.

Thoughts ?

dracut/modules.d/90kiwi-live/module-setup.sh
```diff
@@ -29,7 +29,7 @@ install() {
     inst_multiple \
         umount dmsetup blockdev blkid lsblk dd losetup \
         grep cut partprobe find wc fdisk tail mkfs.ext4 mkfs.xfs \
-        dialog cat mountpoint
+        dialog cat mountpoint partx
```
schaefi, line 32: If you consider to use blockdev no changes to this file would be needed

davidcassany: > I have not used blockdev --rereadpt because on my TW tests it just fails, it complains about a busy device, could it be another issue? 🤷🏽‍♂️

Yes, in fact `blockdev` was my first attempt, however I could not make it work... it complains of a busy device, probably it actually does the re-read operation, but complains of a busy device and returns non zero code. In fact, the `fdisk` call just before is also complaining and outputs a warning message suggesting that partitioning a device in use is bad idea.

schaefi: That fdisk complains is a known issue for years and the reason why we call another tool to re-read the partition table and inform the kernel. ok so this this is good information that blockdev did not work for you. In this case I suggest the other way round and use ```partx -u``` everywhere

davidcassany: Yes I am also wondering if we should consider refactoring a little the persistent partition creation to manipulate the disk without having the disk mounted.

So something like:

1. compute if persistent partition is possible and its setup
2. umount anything from the disk
3. repartition
4. mount all the required stuff

I know this is not a straight forward and probably requires a non easy refactor that expands the scope of the issue, but this would get rid of busy state of the block device. This would also get rid of the fdisk warnings and allow `blockdev --rereadpt` operation.

Thoughts on that?

Probably we can do a first attempt with `partx -u` and then consider if such a refactor is worth. Since this is hitting 15.5 code base I think having a minimalistic solution there is also valuable, I doubt we want to apply a refactor there.

davidcassany:
Closing in favor of #2387

PR #2387, opened 2023-10-26
Create live persistent storage without busy state

schaefi:
With the former logic the live ISO was already mounted when an eventual persistent storage partition was created. This leads to an issue on re-reading the partition table, not for all but several tools. This commit changes the order of tasks such that the setup of the persistent write storage is performed prior mounting the live ISO. In addition to this change an alternative method using blockdev to re-read the partition was added in case partprobe is not present. This also allows to get rid of the parted dependency which provides partprobe

### DIFF
--- a/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
@@ -116,6 +116,7 @@ function mountIso {
 function mountCompressedContainerFromIso {
     local iso_mount_point=$1
     local overlay_base
+    local squashfs_container
     overlay_base=$(getOverlayBaseDirectory)
     local container_mount_point="${overlay_base}/squashfs_container"
     squashfs_container="${iso_mount_point}/${live_dir}/${squash_image}"
@@ -205,8 +206,14 @@ function preparePersistentOverlayDiskBoot {
     if ! mount -L cow "${overlay_mount_point}"; then
         partitions_before_cow_part=$(_partition_count)
         echo -e "n\np\n\n\n\nw\nq" | fdisk "${isodiskdev}"
-        if ! partprobe "${isodiskdev}"; then
-            return 1
+        if type partprobe &> /dev/null;then
+            if ! partprobe "${isodiskdev}"; then
+                return 1
+            fi
+        else
+            if ! blockdev --rereadpt "${isodiskdev}"; then
+                return 1
+            fi
         fi
         if [ "$(_partition_count)" -le "${partitions_before_cow_part}" ];then
             return 1
--- a/dracut/modules.d/90kiwi-live/kiwi-live-root.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-live-root.sh
@@ -20,17 +20,6 @@ initGlobalDevices "$1"
 # live options and their default values
 initGlobalOptions
 
-# mount ISO device
-iso_mount_point=$(mountIso)
-
-# mount squashfs compressed container
-container_mount_point=$(
-    mountCompressedContainerFromIso "${iso_mount_point}"
-)
-
-# mount readonly root filesystem from container
-mountReadOnlyRootImageFromContainer "${container_mount_point}"
-
 # prepare overlay for generated systemd LiveOS_rootfs service
 declare isodiskdev=${isodiskdev}
 if getargbool 0 rd.live.overlay.persistent && [ -n "${isodiskdev}" ]; then
@@ -42,6 +31,17 @@ if getargbool 0 rd.live.overlay.persistent && [ -n "${isodiskdev}" ]; then
 else
     prepareTemporaryOverlay
 fi
+
+# mount ISO device
+iso_mount_point=$(mountIso)
+
+# mount squashfs compressed container
+container_mount_point=$(
+    mountCompressedContainerFromIso "${iso_mount_point}"
+)
+
+# mount readonly root filesystem from container
+mountReadOnlyRootImageFromContainer "${container_mount_point}"
 
 need_shutdown
 


schaefi:
This relates to the proposed changes in #2385. On a conversation about the topic we think that apart from the tooling alternatives also the logic needs to change.

schaefi:
@davidcassany I have done some local build tests with this code in action and it worked for me

davidcassany (commented):
Oh this looks really good, I did not expect it to be that small the required changes :+1:

Gonna give it a run before approving it if you don't mind, since I already prepared my TW test it should be fairly simple. Just to double check it fixes the blockdev issue I encountered.

schaefi:
> Gonna give it a run before approving it if you don't mind,

yes yes yes :) I really appreciate your testing, we both know how sensitive initrd code can be :)

davidcassany:
Allright it works fine on my end :+1: I did few tests:

Build a TW iso with the patch of this PR and without `parted` package, these are the logs:

```
[    4.330863] dracut-initqueue[551]: + fdisk /dev/sda
[    4.333883] dracut-initqueue[550]: + echo -e 'n\np\n\n\n\nw\nq'
[    4.341807] dracut-initqueue[551]: The device contains 'iso9660' signature and it may remain on the device. It is recommended to wipe the device with wipefs(8) or fdisk --wipe, in order to avoid possible collisions.
[    4.349359] dracut-initqueue[551]: Partition #3 contains a ext4 signature.
[    4.464041][  T551]  sda: sda1 sda2 sda3
[    4.395170] dracut-initqueue[506]: + type partprobe
[    4.485298][  T415]  sda: sda1 sda2 sda3ckdev --rereadpt /dev/sda

[    4.494607][  T561]  sda: sda1 sda2 sda3
[    4.410516] dracut-initqueue[563]: ++ _partition_count
[    4.412661] dracut-initqueue[563]: ++ '[' -z /dev/sda ']'
[    4.418707] dracut-initqueue[565]: ++ grep -c part
[    4.419770] dracut-initqueue[564]: ++ lsblk /dev/sda -p -r -n -o TYPE
[    4.428823] dracut-initqueue[506]: + '[' 3 -le 2 ']'
[    4.428889] dracut-initqueue[506]: + local write_partition
[    4.433660] dracut-initqueue[569]: ++ tail -n1
[    4.435134] dracut-initqueue[568]: ++ lsblk /dev/sda -p -l -n -o NAME
[    4.440494] dracut-initqueue[506]: + write_partition=/dev/sda3
[    4.440545] dracut-initqueue[506]: + mkfs.ext4 -L cow /dev/sda3
[    4.442987] dracut-initqueue[572]: mke2fs 1.47.0 (5-Feb-2023)
```

persistent partition is created and fdisk still produces the ISO filsystem warning (which makes sense) but it does not produce the warning for the device being used. It just works :tada:

Then build the same TW iso without this patch, these are the logs:

```
[    4.711941] dracut-initqueue[575]: + fdisk /dev/sda
[    4.713748] dracut-initqueue[574]: + echo -e 'n\np\n\n\n\nw\nq'
[    4.723079] dracut-initqueue[575]: This disk is currently in use - repartitioning is probably a bad idea.
[    4.723147] dracut-initqueue[575]: It's recommended to umount all file systems, and swapoff all swap
[    4.723177] dracut-initqueue[575]: partitions on this disk.
[    4.725387] dracut-initqueue[575]: The device contains 'iso9660' signature and it may remain on the device. It is recommended to wipe the device with wipefs(8) or fdisk --wipe, in order to avoid possible collisions.
[    4.757811] dracut-initqueue[508]: + partprobe /dev/sda
[    4.762644] dracut-initqueue[576]: /lib/kiwi-live-lib.sh: line 208: partprobe: command not found
[    4.764010] dracut-initqueue[508]: + return 1
[    4.764065] dracut-initqueue[508]: + return 1
[    4.764096] dracut-initqueue[508]: + warn 'Failed to setup persistent write space !'
[    4.764122] dracut-initqueue[508]: + echo 'Warning: Failed to setup persistent write space !'
[    4.764151] dracut-initqueue[508]: Warning: Failed to setup persistent write space !
[    4.764181] dracut-initqueue[508]: + warn 'Falling back to temporary overlay'
[    4.764206] dracut-initqueue[508]: + echo 'Warning: Falling back to temporary overlay'
[    4.764229] dracut-initqueue[508]: Warning: Falling back to temporary overlay
[    4.764253] dracut-initqueue[508]: + prepareTemporaryOverlay
[    4.764280] dracut-initqueue[508]: + local overlay_base
[    4.773452] dracut-initqueue[577]: ++ getOverlayBaseDirectory
[    4.773924] dracut-initqueue[577]: ++ local overlay_base=/run/overlay
[    4.774116] dracut-initqueue[577]: ++ local overlay_size
[    4.774203] dracut-initqueue[577]: ++ mkdir -p /run/overlay
[    4.781631] dracut-initqueue[577]: ++ mountpoint -q /run/overlay
[    4.788125] dracut-initqueue[577]: ++ echo /run/overlay
[    4.788494] dracut-initqueue[508]: + overlay_base=/run/overlay
[    4.788539] dracut-initqueue[508]: + mkdir -m 0755 -p /run/overlay/overlayfs/rw
[    4.791252] dracut-initqueue[508]: + mkdir -m 0755 -p /run/overlay/overlayfs/work
[    4.797092] dracut-initqueue[508]: + need_shutdown
[    4.797141] dracut-initqueue[508]: + :
[    4.797171] dracut-initqueue[508]: + exit 0
```

`fdisk` complains about partitioning a mounted device but continues then partprobe fails because the tool is just not there.

Finally, just for completion, I did a last test just using the `blockdev --rereadpt` while the device is still mounted (the change and tried first that failed) and I got the following:

```
[    4.589230] dracut-initqueue[573]: + fdisk /dev/sda
[    4.593199] dracut-initqueue[572]: + echo -e 'n\np\n\n\n\nw\nq'
[    4.600784] dracut-initqueue[573]: This disk is currently in use - repartitioning is probably a bad idea.
[    4.600847] dracut-initqueue[573]: It's recommended to umount all file systems, and swapoff all swap
[    4.600874] dracut-initqueue[573]: partitions on this disk.
[    4.603157] dracut-initqueue[573]: The device contains 'iso9660' signature and it may remain on the device. It is recommended to wipe the device with wipefs(8) or fdisk --wipe, in order to avoid possible collisions.
[    4.609423] dracut-initqueue[573]: Partition #3 contains a ext4 signature.
[    4.641012] dracut-initqueue[506]: + type partprobe
[    4.641076] dracut-initqueue[506]: + blockdev --rereadpt /dev/sda
[    4.646510] dracut-initqueue[574]: blockdev: ioctl error on BLKRRPART: Device or resource busy
[    4.648177] dracut-initqueue[506]: + return 1
[    4.648225] dracut-initqueue[506]: + return 1
[    4.648253] dracut-initqueue[506]: + warn 'Failed to setup persistent write space !'
[    4.648277] dracut-initqueue[506]: + echo 'Warning: Failed to setup persistent write space !'
[    4.648301] dracut-initqueue[506]: Warning: Failed to setup persistent write space !
[    4.648331] dracut-initqueue[506]: + warn 'Falling back to temporary overlay'
[    4.648356] dracut-initqueue[506]: + echo 'Warning: Falling back to temporary overlay'
[    4.648379] dracut-initqueue[506]: Warning: Falling back to temporary overlay
[    4.648401] dracut-initqueue[506]: + prepareTemporaryOverlay
[    4.648423] dracut-initqueue[506]: + local overlay_base
[    4.661894] dracut-initqueue[575]: ++ getOverlayBaseDirectory
[    4.662059] dracut-initqueue[575]: ++ local overlay_base=/run/overlay
[    4.662141] dracut-initqueue[575]: ++ local overlay_size
[    4.662220] dracut-initqueue[575]: ++ mkdir -p /run/overlay
[    4.671146] dracut-initqueue[575]: ++ mountpoint -q /run/overlay
[    4.672802] dracut-initqueue[575]: ++ echo /run/overlay
[    4.673175] dracut-initqueue[506]: + overlay_base=/run/overlay
[    4.673220] dracut-initqueue[506]: + mkdir -m 0755 -p /run/overlay/overlayfs/rw
[    4.678184] dracut-initqueue[506]: + mkdir -m 0755 -p /run/overlay/overlayfs/work
[    4.683243] dracut-initqueue[506]: + need_shutdown
[    4.683300] dracut-initqueue[506]: + :
[    4.683329] dracut-initqueue[506]: + exit 0
```

Setting the persistent partition fails because `blockdev --rereadpt` refuses to do anything on a busy device and errors out.

davidcassany (approved):

schaefi:
@davidcassany great thanks much for testing. Especially the last test tells us we will never be successful with the blockdev fallback if the device is in a busy state. That brings me to the question if we should change the partprobe fallback to partx rather than blockdev ? Informing the kernel about a new partition geometry can be done in a busy state and is also effective for new devices and new mounts based on the new geometry. busy state mounts of course will not see it until swap of the busy state but a complete refuse of operation is imho not required. iirc ```partx -u``` is less restrictive than blockdev in this situation. So I think a change of the fallback would be good

Thoughts ?

schaefi:
@davidcassany I added a commit to the idea of the changed default

davidcassany:
@schaefi recently in other projects I switched from `partprobe` to `partx`, because we noted a behavior change from leap 15.4 to 15.5 in `parted` regarding how loop devices are managed. `parted` and `partprobe` no longer inform the kernel on loop device partitions, for whatever reason parittioning is possible not it does not notify the kernel or does it in an ineffective way. We haven't seen that on `partx` neither I have seen now in `blockdev`. On on this story my take away is that `parted` is not that reliable and, in fact, before facing this we were already fading it out in KIWI I guess for similar motivations.

Then regarding partx vs partprobe I think is good for two reasons:

* in this tests looks like it behaves in a better way, feels like it is only parsing partition headers without requiring the device to be an actual block device. In other words, `partx` works fine directly on disk files without creating a loop device. `blockdev` requires block devices always. I guess blockdev uses a blockdevice kernel interface and partx just parses partition table headers (just wild guess).
* `partx` is already used within the KIWI code to actually read and map partitions, hence we get a consistent behavior in both sides of the code, kiwi core and its dracut modules.

davidcassany:
> @davidcassany I added a commit to the idea of the changed default

Yes I think this is a good move too.

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -25,7 +25,7 @@ function create_partitions {
     if type partprobe &> /dev/null;then
         partprobe "${disk_device}"
     else
-        blockdev --rereadpt "${disk_device}"
+        partx -u "${disk_device}"
```
davidcassany, line 28: @schaefi I think partx should be added to module-setup.sh for both modules. For the live module it is enough to swap it for blockdev (since it is no longer used there). For kiwi-lib then both are required `blockdev` and `partx`. Both tools a provided by util-linux, hence no spec changes are required.

schaefi: yes, fixed

dracut/modules.d/90kiwi-live/module-setup.sh
```diff
@@ -27,7 +27,7 @@ install() {
     local dmsquashdir=
     inst_multiple -o checkmedia
     inst_multiple \
-        umount dmsetup blockdev blkid lsblk dd losetup \
+        umount dmsetup partx blkid lsblk dd losetup \
```
davidcassany, line 30: Isn't a similar change also required to `dracut/modules.d/99kiwi-lib/module-setup.sh`. I think `partx` should be added to list for `kiwi-lib` module too.

schaefi: @davidcassany It's so good that you are also looking at this. Yes I missed to add it to the kiwi-lib module. Thanks a ton. I think we are good now. Can you do a final review ? Thanks much

dracut/modules.d/99kiwi-lib/module-setup.sh
```diff
@@ -15,7 +15,7 @@ depends() {
 install() {
     declare moddir=${moddir}
     inst_multiple \
-        blkid blockdev dd mkdir rmdir \
+        blkid partx blockdev dd mkdir rmdir \
```
davidcassany, line 18: sorry @schaefi this change is not good yet, in kiwi-lib we are not replacing blockdev by partx completely. `blockdev` is still used to get block devices size. I could find a way to do the same with `partx`, hence we can't replace one by the other, we need both. In fact the current change as is in this commit has potential to break expandable disks. See where blockdev is used:

```bash
grep -rn blockdev dracut/modules.d/
dracut/modules.d/99kiwi-lib/module-setup.sh:18:        blkid blockdev dd mkdir rmdir \
dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh:244:    echo $(($(blockdev --getsize64 "${device}") / 1024))
dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh:253:    disk_bytes=$(blockdev --getsize64 "${disk}")
dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh:284:            part_bytes=$((part_bytes + $(blockdev --getsize64 "${part}")))
```

`blockdev --getsize64` is a critical call. We can't risk missing blockdev in kiwi-lib. Probably some other module already pulls it in, hence adding it here probably does not make a difference in most cases, but still a hard requirement we need to include.

davidcassany (changes requested):
Sorry there is still a little detail to fix, in kiwi-lib module we can't remove blockdev requirement.

dracut/modules.d/99kiwi-lib/module-setup.sh
schaefi: I know and I did not remove it from kiwi-lib. Maybe you overlooked it but blockdev is still in the list to be included in module-setup.sh. blockdev got only deleted from the kiwi-live module as it is not used there anymore. Or did I miss something ?

davidcassany: Damn it! you are absolutely right :+1: I don't know how I looked at diff that I had impression that partx was not added but subsituted. Sorry for the noise 🙏🏽

davidcassany (approved):
I was wrong on my last review, the changes are correct, I overlooked it, sorry.

Thanks Marcus

PR #2388, opened 2023-10-30
Rewrite loader linux/initrd entry relabelling without regex to improve reliability

KaliumPuceon:
Replacing the regex-based loader entry fix with string parsing.

A user building RHEL images ran into issues with the initrd. It turns out that RHEL uses some patches that mean the initrd/linux files in RHEL are not installed to /boot, which trips up the original regex. The new fix doesn't rely on matching the path in boot, instead just finding the initrd/linux files and rewriting them in place.

This change also adds the pre-and-post fix loader entries to the debug logs.

Original bug report https://bugzilla.suse.com/show_bug.cgi?id=1208701

Fixes suse bsc#1208701

* Replace loader fix regex with python string handling.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -1459,19 +1459,24 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         )
         for menu_entry_file in glob.iglob(loader_entries_pattern):
             with open(menu_entry_file) as grub_menu_entry_file:
-                menu_entry = grub_menu_entry_file.read()
-                if self.bootpartition:
-                    menu_entry = re.sub(
-                        r'(linux|initrd) .*/boot(.*)',
-                        r'\1 \2',
-                        menu_entry
-                    )
-                else:
-                    menu_entry = re.sub(
-                        r'(linux|initrd) .*(/boot.*)',
-                        r'\1 \2',
-                        menu_entry
-                    )
+                menu_entry = grub_menu_entry_file.read().split(os.linesep)
+
+            for line_number, menu_entry_line in enumerate(menu_entry):
+                if bool(re.match(r'^(linux|initrd) .*', menu_entry_line)):
+
+                    log.debug(f'Existing loader entry: {menu_entry_line}')
+                    config_path = menu_entry_line.split(' ', 1)[0]
+
+                    basename = os.path.basename(menu_entry_line)
+                    if self.bootpartition:
+                        config_path = (f'{config_path} /{basename}')
+                    else:
+                        config_path = (f'{config_path} /boot/{basename}')
+
+                    menu_entry[line_number] = config_path
+                    log.debug(f'Updated loader entry: {config_path}')
+
+            menu_entry = os.linesep.join(menu_entry)
             with open(menu_entry_file, 'w') as grub_menu_entry_file:
                 grub_menu_entry_file.write(menu_entry)
 


schaefi:
Thanks for the work done here. As discussed internally on slack I appreciate the change. I'll do some tests. Could you please fix the white space issues reported by the linters ? Thanks much

@davidcassany if you can take a look at the changes would be great. I'm on vacation this week and will take a look when I'm back

kiwi/bootloader/config/grub2.py
```diff
@@ -1459,19 +1459,24 @@ def _fix_grub_loader_entries_linux_and_initrd_paths(self):
...
+                    basename = os.path.basename(menu_entry_line)
+                    if self.bootpartition:
+                        config_path += (" /%s" % basename)
+                    else:
+                        config_path += (" /boot/%s" % basename)
+
+                    menu_entry[line_number] = config_path
+                    log.debug("Updated loader entry: %s", config_path)
+
+            menu_entry = "\n".join(menu_entry)
```
schaefi, line 1479: Can we use ```os.linesep``` instead of ```\n``` here

kiwi/bootloader/config/grub2.py
```diff
@@ -1459,19 +1459,24 @@ def _fix_grub_loader_entries_linux_and_initrd_paths(self):
...
+
+            for line_number, menu_entry_line in enumerate(menu_entry):
+                if bool(re.match(r'^(linux|initrd) .*', menu_entry_line)):
+
+                    log.debug("Existing loader entry: %s", menu_entry_line)
+                    config_path = menu_entry_line.split(" ", 1)[0]
+
+                    basename = os.path.basename(menu_entry_line)
+                    if self.bootpartition:
+                        config_path += (" /%s" % basename)
```
schaefi, line 1472: Can we use python's f strings for consistency reasons

```python
config_path.append(f' /{basename}')
```

kiwi/bootloader/config/grub2.py
```diff
@@ -1459,19 +1459,24 @@ def _fix_grub_loader_entries_linux_and_initrd_paths(self):
...
+                    else:
+                        config_path += (" /boot/%s" % basename)
```
schaefi, line 1474: same here, can we use f strings

kiwi/bootloader/config/grub2.py
```diff
@@ -1459,19 +1459,24 @@ def _fix_grub_loader_entries_linux_and_initrd_paths(self):
...
-                        r'\1 \2',
-                        menu_entry
-                    )
-                else:
-                    menu_entry = re.sub(
-                        r'(linux|initrd) .*(/boot.*)',
-                        r'\1 \2',
-                        menu_entry
-                    )
+                menu_entry = grub_menu_entry_file.read().split("\n")
```
schaefi, line 1462: sorry overlooked this one, ```os.linesep``` please, thanks

kiwi/bootloader/config/grub2.py
```diff
@@ -1459,19 +1459,24 @@ def _fix_grub_loader_entries_linux_and_initrd_paths(self):
...
-                        r'\1 \2',
-                        menu_entry
-                    )
+                menu_entry = grub_menu_entry_file.read().split(os.linesep)
+
+            for line_number, menu_entry_line in enumerate(menu_entry):
+                if bool(re.match(r'^(linux|initrd) .*', menu_entry_line)):
+
+                    log.debug(f'Existing loader entry: {menu_entry_line}')
+                    config_path = menu_entry_line.split(" ", 1)[0]
```
schaefi, line 1468: single quotes please ```' '``` instead of ```" "```. Reason here is also consistency and that double quotes for strings in python are typically used when variables or other types of substitution is part of the string expression. As this is not the case here I recommend the single quotes. Sorry if this is nitpick but the more consistent we are the more robust the code will grow. Thanks

After that it looks good to me. Thanks much for your patience :)

schaefi:
@davidcassany can you also take a look on the proposed changes ? Thanks much

schaefi (approved):
Given the last nit pick change got addressed, this looks good to me. Thanks for your effort

Conan-Kudo (changes requested):
Please squash all this into a single commit.

Conan-Kudo:
@KaliumPuceon Can you put your pull request description in the commit message? Like so in your editor:

```
Replace the regex-based loader entry fix with string parsing

A user building RHEL images ran into issues with the initrd. 
It turns out that RHEL uses some patches that mean the
initrd/linux files in RHEL are not installed to /boot, which trips
up the original regex. The new fix doesn't rely on matching the
path in boot, instead just finding the initrd/linux files and rewriting
them in place.

This change also adds the pre-and-post fix loader entries to the debug logs.

Reference: https://bugzilla.suse.com/1208701

Fixes suse bsc#1208701
```

Conan-Kudo:
I just took care of it via squash merge.

schaefi:
ok, this change now needs good testing through the integration test matrix. As it will be visible at boot time I would wish we can just trigger the build tests done by @dcermak somehow ? But I think we need Dan to do it. I will update kiwi in Staging today so that the integration tests rebuild

PR #2393, opened 2023-11-10
Container installation support

rjschwei:
As systems become leaner and functionality moves from system installed packages (RPM, deb or other) to containers it is important that kiwi can pre-install conatiners from specified registries to build system images.

* Schema support for container installation with podman
*

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -65,7 +65,7 @@ div {
         attribute xsi:schemaLocation { xsd:anyURI }
     k.image.schemaversion.attribute =
         ## The allowed Schema version (fixed value)
-        attribute schemaversion { "7.6" }
+        attribute schemaversion { "7.7" }
     k.image.id =
         ## An identification number which is represented in a file
         ## named /etc/ImageID
@@ -89,6 +89,7 @@ div {
             k.drivers* &
             k.strip* &
             k.repository* &
+            k.registry* &
             k.packages* &
             k.extension?
         }
@@ -938,6 +939,21 @@ div {
 }
 
 #==========================================
+# common element <container>
+#
+div {
+    k.container.name.attribute = k.name.attribute
+    k.container.attlist =
+        k.container.name.attribute
+    k.container =
+        ## Name of container to install in the image
+        element container {
+            k.container.attlist,
+            empty
+        }
+}
+
+#==========================================
 # common element <package>
 #
 div {
@@ -955,6 +971,22 @@ div {
         element package {
             k.package.attlist,
             empty
+        }
+}
+
+#==========================================
+# common element <containerruntime>
+#
+div {
+    k.containerruntime.content =
+        "podman"
+    k.containerruntime.attlist = empty
+    k.containerruntime =
+        ## Name of the container orchestrator
+        ## Must be installed in the image
+        element containerruntime {
+            k.containerruntime.attlist,
+            k.containerruntime.content
         }
 }
 
@@ -1035,6 +1067,23 @@ div {
         }
 }
 
+#==========================================
+# common element <registry>
+#
+div {
+    k.registry.profiles.attribute = k.profiles.attribute
+    k.registry.path.attribute = k.path.attribute
+    k.registry.attlist =
+        k.registry.profiles.attribute? &
+        k.registry.path.attribute
+
+    k.registry =
+        ## The path to an OCI container registry
+        element registry {
+            k.registry.attlist
+    }
+}
+    
 #==========================================
 # common element <repository>
 #
@@ -3516,6 +3565,7 @@ div {
             k.namedCollection* &
             k.collectionModule* &
             k.product* &
+            k.container* &
             k.package*
         }
 }
@@ -3541,6 +3591,7 @@ div {
             k.keytable? &
             k.locale? &
             k.packagemanager? &
+            k.containerruntime? &
             k.release-version? &
             k.rpm-locale-filtering? &
             k.rpm-check-signatures? &
--- a/kiwi/schema/kiwi.rng
+++ b/kiwi/schema/kiwi.rng
@@ -109,7 +109,7 @@
   <!--
     ==========================================
     start with image description
-
+    
   -->
   <start>
     <ref name="k.image">
@@ -119,7 +119,7 @@
   <!--
     ==========================================
     main block: <image>
-
+    
   -->
   <div>
     <define name="k.image.name.attribute">
@@ -145,7 +145,7 @@ second the location of the XSD Schema
     <define name="k.image.schemaversion.attribute">
       <attribute name="schemaversion">
         <a:documentation>The allowed Schema version (fixed value)</a:documentation>
-        <value>7.6</value>
+        <value>7.7</value>
       </attribute>
     </define>
     <define name="k.image.id">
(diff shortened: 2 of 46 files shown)


rjschwei:
@schaefi can you give this a once over to see if I am going in the right direction? The idea is to eventually have kiwi execute `podman pull` in the chroot image environment after package install is finished. This sets up XML support.

schaefi:
@rjschwei thanks for starting this effort. I took a look to the schema changes you are about to add and those for sure makes sense when the task is to pull a container into the local registry of the image. However, I've been tasked with this several times in the past and had to deal with some hurdles which held me back to do what you started now. Let me share some thoughts:

1. OBS. It's not possible to run outgoing connections in OBS, this includes fetching an arbitrary container from a container registry, even if that registry is the suse registry. Thus your new container elements would imply a magic handling inside of OBS like it is done with packages and that doesn't exist
2. The container registry is connected to a storage driver. That storage driver is configurable and by default set to overlayfs. When using overlayfs you can pre-load containers to the local registry in kiwi, however if you use e.g btrfs then loading of containers requires the btrfs layer to be present at the time of the pulling. In kiwi this means a generic solution to pull containers has to run with the device layer active. This limits the feature to disk images because they expose the real devices (through loops) and should allow the pull for any storage driver. Either this limitation or forcing into overlay as storage driver is a decision that needs to be taken
3. Containers. Many people build images in CI/CD systems that are based on containers. Running kiwi in a container and building an image that itself pulls containers has caused very weird error conditions in the past for which I could not come up with a solution other than "build in a VM" ... so yet another limitation

There is more but I don't want to scare you ;)

With the above in mind I did not add an extra container tag to the kiwi schema so far. The concept we use so far uses a different approach based on the following design:

1. Package the container. We can do this in a very simple way using OBS, any other way to put a container into a package would also work. If you stick with OBS you only need to add ```support: containment-rpm-docker``` to your prjconf. Any container build in that project will also be provided as a package
2. Preload the container as a package. Instead of a new tag for handling containers you just add ```<package name="the-container-package"/>``` This now also works inside of OBS, outside of OBS and with any other package that contains the container
3. Image description. In the image description you need to add code that does ```podman load -i the-container-from-the-package``` and if the raw container should not be kept you also need to put the package into the ```<packages type="uninstall"/>``` section. For that podman call I thought a helper method in ```kiwi/config/functions.sh``` might be useful but as it's only one line I did not add something like that so far. We could think of an extra attribute that marks the package as a container package and automatically includes it into the registry. Something like ```<package name="foo" oci_import="true"/>```
4. The problem with the storage driver, container policy, etc etc. If you have a container package you can build the image and store the container locally. You can then load the container at a later point in time e.g on first boot which will then eliminate all storage driver issues and more. It's like a pre-provision and set active later approach

I setup all images that needs containers that way. Yes it requires putting the container into a package which might be considered cumbersome to people

If we add a new container tag I strongly recommend that we don't aim for a direct pull into the local registry but load the container archive as file first and then allow via an attribute what should happen with it: stay as archive, try-loading-but-might-fail---document-exceptions. I could also imagine if we add a new mime type to reference the container name e.g ```obsregistry://..." that the obs team will add some magic such that it can be fetched from a worker.

Thus from my perspective a step that we can go is something like this

```xml
<package type="image">
<container name="URI" save_as="/usr/share/my/containers/" preload="true|false"/>
</packages>
```

The URI references from where to fetch it and supports the format that you can use with podman pull plus additional mime types that we might need for integration with OBS ... or others

Sorry long story, but I had some experience and trouble with all this already :smiley:

kiwi/schema/kiwi.rnc
```diff
@@ -65,7 +65,7 @@ div {
         attribute xsi:schemaLocation { xsd:anyURI }
     k.image.schemaversion.attribute =
         ## The allowed Schema version (fixed value)
-        attribute schemaversion { "7.6" }
+        attribute schemaversion { "7.7" }
```
schaefi, line 68: We only change the schema version on incompatible schema changes which requires an XSLT transformation.  You are only adding stuff which is not an incompatible change. For tracking features according to the version of kiwi we use the main kiwi version not the schema version. Please keep the version as it is

This also obsoletes your XSLT transformation which does basically nothing than changing the version. So please drop the XSLT

Also please drop the adaptions to the various XML snippets that only changes the schema version.
It will make the pull request a lot smaller :)

Thanks much

kiwi/schema/kiwi.rnc
```diff
@@ -3516,6 +3565,7 @@ div {
             k.namedCollection* &
             k.collectionModule* &
             k.product* &
+            k.container* &
```
schaefi, line 3568: As I wrote in my long first comment I think we can make our live a bit easier if we just add a new container element inside of the ```<packages>``` section. A packages sub-element in kiwi speak can be a package, an archive and now also a container. The registry information should be imho placed in an URI reference which is the way we support source references in kiwi. I don't think we need to invent a different way to do this unless I forgot something ?

schaefi (changes requested):
I like the idea to reference a container and operate on it, but I think for consistency and simplicity the way to describe it should be changed. Also see my initial comment. Thoughts ?

rjschwei:
@schaefi thanks for sharing your thoughts. On the OBS side, yes while we want this feature in OBS, OBS is not the only place where to build images. Therefore, if kiwi would have a feature that doesn't work in OBS then that would just be it. I am not certain we should let potential limitations of 1 tool dictate features in kiwi. That said, the problem can certainly be solved in a similar way to repositories, as you already stated `obsregistry://`. 

At first I also thought about adding the location of the container to the container tag but then I decided that this could lead to a lot of repetition, if for every container I want to get installed I need to add `some-registry` then this gets really old. Also a reason I started with `podman` as it has a config file to write registries from which to search to. That said, that may of course lead to duplication which would present trouble during install, for example on a deafult openSUSE setup one gets

```
# podman pull distrobox
? Please select an image: 
  ▸ registry.opensuse.org/distrobox:latest
    registry.suse.com/distrobox:latest
    docker.io/library/distrobox:latest
```
With my current thought process of eliminating the duplication for registry specification it becomes the user's responsibility to avoid using registries that have the same container name. That said this may be an untenable task for a user as shown in the example above. If a user wants the openSUSE `distrobox` container but also a container that is only available on docker hub then the user cannot avoid the search conflict for distrobox and since it is not possible to the best of my knowledge to specify some kind of priority the idea of avoiding duplication goes out the window. And that makes the integration with OBS also more difficult.

The packaging of containers in RPMs brings it's own problems and there have been a number of discussions about that.

Well at least we have this documented now for referencing. I am going to close this PR and we'll see where the conversation in other places around containers wrapped in RPMs leads.

schaefi:
> Therefore, if kiwi would have a feature that doesn't work in OBS then that would just be it

I agree and have no problem with that. The feature should just at least in theory allow to be expanded for the OBS use case. I think the idea would have covered that part

>While the default install would render a choice problem. Meaning kiwi would have to overwrite the default distribution 
setup.

yes and as you explained this can be done in kiwi code for the time the image is being built. I don't think it would be problematic to implement that. We had to do it for all package managers we support too

> The packaging of containers in RPMs brings it's own problems and there have been a number of discussions about that.

right and it has one advantage in a way that it does not require extra implementations in kiwi. The handling of the installed data via package or by a pull from a registry would be the same code

I wonder why you closed your PR. I think adding a feature that allows to fetch a container from a registry (without looking at OBS for the moment) would be a nice and great addition. We only need to agree on the format to describe it in XML speak and you could go ahead.

I hope I did not demotivate with you with my feedback, it was not my intention

rjschwei:
@schaefi rather than changing this PR I created a new branch https://github.com/OSInside/kiwi/commit/4ff11a5d95e8d97f14f75b3e9795a201d455e601 in light of what I learned today this is getting a bit more pressing as well. So no you did not break my motivation. Just thought it would be nicer to start with the smaller scope schema changes rather than undoing some of the schema work I did here.

PR #2398, opened 2023-11-22
Support install disk overwrite per cmdline

schaefi:
Allow install disk overwrite from cmdline
    
Add rd.kiwi.oem.installdevice=DEVICE. Configures the disk device that should be used in an OEM installation. This oerwrites any other oem device setting, e.g device filter or maxdisk and just  continues the installation on the given device. However, the device must exist and must be a block special.

### DIFF
--- a/doc/source/building_images/build_expandable_disk.rst
+++ b/doc/source/building_images/build_expandable_disk.rst
@@ -393,6 +393,15 @@ oemconfig.oem-unattended Element
   possible target devices are discovered the image is deployed to
   the first device. ``kiwi_oemunattended`` in the initrd
 
+oemconfig.oem-unattended-id Element
+  The target disk device for the installation is selected according
+  to the specified device ID. The device ID corresponds to the name of
+  the device as it exists for the configured `devicepersistency`. By
+  default this is the `by-uuid` device name. If no such representation
+  exist e.g for ramdisk devices, the unix device node can be used
+  to select. The given name must be present in the device list detected
+  by kiwi.
+
 oemconfig.oem-skip-verify Element
   Do not perform the checksum verification process after install
   of the image to the target disk. The verification process computes
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -20,6 +20,8 @@ function get_disk_list {
     declare kiwi_oemmultipath_scan=${kiwi_oemmultipath_scan}
     declare kiwi_devicepersistency=${kiwi_devicepersistency}
     declare kiwi_install_volid=${kiwi_install_volid}
+    declare kiwi_oemunattended=${kiwi_oemunattended}
+    declare kiwi_oemunattended_id=${kiwi_oemunattended_id}
     local disk_id="by-id"
     local disk_size
     local disk_device
@@ -38,6 +40,7 @@ function get_disk_list {
     max_disk=0
     kiwi_oemmultipath_scan=$(bool "${kiwi_oemmultipath_scan}")
     kiwi_oem_maxdisk=$(getarg rd.kiwi.oem.maxdisk=)
+    kiwi_oem_installdevice=$(getarg rd.kiwi.oem.installdevice=)
     if [ -n "${kiwi_oem_maxdisk}" ]; then
         max_disk=$(binsize_to_bytesize "${kiwi_oem_maxdisk}") || max_disk=0
     fi
@@ -107,6 +110,39 @@ function get_disk_list {
         fi
         list_items="${list_items} ${disk_device} ${disk_size}"
     done
+    if [ -n "${kiwi_oem_installdevice}" ];then
+        # install device overwritten by cmdline.
+        local device=${kiwi_oem_installdevice}
+        local device_meta
+        local device_size
+        if [ ! -e "${device}" ];then
+            local no_dev="Given device ${device} does not exist"
+            report_and_quit "${no_dev}"
+        fi
+        if [ ! -b "${device}" ];then
+            local no_block_dev="Given device ${device} is not a block special"
+            report_and_quit "${no_block_dev}"
+        fi
+        device_meta=$(
+            eval lsblk "${blk_opts}" "${device}" |\
+            grep -E "disk|raid" | tr ' ' ":"
+        )
+        device_size=$(echo "${device_meta}" | cut -f2 -d:)
+        list_items="${device} ${device_size}"
+        # activate unattended mode. In case a user explicitly
+        # provides the device name to deploy the image to via
+        # the kernel commandline, no further questions if this
+        # is wanted should appear
+        message="rd.kiwi.oem.installdevice set via cmdline to: ${device}"
+        message="${message} The following OEM device settings will be ignored:"
+        message="${message} oem-unattended,"
+        message="${message} oem-unattended-id,"
+        message="${message} oem-device-filter"
+        export kiwi_oemunattended="true"
+        export kiwi_oemunattended_id=""
+        export kiwi_oemdevicefilter=""
+        info "${message}" >&2
+    fi
     if [ -z "${list_items}" ];then
         local no_device_text="No device(s) for installation found"
         report_and_quit "${no_device_text}"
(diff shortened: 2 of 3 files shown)


dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
```diff
@@ -107,6 +108,26 @@ function get_disk_list {
...
+            local no_block_dev="Given device ${device} is not a block special"
+            report_and_quit "${no_block_dev}"
+        fi
+        device_meta=$(
+            eval lsblk "${blk_opts}" "${kiwi_oem_installdevice}" |\
+            grep -E "disk|raid" | tr ' ' ":"
+        )
+        device_size=$(echo "${device_meta}" | cut -f2 -d:)
+        list_items="${device} ${device_size}"
+    fi
```
e-minguez, line 130: Shouldn't `rd.kiwi.oem.installdevice` override the `kiwi_oemunattended` variable (to true)  as well?

schaefi: Hmm, why should it do that ? If you overwrite the device to install to, how does this affect the "I want to confirm the selected install target interactively yes|no ?" feature. By default kiwi asks before wiping your disk. I think it would be relatively unexpected that it does not do this anymore just because you set rd.kiwi.oem.installdevice right ?

e-minguez: I believe it is quite the opposite :) If you are adding the device to the boot parameters you are aware that the device will be written, why would you want to confirm it a second time?

schaefi: ok that actually makes sense to me, you won :-) I agree that someone setting a boot parameter usually should know about the consequences. But let us also add a big fat note about the behavior in the documentation ok ?
I'll adapt the PR according to what you said, thanks much

doc/source/building_images/build_expandable_disk.rst
```diff
@@ -393,6 +393,15 @@ oemconfig.oem-unattended Element
...
 
+oemconfig.oem-unattended-id Element
+  The target disk device for the installation is selected according
+  to the specified device ID. The device ID corresponds to the name of
+  the device as it exists for the configured `devicepersistency`. By
+  default this is the `by-uuid` device name. If no such representation
+  exist e.g for ramdisk devices, the unix device node can be used
+  to select. The given name must be present in the device list detected
+  by kiwi.
+
```
e-minguez, line 404:
```suggestion
.. warning::
    Setting this parameter will perform the installation without any confirmation whatsoever.

```

e-minguez: Not sure if this is the the proper rst format tho (never used it but it seems like it supports it as per https://docutils.sourceforge.io/docs/ref/rst/directives.html#warning)

dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
e-minguez: Sounds great, thanks!

e-minguez (approved):

schaefi:
cool, thanks for the review. As this is initrd code I'm always a bit concerned and therefore I also like to have two people approving. @davidcassany could you have a short look if you see something problematic ? I think we are good though. Thanks

doc/source/building_images/build_expandable_disk.rst
davidcassany, line 404: Added documentation, nice :+1:

 Probably we should note that  `oem-unattended-id` has preference over `rd.kiwi.oem.installdevice` (or the other way round if you think it should be the opposite)

davidcassany (approved):
Looks good to me.

Just a comment regarding the current logic. From lines https://github.com/OSInside/kiwi/blob/488be00c8aa5d22fbef78586ed0313c457bcca8f/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh#L138-L152

I understand that `oem-unattended-id` element from the XML will always have precedence over `rd.kiwi.oem.installdevice` , I am not sure this is intended. In general, my expectation as a user would be that runtime choices in kernel command parameters overwrite buildtime choices coming from the XML, so that the same image build turns to be usable in a wider range of use cases.

doc/source/building_images/build_expandable_disk.rst
schaefi: > I understand that oem-unattended-id element from the XML will always have precedence over rd.kiwi.oem.installdevice

I believe this is not the case. From a code logic we only evaluate kiwi_oemunattended_id if there is more than one device found. As we set the list of devices to exactly one in the rd.kiwi.oem.installdevice case we do not enter that code path.

But I agree to you this should be done more explicit. I suggest to reset any eventually given kiwi_oemunattended_id and also kiwi_oemdevicefilter in case rd.kiwi.oem.installdevice is set.

schaefi: @davidcassany I added a commit which addresses your issue. I think we want that rd.kiwi.oem.installdevice as a kernel parameter takes precedence over any other setting and the docs should also be clear on that. What do you think ?

davidcassany: > I believe this is not the case. From a code logic we only evaluate kiwi_oemunattended_id if there is more than one device found. As we set the list of devices to exactly one in the rd.kiwi.oem.installdevice case we do not enter that code path.

Oh, indeed, you are right. Then from a code perspective also setting `kiwi_oemunattended` to `true` had no real effect, I guess that's why I got confused as I directly looked when this variable was used without reading the full code 😅 

Yes the current approach looks all good to me. Resetting all overwritten (or ignored) parameters looks like the most readable and maintainable approach :+1:.

schaefi: >Then from a code perspective also setting kiwi_oemunattended to true had no real effect
Well it had one at the code path that asks to continue later
```bash
if [ "${kiwi_oemunattended}" = "false" ];then
        local ack_dump_text="Destroying ALL data on ${image_target}, continue ?"
        ...
fi
```
So this setting was indeed needed :) For all the others you are right, but now it's explicit and better to read and understand. Thanks for the feedback and review :+1:

schaefi:
my tests were successful. I tested:
* normal operation, no regression found
* with ```rd.kiwi.oem.installdevice=/dev/foo``` ends with expected error message on the console
* with ```rd.kiwi.oem.installdevice=/dev/sda``` installs like in normal case because my qemu disk appears as /dev/sda
* with ```rd.kiwi.oem.installdevice=/dev/disk/by-id/ata-QEMU_HARDDISK_QM00001``` installs like in normal case because ata-QEMU_HARDDISK_QM00001 -> ../../sda

If you have more specific tests in my mind just try them with the proposed integration test. Thanks

e-minguez:
> my tests were successful. I tested:
> 
> * normal operation, no regression found
> * with `rd.kiwi.oem.installdevice=/dev/foo` ends with expected error message on the console
> * with `rd.kiwi.oem.installdevice=/dev/sda` installs like in normal case because my qemu disk appears as /dev/sda
> * with `rd.kiwi.oem.installdevice=/dev/disk/by-id/ata-QEMU_HARDDISK_QM00001` installs like in normal case because ata-QEMU_HARDDISK_QM00001 -> ../../sda
> 
> If you have more specific tests in my mind just try them with the proposed integration test. Thanks

Just in case it would be nice to test installing it into a 'second' hard disk such as /dev/sdb to verify that it doesn't use the default one :) Thanks!

schaefi:
> Just in case it would be nice to test installing it into a 'second' hard disk such as /dev/sdb to verify that it doesn't use the default one :) Thanks!

yep good idea. I did this as follows

```
qemu-img create mydisk_a 20g
qemu-img create mydisk_b 20g

kvm -cdrom kiwi-test-image-disk.x86_64-1.42.1-Build62.13.install.iso -hda mydisk_a -hdb mydisk_b -boot d

and on the cmdline:

rd.kiwi.oem.installdevice=/dev/sdb

then

kvm -hda mydisk_b
```

* and it boots with the system deployed to the second disk
* the device switch via hda and hdb does not impact the system boot as the UUID of the deployed root device did not change, so it's still safe to deploy a target using its unix node name even if that is not a persistent device name

So I think it does what we wanted it to do :)

PR #2399, opened 2023-11-23
Fix typo in workflow overview doc section

ahresse:
I think there is a typo in here.


iammattcoleman (approved):

schaefi:
> I think there is a typo in here.

I think you would find a lot more :grin: Thanks much Matt :+1:

PR #2401, opened 2023-12-06
Don't consider missing partprobe a dracut error

### LINKED ISSUE #2400: dracut-install: ERROR: installing 'partprobe'
The `dracut-kiwi-oem-repart` package doesn't depend on `parted` but `99kiwi-lib/module-setup.sh` installs it as non-optional component. Apparently the module uses `partx` if `partprobe` is not installed, so it should probably check which of them to install.

https://github.com/OSInside/kiwi/blob/39fde530f606d71e013c686b08689736df91e323/dracut/modules.d/99kiwi-lib/module-setup.sh#L20

schaefi:
The code in the 99kiwi-lib dracut module can either use partprobe or partx to do the job. We prefer partx over partprobe and want the error on a missing partx but not on a missing partprobe. This Fixes #2400

### DIFF
--- a/dracut/modules.d/99kiwi-lib/module-setup.sh
+++ b/dracut/modules.d/99kiwi-lib/module-setup.sh
@@ -15,9 +15,9 @@ depends() {
 install() {
     declare moddir=${moddir}
     inst_multiple \
-        blkid partx blockdev dd mkdir rmdir \
+        blkid blockdev dd mkdir rmdir \
         grep cut tail head tr bc true false mountpoint \
-        basename partprobe sfdisk sgdisk mkswap readlink lsblk \
+        basename sfdisk sgdisk mkswap readlink lsblk \
         btrfs xfs_growfs resize2fs \
         e2fsck btrfsck xfs_repair \
         vgs vgchange lvextend lvcreate lvresize pvresize \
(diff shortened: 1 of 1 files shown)


davidcassany:
> Wouldn't it make sense to have `inst_multiple -o partx partprobe` and separately make an error if either one were missing?

Does dracut offer some utilities to do so? or shall we code our own validation, on a quick look I have not found if `inst_module -o` reports if it succeeded or not.

Conan-Kudo:
We would need to write a conditional to check ourselves.

schaefi:
I wanted to prevent adding self written code to do the probing and thought we can live with the capabilities of inst-multiple. I'm also not sure if we will find an exit code in the module_setup that will lead to an exit criteria for dracut in any case. @davidcassany I remember you investigated on the error codes in the past, do you still have it in your mind ? Thanks

schaefi:
I added a patch to check ourselves and read how upstream 00systemd/module-setup.sh does it. Please review again. Thanks

Vogtinator (approved):
FWICT this is the best way. Dracut is just a bit weird, otherwise `set -e` would've been enough...

Conan-Kudo (approved):

PR #2403, opened 2023-12-11
Set kpartx as default mapper tool for s390

### LINKED ISSUE #2277: Cant build image by new version kiwi
## Problem description
After update kiwi 9.24.41 -> 9.24.57 build failed with error:

```
...
[  192s] [ INFO    ]: 15:14:50 | --> EFI partition adding 20 MB
[  192s] [ INFO    ]: 15:14:50 | Using calculated disk size: 2237 MB
[  192s] [ INFO    ]: 15:14:50 | Creating raw disk image /usr/src/packages/KIWI-oem/kiwi-test-image-live-disk-v8.x86_64-1.3.0.raw
[  192s] [ DEBUG   ]: 15:14:50 | EXEC: [qemu-img create /usr/src/packages/KIWI-oem/kiwi-test-image-live-disk-v8.x86_64-1.3.0.raw 2237M]
[  192s] [ DEBUG   ]: 15:14:50 | EXEC: [losetup -f --show /usr/src/packages/KIWI-oem/kiwi-test-image-live-disk-v8.x86_64-1.3.0.raw]
[  192s] [ DEBUG   ]: 15:14:50 | Initialize gpt disk
[  192s] [ DEBUG   ]: 15:14:50 | EXEC: [sgdisk --zap-all /dev/loop0]
[  193s] [ INFO    ]: 15:14:51 | --> creating EFI CSM(legacy bios) partition
[  193s] [ DEBUG   ]: 15:14:51 | EXEC: [sgdisk -n 1:2048:+2M -c 1:p.legacy /dev/loop0]
[  194s] [ DEBUG   ]: 15:14:52 | EXEC: [sgdisk -t 1:EF02 /dev/loop0]
[  196s] [ INFO    ]: 15:14:54 | --> creating EFI partition
[  196s] [ DEBUG   ]: 15:14:54 | EXEC: [sgdisk -n 2:0:+20M -c 2:p.UEFI /dev/loop0]
[  197s] [ DEBUG   ]: 15:14:55 | EXEC: [sgdisk -t 2:EF00 /dev/loop0]
[  198s] [ INFO    ]: 15:14:56 | --> using all_freeMB for the root(rw) partition if present
[  198s] [ INFO    ]: 15:14:56 | --> creating root partition [with 0 clone(s)]
[  198s] [ DEBUG   ]: 15:14:56 | EXEC: [sgdisk -n 3:0:0 -c 3:p.lxroot /dev/loop0]
[  199s] [ DEBUG   ]: 15:14:57 | EXEC: [sgdisk -t 3:8300 /dev/loop0]
[  200s] [ DEBUG   ]: 15:14:58 | EXEC: [partx --add /dev/loop0]
[  200s] [ DEBUG   ]: 15:14:58 | EXEC: Failed with stderr: partx: /dev/loop0: error adding partitions 1-3
[  200s] , stdout: (no output on stdout)
[  200s] [ ERROR   ]: 15:14:58 | KiwiCommandError: partx: stderr: partx: /dev/loop0: error adding partitions 1-3
[  200s] , stdout: (no output on stdout)
[  200s] [ INFO    ]: 15:14:58 | Cleaning up LoopDevice instance
[  200s] [ DEBUG   ]: 15:14:58 | EXEC: [losetup -d /dev/loop0]
[  200s] [ INFO    ]: 15:14:58 | Cleaning up BootImageDracut instance
[  200s] [ INFO    ]: 15:14:58 | Cleaning up BootLoaderConfigGrub2 instance
[  200s] 
[  200s] obs01p failed "build appliance.kiwi" at Fri Apr  7 15:14:58 UTC 2023.
```

## Expected behaviour
by kiwi 9.24.41 built fine

* KIWI version:  9.24.57
* Operating system target version:  Oracle Linux 7 & 8
* Open Build Service version (N/A if not using OBS):  2.10.15

schaefi:
Some time ago we moved the default partition mapper from kpartx to partx to reduce a package dependencies. Some months and issues later this change will now be reverted for the following reason. partx exists in util-linux in different versions and with different feature sets for different distributions. This makes it very hard to provide a consistently working partition mapper utility across the distributions in the way we had it with kpartx before. Changing the default for a tool that has issues and less support as the one before is a bad move and causes trouble to our users.
This Fixes #2277

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -1791,6 +1791,9 @@ class Defaults:
 
         :rtype: str
         """
+        host_architecture = Defaults.get_platform_name()
+        if 's390' in host_architecture:
+            return 'kpartx'
         return 'partx'
 
     @staticmethod
--- a/test/unit/runtime_config_test.py
+++ b/test/unit/runtime_config_test.py
@@ -78,7 +78,11 @@ class TestRuntimeConfig:
         ]
 
     @patch('kiwi.runtime_checker.Defaults.is_buildservice_worker')
-    def test_config_sections_defaults(self, mock_is_buildservice_worker):
+    @patch('kiwi.runtime_checker.Defaults.get_platform_name')
+    def test_config_sections_defaults(
+        self, mock_get_platform_name, mock_is_buildservice_worker
+    ):
+        mock_get_platform_name.return_value = 's390x'
         mock_is_buildservice_worker.return_value = True
         with patch.dict('os.environ', {'HOME': '../data/kiwi_config/defaults'}):
             runtime_config = RuntimeConfig(reread=True)
@@ -93,10 +97,15 @@ class TestRuntimeConfig:
         assert runtime_config.get_container_compression() is True
         assert runtime_config.get_iso_tool_category() == 'xorriso'
         assert runtime_config.get_oci_archive_tool() == 'umoci'
-        assert runtime_config.get_mapper_tool() == 'partx'
+        assert runtime_config.get_mapper_tool() == 'kpartx'
         assert runtime_config.get_package_changes() is False
         assert runtime_config.\
             get_credentials_verification_metadata_signing_key_file() == ''
+
+        mock_get_platform_name.return_value = 'x86_64'
+        with patch.dict('os.environ', {'HOME': '../data/kiwi_config/defaults'}):
+            runtime_config = RuntimeConfig(reread=True)
+        assert runtime_config.get_mapper_tool() == 'partx'
 
     def test_config_sections_invalid(self):
         with patch.dict('os.environ', {'HOME': '../data/kiwi_config/invalid'}):


Conan-Kudo (changes requested):
I don't think this is the appropriate response to the bug report. Nothing we're doing is particularly complex, and we explicitly changed this to make it easier by default to run in a container.

The bug this references seems to be related to something that the OBS builder VM does, so I don't think the problem is with us.

schaefi:
The problem here is that if we change to a tool to make one area easier and broke another completely we should revert the decision. So partx in some versions has no support for other disk types e.g DASD on s390. Thus this default broke the build for an entire architecture which is not acceptable imho. Also I got reports from people who uses a non standard start sector for the disk and also had issues when partx did the mapping. So the report I referenced here is just one out of a collection of others who had trouble.

I thought it's better to switch to a default that has proven to work and allow to switch to partx for simplicity e.g in the container world rather than vice versa

If this is not acceptable than we should at least move the default to kpartx for the s390 architecture

Thoughts ?

schaefi:
I don't believe OBS is a problem here. The setup of the mapping device nodes was no problem. Issues appeared when using the created device nodes and that's usually a kernel, ioctl or mapper code issue which is not related to the obs worker setup

Conan-Kudo:
> So partx in some versions has no support for other disk types e.g DASD on s390.

Why does `partx` care about disk types? That's all in the kernel, isn't it?

schaefi:
> Why does `partx` care about disk types? That's all in the kernel, isn't it?

To my understanding the partition mapper needs to understand the partition table, read the offsets and creates a device map starting at the correct offsets per partition. In case of DASD kpartx does it right, partx does something but wrong. On the subsequent processing all kinds of issues arises because the data behind the mapped device is just bogus

Conan-Kudo:
Is there a bug filed with [util-linux](https://github.com/util-linux/util-linux/issues) about it?

schaefi:
> Is there a bug filed with [util-linux](https://github.com/util-linux/util-linux/issues) about it?

In the response on the bugzilla conversation at SUSE, Hannes Reinecke mentioned that upstream util-linux seems to work. I could not find a bug report that specifically fixes this particular issue though. Even if upstream works we have a number of distros out there that will not be able to build for this architecture.

I'm going to change the PR here to change the default to kpartx for s390 only. I think this is a fair compromise given that updates to these platforms usually moves at the speed of tectonic plates ;)

schaefi:
@Conan-Kudo ok so the current code changes the default to kpartx for s390 which is an important move as I said on the enterprise end, the move to partx broke all projects. So this needs to change even though an upstream update of util-linux will fix it I don't see that we can wait for this update to happen, we need to move quicker.

On the other this now still leaves the reporter from issue #2277 in a trouble state. As he says that he builds the image on his host there is no obs or koji involved. We have integration tests that works on leap as well as on TW with partx here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:leap/test-image-disk
* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-disk

It's obvious that partx is the reason why it fails for him but I don't understand why.

Please let us clarify this first with help from the reporter and then decided how to go forward.

Thanks

Conan-Kudo:
> Please let us clarify this first with help from the reporter and then decided how to go forward.

Yes, I want to understand what's happening here before we go further, that said this is good to go since we do need s390x to work properly.

Conan-Kudo:
Can you please squash the two commits and rewrite the commit message for indicating just that we're changing it for s390x?

schaefi:
> Can you please squash the two commits and rewrite the commit message for indicating just that we're changing it for s390x?

@Conan-Kudo done

Conan-Kudo (approved):

PR #2405, opened 2023-12-13
package: Drop unused xorriso dep on dracut-kiwi-live subpackage

### LINKED ISSUE #2404: dracut-kiwi-live depends on xorriso
The `dracut-kiwi-live` package has `Requires:       xorriso` which means that `xorriso` and its dependencies are installed inside the live system. FWICT this is unnecessary.

* KIWI version: 9.25.19
* Operating system host version:  openSUSE Tumbleweed
* Operating system target version: openSUSE Tumbleweed

Conan-Kudo:
We do not actually use xorriso anywhere inside of the dracut module, nor do we pull in any utilities from the xorriso package into the generated initramfs anyway.

Fixes #2404

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -561,7 +561,6 @@ Requires:       dracut-network
 Requires:       device-mapper
 %endif
 Requires:       dracut
-Requires:       xorriso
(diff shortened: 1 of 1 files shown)


schaefi (approved):
Thanks :+1:

PR #2406, opened 2023-12-15
Make sure selinux policy is effectively applied

schaefi:
setup_selinux_file_contexts is now called after the config.sh script. This makes sure that eventual policy related changes done in the optional config.sh are covered by a late setfiles call. In addition setup_selinux_file_contexts is called again at the end of any chroot based script hook. So we assume that any optional script target can change the system in a way that a new setfiles call might be required. It can happen that setfiles is called more often than required but as we cannot know what custom scripts does, it's better to call it more often compared to not often enough. This Fixes bsc#1210604

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -549,7 +549,12 @@ class SystemSetup:
         """
         security_context = '/etc/selinux/targeted/contexts/files/file_contexts'
         if os.path.exists(self.root_dir + security_context):
-            self.set_selinux_file_contexts(security_context)
+            if Path.which(filename='setfiles', access_mode=os.X_OK):
+                self.set_selinux_file_contexts(security_context)
+            else:
+                log.warning(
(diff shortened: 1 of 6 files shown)


Vogtinator:
Does this effectively run after `fstab.script` now? If so, should work

schaefi:
> Does this effectively run after `fstab.script` now? If so, should work

Good point, the fstab script execution is done differently than all others, this also needs to change. I will add that.

schaefi:
@Vogtinator ok done, like with any other script hook the fstab script is called through the same method and runs setfiles after execution if there is a policy and the tooling

schaefi:
I have to test the change to make sure no regressions are introduced

Conan-Kudo (dismissed):

schaefi:
Tested the code and works for me now

Conan-Kudo (approved):

PR #2409, opened 2023-12-26
Allow target_blocksize to be available for kis image

lujianhua:
target_blocksize attribute shouble be also avaliable for kis image, so add kis to target_blocksize match pattern.


Conan-Kudo (approved):
This seems okay to me.

schaefi:
Yes this change makes sense as we effectively use the target blocksize for the filesystem image component of a kis image. Good catch :+1:

schaefi (approved):
lgtm

PR #2413, opened 2023-12-29
Switch to pyproject.toml

schaefi:
Calling ```setup.py``` is marked obsolete for a long time and with this commit ```pyproject.toml``` is introduced and used for the development environment setup. On the packaging side the spec file is kept compatible with setup.py in a way that we try the ```build``` and ```installer``` python modules first and only if those are not present setup.py is called. For the publishing on pypi both targets wheel and sdist are used which is the default action of ```python -m build```.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -6,3 +6,5 @@ tag = True
 [bumpversion:file:kiwi/version.py]
 
 [bumpversion:file:doc/source/conf.py]
+
+[bumpversion:file:kiwi/pyproject.toml]
--- a/.virtualenv.dev-requirements.txt
+++ b/.virtualenv.dev-requirements.txt
@@ -6,6 +6,8 @@
 
 # setuptools for pypi upload
 setuptools
+build
+installer
 
 # virtualenv-based automation of test activities
 tox
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -37,12 +37,10 @@ include tox.ini
 include kiwi.yml
 
 recursive-include doc/build/man *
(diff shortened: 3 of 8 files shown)


schaefi:
This PR is a followup from the conversation on #2376

Conan-Kudo:
If we want to start working on 10, maybe we should endeavor to close out the remaining stuff targeting master and branch that for kiwi-9 so master can be 10. If we do that, then I can just go ahead and make a pull request to convert the project to poetry and drastically reduce our host distro support.

schaefi:
> If we want to start working on 10, maybe we should endeavor to close out the remaining stuff targeting master and branch that for kiwi-9 so master can be 10. If we do that, then I can just go ahead and make a pull request to convert the project to poetry and drastically reduce our host distro support.

Problem here is that the master branch is used for publishing the docs, to pypi, for obs CI etc etc. All these systems would need to change their branch reference. I would like to avoid that because I'm sure mistakes will happen and they are painful. I thought we create a kiwi-10 branch (already done) and perform the changes we want over time. I'm confident that regular rebase of the kiwi-10 to master will not be painful.

I can imagine some conversation on the topic so let's do that in the next community meeting

Conan-Kudo:
Maybe we can use kiwi 10 as the point to switch to "main" then, since I'll be changing almost everything about how this project builds...

schaefi:
> Maybe we can use kiwi 10 as the point to switch to "main" then, since I'll be changing almost everything about how this project builds...

That sounds like a nice idea

schaefi:
@Conan-Kudo I'm closing this branch now as we agreed in the meeting and I assume you will tackle the move to poetry in #2370. Thanks much

Issue #2412, opened 2023-12-29
Stop using __del__ for cleanup and switch to context managers


dcermak:
## Problem description

Currently, the kiwi code is relying on `__del__` for cleanup. While this is a typical idiom in C++ (RAII), it is highly uncommon in Python or garbage collected languages in general, as there are generally no guarantees when or if a deleter is ever run. We ensure this in kiwi at the moment via explicit calls to `del foo`.

This is undesirable for multiple reasons:
- unpythonic, i.e. API users will simply not think about explicitly deleting the object, because it is atypical
- thrown exceptions result in cleanup code never running or running in the wrong order

I would thus suggest that we switch to context managers instead.

As this is a potentially breaking API change, we might have to defer this to kiwi 10

schaefi:
Definitely a kiwi-10 topic

I disagree with what you said on "no guarantees when or if a deleter is ever run". Python is an OO language and it's clearly defined when the destructor (__del__) is called and that also works in the designed way. The cases in code where we intentionally call ```del``` on the object instance have their reason because we want the destructor to run prior it going out of scope by the flow of code. You can argue that the implementation which requires an explicit ```del``` is questionable and I would agree with you on that point but the OO design pattern in python works well :)

dcermak:
> I disagree with what you said on "no guarantees when or if a deleter is ever run". Python is an OO language and it's clearly defined when the destructor (`__del__`) is called and that also works in the designed way. 

This is incorrect: https://docs.python.org/3/reference/datamodel.html#object.__del__:
> It is not guaranteed that [__del__()](https://docs.python.org/3/reference/datamodel.html#object.__del__) methods are called for objects that still exist when the interpreter exits.
 
> The cases in code where we intentionally call `del` on the object instance have their reason because we want the destructor to run prior it going out of scope by the flow of code.

This only invokes `__del__` if the reference count reached zero (see the above link). Hopefully that's the case wherever we use finalizers, but it is **not** guaranted.

> You can argue that the implementation which requires an explicit `del` is questionable and I would agree with you on that point but the OO design pattern in python works well :)

Good that we agree, but note that this type of design pattern is explicitly discouraged in Python due to reliance on GC running. In Python Context Managers were added for RAII, relying on finalizers is too brittle.

schaefi:
> relying on finalizers is too brittle.

ok, thanks for the pointers, makes sense

schaefi:
[...]

PR #2416, opened 2023-12-30
Move LoopDevice class to context manager

schaefi:
Change the LoopDevice class to be a context manager. All code using LoopDevice was updated to the following with statement:

```python
with LoopDevice(...) as loop_provider:
    loop_provider.some_member()
```

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 9.25.22
+current_version = 10.0.0
 commit = True
 tag = True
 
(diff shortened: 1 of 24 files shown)


schaefi:
@dcermak I believe this is what you have in mind regarding the context manager refactoring ? making these changes of course has a big impact on the code using the classes. I started with ```LoopDevice``` which is the entry point for alomst everything, thus the biggest change. Others will follow but I'd like you to take a look first if the direction is correct. Thanks

schaefi:
I did some test builds with the change and so far works for me

PR #2418, opened 2024-01-04
Fix overwrite of kiwi_oemunattended

schaefi:
In case rd.kiwi.oem.installdevice is set, there is an overwrite of the kiwi_oemunattended setting. However the variable was set in local scope of a function and therefore the change was not effective in other methods which also evaluates this variable. This commit fixes it such that the overwrite happens in the early initialize method which provides the environment for all code running in the dracut module. This is related to jira#PED-7180


rdoxenham (approved):
Confirming that this works, thanks so much!

PR #2420, opened 2024-01-08
Fixed grub terminal setup

### LINKED ISSUE #2419: Setup of GRUB_TERMINAL is invalid
## Problem description

report from @leifliddy on the matrix chat:

I'm looking at the kiwi documentation here -- which shows which values can be specified by console

```
console=”console|gfxterm|serial”
```

kiwi puts that value under GRUB_TERMINAL the problem with this is that GRUB_TERMINAL overrides both GRUB_TERMINAL_OUTPUT and GRUB_TERMINAL_INPUT If you look at the grub manual you'll see there are some options value available for GRUB_TERMINAL_OUTPUT that are not available for GRUB_TERMINAL_INPUT

* https://www.gnu.org/software/grub/manual/grub/grub.html

I tried setting console=gfxterm but receive an error at bootup stating
error: ../../grub-core/commands/terminal.c:138:terminal 'gfxterm isn't found.....and that's because it's not a supported option for GRUB_TERMINAL_INPUT

I mean gfxterm is literally the default value on most systems for GRUB_TERMINAL_OUTPUT -- perhaps kiwi should not use GRUB_TERMINAL at all as there's way too many differences between what GRUB_TERMINAL_INPUT and GRUB_TERMINAL_OUTPUT support

schaefi:
The grub terminal setup is divided into the setting for the output and the input console. For both settings different parameters exists. So far kiwi did not differentiate between the two parts of the console setup and that could lead to a wrong setting if only one value is provided in kiwi's console= attribute which lead to the grub setting, GRUB_TERMINAL=value. If value is set to e.g gfxterm grub takes this for both input and output and it's obviously wrong for the input.

To make this less error prune the kiwi code changes with this commit to set GRUB_TERMINAL_INPUT and GRUB_TERMINAL_OUTPUT rather than GRUB_TERMINAL and also runs sanity checks on the provided values if they are applicable. The information for setting up the console in the schema stays untouched though. That's because it's used for all bootloaders and also because grub supports multiple values for the console in/out setting in one GRUB_TERMINAL variable even though kiwi does no longer use it.

To make this clear for the users also the documentation for the console attribute setup has been updated. If we want to wish two distinct attributes for input and output console settings a schema change and also differentiation between bootloaders is needed and that I only see for the kiwi-10 branch if at all. 

This Fixes #2419


Conan-Kudo:
I think we should also consider for kiwi-10 dropping isolinux support. The syslinux project has been dead for a long time, and neither Fedora nor openSUSE use it for anything anymore.

Conan-Kudo (approved):

kiwi/bootloader/template/grub2.py
```diff
@@ -73,8 +73,8 @@ def __init__(self):
         ''').strip() + os.linesep
 
         self.header_terminal_setup = dedent('''
-            terminal_input ${terminal_setup}
-            terminal_output ${terminal_setup}
+            terminal_input ${terminal_input}
```
Vogtinator, line 76: This breaks if `terminal_input` or `terminal_output` are empty

schaefi: If not set this leads to the following settings in the template
```
...
set timeout=10
set timeout_style=menu
terminal_input 
terminal_output 
menuentry ...
```

According to the grub docs this leads to the default settings for in and output. In my tests it does that and did not cause an issue. Do you have different experience ?

PR #2423, opened 2024-01-09
Move loop device to context manager

schaefi:
Change the LoopDevice class to be a context manager. All code using LoopDevice was updated to the following
with statement:
   
```python
with LoopDevice(...) as loop_provider:
    loop_provider.some_member()
```
    
### DIFF
--- a/kiwi/storage/loop_device.py
+++ b/kiwi/storage/loop_device.py
@@ -23,9 +23,7 @@ from kiwi.command import Command
 from kiwi.storage.device_provider import DeviceProvider
 from kiwi.utils.command_capabilities import CommandCapabilities
 
-from kiwi.exceptions import (
-    KiwiLoopSetupError
-)
+from kiwi.exceptions import KiwiLoopSetupError
 
 log = logging.getLogger('kiwi')
 
@@ -50,6 +48,9 @@ class LoopDevice(DeviceProvider):
         self.filename = filename
         self.filesize_mbytes = filesize_mbytes
         self.blocksize_bytes = blocksize_bytes
+
+    def __enter__(self):
+        return self
 
     def get_device(self) -> str:
         """
@@ -98,12 +99,13 @@ class LoopDevice(DeviceProvider):
         )
         self.node_name = loop_call.output.rstrip(os.linesep)
 
-    def __del__(self):
+    def __exit__(self, exc_type, exc_value, traceback):
         if self.node_name:
-            log.info('Cleaning up %s instance', type(self).__name__)
             try:
                 Command.run(['losetup', '-d', self.node_name])
-            except Exception:
-                log.warning(
-                    'loop device %s still busy', self.node_name
+            except Exception as issue:
+                log.error(
+                    'loop cleanup on {0} failed with: {1}'.format(
+                        self.node_name, issue
+                    )
                 )
--- a/doc/source/contributing/kiwi_from_python.rst
+++ b/doc/source/contributing/kiwi_from_python.rst
@@ -56,15 +56,15 @@ which contains your host `tmp` directory.
     from kiwi.storage.loop_device import LoopDevice
     from kiwi.filesystem import FileSystem
 
-    loop_provider = LoopDevice(
+    with LoopDevice(
         filename='my_tmp.ext4', filesize_mbytes=100
-    )
-    loop_provider.create()
+    ) as loop_provider:
+        loop_provider.create()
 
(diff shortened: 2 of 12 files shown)


kiwi/storage/loop_device.py
```diff
@@ -98,12 +101,13 @@ def create(self, overwrite: bool = True):
         )
         self.node_name = loop_call.output.rstrip(os.linesep)
 
-    def __del__(self):
+    def __exit__(self, exc_type, exc_value, exc_traceback):
         if self.node_name:
-            log.info('Cleaning up %s instance', type(self).__name__)
             try:
                 Command.run(['losetup', '-d', self.node_name])
             except Exception:
```
dcermak, line 108:
```suggestion
            except Exception as exc:
```

kiwi/storage/loop_device.py
```diff
@@ -98,12 +101,13 @@ def create(self, overwrite: bool = True):
...
                 log.warning(
-                    'loop device %s still busy', self.node_name
+                    f'loop device {self.node_name} still busy'
```
dcermak, line 110:
```suggestion
                    f'loop device {self.node_name} still busy, got error: %s', exc
```

kiwi/storage/loop_device.py
```diff
@@ -98,12 +101,13 @@ def create(self, overwrite: bool = True):
         )
         self.node_name = loop_call.output.rstrip(os.linesep)
```
dcermak, line 102: Would it make sense to move `create` into `__init__` or `__enter__`?

From my PoV there's little use for a `LoopDevice` that is not created. However, I am unsure whether this would be a good API or whether that is considered good python style

kiwi/storage/loop_device.py
```diff
@@ -98,12 +101,13 @@ def create(self, overwrite: bool = True):
...
-            log.info('Cleaning up %s instance', type(self).__name__)
             try:
                 Command.run(['losetup', '-d', self.node_name])
             except Exception:
                 log.warning(
-                    'loop device %s still busy', self.node_name
+                    f'loop device {self.node_name} still busy'
                 )
+            # exception got handled, return gracefully
+            return True
```
dcermak, line 113: This is dangerous, returning `True` from `__exit__()` indicates that the exception thrown in the `with` block got handled. I.e. you are hereby swallowing **all** exceptions raised in every `with LoopDevice() as loop_provider:` block. I am pretty certain we do **not** want that.

```suggestion
```

dcermak (changes requested):
There's one potentially dangerous change in `__exit__()`, besides that LGTM + a few nits/ideas.

kiwi/storage/loop_device.py
schaefi: yep makes a lot sense, I fixed it

kiwi/storage/loop_device.py
schaefi: done

kiwi/storage/loop_device.py
schaefi: That part I'd like to keep for now. Reason is mostly consistency. All Device classes works in the way that you can create the object instance and it does not immediately run the device magic but there is a method to initiate it. For the loop device class it's true there is not much benefit but other device classes provides methods that could be called prior the creation/initialization code. As a followup refactoring we can look again into this topic but for the move from finalizers to context manager I think the basic class behavior should not change

kiwi/storage/loop_device.py
```diff
@@ -98,12 +99,13 @@ def create(self, overwrite: bool = True):
...
             try:
                 Command.run(['losetup', '-d', self.node_name])
-            except Exception:
-                log.warning(
-                    'loop device %s still busy', self.node_name
+            except Exception as issue:
+                log.error(
+                    'loop cleanup on {0} failed with: {1}'.format(
+                        self.node_name, issue
+                    )
```
dcermak, line 110: you don't like f-strings, do you :stuck_out_tongue: ?
```suggestion
                    f"loop cleanup on {self.node_name} failed with: {issue}'
                    )
```

dcermak:
LGTM, but there are merge conflicts now

kiwi/storage/loop_device.py
schaefi: I like them, but I don't like the long lines and if there are many substitutions in it. In this case I think it's an easier read to use format() :smile_cat:

schaefi:
I'll fix the merge conflict

Issue #2335, opened 2023-07-25
CentOS 9 Images in the staging project are unresolvable


dcermak:
## Problem description

All CentOS 9 images from https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:centos/test-image-live-disk-v9 are unresolvable:
```
 unresolvable: nothing provides kbd-legacy = 2.4.0-9.el9 needed by kbd, (got version 2.4.0-8.el9) 
```

Conan-Kudo:
That sounds like OBS is failing to sync the repos internally again. 

cc: @adrianschroeter

schaefi:
yes something is wrong. I sent an e-mail  to the OBS list, no answer so far. Let's wait. This one should self heal and is not something we can do much about it other than reporting

schaefi:
@Conan-Kudo We are getting ignored on that one... no idea how to fix it

PR #2426, opened 2024-01-10
Use obsrepositories for the CentOS 9 image

dcermak:
We were relying on the repos being parsed from the kiwi description but this can cause problems and gives us less flexibility when it comes to fixing repository setup issues.

This fixes https://github.com/OSInside/kiwi/issues/2335


schaefi (approved):
lgtm, thanks

PR #2427, opened 2024-01-11
Drop clicfs

schaefi:
This patch is two fold, please review commit based

**Drop clicfs support**
    
clicfs was a fuse userspace implementation of an overlay filesystem. Since overlayfs became the default overlay filesystem there is no reason for kiwi to keep the clicfs support

**Update major schema version from 7.6 -> 8.0**
    
With the switch to schema v8.x we allow for non compatible changes in regards to the v7.x stream. Thus this commit also     deletes the support for clicfs from the schema


schaefi:
Please ignore:

* xml_parse.py
* kiwi.rng

Those files are auto generated

Conan-Kudo (approved):

PR #2428, opened 2024-01-12
Move FileSystem to context manager

schaefi:
Change the FileSystem class to be a context manager. All code using FileSystem was updated to the following with statement:

with FileSystem.new(...) as filesystem:
    filesystem.some_member()

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -18,7 +18,7 @@
 import os
 import logging
 from typing import (
-    Dict, List, Optional, Tuple, Any
+    Dict, List, Optional, Tuple, Union
 )
 
 # project
@@ -49,6 +49,7 @@ from kiwi.storage.device_provider import DeviceProvider
 from kiwi.filesystem import FileSystem
 from kiwi.filesystem.squashfs import FileSystemSquashFs
 from kiwi.volume_manager import VolumeManager
+from kiwi.volume_manager.base import VolumeManagerBase
 from kiwi.command import Command
 from kiwi.system.setup import SystemSetup
 from kiwi.builder.install import InstallImageBuilder
@@ -234,7 +235,8 @@ class DiskBuilder:
         # an instance of a class with the sync_data capability
         # representing the entire image system except for the boot/ area
         # which could live on another part of the disk
-        system: Any = None
+        system: Union[FileSystemBase, VolumeManagerBase] = \
+            FileSystemBase(DeviceProvider())
 
         # an instance of a class with the sync_data capability
         # representing the boot/ area of the disk if not part of
@@ -644,7 +646,7 @@ class DiskBuilder:
         self,
         device_map: Dict,
         disk: Disk,
-        system: Any,
+        system: Union[FileSystemBase, VolumeManagerBase],
         system_boot: Optional[FileSystemBase],
         system_efi: Optional[FileSystemBase],
         system_spare: Optional[FileSystemBase],
@@ -1136,13 +1138,15 @@ class DiskBuilder:
             )
 
     def _write_generic_fstab_to_system_image(
-        self, device_map: Dict, system: Any
+        self, device_map: Dict,
+        system: Union[FileSystemBase, VolumeManagerBase]
     ) -> None:
         log.info('Creating generic system etc/fstab')
         self._write_generic_fstab(device_map, self.system_setup, system)
 
     def _write_generic_fstab_to_boot_image(
-        self, device_map: Dict, system: Any
+        self, device_map: Dict,
+        system: Union[FileSystemBase, VolumeManagerBase]
     ) -> None:
         if self.initrd_system == 'kiwi':
             log.info('Creating generic boot image etc/fstab')
@@ -1152,7 +1156,7 @@ class DiskBuilder:
 
     def _write_generic_fstab(
         self, device_map: Dict, setup: SystemSetup,
-        system: Any
+        system: Union[FileSystemBase, VolumeManagerBase]
     ) -> None:
         root_is_snapshot = \
             self.xml_state.build_type.get_btrfs_root_is_snapshot()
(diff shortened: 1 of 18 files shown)


kiwi/builder/disk.py
```diff
@@ -444,144 +445,65 @@ def create_disk(self) -> Result:
...
+                device_map,
+                disk,
+                system,
+                system_boot,
+                system_efi,
+                system_spare,
+                system_custom_parts,
+                luks_root,
+                raid_root
+            )
```
dcermak, line 493: This does not look like a good idea. `__exit__()` of Filesystem is called in line 480, but you are reusing the Filesystem instance in the else case in line 486.

kiwi/builder/disk.py
```diff
@@ -718,6 +640,121 @@ def create_install_media(self, result_instance: Result) -> Result:
 
         return result_instance
 
+    def _build_main_system(
+        self,
+        device_map: Dict,
+        disk: Disk,
+        system: Any,
```
dcermak, line 647: Do we really not know the concrete type of `system` here?

kiwi/builder/disk.py
```diff
@@ -814,16 +851,16 @@ def _build_spare_filesystem(self, device_map: Dict) -> Optional[FileSystemBase]:
...
-            )
-            filesystem.create_on_device(
-                label='SPARE'
-            )
-            return filesystem
+            ) as filesystem:
+                filesystem.create_on_device(
+                    label='SPARE'
+                )
+                return filesystem
```
dcermak, line 863: The filesystem is unmounted on return. Are you sure that is the right thing to do?

kiwi/builder/disk.py
```diff
@@ -855,13 +892,13 @@ def _build_boot_filesystems(
...
-            filesystem.create_on_device(
-                label=self.disk_setup.get_boot_label()
-            )
-            system_boot = filesystem
+            ) as filesystem:
+                filesystem.create_on_device(
+                    label=self.disk_setup.get_boot_label()
+                )
+                system_boot = filesystem
         return system_boot, system_efi
```
dcermak, line 902: Same here, `system_boot` and `system_efi` are now no longer mounted. This is changed behavior in contrast to previously, where they were mounted until they are garbage collected.

dcermak (changes requested):
There are a few places that look wonky or dangerous to me as the filesystem would be guaranteed to be unmounted far earlier than it was before (where it might never be unmounted).

kiwi/builder/disk.py
schaefi: no it can be FileSystem or VolumeManager. Dan really to make this absolutely nice a bigger refactoring is required which I'm not starting together with the context manager switch, sorry

kiwi/builder/disk.py
schaefi: yes. I know it's not perfect but the FileSystem class supports filesystems that requires mounting to receive data and filesystems that does not e.g mksquashfs. Thus it is save to leave the context of the filesystem AND keep the instance of filesystem for later actions. I really need that and I cannot change it without further refactoring. The context manager makes sure the filesystem is not left in a mounted state. Any subsequent actions on the instance of filesystem makes sure it gets mounted when needed e.g sync_data().

kiwi/builder/disk.py
schaefi: correct and a wanted behavior. The former implementation left mounts way too long and several cleanup were racy to fail. This is now much better. I have done many tests to the code hitting Ctrl-C in between of operations which left dead loops or mounts, it's now way better

kiwi/builder/disk.py
schaefi: See the other comments which should make this more clear. It's ok and wanted to use the instance of a filesystem after the context exit is done

kiwi/builder/disk.py
schaefi: I have switched all kiwi integration tests to main which allows me to build all the images we have with the changed code. So the version of kiwi in Staging is at 10.x.x and I plan to let it run with each iteration of context manager changes and only continue if there are zero failures.

These type of changes are very intrusive to the code. It's different to run context manager based or trust on finalizers. Thus running all build tests with the pull request is imho the only way to do it without too many regressions. I did so with the LoopDevice and that succeeded completely. I would do it with this PR too

kiwi/builder/disk.py
schaefi: ok that I could fix with a cast()

dcermak: If it's either of the two classes, how about this then:
```suggestion
        system: Union[FileSystem, VolumeManager],
```

kiwi/builder/disk.py
dcermak: Hm, out of interest, if `sync_data()` mounts the file system again, I guess it will remain mounted for ever then, right?

kiwi/builder/disk.py
schaefi: I'll try

kiwi/builder/disk.py
schaefi: You are partially right. The point is, the ```umount``` is not a responsibility of the sync_data() method and that I also consider a design bug. The way the several sync_data() methods are called ensures that at the end (or in case of an exception) a cleanup runs which makes sure we are umounted after sync. It's the builder implementation that has to ensure this and you will find it in each kiwi/builder/... implementation. 

In the beginning I thought it will be one class that is responsible for mounting the system, sync and umount. But this was a bad design decision because over time I understand that we need mount states of the system also after sync, e.g for certain boot-loaders or to apply specific attributes on certain filesystems. At that time I decided for the shortcut, keep the mount state into the builder and let the builder decide for the right time to umount. It works but it's not super nice probably was the wrong decision back in time.

So all this would need a complete refactor of how the builders are implemented. I simply don't have the time to start this massive reconstruction and to be honest I also feel mostly left alone with the actual work. I'd like to finish the context manager move because it will increase stability but I fear along the way and as you read more of my code you will find more stuff you will not like :) The context manager move is already a big big change as it also touches everything

kiwi/builder/disk.py
schaefi: done

kiwi/builder/disk.py
schaefi: @dcermak so from the perspective of the context manager initiative, can this go on ? See the build matrix it all builds. I'm fine continuing on further refactoring tasks but I also like to move forward with this one here. Thanks

dcermak (approved):

PR #2429, opened 2024-01-12
Reduce XSL transformation list

### LINKED ISSUE #2368: Reduce XSL transformation list
## Problem description

So far we support transformation of XML schema versions from the beginning of time. These transformation steps takes time in reading as they are all based on XSL recursions. I think with the next major release we can consider to reduce the
entry version for supported schemas

schaefi:
Minimum schema version to start with is v74
This Fixes #2368


Conan-Kudo (approved):

Conan-Kudo:
It might be worth considering splitting this out somewhere else and providing instructions on how to manually run them.

PR #2434, opened 2024-01-17
Refactor disk builder for use with context manager

schaefi:
In preparation to further context manager related changes in VolumeManager, LuksDevice, RaidDevice and more the disk builder code which uses these classes needs to be refactored beforehand to allow switching to context manager based cascading of the storage device classes. This commit does the refactoring and is related to Issue #2412

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -16,10 +16,15 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
+import sys
 import logging
 from typing import (
     Dict, List, Optional, Tuple, Union
 )
+if sys.version_info >= (3, 8):
+    from typing import TypedDict  # pragma: no cover
+else:  # pragma: no cover
+    from typing_extensions import TypedDict  # pragma: no cover
 
 # project
 import kiwi.defaults as defaults
@@ -31,6 +36,7 @@ from kiwi.storage.disk import ptable_entry_type
 from kiwi.defaults import Defaults
 from kiwi.filesystem.base import FileSystemBase
 from kiwi.bootloader.config import BootLoaderConfig
+from kiwi.bootloader.config.base import BootLoaderConfigBase
 from kiwi.bootloader.install import BootLoaderInstall
 from kiwi.system.identifier import SystemIdentifier
 from kiwi.boot.image import BootImage
@@ -69,6 +75,25 @@ from kiwi.exceptions import (
 )
 
 log = logging.getLogger('kiwi')
+
+
+class StorageMap(TypedDict):
+    system: \
+        Optional[Union[FileSystemBase, VolumeManagerBase]]
+    system_boot: \
+        Optional[FileSystemBase]
+    system_efi: \
+        Optional[FileSystemBase]
+    system_spare: \
+        Optional[FileSystemBase]
+    system_custom_parts: \
+        Dict[str, FileSystemBase]
+    luks_root: \
+        Optional[LuksDevice]
+    raid_root: \
+        Optional[RaidDevice]
+    integrity_root: \
+        Optional[IntegrityDevice]
 
 
 class DiskBuilder:
@@ -129,7 +154,7 @@ class DiskBuilder:
         self.luks = xml_state.get_luks_credentials()
         self.use_disk_password = \
             xml_state.get_build_type_bootloader_use_disk_password()
-        self.integrity_root = xml_state.build_type.get_standalone_integrity()
+        self.integrity = xml_state.build_type.get_standalone_integrity()
         self.integrity_legacy_hmac = \
             xml_state.build_type.get_integrity_legacy_hmac()
         self.integrity_keyfile = xml_state.build_type.get_integrity_keyfile()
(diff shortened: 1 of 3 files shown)


schaefi:
@dcermak I put you as a reviewer here but I think this sort of changes are not really reviewable. I plan to create a Staging release with this changes and let the integration tests run to see of any of them fails. So if you immediately spot something stupid that I have done feel free to comment :) Thanks much

kiwi/builder/disk.py
```diff
@@ -68,6 +73,19 @@
...
+        'system': Optional[Union[FileSystemBase, VolumeManagerBase]],
+        'system_boot': Optional[FileSystemBase],
+        'system_efi': Optional[FileSystemBase],
+        'system_spare': Optional[FileSystemBase],
+        'system_custom_parts': Dict[str, FileSystemBase],
+        'luks_root': Optional[LuksDevice],
+        'raid_root': Optional[RaidDevice],
+        'integrity_root': Optional[IntegrityDevice]
+    }
+)
```
dcermak, line 87: The namedtuple like syntax is deprecated as of Python 3.11 and will be removed in python 3.13
```suggestion
class StorageMap(TypedDict):
    system: Optional[Union[FileSystemBase, VolumeManagerBase]],
    system_boot: Optional[FileSystemBase],
    system_efi: Optional[FileSystemBase],
    system_spare: Optional[FileSystemBase],
    system_custom_parts: Dict[str, FileSystemBase],
    luks_root: Optional[LuksDevice],
    raid_root: Optional[RaidDevice],
    integrity_root: Optional[IntegrityDevice]
```

kiwi/builder/disk.py
```diff
@@ -287,225 +327,70 @@ def create_disk(self) -> Result:
...
-            )
+            bootloader_custom_parameters = {
+                'targetbase':
+                    loop_provider.get_device(),
+                'grub_directory_name':
+                    Defaults.get_grub_boot_directory_name(
+                        self.root_dir
+                    ),
+                'crypto_disk':
+                    True if self.luks is not None else False,
```
dcermak, line 338: This should do the trick too
```suggestion
                'crypto_disk': bool(self.luks),
```

dcermak (approved):
Overall LGTM, thanks for this. I've added a few comments but nothing really substantial

kiwi/builder/disk.py
schaefi: I tried this on py3.11 and only got a E999 SyntaxError: invalid syntax

dcermak: That's because I'm stupid and left the trailing commas in there :facepalm:

kiwi/builder/disk.py
```diff
@@ -642,17 +526,243 @@ def create_install_media(self, result_instance: Result) -> Result:
 
         return result_instance
 
+    def _raid_instance(self, device_map: Dict) -> RaidDevice:
+        return RaidDevice(device_map['root'])
+
+    def _luks_instance(self, device_map: Dict) -> LuksDevice:
+        return LuksDevice(device_map['root'])
```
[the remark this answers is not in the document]
schaefi: The reason why I put the construction of instances into an extra method is that I expect those to receive more parameters in the future. When we change to context managers this will result in a nested structure of "with" statements and in contrast to today's implementation there will then be more object constructions of the same class. If it still stays that simple after the refactoring into context managers I will delete the one liner instance creation methods again. It will not hurt because they "sort of" private with the "_" and will not be exposed to the public API, at least not from a documentation perspective

kiwi/builder/disk.py
schaefi: nope, an empty passphrase for luks is allowed and still crypto

```python
luks = ''
print(luks)
x = True if luks is not None else False
print(x)
x = bool(luks)
print(x)
```

I stumbled over exactly that problem in the past ;)

kiwi/builder/disk.py
```diff
@@ -642,17 +526,243 @@ def create_install_media(self, result_instance: Result) -> Result:
...
+                keydescription=self.integrity_key_description,
+                keyfile=self.integrity_keyfile,
+                keyfile_algorithm=defaults.INTEGRITY_KEY_ALGORITHM,
+                options=[
+                    'legacy_hmac'
+                ] if self.integrity_legacy_hmac else []
+            )
+        )
+
+    def _system_instance(self, device_map: Dict) -> None:
```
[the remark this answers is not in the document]
schaefi: yep good point, done

kiwi/builder/disk.py
```diff
@@ -642,17 +526,243 @@ def create_install_media(self, result_instance: Result) -> Result:
...
+            filesystem.create_on_device(
+                label=self.disk_setup.get_root_label(),
+            )
+        self.storage_map['system'] = filesystem
+
+    def _process_build(self, device_map: Dict, disk: Disk) -> None:
+        # representing the entire image system except for the boot/ area
+        # which could live on another part of the disk
+        system: Optional[Union[FileSystemBase, VolumeManagerBase]] = \
+            self.storage_map.get('system')
```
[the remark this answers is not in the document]
schaefi: done

kiwi/builder/disk.py
```diff
@@ -802,47 +913,54 @@ def _build_custom_parts_filesystem(
...
+                                    readonly_target_bytesize,
+                                    'bytes'
                                 )
-                            filesystem_dict[map_name] = filesystem
+                            )
+                            Command.run(
+                                [
+                                    'dd',
+                                    'if=%s' % squashed_root_file.name,
+                                    'of=%s' % readonly_target
```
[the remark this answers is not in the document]
schaefi: lovely f strings :) done

kiwi/builder/disk.py
schaefi: ah yes, thanks :) and done

PR #2435, opened 2024-01-18
Convert the MountManager to context manager

dcermak:


kiwi/mount_manager.py
```diff
@@ -148,10 +155,13 @@ def umount(self, raise_on_busy: bool = True) -> bool:
...
                     break
-                except Exception:
+                # we only want to catch KiwiCommandError, everything else
+                # indicates either a bug in the code or a bigger problem, like
+                # umount not being available
+                except KiwiCommandError as kw_err:
                     log.warning(
-                        '%d umount of %s failed, try again in 1sec',
-                        busy, self.mountpoint
+                        '%d umount of %s failed with %s, try again in 1sec',
```
schaefi, line 163: As you are on it, can you also change the message to either use f-string or format()

kiwi/volume_manager/base.py
```diff
@@ -41,58 +42,61 @@ class VolumeManagerBase(DeviceProvider):
...
-    :param list custom_filesystem_args: custom filesystem creation and mount
-        arguments, subset of the custom_args information suitable to
-        be passed to a FileSystem instance
 
     :raises KiwiVolumeManagerSetupError: if the given root_dir doesn't exist
     """
-    def __init__(self, device_map, root_dir, volumes, custom_args=None):
-        self.temp_directories = []
+
+    def __init__(self, device_map: Dict[str, DeviceProvider], root_dir: str, volumes: List[volume_type], custom_args: Optional[Dict[str, Any]] = None) -> None:
```
schaefi, line 55: Can we please break up the long line a bit, thanks ;)

schaefi (changes requested):
Looks good, just nitpick at two places

I like the ExitStack idea. I'm not sure if the order of umounts in the stack will still be correct for nested mounts but I would test this by running the integration tests.

Is this PR rebased to the current main ? If yes I would zip a new Staging package from your branch and let the integration tests build with it

schaefi (approved):
Looks good :+1:

schaefi:
I will create a new Staging release to see if it breaks things. Until then please don't merge

schaefi:
@dcermak all integration test builds are successful :+1: Can you please squash your commits into a good set and a description to the pull request referencing #2412 would be nice.

Thanks

PR #2437, opened 2024-01-19
[docs] make copyright year dynamic

dcermak:
Just a tiny cosmetic change that the docs show `copyright 2020-$cur_year`


Conan-Kudo:
We should just drop the year instead. We don't need it.

schaefi (approved):
Looks good :+1:

PR #2438, opened 2024-01-19
Move BootLoaderConfig to context manager

schaefi:
Change the BootLoaderConfig class to be a context manager. All code using BootLoaderConfig was updated to the following with statement:

```python
with BootLoaderConfig.new(...) as bootloader_config:
    bootloader_config.some_member()
````

### DIFF
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -23,6 +23,7 @@ import shutil
 # project
 from kiwi.utils.temporary import Temporary
 from kiwi.bootloader.config import BootLoaderConfig
+from kiwi.bootloader.config.base import BootLoaderConfigBase
 from kiwi.filesystem import FileSystem
 from kiwi.filesystem.isofs import FileSystemIsoFs
 from kiwi.filesystem.setup import FileSystemSetup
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by dcermak — COMMENTED** [review]
I think it would be cleaner if the `BootLoaderConfigBase` would use a `contextlib.ExitStack`, as the `volume_mounts` are already context managers

**Comment by schaefi** [comment]
> I think it would be cleaner if the `BootLoaderConfigBase` would use a `contextlib.ExitStack`, as the `volume_mounts` are already context managers

The reason why I did not push it into an ExitStack was because I think the elements of the ExitStack will disappear one after the other when we go forward moving the other classes into context managers. With each follup pull request my plan is to come more and more into that nested structure of context managers which will reduce the existing ExitStack until we reach the point where it is empty.

The current implementation of some builders keeps resources active for way too long. BootLoaderConfig is one example. It was created super early and kept until the end of the complete build process. But within that process there is only one time where it actually operates and depending on the bootloader mounts stuff, does its bootloader config job and should now free the resources but it holds it too long. One of the nice effects of moving into the context managers is to find out how to release as early as possible.

That's why I did not use the ExitStack because for the BootLoaderConfig I knew when I need it and when I can release.

Hope this makes sense to you too ?

**Comment by dcermak** [comment]
> With each follup pull request my plan is to come more and more into that nested structure of context managers which will reduce the existing ExitStack until we reach the point where it is empty.

Well, the idea of the `ExitStack` is that you do **not** have to nest contextmanagers and just have one exitstack that cleans them all up. But I think we can add these later on if they turn out to be useful.

File: kiwi/builder/live.py [hunk 1]
```diff
@@ -146,93 +147,80 @@ def create(self) -> Result:
...
+                )
+
+            IsoToolsBase.setup_media_loader_directory(
+                self.boot_image.boot_root_directory, self.media_dir.name,
+                bootloader_config.get_boot_theme()
             )
-            bootloader_config.setup_live_boot_images(
-                mbrid=self.mbrid, lookup_path=self.root_dir
+            if self.firmware.bios_mode():
+                Iso(self.media_dir.name).setup_isolinux_boot_path()
```
dcermak, line 161: These lines appear exactly the same way in `kiwi/builder/install.py` lines 210 & following

kiwi/builder/live.py
```diff
@@ -359,6 +347,30 @@ def create(self) -> Result:
         )
         return self.result
 
+    def _bootloader_instance(self) -> BootLoaderConfigBase:
```
dcermak, line 350: Maybe use a name that suggests that an instance is explicitly created and this is not just a getter?
```suggestion
    def _create_bootloader_instance(self) -> BootLoaderConfigBase:
```

dcermak (approved):
LGTM, I've added two small nit-picky comments

kiwi/builder/live.py
schaefi: yes correct and intentional because this action needs to be done for building live ISO images (builder/live) and for building install ISO image (builder/install). Those are two different builders but they both produce some sort of ISO, which is the reason why you find the same function call in both builders, is this a problem ?

btw, in the other open PR #2439 these will also disappear

kiwi/builder/live.py
schaefi: yep done

kiwi/builder/live.py
dcermak: Code duplication is generally never a good idea, as changing one instance requires changing all instances.

However, if this code will go, then I think it's fine to keep it as is for now.

PR #2439, opened 2024-01-22
Drop isolinux

schaefi:
Drop support for syslinux/isolinux
    
SYSLINUX has been abandoned for several years now, and it does not even work with modern versions of the filesystems supported in kiwi. This commit drops it and Fixes #2433

PR #2440, opened 2024-01-26
convert BootLoaderConfig to an abstract base class

dcermak:
These are mostly changes wrt style, but they allow us to have more concrete types of the `BootLoaderConfigBase` subclasses.

* convert `BootLoaderConfig.new` to a free function. This class really has no purpose except for being having this one static method. This is very C++ pre 98 when there were no namespaces. There's no reason for replicating this style in Python
* convert `BootLoaderConfigBase` into an abstract base class. Thereby we can ensure that all methods must be implemented. This shows a few shortcomings in the current code: certain methods only make sense for certain bootloaders but have to be implemented by all bootloader-configs. It would imho make sense to further refactor the source code to remove these issues. But that probably only makes sense once #2439 is merged
* a few minor changes that showed up during refactoring (wrong type hint, missing `assert` in a test)

### DIFF
--- a/kiwi/bootloader/config/__init__.py
+++ b/kiwi/bootloader/config/__init__.py
@@ -15,51 +15,61 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
-import importlib
-from typing import Dict
-from abc import (
-    ABCMeta,
-    abstractmethod
-)
+from typing import TYPE_CHECKING, Dict, Literal, Union, overload
 
 from kiwi.exceptions import (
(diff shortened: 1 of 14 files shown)


Conan-Kudo:
I'm blocking this on #2439.

schaefi:
@dcermak great effort thanks :+1: , let's wait for #2439 to be merged and then I fear you need to fix inevitable conflicts ;)

Conan-Kudo:
This now needs to be rebased and then we can look at landing this!

dcermak:
I've rebased the PR

Conan-Kudo (approved):

PR #2443, opened 2024-01-31
Fix reading of os-release file

schaefi:
If the /etc/os-release file contains comments or spaces python's csv reader will throw an exception. Thus this data must be ripped out prior reading


schaefi:
This bug currently prevents the following integration test to succeed
* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-systemd-boot

PR #2446, opened 2024-01-31
Fix crypto LUKS integration test

schaefi:
The integration test build also encrypts /boot which requires grub to open the LUKS pool using cryptomount. grub does not support the argonID salted password hashes. Thus the integration test description configures pbkdf2 instead


schaefi:
I need to move on with this and the respective test builds and works, guess it's ok to merge under this conditions

PR #2448, opened 2024-02-02
Drop PDF build of the documentation

schaefi:
The kiwi.pdf file as provided in the kiwi-man-pages package was never really used by our users but comes with a huge dependency chain to LaTex. We provide the documentation online as well as converted to DocBook from where all other formats could be derived. Thus there is no need for us to produce an extra PDF document which is imho never consulted. Along with this change there is also a new meta package called kiwi-devel which should make it easier for people to pull in packages for kiwi development. This was formerly done by the helper/install_devel_packages script which was pretty much outdated. I expect there is more attention on the package than on a custom script in some helper directory. Having a kiwi-devel also simplifies the contribution chapter which was not really user friendly regarding the needed development dependencies


schaefi:
@Conan-Kudo this should also make your poetry work a little easier as it drops some parts from the Makefile and fixes the MANIFEST.in which had a broken recursive_include of stuff we never needed

package/python-kiwi-spec-template
```diff
@@ -75,6 +75,23 @@ The KIWI Image System provides an operating system image builder
...
+Requires:       trang
+Requires:       libxml2-devel
+Requires:       libxslt-devel
+Requires:       python%{python3_pkgversion}-tox
+Requires:       python%{python3_pkgversion}-%{develsuffix} >= 3.6
+Requires:       python%{python3_pkgversion}-setuptools
+
+%description
+This metapackage installs the necessary dependencies
+to develop KIWI.
```
Conan-Kudo, line 93: This will confuse people in the spec file. You could just as easily get everything you need to develop kiwi by using `dnf builddep kiwi` or similar.

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: Sorry I can't follow you, what is confusing people ? 

iirc builddep or source-install for zypper installs everything that is needed to build the package, but this is not the same as developing kiwi. For example; trang is used to convert the schema if you develop a schema change. For building the package this is not needed at all. Thus I think a -devel package makes sense to track all that and pull it in

Conan-Kudo: `-devel` packages are typically used for building things _on_ the thing, not for developing the thing itself.

schaefi: ok, fine for me and I'm open to any other name that solves the confusion. But the motivation for such a package rather than the former ```helper/install_devel_packages```, you would agree to be a good move, no ?

I'm convinced providing a package to pull the "for kiwi development" packages will be better maintained that this script which was old, outdated and non functional.

So what would be a good name then ?

schaefi: ```kiwi-develdeps``` in alignment to our kiwi-systemdeps packages ?

schaefi: I renamed to kiwi-develdeps in alignment with kiwi-systemdeps, thoughts ?

package/python-kiwi-spec-template
```diff
@@ -75,6 +75,23 @@ The KIWI Image System provides an operating system image builder
...
+Requires:       trang
+Requires:       libxml2-devel
+Requires:       libxslt-devel
+Requires:       python%{python3_pkgversion}-tox
+Requires:       python%{python3_pkgversion}-%{develsuffix} >= 3.6
+Requires:       python%{python3_pkgversion}-setuptools
+
+%description -n kiwi-develdeps
+This metapackage installs the necessary dependencies
+to develop KIWI.
```
Conan-Kudo, line 93: I don't think this will be useful/necessary once I've gotten #2441 in place.

schaefi: ok so you are saying poetry also cares for the non python based requirements like "trang" to be present ? How does it do that ?

Conan-Kudo: Hmm, well, I guess not trang. But isn't trang already a weak dependency we use at runtime?

schaefi: no for development this is a strict dependency. See the Makefile
```
kiwi/schema/kiwi.rng: kiwi/schema/kiwi.rnc
    # whenever the schema is changed this target will convert
    # the short form of the RelaxNG schema to the format used
    # in code and auto generates the python data structures
    trang -I rnc -O rng kiwi/schema/kiwi.rnc kiwi/schema/kiwi.rng
```
It's required to convert the schema to RelaxNG which is the format we use. So if you as a developer do a single schema change you require this tool to convert it into the right format.

I guess not many people know that though ;)

schaefi: If you don't work with the schema you don't need it, but it would be good if we can come up with a concept that provides the full environment not only parts of it

schaefi: That's why I thought it would be handy that we provide this as a package in the same way you did it with the systemdeps packages and reference to it in the documentation

schaefi: gcc is needed for the C tools/ that we still have in the project to support users with custom initrds. But I would say all that C code should be dropped.

In the end outside of python I only see

* trang

The -devel packages are only in there when "pip install ..." starts to compile code from the module it fetches. I never completely understood why it sometimes compiles stuff and sometimes not, maybe poetry fixes that ? but it imho depends on how the module is provided on pypi

schaefi: @Conan-Kudo I'm down to trang as the only package outside of python. I was thinking we could add trang to the systemdeps-core even if that would not be 100% accurate. But if you would agree I can add the one package there and drop the -develdeps. Thoughts ?

Conan-Kudo: If we have only one dependency, we could express the check in the Makefile, no? Because trang is packaged nowhere but in openSUSE.

schaefi: ok that can also be done

schaefi: @Conan-Kudo ok done, this should be acceptable now, please check. Thanks

Conan-Kudo (approved):

PR #2450, opened 2024-02-04
Move Raid Luks and Integrity to context manager

schaefi:
Change the RaidDevice, LuksDevice and IntegrityDevice classes to context manager. All code using this classes moved to:
```python
with RaidDevice(...) as raid:
    raid.some_member()

with LuksDevice(...) as luks:
    luks.some_member()

with IntegrityDevice(...) as integrity:
    integrity.some_member()
```

In addition to the context manager move, the use of them has been refactored into a nested structure of supported device layouts inside the disk builder.

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -349,26 +349,33 @@ class DiskBuilder:
                     disk.public_partition_id_map['kiwi_RootPart'] = \
                         disk.public_partition_id_map['kiwi_ROPart']
 
-                # create raid on current root device if requested
-                if self.mdraid:
-                    raid_root = self._raid_instance(device_map)
-                    device_map = self._map_raid(device_map, disk, raid_root)
+                with ExitStack() as stack:
+                    if self.mdraid:
+                        # create raid on current root device
+                        raid_root = self._raid_instance(device_map)
+                        stack.push(raid_root)
+                        device_map = self._map_raid(
+                            device_map, disk, raid_root
+                        )
 
-                # create integrity on current root device if requested
-                if self.integrity:
-                    integrity_root = self._integrity_instance(device_map)
-                    device_map = self._map_integrity(device_map, integrity_root)
+                    if self.integrity:
+                        # create integrity on current root device
+                        integrity_root = self._integrity_instance(device_map)
+                        stack.push(integrity_root)
+                        device_map = self._map_integrity(
+                            device_map, integrity_root
+                        )
 
-                # create luks on current root device if requested
-                if self.luks is not None:
-                    luks_root = self._luks_instance(device_map)
-                    device_map = self._map_luks(device_map, luks_root)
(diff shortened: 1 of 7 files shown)


kiwi/builder/disk.py
```diff
@@ -349,26 +349,66 @@ def create_disk(self) -> Result:
...
+                            self._process_build(device_map, disk)
+                elif self.luks is not None:
+                    with self._luks_instance(device_map) as luks_root:
+                        # create luks on root dev
+                        device_map = self._map_luks(device_map, luks_root)
+                        # build bootable luks disk
+                        self._process_build(device_map, disk)
+                else:
+                    # build bootable disk
+                    self._process_build(device_map, disk)
```
dcermak, line 411: This part looks rather convoluted and a bit too deeply nested for my taste. Also, the `self._process_build()` call at the end of every exit from the context manager looks like there should be a better way to do this.

Have you considered using a single exit stack, use `push()` to add all the Raid/Integrity/LUKS but thereby preserve the previous & much easier to follow code structure?

dcermak (changes requested):
I think the new code in disk.py could use a bit of simplification

kiwi/builder/disk.py
schaefi: I knew you would say that ;) I also thought while writing it that this is not optimal. However, from the conversation about context managers I had the impression that you wanted this type of nested structure, no ?

I'm fine with an ExitStack that is much better. Will adapt the PR

schaefi: done

dcermak: > However, from the conversation about context managers I had the impression that you wanted this type of nested structure, no ?

No, I actually don't prefer this deep nesting that context managers cause. Hence I am very happy that the exit stack exists, as it allows us to combine the advantages of context managers and not having a five level deep nesting.

kiwi/builder/disk.py
```diff
@@ -703,6 +710,9 @@ def _map_root_filesystem(
         self.storage_map['system'] = filesystem
 
     def _process_build(self, device_map: Dict, disk: Disk) -> None:
+        # create system layout for root system
+        self._create_system_instance(device_map)
```
dcermak, line 714: Why is this suddenly necessary?

dcermak (approved):
Looks much better, thanks!

Just added one question about a code addition

kiwi/builder/disk.py
schaefi: It was there before. The call moved into the function as it imho makes more sense to have it there

schaefi: see the red lines from the diff, you will see the call as an extra line but it only makes sense in combination with the build of the disk

dcermak: Ah, sorry, I completely missed it :facepalm:

PR #2452, opened 2024-02-05
Remove destructor from BootLoaderInstallGrub2

schaefi:
With MountManager as context manager the BootLoaderInstallGrub2 class doesn't need a destructor anymore. This is related to Issue #2412

### DIFF
--- a/kiwi/bootloader/install/grub2.py
+++ b/kiwi/bootloader/install/grub2.py
@@ -19,6 +19,7 @@ import glob
 import os
 import re
 import logging
+from contextlib import ExitStack
 
 # project
 from kiwi.bootloader.install.base import BootLoaderInstallBase
@@ -65,15 +66,9 @@ class BootLoaderInstallGrub2(BootLoaderInstallBase):
         self.install_arguments = []
         self.shim_install_arguments = []
         self.firmware = None
-        self.efi_mount = None
         self.root_mount = None
-        self.boot_mount = None
-        self.device_mount = None
-        self.proc_mount = None
-        self.sysfs_mount = None
         self.volumes = None
         self.root_volume_name = None
-        self.volumes_mount = []
         self.target_removable = None
         if custom_args and 'target_removable' in custom_args:
             self.target_removable = custom_args['target_removable']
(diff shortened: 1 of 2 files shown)


kiwi/bootloader/install/grub2.py
```diff
@@ -174,114 +169,119 @@ def install(self):
...
+                            [
+                                'chroot', self.root_mount.mountpoint,
+                                'shim-install', '--removable'
+                            ] + self.shim_install_arguments + [
+                                self.device
+                            ]
+                        )
+                    finally:
+                        # restore the grub installer noop
+                        self._enable_grub2_install(self.root_mount.mountpoint)
```
dcermak, line 284: Does it really make sense to always perform the grub2 installer enablement? If an exception is thrown, then the build will be abandoned and the buildroot is wiped

schaefi: but I should move it into the try block

schaefi: yes this is correct but if the shim-install succeeds you have to revert the crap. What's wrong in trying to revert a really bad change to the root such that shim-install works in an image build ? I really dislike this disable/enable code and would like to get rid of it rather sooner than later but as long as it's needed I think we should try hard to revert any of the stupid changes to the root, no ?

schaefi: @dcermak I'd like to give this change and the other PR that already got merged a test with the integration tests. What's your take on this one, can we continue ? Thanks

dcermak: Sorry, forgot about this :see_no_evil: 

So tbh, I find it a bit odd to always perform this cleanup, even if a failure will be catastrophic and won't have any impact on the host. But it doesn't hurt either. If you think it's better to do it, then feel free to merge. It's not like it's bad, I just find it peculiar and it confuses me.

schaefi: ok, I'd like to keep it. The reason is I think temporary changes done to the system and only needed because of the weakness of a tool should be reverted in any case because I don't want to leave the root system in this state even if the tool call has failed and the build process exited. imho it's strange and confusing that we need this extra code to let shim-install work in an appliance build environment. It is yet another one of these tools that thinks it is only called on target and nowhere else... a wrong assumption which also applies to grub and makes all this crap code necessary. But as I said I don't want to leave the build process with that "crap changes" active in any case, mostly because I don't want people to work with it after the build has failed.

PR #2455, opened 2024-02-11
Drop obsolete C tools from kiwi

schaefi:
For building custom initrd images kiwi provided some optional helper tools. All these C written tools are old and outdated and either not useful anymore or
replaced by other tools that are part of the distribution matrix we support. Thus with the move to the next
major release those can go away. This also include the package kiwi-tools which was used to provide them


schaefi:
@Conan-Kudo this should also make your work with regards to #2441 easier

Conan-Kudo:
These aren't used by the kiwi dracut modules right?

PR #2457, opened 2024-02-12
Drop destructor from CommandProcess

schaefi:
There is code that sends a SIGTERM to the process in case there is no error code information. I believe in this case sending SIGTERM will not kill the process (defunct) and I also don't see in what good condition we would be entering this state.


schaefi:
Running a Staging build after the rebase

schaefi:
All tests passing, @dcermak good to go ?

Issue #2447, opened 2024-02-01
Create live images with direct squashfs


Vogtinator:
Currently kiwi puts the LiveOS inside an ext4 filesystem image which is then put inside a squashfs and then inside ISO9660/UDF.
I don't know the rationale for the current design, but the only reasons I can think of do not apply (anymore):

- Overlay on squashfs should work just fine
- squashfs supports all necessary FS features (hardlinks, sparse files, xattrs). ACLs aren't supported but I don't see any in the current live images.

Avoiding the indirection results in a smaller image (Leap 15.6 went from 588M to 575M), and should also noticably improve performance. It should also make .iso builds quicker.

Currently it looks like there's no existing option that could be reused to switch it off, the FS is hardcoded to ext4 in defaults.py.

Vogtinator:
> The reason for the layout was the compatibility with the upstream dmsquash dracut module. With kiwi-10 (current main branch) we have an opportunity to break compatibility, so maybe a good idea to rethink the live layout

FTR, I started experimenting with that, will try a PR if it works.

schaefi:
Thanks for the update :+1: I was about to also look into it, but if you do it I will stop for the moment. As a hint we have the ```flags="overlay|dmsquash"``` attribute in kiwi. In the ```builder/live.py``` code this is available in the ```self.live_type``` variable. So in your experiments you can say

```
if self.live_type == 'dmsquash':
    # keep the layout compatible with dracut modules upstream
else:
    # introduce a new layout together with the kiwi-live provided dracut module
```

The kiwi-live dracut module exists for some time and we had planned to drop it because meanwhile all features of our own module are available in upstream dmsquash too. However I agree with you we could provide a smaller/better/faster layout as an alternative in the kiwi-live dracut module when upstream cannot afford breaking the compat layer. Adding another layout to the upstream dracut dmsquash is not so easy because the code already has quite some niche cases for former overlay with device-mapper and it was less intrusive to have a small and clean live module. So long story short I think we can have an alternative live module in combination with upstream and customers could decide what they prefer. I think this wouldn't be a bad move for the existing kiwi live dracut and builder code

Vogtinator:
PR created: https://github.com/OSInside/kiwi/pull/2460

It doesn't quite follow what you had in mind, but it's rather simple and small (if you hide whitespace changes).

Vogtinator:
[...]

PR #2460, opened 2024-02-13
Allow specifying the filesystem for live image, also direct squashfs

Vogtinator:
By setting <type image="iso" filesystem="FSTYPE" .../> it's now possible
to specify the filesystem used for live images. By using "squashfs", the
rootfs container is skipped entirely.

Fixes https://github.com/OSInside/kiwi/issues/2447

### DIFF
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -230,7 +230,9 @@ class LiveImageBuilder:
                 self.live_type
             )
         )
-        root_filesystem = Defaults.get_default_live_iso_root_filesystem()
+        root_filesystem = self.xml_state.build_type.get_filesystem()
+        root_filesystem = root_filesystem if root_filesystem else \
+            Defaults.get_default_live_iso_root_filesystem()
         filesystem_custom_parameters = {
             'mount_options': self.xml_state.get_fs_mount_option_list(),
             'create_options': self.xml_state.get_fs_create_option_list()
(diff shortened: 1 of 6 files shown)


kiwi/xml_parse.py
```diff
@@ -16,7 +16,7 @@
 #   kiwi/schema/kiwi_for_generateDS.xsd
 #
 # Command line:
-#   /home/ms/Project/kiwi/.tox/unit_py3_11/bin/generateDS.py -f --external-encoding="utf-8" --no-dates --no-warnings -o "kiwi/xml_parse.py" kiwi/schema/kiwi_for_generateDS.xsd
+#   /data/git/kiwi/env/bin/generateDS.py -f --external-encoding="utf-8" --no-dates --no-warnings -o "kiwi/xml_parse.py" kiwi/schema/kiwi_for_generateDS.xsd
 #
 # Current working directory (os.getcwd()):
 #   kiwi
```
schaefi, line 22: please drop this change from the pull request, as you only changed the schematron rule there are no real changes to the generated API except for the date and the python version you are using.

kiwi/builder/live.py
```diff
@@ -230,65 +230,91 @@ def create(self) -> Result:
...
+                    container_image.name
+                )
+                Path.create(self.media_dir.name + '/LiveOS')
+                os.chmod(container_image.name, 0o644)
+                shutil.copy(
+                    container_image.name,
+                    self.media_dir.name + '/LiveOS/squashfs.img'
+                )
+        else:
+            # Put the root filesystem into SquashFS directly
```
schaefi, line 297: I like the way you added the feature in code but I think it's confusing that the filesystem attribute changes the layout of the live ISO. There is the ```flags=overlay|dmsquash``` attribute which we used to control the layout of the live ISO and there was a reason why ```filesystem``` was forbidden because you actually have no choice for the filesystem, it will always be some squashfs+overlay layout. So if you don't mind please use the value for the flags to differentiate the layout and not filesystem. You have the current value available in the variable ```self.live_type```. 

Feel free to run your new layout with ```if self.live_type == 'overlay':```.

schaefi: That way you also don't have to change the schema and the only missing part is then to double check if the documentation still matches the code for flags="overlay"

kiwi/xml_parse.py
schaefi: overall I think we don't need a schema change, see my previous comments

schaefi (changes requested):
I like the implementation, thanks much :+1: I think we should change the way how this is configured in the image description.

kiwi/builder/live.py
schaefi: If you don't like this you can also add another flag value but please don't let us abuse the filesystem attribute :)

Vogtinator: `overlay` vs. `dmsquash` is independent of the actual file system, it's just how writing and persistence are implemented. With `overlay`, the overlay is on top of the filesystem, while with `dmsquash` the overlay is on the block level below the filesystem. So it's perfectly fine to use `flags="overlay" filesystem="xfs"` or `flags="dmsquash" filesystem="btrfs"`.

IMO using `filesystem` for this is intuitive, because that is what the running live system will see as effective file system for `/`.

Even the `flags="dmsquash" filesystem="squashfs"` combination should work according to the [dracut.cmdline man page](https://manpages.opensuse.org/Tumbleweed/dracut/dracut.cmdline.7.en.html#Booting_live_images):

```
 For OverlayFS mount overlays, the filesystem structure may also be a direct compression of the root filesystem:

     squashfs.img          |  SquashFS from LiveCD .iso
        !(mount)
        /bin               |  Live filesystem
        /boot              |
        /dev               |
        ...                |
```

I didn't test that yet, but I'll try to do a build.

kiwi/xml_parse.py
Vogtinator: AFAICT we need a schema update no matter how it's implemented, as there needs to be some way to configure that.

kiwi/builder/live.py
schaefi: > IMO using filesystem for this is intuitive, because that is what the running live system will see as effective file system 
for /

This is true but it will always be a squashfs. If you say ```flags="dmsquash" filesystem="ext3"``` it will be accepted and ignored. Why not just add a new flag e.g

```xml
<type image="iso" ... flags="direct-squashfs"/>
```
or any name that makes it clear.

schaefi: > the overlay is on the block level below the filesystem. So it's perfectly fine to use flags="overlay" filesystem="xfs"

This is also true but I don't think we want to support that now or in the future. The squashfs embedded filesystem image was a layout only introduced to stay compatible with the old concept which uses only the device-mapper. I don't see any benefit to place an XX filesystem image inside of the squashfs. That's also the reason why I really like your change, it gets rid of this unneeded indirection. So why add code that allows for a setting we don't want to support ?

schaefi: ok, I was wrong when I wrote the filesystem setting got ignored, sorry I did no read carefully enough. But the question stays, why do we want to support this ? Does it make sense ?

kiwi/xml_parse.py
Vogtinator: Would be nice if `generateDS.py` had an option to omit the header, or at least the system specific content. That way the CI could regenerate both kiwi.rng as well as `xml_parse.py` and check that they stay unchanged, to make sure they match the source files.

kiwi/builder/live.py
schaefi: ok, I agree so let's keep it that way. Can you add a few words in the documentation about the change ? I think this chapter here would be a good place: https://osinside.github.io/kiwi/building_images/build_live_iso.html
Thanks

schaefi:
@Vogtinator ignore the CI-Compliant-To-Kiwi-9 github action. I realized that test does not work when the pull request comes from a fork

schaefi (approved):
Looks good to me :+1: 

Some words in the documentation and we can merge.

schaefi:
Thanks, just perfect :+1:

PR #2464, opened 2024-02-14
Add support for reading .profile.extra in initrd

schaefi:
If there is the file /.profile.extra available in the initrd, kiwi will import this additional environment file after the import of the standard /.profile file. This is related to bsc#1218095


schaefi:
@Conan-Kudo this is an enhancement requested from SUSE colleagues. I tested it with an internal example. I think it's a clean change. Thanks

Conan-Kudo (approved):
Meh, I don't know what you need this for, but sure.

Conan-Kudo:
FWIW, you could also just as easily just extend this with config.sh or ship your own overlay systemd units...

PR #2466, opened 2024-02-14
Don't use LUKS keyfile with disk password

schaefi:
In case a disk password is configured, no keyfile will be used. The setup of the disk password is a method to store an insecure password into the bootloader configuration and directly call cryptomount with that password to boot into the system once. The
insecure credentials will then be replaced by a secure version e.g from a TPM and a final keyfile will be placed into the luks pool. Thus there is no need for kiwi to create an initial keyfile as it's insecure and overwritten anyway.

This is related to bsc#1218181

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -644,6 +644,21 @@ class DiskBuilder:
         #
         luks_need_keyfile = \
             True if self.boot_is_crypto or self.luks == '' else False
+        if self.use_disk_password and self.luks:
+            # In case a disk password is configured, no keyfile
+            # will be used. The setup of the disk password is a
+            # method to store an insecure password into the bootloader
+            # configuration and directly call cryptomount with
+            # that password to boot into the system once. The
+            # insecure credentials will then be replaced by a
+            # secure version e.g from a TPM. This concept for
+            # non-interactive boot of an encrypted system is similar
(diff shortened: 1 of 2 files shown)


schaefi:
This needs confirmation from the people involved in the mentioned suse bug ticket

Vogtinator:
GRUB2 with recent openSUSE patches injects a keyfile into the initrd to avoid the second passphrase entry, which is basically the same during runtime as what kiwi does here during the image build.

Removing this kiwi behaviour would cause two regressions:

- A 2nd passphrase entry is necessary if GRUB2 without openSUSE patches is used
- kiwi-repart might also prompt for the passphrase without the keyfile supplied here ([bsc#1219009](https://bugzilla.suse.com/show_bug.cgi?id=1219009))

schaefi:
yeah you are right, I did not think about the fact that this only works with the grub patches specific to SUSE. So I think this means we should not do the workaround change in kiwi and add a comment to the bug in #1218181, that a real fix is needed. Thanks

Conan-Kudo:
This also means we should try to get these patches into at least the Red Hat grub2 tree, which is a shared upstream for all major distributions...

schaefi:
ok, so we are not doing the proposed change. There are more discussions regarding the topic here: https://bugzilla.suse.com/show_bug.cgi?id=1219009

A different solution is needed

PR #2467, opened 2024-02-15
Refactor Command.run()

dcermak:
`Command.run()` currently has a bit of a confusing behavior: if `raise_on_error` is `False` and the executable is not found, then a weird `CommandT` is returned (return code is -1 and stdout+stderr is None). This makes it possible to handle command not found errors separately, but it makes that needlessly verbose. So instead, let's just return `None` in **this** special case.

That in turn uncovered, that in most cases when we set `raise_on_error=True`, we actually want an error if the command is not present but no error if the command fails to execute (e.g. because it returns -1 if you run `$cmd --version` 🙄). Hence we introduce the flag `raise_on_command_not_found`, which causes an exception to be raised if the command is not found. This makes it independent of the `raise_on_error` flag.

Additionally, we add a small optimization: if command starts with /, then we assume it's a full path and we omit the call to which (and just check whether it exists).

These changes require additional tests and a few assertions in places where mypy cannot infer that a certain value is of the correct type.
Also, add type hints in a few (mostly random) places.

### DIFF
--- a/kiwi/builder/install.py
+++ b/kiwi/builder/install.py
@@ -280,11 +280,9 @@ class InstallImageBuilder:
             source_filename=self.diskname,
             keep_source_on_compress=True
         )
-        compress.xz(self.xz_options)
-        # set by compress.xz
-        assert compress.compressed_filename
+        xz_archive = compress.xz(self.xz_options)
         Command.run(
-            ['mv', compress.compressed_filename, pxe_image_filename]
+            ['mv', xz_archive, pxe_image_filename]
         )
 
         # the system image transfer is checked against a checksum
--- a/kiwi/builder/install.py
+++ b/kiwi/builder/install.py
@@ -281,6 +281,8 @@ class InstallImageBuilder:
             keep_source_on_compress=True
         )
         compress.xz(self.xz_options)
+        # set by compress.xz
+        assert compress.compressed_filename
         Command.run(
             ['mv', compress.compressed_filename, pxe_image_filename]
         )
--- a/.github/workflows/ci-kiwi-9-compliant.yml
+++ b/.github/workflows/ci-kiwi-9-compliant.yml
@@ -12,4 +12,4 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Run rebase
-        run: git fetch --all; git config user.email cherrypick@action.com; git config user.name git_action; git checkout master; git log --pretty=oneline --no-merges "origin/master..origin/${GITHUB_HEAD_REF}" | cut -f1 -d " " | head -n -1 > commit.list; git cherry-pick $(tac commit.list | tr "\n" " ")
+        run: git fetch --all; git config user.email cherrypick@action.com; git config user.name git_action; git checkout master; git log --pretty=oneline --no-merges "origin/master..origin/${GITHUB_HEAD_REF}" | cut -f1 -d " " | head -n -1 > commit.list; git cherry-pick --keep-redundant-commits $(tac commit.list | tr "\n" " ")
--- a/.github/workflows/ci-units-types.yml
+++ b/.github/workflows/ci-units-types.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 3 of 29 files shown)


schaefi:
> Umm wow. @schaefi can we have this staged to run image build tests before I approve and merge this?

yes I will stage this. It's a big change

kiwi/builder/install.py
```diff
@@ -281,6 +281,8 @@ def create_install_pxe_archive(self) -> None:
             keep_source_on_compress=True
         )
         compress.xz(self.xz_options)
+        # set by compress.xz
+        assert compress.compressed_filename
```
schaefi, line 285: Is this assert useful ?

* if compress.xz() fails it will raise an Exception from the Command.run
* I don't see how the AssertionError could be reached
* if it is reached all we will get is "AssertionError" without any further information

compress.xz() also returns ```compressed_filename``` I think it might be better to refactor the code in another pull request that to use the return value from ```compress.xz()``` rather than accessing the member of the compress object after the call.

In general I'm not a friend of assertions, they are usually placed at code that should never be reached but "who knows" :) if possible I think such code elements should be avoided. Thoughts ?

schaefi (changes requested):
Thanks, this is a nice change :+1: 

I will run the staging builds with it to check if anything breaks. Other than that I only have one question on the use of assert

kiwi/builder/install.py
dcermak: You are right, I only added this to work around the type checker, as it would otherwise not accept my code. I have pushed another commit that removes the usage of `Compress.compressed_filename` everywhere and hence removes this problem.

schaefi:
I submitted to Staging, let's see what happens :)

schaefi:
@dcermak The changes works except for distros with py <3.8, there it ends with
```
[   62s] ImportError: cannot import name 'Literal'
```

This is easy to fix and as long as it's easy to stay compatible we should do this. Thus can you please add

```patch
-from typing import TYPE_CHECKING, Dict, Literal, Union, overload
+import sys
+from typing import TYPE_CHECKING, Dict, Union, overload
+
+if sys.version_info >= (3, 8):
+    from typing import Literal  # pragma: no cover
+else:  # pragma: no cover
+    from typing_extensions import Literal  # pragma: no cover
 ```

This will fix it.

Thanks

dcermak:
> @dcermak The changes works except for distros with py <3.8, there it ends with
> 
> ```
> [   62s] ImportError: cannot import name 'Literal'
> ```
> 
> This is easy to fix and as long as it's easy to stay compatible we should do this. Thus can you please add
> 
> ```diff
> -from typing import TYPE_CHECKING, Dict, Literal, Union, overload
> +import sys
> +from typing import TYPE_CHECKING, Dict, Union, overload
> +
> +if sys.version_info >= (3, 8):
> +    from typing import Literal  # pragma: no cover
> +else:  # pragma: no cover
> +    from typing_extensions import Literal  # pragma: no cover
> ```
> 
> This will fix it.
> 
> Thanks

Done. I have also added a test run for both python 3.8 & 3.7, as we'll otherwise get more of these failures in the future.

schaefi:
@dcermak please drop 3.7 from the matrix. Due to the fb69627ad3b4a3dfd66b35b64252ed52f81fd40b 3.7 is out for the tests

Conan-Kudo:
IIRC, we're also dropping Python 3.8 since our new floor will be Python 3.9?

schaefi:
>IIRC, we're also dropping Python 3.8 since our new floor will be Python 3.9?

we said so but there are a few code parts that differentiate on the import for the typing vs. typing_extensions and the boundary here is py3.8. So for the unit test matrix I agree with Dan that we should keep 3.8 in when we have code for it. I don't want to delete the code that differentiates on the typing vs typing_extensions because that also keeps it functional for <3.8 even if that is something we don't care so much. It's very simple to keep that and there is no point in removing imho

schaefi:
Creating a new Staging build...

dcermak:
How about we just run mypy for python 3.7? I don't like the idea of having code that we don't test in production, but that will give us at least errors on these kind of import issues.

schaefi:
> How about we just run mypy for python 3.7? I don't like the idea of having code that we don't test in production, but that will give us at least errors on these kind of import issues.

sounds good to me :+1:

schaefi (approved):

PR #2468, opened 2024-02-16
Include partprobe in initrd for s390

davidcassany:
This commit includes partprobe, in addition to parted, on s390 based systems. Otherwise partx is used and apparently it does not properly support s390.

Fixes bsc#1219798


schaefi (approved):
looks good to me

Vogtinator:
IMO should gain a comment to explain why it's necessary, otherwise some cleanup like #2401 might remove it again...

PR #2470, opened 2024-02-17
use the 3.8 unittest.mock everywhere

a-detiste:


schaefi:
Hi, thanks for your effort. @a-detiste could you share a bit of an explanation what's the motivation for the change ?
As you can see from the workflow runs it causes some issues in the tests. Can you double check on the faulty tests too. Thanks much

a-detiste:
Hi, `mock` was an independent module that has been merged into Python standard library for 3.8 under the nam `unittest.mock`; the independent modules is not anymore developed and will continue to slowly bitrot.

The main question is not wether or not you want to switch to unittest.mock but _WHEN_ ; because it means dropping support for Python <= 3.7.

a-detiste:
I saw that `unittest.mock` was already used once in the codebase before this PR/ most likely by someone that is used to newer codebases. The CI tests passed for this one.

a-detiste:
Many other projects have been sprinkling this hybridized import code all over the place, but I think that by now it's not worth it anymore:

```
try:
    from unittest.mock import patch
except:
    from mock import patch
```

schaefi:
ok great, got it and I agree

schaefi:
@a-detiste can you please add the following patch on top of your changes, it fixed the broken unit test for me
```patch
diff --git a/test/unit/solver/sat_test.py b/test/unit/solver/sat_test.py
index f063c306..7f156b87 100644
--- a/test/unit/solver/sat_test.py
+++ b/test/unit/solver/sat_test.py
@@ -1,4 +1,5 @@
 import logging
+import importlib
 from unittest.mock import (
     patch, Mock, MagicMock
 )
@@ -21,7 +22,7 @@ class TestSat:
     def inject_fixtures(self, caplog):
         self._caplog = caplog
 
-    @patch('importlib.import_module')
+    @patch.object(importlib, 'import_module')
     def setup(self, mock_import_module):
         self.sat = Sat()
         self.solver = MagicMock()
@@ -44,11 +45,11 @@ class TestSat:
         self.sat.pool.setarch.assert_called_once_with()
         self.sat.pool.setarch.reset_mock()
 
-    @patch('importlib.import_module')
+    @patch.object(importlib, 'import_module')
     def setup_method(self, cls, mock_import_module):
         self.setup()
 
-    @patch('importlib.import_module')
+    @patch.object(importlib, 'import_module')
     def test_setup_no_sat_plugin(self, mock_import_module):
         mock_import_module.side_effect = Exception
         with raises(KiwiSatSolverPluginError):
```

Then we are ready for merge :)

Thanks

schaefi:
btw you can run the unit tests yourself via

```
tox -e unit_py3_11 -- "-n 5"
```

a-detiste:
In fact this all happened even years earlier than I thought:

https://github.com/testing-cabal/mock 'mock is now part of the Python standard library, available as [unittest.mock](https://docs.python.org/dev/library/unittest.mock.html) in Python 3.3 onwards.'

https://news.ycombinator.com/item?id=3913182

[The Python Standard Library - Where Modules Go To Die](http://www.leancrew.com/all-this/)


https://qa.debian.org/popcon.php?package=python-mock

schaefi:
> In fact this all happened even years earlier than I thought:

yeah wow, I completely missed that, probably because it did not cause an issue so far, but good to address and fix it

PR #2475, opened 2024-02-20
Fixed implementation for oem-shutdown

### LINKED ISSUE #2474: oem-shutdown doesn't shutdown but reboot the machine
## Problem description
`<oem-shutdown>true</oem-shutdown>` doesn't really **shutdown** but **reboot** the machine.

## Expected behaviour
The machine is to be shutdown after installation.

1. Create a hybrid iso image. With "oem-shutdown" set to true.
```xml
        <type image="oem" filesystem="ext4" installiso="true" installboot="install" firmware="uefi" kernelcmdline="console=ttyS0">
            <bootloader name="grub2" console="serial" timeout="10"/>
            <oemconfig>
                <oem-systemsize>2048</oem-systemsize>
                <oem-swap>true</oem-swap>
                <oem-swapsize>200</oem-swapsize>
                <oem-shutdown>true</oem-shutdown>
            </oemconfig>
        </type>
```
2. kiwi build
3. Start a virtual machine with qemu
```bash
sudo /usr/libexec/qemu-kvm -m 20480 -drive file=./xxx.qcow2,format=qcow2  -cdrom ./build/xxx.x86_64-1.0.0.install.iso -net nic -net user --enable-kvm -cpu host -boot d -serial mon:stdio
```
5. Waiting for it to finish installation
6. It reboots

* KIWI version: 9.25.21
* Operating system host version: Rocky Linux 9.2
* Operating system target version: CentOS 8.5.2111

I suspect the issue is resulted from this line, where `reboot -p` might mean reboot rather than poweroff.
https://github.com/OSInside/kiwi/blob/5e40229b1c75d0a36df17c2565bd23304168ad81/dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh#L33

schaefi:
If specified oem-shutdown caused a reboot -f -p which is a powerdown but not a graceful shutdown. This commit fixes this by using systemctl halt for a clean shutdown. This Fixes #2474

### DIFF
--- a/dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
+++ b/dracut/modules.d/99kiwi-dump-reboot/kiwi-dump-reboot-system.sh
@@ -30,7 +30,7 @@ function boot_installed_system {
         ask_and_shutdown "${ask_shutdown_text}"
     fi
     if [ "${kiwi_oemshutdown}" = "true" ];then
-        reboot -f -p
+        systemctl halt
     fi
 
     # if rd.kiwi.install.pass.bootparam is given, pass on most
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
This means we need systemd as a dependency for the module, right?

PR #2477, opened 2024-02-22
Move Repository to context manager

schaefi:
Change the Repository Factory to be a context manager. All code using Repository was updated to the following with statement:

```python
with Repository(...).new as repo:
    repo.some_member()
```

### DIFF
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -17,6 +17,7 @@
 #
 import os
 import logging
+from contextlib import ExitStack
 from typing import (
     List, Any, Optional
 )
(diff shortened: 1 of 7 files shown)


kiwi/system/prepare.py
```diff
@@ -404,19 +407,20 @@ def pinch_system(
...
+                    if manager is None:
+                        package_manager = self.xml_state.get_package_manager()
+                        release_version = self.xml_state.get_release_version()
+                        repo = Repository.new(self.root_bind, package_manager)
+                        stack.push(repo)
+                        manager = PackageManager.new(
+                            repository=repo,
+                            package_manager_name=package_manager,
+                            release_version=release_version
+                        )
```
dcermak, line 420: How is this branch reachable? The type hint says that `manager` is an instance of `PackageManagerBase` not `Optional[PackageManagerBase]`

dcermak (approved):
LGTM, just one question about a related issue

kiwi/system/prepare.py
schaefi: The type hint is wrong. pinch_system is allowed to be called without a pre-created PackageManager instance. I will fix it

Conan-Kudo (approved):

PR #2479, opened 2024-02-24
Add support for discoverable partitions

### LINKED ISSUE #1385: add support for discoverable partitions
Systemd has put up a number of standard UUIDs for GPT partitions:
https://systemd.io/DISCOVERABLE_PARTITIONS/

Quoting the link:

*   OS installers can automatically discover and make sense of partitions of existing Linux installations.
*    The OS can discover and mount the necessary file systems with a non-existing or incomplete /etc/fstab file and without the root= kernel command line option.
*   Container managers (such as nspawn and libvirt-lxc) can decode and set up file systems contained in GPT disk images automatically and mount them to the right places, thus allowing booting the same, identical images on bare-metal and in Linux containers. This enables true, natural portability of disk images between physical machines and Linux containers.
 *  As a help to administrators and users partition manager tools can show more descriptive information about partitions tables.

Would make sense to support that from kiwi I guess.

schaefi:
Set PARTUUID according to systemd-id128 if applicable

This Fixes #1385


Conan-Kudo:
Can we get test Fedora and openSUSE images for this?

schaefi:
> Can we get test Fedora and openSUSE images for this?

yes I will put it to Staging

schaefi:
The following distros does not have ```systemd-id128```

* Testing_x86:archlinux - http://ftp.hosteurope.de/mirror/ftp.archlinux.org/core/os/x86_64
* Testing_x86:debian - obs://Debian:11/standard
* Testing_x86:ubuntu - obs://Ubuntu:22.04/universe
* Testing_x86:centos - CentOS-8

I currently clarify with Daan if this is expected or if systemd provided the information differently.

mwilck:
`systemd-id128` just prints the list of UUIDs that is maintained at the UAPI group. I don't think it matters if a distro ship the tool.

schaefi:
> `systemd-id128` just prints the list of UUIDs that is maintained at the UAPI group. I don't think it matters if a distro ship the tool.

ok so you are proposing that we maintain a static list of IDs as defined by the UAPI group and do not call a tool to retrieve them ?

mwilck:
> ok so you are proposing that we maintain a static list of IDs as defined by the UAPI group and do not call a tool to retrieve them ?

We could use the tool if available, and fall back to a static list otherwise. To my understanding these UUIDs are considered "stable for eternity".

schaefi:
ok I'm adapting the PR to have a fallback

schaefi:
All Staging builds pass. @Conan-Kudo if you would like to double check the images, feel free to look at

* https://build.opensuse.org/package/live_build_log/Virtualization:Appliances:Images:Testing_x86:fedora/test-image-live-disk:Disk/images/x86_64
* https://build.opensuse.org/package/live_build_log/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-disk/images/x86_64

search for ```partition-guid``` in the log or check the result via loop mount and blkid

Conan-Kudo (approved):

PR #2487, opened 2024-02-28
Set default output console to gfxterm for grub

schaefi:
If no console setting is done in the image description for grub the default output console is set to: gfxterm and the default input console is set to: console. This Fixes bsc#1219074

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -122,9 +122,9 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         ]
 
         self.terminal_output = \
-            terminal_output if terminal_output in terminal_output_grub else ''
+            terminal_output if terminal_output in terminal_output_grub else 'gfxterm'
         self.terminal_input = \
-            terminal_input if terminal_input in terminal_input_grub else ''
+            terminal_input if terminal_input in terminal_input_grub else 'console'
 
         self.gfxmode = self.get_gfxmode('grub2')
         self.theme = self.get_boot_theme()
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -894,7 +894,7 @@ class TestBootLoaderConfigGrub2:
         self.bootloader.multiboot = True
         self.bootloader.setup_live_image_config(self.mbrid)
         self.grub2.get_multiboot_iso_template.assert_called_once_with(
-            True, False, False, None
+            True, True, False, None
         )
 
     @patch.object(BootLoaderConfigGrub2, '_copy_grub_config_to_efi_path')
@@ -1043,7 +1043,7 @@ class TestBootLoaderConfigGrub2:
         self.bootloader.multiboot = False
         self.bootloader.setup_install_image_config(self.mbrid)
         self.grub2.get_install_template.assert_called_once_with(
-            True, False, False, True
+            True, True, False, True
         )
         mock_copy_grub_config_to_efi_path.assert_called_once_with(
             'root_dir', 'earlyboot.cfg', 'iso'


Conan-Kudo:
We should only default to `gfxterm` if a theme is set, otherwise we should default to `console`.

schaefi:
> We should only default to `gfxterm` if a theme is set, otherwise we should default to `console`, I think.

gfxterm can also render "text mode" in a nicer way. I don't think it should be coupled to a theme. Also the reason why this was brought up is that it's a regression to former behavior. Thus the defaults should stay

Conan-Kudo:
Using gfxterm has the consequence of breaking flicker-free boot, though maybe we don't care here?

schaefi:
> Using gfxterm has the consequence of breaking flicker-free boot, though maybe we don't care here?

Problem is that I don't have much of a choice. This was considered a regression and I need to do something. See here
https://bugzilla.suse.com/show_bug.cgi?id=1219074 

It seems to have a bad effect on QA and I won't get any package into TW or ALP without a fix.

Personally I think having console to be the default is better and if you want graphics well then add the attribute to the image description and be done. But it's not as easy as one think and I can't say much against the regression. We had gfxterm in from day one and it even was the default in grub-mkconfig for some time... until they changed it. So I could also argue, well grub has changed too so why not we ? But honestly I don't see the value in arguing on this regard because any group can make a point and we won't move forward

Conan-Kudo:
Wouldn't it make sense to just adjust the kiwi descriptions for openSUSE and SLE then?

schaefi:
> Wouldn't it make sense to just adjust the kiwi descriptions for openSUSE and SLE then? That's what I do for the Fedora and CentOS ones when we make upstream changes.

yes I agree and on the suse side it's considered a regression. As I said both sides has a fair point. From the suse side I got a clear message: "We can't have any newer kiwi in Tumbleweed without this fixed" (see the bug). Thus I doubt there is willingness to change all affected image descriptions.

From my perspective the less troublesome and most effective change is merging this PR

davidcassany (approved):
I agree with @schaefi that if this is perceived as a regression that clearly points KIWI we should fix it or either have strong arguments to not do so. IMHO there are no definitive arguments in favor of one approach or the other, hence requiring image definitions changed without a planning or advising it ahead feels like a bad approach. At least at openSUSE side this implies quite some work (there several images and in different code streams) and it can potentially be also the case for other organizations making use of KIWI.

Moreover we are currently planning v10, hence IMHO this is the chance to craft defaults as we want. So wouldn't it make sense having this change for v9.x only? v10 should have the defaults we want and moving from v9 to v10 including some description adaptations makes total sense to me and IMHO also puts as in a way stronger position to argue that's one of the reasons for a new major version, because little adaptations for some defaults require manual intervention.

Approving the PR as I do not disagree with change applied here, however, given the current discussion, I believe this PR makes more sense in a v9 scope only, so merging to _master_. Thoughts?

Conan-Kudo:
I am fine with this for 9.x and then making downstreams deal with this properly in 10.

schaefi:
Guys it's the TW people who complained about the regression. Doing this for v9 and introduce it back in v10 a few weeks later is not a solution just a delay to a tough conversation. This is a regression in v9 AND v10 because we never created any issue pointing out that we will be changing the bootloader default console. I really cannot create an argumentation based on a regression.

So it's time to make a next step and move forward. Here is what I will do now

1. This PR gets merged in main
2. This commit gets cherry-picked to master
3. I will submit master with the fix to TW such that we get unblocked and everybody is happy again and people stop sending me e-mails when it finally gets fixed
4. I will open an issue marked as kiwi10 to officially change the bootloader default console to be text based

Next we can create a PR to change the default, now everyone knows and the next time they complain I point them to the process of change which is now no longer based on a regression but on a community decision

Conan-Kudo:
> I will submit master with the fix to TW such that we get unblocked and everybody is happy again and people stop sending me e-mails when it finally gets fixed

Oof. ☹️

PR #2489, opened 2024-02-28
Update changes reference file

davidcassany:
Note first commit is an update of the changes reference file and second is a the actual change to fix it.


schaefi:
Unfortunately I found some issues with regards to the .changes update. The reference data is now newer than some of the changes done in main. The creation of the changes file will skip those which is something we don't want. Having two different changes reference files in master and main will produce conflicts.

I have an idea how to handle the problem but need to do some tests first

PR #2490, opened 2024-02-28
Add changelog fix file for commit deb6ca

schaefi:
The commit used a wrong e-mail address which should not land in the created changes file for the packaging

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,8 +102,10 @@ build: clean tox
 	# provide rpm source tarball
 	mv dist/kiwi-${version}.tar.gz dist/python-kiwi.tar.gz
 	# update rpm changelog using reference file
-	helper/update_changelog.py --since package/python-kiwi.changes > \
-		dist/python-kiwi.changes
+	helper/update_changelog.py \
+		--since package/python-kiwi.changes \
+		--fix package/python-kiwi.changes.deb6ca.fix \
+	> dist/python-kiwi.changes
 	helper/update_changelog.py --file package/python-kiwi.changes >> \
 		dist/python-kiwi.changes
 	# update package version in spec file
--- a/helper/update_changelog.py
+++ b/helper/update_changelog.py
@@ -2,12 +2,17 @@
 """
 usage: update_changelog (--since=<reference_file>|--file=<reference_file>)
             [--utc]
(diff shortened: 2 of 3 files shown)


schaefi:
@davidcassany I had some trouble with the update of the reference changelog and I think we should not touch it. Instead I propose a concept that allows an overwrite per commit. Thoughts ?

davidcassany:
@schaefi I am fine with not touching the reference file, I was also not happy with such an approach. I think your idea is good enough as long as it doesn't have to be used often. If additional changes like that in the future are needed it might turn to be unmanageable having a file per commit fix, but if that happens (I really hope it won't) then probably the whole changes file approach should be revisited.

A bit unrelated but something I have also realized with this issue is that we are building the change log and ordering it based on the author date, not the committer date. Which is fine for _main_, but could to be tricky for _master_ if commits are not cherry-picked with the former order, in _master_ cherry-picked commits not appear with the order of the cherry-pick action but with the other of the author date. I don't think we can do much about it other than paying attention to it, this could cause submissions that append interleaved change log entries, which shouldn't be that bad, but still not a clean pre-append.

davidcassany (approved):
Looks good :+1:

I also gave it a run and in also included additional commit fixes to exercise it myself and how it should be used. So far it's been easy to understand and exercise and the result was exactly the expected one.

schaefi:
> A bit unrelated but something I have also realized with this issue is that we are building the change log and ordering it based on the author date, not the committer date. Which is fine for main, but could to be tricky for master if commits are not cherry-picked with the former order, in master cherry-picked commits not appear with the order of the cherry-pick action but with the other of the author date. I don't think we can do much about it other than paying attention to it, this could cause submissions that append interleaved change log entries, which shouldn't be that bad, but still not a clean pre-append.

Yes you are absolutely right on this. Actually I hope that the move to poetry will finish soon. I'm finished with the "no finalizer" refactor and we already have dropped what we wanted to drop. Thus I would like to submit kiwi10 to ALP and TW(Factory) asap which then stops maintaining and cherry-picking to master. We had that interleaved changelog entries in the past and it fortunately was not an issue for suse's changelog management.

I also agree if there are more issues around the changes creation in the future we need to revisit the concept

PR #2496, opened 2024-02-29
Fix activation of luks pool in the initrd

schaefi:
kiwi called systemd-cryptsetup directly which does not take the settings available in /etc/crypttab into account. This commit changes the activation procedure in a way that we ensure all crypttab settings are taken into account by calling the systemd-cryptsetup-generator first, followed by the activation through the generated systemd-cryptsetup@... unit. This Fixes bsc#1219009


dracut/modules.d/99kiwi-lib/kiwi-luks-lib.sh
```diff
@@ -39,7 +39,10 @@ function activate_luks {
         luksOpen "${device}" luks
     else
         # There is a keyfile and we need to get prompted to enter the passphrase
-        /usr/lib/systemd/systemd-cryptsetup attach luks "${device}"
+        /usr/lib/systemd/system-generators/systemd-cryptsetup-generator \
```
Vogtinator, line 42: The unit should already exist, so this call shouldn't be necessary

schaefi: yes and you are absolutely sure that between the first generator call and the time we resize that there is no combustion, ignition, some-other-magic that kicks something in the background and changes ```/etc/crypttab``` ? If you say yes, I delete it :-)

schaefi: ok good point, I'll drop it

schaefi:
I tested the current change and it worked so far for me

Vogtinator (approved):

PR #2498, opened 2024-03-01
Building images revision

dmpop:
Revise all `.rst` files in the `doc/building_images` directory.


schaefi:
Thanks much. I read through the docs after rendering with sphinx and it is a much better read at all places that you have changed :+1:

schaefi (approved):

PR #2502, opened 2024-03-11
Lookup distro provided BLS entries for zipl

schaefi:
Check if there is already an entry file name produced by the  distribution logic/policy. If we can find an already present  entry name, prefer this name over the kiwi naming policy


schaefi:
@Conan-Kudo During my tests on s390 I understand that some distros doesn't have a concept for the loader entries of BLS zipl and some others have. On Fedora the policy is based on a machine-id naming convention as documented by IBM. kiwi's default naming policy is based on the OS ID which I think is the better concept. Nevertheless kiwi has to respect the distro over its own policies. Thus this change looks up an eventually created entries.conf and takes this name rather than producing its own version of it.

This is important because if there are multiple BLS entry configs at the time zipl is called this leads to a menu selection and it becomes unclear if the correct entry is set as the default

schaefi:
I will run the integration test builds to make sure the boot has only one entry for fedora and it uses the fedora naming policy. No merge until I have tested it please

schaefi:
> Sure! Just leave a comment once you've verified it!

verified and works

kiwi/bootloader/config/zipl.py
```diff
@@ -101,9 +101,15 @@ def set_loader_entry(self, root_dir: str, target: str) -> None:
                 f'Kernel lookup in {root_dir}/lib/modules failed with {issue}'
             )
         default_entry = f'{os_release.get("ID")}-{kernel_name}.conf'
+        # Check if there is already an entry file name produced by the
+        # distribution logic/policy. If we can find an already present
+        # entry name, prefer this name over the kiwi naming policy
+        for dist_entry in glob.iglob(f'{root_dir}/{self.entries_dir}/*'):
+            default_entry = os.path.basename(dist_entry)
+            break
```
Conan-Kudo, line 109: Shouldn't this be in bootloader_spec_base for all bls bootloaders to use?

Conan-Kudo (changes requested):

kiwi/bootloader/config/zipl.py
schaefi: good point, yes

schaefi: done

Conan-Kudo (approved):

PR #2504, opened 2024-03-12
Fixup use of /boot/zipl

schaefi:
Only for the grub2_s390x_emu bootloader setup an optional boot partition setup must be put to /boot/zipl rather than to /boot


schaefi:
Testing required, no merge please. I need to check on suse and fedora this time

Conan-Kudo (approved):

PR #2505, opened 2024-03-12
Drop rpm-dir from allowed repository type

### LINKED ISSUE #1926: Remove "rpm-dir" repository type support
KIWI has a generic "rpm-dir" type repository, which in practice only works with Zypper's special "rpm-dir" repository type. No other RPM package manager has this capability, and OBS does not use it, even with Zypper.

We should _immediately_ deprecate this in KIWI 9 for removal in KIWI 10.

schaefi:
Using an arbitrary list of rpm packages as repository is a zypper only feature, barely tested and from our pov not really needed as a simple createrepo call turns any custom list of packages into a clean rpm-md repo including metadata. This commit drops rpm-dir from the list of allowed repository types and auto converts those image descriptions which makes use of it. Please note this does not prevent users from using flat package directories with zypper, because the type argument in the repository section is an optional attribute. In case there is no type specification zypper auto-detects and handles the data as it handles it. This Fixes #1926

### DIFF
--- a/doc/source/commands/system_build.rst
+++ b/doc/source/commands/system_build.rst
@@ -134,7 +134,7 @@ OPTIONS
 
   - **type**
 
-    repository type, could be one of `rpm-md`, `rpm-dir` or `yast2`.
+    repository type, could be one of `rpm-md`, or `apt-deb`.
 
   - **alias**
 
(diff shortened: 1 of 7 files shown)


Conan-Kudo:
@schaefi You need to bump the schema version in the spec file.

schaefi:
yeah, thanks I forgot about that provides completely

PR #2508, opened 2024-03-18
Fixup cleanup of zipl templates

schaefi:
Make sure temporary modifications to the zipl template and config file are not effective in the later system. This Fixes bsc#1221469

### DIFF
--- a/kiwi/bootloader/install/grub2.py
+++ b/kiwi/bootloader/install/grub2.py
@@ -210,28 +210,6 @@ class BootLoaderInstallGrub2(BootLoaderInstallBase):
                         'grub2-zipl-setup', '--keep'
                     ]
                 )
-                zipl_config_file = ''.join(
-                    [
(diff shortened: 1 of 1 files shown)


Vogtinator:
LGTM, but I'm a bit surprised that it didn't impact any existing tests

schaefi:
> LGTM, but I'm a bit surprised that it didn't impact any existing tests

Thanks. The problem was that the restore of the template files only happened for CDL targets when it should happen always. The tests we had tested the cleanup for CDL and we didn't had a test to test the "no cleanup" situation. The change here causes  the cleanup code to happen always which still keeps the existing test functional. I think the issue was that we didn't had a test for the "no cleanup" condition. Maybe if we would have had that test we would have asked ourselves if that's not a mistake... sounds philosophical doesn't it :)

Conan-Kudo (approved):

PR #2516, opened 2024-03-25
Wait for udev event queue to become empty

schaefi:
Make sure to wait for the event queue to become empty prior accessing the device nodes

### DIFF
--- a/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
+++ b/dracut/modules.d/90kiwi-live/kiwi-live-lib.sh
@@ -66,6 +66,15 @@ function loadKernelModules {
     modprobe squashfs
 }
 
+function udev_pending {
+    declare DEVICE_TIMEOUT=${DEVICE_TIMEOUT}
+    local limit=120
+    if [[ "${DEVICE_TIMEOUT}" =~ ^[0-9]+$ ]]; then
+        limit="${DEVICE_TIMEOUT}"
+    fi
+    udevadm settle --timeout="${limit}"
+}
+
 function initGlobalDevices {
     if [ -z "$1" ]; then
         die "No root device for operation given"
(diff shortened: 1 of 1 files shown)


schaefi:
There seems to be a race condition in the live iso code for which this change could be a solution. However, it's not clear yet and needs proper testing in the case of the race

Vogtinator:
This should be a noop as the script is only called from the initqueue --settled hook

schaefi:
> This should be a noop as the script is only called from the initqueue --settled hook

Yes I agree it doesn't make sense... in theory everything should have already settled

schaefi:
I changed the code to wait for the event queue after an eventual partition creation. I think this could really cause the race condition. When we add the write partition and trigger the re-reading of the table this will queue new events for udev. Waiting for them to settle is imho important. I'll check the builds with this patch.

schaefi:
Can someone else double check on it ? Thanks much

aafeijoo-suse:
> I changed the code to wait for the event queue after an eventual partition creation.

If my feedback can be of some use, the first approach of this PR didn't work, but this last change worked for me (I tried it twice).

Vogtinator:
The bug https://bugzilla.opensuse.org/show_bug.cgi?id=1219074 just arrived in 15.x, breaking live cds even more :-/

Vogtinator:
I backported 69cc1c35d77ca56b930f1a0d79f8c19b9d43e754 as well as this PR to 15.6 and the resulting rescue-cd image booted successfully 3/3 times. So it's most likely fixed!

schaefi:
Yay, thanks much for testing :+1:  Also all my tests worked.

schaefi:
@Vogtinator I rebased and provided more details in the commit message. I'd like to add the bugzilla reference connected with my fix but can't remember which bug it was. Can you give me a hint ? Thanks

Vogtinator:
> @Vogtinator I rebased and provided more details in the commit message. I'd like to add the bugzilla reference connected with my fix but can't remember which bug it was. Can you give me a hint ? Thanks

https://bugzilla.opensuse.org/show_bug.cgi?id=1213595

Vogtinator (approved):

PR #2518, opened 2024-03-27
Fix setup of discoverable partitions

### LINKED ISSUE #2517: discoverable partitions support is broken
## Problem description
Commit 67bf130cd1da5372079810f87445dd70f26b3a24 has a bug, it sets the wrong uuid.

The gpt partition entries have /two/ guids. The partition uuid (which shows up in /dev/disk/by-partuuid), and the partition /type/ uuid. The discoverable partition uuids should go to the latter, the patch sets the former though.

```
fdisk -lx
[ ... ]
Device       Start      End Sectors Type-UUID                            UUID                                 Name     Attrs
/dev/sda1     2048     6143    4096 21686148-6449-6E6F-744E-656564454649 F77EF7E3-7096-4DC1-BB0D-F234380ACA27 p.legacy 
/dev/sda2     6144   210943  204800 C12A7328-F81F-11D2-BA4B-00A0C93EC93B C12A7328-F81F-11D2-BA4B-00A0C93EC93B p.UEFI   
/dev/sda3   210944  2258943 2048000 0FC63DAF-8483-4772-8E79-3D69D8477DE4 BC13C2FF-59E6-4262-A352-B275FD6F7172 p.lxboot 
/dev/sda4  2258944 10485726 8226783 0FC63DAF-8483-4772-8E79-3D69D8477DE4 4F68BCE3-E8CD-4DB1-96E7-FBCAF984B709 p.lxroot 
```
The `4F68...` discoverable partition uuid for the x86_64 root filesystem is in the `UUID` not the `Type-UUID` column

* KIWI version: 10.0.8
* Operating system host version:  Fedora 39
* Operating system target version: Fedora 40

schaefi:
Make sure GUI and typecode are set. This Fixes #2517

### DIFF
--- a/kiwi/partitioner/gpt.py
+++ b/kiwi/partitioner/gpt.py
@@ -116,7 +116,7 @@ class PartitionerGpt(PartitionerBase):
 
     def set_uuid(self, partition_id: int, uuid: str) -> None:
         """
-        Set partition UUID (GUID)
+        Set partition UUID (TypeCode)
 
         :param int partition_id: partition number
         :param string uuid: UUID
@@ -124,7 +124,7 @@ class PartitionerGpt(PartitionerBase):
         Command.run(
             [
                 'sgdisk',
-                '--partition-guid', f'{partition_id}:{uuid}',
+                '--typecode', f'{partition_id}:{uuid}',
                 self.disk_device
             ]
         )
--- a/test/unit/partitioner/gpt_test.py
+++ b/test/unit/partitioner/gpt_test.py
@@ -121,5 +121,5 @@ class TestPartitionerGpt:
     def test_set_uuid(self, mock_Command_run):
         self.partitioner.set_uuid(42, 'ID')
         mock_Command_run.assert_called_once_with(
-            ['sgdisk', '--partition-guid', '42:ID', '/dev/loop0']
+            ['sgdisk', '--typecode', '42:ID', '/dev/loop0']
         )


Conan-Kudo:
Do we want the partition UUID and the type UUID to be the same? I don't think so... I think the idea here is that we _only_ set the type UUID, right @kraxel?

schaefi:
I think it looks inconsistent if the typecode and the GUID are different. If they vary systemd-boot as an example would be able to identify the root device and boot from it, but the same device would have a different /dev/disk/by-partuuid device node name. So I did that for consistency and think it doesn't hurt

Conan-Kudo:
Well it creates the problem that Linux will get confused when trying to enumerate partitions when multiple formatted disks are present.

schaefi:
I will adapt ... tomorrow ;) late here

kraxel:
> Do we want the partition UUID and the type UUID to be the same? I don't think so... I think the idea here is that we _only_ set the type UUID, right @kraxel?

Yes.  The Type-UUID specified what kind of partition this is, and the UUID should be unique for each partition.  I think sfdisk already generates a UUID when creating a gpt entry , so there should be nothing to do.

schaefi:
ok, so the updated patch should address it correctly now

kiwi/partitioner/gpt.py
```diff
@@ -116,15 +116,15 @@ def set_flag(self, partition_id: int, flag_name: str) -> None:
 
     def set_uuid(self, partition_id: int, uuid: str) -> None:
```
Conan-Kudo, line 117: Let's change this function name. That way it's clear what it is.

Conan-Kudo (changes requested):

kiwi/partitioner/gpt.py
schaefi: Sorry but this would change the public interface and I'm not very keen on creating a deprecation decorator. Also this name is part of the partitioner interface and exists for msdos, gpt, dasd and has a different meaning according to the different table types anyway. It's a UUID, and what it technically means we can document but I don't think you are able to come up with any name that could not be misinterpreted. So let's stick with the simplest: It's a UUID

Conan-Kudo: Okay.

Conan-Kudo (approved):

PR #2520, opened 2024-04-03
Fix efifatimagesize attribute type

imcsk8:
The efifatimagesize attribute type value is set to "oem" but the documentation says that it is intended to be used for creating ISO images. This causes a schema error when this attribute is set on a profile with type "iso" and blocks changing the EFI boot image size which is a problem if the image is bigger than 20M.

This patch changes the efifatimagesize attribute type to "iso" with the benefit of being able to change the EFI boot image size and avoid problems when the size is bigger than 20M.

Fixes # No issue for this problem.

* Change the value for the type of the atrribute `efifatimagesize` to `iso`.

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -1588,7 +1588,7 @@ div {
         attribute efifatimagesize { xsd:nonNegativeInteger }
         >> sch:pattern [ id = "efifatimagesize" is-a = "image_type"
             sch:param [ name = "attr" value = "efifatimagesize" ]
-            sch:param [ name = "types" value = "oem" ]
+            sch:param [ name = "types" value = "iso" ]
         ]
     k.type.dosparttable_extended_layout.attribute =
         ## Specify to make use of logical partitions inside of an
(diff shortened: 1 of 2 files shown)


kiwi/schema/kiwi.rnc
```diff
@@ -1588,7 +1588,7 @@ div {
         attribute efifatimagesize { xsd:nonNegativeInteger }
         >> sch:pattern [ id = "efifatimagesize" is-a = "image_type"
             sch:param [ name = "attr" value = "efifatimagesize" ]
-            sch:param [ name = "types" value = "oem" ]
+            sch:param [ name = "types" value = "iso" ]
```
schaefi, line 1591: I think the way you set the schematron rule will overwrite it, so it should be set as

```
sch:param [ name = "types" value = "oem iso" ]
```

schaefi (changes requested):
Hi, thanks for looking into that. Per comment I think the schematron rule needs to be set differently.

kiwi/schema/kiwi.rnc
imcsk8: Thanks!
I'll add the suggestion to the PR

schaefi: Thanks, any news on this one ?

imcsk8: Took me a little time to properly test the change. I've uploaded another commit with your suggestion

schaefi (approved):
Looks good :+1:  Thanks

PR #2523, opened 2024-04-03
Add schema upgrade opportunity for old schemas

schaefi:
kiwi files using a schema version < 7.4 are no longer supported by kiwi >= v10.x.x. Thus this commit provides the required XSL stylesheets to upgrade older schemas to v74 such that they can be consumed by the latest kiwi version. The needed xsltproc instruction is placed on the main page of the documentation

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -724,11 +724,13 @@ fi
 
 %files -n python%{python3_pkgversion}-kiwi
 %dir %{_defaultdocdir}/python-kiwi
+%dir %{_usr}/share/kiwi
 %{_bindir}/kiwi
 %{_bindir}/kiwi-ng
 %{_bindir}/kiwi-ng-3*
 %{python3_sitelib}/kiwi*
 %{_usr}/share/bash-completion/completions/kiwi-ng
+%{_usr}/share/kiwi/xsl_to_v74/
 %{_defaultdocdir}/python-kiwi/LICENSE
 %{_defaultdocdir}/python-kiwi/README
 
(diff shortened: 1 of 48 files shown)


Conan-Kudo (approved):

package/python-kiwi-spec-template
```diff
@@ -724,11 +724,13 @@ fi
 
 %files -n python%{python3_pkgversion}-kiwi
 %dir %{_defaultdocdir}/python-kiwi
+%dir %{_usr}/share/kiwi
 %{_bindir}/kiwi
 %{_bindir}/kiwi-ng
 %{_bindir}/kiwi-ng-3*
 %{python3_sitelib}/kiwi*
 %{_usr}/share/bash-completion/completions/kiwi-ng
+%{_usr}/share/kiwi/xsl_to_v74
```
Conan-Kudo, line 733: Please put a trailing slash here.

Conan-Kudo (changes requested):

dcermak (commented):
All these `convertNtoM.xsl` files look autogenerated, wouldn't it make sense to create them here in the repo?

schaefi:
> All these `convertNtoM.xsl` files look autogenerated, wouldn't it make sense to create them here in the repo?

Oh no they are not auto generated. It was hard manual work to write these style sheets since kiwi exists :) Why do you think that is auto generated code ?

package/python-kiwi-spec-template
schaefi: I will, thanks

schaefi: @Conan-Kudo done

Conan-Kudo (approved):

PR #2526, opened 2024-04-08
Fixed runtime checks for dracut module packages

### LINKED ISSUE #2524: dracut-kiwi-oem-repart not found because its  kiwi-dracut-oem-repart
kiwi shows the following error:

```
Required dracut module package missing in package list

The package 'dracut-kiwi-oem-repart' is required for the selected
oem image type. Please add the following in your
<packages type="image"> section to your system XML
description:

```

The problem is, when building a Debian image `dracut-kiwi-oem-repart` does not exist becuase it is actually `kiwi-dracut-oem-repart` (https://packages.debian.org/bookworm/kiwi-dracut-oem-repart)

Therefore the underlying logic should be updated to check `dracut-kiwi-oem-repart || kiwi-dracut-oem-repart`

Your Debian build-test does not pick this up because you have `<oem-resize>false</oem-resize>` set in your `appliance.kiwi` for the Debian build test. 

I am not a Python coder, so I cannot offer a PR.  However a quick scan of your source code suggest this is the offending section: https://github.com/OSInside/kiwi/blob/2bd0ebf6a54db397a875d8ef807cb6aa3783a92d/kiwi/runtime_checker.py#L757C9-L757C32

schaefi:
Unfortunately the packaging of kiwi on Debian follows different naming conventions for dracut module packages which causes the runtime check to fail. This commit allows to check for multiple package names and adds the variants used on Debian. This Fixes #2524

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,8 +39,11 @@ install:
 	# kiwi default configuration
 	install -d -m 755 ${buildroot}etc
 	install -m 644 kiwi.yml ${buildroot}etc/kiwi.yml
-	# kiwi old XSL stylesheets for upgrade
+	# kiwi runtime checker metadata
 	install -d -m 755 ${buildroot}usr/share/kiwi
+	install -m 644 kiwi/runtime_checker_metadata.yml \
+		${buildroot}usr/share/kiwi/runtime_checker_metadata.yml
+	# kiwi old XSL stylesheets for upgrade
 	cp -a helper/xsl_to_v74 ${buildroot}usr/share/kiwi/
 
 tox:
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -82,6 +82,7 @@ EDIT_BOOT_INSTALL_SCRIPT = 'edit_boot_install.sh'
 IMAGE_METADATA_DIR = 'image'
 ROOT_VOLUME_NAME = 'LVRoot'
 SHARED_CACHE_DIR = '/var/cache/kiwi'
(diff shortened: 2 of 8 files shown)


Conan-Kudo:
Did we try asking the Debian maintainer to fix it?

schaefi:
> Did we try asking the Debian maintainer to fix it?

We can try. @glaubitz What's your take on the issue ? Thanks

glaubitz:
> Did we try asking the Debian maintainer to fix it?

Renaming packages in Debian is not trivial, so I prefer not to rename these packages. It also makes sense to keep the naming scheme.

> > Did we try asking the Debian maintainer to fix it?
> 
> We can try. @glaubitz What's your take on the issue ? Thanks

I prefer your fix.

schaefi:
@Conan-Kudo expected feedback, renaming is hard and will also take a long time. I would like to go ahead with the proposed solution. Thoughts ?

Conan-Kudo:
If we're going with the rule "don't assume package names" then we have to change this fix completely. The package names need to be shifted to configuration to make it patchable.

schaefi:
> If we're going with the rule "don't assume package names" then we have to change this fix completely. The package names need to be shifted to configuration to make it patchable.

Hmm, yeah ... I was too lazy but you are right

schaefi:
@Conan-Kudo I refactored in a way that we provide a yaml file with information about the package names for consumption by the runtime checker code: ```runtime_checker_metadata.yml``` If need be the file can be extended with further information used by runtime checks. For the moment only package_names are added

Conan-Kudo (approved):

PR #2530, opened 2024-04-15
Prevent extra volume mount/umount on btrfs

### LINKED ISSUE #2529: kiwi 10 fails to umount some subvolumes
## Problem description

https://build.opensuse.org/package/live_build_log/openSUSE:Factory/openSUSE-MicroOS:kvm-and-xen-sdboot/images/x86_64

Has errors like this:

```
[  104s] [ DEBUG   ]: 16:21:35 | EXEC: [umount /var/tmp/kiwi_volumes.smcpb8p1]
[  104s] [ DEBUG   ]: 16:21:35 | EXEC: Failed with stderr: umount: /var/tmp/kiwi_volumes.smcpb8p1: target is busy.
[  104s] , stdout: (no output on stdout)
[  104s] [ WARNING ]: 16:21:35 | 0 umount of /var/tmp/kiwi_volumes.smcpb8p1 failed with: umount: stderr: umount: /var/tmp/kiwi_volumes.smcpb8p1: target is busy.
[  104s] , stdout: (no output on stdout)
```

That does not happen with python3-kiwi-9.25.22-2.2

Comparing the logs one can see that the new version does not umount certain subvolume underneath `/@/.snapshots/1/snapshot`, eg there's a a line

```
[   74s] [ DEBUG   ]: 12:11:46 | EXEC: [mount -o subvol=@/home /dev/loop0p3 /var/tmp/kiwi_volumes.tis_hfxv/@/.snapshots/1/snapshot/home]
```
but no matching umount. In kiwi9 it did umount.

Build https://build.opensuse.org/package/show/openSUSE:Factory/openSUSE-MicroOS using profile kvm-and-xen-sdboot
Reproducible locally.

* KIWI version: 10.0.10-2.1
* Operating system host version: Tumbleweed
* Operating system target version: Tumbleweed

schaefi:
For setting up the read-only property an extra mount of the btrfs sub-volumes was issued. However, all volumes are mounted at that time. Thus it's not required to mount them again, resulting in a busy state because of the auto-snapshot mounts which does not get umounted and keeps a busy state until the lazy umount kicks in. This Fixes #2529

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -1727,7 +1727,6 @@ class DiskBuilder:
             )
 
         if system and self.volume_manager_name:
-            system.umount_volumes()
             custom_install_arguments.update(
                 {
                     'system_volumes': system.get_volumes(),
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -1780,9 +1780,7 @@ class DiskBuilder:
                 log.info(
                     'Setting root filesystem into read-only mode'
                 )
-                system.mount_volumes()
                 system.set_property_readonly_root()
-                system.umount_volumes()
 
     def _copy_first_boot_files_to_system_image(self) -> None:
         boot_names = self.boot_image.get_boot_names()
--- a/kiwi/volume_manager/base.py
+++ b/kiwi/volume_manager/base.py
@@ -72,10 +72,6 @@ class VolumeManagerBase(DeviceProvider):
         #: the underlaying device provider
         self.device_provider_root = device_map['root']
 
-        #: An indicator for the mount of the filesystem and its volumes
-        #: when mounted for the first time
-        self.volumes_mounted_initially = False
-
         #: root directory path name
         self.root_dir = root_dir
         #: list of volumes from :class:`XMLState::get_volumes()`
--- a/kiwi/volume_manager/btrfs.py
+++ b/kiwi/volume_manager/btrfs.py
@@ -355,13 +355,6 @@ class VolumeManagerBtrfs(VolumeManagerBase):
         )
 
         for volume_mount in self.subvol_mount_list:
-            if self.volumes_mounted_initially:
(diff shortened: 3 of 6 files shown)


lnussel:
It fixes the busy issue but the real problem is still lurking. `sync_target` is actually the root subvolume rather than snapshot 1. therefore the property ends up set on the wrong subvolume

lnussel:
I think it needs `sync_target = self.get_mountpoint()` instead of `sync_target = self.mountpoint`

schaefi:
yes it this makes sense. I'll give it a test

lnussel:
yup, with self.get_mountpoint() it works

schaefi:
@lnussel also tested this successfully if the bootloader is not custom. To make sure there are no regressions by this change I will create a Staging build after this one has been merged.

Thanks

schaefi:
waiting for staging builds to complete

schaefi:
integration tests looks good

lnussel:
any ETA? currently systemd-boot based images in Factory fail

schaefi:
> any ETA? currently systemd-boot based images in Factory fail

I was waiting for an approval from at least one involved person. I think we are good and I have a version in Staging and all integration test builds looks good. If you or Neal agrees I'll continue and make a submission today.

Conan-Kudo:
Won't this mess up non-Btrfs stuff (e.g. LVM disks)? I vaguely recall that we have this specifically to ensure that everything cleans up properly...

schaefi:
> Won't this mess up non-Btrfs stuff (e.g. LVM disks)? I vaguely recall that we have this specifically to ensure that everything cleans up properly...

I was also concerned about the other volume manager but due to the refactoring into the context managers we have proper cleanup which allowed to cleanup some places of explicit mount/umount calls. There is an integration test for testing LVM and I did a closer look to that build to check if there are any regressions. I could not find one.

Conan-Kudo (approved):

PR #2535, opened 2024-04-25
Only enable BLSCFG if the new 'bls' attribute is set

### LINKED ISSUE #2497: Add support for grub2 with BLS
Kindly be aware that this issue has been created as a reminder. The work on the grub2 component is currently in progress and considers unfinished at this time.

The [Boot Loader Specification (BLS)](https://uapi-group.org/specifications/specs/boot_loader_specification/) defines a standardized format for individual boot fragments. These fragments can be stored in a shared drop-in directory, enabling a pluggable configuration system for adding and removing entries.

One significant advantage is the improved cohesion between various boot-related tasks, as boot fragments are generated and maintained separately. This approach is architecture-agnostic, allowing seamless integration with different boot loaders.

In comparison to GRUB, which also supports loading boot fragments through the **source** command or by placing files in `/etc/grub.d` for `mkconfig` hooks, BLS has the potential to gain broader acceptance in the future. Particularly, developers, especially those associated with systemd, often prefer the BLS format for boot loaders.

## Background

The BLS integration in GRUB has been developed by Fedora. For additional context, refer to [the Fedora wiki page](https://fedoraproject.org/wiki/Changes/BootLoaderSpecByDefault) related to this topic.

It's essential to note that openSUSE do not utilize **grubby** as a unified manager for many loader configurations. Therefore, the transition from grubby to BLS may not be applicable. Instead, GRUB serves as the all-in-one loader that unifies configurations across the architectures supported by SUSE/openSUSE.

The primary motivation behind incorporating BLS into GRUB is the continual expansion of features in **systemd** that exclusively focus on producing BLS-compatible fragments. This change marks the initial step towards bridging the gap among different open-source prospects in the future.

## Locating BLS

Before proceeding, review Fedora's page on [Differences from BootloaderSpec](https://fedoraproject.org/wiki/Changes/BootLoaderSpecByDefault#Differences_from_BootLoaderSpec). This outlines where BLS is searched for when the EFI System Partition is not mounted from a Linux system. Instead, the `/boot` partition, where `grub.cfg` is loaded (i.e., the GRUB `$root` device), is used for the BLS device.

To enhance collaboration around BLS tooling, we've introduced an additional patch in GRUB to enable BLS discovery over the EFI System Partition (`/boot/efi`) in addition to the Linux system boot (`/boot`) partition.

## Getting Started

There are two ways to enable BLS testing for GRUB. One is to use `grub2-switch-to_blscfg`, a utility native to GRUB, to facilitate the transition. Another option is to try out the actual tool, `kernel-install`, which is a common method for enrolling kernels and initrd into the BLS structure.

### Installing the Test Package

```bash
sudo zypper ar https://download.opensuse.org/repositories/home:/michael-chang:/efi:/grub:/blscfg/standard/home:michael-chang:efi:grub:blscfg.repo
sudo zypper ref
sudo zypper dup --allow-vendor-change --from home_michael-chang_efi_grub_blscfg
```

### `grub2-switch-to-blscfg`

This script checks if BLS works for your GRUB installation with a single command:

```bash
sudo grub2-switch-to-blscfg
```

It creates **/loader/entries** under `/boot`, filling it with BLS fragments of discovered kernel images found in `/boot`. During this process, it enables BLS support in GRUB by setting `GRUB_ENABLE_BLSCFG=true` in `/etc/default/grub` and makes a backup copy of the original configuration to `grub.cfg.backup`.

After ensuring everything works, reboot to see if BLS entries appear in the GRUB menu:

```bash
sudo reboot
```

In case of failure after enabling BLS support, you can rescue the system at GRUB runtime by loading the backup:

```bash
configfile $prefix/grub.cfg.backup
```

### `kernel-install`

If `grub2-switch-to-blscfg` works, you can proceed to test `kernel-install`, which is part of the udev package. It is primarily used to register the kernel and initrd for systemd-boot but can also be utilized for GRUB if properly integrated with BLS.

```bash
sudo bash -c 'echo layout=bls >>/etc/kernel/install.conf'
sudo kernel-install --verbose --make-entry-directory=yes --entry-token=machine-id add 6.8.0-rc4-2.g6b6d2be-default /boot/vmlinuz-6.8.0-rc3-1.gae4495f-default
```

You can clean up the `/boot/loader` `blsdir` created by `grub2-switch-to-blscfg` as it won't interfere. After all, `kernel-install` takes over the BLS update for new kernel packages:

```bash
sudo rm -rf /boot/loader
```

### How it works

Please note that the following information reflects the current stance and may be subject to change in the future.

- The `grubbls` currently does not support the BLS generator, except for `grub2-switch-to-bls`, which generates one-off entries for testing purposes. In contrast, it collects fragments from well-known locations such as `/boot/loader/entries/` and `$esp/loader/entries` if there is an ESP present.

- The BLS menu entries are currently displayed in version order, with the newest entries at the top.

- There is no visible change in how you can update `grub.cfg`. `grub2-mkconfig` remains the go-to tool, but it is now utilized less often due to the outsourcing of updates to BLS generators.

- The default entry is set by `grub2-set-default` and stored in `/boot/grub2/grubenv`. This makes sense as it needs to support more than just BLS entries and includes mechanisms surrounding that, such as `grub2-once`.

- Currently, BLS generators are implemented by `kernel-install` for kernel updates, and `sdbootutil` has enhanced it to be Btrfs snapshot-aware.

- It is enforced that the boot target must reside on the same device where its BLS fragment is discovered.

- It is possible to support non-EFI systems if the BLS generator can output to `/boot/loader/entries`, although it's unclear if any generators in openSUSE perform this. Fedora seems to default to this approach.

- s390 support depends on `grub2-emu`, which is covered.

- The BLS generator now takes full responsibility for managing Btrfs snapshot booting, with Grub no longer handling the extensive tasks it used to perform in this process.

- The kernel command line options is totally up to the BLS generator to provide the needed config. The /etc/default/grub remains to be the source for xen hypervisor and xen kernel command lines, given BLS didn't provide corresponding function (eg multiboot2 protocol).

- Full Disk Encryption (FDE) might benefit from the change, given that the BLS format is more likely supported and welcomed by systemd.

- The `/boot/loader/loader.conf` file is unused, as it is a property of systemd-boot.

- Ultimately, it is still grub2, maintaining familiarity, and not a systemd-boot clone.

aplanas:
Partially fixes #2497 

It is not properly tested, as I did not generate images with grub2-bls (not sure how to do that outside OBS), but I added some tests to cover this feature.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -133,6 +133,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         self.timeout_style = \
             self.xml_state.get_build_type_bootloader_timeout_style()
         self.displayname = self.xml_state.xml_data.get_displayname()
+        self.bls = self.xml_state.get_build_type_bootloader_bls()
         self.serial_line_setup = \
             self.xml_state.get_build_type_bootloader_serial_line_setup()
         self.continue_on_timeout = self.get_continue_on_timeout()
@@ -794,7 +795,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 self._get_grub2_mkconfig_tool()
             ], raise_on_error=False
         )
-        if enable_blscfg_implemented.returncode == 0:
+        if self.bls and enable_blscfg_implemented.returncode == 0:
             grub_default_entries['GRUB_ENABLE_BLSCFG'] = 'true'
 
         if grub_default_entries:
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -2757,6 +2757,13 @@ div {
             sch:param [ name = "attr" value = "grub_template" ]
             sch:param [ name = "types" value = "oem iso" ]
         ]
+    k.bootloader.bls.attribute =
+        ## Specifies if the GRUB2 bootloader uses the boot loader specification
+        attribute bls { xsd:boolean }
+        >> sch:pattern [ id = "bls" is-a = "bootloader_name_type"
+            sch:param [ name = "attr" value = "bls" ]
+            sch:param [ name = "types" value = "grub2" ]
+        ]
     k.bootloader.console.attribute =
         ## Specifies the bootloader console.
         attribute console { grub_console }
@@ -2814,6 +2821,7 @@ div {
         attribute use_disk_password { xsd:boolean }
     k.bootloader.attlist =
         k.bootloader.name.attribute &
+        k.bootloader.bls.attribute? &
         k.bootloader.console.attribute? &
         k.bootloader.serial_line.attribute? &
         k.bootloader.timeout.attribute? &
--- a/kiwi/schema/kiwi.rng
+++ b/kiwi/schema/kiwi.rng
@@ -4128,6 +4128,16 @@ setting only affects the oem and iso image types.</a:documentation>
         <sch:param name="types" value="oem iso"/>
       </sch:pattern>
     </define>
+    <define name="k.bootloader.bls.attribute">
+      <attribute name="bls">
+        <a:documentation>Specifies if the GRUB2 bootloader uses the boot loader specification</a:documentation>
+        <data type="boolean"/>
+      </attribute>
+      <sch:pattern id="bls" is-a="bootloader_name_type">
+        <sch:param name="attr" value="bls"/>
+        <sch:param name="types" value="grub2"/>
+      </sch:pattern>
+    </define>
     <define name="k.bootloader.console.attribute">
       <attribute name="console">
         <a:documentation>Specifies the bootloader console.</a:documentation>
@@ -4207,6 +4217,9 @@ is useful for full disk encryption images</a:documentation>
     <define name="k.bootloader.attlist">
       <interleave>
         <ref name="k.bootloader.name.attribute"/>
+        <optional>
+          <ref name="k.bootloader.bls.attribute"/>
+        </optional>
         <optional>
           <ref name="k.bootloader.console.attribute"/>
         </optional>
--- a/kiwi/xml_parse.py
+++ b/kiwi/xml_parse.py
@@ -3,7 +3,7 @@
 
 #
 # Generated  by generateDS.py version 2.29.24.
-# Python 3.11.5 (main, Sep 06 2023, 11:21:05) [GCC]
+# Python 3.11.9 (main, Apr 08 2024, 06:18:15) [GCC]
 #
 # Command line options:
 #   ('-f', '')
@@ -16,7 +16,7 @@
 #   kiwi/schema/kiwi_for_generateDS.xsd
 #
 # Command line:
-#   /home/ms/Project/kiwi/.tox/unit_py3_11/bin/generateDS.py -f --external-encoding="utf-8" --no-dates --no-warnings -o "kiwi/xml_parse.py" kiwi/schema/kiwi_for_generateDS.xsd
+#   /home/aplanas/mnt/Documents/Work/kiwi/venv/bin/generateDS.py -f --external-encoding="utf-8" --no-dates --no-warnings -o "kiwi/xml_parse.py" kiwi/schema/kiwi_for_generateDS.xsd
 #
 # Current working directory (os.getcwd()):
 #   kiwi
@@ -5560,9 +5560,10 @@ class bootloader(GeneratedsSuper):
     provide configuration parameters for it"""
     subclass = None
     superclass = None
-    def __init__(self, name=None, console=None, serial_line=None, timeout=None, timeout_style=None, targettype=None, use_disk_password=None, grub_template=None, bootloadersettings=None):
+    def __init__(self, name=None, bls=None, console=None, serial_line=None, timeout=None, timeout_style=None, targettype=None, use_disk_password=None, grub_template=None, bootloadersettings=None):
         self.original_tagname_ = None
         self.name = _cast(None, name)
+        self.bls = _cast(bool, bls)
         self.console = _cast(None, console)
         self.serial_line = _cast(None, serial_line)
         self.timeout = _cast(int, timeout)
@@ -5586,6 +5587,8 @@ class bootloader(GeneratedsSuper):
     def set_bootloadersettings(self, bootloadersettings): self.bootloadersettings = bootloadersettings
     def get_name(self): return self.name
     def set_name(self, name): self.name = name
+    def get_bls(self): return self.bls
+    def set_bls(self, bls): self.bls = bls
     def get_console(self): return self.console
     def set_console(self, console): self.console = console
     def get_serial_line(self): return self.serial_line
@@ -5639,6 +5642,9 @@ class bootloader(GeneratedsSuper):
         if self.name is not None and 'name' not in already_processed:
             already_processed.add('name')
             outfile.write(' name=%s' % (self.gds_encode(self.gds_format_string(quote_attrib(self.name), input_name='name')), ))
+        if self.bls is not None and 'bls' not in already_processed:
+            already_processed.add('bls')
+            outfile.write(' bls="%s"' % self.gds_format_boolean(self.bls, input_name='bls'))
         if self.console is not None and 'console' not in already_processed:
             already_processed.add('console')
             outfile.write(' console=%s' % (quote_attrib(self.console), ))
@@ -5680,6 +5686,15 @@ class bootloader(GeneratedsSuper):
             already_processed.add('name')
             self.name = value
             self.name = ' '.join(self.name.split())
+        value = find_attr_value_('bls', node)
+        if value is not None and 'bls' not in already_processed:
+            already_processed.add('bls')
+            if value in ('true', '1'):
+                self.bls = True
+            elif value in ('false', '0'):
+                self.bls = False
+            else:
+                raise_parse_error(node, 'Bad boolean attribute')
         value = find_attr_value_('console', node)
         if value is not None and 'console' not in already_processed:
             already_processed.add('console')
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -998,6 +998,19 @@ class XMLState:
         return bootloader.get_name() if bootloader else \
             Defaults.get_default_bootloader()
 
+    def get_build_type_bootloader_bls(self) -> bool:
+        """
+        Return bootloader bls setting for selected build type
+
+        :return: True or False
+
+        :rtype: bool
+        """
+        bootloader = self.get_build_type_bootloader_section()
+        if bootloader:
+            return bootloader.get_bls()
+        return False
+
     def get_build_type_bootloader_console(self) -> List[str]:
         """
         Return bootloader console setting for selected build type
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -600,7 +600,6 @@ class TestBootLoaderConfigGrub2:
             'GRUB_BACKGROUND': '/boot/grub2/themes/openSUSE/background.png',
             'GRUB_CMDLINE_LINUX_DEFAULT': '"some-cmdline"',
             'GRUB_DISTRIBUTOR': '"Bob"',
-            'GRUB_ENABLE_BLSCFG': 'true',
             'GRUB_ENABLE_CRYPTODISK': 'y',
             'GRUB_GFXMODE': '800x600',
             'GRUB_SERIAL_COMMAND': '"serial --speed=38400"',
@@ -643,7 +642,6 @@ class TestBootLoaderConfigGrub2:
             call('GRUB_CMDLINE_LINUX', '"root=LABEL=some-label"'),
             call('GRUB_DISABLE_LINUX_UUID', 'true'),
             call('GRUB_DISTRIBUTOR', '"Bob"'),
-            call('GRUB_ENABLE_BLSCFG', 'true'),
             call('GRUB_ENABLE_CRYPTODISK', 'y'),
             call('GRUB_ENABLE_LINUX_LABEL', 'true'),
             call('GRUB_GFXMODE', '800x600'),
@@ -688,7 +686,6 @@ class TestBootLoaderConfigGrub2:
             call('GRUB_DISABLE_LINUX_PARTUUID', 'false'),
             call('GRUB_DISABLE_LINUX_UUID', 'true'),
             call('GRUB_DISTRIBUTOR', '"Bob"'),
-            call('GRUB_ENABLE_BLSCFG', 'true'),
             call('GRUB_ENABLE_CRYPTODISK', 'y'),
             call('GRUB_GFXMODE', '800x600'),
             call(
@@ -732,7 +729,6 @@ class TestBootLoaderConfigGrub2:
             call('GRUB_CMDLINE_LINUX_DEFAULT', '"abcd console=tty0"'),
             call('GRUB_DISABLE_LINUX_UUID', 'true'),
             call('GRUB_DISTRIBUTOR', '"Bob"'),
-            call('GRUB_ENABLE_BLSCFG', 'true'),
             call('GRUB_ENABLE_CRYPTODISK', 'y'),
             call('GRUB_ENABLE_LINUX_LABEL', 'true'),
             call('GRUB_GFXMODE', '800x600'),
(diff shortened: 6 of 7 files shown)


Conan-Kudo:
Aside from the unit test, this is going to break existing descriptions in the wild for Fedora and CentOS, especially the ones now used for Fedora 40 (the first release that uses kiwi to build its cloud images) since it turns _off_ bls by default.

I would prefer that we don't do that and instead preserve the old behavior if it's unset.

aplanas:
The old behavior (activate bls unconditionally) is the bug. Now adding a grub2 version that adds a feature (BLS) breaks all the openSUSE images and is blocking it into factory.

Fedora can fix this with one single change: adding `bls="true"`, openSUSE cannot fix it this without forcing the move to Factory.

Sadly I do not see a workaround here.

aplanas:
> Sadly I do not see a workaround here.

I want to extend and share this argument in case I am missing a better idea:

**bug**: If BLS feature is detected in GRUB2, KIWI enable this unconditionally

* Fix 1: Add a `bls="true"` attribute to the grub2 bootloader XML entry and keep `bls="false"` as default (this PR).  This will keep the upstream behavior of grub2 intact, and if you have the bls patches you need to opt-in.  Breaks Feadora as it depends on the bug description.
* Fix 2: Add a `bls="false"` and keep `true` as default. This will keep the current kiwi behavior but distributions that includes the bls patches needs to opt out.  Break all the openSUSE images by default, as for a period of time it will no use BLS by default, and so the GRUB2 patches cannot be merged in Factory
* Fix 3: Add a different bootloader type ("grub2-bls", for example).  If the bls patches are included in the normal grub2, then the result is the same as Fix 2. But then the only way to opt in is to have two GRUB2 packages in Fedora, one "grub2" without the patches and another "grub2-bls" that enable this feature.

Evaluating the cost of Fix 4 is something that the SUSE / openSUSE GRUB2 maintainers needs to calculate, but seems illogical to create a subpackage for a feature that is correctly implemented in GRUB2 (opt in) because is incorrectly implemented in KIWI (opt out)

cc: @WenhuaChang

Conan-Kudo:
Note that as the code works right now, bls support has to be detected for the flag to mean anything, so it breaks nobody who doesn't have the feature.

aplanas:
@Conan-Kudo I do not think that you are understanding what it is happening:

* openSUSE has backported the BLS patches into GRUB2 ([see entry from Tue Mar 5](https://build.opensuse.org/projects/home:michael-chang:grub:devel/packages/grub2/files/grub2.changes?expand=1))
* openSUSE wants the patch in Factory
* Because of https://github.com/OSInside/kiwi/issues/2497#issuecomment-1978609900 this **breaks** Factory
* That means that any attempt to update GRUB2 with the BLS patches gets [rejected](https://build.opensuse.org/request/show/1163418) in Factory

We want the `blscfg` command from GRUB2, but not for all the images: we want to opt in for now for a subset of images.

Do you see the issue? (Please, do not bring `grubby` in this thread)

Conan-Kudo:
I understand the issue just fine. This has nothing to do with `grubby` and everything to do with you not wanting to make a full cut-over. And the reason you don't want to has to do with the custom tooling you have in place on top of the bootloader which is not adapted fully over to BLS.

But here's the reality from the Fedora side _who also uses kiwi in production on stable images_:

* BLS is used unconditionally everywhere
* Non-BLS is effectively not supported
* A change like this effectively breaks the ability to build Fedora 40 images because the bls option didn't exist before.

Changing the default behavior for openSUSE to break Fedora is not acceptable. The other way around has not been _generally_ accepted without heads-up but has been less of a problem because openSUSE Tumbleweed is a rolling release platform with no expectations for stability.

You could have also considered subpackaging the `blscfg` GRUB module in _your_ grub2 package, but you threw that choice out and didn't present it as an option for consideration.

Conan-Kudo:
Also, as a general rule, I'm inclined to reject this if you can't even test your changes on your own images. You don't know how to build images outside of OBS? That's critically important because OBS changes the way that kiwi is used and invokes slightly different behaviors.

aplanas:
> Also, as a general rule, I'm inclined to reject this if you can't even test your changes on your own images

Why do you think that I cannot test it? I assure you that I can test it.

> You don't know how to build images outside of OBS?

....

> That's critically important because OBS changes the way that kiwi is used and invokes slightly different behaviors.

....

aplanas:
Is there a chance to refocus this in a constructive way?

Besides the options from https://github.com/OSInside/kiwi/pull/2535#issuecomment-2079474621 I guess that there is another one:

* Fix 4: The automatic detection is using `grep -q GRUB_ENABLE_BLSCFG`, so one solution is to change this name in the openSUSE package, to something like `SUSE_ENABLE_BLSCFG`, but this is more work for the GRUB2 maintainer, and I had a clear message that they want to not modify the patches, to simplify the maintenance.

aplanas:
(cont.)

I think that is very unfair to put more work on the GRUB2 maintainers (fixes 3 and 4), so this can be an alternative too:

* Fix 5: Merge this PR (or a variation of it) and in the Fedora package add a patch that revert to the previous behavior that now is a requirement.

Conan-Kudo:
And it's fair to make us deal with it? How does that make sense?! This is only happening because you don't want to cut over everything to BLS right now. Otherwise you wouldn't care.

Anyway, screw it. Clearly my needs as the community maintainer, developer, and contributor don't matter.

Let's merge this and then I can suffer for this downstream. But note that we are actively removing SUSE-specific stuff from kiwi upstream too.

Conan-Kudo (commented):
Accepted under protest.

aplanas:
> And it's fair to make us deal with it?

No. I am proposing alternatives. I want to work in the most cost effective solution.

> How does that make sense?!

It does not. I am trying to work here.

> This is only happening because you don't want to cut over everything to BLS right now.

Forcing openSUSE to move NOW to BLS is something that you perceive as fair?

> Otherwise you wouldn't care.

I am working on exploring more the alternatives, is this a sign that I do not care?
 
> Anyway, screw it. Clearly my needs as the community maintainer, developer, and contributor don't matter.

....

> Let's merge this and then I can suffer for this downstream. But note that we are actively removing SUSE-specific stuff from kiwi upstream too.

Can we please work on this before merging it? I do not think that this is time critical, can we give us one week to explore better the space problem?

aplanas:
Oh my ... 

Why has this been merged?

@Conan-Kudo can we please work in a good way? I do really want a good solution. Can we please do the work, instead merging this without a proper review?

aplanas:
@Conan-Kudo can you revert this and give us another opportunity to work together?

I do think that this is what is hurting the project and the community. Instead of working we are wasting time in discussions and it is not helping in exploring the solution space. A PR is an opportunity to review the code (in this case I had none), the alternatives and the cost of them.

Can we please do the job?

Conan-Kudo:
It's been merged because the code is actually fine, the implications are what suck. But I am the only kiwi maintainer not employed by SUSE, and if I blocked this, it would continue to become uncomfortable for everyone. This is not the first time we've had this problem related to grub stuff, and I expect it won't be the last.

I can suck it up and deal with it downstream in Fedora, Fedora Asahi, CentOS Stream, and CentOS Hyperscale somehow. it's going to be a bit of a mess because I have no synchronization point for Fedora.

At the end of the day, the real problem is that despite repeated attempts to encourage collaboration both upstream and downstream on grub2 work, nobody wants to. So we continue having this mess and kiwi is caught in the middle of it.

It has been 10 years, and none of the SUSE-specific Btrfs features for grub have been submitted upstream. To be fair, the same is also true for blscfg, but at least that's just a module instead of changes to multiple things in GRUB.

I'm tired of this. I'm tired of nobody wanting to work together on anything for the boot stuff. I'm tired being caught in the middle over it too. It's because of this that GRUB development upstream slowed to a crawl. Nobody who does this work for distributions wants to do anything upstream.

Conan-Kudo:
I will also add a final point: I asked multiple times about rolling out BLS support in openSUSE's GRUB when we first added it to kiwi years ago with RHEL 8. Nobody responded to my messages _at all_. I wasn't even the only person asking for it.

Nobody commented then, nobody cared. And here we are again because nobody pays attention. If nothing changes, this will keep happening.

aplanas:
> It's been merged because the code is actually fine, the implications are what suck. 

Before rushing the merge, can we work on alternatives with cheaper implications? The cost of this one is to add a field in the XML in the KIWI files controlled by Fedora, but I understand that there are more like that in the wild.

> But I am the only kiwi maintainer not employed by SUSE, and if I blocked this, it would continue to become uncomfortable for everyone.

There is not reason to be uncomfortable. This PR is me working with upstream (you). This change is not for SUSE but for openSUSE (Factory).
 
> I can suck it up and deal with it downstream in Fedora, Fedora Asahi, CentOS Stream, and CentOS Hyperscale somehow. it's going to be a bit of a mess because I have no synchronization point for Fedora.

You mean adding a patch in those packages that revert this change? Can this be, somehow, be alleviated?

If your decision is to add this patch that revert this change, I propose to add in those projects the "bls=true" in the XML that will imply that the patch in the package will be just one line: the `and` in the condition, keeping the change in the XML parse and the rest.  This will give time to this change spread.

> At the end of the day, the real problem is that despite repeated attempts to encourage collaboration both upstream and downstream on grub2 work, nobody wants to. 

But I am willing! Why are you assuming otherwise? In what moment did I give you this impression? I am not forcing this PR, and I was opening other alternatives. When did I pushed the message that I do not want to collaborate?

> It has been 10 years, and none of the SUSE-specific Btrfs features for grub have been submitted upstream. To be fair, the same is also true for blscfg, but at least that's just a module instead of changes to multiple things in GRUB.

I am not a grub2 maintainer but I talked about this issue long enough (https://news.opensuse.org/2023/12/20/systemd-fde/). This is a big issue, and upstream GRUB2 and the openSUSE maintainer are aware of it and they are making big efforts to fix it (check the FOSDEM talk to see the upstream position, and I guarantee you that the downstream maintainer are not happy to have more that 200 patches to take care of)

> I'm tired of this. I'm tired of nobody wanting to work together on anything for the boot stuff. I'm tired being caught in the middle over it too. It's because of this that GRUB development upstream slowed to a crawl. Nobody who does this work for distributions wants to do anything upstream.

I do not have a solution neither ... but I feel you

aplanas:
> I will also add a final point: I asked multiple times about rolling out BLS support in openSUSE's GRUB when we first added it to kiwi years ago with RHEL 8 support. Nobody responded to my messages _at all_. I wasn't even the only person asking for it.

But what do you think that we are doing??!!

We are doing steps: we have a design that requires systemd-boot for FDE, we relaxed this requirement to just have BLS, we moved the BLS patches to the openSUSE GRUB2, and we changed KIWI to have this patches in Factory. Don't you see the direction??

Conan-Kudo:
Rea

> > It's been merged because the code is actually fine, the implications are what suck.
> 
> Before rushing the merge, can we work on alternatives with cheaper implications? The cost of this one is to add a field in the XML in the KIWI files controlled by Fedora, but I understand that there are more like that in the wild.
> 
> > But I am the only kiwi maintainer not employed by SUSE, and if I blocked this, it would continue to become uncomfortable for everyone.
> 
> There is not reason to be uncomfortable. This PR is me working with upstream (you). This change is not for SUSE but for openSUSE (Factory).
> 
> > I can suck it up and deal with it downstream in Fedora, Fedora Asahi, CentOS Stream, and CentOS Hyperscale somehow. it's going to be a bit of a mess because I have no synchronization point for Fedora.
> 
> You mean adding a patch in those packages that revert this change? Can this be, somehow, be alleviated?
> 
> If your decision is to add this patch that revert this change, I propose to add in those projects the "bls=true" in the XML that will imply that the patch in the package will be just one line: the `and` in the condition, keeping the change in the XML parse and the rest. This will give time to this change spread.
> 

Realistically, there are only two workable options for compatibility now that we've locked in kiwi 10's behaviors:

* Unset `bls` goes to old behavior (effectively default to `bls="true"` because there's lots of private kiwi descriptions that will break otherwise), we would switch it for SUSE distributions only to default to false until grub2 with bls is propagated
* Total revert and we add a `grub2-no-bls` bootloader choice that removes the bls setting

> > At the end of the day, the real problem is that despite repeated attempts to encourage collaboration both upstream and downstream on grub2 work, nobody wants to.
> 
> But I am willing! Why are you assuming otherwise? In what moment did I give you this impression? I am not forcing this PR, and I was opening other alternatives. When did I pushed the message that I do not want to collaborate?
> 
> > It has been 10 years, and none of the SUSE-specific Btrfs features for grub have been submitted upstream. To be fair, the same is also true for blscfg, but at least that's just a module instead of changes to multiple things in GRUB.
> 
> I am not a grub2 maintainer but I talked about this issue long enough (https://news.opensuse.org/2023/12/20/systemd-fde/). This is a big issue, and upstream GRUB2 and the openSUSE maintainer are aware of it and they are making big efforts to fix it (check the FOSDEM talk to see the upstream position, and I guarantee you that the downstream maintainer are not happy to have more that 200 patches to take care of)
> 

But I asked about this in 2019. And back then the GRUB maintainers in openSUSE ignored my requests.

I know nobody is happy about 200+ patches, but it doesn't seem to motivate anyone to do anything about it.

> > I'm tired of this. I'm tired of nobody wanting to work together on anything for the boot stuff. I'm tired being caught in the middle over it too. It's because of this that GRUB development upstream slowed to a crawl. Nobody who does this work for distributions wants to do anything upstream.
> 
> I do not have a solution neither ... but I feel you

If only I could get everyone into a room...

aplanas:
>     * Unset `bls` goes to old behavior (effectively default to `bls="true"` because there's lots of private kiwi descriptions that will break otherwise), we would switch it for SUSE distributions only to default to false until grub2 with bls is propagated

Yes, that was my proposal too.

>     * Total revert and we add a `grub2-no-bls` bootloader choice that removes the bls setting

That is a bad idea as will make valid XMLs invalid (there is an option that is not recognized).

I propose this patch for Fedora, this will keep the old behavior but will accept the new parameter (

```diff
diff --git a/kiwi/bootloader/config/grub2.py b/kiwi/bootloader/config/grub2.py
index dcd03534f..101b28a96 100644
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -795,7 +795,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                 self._get_grub2_mkconfig_tool()
             ], raise_on_error=False
         )
-        if self.bls and enable_blscfg_implemented.returncode == 0:
+        if enable_blscfg_implemented.returncode == 0:
             grub_default_entries['GRUB_ENABLE_BLSCFG'] = 'true'
 
         if grub_default_entries:
diff --git a/test/unit/bootloader/config/grub2_test.py b/test/unit/bootloader/config/grub2_test.py
index d81d76ec7..e9189a609 100644
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -600,6 +600,7 @@ class TestBootLoaderConfigGrub2:
             'GRUB_BACKGROUND': '/boot/grub2/themes/openSUSE/background.png',
             'GRUB_CMDLINE_LINUX_DEFAULT': '"some-cmdline"',
             'GRUB_DISTRIBUTOR': '"Bob"',
+            'GRUB_ENABLE_BLSCFG': 'true',
             'GRUB_ENABLE_CRYPTODISK': 'y',
             'GRUB_GFXMODE': '800x600',
             'GRUB_SERIAL_COMMAND': '"serial --speed=38400"',
@@ -642,6 +643,7 @@ class TestBootLoaderConfigGrub2:
             call('GRUB_CMDLINE_LINUX', '"root=LABEL=some-label"'),
             call('GRUB_DISABLE_LINUX_UUID', 'true'),
             call('GRUB_DISTRIBUTOR', '"Bob"'),
+            call('GRUB_ENABLE_BLSCFG', 'true'),
             call('GRUB_ENABLE_CRYPTODISK', 'y'),
             call('GRUB_ENABLE_LINUX_LABEL', 'true'),
             call('GRUB_GFXMODE', '800x600'),
@@ -686,6 +688,7 @@ class TestBootLoaderConfigGrub2:
             call('GRUB_DISABLE_LINUX_PARTUUID', 'false'),
             call('GRUB_DISABLE_LINUX_UUID', 'true'),
             call('GRUB_DISTRIBUTOR', '"Bob"'),
+            call('GRUB_ENABLE_BLSCFG', 'true'),
             call('GRUB_ENABLE_CRYPTODISK', 'y'),
             call('GRUB_GFXMODE', '800x600'),
             call(
@@ -729,6 +732,7 @@ class TestBootLoaderConfigGrub2:
             call('GRUB_CMDLINE_LINUX_DEFAULT', '"abcd console=tty0"'),
             call('GRUB_DISABLE_LINUX_UUID', 'true'),
             call('GRUB_DISTRIBUTOR', '"Bob"'),
+            call('GRUB_ENABLE_BLSCFG', 'true'),
             call('GRUB_ENABLE_CRYPTODISK', 'y'),
             call('GRUB_ENABLE_LINUX_LABEL', 'true'),
             call('GRUB_GFXMODE', '800x600'),

```

WenhuaChang:
> At the end of the day, the real problem is that despite repeated attempts to encourage collaboration both upstream and downstream on grub2 work, nobody wants to. So we continue having this mess and kiwi is caught in the middle of it.
> 
> It has been 10 years, and none of the SUSE-specific Btrfs features for grub have been submitted upstream. To be fair, the same is also true for blscfg, but at least that's just a module instead of changes to multiple things in GRUB.
> 
> I'm tired of this. I'm tired of nobody wanting to work together on anything for the boot stuff. I'm tired being caught in the middle over it too. It's because of this that GRUB development upstream slowed to a crawl. Nobody who does this work for distributions wants to do anything upstream.

I am sorry to make you feel so. But we did before submitting a series of patches, we initially sent a foundational patch to the upstream, but it did not proceed as expected. You can view this initial communication here:

https://mail.gnu.org/archive/html/grub-devel/2020-07/msg00000.html

To clarify the issue with the upstream patch: the upstream btrfs file system uses a clear, unambiguous path system that originates from the top-level tree. This approach is preferred because it allows files to be consistently located and shared with other grub utilities without the need for specific modifications.

However, for the btrfs snapshots, particularly in systems like SUSE/openSUSE, it is necessary to use paths relative to the subvolume. This requirement arises because it is not possible to modify the grub.cfg file within a read-only (ro) snapshot to update hardcoded absolute paths to reflect their new locations within a snapshot.

Meanwhile the downside of using relative paths is evident. It introduces the need for additional constructs during grub's runtime, such as "get-default", "set-subvolume" and "mount-subvolume". These additions complicate automatic setup tools like grub-install, grub-mkconfig and grub-mkrelpath, which now require more sophisticated handling, therefore increasing the overall complexity — particularly in environments that do not use similar btrfs boot configuration introducing the change is not welcome.

To be honest, I have revisited this issue regularly but have not yet been able to devise a satisfactory solution.

WenhuaChang:
Why can't we leave GRUB_ENABLE_BLSCFG unset and allow it to follow the distribution's default settings? For example, Fedora enables BLS by default if grubby is not used, while openSUSE does not enable BLS at all.

https://github.com/rhboot/grub2/blob/fedora-39/util/grub.d/10_linux.in#L203

Conan-Kudo:
Because in practice, `grubby` has to be installed on Fedora systems or the kernel package installs things to the wrong place. That's why our kiwi descriptions ensure it: https://pagure.io/fedora-kiwi-descriptions/c/31f9f8a7437ffe6cb5a9a761dfac62949f8b07ff

PR #2536, opened 2024-04-25
Fix package removal with dnf5

AdamWill:
dnf5 does not implement `dnf autoremove <package>` as a synonym for `--setopt=clean_requirements_on_remove=true remove <package>` as dnf4 did. So, we should do it this way instead.


AdamWill:
btw, why on earth are both 'main' and 'master' active branches in this repo, with differing commit histories? that seems very odd.

Conan-Kudo:
`master` is the old kiwi9 head, `main` is the current tree.

Conan-Kudo:
Take it up with @schaefi 😄

PR #2538, opened 2024-04-26
Stop leaking plugins/{priorities,versionlock}.conf file with dnf4/5.

Romain-Geissler-1A:
This fixes the fedora issue
https://bugzilla.redhat.com/show_bug.cgi?id=2270364

Description of the issue:
With Fedora 40 it seems the Fedora base image moved to being created with Kiwi while apparently Fedora <= 39 used something else. Starting with fedora 40, we can see that all OCI fedora images have one new file `/etc/dnf/plugins/versionlock.conf` which explicitly disables the versionlock plugin. If later, with this generated OCI fedora image you try to install the versionlock plugin via dnf (ie package `python3-dnf-plugin-versionlock` then since the config file `/etc/dnf/plugins/versionlock.conf` exists already, the one from the package is ignored. The end result is that when users of fedora try to install `python3-dnf-plugin-versionlock` in the end the versionlock plugin is still disabled, which is not expected.

In order to fix this, I can see two possible ways. Either adding some explicit code in the `cleanup` method to correctly remove the plugin configuration files. However I really don't know your codebase and I am quite unsure when exactly is cleanup called, and to be honest it would strange to me to explicitly remove files here. Indeed if one day in the fedora base image the package `python3-dnf-plugin-versionlock` is pre-installed then... you end up removing an actual packaged file.

The other way to fix this, and the solution I propose here is to stop creating any plugin configuration file, since the only thing kiwi seems to care about is enabling/disabling plugins, and instead achieve the same activation/deactivation via command line arguments.

I discovered kiwi today, I am just trying to fix an issue I have with fedora OCI images, so excuse my lack of internal knowledge in your project ;)

Note: I tested only the dnf4 code (re-building locally the fedora images to fix my initial versionlock issue), the dnf5 code change was made blindly, reading the doc as it seems the `--disable-plugin` flag name has been renamed compared to dnf4.

### DIFF
--- a/kiwi/repository/dnf4.py
+++ b/kiwi/repository/dnf4.py
@@ -100,14 +100,13 @@ class RepositoryDnf4(RepositoryBase):
         ).unmanaged_file()
 
         self.dnf_args = [
-            '--config', self.runtime_dnf_config_file.name, '-y'
+            '--config', self.runtime_dnf_config_file.name, '-y', '--enableplugin=priorities', '--disableplugin=versionlock'
         ] + self.custom_args
 
         self.command_env = self._create_dnf_runtime_environment()
 
         # config file parameters for dnf tool
         self._create_runtime_config_parser()
-        self._create_runtime_plugin_config_parsers()
         self._write_runtime_config()
 
     def setup_package_database_configuration(self) -> None:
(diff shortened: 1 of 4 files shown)


Conan-Kudo:
This is on the right track, but it needs updates to the unit tests too. Can you try to fix them?

Romain-Geissler-1A:
Ah sorry, I saw a green CI so I didn't bother looking at this more thoroughly, I didn't expect your CI only runs if changes are kind of "pre-approved". I guess it should be ok now, at least locally it seems better now.

Romain-Geissler-1A:
Note: if this is accepted here, since it seems you also maintain the fedora package, I already submitted the backport on fedora side here ;) https://src.fedoraproject.org/rpms/kiwi/pull-request/4

Conan-Kudo (approved):
Looks reasonable to me.

PR #2541, opened 2024-04-29
Overwrite compression setting only if randomized

### LINKED ISSUE #2540: Build compressed LUKS images
## Problem description

When building encrypted images, the final image is not compressed. Due to this condition here:

https://github.com/OSInside/kiwi/blob/main/kiwi/builder/disk.py#L402-L403

## Expected behaviour

Encrypted images are usually quite good to compress. For two reasons:

1. The file system is not 100% full. For example, SUSE Micro images are typically filled about 60%.
2. You might choose a weak encryption cipher intentionally. If you distribute images publicly you force the user to reencrypt the LUKS volume (typically on first start) anyway, as the master key is obviously publicly known. So you can use the `null` cipher, for example. `cryptsetup reencrypt` even supports this scheme by switching automatically from `null`  to the default cipher.

It might make more sense to depend on the `luks_randomize` setting instead, as this makes the image truly incompressible.

Build any LUKS encrypted OEM image.

* latest KIWI

schaefi:
When building an encrypted image, the bundler never compressed the result. This overwrite from the runtime configuration and the default compression setting actually only makes sense when the image is randomized because only then a compression is for sure useless. This Fixes #2540

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -399,7 +399,7 @@ class DiskBuilder:
 
         # store image file name in result
         compression = self.runtime_config.get_bundle_compression(default=True)
-        if self.luks is not None:
+        if self.luks is not None and self.luks_randomize:
             compression = False
         self.result.add(
             key='disk_image',


wfeldt (approved):
tested and works

PR #2544, opened 2024-05-02
Fix set_disk_password() to really add the password to grub.cfg.

wfeldt:
## Problem

The SL-Micro -6.0 encrypted images - when built with latest KIWI 10.0 - ask the user for a decryption password during initial boot.

The reason is that usually the password is stored in `grub.cfg` and passed to grub's `cryptomount`.

But KIWI 10.0 fails to create `grub.cfg` correctly. KIWI 9.25 works.

## Solution

Since commit 8aa517eb7 it is necessary to call _mount_device_and_volumes() in methods that want to modify the grub setup to ensure the current boot files are mounted.

Also, raise an error if the config file could not be modified.

### DIFF
--- a/kiwi/bootloader/install/grub2.py
+++ b/kiwi/bootloader/install/grub2.py
@@ -287,22 +287,26 @@ class BootLoaderInstallGrub2(BootLoaderInstallBase):
                         self._enable_grub2_install(self.root_mount.mountpoint)
 
     def set_disk_password(self, password: str):
-        if self.root_mount and password is not None:
-            log.debug('Include cryptomount credentials...')
-            config_file = '{0}/boot/efi/EFI/BOOT/grub.cfg'.format(
-                self.root_mount.mountpoint
-            )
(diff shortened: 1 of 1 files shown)


schaefi:
good catch, thanks for finding and fixing. You did not update the unit tests for the change, thus it fails there. Let me know if you want help for fixing those. Thanks

wfeldt:
Uhm, I see this involves some mock-ups I'm not familiar with in KIWI, so it'd be great if you could adjust the tests. :-)

schaefi:
> Uhm, I see this involves some mock-ups I'm not familiar with in KIWI, so it'd be great if you could adjust the tests. :-)

ok I will. I think I cannot push to your fork. I'll move your change into a branch. btw I added you with write permissions to the project so you can create PRs based on branches :)

schaefi:
continue work on #2546

wfeldt:
Thanks! Let's hope I won't need it. :-)

PR #2546, opened 2024-05-03
Fix set_disk_password() to really add the password to grub.cfg.

schaefi:
Since commit 8aa517eb7 it is necessary to call _mount_device_and_volumes() in methods that want to modify the grub setup to ensure the current boot files are mounted.

Also, raise an error if the config file could not be modified.

### DIFF
--- a/test/unit/bootloader/install/grub2_test.py
+++ b/test/unit/bootloader/install/grub2_test.py
@@ -11,7 +11,8 @@ from kiwi.defaults import Defaults
 from kiwi.exceptions import (
     KiwiBootLoaderGrubInstallError,
     KiwiBootLoaderGrubDataError,
-    KiwiBootLoaderGrubPlatformError
+    KiwiBootLoaderGrubPlatformError,
+    KiwiBootLoaderDiskPasswordError
 )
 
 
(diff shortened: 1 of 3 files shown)


schaefi:
@wfeldt will commit the test update next...

test/unit/bootloader/install/grub2_test.py
```diff
@@ -448,8 +449,19 @@ def side_effect(device, mountpoint=None):
             filename='shim-install', root_dir='tmp_root'
         )
 
+    @patch('kiwi.bootloader.install.grub2.Command.run')
+    @patch('kiwi.bootloader.install.grub2.MountManager')
     @patch('os.path.isfile')
-    def test_set_disk_password(self, mock_os_path_is_file):
+    def test_set_disk_password(
+        self, mock_os_path_is_file, mock_mount_manager, mock_command
```
wfeldt, line 456: Since you optimized out `os.path.isfile()`, `mock_os_path_is_file` is no longer needed, or?

schaefi: right, this mock can be deleted from the test

schaefi: done

wfeldt (approved):
lgtm; I've run test builds for the working and non-working case and it works as expected.

PR #2547, opened 2024-05-08
Fix spec file

schaefi:
Require docopt-ng for Fedora 41+


schaefi:
The spec contains the modification to the pyproject.toml but not to the requires of the package. When building for Rawhide I get a conflict saying that docopt conflicts with docopt-ng, I was not able to set an ignore statement or something to workaround this in the buildservice and actually I think it's a bug in the spec

Conan-Kudo (approved):

PR #2555, opened 2024-05-29
package: adjust openSUSE patch

aplanas:
Add the openSUSE patch as a `Patch1` element, and set `%autosetup` to `-p1`

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -57,6 +57,11 @@ Packager:       Marcus Schaefer <marcus.schaefer@suse.com>
 Group:          %{pygroup}
 Source:         %{name}.tar.gz
 Source1:        %{name}-rpmlintrc
+# PATCH-FIX-OPENSUSE kiwi-revert-bls-default-for-suse.patch
+# Temporarily revert grub-bls default to false for SUSE distributions
+%if 0%{?suse_version}
+Patch:          kiwi-revert-bls-default-for-suse.patch
+%endif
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
This commit means that SRPMs now can conditionally be broken by not including the patch. I did it differently _on purpose_ to avoid that.

aplanas:
> This commit means that SRPMs now can conditionally be broken by not including the patch. 

The patch is under a macro guard. Would not this resolve this issue?

> I did it differently _on purpose_ to avoid that.

Right, the problem is that the Factory / OBS packaging policies[1] require certain properties for a patch: needs to be referenced as a "Patch:", needs to be in the changelog, and optionally tagged.

[1] https://en.opensuse.org/openSUSE:Packaging_Patches_guidelines

Conan-Kudo:
Macro guards are the problem. That's not allowed for `Patch` or `Source` because it means the SRPM can be generated in a way that's broken.

aplanas:
Without breaking the policy (that makes the submission automatically declined), what can be a good solution then? Maybe creating a different spec template only for openSUSE?

aplanas:
Maybe unconditionally add the patch in the spec file, remove the autosetup and manually apply the patch with "%patch"? That seems a bit of step backwards to me (I like autosetup)

Conan-Kudo:
I like autosetup too, but `%autosetup -N` and `%autosetup -m/-M` aren't supported everywhere so we have to roll back a bit.

https://github.com/OSInside/kiwi/pull/2556

PR #2556, opened 2024-05-29
package: Unconditionally include patches and number all sources and patches

Conan-Kudo:
This ensures that stuff is applied reliably and all sources and patches are included as expected.

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -55,13 +55,11 @@ License:        GPL-3.0-or-later
 Packager:       Marcus Schaefer <marcus.schaefer@suse.com>
 %endif
 Group:          %{pygroup}
-Source:         %{name}.tar.gz
+Source0:        %{name}.tar.gz
 Source1:        %{name}-rpmlintrc
-# PATCH-FIX-OPENSUSE kiwi-revert-bls-default-for-suse.patch
-# Temporarily revert grub-bls default to false for SUSE distributions
-%if 0%{?suse_version}
-Patch:          kiwi-revert-bls-default-for-suse.patch
-%endif
+# SUSE-specific patches (1001+)
+## PATCH-FIX-OPENSUSE kiwi-revert-bls-default-for-suse.patch -- temporary until opensuse has bls
+Patch1001:      kiwi-revert-bls-default-for-suse.patch
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
 %if 0%{?fedora} || 0%{?suse_version}
 BuildRequires:  fdupes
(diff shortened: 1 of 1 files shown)


package/python-kiwi-spec-template
```diff
@@ -641,7 +639,12 @@ Group:          %{sysgroup}
 Provides manual pages to describe the kiwi commands
 
 %prep
-%autosetup -p1 -n kiwi-%{version}
+%setup -q -n kiwi-%{version}
+
+%if 0%{?suse_version}
+# Apply SUSE specific patches
```
aplanas, line 645: Maybe you want to replace the comment from the one in L61, to avoid the repetition of being SUSE specific.

package/python-kiwi-spec-template
```diff
@@ -641,7 +639,12 @@ Group:          %{sysgroup}
 Provides manual pages to describe the kiwi commands
 
 %prep
-%autosetup -p1 -n kiwi-%{version}
+%setup -q -n kiwi-%{version}
+
+%if 0%{?suse_version}
+# Temporarily revert grub-bls default to false for SUSE distributions
```
aplanas, line 645: : D I mean move (sorry)

package/python-kiwi-spec-template
```diff
@@ -55,13 +55,11 @@ License:        GPL-3.0-or-later
...
-Source:         %{name}.tar.gz
+Source0:        %{name}.tar.gz
 Source1:        %{name}-rpmlintrc
-# PATCH-FIX-OPENSUSE kiwi-revert-bls-default-for-suse.patch
-# Temporarily revert grub-bls default to false for SUSE distributions
-%if 0%{?suse_version}
-Patch:          kiwi-revert-bls-default-for-suse.patch
-%endif
+# SUSE-specific patches (1001+)
+## PATCH-FIX-OPENSUSE kiwi-revert-bls-default-for-suse.patch -- temporary until opensuse has bls
```
aplanas, line 61: Not sure that openSUSE nor SUSE will have BLS generally ...

aplanas (approved):

package/python-kiwi-spec-template
Conan-Kudo: It'll happen. If for nothing else, because we in kiwi will push for it harder as we don't want to support more weird ways to do things.

schaefi:
@Conan-Kudo the SUSE package checks also checks the changelog for an entry of the form

```
Added patch kiwi-revert-bls-default-for-suse.patch
```

In kiwi we create the changelog from the git history. My latest submission got declined because we added the patch file but the produced changelog missed the information from above

Could you write the commit message of this PR in a way that the sentence from above is part of the commit message text ?

Thanks a ton

Conan-Kudo:
> Looks good, thanks
> 
> out of curiosity, why does the patch numbering starts with 1001 ?

If this package is forked by SLE downstream and backports are made, they can be in 0~1000.

schaefi:
> > Looks good, thanks
> > out of curiosity, why does the patch numbering starts with 1001 ?
> 
> If this package is forked by SLE downstream and backports are made, they can be in 0~1000.

Ah ok got it

Conan-Kudo:
For example: https://src.fedoraproject.org/rpms/kiwi/blob/rawhide/f/kiwi.spec

schaefi:
> @schaefi is the commit message sufficient now?

Only God knows that. @aplanas can you double check if it's enough that the name of the patch file appears in the changelog or do we also need a specific format sentence like "Added patch ..." ? Thanks

aplanas (approved):
LGTM!

Issue #38, opened 2016-03-01
OSTree - git for operating system binaries


schaefi:
With rpm-ostree, you can create versioned filesystem trees to provide atomic, reliable upgrades. "git-like" management of appliances would be pretty sweet.

schaefi:
https://wiki.gnome.org/Projects/OSTree

Conan-Kudo:
RPM-OSTree package for openSUSE Leap 15.0 / Tumbleweed: https://build.opensuse.org/package/show/home:Pharaoh_Atem:SUSE_RPMOSTree/rpm-ostree

Docs and examples:

* https://github.com/projectatomic/rpm-ostree/blob/master/docs/manual/treefile.md
* https://pagure.io/fedora-atomic/tree/master
* https://pagure.io/workstation-ostree-config/tree/master
* https://pagure.io/fork/siosm/workstation-ostree-config/tree/f28-kinoite

davidcassany:
@Conan-Kudo I am still wondering two things:

1. What actually means supporting rpm-ostree in KIWI? Probably this caused by my lack of knowledge regarding rpm-ostree. I assume this is a new package manager right? or at most an extension of the feature set of DNF and/or any other that has some layer of cooperation.

2. Which is the use case? According to what I understand this is not much different of building a root-tree over an ostree, thus building rootfs in some container like layered filesystem, correct? So I am wondering what is actually the result of using rpm-ostree as a package manager for an e.g. Fedora build. Mostly because if rpm-ostree is a package manager (from KIWI POV) then it should not have impact on the build procedure. It should still be possible to build isos, vms, oem, etc. 

I can imagine features like usign the `derived_from` attribute to import an ostree and create a new images out of it by adding a new transaction on top with rpm-ostree. Am I pointing to the right direction?

Also stuff like making each `<packages>` section as an individual transaction, even I am having hard time to see the use case and potential benefits of it.

We need to refine this support to narrow what actually means and also understand the use cases we are covering with it. @Conan-Kudo would be appreciated if you can help us on that regard.

Conan-Kudo:
[RPM-OSTree](https://rpm-ostree.readthedocs.io/en/latest/) produces an OSTree, which is a special rootfs type representing a largely immutable operating system environment. The `rpm-ostree` program has features for producing and managing OSTrees from RPM content. It does leverage DNF stack libraries, but it does not have the same interface as DNF.

[...]

PR #2557, opened 2024-05-30
Initial support for ostree based systems

travier:
defaults: Add method to check for ostree case

Temporary method to place all ostree specific code paths behind a single
check. May be converted to configuration option later.

---

bootloader/config/grub2: Handle ostree case for BLS configs

- Do not replace but append to the existing kernel command line as it is
  correct
- Do not fix the path to the kernel & initrd as those are also correct

---

system/kernel: Find kernel in ostree case

---

kiwi/builder/live: Handle ostree case

Do not rebuild the initrd in the ostree case and use the one from the
ostree commit instead. It is expected that it includes all the needed
modules.

---

kiwi/bootloader/config/grub2: Handle ostree case

---

See: https://github.com/OSInside/kiwi/issues/38

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -319,13 +319,10 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         # be deleted...
         if self.xml_state.build_type.get_overlayroot_write_partition() is not False:
             self._fix_grub_root_device_reference(config_file, boot_options)
-            if not Defaults.is_ostree(self.root_dir):
-                self._fix_grub_loader_entries_boot_cmdline()
-                self._fix_grub_loader_entries_linux_and_initrd_paths()
-            else:
-                # In the ostree case, the BLS entries are correct, but we still
-                # need to append the additionnal kernel command line arguments
-                self.grub_loader_entries_boot_cmdline_append()
+            self._fix_grub_loader_entries_boot_cmdline(
+                keep_existing_options=True if Defaults.is_ostree(self.root_dir) else False
+            )
+            self._fix_grub_loader_entries_linux_and_initrd_paths()
 
         if self.firmware.efi_mode() and self.early_boot_script_efi:
             self._copy_grub_config_to_efi_path(
@@ -425,7 +422,8 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             has_serial = True
 
         # Find the ostree=... parameter from the BLS config and add it to the boot options
-        if not Defaults.is_ostree(self.root_dir):
+        # TODO: ostree: This code looks like it is required for any ostree based boot process and should live in its own scope e.g. an OSTree class
+        if Defaults.is_ostree(self.root_dir):  # pragma: nocover
             loader_entries_pattern = os.sep.join(
                 [
                     self.root_mount.mountpoint,
@@ -1482,7 +1480,9 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                     with open(vendor_grubenv_file, 'w') as vendor_grubenv:
                         vendor_grubenv.write(grubenv)
 
-    def _fix_grub_loader_entries_boot_cmdline(self):
+    def _fix_grub_loader_entries_boot_cmdline(
+        self, keep_existing_options=False
+    ):
         if self.cmdline:
             # For distributions that follows the bootloader spec here:
             # https://www.freedesktop.org/wiki/Specifications/BootLoaderSpec
@@ -1506,30 +1506,18 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             for menu_entry_file in glob.iglob(loader_entries_pattern):
                 with open(menu_entry_file) as grub_menu_entry_file:
                     menu_entry = grub_menu_entry_file.read()
-                    menu_entry = re.sub(
-                        r'options (.*)',
-                        r'options {0}'.format(self.cmdline),
-                        menu_entry
-                    )
-                with open(menu_entry_file, 'w') as grub_menu_entry_file:
-                    grub_menu_entry_file.write(menu_entry)
-
-    def grub_loader_entries_boot_cmdline_append(self):
-        if self.cmdline:
-            loader_entries_pattern = os.sep.join(
-                [
-                    self.root_mount.mountpoint,
-                    'boot', 'loader', 'entries', '*.conf'
-                ]
-            )
-            for menu_entry_file in glob.iglob(loader_entries_pattern):
-                with open(menu_entry_file) as grub_menu_entry_file:
-                    menu_entry = grub_menu_entry_file.read()
-                    menu_entry = re.sub(
-                        r'options (.*)',
-                        r'options \1 {0}'.format(self.cmdline),
-                        menu_entry
-                    )
+                    if keep_existing_options:
+                        menu_entry = re.sub(
+                            r'options (.*)',
+                            r'options \1 {0}'.format(self.cmdline),
+                            menu_entry
+                        )
+                    else:
+                        menu_entry = re.sub(
+                            r'options (.*)',
+                            r'options {0}'.format(self.cmdline),
+                            menu_entry
+                        )
                 with open(menu_entry_file, 'w') as grub_menu_entry_file:
                     grub_menu_entry_file.write(menu_entry)
 
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -209,8 +209,8 @@ class LiveImageBuilder:
                                 )
                             ]
                         )
-            else:
-                # find the existing initrd in the ostree case
+            else:  # pragma: nocover
+                # TODO: ostree: this code should be part of an OSTree class find the existing initrd in the ostree case
                 boot_ostree_dir = os.sep.join([self.root_dir, 'boot/ostree'])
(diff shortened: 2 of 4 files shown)


travier:
This is definitely work in progress. I'm looking for suggestions on how we should implement this properly.

kiwi/system/kernel.py
```diff
@@ -90,6 +90,27 @@ def get_kernel(
...
+                        version_match = re.match(kernel_ostree_pattern, kernel_file)
+                        if version_match:
+                            version = version_match.group(1)
+                            return kernel_type(
+                                name=os.path.basename(
+                                    os.path.realpath(kernel_file)
+                                ),
+                                filename=kernel_file,
+                                version=version
+                            )
```
schaefi, line 112: I think we should move that code into an extra method.

Other than that I believe it does not have unwanted side effects as it only runs if ```boot/ostree``` exists which I think is a fair check into ostree specific code.

What I wondered about is if the version lookup makes sense/is useful. iirc the code walks through all deployment directories and the first one matching ```'.*/boot/ostree/.*/vmlinuz-(.*)'``` is going to be used. Is this accurate ? 
Can there be multiple different kernel versions in the deployments ?

in kiwi the kernel version lookup is not only used for bootloader code it's for example also used to construct the proper initrd file name when calling dracut and at other places. Thus fetching the information about the later running kernel of the OS is important for the build process of images that needs a kernel (typically disk images) 

I guess it's also because I don't know the inner works of ostree, but correct me when wrong, in the later running system there is still a kernel and its modules at the expected places ```/lib/modules/<kver>/...``` or the ```usr``` as I cannot imagine that ostree systems has moved everything to ```/boot/ostree/<deployment-id>/...```, correct ?

If so I also wondered why the kiwi built could not find a kernel at the usual place

kiwi/bootloader/config/grub2.py
```diff
@@ -1441,15 +1441,18 @@ def _fix_grub_loader_entries_boot_cmdline(self):
...
-                        r'options {0}'.format(self.cmdline),
+                        r'options \1 {0}'.format(self.cmdline),
                         menu_entry
                     )
                 with open(menu_entry_file, 'w') as grub_menu_entry_file:
                     grub_menu_entry_file.write(menu_entry)
 
     def _fix_grub_loader_entries_linux_and_initrd_paths(self):
+        # Temporarily nuked
+        return
```
schaefi, line 1455: All the _fix methods are called in ```bootloader/config/grub2.py``` 

```
# Patch the written grub config file to actually work:
# Unfortunately the grub tooling has several bugs and issues
# which prevents it to work in image build environments.
# More details can be found in the individual methods.
# One fine day the following fix methods can hopefully
# be deleted...
if self.xml_state.build_type.get_overlayroot_write_partition() is not False:
    self._fix_grub_root_device_reference(config_file, boot_options)
    self._fix_grub_loader_entries_boot_cmdline()
    self._fix_grub_loader_entries_linux_and_initrd_paths()
```

I think it would be good if the skipping of them per if ```Defaults.is_ostree()``` can be done at this place in the code

kiwi/bootloader/config/grub2.py
```diff
@@ -1441,15 +1441,18 @@ def _fix_grub_loader_entries_boot_cmdline(self):
             for menu_entry_file in glob.iglob(loader_entries_pattern):
                 with open(menu_entry_file) as grub_menu_entry_file:
                     menu_entry = grub_menu_entry_file.read()
+                    # Made append only
                     menu_entry = re.sub(
                         r'options (.*)',
-                        r'options {0}'.format(self.cmdline),
+                        r'options \1 {0}'.format(self.cmdline),
```
schaefi, line 1447: This seems to be a generic change not ostree specific and it changes the behavior as we intentionally wanted to replace the complete cmdline with this _fix method. I'm so happy when we can get rid of this _fix stuff

From an implementation point of view I think the different expression should be only applied in case of  a ```boot/ostree``` system. As it seems we have this differentiation at some places it might be a good idea to provide a method in the ```defaults.py``` that offers something like ```def is_ostree() -> bool:```

schaefi:
> This is definitely work in progress. I'm looking for suggestions on how we should implement this properly.

Thanks for looking into this :+1: much appreciated. I made some comments and maybe I understand the ostree better when we meet at oSC :)

travier:
> ... maybe I understand the ostree better when we meet at oSC :)

Unfortunately I'm struggling to get funding for travelling to DevConf.CZ so it's unlikely that I'll be able to go to oSC this year :(

kiwi/system/kernel.py
travier: > Can there be multiple different kernel versions in the deployments ?

ostree only supports having a single kernel per deployment.

> in the later running system there is still a kernel and its modules at the expected places

Yes, on the final system, things are mounted such that you have the impression of being on a classic system. You don't see the "real root" layout with `deploy`, etc.

> If so I also wondered why the kiwi built could not find a kernel at the usual place

Kiwi is looking at the "real root" layout, not at the final layout that is setup in the initramfs by ostree for the final system, so `/lib/modules` does not exists there, only in the deployment.

Overall, it might make sense to completely skip any kernel lookup in the ostree case.

kiwi/bootloader/config/grub2.py
travier: Sounds like a good idea. As an alternative, we could also check if the files exists as set in the BLS config and only fix them if they don't?

kiwi/bootloader/config/grub2.py
travier: An option might be to add a `kernelcmdlineappend` option that will only append to the command line instead of replacing it entirely. This would not change the behavior of the existing one and would let us use it for the ostree case.

travier:
This is still hackish but a bit less 😅

kiwi/bootloader/config/grub2.py
travier: Done

kiwi/bootloader/config/grub2.py
travier: Done

kiwi/builder/live.py
```diff
@@ -168,39 +169,56 @@ def create(self) -> Result:
...
+                        deployment_dir = os.sep.join([self.root_dir, 'boot/ostree', deployment])
+                        if os.path.isdir(deployment_dir):
+                            files = sorted(os.listdir(deployment_dir))
+                            for f in files:
+                                initramfs_file = os.sep.join([self.root_dir, 'boot/ostree', deployment, f])
+                                version_match = re.match(initramfs_ostree_pattern, initramfs_file)
+                                if version_match:
+                                    initrd_dest = os.sep.join([self.root_dir, 'boot/initramfs'])
+                                    Command.run(['cp', initramfs_file, initrd_dest])
+                                    self.boot_image.initrd_filename = initrd_dest
```
schaefi, line 221: If I understand it correctly we don't need to produce our own initrd in the ostree case but have to use the existing one in the ostree. You adapted this in the ```live``` builder but I think this would be effectively needed when other builders should support ostree. From that perspective do you think the above code could be moved into a method or maybe even better into its own class ?

A logging information when the initrd in the ostree was found would also be nice :)

Conan-Kudo:
@travier Couple of things here:

* Could you rebase this to fix the conflicts?
* What are you missing to turn this into something that we can review and merge?

travier:
> * Could you rebase this to fix the conflicts?

Will do.

> * What are you missing to turn this into something that we can review and merge?

I need help figuring out:
- How to fix the Live ISO build: https://github.com/OSInside/kiwi/issues/38#issuecomment-2144841679
- How to make a clean integration: https://github.com/OSInside/kiwi/issues/38#issuecomment-2179353264
	- This probably needs a general design discussion as I'm not familiar enough with kiwi's internals yet

Conan-Kudo:
@schaefi @travier Is there a time where we could have a dedicated meeting for this? We should try to get this hammered out asap as we want to start moving things for F42 in the spring.

travier:
Friday is usually a good time for me.

schaefi:
> @schaefi @travier Is there a time where we could have a dedicated meeting for this? We should try to get this hammered out asap as we want to start moving things for F42 in the spring.

This week and next week I won't be able to make it. I will be traveling to NUE and next week I'm in Portland for a workshop. The week after we can schedule a meeting and Friday is also a good day for me. That would mean October 18th

Conan-Kudo:
That works for me too.

schaefi:
@travier @Conan-Kudo Hi, unfortunately I still don't get my voice back and sound like a mouse. I think we need to move the meeting to another day

schaefi:
@travier in preparation to the conversation I was so free to rebase your code to the upstream main and fixed the conflict. Hope this is ok ?

kiwi/builder/live.py
```diff
@@ -172,41 +173,58 @@ def create(self) -> Result:
...
+                        Command.run(
+                            [
+                                'mv', self.boot_image.initrd_filename,
+                                self.root_dir + ''.join(
+                                    ['/boot/', boot_names.initrd_name]
+                                )
+                            ]
+                        )
+            else:  # pragma: nocover
+                # TODO: ostree: this code should be part of an OSTree class find the existing initrd in the ostree case
```
schaefi, line 213: Open for the Friday conversation

travier:
I'll close this one to clean things up for the project as I'm not working on it right now and as per the discussion, we should take a different approach.

PR #2558, opened 2024-05-31
Add --list-profiles to image info

schaefi:
Allow to list available profiles from the processed image description


schaefi:
This came in handy when working with the Fedora image description

PR #2561, opened 2024-06-03
Fix method ineffective

schaefi:
**Make sure BootLoaderConfig fixes are effective**
    
The BootLoaderConfigGrub2 class has methods to fix the grub-mkconfig generated files. It does that by mounting the system and changing the respective files after the mkconfig call. However, after the change  the class instance stays open in combination with BootLoaderInstallGrub2  instance which itself under certain circumstances also mounts the system to call grub-install. At the time grub-install is called it cannot be guaranteed that all changes has been written unless an explicit umount in the BootLoaderConfigGrub2 class instance happened. This commit address the potential race condition.


Conan-Kudo (approved):

Conan-Kudo:
Once the tests pass, I'll merge it.

PR #2562, opened 2024-06-03
Add missing write_meta_data method to BLS base

schaefi:
The standard bootloader interface class provided a method named write_meta_data which is expected to be implemented in the specialized bootloader implementation. For BLS bootloaders this method was missing in the BLS base class. write_meta_data can provide additional cmdline options for booting. If not covered some boot options might be missing. This patch fixes it

### DIFF
--- a/kiwi/bootloader/config/bootloader_spec_base.py
+++ b/kiwi/bootloader/config/bootloader_spec_base.py
@@ -89,6 +89,19 @@ class BootLoaderSpecBase(BootLoaderConfigBase):
         """
         pass
 
+    def write_meta_data(
+        self, root_device=None, write_device=None, boot_options=''
+    ):
+        """
+        For bootloaders following the bootloader spec take over
+        the cmdline options and store them for later use
+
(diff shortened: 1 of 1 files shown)


kiwi/bootloader/config/bootloader_spec_base.py
```diff
@@ -89,6 +89,19 @@ def setup_sysconfig_bootloader(self) -> None:
         """
         pass
 
+    def write_meta_data(
```
Conan-Kudo, line 92: What calls this function?

Conan-Kudo (changes requested):

kiwi/bootloader/config/bootloader_spec_base.py
schaefi: The builder. https://github.com/OSInside/kiwi/blob/main/kiwi/builder/disk.py#L1462

Conan-Kudo (approved):

PR #2567, opened 2024-06-10
Better error handling on grub vendor dir lookup

### LINKED ISSUE #2565: Missing `strings` command results in non-bootable ISOs
## Problem description

I am using a Fedora 40 image in Hetzner Cloud in a pipeline to build ISO live systems in an automated way.

Builds runs through smoothly.

Resulting ISOs didn't boot up correctly because of missing `(cd0)/EFI/ubuntu` (here for Ubuntu OS) path for GRUB2. Only the default path `(cd0)/EFI/BOOT` is available. Boot stops at GRUB2 rescue shell.

Root cause is missing `strings` command on the build host which extracts the path from GRUB2 binary:
https://github.com/OSInside/kiwi/blob/98608861e4b612b19f533ad4ac0a38cbb1e16f54/kiwi/bootloader/config/grub2.py#L576

Fedora image might be stripped down by Hetzner, however I would expect at least a warning if a command invocation fails in case of "command not found" errors (or any error) and/or a dependency to `strings` command - which is in the most cases part of `binutils`.

According to the logs, there is no issue with missing `strings` command at all:
```
...
[ INFO    ]: 07:09:23 | --> Using standard boot template
[ DEBUG   ]: 07:09:23 | EXEC: [bash -c strings /root/image/build/image-root/usr/lib/grub/x86_64-efi-signed/grubx64.efi.signed | grep EFI\/]
[ DEBUG   ]: 07:09:23 | EXEC: [mkdir -p /root/image/live-media.ypc4hwrk/EFI/ubuntu]
[ INFO    ]: 07:09:23 | Copying /root/image/live-media.ypc4hwrk/EFI/BOOT/grub.cfg -> /root/image/live-media.ypc4hwrk/EFI/ubuntu/grub.cfg to be found by EFI
[ INFO    ]: 07:09:23 | Writing KIWI template grub.cfg file
...
```

Installing `binutils` fixed the issue for me.

## Expected behaviour

Bootable ISO after successful built.

- absent `strings` command on build host
- build a live ISO (dmsquash)
- boot the ISO and end up in GRUB2 rescue shell
- check for existing paths with `grub> ls (cd0)/EFI/`

* KIWI version: 10.0.11
* Operating system host version: Fedora Linux 40 (Hetzner Cloud image)
* Operating system target version: Ubuntu 20.04

schaefi:
The strings command is used to lookup the in-efi binary encoded vendor path. However, if the strings or bash command is not availabe on the build host, the command silently failed and moved into the standard (non vendored) EFI boot path. This can lead to a broken boot for those distros and image targets which requires a vendor directory and should lead to an error message instead of a successful image build. This Fixes #2565

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -41,7 +41,8 @@ from kiwi.exceptions import (
     KiwiBootLoaderGrubModulesError,
     KiwiBootLoaderGrubSecureBootError,
     KiwiBootLoaderGrubFontError,
-    KiwiDiskGeometryError
+    KiwiDiskGeometryError,
+    KiwiCommandNotFound
 )
 
 import kiwi.defaults as defaults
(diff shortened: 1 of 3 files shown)


Conan-Kudo (commented):
What's the bash command you're referring to? Could we depend on that instead?

schaefi:
> What's the bash command you're referring to? Could we depend on that instead?

That is covered by the ```raise_on_command_not_found=True``` when calling ```strings``` as a script through bash.

We have no requirement on the packaging level for any shell interpreter but we also offer a bash completion for kiwi, so maybe it would be a good idea to add a requirement ?

Conan-Kudo:
I think we probably should just express a dependency on bash.

schaefi:
> I think we probably should just express a dependency on bash.

ok, I updated the spec template. Not sure if the names needs to be differentiated between distros ?

Conan-Kudo (approved):

PR #2571, opened 2024-06-12
test-image-live: add shadow package

landgraf:
Fixes:
KiwiCommandError: chroot: stderr: /usr/bin/chroot: failed to run command ‘usermod’: No such file or directory


schaefi:
Thanks, yes I also dislike that package resolution is different compared to when the buildservice builds this image or if you build it locally. I updated the package setup for some build-tests but not for all. Thanks for spotting this one :+1:

schaefi (approved):

PR #2581, opened 2024-07-03
Fixed spec file for TW build

schaefi:
On openSUSE Factory/TW the python packages are versioned and should be named according to the python interpreter

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -20,7 +20,7 @@
 # If they aren't provided by a system installed macro, define them
 %{!?_defaultdocdir: %global _defaultdocdir %{_datadir}/doc}
 
-%if 0%{?suse_version} && 0%{?suse_version} < 1600
+%if 0%{?suse_version} && 0%{?suse_version} >= 1500
 %global __python3 /usr/bin/python3.11
 %global python3_pkgversion 311
 %else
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
This should not be required. The default python flavor should always have a `python3-` Provides.

Conan-Kudo:
If you do this, then kiwi will continually break when python versions are swapped out.

schaefi:
> This should not be required. The default python flavor should always have a `python3-` Provides.

Yes you are right. So I added a Provides tag such that we always have python3-kiwi either as a direct package or through a provides tag. You may ask why. The reason is for suse many python packages uses the suse provided macros to make dealing with python and dependencies easier. There is one that does for example ````%{python_module Sphinx}``` and it resolves to ```python311-Sphinx``` or whatever python default version the provided distro has. Thus when we built kiwi for TW and also for Leap/SLES it should be able to resolve both, versioned and non-versioned names.

I have submitted this spec change to :Staging to see if it causes issues on the integration tests, but it seems to work

Thoughts ?

Conan-Kudo:
What problem are we trying to solve here? Because for Tumbleweed and Leap 16, we absolutely want to always use the default version, and for Leap 15, we need to override to use a non-default version of Python.

schaefi:
I tried to explain the problem. There is a package and it has this Requires

```
Requires:    %{python_module kiwi}
```

That doesn't work for TW because the above macro resolves to ```python311-kiwi``` which does not exist because at the moment our package is named ```python3-kiwi```.

I know that this change here will make the TW kiwi build to stick with py311 and will not automatically follow new default interpreters, but would that hurt ? they will not drop 311 from TW

I'm open for anything that solves this mess

Conan-Kudo:
So then the simple fix is to add `pythonXY-kiwi = %{version}-%{release}` provides rather than what this is. Let me come up with something, give me a tick.

schaefi:
> So then the simple fix is to add `pythonXY-kiwi = %{version}-%{release}` provides rather than what this is. Let me come up with something, give me a tick.

ok yes sounds better. Thanks for helping

PR #2583, opened 2024-07-04
package: Add fully qualified provides for python3-kiwi in spec

Conan-Kudo:
On SUSE distributions, currently the expectation is that packages built against the Python interpreter should have fully qualified names in the form of `pythonXY-<modulename>`. Additionally, all other Linux distributions prefer something similar in the form of `pythonX.Y-<modulename>`.

This ensures we have those names so that distribution dependency generation works as expected.


schaefi:
I'm so happy you are managing this :+1: big thanks as I usually produce issues when I touch a spec file :)

schaefi:
Once the tests pass I will create  a Staging build with this patch

schaefi (approved):

PR #2584, opened 2024-07-04
Split out bash completion into a sub-package

schaefi:
Per review of the SUSE packaging team we should split out the bash completion into its own sub-package to give users better control over the completion feature.

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -412,7 +412,6 @@ Requires:       screen
 Requires:       file
 Requires:       sed
 Requires:       bash
-Recommends:     bash-completion
 Requires:       python%{python3_pkgversion} >= 3.9
 %if 0%{?ubuntu} || 0%{?debian}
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Where did this commentary come from?

schaefi:
> Where did this commentary come from?

See ```osc rq show 1185374```. I was about to update kiwi in TW including the latest fixes we did and one comment of this submission is

```
( 1956032 ) On 2024-07-04 11:53:03 UTC dimstar wrote:
Inline comment for target: 'openSUSE:Factory/python-kiwi', file: 'python-kiwi.spec', and line: 43:

Recommending bash-completion is definitively not a good thing; -bash-completions are commonly split out and installed via Sipplements (%name and bash-completion)

giving users proper control over this feature.
```

According to this information I created this followup PR here

Conan-Kudo (approved):
I think this is overkill, but whatever...

PR #2590, opened 2024-07-10
Add new eficsm type attribute

### LINKED ISSUE #2407: RFE: Flag to tell kiwi to create UEFI-only images for x86_64
## Problem description
As part of the effort to support [UKIs in Fedora Cloud](https://fedoraproject.org/wiki/Changes/Unified_Kernel_Support_Phase_2), we need to be able to create images that are not hybrid UEFI+BIOS, but instead are UEFI-only. Currently, when setting an image firmware to `efi` or `uefi`, we produce hybrid images and image builds fail if we do not provide GRUB2 for BIOS boot. 

One way to solve this would be to add a `uefi_nocsm` variant to not produce hybrid boot images would resolve this issue. Another way would be to add a `csm=` flag that defaults to `on` but when set to `off` would not produce hybrid images.

* KIWI version: 9.25.16
* Operating system target version: Fedora Linux 40
* Koji version (N/A if not using Koji): 1.33.1

schaefi:
Allow to produce EFI/UEFI images without hybrid CSM capabilities. This Fixes #2407

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-disk-simple/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-disk-simple/appliance.kiwi
@@ -16,7 +16,7 @@
         <rpm-check-signatures>false</rpm-check-signatures>
         <bootsplash-theme>breeze</bootsplash-theme>
         <bootloader-theme>openSUSE</bootloader-theme>
-        <type image="oem" filesystem="ext3" kernelcmdline="console=ttyS0" firmware="uefi" format="vmdk">
+        <type image="oem" filesystem="ext3" kernelcmdline="console=ttyS0" firmware="uefi" eficsm="false" format="vmdk">
             <oemconfig>
                 <oem-resize>false</oem-resize>
             </oemconfig>
--- a/build-tests/x86/tumbleweed/test-image-live/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-live/appliance.kiwi
@@ -37,7 +37,7 @@
         </type>
     </preferences>
     <preferences profiles="Secure">
-        <type image="iso" flags="overlay" firmware="uefi" kernelcmdline="console=ttyS0" hybridpersistent_filesystem="ext4" hybridpersistent="true" mediacheck="true">
+        <type image="iso" flags="overlay" firmware="uefi" eficsm="false" kernelcmdline="console=ttyS0" hybridpersistent_filesystem="ext4" hybridpersistent="true" mediacheck="true">
             <bootloader name="grub2" console="serial" timeout="10"/>
         </type>
     </preferences>
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -384,6 +384,11 @@ bootpartsize="nonNegativeInteger":
   specifies the size in MB. If not set the boot partition
   size is set to 200 MB
 
+eficsm="true|false":
+  For images with an EFI layout, specify if the legacy
+  CSM (BIOS) mode should be supported or not. By default
+  CSM mode is enabled.
+
 efipartsize="nonNegativeInteger":
   For images with an EFI fat partition this attribute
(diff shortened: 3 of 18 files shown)


Conan-Kudo:
Why add new modes instead of another control flag?

schaefi:
> Why add new modes instead of another control flag?

I decided so because a new control attribute makes sense when it can be combined with a number of different settings. However, in this case a new attribute, let's name it ```csm="true|false"``` would only make sense in combination with ```firmware="efi|uefi"``` on x86 and would be meaningless for anything else. That's why I rather added the two new mode names

schaefi:
still testing

Conan-Kudo:
We already have arch-specific attributes, though. `efi*` attributes and `force_mbr` are already specific to some arches. I don't think having `efi_csm` would be that much worse than what we have now. It can be a no-op on non-x86 arches, which is fine.

Conan-Kudo:
And `efi_only|uefi_only` is going to be confusing for people and cause problems down the road when we have to rip it out once we can't support hybrid boot anymore.

schaefi:
> We already have arch-specific attributes, though. `efi*` attributes and `force_mbr` are already specific to some arches. I don't think having `efi_csm` would be that much worse than what we have now. It can be a no-op on non-x86 arches, which is fine.

I'm not referring to the arch specific bit, my point is that csm yes or no is only relevant for EFI. EFI is a firmware setting connected to a layout design and that's what the firmware specification indicates. Also see the updated documentation which should make it clear for every user.

schaefi:
> And `efi_only|uefi_only` is going to be confusing for people and cause problems down the road when we have to rip it out once we can't support hybrid boot anymore.

why is this confusing ? I can't follow why it should be problematic to rip it out ?

ripping out a supported value from an attribute is easy, a simple XSL stylesheet that changes the value to whatever we want
ripping out an attribute is in the same way easy, a simple XSL stylesheet that drops the attribute

The complexity of moving forward regarding the schema is the same

So we need to get clarity on what's confusing about "efi_only" ?

schaefi:
Given there is a csm attribute people could specify 

firmware="ofw" csm="true"

That's confusing and not even possible. And adding a rule set in the schema for useful combinations is difficult and an effort I'm not willing to implement because we could just come up with a none confusing firmware name for EFI without CSM

Conan-Kudo:
But don't we already have that problem with the other attributes?

schaefi:
> But don't we already have that problem with the other attributes?

Which attribute combination do you have in mind ?

Conan-Kudo:
`efipartsize` and `force_mbr` are both two attributes that make no sense with `firmware="ofw"`, but are still a valid combination that can be set.

Conan-Kudo:
> Given there is a csm attribute people could specify
> 
> firmware="ofw" csm="true"
> 
> That's confusing and not even possible. And adding a rule set in the schema for useful combinations is difficult and an effort I'm not willing to implement because we could just come up with a not confusing firmware name for EFI without CSM, for which I thought "efi_only" plus docs should do the trick

If we call it `efi_csm` then I think it falls under the same bucket as the other `efi*` options we have as attributes that obviously don't make sense for non-EFI systems but exist alongside them anyway.

schaefi:
yes it's true, there are more attributes that can be combined into non-sense combinations. Ok let's add another attribute but at least name it in a way that sort of tells the user it has something in common with EFI. I like ```eficsm=true|false```

Conan-Kudo:
Can we squash this down to two commits for eficsm + tests? We don't need the historical stuff in the final merge.

schaefi:
> Can we squash this down to two commits for eficsm + tests? We don't need the historical stuff in the final merge.

yes sure, I just need to complete my local build tests and once all is good I will squash into a good set

schaefi:
@Conan-Kudo ok done so far. All other tests will be done through the Staging builds

Conan-Kudo (approved):

PR #2591, opened 2024-07-11
Add --set-type-attr and --set-release-version

### LINKED ISSUE #2478: RFE: Add a way to override/set `release-version` from the CLI
## Problem description
We can set the `release-version` in the description, but if you want to have a one-off build with a different value (for example testing before branching or something), you cannot pass it via the commandline to test. Please add some kind of `--release-version`/`--releasever` option so this can be done.

* KIWI version: 9.25.21
* Operating system host version: Fedora 41
* Operating system target version: Fedora 40

### LINKED ISSUE #2588: RFE: Add a way to override/set ISO `volid` from the CLI
## Problem description
We can set the `volid` in the description for an ISO build, but if you want to have a one-off build with a different value (for example testing before branching or something) or have the value externally controlled (via something like Koji), you cannot pass it via the commandline to test. Please add some kind of `--iso-volid` option so this can be done.

* KIWI version: 10.0.21
* Operating system host version: Fedora 41
* Operating system target version: Fedora 40

schaefi:
Allow to set/overwrite type section attributes via the cmdline. Allow to set/add the release-version element via the cmdline.
This Fixes #2478 and Fixes #2588

### DIFF
--- a/doc/source/commands/system_build.rst
+++ b/doc/source/commands/system_build.rst
@@ -28,6 +28,8 @@ SYNOPSIS
        [--set-container-derived-from=<uri>]
        [--set-container-tag=<name>]
        [--add-container-label=<label>...]
+       [--set-type-attr=<attribute=value>...]
+       [--set-release-version=<version>]
        [--signing-key=<key-file>...]
    kiwi-ng system build help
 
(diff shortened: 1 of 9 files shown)


schaefi:
This allows to run kiwi with e.g

```
kiwi-ng system build ... --set-type-attr volid=some --set-release-version 42
```

Conan-Kudo (approved):

PR #2592, opened 2024-07-11
Added integration test for SUSE agama installer

schaefi:
This integration test builds a self-install ISO image which drops the SUSE Agama installer into a ramdisk for performing an interactive installation procedure to test Agama


schaefi:
This is to a large degree a copy of the agama image that we have in the systemsmanagement project and I wanted to have an image build here in kiwi because Agama makes use of several kiwi features which I'd like to be able to test and check in its own integration test. The image here builds at:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-agama

and the test can be extended to other architectures as we go forward.

For testing the installation the following procedure on x86_64 can be used

```
qemu-img create mydisk 20g
qemu-kvm -cpu Broadwell-v2 -m 4096 -netdev user,id=user0 -device virtio-net-pci,netdev=user0 -cdrom test-image-agama.x86_64-1.0.1.install.iso -hda mydisk
```

The image build also produces image files in an ```install.tar``` tarball which can be used to setup a remote/PXE-like installation procedure which should work on any architecture that supports remote loading of an initrd+kernel

PR #2593, opened 2024-07-12
Allow to set custom ISO Application ID

### LINKED ISSUE #1810: Enable custom Application IDs in ISOs
## Problem description

Currently in live images KIWI allows to customize `Volume ID` and `Publisher` fields of a generated ISO, however ISO supports a wider range of metadata fields. The problem seen in [bsc#1184581](https://bugzilla.suse.com/show_bug.cgi?id=1184581) is that the Volume ID is not sufficient to express all the data the developer wanted, mostly because it is restricted to 32 characters only.

As an alternative other fields that are not that constrained could be used, in particular Application ID (restricted to 128 characters) seams to be a good match for the purpose. I believe it would be good that KIWI allows to set a custom Application ID.

Currently KIWI include the MBR ID into the Application ID, I could not really figure out why, probably for debugging purposes back at the time all this part was coded?

I did the quick hack of adding random string as the Application ID to the xorriso call and all worked fine, EFI or BIOS booting as CROM or as a disk. I did not test with cdrtools but the implementation seams to be absolutely parallel to the xorriso implementation regarding this metadata attribute.

## Expected behaviour

There is a way to set a custom Application ID in KIWI ISOs

schaefi:
Add new <type ... application_id="..."/> attribute to be set in the ISO header main block. The application ID was used as identifier in the legacy initrd code from former kiwi versions. Because of this there is still the compat layer which sets an App ID as MBR identifier string unless the new application_id overwrites it. This Fixes #1810

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -943,6 +943,11 @@ volid="string":
   For the ISO type only, specifies the volume ID (volume name or label)
   to be written into the master block. There is space for 32 characters.
 
+application_id="string":
+  For the ISO/(oem install ISO) type only, specifies the Application
+  ID to be written into the master block. There is space for
+  128 characters.
+
 vhdfixedtag="GUID_string":
   For the VHD disk format, specifies the GUID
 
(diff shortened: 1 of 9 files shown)


schaefi:
To set the application ID on the command the new ```--set-type-attr``` option can be used as follows

```
kiwi-ng ... --set-type-attr application_id=some
```

PR #2594, opened 2024-07-12
Fixed logging behavior of Compress::get_format

schaefi:
The get_format() method allows to check which compression format a given input stream has. This is done by calling the supported compression tools in a row and let them check if they can deal with the provided data or not. As a result error messages are logged for streams that some tool doesn't understand. However, those error messages are no errors and only the result of the checking. This information in the kiwi log file is confusing and several users already complained when they see information like:

```
EXEC: Failed with stderr: /usr/bin/xz: ...: File format not recognized
```

This commit changes how the compression tooling is called in a way that no exception is raised (which leads to the above error message) but the result returncode is used to decide on the success or error of the respective compression tooling.

### DIFF
--- a/test/unit/utils/compress_test.py
+++ b/test/unit/utils/compress_test.py
@@ -105,14 +105,14 @@ class TestCompress:
 
     @patch('kiwi.command.Command.run')
     def test_get_format_invalid_format(self, mock_run):
-        mock_run.side_effect = ValueError("nothing")
+        result = Mock()
+        result.returncode = 1
+        mock_run.return_value = result
         invalid = Compress("../data/gz_data.gz")
         invalid.supported_zipper = ["mock_zip"]
 
         with self._caplog.at_level(logging.DEBUG):
             assert invalid.get_format() is None
             mock_run.assert_called_once_with(
-                ['mock_zip', '-l', '../data/gz_data.gz']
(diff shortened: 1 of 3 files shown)


test/unit/utils/compress_test.py
```diff
@@ -105,14 +105,16 @@ def test_get_format(self, mock_which):
...
         invalid = Compress("../data/gz_data.gz")
         invalid.supported_zipper = ["mock_zip"]
 
         with self._caplog.at_level(logging.DEBUG):
             assert invalid.get_format() is None
             mock_run.assert_called_once_with(
-                ['mock_zip', '-l', '../data/gz_data.gz']
+                ['mock_zip', '-l', '../data/gz_data.gz'], raise_on_error=False
             )
             assert 'Error running "mock_zip -l ../data/gz_data.gz", got a'
```
Vogtinator, line 119: This should be gone, right?

schaefi: Uh yes and that also uncovers a mistake in the test.  The line break in the assert statement caused it to be always correct.

schaefi: fixed

Vogtinator (approved):

Conan-Kudo (approved):

PR #2595, opened 2024-07-15
Support file includes

### LINKED ISSUE #1953: Allow to copy files to containers from image description without archive
## Problem description
It is currently difficult to do the equivalent from Dockerfile
`COPY this_file / `
directly in kiwi image description.

This requires usage of "archive" or config.sh
## Expected behaviour
Allowing to specific directly files in the kiwi description to be copied to a specific location on the image.

schaefi:
Add ```<file>``` directive to incorporate custom files
    
Usually custom files are managed by placing them as overlay files or archives. However, overlay files must be structured inside of a root/ subdirectory and archive files are binary data. It is therefore not straight forward to just reference one or more files as source files to the image description  to be placed into the image. This commit adds a new ```<file>```  element which allows to do this.

This Fixes #1953

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-disk-simple/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-disk-simple/appliance.kiwi
@@ -25,6 +25,7 @@
     </preferences>
     <users>
         <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root" groups="root"/>
+        <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/home/bob" name="bob" groups="users"/>
     </users>
     <repository type="rpm-md">
         <source path="obsrepositories:/"/>
@@ -56,6 +57,7 @@
         <package name="kernel-default"/>
         <package name="shim"/>
         <package name="timezone"/>
+        <file name="a_custom_file" target="/home/bob/some_file" owner="bob:users" permissions="444"/>
     </packages>
     <packages type="bootstrap">
         <package name="gawk"/>
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -1536,6 +1536,37 @@ elements:
 * The `enable` attribute is mandatory because it should be an explicit
   setting if a module is effectively used or not.
 
+<packages><file>
+~~~~~~~~~~~~~~~~~~~
+.. code:: xml
+
+   <packages type="image"/>
+     <file name="name"/>
+   </packages>
+
+The file element takes the `name` attribute and looks up the
+given name as file on the system. If specified relative {kiwi}
+looks up the name in the image description directory. The file
+is installed using the `rsync` program. The file element has the
+following optional attributes:
+
+owner="user:group"
+  The `owner` attribute can be specified to make the file
+  belonging to the specified owner and group. The ownership of
+  the original file is meaningless in this case. The provided
+  value is passed along to the `chown` program.
+
+permissions="perms"
+  The `permissions` attribute can be specified to store the file
+  with the provided permission. The permission bits of the original
+  file are meaningless in this case. The provided value is passed
+  along to the `chmod` program.
+
+target="some/path"
+  The `target` attribute can be used to specify a target path to
+  install the file to the specified directory and name. Eventually
+  missing parent directories will be created.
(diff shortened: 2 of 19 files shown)


build-tests/x86/tumbleweed/test-image-disk-simple/appliance.kiwi
```diff
@@ -56,6 +56,7 @@
         <package name="kernel-default"/>
         <package name="shim"/>
         <package name="timezone"/>
+        <file name="a_custom_file" target="/etc"/>
```
Conan-Kudo, line 59: This is not sufficient. You need to be able to set the destination filename.

Conan-Kudo (changes requested):

build-tests/x86/tumbleweed/test-image-disk-simple/appliance.kiwi
schaefi: you can. See the implementation, it uses rsync which supports all that, also the parent paths would be created. So all of the following would work:

```
<file name="a_custom_file" target="/etc"/>
<file name="a_custom_file" target="/etc/new_name"/>
<file name="a_custom_file" target="/some/new/path/foo"/>
<file name="a_custom_file" target="/some/new/path/"/>
```

If you want to place the file without a name change into a new new directory your path statement has to end with a "/" such that rsync knows this

From my perspective we are good here

Conan-Kudo: Hmm, what about mode and ownership?

schaefi: The sync happens with the default settings that are applied to other sync operations, e.g the overlay tree. These default sync options are:

```
return [
            '--archive', '--hard-links', '--xattrs', '--acls',
            '--one-file-system', '--inplace'
        ]
```

selinux labels are covered by the labeling process and are not related by the time of the syncing of files into the root tree

Conan-Kudo: Right, but that doesn't work for the nature of why this PR exists: the OBS VCS doesn't store that data, so we can assume it's always wrong.

schaefi: Sorry I can't follow you. If you place a file as obs source file it's permissions, acls, xattrs should be taken as the file is stored and not changed. To be really honest if obs is not able to store the file with its permissions it's not a kiwi job to fix this. In other words if the file has the wrong permissions after it was imported by the kiwi process and it's because the obs storage backend doesn't care for the origin permissions (which I hardly can believe) then a user needs to fix the permissions via script in config.sh.

In addition we can add another attribute for ```<file>``` that allows to provide a perm and mode mask which we apply after copy. But that I would do in a follow up pull request, not here

schaefi: Uhm it seems really true, file permissions of source files are not preserved when stored in obs.

@Conan-Kudo I added ```owner``` and ```permissions``` attribute to the ```<file>``` element

doc/source/image_description/elements.rst
```diff
@@ -1536,6 +1536,37 @@ elements:
...
+
+permissions="perms"
+  The `permissions` attribute can be specified to store the file
+  with the provided permission. The permission bits of the original
+  file are meaningless in this case. The provided value is passed
+  along to the `chmod` program.
+
+target="some/path"
+  The `target` attribute can be used to specify a target path to
+  unpack the file to the specified directory and name. Eventually
```
Conan-Kudo, line 1567: `unpack` -> `install`

Conan-Kudo (changes requested):

Conan-Kudo:
Please add a test for owner and mode setting.

Conan-Kudo:
Things look good now, can you squash the doc fixes into the main commit, so then I can approve it?

schaefi:
I found an issue... need to rework the code a bit. Problem is when we set the owner the user has to exist at that time already, which is currently not the case as the users gets created later...

doc/source/image_description/elements.rst

schaefi:
@Conan-Kudo ok ready for final review now

Conan-Kudo (approved):

PR #2601, opened 2024-07-30
Fix dracut-interactive with systemd 256

Vogtinator:
With systemd 256, /usr (and thus also /bin/) is read-only in the initrd. Move dracut-interactive into /run instead.

~Draft because not tested yet.~


schaefi (approved):
looks good to me :+1:

PR #2604, opened 2024-07-31
Fix kiwi-repart restrictions

schaefi:
The kiwi repart dracut module reads a profile file and if it does not exists it dies in the initrd. However, that profile file is not mandatory for the main resize functionality. Thus this commit turns this into a warning message. In addition a post script for the dracut-kiwi-oem-repart sub-package was added which creates a 40-kiwi-resize.conf to allow re-use of the kiwi resize module when it was requested.
This Fixes bsc#1228118


package/python-kiwi-spec-template
```diff
@@ -746,6 +746,16 @@ fi
 %endif
 %endif
 
+%post -n dracut-kiwi-oem-repart
+#============================================================
+# create /etc/dracut.conf.d/40-kiwi-resize.conf
+mkdir -p /etc/dracut.conf.d
+echo 'add_dracutmodules+=" kiwi-repart "' > /etc/dracut.conf.d/40-kiwi-resize.conf
+echo 'install_items+=" /config.partids "' >>/etc/dracut.conf.d/40-kiwi-resize.conf
+if [ -e "/.profile" ];then
```
davidcassany, line 755: I understand this is only likely to be there if someone installs this package as part of a KIWI build. So this looks safe good to me.

package/python-kiwi-spec-template
```diff
@@ -746,6 +746,16 @@ fi
 %endif
 %endif
 
+%post -n dracut-kiwi-oem-repart
+#============================================================
+# create /etc/dracut.conf.d/40-kiwi-resize.conf
+mkdir -p /etc/dracut.conf.d
+echo 'add_dracutmodules+=" kiwi-repart "' > /etc/dracut.conf.d/40-kiwi-resize.conf
```
davidcassany, line 753: I wonder what happens if someone installs this RPM in a host, won't this result into a generating a new initrd including the resize module? Probably even worse because the module would be only added to initrd if it is regenarated which does not happen automatically by installing this package.

The fact that installing this RPM in a regular host day A adding a dracut setup that on day B (e.g. after a kernel update) will cause resize module to be included is a bit scary to my eyes. In most cases this should not really apply any change, but on certain cases I fear it could silently re-partition a host...

IIRC the reason we run `check() { return 255 }` in our modules is to prevent auto inclusion of such a module in initrd. So they are only included into the initrd if it is explicitly required.

davidcassany (commented):
I wonder if always adding
```
add_dracutmodules+=" kiwi-repart "
```
to dracut confg is a good move... IMHO this should only be applied when build images, which the main use case of this package, but nothing prevents the user of installing  this KIWI package in the builder host (imagine someone just installing all `*kiwi*` packages)

package/python-kiwi-spec-template
schaefi: @davidcassany I understand, how about changing the %post to add this dracut conf if it can find `/config.partids` in the system ? This clearly states it's a kiwi image that got deployed and it has the resize activated in the original image and has also successfully booted into the system with that resize code active

schaefi: I pushed a commit to that idea

package/python-kiwi-spec-template
```diff
@@ -746,6 +746,18 @@ fi
 %endif
 %endif
 
+%post -n dracut-kiwi-oem-repart
+#============================================================
+# create /etc/dracut.conf.d/40-kiwi-resize.conf if applicable
+if [ -e "/config.partids" ];then
```
davidcassany, line 752: Yes, this looks safe to me :+1:

davidcassany (approved):
LGTM

package/python-kiwi-spec-template
schaefi: Thanks

PR #2608, opened 2024-08-06
Remove Path.remove, Path.rename and refactor Path.which & Path.create

dcermak:
* Drop `Path.remove` & `Path.rename`: each method is only used in one place each and they were subprocessing into shell commands. Let's use the pathlib builtins instead (more pythonic & efficient)
* Refactor `Path.create` to not subprocess into `mkdir`
* Refactor `Path.which` to use the shutil builtin `which`

### DIFF
--- a/kiwi/path.py
+++ b/kiwi/path.py
@@ -122,29 +122,6 @@ class Path:
             )
 
     @staticmethod
-    def remove(path: str) -> None:
-        """
-        Delete empty path, causes an error if target is not empty
-
-        :param string path: path name
-        """
-        Command.run(
-            ['rmdir', path]
-        )
-
-    @staticmethod
-    def rename(cur: str, new: str) -> None:
-        """
-        Move path from cur name to new name
-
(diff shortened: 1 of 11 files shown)


schaefi:
@dcermak I like it and makese sense :+1: 

Just one point to mention. Any `Command` call was also logged into the kiwi log file. it was very handy for debugging to see which directories where created, renamed, searched etc... With your changes I think we loose some of the logging information. Could you add a `log.debug` message for those parts in the code that replaced the `Command`  calls ?

After that I'm happy to merge it :) 

Thanks much

dcermak:
> Just one point to mention. Any `Command` call was also logged into the kiwi log file. it was very handy for debugging to see which directories where created, renamed, searched etc... With your changes I think we loose some of the logging information. Could you add a `log.debug` message for those parts in the code that replaced the `Command` calls ?

I've added logging calls to the code

schaefi (approved):
Thanks looks good to me :+1:

PR #2612, opened 2024-08-09
Fix handling of zipl.conf in plain zipl bootloader

### LINKED ISSUE #2597: invalid /etc/zipl.conf in Fedora images
## Problem description
An invalid `/etc/zipl.conf` is created for Fedora cloud images. When processing `zipl` complains about `keyword 'targetbase' required`.
````
[fedora@localhost ssh]$ cat /etc/zipl.conf
[defaultboot]
defaultauto
prompt=1
target=/boot

targettype=SCSI
targetblocksize=512
targetoffset=2048

timeout=1
secure=auto
default=(6.10.0-64.fc41.s390x) 
````
````
[fedora@localhost ssh]$ sudo zipl -V
Looking for components in '/lib/s390-tools'
Using config file '/etc/zipl.conf'
Using BLS config file '/boot/loader/entries/12bf46c0e4b74221a8084cacc28780cc-6.10.0-64.fc41.s390x.conf'
Error: Config file '/etc/zipl.conf': Line 6: keyword 'targetbase' required when specifying 'targettype'
````

## Expected behaviour
bootloader reinstalled

Just rerun `zipl` after booting the image.

* KIWI version: 10.0.24
* Operating system target version: Fedora rawhide
* Koji version (N/A if not using Koji): 1.34.1 (Fedora production instance)

for full details please see https://koji.fedoraproject.org/koji/taskinfo?taskID=120569240

schaefi:
When using the plain zipl bootloader kiwi created a /etc/zipl.conf file. However, this file was only useful during image build as it points to a loop target device and geometry but does not represent a proper config file to be used in the running system. In addition the different distributors provides their own version and layout of the zipl.conf to be used inside of the system and with their respective tools. Thus this commit changes the way how kiwi operates in a way that the zipl.conf used in the initial image only exists during the image build process. An eventual present /etc/zipl.conf will not be touched by kiwi. This Fixes #2597

### DIFF
--- a/kiwi/bootloader/config/zipl.py
+++ b/kiwi/bootloader/config/zipl.py
@@ -25,6 +25,7 @@ from kiwi.boot.image.base import BootImageBase
 from kiwi.bootloader.template.zipl import BootLoaderTemplateZipl
 from kiwi.bootloader.config.bootloader_spec_base import BootLoaderSpecBase
 from kiwi.command import Command
+from kiwi.utils.temporary import Temporary
 
 from kiwi.exceptions import (
     KiwiTemplateError,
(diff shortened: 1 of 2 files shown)


schaefi:
This PR must be used in Staging to check that no integration test broke. No merge before we have a green check mark here.

Conan-Kudo (approved):

Conan-Kudo:
`shellcheck Expected — Waiting for status to be reported` seems to be stuck, so I'm force merging it.

PR #2615, opened 2024-08-19
Add support for isomd5sum for tagging iso files

Conan-Kudo:
The `isomd5sum` tool suite is used and available on all supported distributions except SUSE distributions, and is necessary to produce conformant ISOs for most Linux distributions.

This change adds support for `isomd5sum` tool suite for kiwi, though it does not extend the kiwi-live dracut module to use it. The upstream dracut `dmsquash-live` module must be used instead.

### DIFF
--- a/kiwi/runtime_config.py
+++ b/kiwi/runtime_config.py
@@ -17,6 +17,7 @@
 #
 import os
 import logging
+from typing import Literal
 import yaml
 
 # project
@@ -308,6 +309,37 @@ class RuntimeConfig:
                 )
             )
             return Defaults.get_iso_tool_category()
+
+    def get_iso_media_tag_tool(self) -> Literal['checkmedia', 'isomd5sum']:
+        """
+        Return media tag tool used to checksum iso images
+
+        iso:
+          - media_tag_tool: checkmedia
+
+        if no or invalid configuration exists the default media tagger
+        from the Defaults class is returned
+
+        :return: A name
+
+        :rtype: str
+        """
+        iso_media_tag_tool = self._get_attribute(
+            element='iso', attribute='media_tag_tool'
+        )
+        if not iso_media_tag_tool:
+            return Defaults.get_iso_media_tag_tool()
+        elif 'checkmedia' in iso_media_tag_tool:
+            return iso_media_tag_tool
+        elif 'isomd5sum' in iso_media_tag_tool:
+            return iso_media_tag_tool
+        else:
+            log.warning(
+                'Skipping invalid iso media tag tool: {0}'.format(
+                    iso_media_tag_tool
+                )
+            )
+            return Defaults.get_iso_media_tag_tool()
 
     def get_oci_archive_tool(self):
         """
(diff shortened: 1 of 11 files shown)


kiwi/runtime_config.py
```diff
@@ -309,6 +309,25 @@ def get_iso_tool_category(self):
...
+        from the Defaults class is returned
+
+        :return: A name
+
+        :rtype: str
+        """
+        iso_media_tag_tool = self._get_attribute(
+            element='iso', attribute='media_tag_tool'
+        )
+        return iso_media_tag_tool or Defaults.get_iso_media_tag_tool()
```
dcermak, line 329: I'd suggest that we fail if the result is not in `['checkmedia', 'isomd5sum']`

dcermak (changes requested):
I believe we have to change `RuntimeChecker.check_mediacheck_installed`, as I currently get this error:

```
[ ERROR   ]: 15:23:31 | KiwiRuntimeError: 

Required tool tagmedia not found in caller environment

The attribute 'mediacheck' is set to 'true' which requires
the above tool to be installed on the build system

```

kiwi/runtime_config.py
Conan-Kudo: I tried to do this earlier, but I couldn't get the unit test coverage to be complete for it, so I dropped the logic since we don't do this for several other ones either...

Conan-Kudo: We don't fail, we select the default if it's not a valid choice. But I've put the logic back.

kiwi/runtime_config.py
```diff
@@ -309,6 +309,25 @@ def get_iso_tool_category(self):
             )
             return Defaults.get_iso_tool_category()
 
+    def get_iso_media_tag_tool(self):
```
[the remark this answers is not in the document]
Conan-Kudo: I tried to use this and the tests choked on it.

Conan-Kudo:
Tests pass again, but now coverage fails... 😦 

```
/home/runner/work/kiwi/kiwi/kiwi/runtime_checker.py                                333      2    99%   924, 926
/home/runner/work/kiwi/kiwi/kiwi/runtime_config.py                                 110      3    97%   332, 336-341
```

Conan-Kudo:
@schaefi, @davidcassany: either of you able to help me figure out how to close this coverage gap so this becomes green?

dcermak:
> @schaefi, @davidcassany: either of you able to help me figure out how to close this coverage gap so this becomes green?

The mocks were not setup correctly, I have force pushed a fix for `runtime_checker_test`

dcermak (commented):
I think this should work now, at least a test build doesn't die immediately.

Does the documentation need updating somewhere?

Conan-Kudo:
> I think this should work now, at least a test build doesn't die immediately.
> 
> Does the documentation need updating somewhere?

Nope. The documentation for the kiwi.yml file is in the kiwi.yml file itself.

dcermak (approved):

PR #2618, opened 2024-08-23
kiwi/schema: Fix allowed value type for ISO publisher and application ID

### LINKED ISSUE #2617: ISO application ID disallows spaces
## Problem description
For ISO9660, the Application ID has the same constraints as Publisher, and thus allows spaces. However, our schema is too strict and disallows this. This caused Fedora's image builds to fail as we unexpectedly failed validation.

## Expected behavior
Application ID "Fedora Linux" should be allowed.

Build `Fedora-KDE-Desktop-Live` from [fedora-kiwi-descriptions](https://pagure.io/fedora-kiwi-descriptions) at 70edf9c30104a3dfb709fdefd7adf3285753a90b.

* KIWI version: 10.0.24-1.fc40
* Operating system target version: Fedora Linux 42
* Koji version (N/A if not using Koji): 1.35

Conan-Kudo:
According to the spec, this should be constrained to 128 characters but also allow whitespaces. We didn't allow spaces in application ID, but allowed too much for Publisher.

Now we set up both correctly.

Fixes #2617.

### DIFF
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -26,7 +26,7 @@ namespace nul = ""
 
 safe-posix-name = xsd:token {pattern = "[a-zA-Z0-9_\-\.]+"}
 safe-posix-short-name = xsd:token {pattern = "[a-zA-Z0-9_\-\.]{1,32}"}
-safe-posix-long-name = xsd:token {pattern = "[a-zA-Z0-9_\-\.]{1,128}"}
+ecma-119-achar-128-text = xsd:token {pattern = "[a-zA-Z0-9_\-\. ]{1,128}"}
 locale-name = xsd:token {pattern = "(POSIX|[a-z]{2,3}_[A-Z]{2})(,[a-z]{2,3}_[A-Z]{2})*"}
 mac-address-type = xsd:token {pattern = "([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2}"}
 size-type = xsd:token {pattern = "(\d*|image)"}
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
@schaefi, @davidcassany: I'm pretty sure this necessitates a schema version bump, but I'm not sure how we handle implementing those.

AdamWill:
note this still isn't strictly correct. Per [the spec](https://ecma-international.org/wp-content/uploads/ECMA-119_4th_edition_june_2019.pdf), several other characters are allowed - `! " % & ' ( ) * + , - . / : ; < = > ?`. The more restrictive character set (just upper-case letters, numbers, and the underscore) is referred to in the spec as "d-characters", and used for e.g. the volume ID. The less restrictive set is referred to as "a-characters", and used for the application ID and publisher (and also the data preparer, the system identifier, and the "standard identifier", though that is always the same string anyway).

Conan-Kudo:
Well, this isn't merged yet, so I can certainly add those in too. 😄

schaefi:
> @schaefi, @davidcassany: I'm pretty sure this necessitates a schema version bump, but I'm not sure how we handle implementing those.

Well so far we did not bump the schema version for this cases. You changed the strictness of a pattern applied to an attribute. Strictly speaking the schema did not change on any attribute or element or semantic structure. From that perspective no XSLT conversion is needed from `old` to `new` and thus we usually don't bump the schema version for such a change.

The version of kiwi itself will be bumped on a new tag that includes your change. Thus the increase of the pattern strictness can still be detected by a version compare of kiwi itself.

From my perspective this change can go in as soon as @AdamWill is fine with the final pattern

Conan-Kudo:
The only character not added was the `"` character as I wasn't sure how to get that through with the file format (ie. I didn't know how to escape it properly).

Conan-Kudo:
And now all the tests fail anyway because the regex is somehow invalid 😩

Conan-Kudo:
Now everything should be good now! 😄

schaefi (approved):
Looks good to me, thanks much :+1:

PR #2623, opened 2024-08-24
kiwi/builder/live: Cleanup leftover dracut config file and log correct application id

Conan-Kudo:
This pull request contains two small fixes for building live ISOs:
* A commit to clean up a leftover dracut config file that caused Fedora live media to become non-installable
* A commit to correctly log the application id whenever a custom one is set


Conan-Kudo:
FYI: @tdawson

Conan-Kudo:
I've verified with Fedora KDE image builds that this fixes live images so that they are properly installable.

schaefi (approved):
Looks good to me :+1:

PR #2625, opened 2024-08-27
remove dependency on /usr/bin/python

pgajdos:
Fixes bsc#1212476.

* shebangs include /usr/bin/python3.11 instead of /usr/bin/python3, thus 
```
:/ # rpm -qR python3-kiwi | grep bin
/usr/bin/python3.11
:/ #
```

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -729,6 +729,11 @@ done
 %fdupes %{buildroot}/srv/tftpboot
 %endif
 
+%if 0%{?suse_version}
+# bsc#1212476
+%python3_fix_shebang
+%endif
+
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Why aren't you doing this as a buildroot policy?

pgajdos:
Not sure I understand, sorry.
To give you more context: https://trello.com/c/RE0ZBjLe/162-remove-dependency-on-usr-bin-python3-from-factory-packages

Conan-Kudo:
In this case, we call our package `python3-kiwi` and therefore it makes sense to link to `/usr/bin/python3` shebang.

Conan-Kudo (approved):
Meh, even if this should be a brp instead of a script, I don't care enough to quibble over it.

PR #2626, opened 2024-08-27
Mount external /etc and /var partitions

### LINKED ISSUE #2619: /var is was not mounted when disk.sh was called
## Problem description

A MicroOS variant with a separated /var partition was created via:

```
            <systemdisk>
                <volume name="home"/>
                <volume name="root"/>
                <volume name="opt"/>
                <volume name="srv"/>
                <volume name="usr/local"/>
            </systemdisk>
	    <partitions>
	      <partition name="var" size="1G" mountpoint="/var" filesystem="ext4"/>
	    </partitions>
```

Note that `<volume name="var" copy_on_write="false"/>` was dropped from `<systemdisk>` and the partition was added instead.

This image is using systemd-boot as a boot loader.  This is relevant because we will need to call in the `disk.sh` script the tool `sdbootutil` to install the bootloader and copy the kernels from the rootfs to the ESP.

```
    echo "add kernels"
    export hostonly_l=no # for dracut
    sdbootutil -v --arch "$arch" --esp-path /boot/efi --entry-token=auto add-all-kernels
```

This call will try to generate an `initrd` for the image calling `dracut`, and for that needs `/etc` and `/var` mounted, as the `/etc` overlays are in `/var`.  The commands will fails because `/var` was not mounted:

```
mount: /.snapshots/1/snapshot/etc: special device overlay does not exist.
```

Printing the mounts and the `/etc/fstab` from `disk.sh` we can see that this is the case:

```
# mount
[  207s] [ DEBUG   ]: 10:56:55 | /dev/loop0p4 on / type btrfs (rw,relatime,discard=async,space_cache=v2,subvolid=258,subvol=/@/.snapshots/1/snapshot)
[  207s] [ DEBUG   ]: 10:56:55 | /dev/loop0p2 on /boot/efi type vfat (rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=iso8859-1,shortname=mixed,errors=remount-ro)
[  207s] [ DEBUG   ]: 10:56:55 | /dev/loop0p4 on /.snapshots type btrfs (rw,relatime,discard=async,space_cache=v2,subvolid=257,subvol=/@/.snapshots)
[  207s] [ DEBUG   ]: 10:56:55 | /dev/loop0p4 on /home type btrfs (rw,relatime,discard=async,space_cache=v2,subvolid=259,subvol=/@/home)
[  207s] [ DEBUG   ]: 10:56:55 | /dev/loop0p4 on /opt type btrfs (rw,relatime,discard=async,space_cache=v2,subvolid=260,subvol=/@/opt)
[  207s] [ DEBUG   ]: 10:56:55 | /dev/loop0p4 on /root type btrfs (rw,relatime,discard=async,space_cache=v2,subvolid=261,subvol=/@/root)
[  207s] [ DEBUG   ]: 10:56:55 | /dev/loop0p4 on /srv type btrfs (rw,relatime,discard=async,space_cache=v2,subvolid=262,subvol=/@/srv)
[  207s] [ DEBUG   ]: 10:56:55 | /dev/loop0p4 on /usr/local type btrfs (rw,relatime,discard=async,space_cache=v2,subvolid=263,subvol=/@/usr/local)
[  207s] [ DEBUG   ]: 10:56:55 | /dev/vda on /image type ext4 (rw,noatime)
[  207s] [ DEBUG   ]: 10:56:55 | tmpfs on /tmp type tmpfs (rw,relatime,inode64)
[  207s] [ DEBUG   ]: 10:56:55 | tmpfs on /var/tmp type tmpfs (rw,relatime,inode64)
[  207s] [ DEBUG   ]: 10:56:55 | proc on /proc type proc (rw,nosuid,nodev,noexec,relatime)
[  207s] [ DEBUG   ]: 10:56:55 | sysfs on /sys type sysfs (rw,nosuid,nodev,noexec,relatime)
[  207s] [ DEBUG   ]: 10:56:55 | devtmpfs on /dev type devtmpfs (rw,nosuid,size=4096k,nr_inodes=249301,mode=755,inode64)

# /etc/fstab
[  207s] [ DEBUG   ]: 10:56:55 | UUID=dd4b5ea9-7980-4f80-89ec-b331645e40c8 / btrfs ro 0 0
[  207s] [ DEBUG   ]: 10:56:55 | UUID=dd4b5ea9-7980-4f80-89ec-b331645e40c8 /.snapshots btrfs defaults,subvol=@/.snapshots 0 0
[  207s] [ DEBUG   ]: 10:56:55 | UUID=dd4b5ea9-7980-4f80-89ec-b331645e40c8 /home btrfs defaults,subvol=@/home 0 0
[  207s] [ DEBUG   ]: 10:56:55 | UUID=dd4b5ea9-7980-4f80-89ec-b331645e40c8 /opt btrfs defaults,subvol=@/opt 0 0
[  207s] [ DEBUG   ]: 10:56:55 | UUID=dd4b5ea9-7980-4f80-89ec-b331645e40c8 /root btrfs defaults,subvol=@/root,x-initrd.mount 0 0
[  207s] [ DEBUG   ]: 10:56:55 | UUID=dd4b5ea9-7980-4f80-89ec-b331645e40c8 /srv btrfs defaults,subvol=@/srv 0 0
[  207s] [ DEBUG   ]: 10:56:55 | UUID=9375a160-9fdb-4f44-9ff9-99f996e38c05 /var ext4 defaults,x-initrd.mount,x-growpart.grow,x-systemd.growfs 0 0
[  207s] [ DEBUG   ]: 10:56:55 | UUID=AE42-E9CC /boot/efi vfat defaults 0 0
[  207s] [ DEBUG   ]: 10:56:55 | UUID=dd4b5ea9-7980-4f80-89ec-b331645e40c8 /usr/local btrfs defaults,subvol=@/usr/local 0 0
[  207s] [ DEBUG   ]: 10:56:55 | overlay /etc overlay defaults,lowerdir=/sysroot/etc,upperdir=/sysroot/var/lib/overlay/1/etc,workdir=/sysroot/var/lib/overlay/work-etc,x-systemd.requires-mounts-for=/var,x-systemd.requires-mounts-for=/sysroot/var,x-initrd.mount 0 0

```

## Expected behaviour

I would expect `/var` mounted, or something that can temporally used as such (and later discarded)

aplanas:
If there are external /etc and /var partitions when the post disk preparation disk script would be called, mount them.

Fixes #2619

### DIFF
--- a/kiwi/system/mount.py
+++ b/kiwi/system/mount.py
@@ -70,7 +70,7 @@ class ImageSystem:
         Mount image system from current block layers
         """
         # mount root boot and efi devices as they are present
-        (root_device, boot_device, efi_device) = self._setup_device_names()
+        (root_device, boot_device, efi_device, other_devices) = self._setup_device_names()
         root_mount = MountManager(
             device=root_device
         )
@@ -92,6 +92,11 @@ class ImageSystem:
                     root_mount.mountpoint, 'boot', 'efi'
                 )
             )
+        other_mounts = [
+            MountManager(
+                device=d, mountpoint=os.path.join(
+                    root_mount.mountpoint, m)) for (m, d) in other_devices.items()
+        ]
 
         self.mount_list.append(root_mount)
         root_mount.mount()
@@ -103,6 +108,10 @@ class ImageSystem:
         if efi_device:
             self.mount_list.append(efi_mount)
             efi_mount.mount()
+
+        for mount in other_mounts:
+            self.mount_list.append(mount)
+            mount.mount()
 
         if self.volumes:
             self._mount_volumes(root_mount.mountpoint)
@@ -163,7 +172,12 @@ class ImageSystem:
             root_device = self.device_map['readonly'].get_device()
         if 'efi' in self.device_map:
             efi_device = self.device_map['efi'].get_device()
-        return (root_device, boot_device, efi_device)
+        other_devices = {
+            d: self.device_map[d].get_device() for d in self.device_map if d not in (
+                'root', 'origin_root', 'boot', 'readonly', 'efi', 'efi_csm', 'swap'
+            ) and self.device_map[d]
+        }
+        return (root_device, boot_device, efi_device, other_devices)
 
     def _mount_volumes(self, mountpoint) -> None:
         for volume_path in Path.sort_by_hierarchy(sorted(self.volumes.keys())):


Conan-Kudo:
Can we generalize this to just mount all defined partitions? There's no reason to be this specific.

aplanas:
@Conan-Kudo The second commit provides a more generic approach. Sadly `device_map` keys can contain some magic stuff (`efi_csm`, `origin_root`, `readonly`, etc) or not mountable devices (`swap`) that make this approach brittle.

Conan-Kudo:
I'm kind of surprised this didn't require extending the tests...

Conan-Kudo:
The approach is sound, I would prefer this be squashed into a single commit, though.

aplanas:
I can squash, I just separated to compare both approaches. The `not in ('root', 'origin_root', ...)` makes me uneasy, as there is no guarantee that tomorrow we found three more keys that needs to be excluded.

But if you find OK, I am OK too.

Squasing

Conan-Kudo:
I am uneasy about it too. I'm wondering if @davidcassany might have a better solution for this.

schaefi (changes requested):
Hmm, I'm not in favor of the `other_devices` parameter. This should be done differently. All partition information is stored as

```python
custom_partitions = xml_state.get_partitions()
```

This custom_partitions should be handed over to an instance of class `ImageSystem` This means the class constructor should be changed to

```python
def __init__(
    self, device_map: Dict, root_dir: str, volumes: Dict = {}, partitions: Dict[str, ptable_entry_type] = {}
) -> None:
    ...
    self.partitions = partitions
```

Now all information to partitions is available as `self.partitions` and you can use this information in the `mount()` method to do the mounting

Thanks

aplanas:
Closed if favor of https://github.com/OSInside/kiwi/pull/2629

PR #2631, opened 2024-09-02
Fixed resize of dos table type on s390

schaefi:
On s390, parted is used to detect the partition table type. In contrast to blkid the name for DOS tables is reported as 'msdos' and not 'dos' which impacts several conditions in the kiwi initrd code which checks for 'dos'. This commit fixes the get_partition_table_type() method to return a consistent table name for DOS tables. This Fixes bsc#1228729

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -292,11 +292,22 @@ function get_free_disk_bytes {
 
 function get_partition_table_type {
     local disk_device="$1"
+    local table_type
 
     # blkid doesn't work for dasd partitions, so on s390 parted is used
     # to detect the partition type (bsc#1209247)
     if [[ "$(uname -m)" =~ s390 ]];then
-        parted --script --machine "${disk_device}" print | grep "^${disk_device}" | cut -d":" -f6
+        table_type=$(
+            parted --script --machine "${disk_device}" print |\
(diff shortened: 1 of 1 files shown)


schaefi:
bug reference: https://bugzilla.suse.com/show_bug.cgi?id=1228729

davidcassany (approved):
Looks good to me.

I guess we can't really do it any better, as there is no set of tools that works and is consistent in all supported archs.

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -292,11 +292,22 @@ function get_free_disk_bytes {
...
     # to detect the partition type (bsc#1209247)
     if [[ "$(uname -m)" =~ s390 ]];then
-        parted --script --machine "${disk_device}" print | grep "^${disk_device}" | cut -d":" -f6
+        table_type=$(
+            parted --script --machine "${disk_device}" print |\
+                grep "^${disk_device}" | cut -d":" -f6
+        )
+        if [ "${table_type}" = "msdos" ];then
+            # make sure to set a consistent table name for the
+            # DOS table type. Some tools reports it as either
```
Vogtinator, line 306: > Some tools reports it as either

I'd hope that at least `parted` reports it as `msdos` consistently

schaefi: yes it does and I agree the comment could be misleading. I changed the comment text to be more clear

PR #2632, opened 2024-09-02
Documentation: Add note about guestOS

kastakhov:

- Update documentation with note about correct values for guestOS.


schaefi (approved):
Thanks, looks good to me :+1:

PR #2634, opened 2024-09-03
Fixed ramdisk size setup

schaefi:
For setting up the brd rd_size option kiwi creates 99-brd.conf used at load time of the kernel brd driver. The location for the conf file is set to /etc/modprobe.d/ However, in newer versions the location has changed to /usr/lib/modprobe.d/ and /etc/modprobe.d no longer exists. This commit allows to handle both systems

### DIFF
--- a/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
+++ b/dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
@@ -50,6 +50,7 @@ function get_disk_list {
         local rd_size
         local modfile=/etc/modprobe.d/99-brd.conf
         rd_size=$(getarg ramdisk_size=)
+        mkdir -p /etc/modprobe.d
         if [ -n "${rd_size}" ];then
             echo "options brd rd_size=${rd_size}" > ${modfile}
         fi


Conan-Kudo (changes requested):
We should *never* create files in `/usr/lib/modprobe.d`. If the directory in `/etc` doesn't exist and we need to put a file there, we need to create the directory.

davidcassany:
I agree with @Conan-Kudo , creating a file in `/usr/lib/modprobe.d` feels wrong, moreover after some time we (or someone else) are likely to be puzzled by this file not provided by any package and mysteriously added into a sensitive location.
If the path `/etc/modprobe.d` is is still honored I'd say that's the way to go and relay on a `mkdir -p /etc/modprobe.d` or similar approach to ensure the path exists.

schaefi:
ok, thanks I will do so. Let's hope the /etc location is still honored. Thanks

schaefi:
@Conan-Kudo @davidcassany ok, I updated the PR, can you take a look again ? Thanks

Conan-Kudo (approved):

PR #2638, opened 2024-09-04
CHRP boot

wfeldt:
## This adds CHRP boot support to get ppc64 ISO media bootable

What is needed for that is a file `/ppc/bootinfo.txt` with this sample content:
```
<chrp-boot>
<description>Some Name</description>
<os-name>Some Name</os-name>
<boot-script>boot &device;:1,\boot\grub2\powerpc-ieee1275\grub.elf</boot-script>
</chrp-boot>
```
Note the backslashes in the path to `grub.elf`. `grub.elf` is the default grub boot image as provided in the `grub2-powerpc-ieee1275` package.

Note also that the grub2 directory `powerpc-ieee1275` contains a '`-`' - this is not possible with the default setting which  merges file names into the ancient 8.3 format. To avoid this, `-compliance  untranslated_names` is added to xorriso.
I've added this only to ppc64 for now but IMHO this should be done for all architectures.

And, these are the CHRP xorriso boot options:
```
-boot_image any chrp_boot_part=on
```

Finally, fix  a bug that creates an empty `EFI/BOOT` directory on all architectures.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -16,6 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 from string import Template
+from textwrap import dedent
 import re
 import os
 import logging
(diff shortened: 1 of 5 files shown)


schaefi:
@wfeldt Great job :+1: thanks much for digging into this. If you want me to kick in regarding the tests let me know

glaubitz:
Adding support for sparc64 shouldn't be too different here as we use GRUB here as well.

See: https://sources.debian.org/src/debian-cd/3.2.1/tools/boot/bookworm/boot-sparc64/

Boot `cdboot.img` and `core.img` are part of the grub2 package on sparc64.

schaefi (approved):
ok looks good now

@glaubitz for sparc64 I suggest a new pull request

PR #2639, opened 2024-09-05
Fixed Debian bootstrap for older versions

schaefi:
For older targets e.g Ubuntu Jammy, the apt version does not ignore the absence of a non standard status file. Thus it needs to be created empty and removed later.

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -259,6 +259,7 @@ class PackageManagerApt(PackageManagerBase):
                     echo "Unpacking $deb"
                     dpkg-deb --fsys-tarfile $deb | tar -C {0} -x
                 done < {1}
+                export PATH=/usr/bin:/bin:/usr/sbin:/sbin
                 while read -r deb;do
                     pushd "$(dirname "$deb")" >/dev/null || exit 1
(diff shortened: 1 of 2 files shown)


schaefi:
This change is related to an issue found by @IsaacJT , please keep open until we can successfully build the target. Thanks

Conan-Kudo:
Do we still support these targets?

schaefi:
> Do we still support these targets?

I'm not sure, I think we decided to no longer support them. However, the change here is not so problematic. But if we need more to make it build I would also suggest to use an older kiwi still based on debootstrap to build it. I wanted to make this depending on the referenced kiwi-descriptions issue

Conan-Kudo:
Fine with me, but let's make sure we don't create a rabbit-hole here. I'm pretty sure our floor with kiwi 10 is CentOS/RHEL 9, Ubuntu 24.04, Debian 12, and SLE 15.6/16.

schaefi:
> Fine with me, but let's make sure we don't create a rabbit-hole here. I'm pretty sure our floor with kiwi 10 is CentOS/RHEL 9, Ubuntu 24.04, Debian 12, and SLE 15.6/16.

I agree, I'm closing this one

PR #2642, opened 2024-09-11
Add new build type provides for enclave

schaefi:
Add a provides tag (read by the open buildservice) for the new enclave builder. Also add a recommends to eif_builder in the systemdeps-core meta package

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -122,6 +122,10 @@ Obsoletes:      kiwi-image-tbz-requires < %{version}-%{release}
 %if "%{_vendor}" != "debbuild"
 Provides:       kiwi-image:tbz
 %endif
+%if 0%{?fedora}
+Provides:       kiwi-image:enclave
+Requires:       eif_build
+%endif
 # tools conditionally used by kiwi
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

package/python-kiwi-spec-template
```diff
@@ -121,12 +121,14 @@ Provides:       kiwi-image-tbz-requires = %{version}-%{release}
 Obsoletes:      kiwi-image-tbz-requires < %{version}-%{release}
 %if "%{_vendor}" != "debbuild"
 Provides:       kiwi-image:tbz
+Provides:       kiwi-image:enclave
```
Conan-Kudo, line 124: This should also be under the same conditional as what provides eif_build.

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: ok, done

schaefi: once we have eif_build for other distros it's easier to increase the scope. Yes that's better

PR #2643, opened 2024-09-11
Fix oci security sync filters

schaefi:
This patch is two fold

**Fixed container sync options**

Do not exclude/filter any security/xattr capabilities.

**Update container integration test**
    
Add getcap to check on filesystem capabilities


darix:
LGTM!

Thank you for restoring fscaps again. we moved a lot of setuid binaries to only use fscaps to limit the attack surface.

Conan-Kudo (approved):

PR #2645, opened 2024-09-13
Add support for erofs

### LINKED ISSUE #2633: EROFS output type support
## Problem description
Kiwi doesn't currently support creating [EROFS](https://docs.kernel.org/filesystems/erofs.html) images. This would be useful for the [FEX RootFS](https://pagure.io/fedora-kiwi-descriptions/pull-request/83) in Fedora, among other things, as EROFS is more efficient than squashfs.

## Expected behaviour
`<type image="erofs"/>` is supported

This is a feature request

* KIWI version: 10.1.2
* Operating system target version: Fedora Linux 42

schaefi:
erofs is an alternative readonly filesystem that can be used as alternative to squashfs. This Fixes #2633

Please find an integration test build here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-live

### DIFF
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -246,7 +246,7 @@ class LiveImageBuilder:
         filesystem_setup = FileSystemSetup(
             self.xml_state, self.root_dir
         )
-        if root_filesystem != 'squashfs':
+        if root_filesystem not in ['squashfs', 'erofs']:
             # Create a filesystem image of the specified type
             # and put it into a SquashFS container
             root_image = Temporary().new_file()
@@ -302,12 +302,15 @@ class LiveImageBuilder:
         else:
             # Put the root filesystem into SquashFS directly
             with FileSystem.new(
-                name='squashfs',
+                name=root_filesystem,
                 device_provider=DeviceProvider(),
                 root_dir=self.root_dir + os.sep,
                 custom_args={
                     'compression':
                         self.xml_state.build_type.get_squashfscompression()
+                } if root_filesystem == 'squashfs' else {
+                    'compression':
+                        self.xml_state.build_type.get_erofscompression()
                 }
             ) as live_container_image:
                 container_image = Temporary().new_file()
@@ -316,6 +319,12 @@ class LiveImageBuilder:
                 )
                 Path.create(self.media_dir.name + '/LiveOS')
                 os.chmod(container_image.name, 0o644)
+                # Note: we keep the filename of the read-only image as it is
+                # even if another read-only filesystem not matching this
+                # filename is used. This is because the following filename
+                # is also used in the initrd code for the kiwi-live and
+                # dmsquash dracut modules. The name can be overwritten
+                # with the rd.live.squashimg boot option though.
                 shutil.copy(
                     container_image.name,
                     self.media_dir.name + '/LiveOS/squashfs.img'
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -291,10 +291,17 @@ Provides:       kiwi-filesystem:ext3
 Provides:       kiwi-filesystem:ext4
 Provides:       kiwi-filesystem:squashfs
 Provides:       kiwi-filesystem:xfs
+%if ! (0%{?suse_version} && 0%{?suse_version} < 1600)
+Provides:       kiwi-filesystem:erofs
+Provides:       kiwi-image:erofs
+%endif
 %endif
 Requires:       dosfstools
 Requires:       e2fsprogs
 Requires:       xfsprogs
+%if ! (0%{?suse_version} && 0%{?suse_version} < 1600)
+Requires:       erofs-utils
+%endif
 %if 0%{?suse_version}
 Requires:       btrfsprogs
 %else
--- a/build-tests/x86/rawhide/test-image-erofs/appliance.kiwi
+++ b/build-tests/x86/rawhide/test-image-erofs/appliance.kiwi
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<image schemaversion="7.5" name="kiwi-test-image-erofs">
+    <description type="system">
+        <author>Marcus Schaefer</author>
+        <contact>marcus.schaefer@suse.com</contact>
+        <specification>Fedora Appliance, Testing erofs filesystem image</specification>
+    </description>
+    <preferences>
+        <version>2.0.0</version>
+        <packagemanager>dnf5</packagemanager>
+        <rpm-check-signatures>false</rpm-check-signatures>
+    </preferences>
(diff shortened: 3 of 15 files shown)


package/python-kiwi-spec-template
```diff
@@ -291,6 +291,7 @@ Provides:       kiwi-filesystem:ext3
 Provides:       kiwi-filesystem:ext4
 Provides:       kiwi-filesystem:squashfs
 Provides:       kiwi-filesystem:xfs
+Provides:       kiwi-filesystem:erofs
```
Conan-Kudo, line 294: We also need `Requires: erofs-utils`

Conan-Kudo (changes requested):
I'd also like a Rawhide test for this with both live image and filesystem image since these are going to be actively used.

package/python-kiwi-spec-template
schaefi: yes saw it a second before and added it. But probably not correct as I only added for TW, please advise

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
```diff
@@ -291,10 +291,14 @@ Provides:       kiwi-filesystem:ext3
...
 %endif
 Requires:       dosfstools
 Requires:       e2fsprogs
 Requires:       xfsprogs
+%if 0%{?suse_version} >= 1650
```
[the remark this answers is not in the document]

package/python-kiwi-spec-template
Conan-Kudo: I just did in the comment about the conditional. :)

The provides for erofs needs to be similarly wrapped because otherwise it'll be a problem.

schaefi:
> I'd also like a Rawhide test for this with both live image and filesystem image since these are going to be actively used.

What do you mean by "filesystem image" ? a simple filesystem image ? not bootable just for mounting ?

package/python-kiwi-spec-template
```diff
@@ -291,10 +291,14 @@ Provides:       kiwi-filesystem:ext3
 Provides:       kiwi-filesystem:ext4
 Provides:       kiwi-filesystem:squashfs
 Provides:       kiwi-filesystem:xfs
+Provides:       kiwi-filesystem:erofs
```
Conan-Kudo, line 294: This also needs `%if ! (0%{?suse_version} && 0%{?suse_version} < 1600)` otherwise OBS will happily let you run a build guaranteed to fail.

schaefi: yes I did the same mistake again, will put it into the same condition block

Conan-Kudo:
> > I'd also like a Rawhide test for this with both live image and filesystem image since these are going to be actively used.
> 
> What do you mean by "filesystem image" ? a simple filesystem image ? not bootable just for mounting ?

Yes, @davide125 is making a FEX rootfs image with kiwi and wants to use erofs for it (and there's general talk of making container images using erofs), so I'd like for us to make sure this works.

package/python-kiwi-spec-template

schaefi:
ok, I will add those integration tests once the one I have works

package/python-kiwi-spec-template
```diff
@@ -295,6 +295,10 @@ Provides:       kiwi-filesystem:xfs
 Requires:       dosfstools
 Requires:       e2fsprogs
 Requires:       xfsprogs
+%if ! (0%{?suse_version} && 0%{?suse_version} < 1600)
+Provides:       kiwi-filesystem:erofs
```
Conan-Kudo, line 299: This still needs to be with the other Provides lines in the "not debbuild" conditional, since these provides are invalid for debian.

kiwi/builder/live.py
```diff
@@ -302,20 +302,26 @@ def create(self) -> Result:
...
             with FileSystem.new(
-                name='squashfs',
+                name=root_filesystem,
                 device_provider=DeviceProvider(),
                 root_dir=self.root_dir + os.sep,
                 custom_args={
                     'compression':
                         self.xml_state.build_type.get_squashfscompression()
-                }
+                } if root_filesystem == 'squashfs' else {'compression': ''}
```
Conan-Kudo, line 311: Can we not set erofs compression?

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: reading specs is no fun, no indentation between conditions, so hard to read. Thanks

kiwi/builder/live.py
schaefi: We have a `squashfscompression` attribute and so far none for erofs. I go with the default setting in the initial support. In a followup PR we can add custom erofs settings as we see fit

schaefi: If you look at `man mkfs.erofs` there are many options to set for the filesystem. I'd like to have a conversation first which of those are a priority for our use cases

Conan-Kudo: Okay. I'd definitely like to see us expose the ability to configure erofs compression, because otherwise livecds with erofs will be larger than with squashfs...

Conan-Kudo: Here's what I see for erofs compression options:

```
 -zX[,level=Y]         X=compressor (Y=compression level, Z=dictionary size, optional)
    [,dictsize=Z]      alternative compressors can be separated by colons(:)
    [:...]             supported compressors and their option ranges are:
                         lz4
                         lz4hc
                           [,level=<0-12>]		(default=9)
                         lzma
                           [,level=<0-9,100-109>]	0-9=normal, 100-109=extreme (default=6)
                           [,dictsize=<dictsize>]	(default=<auto>, max=8388608)
                         deflate
                           [,level=<0-9>]		(default=1)
                           [,dictsize=<dictsize>]	(default=32768, max=32768)
                         libdeflate
                           [,level=<0-12>]		(default=1)
                         zstd
                           [,level=<0-22>]		(default=3)
                           [,dictsize=<dictsize>]	(default=<auto>, max=1048576)
```

The ones I expect to use are zstd, lz4/lz4hc, or lzma. I don't really expect to use deflate any. I also expect exposing the compression level setting too.

schaefi:
Rawhide LiveEroFS integration test

schaefi:
Filesystem image integration test

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:rawhide/test-image-erofs

kiwi/builder/live.py
schaefi: ok I added the `erofscompression` attribute

schaefi:
@Conan-Kudo all done now, let's wait for the integration tests to pass

Conan-Kudo (approved):

Issue #2652, opened 2024-09-24
No way to delete /dev


alyssarosenzweig:
https://github.com/FEX-Emu/FEX requires an x86 "rootfs" consisting of x86 libraries and binaries, which it internally overlays onto the Arm file system at the directory level to build an accurate emulation environment. In Fedora, we'd like to use Kiwi to build the rootfs.

The problem is that the final compressed rootfs image cannot contain directories like `/dev`, `/proc`, and `/sys`, since the presence of those directories in the image would cause them to be used instead of passing through the real host  /dev, /proc, and /sys filesystems. Upstream FEX solves this in a simple way: after building the image inside the chroot, it exits the chroot and removes those directories outside the chroot right before compressing.

We need a way to do that with Kiwi. We cannot clean up these directories from `config.sh`, because that runs inside the chroot, and we can't delete `/dev` from inside the chroot. Nor can we use the few outside-of-chroot hooks because they are only for special purposes that don't apply here (we are not building a delta image, this is a full rootfs from scratch minus these few special directories).

The feature request is therefore one of:

1. Add a new hook to run a script outside of the chroot to finalize the image, or
2. Add declarative support for removing unwanted files/directories again.
3. Add declarative support for passing `--exclude-regex` to erofs at compression time.

Either option would work for us, the key bit is that it happens just before compressing since it will be impossible to chroot in after running this cleanup.

--

Looks like exclude_files.yaml should do this?

PR #2653, opened 2024-09-24
Honour custom exclude for filesystem builds

alyssarosenzweig:
All other call sites honour the custom exclude file, it's just this one that needs to be fixed. This unblocks use of Kiwi for generating FEX rootfs.

Closes: #2652


alyssarosenzweig:
> @alyssarosenzweig Can you address the style check error?
> 
> ```
> /home/runner/work/kiwi/kiwi/kiwi/builder/filesystem.py:208:64: W504 line break after binary operator
> 1     W504 line break after binary operator
> 1
> ```

Done.

Conan-Kudo (approved):

PR #2654, opened 2024-09-24
Fix globbing with exclude with regex

alyssarosenzweig:

This fixes a collection of bugs when producing erofs images.

On one hand, this ensures that an exclude of `/sys` doesn't accidentally match `/lib/libsystemd.so`, only `/sys/whatever`.

On the other hand, this ensures that `/dev/*` does match `/dev/vda` and not just `/dev///////////`.

This fixes libsystemd.so getting dropped in Kiwi-built FEX images.

### DIFF
--- a/kiwi/filesystem/erofs.py
+++ b/kiwi/filesystem/erofs.py
@@ -45,7 +45,12 @@ class FileSystemEroFs(FileSystemBase):
 
         if exclude:
             for item in exclude:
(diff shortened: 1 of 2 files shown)


schaefi (approved):
Oh, thanks much. Yes I was not paying enough attention to the result of the expression. Looks good to me :+1:

alyssarosenzweig:
Thanks!

PR #2657, opened 2024-09-26
Add quota attribute to volume section

### LINKED ISSUE #2651: Add quota attribute in <volume> definition for Btrfs subvolume usable when btrfs_quota_groups is set to true
## Description
Currently there is no possibility to set Btrfs subvolume quota value in kiwi xml. Quota can be useful to prevent issues with space usage and Btrfs.
Adding a quota attribute in the volume definition like this example:

```
<type
    [...]
    btrfs_quota_groups="true"
    [...]
    >
    <systemdisk>
        <volume name="home"/>
        <volume name="root"/>
        <volume name="tmp"/>
        <volume name="opt"/>
        <volume name="srv"/>
        <volume name="usr/local"/>
        <volume name="var" copy_on_write="false" quota="5G"/>
        <volume name="var/tmp" copy_on_write="false" quota="5G"/>
        <volume name="var/log" copy_on_write="false" quota="10G"/>
        <volume name="var/log/audit" copy_on_write="false" quota="5G"/>
     </systemdisk>
</type>
```

schaefi:
Allow to set quota per volume for the btrfs filesystem This Fixes #2651

### DIFF
--- a/kiwi/volume_manager/btrfs.py
+++ b/kiwi/volume_manager/btrfs.py
@@ -448,7 +448,7 @@ class VolumeManagerBtrfs(VolumeManagerBase):
             if attribute.startswith('quota='):
                 quota = attribute.split('=')[1]
                 Command.run(
-                    ['btrfs', 'quota', 'enable', volume_path]
+                    ['btrfs', 'quota', 'enable', '--simple', volume_path]
                 )
                 Command.run(
                     ['btrfs', 'qgroup', 'limit', quota, volume_path]
--- a/kiwi/volume_manager/btrfs.py
+++ b/kiwi/volume_manager/btrfs.py
@@ -224,6 +224,10 @@ class VolumeManagerBtrfs(VolumeManagerBase):
                         os.path.normpath(toplevel + os.sep + volume.realpath)
                     ]
                 )
+                self._apply_quota(
+                    os.path.normpath(toplevel + os.sep + volume.realpath),
+                    volume.attributes
+                )
                 self.apply_attributes_on_volume(
                     toplevel, volume
                 )
@@ -438,6 +442,17 @@ class VolumeManagerBtrfs(VolumeManagerBase):
             Command.run(
                 ['btrfs', 'property', 'set', sync_target, 'ro', 'true']
             )
+
+    def _apply_quota(self, volume_path: str, attributes: List[str]):
+        for attribute in attributes:
+            if attribute.startswith('quota='):
+                quota = attribute.split('=')[1]
+                Command.run(
+                    ['btrfs', 'quota', 'enable', volume_path]
+                )
+                Command.run(
+                    ['btrfs', 'qgroup', 'limit', quota, volume_path]
+                )
 
     def _has_root_volume(self) -> bool:
         has_root_volume = bool(self.custom_args['root_is_subvolume'])
--- a/build-tests/x86/tumbleweed/test-image-disk/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-disk/appliance.kiwi
@@ -24,7 +24,7 @@
                 <oem-multipath-scan>false</oem-multipath-scan>
             </oemconfig>
             <systemdisk>
-                <volume name="home"/>
+                <volume name="home" quota="5G"/>
             </systemdisk>
         </type>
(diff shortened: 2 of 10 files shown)


kiwi/volume_manager/btrfs.py
```diff
@@ -439,6 +443,17 @@ def set_property_readonly_root(self):
...
+    def _apply_quota(self, volume_path: str, attributes: List[str]):
+        for attribute in attributes:
+            if attribute.startswith('quota='):
+                quota = attribute.split('=')[1]
+                Command.run(
+                    ['btrfs', 'quota', 'enable', volume_path]
+                )
+                Command.run(
+                    ['btrfs', 'qgroup', 'limit', quota, volume_path]
+                )
```
Conan-Kudo, line 455: We should not use qgroups. We should use simple quotas instead. https://lwn.net/Articles/944371/

Conan-Kudo (changes requested):

kiwi/volume_manager/btrfs.py
schaefi: Now I'm having a problem. In the issue description the reported suggested to use qgroups. I was also reading about qgroups and the simple groups and I have no idea what values one over the other.

```
However, due to btrfs's high level decision to tradeoff snapshot
performance against ref-counting performance, qgroups suffer from
non-trivial performance issues that make them unattractive in certain
workloads.
```

Is that the justification to move ?

I'm fine with it, but wanted to double check

Conan-Kudo: Yeah, @boryas added it because of this.

schaefi: And what's the commandline API to make use of squota ? It's somehow quite hard to find commandline docu. I was reading this one: https://btrfs.readthedocs.io/en/latest/Qgroups.html#simple-quotas-squota

Conan-Kudo: Just use `btrfs quota enable --simple` instead of `btrfs quota enable`, I believe.

schaefi: Thanks, that helped to search it 

```
==API==
Squotas reuses the api of qgroups. The only difference is that when you
enable quotas via `btrfs quota enable`, you pass the `--simple` flag.
Squotas will always report exclusive == shared for each qgroup. Squotas
deal with extent_item/metadata_item sizes and thus do not do anything
special with compression. Squotas also introduce auto inheritance for
nested subvols. The API is documented more fully in the documentation
patches in btrfs-progs.
```

I don't think all of our supported distros have it

schaefi: I applied that and will check if the integration test will build

Conan-Kudo: It was added in btrfs-progs v6.5.2, which was a year ago.

schaefi:
```
iosc ls SUSE:SLE-15-SP6:Update/btrfsprogs
boot-btrfs.sh
btrfs-progs-v6.5.1.tar.gz
```

schaefi: anyway let's go this way and react when there are complains

schaefi:
The integration test for TW worked

```
[   70s] [ DEBUG   ]: 12:03:50 | EXEC: [btrfs subvolume create /var/tmp/kiwi_volumes.29npkpkf/@/home]
[   70s] [ DEBUG   ]: 12:03:50 | Looking for btrfs in /sbin:/usr/sbin:/usr/local/sbin:/root/bin:/usr/local/bin:/usr/bin:/bin
[   70s] [ DEBUG   ]: 12:03:50 | EXEC: [btrfs quota enable --simple /var/tmp/kiwi_volumes.29npkpkf/@/home]
[   70s] [ DEBUG   ]: 12:03:50 | Looking for btrfs in /sbin:/usr/sbin:/usr/local/sbin:/root/bin:/usr/local/bin:/usr/bin:/bin
[   70s] [ DEBUG   ]: 12:03:50 | EXEC: [btrfs qgroup limit 5G /var/tmp/kiwi_volumes.29npkpkf/@/home]

```

Conan-Kudo (approved):

PR #2659, opened 2024-09-26
package_manager: apt: run postinst scripts for ca-certificates

IsaacJT:
The ca-certificates package is required when any of the repositories
being used for the image use HTTPS. In order to support this, the
package needs to be added to the bootstrap packages. However, the
bootstrap package installer will not run the postinst script, which is
required to configure the system to use the SSL certificates.

This patch adjusts the bootstrap package installer to additionally run
the postinst scripts for ca-certificates.

If the ca-certificates package has not been configured, the build will
fail with the following error:

` Certificate verification failed: The certificate is NOT trusted. The certificate issuer is unknown.  Could not handshake: Error in the certificate verification. `

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -261,7 +261,7 @@ class PackageManagerApt(PackageManagerBase):
                 done < {1}
                 while read -r deb;do
                     pushd "$(dirname "$deb")" >/dev/null || exit 1
-                    if [[ "$(basename "$deb")" == base-passwd* ]];then
+                    if [[ "$(basename "$deb")" == base-passwd* ]] || [[ "$(basename "$deb")" == ca-certificates* ]]; then
                         echo "Running pre/post package scripts for $(basename "$deb")"
                         dpkg -e "$deb" "{0}/DEBIAN"
                         test -e {0}/DEBIAN/preinst && chroot {0} bash -c "/DEBIAN/preinst install"
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -265,7 +265,7 @@ class PackageManagerApt(PackageManagerBase):
                         echo "Running pre/post package scripts for $(basename "$deb")"
                         dpkg -e "$deb" "{0}/DEBIAN"
                         test -e {0}/DEBIAN/preinst && chroot {0} bash -c "/DEBIAN/preinst install"
-                        test -e {0}/DEBIAN/postinst && chroot {0} bash -c "/DEBIAN/postinst"
+                        test -e {0}/DEBIAN/postinst && chroot {0} bash -c "/DEBIAN/postinst configure"
                         rm -rf {0}/DEBIAN
                     fi
                     popd >/dev/null || exit 1
--- a/test/unit/package_manager/apt_test.py
+++ b/test/unit/package_manager/apt_test.py
@@ -136,7 +136,7 @@ class TestPackageManagerApt:
                 done < {1}
                 while read -r deb;do
                     pushd "$(dirname "$deb")" >/dev/null || exit 1
-                    if [[ "$(basename "$deb")" == base-passwd* ]];then
+                    if [[ "$(basename "$deb")" == base-passwd* ]] || [[ "$(basename "$deb")" == ca-certificates* ]]; then
                         echo "Running pre/post package scripts for $(basename "$deb")"
                         dpkg -e "$deb" "{0}/DEBIAN"
                         test -e {0}/DEBIAN/preinst && chroot {0} bash -c "/DEBIAN/preinst install"
--- a/test/unit/package_manager/apt_test.py
+++ b/test/unit/package_manager/apt_test.py
@@ -140,7 +140,7 @@ class TestPackageManagerApt:
                         echo "Running pre/post package scripts for $(basename "$deb")"
                         dpkg -e "$deb" "{0}/DEBIAN"
                         test -e {0}/DEBIAN/preinst && chroot {0} bash -c "/DEBIAN/preinst install"
-                        test -e {0}/DEBIAN/postinst && chroot {0} bash -c "/DEBIAN/postinst"
+                        test -e {0}/DEBIAN/postinst && chroot {0} bash -c "/DEBIAN/postinst configure"
                         rm -rf {0}/DEBIAN
                     fi
                     popd >/dev/null || exit 1


IsaacJT:
Running Kiwi with these patches allow the build to succeed. The following additional output is printed:

```
Running pre/post package scripts for ca-certificates_20230311_all.deb
 stat: cannot statx '/usr/local': No such file or directory
stat: cannot statx '/usr/local': No such file or directory
Updating certificates in /etc/ssl/certs...
140 added, 0 removed; done.
```

The "No such file or directory" messages are not fatal and don't cause any issues.

IsaacJT:
@schaefi maybe it would make sense to run the post install scripts for all of the packages being installed during bootstrapping?

Conan-Kudo:
This is strictly speaking not required, and why wouldn't you just add it to the bootstrap section?

IsaacJT:
The postinstall scripts aren't run for packages in the bootstrap section, which means that the certificates aren't copied to the right place and `update-ca-certificates` doesn't get run. 
-------- Original Message -------- 
On 26/09/24 21:45, Neal Gompa wrote: 
This is strictly speaking not required, and why wouldn't you just add it to the bootstrap section in your description?

IsaacJT:
Example appliance description (adapted from https://github.com/OSInside/kiwi-descriptions/blob/main/debian/x86_64/debian-bookworm/config.xml) to reproduce the problem this PR fixes:

```

<image schemaversion="7.6" name="Debian-Bookworm_appliance">
    <description type="system">
        <author>Marcus Schäfer</author>
        <contact>marcus.schaefer@gmail.com</contact>
        <specification>Debian Bookworm Appliance</specification>
    </description>
    <profiles>
        <profile name="Live" description="Live image" import="true"/>
        <profile name="Virtual" description="Simple Disk image"/>
        <profile name="Disk" description="Expandable Disk image"/>
    </profiles>
    <preferences>
        <version>12.0</version>
        <packagemanager>apt</packagemanager>
        <bootsplash-theme>fade-in</bootsplash-theme>
        <bootloader-theme>starfield</bootloader-theme>
        <rpm-check-signatures>false</rpm-check-signatures>
        <locale>en_US</locale>
        <keytable>us</keytable>
        <timezone>Europe/Berlin</timezone>
    </preferences>
    <preferences profiles="Live">
        <type image="iso" flags="overlay" hybridpersistent_filesystem="ext4" hybridpersistent="true" firmware="efi"/>
    </preferences>
        <type image="oem" filesystem="ext4" kernelcmdline="console=ttyS0 splash" firmware="efi" format="qcow2">
            <oemconfig>
            </oemconfig>
        </type>
    </preferences>
    <preferences profiles="Disk">
        <type image="oem" filesystem="ext4" initrd_system="dracut" firmware="efi" installiso="true">
            <oemconfig>
                <oem-swap>true</oem-swap>
                <oem-device-filter>/dev/ram</oem-device-filter>
            </oemconfig>
        </type>
    </preferences>
    <users>
        <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root" groups="root"/>
    </users>
    <repository type="apt-deb" distribution="bookworm" components="main contrib non-free" repository_gpgcheck="false">
            <source path="https://deb.debian.org/debian"/>
    </repository>
    <repository type="apt-deb" repository_gpgcheck="false">
        <source path="obs://Virtualization:Appliances:Builder/Debian_12_x86_64"/>
    </repository>
    <packages type="image">
        <package name="grub-theme-starfield"/>
        <package name="plymouth-themes"/>
        <package name="vim"/>
        <package name="plymouth"/>
        <package name="grub-efi-amd64"/>
        <package name="dracut"/>
        <package name="xz-utils"/>
        <package name="binutils"/>
        <package name="linux-image-amd64"/>
        <package name="isolinux"/>
        <package name="syslinux"/>
        <package name="syslinux-common"/>
        <package name="init"/>
        <package name="gnupg"/>
        <package name="iproute2"/>
        <package name="iptables"/>
        <package name="iputils-ping"/>
        <package name="ifupdown"/>
        <package name="isc-dhcp-client"/>
        <package name="bsdmainutils"/>
        <package name="netplan.io"/>
        <package name="networkd-dispatcher"/>
        <package name="net-tools"/>
        <package name="netbase"/>
        <package name="dbus"/>
        <package name="zstd"/>
        <package name="util-linux"/>
        <package name="locales-all"/>
    </packages>
    <packages type="iso">
        <package name="dracut-kiwi-live"/>
    </packages>
    <packages type="oem">
    </packages>
    <packages type="bootstrap">
        <package name="grub2"/>
        <package name="usrmerge"/>
    </packages>
</image>
```

Building this with the current main fails due to the missing certificates

schaefi:
@IsaacJT  @Conan-Kudo 

Thanks for the fix, this makes sense and is another part that I overlooked when we moved away from debootstrap

> @schaefi maybe it would make sense to run the post install scripts for all of the packages being installed during bootstrapping?

I think we should do this. In a former implementation I had it as such but several post install scripts failed to run. I remember that I called them directly after the unpacking and I think this was a mistake. iirc Debian based systems runs all scripts after all packages got installed, is that correct ? If so I think it would be good to do that after all bootstrap packages got unpacked

I also think it would be good to do this as a refactor which moves the unpacking `dpkg-deb` and the call of postscripts into proper python code.

So how about merging this one first, then come up with a refactor ?

Thoughts

IsaacJT:
> iirc Debian based systems runs all scripts after all packages got installed, is that correct ?

Correct - if you use apt, everything gets unpacked and then configured (i.e. the `postinst` script gets run). You could simulate this behaviour when just installing everything one-by-one with dpkg by calling `dpkg --unpack *.deb`, which will only unpack the archives, and then at the end running `dpkg --configure -a`. 

> So how about merging this one first, then come up with a refactor ?

Sounds good to me :)

schaefi:
@IsaacJT I opened #2661 would be great if you can test/assist. Thanks in advance

IsaacJT:
> @IsaacJT I opened #2661 would be great if you can test/assist. Thanks in advance

Thanks a lot! I'll have a look on Monday :)

schaefi:
@IsaacJT I tested the the change from here with your bootstrap section and the log now shows

```
[ DEBUG   ]: 14:05:45 | Running pre/post scripts for: ca-certificates_20240203_all.deb
[ DEBUG   ]: 14:05:46 | Running pre/post scripts for: apt-transport-https_2.7.14build2_all.deb
```

so should be working as expected

schaefi:
Closing this one in favor of #2661

PR #2661, opened 2024-09-27
Run package scripts in apt bootstrap phase

schaefi:
The bootstrap procedure based on apt only runs a manual collection of package scripts. This commit refactors the code that unpacks the bootstrap packages to a python implementation and adds a method to run the bootstrap scripts from all packages resolved by apt.

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -19,6 +19,7 @@ import re
 import os
 import glob
 import logging
+import pathlib
 from typing import (
     List, Dict
 )
@@ -221,6 +222,13 @@ class PackageManagerApt(PackageManagerBase):
         # the essential debs and tell us their full in the apt cache without
         # actually installing them.
         try:
+            # TODO: Drop once apt 2.5.4 is widely available.
+            pathlib.Path(f'{self.root_dir}/var/lib/dpkg').mkdir(
+                parents=True, exist_ok=True
+            )
+            pathlib.Path(f'{self.root_dir}/var/lib/dpkg/status.kiwi').touch()
+            pathlib.Path(f'{self.root_dir}/var/lib/dpkg/available').touch()
+
             if 'apt' not in self.package_requests:
                 self.package_requests.append('apt')
             update_cache = [
@@ -253,6 +261,7 @@ class PackageManagerApt(PackageManagerBase):
             # Extract bootstrap packages
             with open(package_names.name) as packages:
                 solved_debootstrap_packages = [p.rstrip() for p in packages]
+            self.command_env['PATH'] = '$PATH:/usr/bin:/bin:/usr/sbin:/sbin'
             for package in solved_debootstrap_packages:
                 Command.run(
                     [
@@ -260,7 +269,7 @@ class PackageManagerApt(PackageManagerBase):
                         'dpkg-deb --fsys-tarfile {0} | tar -C {1} -x'.format(
                             package, self.root_dir
                         )
-                    ]
+                    ], self.command_env
                 )
             # Run package scripts. Unfortuantely Debian based systems
             # requires special sauce for bootstrap. See the exceptions
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -19,7 +19,6 @@ import re
 import os
 import glob
 import logging
-from textwrap import dedent
 from typing import (
     List, Dict
 )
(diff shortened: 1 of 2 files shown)


kiwi/package_manager/apt.py
```diff
@@ -513,3 +501,54 @@ def _package_requests(self) -> List:
...
+    def _run_bootstrap_scripts(
+        self, solved_debootstrap_packages: List[str], only_for: List[str] = []
+    ):
+        # FIXME: weird stuff
+        # These env variables are required for the package scripts
+        # to work. I don't understand their meaning and I'm not sure
+        # if hardcoding this for libc6 is something we can afford ?
+        # Help is appreciated
+        self.command_env['DPKG_MAINTSCRIPT_NAME'] = 'true'
+        self.command_env['DPKG_MAINTSCRIPT_PACKAGE'] = 'libc6'
```
schaefi, line 514: @IsaacJT As we discussed I created a refactor and implementation for the apt based bootstrap such that all package scripts from the resolved apt list are taken into account.

At this point however I needed two weird environment variables and I had to set libc6 as DPKG_MAINTSCRIPT_PACKAGE otherwise I got usr-merge complains. I really have no idea what these variables really do and my knowledge on the inner works of Debian/Ubuntu packaging is really bad.

I wanted to ask if you can give this PR a try if it would also fix the issue you found and maybe check if the above is ok as a permanent in code setup or if this is all wrong. I believe the way how I set the variables is not a good way to handle the issue that comes up if you disable them. Feel free to run the process without these variables so you can see the errors from apt.

Thanks much

kiwi/package_manager/apt.py
```diff
@@ -513,3 +501,54 @@ def _package_requests(self) -> List:
...
+                    ], self.command_env
+                )
+            # 2. postinst
+            if os.path.exists(script_post):
+                Command.run(
+                    [
+                        'chroot', self.root_dir, 'bash',
+                        f'{script_post.replace(self.root_dir, "")}', 'configure'
+                    ], self.command_env
+                )
```
schaefi, line 554: @IsaacJT If you know a shorter or more straight forward way to call them all I appreciate your feedback.
In any case I had to run the `base-passwd` script first because that creates users, e.g root which is required for any other package script to work. In the apt sequence though it was not first and thus I had this strange two pass system where I run package scripts in a manual order first and then I do it again with the order of the apt resolver. It all feels not really perfect to be honest :)

kiwi/package_manager/apt.py
```diff
@@ -247,42 +253,33 @@ def _process_install_requests_bootstrap(self) -> CommandCallT:
...
-                    script.format(self.root_dir, package_names.name)
+            # Extract bootstrap packages
+            with open(package_names.name) as packages:
+                solved_debootstrap_packages = [p.rstrip() for p in packages]
+            self.command_env['PATH'] = '$PATH:/usr/bin:/bin:/usr/sbin:/sbin'
+            for package in solved_debootstrap_packages:
+                Command.run(
+                    [
+                        'bash', '-c',
+                        'dpkg-deb --fsys-tarfile {0} | tar -C {1} -x'.format(
```
IsaacJT, line 271: Is there a specific reason you're using `tar` here? You can use `dpkg -e <.deb path>` to extract the scripts from a .deb

kiwi/package_manager/apt.py
IsaacJT: AFAIK `DPKG_MAINTSCRIPT_NAME` should be the name of the script (e.g. `postinst`, etc.) and `DPKG_MAINTSCRIPT_PACKAGE` should be the name of the package that the scripts belong to. I would have to double check this though...

Haven't had time to a have a proper look through yet.

kiwi/package_manager/apt.py
schaefi: This extraction is not for the scripts. This extraction "installs" the downloaded bootstrap .deb package into the new root-tree. From the man page of dpkg-deb

```
--fsys-tarfile archive
           Extracts the filesystem tree data from a binary package and
           sends it to standard output in tar format.  Together with
           tar(1) this can be used to extract a particular file from a
           package archive.
```

So this call is not related to the script part of things, which comes later.

kiwi/package_manager/apt.py
schaefi: Ah ok

schaefi: As we run the scripts manually the setting of just `true` would then be ok ?

schaefi: Please also note that `libc6` was the only package in the collection of all packages that complains about the missing variables DPKG_MAINTSCRIPT_*

schaefi: I guess I simply do not understand the concept behind them

kiwi/package_manager/apt.py
schaefi: So the process to bootstrap a debian based system follows the following logic

1. apt-get _with_call_options_to_resolve_packages_and_only_download_them
2. dpkg-deb --fsys-tarfile to extract (install) the set of resolved and downloaded packages
3. run the scripts

kiwi/package_manager/apt.py
schaefi: I updated the comment for this variable settings. I could not find any bad or negative side effect of setting them during the bootstrap of a Debian based system. I tested Ubuntu and Debian and the integration tests pass and boots. Thus I leave it up to the experts if this needs to be changed or can harm the bootstrap procedure.

kiwi/package_manager/apt.py
IsaacJT: Ah, I see. You could instead just use `dpkg -x <.deb> <dest>`  to skip the intermediate tar stage

kiwi/package_manager/apt.py
IsaacJT: Maybe `DPKG_MAINTSCRIPT_PACKAGE` should be set in the for block below so that it represents the name of the package that the script is currently being run for? And `DPKG_MAINTSCRIPT_NAME` could be set 
to the script name immediately before calling the script, e.g. 

```
if os.path.exists(script_pre):
    self.command_env['DPKG_MAINTSCRIPT_PACKAGE'] = 'preinst'
    ...
```

I think this would more closely emulate the environment that the scripts would normally be run in

IsaacJT:
Thanks @schaefi, this seems like a really good solution.  I think if the variables are updated on the go as per my other comments it should be all good to go

Conan-Kudo (approved):

PR #2664, opened 2024-10-15
Add new registry section

### LINKED ISSUE #2663: Container installation during image build
## Problem description

When building a VM image with kiwi there is no feature that supports the inclusion of containers in the image build unless any specific container is wrapped in a package. A use case exists where it is desirable to be able to include containerized application, SUSE Multi Linux Manager for example in an image such that upon startup of an instance the container application can be started without having to pull the container from a registry first. This will reduce the custom work a user of kiwi has to do.

## Expected behaviour

Specify a <registry> element in the kiwi config file
Specify a container to pull into the image with a <container> element in the kiwi config file

During build kiwi will pull the container using skopeo or some other suitable tool. When an instance of the image is booted a kiwi service will automatically import the container present in the image into the container runtime (podman or docker) local repository. This should happen prior to a service the use may include that starts the containerized application. Users would have the requirement to add a service with "After: kiwi-container-load.service" or whatever name may be chosen.

* KIWI version: 
* Operating system target version: 
  - Will need new OBS integration

schaefi:
Allow to specify references to OCI containers in the image description like in the following example:

```xml
<registry source="registry.example.com">
    <container name="some" tag="latest" path="some/path" use_with="podman"/>
    <container name="some_other" use_with="docker" fetch_only="true"/>
</registry>
```

During the kiwi process the containers are fetched into a temporary location and a systemd service is configured to one time load the containers into the local registry at first boot of the system.

This Fixes #2663

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -1233,6 +1233,76 @@ Used to customize the installation media images created for oem images
 deployment.
 For details see: :ref:`installmedia_customize`
 
+.. _sec.registry:
+
+<containers>
+------------
+
+Setup containers to fetch from a registry assigned to one
+of the supported container backends
+
+.. code:: xml
+
+   <containers source="registry.opensuse.org" backend="podman">
+       <container name="some"/>
+   </containers>
+
+The optional containers element specifies the location of one ore
+more containers on a registry `source` server. {kiwi} will take
+this information and fetch the containers as OCI archives to
+the image. On first boot those container archives will be loaded
+into the local container backend store for the selected
+backend and the archive files get deleted.
+
+Supported `backend` values are `docker` and `podman`.
+The `backend` attribute is mandatory and specifies for which
+container backend the image should be available in the system.
+The `containers` element has the following optional attributes:
+
+arch="arch"
+  The containers section can be configured to apply only for a certain
+  architecture. In this case specify the `arch` attribute with a
+  value as it is reported by :command:`uname -m`.
+
+profiles="name[,name]"
+  A list of profiles to which this containers selection applies
+  (see :ref:`image-profiles`).
+
+<containers><container>
+-----------------------
+
+Details about the container
+
+.. code:: xml
+
+   <containers source="registry.opensuse.org" backend="podman">
+       <container name="some"/>
+   </containers>
+
+The `name` attributes is mandatory and specifies
+the name of the container as it exists in the registry.
+The `container` element has the following optional attributes:
+
+path="some/path"
+  The path to the container in the registry. If not specified
+  the value defaults to `/`
+
+fetch_only="true|false"
+  If set to `true` kiwi will only fetch the container but does not
+  setup the systemd unit for loading the container into
+  the local registry. In this mode the container archive file stays
+  in the system and can be handled in a custom way. By default
+  `fetch_only` is set to `false`.
+
+tag="tagname"
+  Specifies the container tag to fetch. If not set the tag name
+  defaults to `latest`
+
+arch="arch"
+  The container section can be configured to apply only for a certain
+  architecture. In this case specify the `arch` attribute with a
+  value as it is reported by :command:`uname -m`.
+
 .. _sec.repository:
 
 <repository>
--- a/kiwi/builder/template/container_import.py
+++ b/kiwi/builder/template/container_import.py
@@ -1,0 +1,68 @@
+# Copyright (c) 2024 SUSE LLC.  All rights reserved.
+#
+# This file is part of kiwi.
+#
+# kiwi is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# kiwi is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with kiwi.  If not, see <http://www.gnu.org/licenses/>
+#
+import os
+from typing import List
+from string import Template
+from textwrap import dedent
+
+
+class BuilderTemplateSystemdUnit:
+    """
+    **systemd unit file templates**
+    """
+    def __init__(self):
+        self.unit = dedent('''
+            # kiwi generated unit file
+            [Unit]
+            Description=Import Local Container(s)
+        ''').strip() + os.linesep
+
+        self.service = dedent('''
+            [Service]
+            Type=oneshot
+        ''').strip() + os.linesep
+
+        self.install = dedent('''
+            [Install]
+            WantedBy=multi-user.target
+        ''').strip() + os.linesep
+
+    def get_container_import_template(
+        self, container_files: List[str], load_commands: List[List[str]],
+        after: List[str]
+    ):
+        template_data = self.unit
+        for container_file in container_files:
+            template_data += 'ConditionPathExists={0}{1}'.format(
+                container_file, os.linesep
+            )
+        if after:
+            template_data += 'After={0}{1}'.format(
+                ' '.join(after), os.linesep
+            )
+        template_data += self.service
+        for load_command in load_commands:
+            template_data += 'ExecStart={0}{1}'.format(
+                ' '.join(load_command), os.linesep
+            )
+        for container_file in container_files:
+            template_data += 'ExecStartPost=/bin/rm -f {0}{1}'.format(
+                container_file, os.linesep
+            )
+        template_data += self.install
+        return Template(template_data)
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -51,6 +51,7 @@ from kiwi.archive.tar import ArchiveTar
 from kiwi.utils.compress import Compress
 from kiwi.utils.command_capabilities import CommandCapabilities
 from kiwi.utils.rpm_database import RpmDataBase
+from kiwi.builder.template.container_import import BuilderTemplateSystemdUnit
 from kiwi.system.profile import Profile
 
 from kiwi.exceptions import (
@@ -86,6 +87,48 @@ class SystemSetup:
         self.root_dir = root_dir
         self._preferences_lookup()
         self._oemconfig_lookup()
+
+    def setup_registry_import(self) -> None:
+        """
+        Fetch container(s) and activate systemd unit to load
+        containers from local oci-archive file during boot
+        """
+        container_files_to_load = []
+        container_execs_to_load = []
+        after_services = set()
+        for container in self.xml_state.get_containers():
+            log.info(f'Fetching container: {container.name}')
+            pathlib.Path(f'{self.root_dir}/{defaults.LOCAL_CONTAINERS}').mkdir(
+                parents=True, exist_ok=True
+            )
+            Command.run(
+                ['chroot', self.root_dir] + container.fetch_command
+            )
+            if container.load_command:
+                container_files_to_load.append(container.container_file)
+                container_execs_to_load.append(container.load_command)
+                if container.backend == 'docker':
+                    after_services.add('docker.service')
+
+        if container_files_to_load:
+            log.info('--> Setup kiwi_containers.service import unit')
+            service = BuilderTemplateSystemdUnit()
+            unit_template = service.get_container_import_template(
+                container_files_to_load, container_execs_to_load,
+                list(after_services)
+            )
+            unit = unit_template.substitute()
+            unit_file = '{0}/etc/systemd/system/{1}.service'.format(
+                self.root_dir, 'kiwi_containers'
+            )
+            with open(unit_file, 'w') as systemd:
+                systemd.write(unit)
+            Command.run(
+                [
+                    'chroot', self.root_dir,
+                    'systemctl', 'enable', 'kiwi_containers'
+                ]
+            )
 
     def import_description(self) -> None:
         """
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -91,6 +91,7 @@ RUNTIME_CHECKER_METADATA = '{}/runtime_checker_metadata.yml'.format(
 ) if MODULE_SPEC else 'unknown'
 
 TEMP_DIR = '/var/tmp'
+LOCAL_CONTAINERS = '/var/tmp/kiwi_containers'
 CUSTOM_RUNTIME_CONFIG_FILE = None
 PLATFORM_MACHINE = platform.machine()
 EFI_FAT_IMAGE_SIZE = 20
--- a/kiwi/schema/kiwi.rnc
+++ b/kiwi/schema/kiwi.rnc
@@ -90,6 +90,7 @@ div {
             k.drivers* &
             k.strip* &
             k.repository* &
+            k.containers* &
             k.packages* &
             k.extension?
         }
@@ -1043,6 +1044,61 @@ div {
         ## definition can be composed by other existing profiles.
         element requires {
             k.requires.attlist,
+            empty
+        }
+}
+
+#==========================================
+# common element <containers>
+#
+div {
+    k.containers.profiles.attribute = k.profiles.attribute
+    k.containers.arch.attribute = k.arch.attribute
+    k.containers.source.attribute =
+        ## Name of registry source server
+        attribute source { text }
+    k.containers.backend.attribute =
+        ## Use container with specified container backend
+        attribute backend { "podman" | "docker" }
+    k.containers.attlist =
+        k.containers.profiles.attribute? &
+        k.containers.arch.attribute? &
+        k.containers.source.attribute &
+        k.containers.backend.attribute
+    k.containers =
+        element containers {
+            k.containers.attlist,
+            k.container+
+        }
+}
+
+#==========================================
+# common element <container>
+#
+div {
+    k.container.arch.attribute = k.arch.attribute
+    k.container.name.attribute =
+        ## Container name
+        attribute name { text }
+    k.container.path.attribute =
+        ## Container path, default to '/' if not specified
+        attribute path { text }
+    k.container.tag.attribute =
+        ## Container tag, defaults to 'latest' if not specified
+        attribute tag { text }
+    k.container.fetch_only.attribute =
+        ## Only fetch the container but do not activate the
+        ## loading of the container at first boot
+        attribute fetch_only { xsd:boolean }
+    k.container.attlist =
+        k.container.name.attribute &
+        k.container.arch.attribute? &
+        k.container.path.attribute? &
+        k.container.tag.attribute? &
+        k.container.fetch_only.attribute?
+    k.container =
+        element container {
+            k.container.attlist,
             empty
         }
 }
--- a/kiwi/schema/kiwi.rng
+++ b/kiwi/schema/kiwi.rng
@@ -209,6 +209,9 @@ named /etc/ImageID</a:documentation>
             <ref name="k.repository"/>
           </zeroOrMore>
           <zeroOrMore>
+            <ref name="k.containers"/>
+          </zeroOrMore>
+          <zeroOrMore>
             <ref name="k.packages"/>
           </zeroOrMore>
           <optional>
@@ -1597,6 +1600,108 @@ definition.</a:documentation>
         <a:documentation>Requires is used to set profiles dependencies, with it a profile
 definition can be composed by other existing profiles.</a:documentation>
         <ref name="k.requires.attlist"/>
+        <empty/>
+      </element>
+    </define>
+  </div>
+  <!--
+    ==========================================
+    common element <containers>
+    
+  -->
+  <div>
+    <define name="k.containers.profiles.attribute">
+      <ref name="k.profiles.attribute"/>
+    </define>
+    <define name="k.containers.arch.attribute">
+      <ref name="k.arch.attribute"/>
+    </define>
+    <define name="k.containers.source.attribute">
+      <attribute name="source">
+        <a:documentation>Name of registry source server</a:documentation>
+      </attribute>
+    </define>
+    <define name="k.containers.backend.attribute">
+      <attribute name="backend">
+        <a:documentation>Use container with specified container backend</a:documentation>
+        <choice>
+          <value>podman</value>
+          <value>docker</value>
+        </choice>
+      </attribute>
+    </define>
+    <define name="k.containers.attlist">
+      <interleave>
+        <optional>
+          <ref name="k.containers.profiles.attribute"/>
+        </optional>
+        <optional>
+          <ref name="k.containers.arch.attribute"/>
+        </optional>
+        <ref name="k.containers.source.attribute"/>
+        <ref name="k.containers.backend.attribute"/>
+      </interleave>
+    </define>
+    <define name="k.containers">
+      <element name="containers">
+        <ref name="k.containers.attlist"/>
+        <oneOrMore>
+          <ref name="k.container"/>
+        </oneOrMore>
+      </element>
+    </define>
+  </div>
+  <!--
+    ==========================================
+    common element <container>
+    
+  -->
+  <div>
+    <define name="k.container.arch.attribute">
+      <ref name="k.arch.attribute"/>
+    </define>
+    <define name="k.container.name.attribute">
+      <attribute name="name">
+        <a:documentation>Container name</a:documentation>
+      </attribute>
+    </define>
+    <define name="k.container.path.attribute">
+      <attribute name="path">
+        <a:documentation>Container path, default to '/' if not specified</a:documentation>
+      </attribute>
+    </define>
+    <define name="k.container.tag.attribute">
+      <attribute name="tag">
+        <a:documentation>Container tag, defaults to 'latest' if not specified</a:documentation>
+      </attribute>
+    </define>
+    <define name="k.container.fetch_only.attribute">
+      <attribute name="fetch_only">
+        <a:documentation>Only fetch the container but do not activate the
+loading of the container at first boot</a:documentation>
+        <data type="boolean"/>
+      </attribute>
+    </define>
+    <define name="k.container.attlist">
+      <interleave>
+        <ref name="k.container.name.attribute"/>
+        <optional>
+          <ref name="k.container.arch.attribute"/>
+        </optional>
+        <optional>
+          <ref name="k.container.path.attribute"/>
+        </optional>
+        <optional>
+          <ref name="k.container.tag.attribute"/>
+        </optional>
+        <optional>
+          <ref name="k.container.fetch_only.attribute"/>
+        </optional>
+      </interleave>
+    </define>
+    <define name="k.container">
+      <element name="container">
+        <ref name="k.container.attlist"/>
         <empty/>
       </element>
     </define>
--- a/kiwi/tasks/system_build.py
+++ b/kiwi/tasks/system_build.py
@@ -307,6 +307,7 @@ class SystemBuildTask(CliTask):
                 setup.setup_timezone()
                 setup.setup_permissions()
                 setup.import_files()
+                setup.setup_registry_import()
 
                 # setup permanent image repositories after cleanup
                 setup.import_repositories_marked_as_imageinclude()
--- a/kiwi/tasks/system_prepare.py
+++ b/kiwi/tasks/system_prepare.py
@@ -293,6 +293,7 @@ class SystemPrepareTask(CliTask):
                 setup.setup_timezone()
                 setup.setup_permissions()
                 setup.import_files()
+                setup.setup_registry_import()
 
                 # setup permanent image repositories after cleanup
                 setup.import_repositories_marked_as_imageinclude()
--- a/kiwi/xml_parse.py
+++ b/kiwi/xml_parse.py
@@ -812,7 +812,7 @@ class image(GeneratedsSuper):
     """The root element of the configuration file"""
     subclass = None
     superclass = None
-    def __init__(self, name=None, displayname=None, id=None, schemaversion=None, noNamespaceSchemaLocation=None, schemaLocation=None, include=None, description=None, preferences=None, profiles=None, users=None, drivers=None, strip=None, repository=None, packages=None, extension=None):
+    def __init__(self, name=None, displayname=None, id=None, schemaversion=None, noNamespaceSchemaLocation=None, schemaLocation=None, include=None, description=None, preferences=None, profiles=None, users=None, drivers=None, strip=None, repository=None, containers=None, packages=None, extension=None):
         self.original_tagname_ = None
         self.name = _cast(None, name)
         self.displayname = _cast(None, displayname)
@@ -852,6 +852,10 @@ class image(GeneratedsSuper):
             self.repository = []
         else:
             self.repository = repository
+        if containers is None:
+            self.containers = []
+        else:
+            self.containers = containers
         if packages is None:
             self.packages = []
         else:
@@ -911,6 +915,11 @@ class image(GeneratedsSuper):
     def add_repository(self, value): self.repository.append(value)
     def insert_repository_at(self, index, value): self.repository.insert(index, value)
     def replace_repository_at(self, index, value): self.repository[index] = value
+    def get_containers(self): return self.containers
+    def set_containers(self, containers): self.containers = containers
+    def add_containers(self, value): self.containers.append(value)
+    def insert_containers_at(self, index, value): self.containers.insert(index, value)
+    def replace_containers_at(self, index, value): self.containers[index] = value
     def get_packages(self): return self.packages
     def set_packages(self, packages): self.packages = packages
     def add_packages(self, value): self.packages.append(value)
@@ -950,6 +959,7 @@ class image(GeneratedsSuper):
             self.drivers or
             self.strip or
             self.repository or
+            self.containers or
             self.packages or
             self.extension
         ):
@@ -1017,6 +1027,8 @@ class image(GeneratedsSuper):
             strip_.export(outfile, level, namespaceprefix_, name_='strip', pretty_print=pretty_print)
         for repository_ in self.repository:
             repository_.export(outfile, level, namespaceprefix_, name_='repository', pretty_print=pretty_print)
+        for containers_ in self.containers:
+            containers_.export(outfile, level, namespaceprefix_, name_='containers', pretty_print=pretty_print)
         for packages_ in self.packages:
             packages_.export(outfile, level, namespaceprefix_, name_='packages', pretty_print=pretty_print)
         for extension_ in self.extension:
@@ -1097,6 +1109,11 @@ class image(GeneratedsSuper):
             obj_.build(child_)
             self.repository.append(obj_)
             obj_.original_tagname_ = 'repository'
+        elif nodeName_ == 'containers':
+            obj_ = containers.factory()
+            obj_.build(child_)
+            self.containers.append(obj_)
+            obj_.original_tagname_ = 'containers'
         elif nodeName_ == 'packages':
             obj_ = packages.factory()
             obj_.build(child_)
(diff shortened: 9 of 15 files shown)


schaefi:
There are two aspects that are not covered by this PR

1. The container tool chain needs to be part of the image. Meaning `skopeo` and the selected backend e.g `podman` needs to be explicitly installed via the package selection. There is no automation that would pull them in
2. Loading the containers into the storage backend requires space to be available in the image. There is no extra space addition done for this loading process because it cannot be predicted as it depends on the storage backend. Thus it's also a user task to give the image enough space such that the on-boot loading operation can succeed

Conan-Kudo (changes requested):
I'm not sure this is the model we want. Why is the registry the top-level element with containers inside? Why isn't this part of sources, and then we can add containers and flatpaks to the same section packages are set? Also, why do we wait until firstboot to load them into container storage?

schaefi:
I'm answering in reverse order:

> Also, why do we wait until firstboot to load them into container storage?

The reason for this is because loading a container archive into the storage area of the container backend requires the configured backend to be up and active. For example in podman your storage backend could be btrfs based and living in a snapshot which would not be able to be created/managed while we are building the offline image root tree. For e.g docker it's required that the docker daemon runs to actually load a container into the storage space, same for containerd... etc etc etc. It simply comes with too many restrictions when we want to load the container into the storage backend while it is not up and running. I don't want this limitation and therefore the model is fetch offline, load online. The price to pay for this model is storage space but that can be handled.

> Why isn't this part of sources, and then we can add containers and flatpaks to the same section packages are set?

The reason why I used a new toplevel element for this is because our sources model is based on `repository` followed by `packages` and a package manager that knows how to resolve using all repositories. When doing the same with containers we have no "container package manager" and have to tell in the image description which container should be fetched from which registry. Because of that I added a new section named `registry` and you need to bind the containers that you want to fetch from this registry via `container` sub-sections to that registry. This concept could grow with other data like flatpacks. In general with all data that is delivered through a registry

I'm open for other definition semantics but would like to ask for an example as you see it.

rjschwei (changes requested):
The "request changes" setting is for the spelling, all other comments are suggestions to consider, primarily from my perspective as a user of the feature.

kiwi/builder/template/container_import.py
```diff
@@ -0,0 +1,42 @@
...
+
+            [Service]
+            Type=oneshot
+            ExecStart=${load_command}
+            ExecStartPost=/bin/rm ${container_file}
+
+            [Install]
+            WantedBy=multi-user.target
+        ''').strip() + os.linesep
+        return Template(template_data)
```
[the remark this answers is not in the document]
Conan-Kudo: We probably want to have dependency relationships somehow expressed in the XML to translate to the units, yes.

Conan-Kudo:
I'd also like @travier to look at this and give feedback because I think we can kill two birds with one stone here and figure out the OCI Flatpak stuff that is needed for Fedora too.

kiwi/system/setup.py
```diff
@@ -87,6 +88,42 @@ def __init__(self, xml_state: XMLState, root_dir: str):
         self._preferences_lookup()
         self._oemconfig_lookup()
 
+    def setup_registry_import(self) -> None:
+        """
+        Fetch container(s) and activate systemd unit to load
+        fetched containers during boot
+        """
+        for container in self.xml_state.get_containers():
```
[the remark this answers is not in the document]
rjschwei: @Conan-Kudo What is your proposal for documentation and what the user is expected to do if they need a service that runs after the container load?

I understand the "better logging" argument. That said if I build an image that includes, let's say Rancher, which has some 60+ containers and I want to run additional setup tasks when all the application containers are installed I end up having to create a `After: A_LIST_OF_60+_SERVICES` unit file. I don't see how we can consider that to be user friendly. Certainly the logging can be solved in a satisfactory way in some other fashion. systemd captures what an executable that is invoked via a unit file outputs, as such some well placed "echo" statements in a generated shell script should provide reasonable debugging information.

kiwi/builder/template/container_import.py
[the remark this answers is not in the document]
schaefi: good point, I'll take a closer look once I've recovered from the *f* cold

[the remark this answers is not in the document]
rjschwei: Also, if `docker` is set the unit file should have a `After: docker.service` entry since dockerd needs to be running.

doc/source/image_description/elements.rst
```diff
@@ -1233,6 +1233,58 @@ Used to customize the installation media images created for oem images
...
+
+Setup containers to fetch from one ore more registry elements
+
+.. code:: xml
+
+   <registry source="registry.opensuse.org">
+       <container name="some" use_with="podman"/>
+   </registry>
+
+The optional registry element specifies the location of one ore
```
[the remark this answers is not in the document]
schaefi: fixed

doc/source/image_description/elements.rst
```diff
@@ -1233,6 +1233,58 @@ Used to customize the installation media images created for oem images
...
+
+<registry><container>
+---------------------
+
+Details about a container to fetch from a given source registry
+
+.. code:: xml
+
+   <registry source="registry.opensuse.org">
+       <container name="some" use_with="podman"/>
```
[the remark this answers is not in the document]

kiwi/builder/template/container_import.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) 2024 SUSE Software Solutions Germany GmbH.  All rights reserved.
```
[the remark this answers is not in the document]
schaefi: fixed

kiwi/builder/template/container_import.py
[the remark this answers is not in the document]
schaefi: I added a commit that implements a solution which I believe is acceptable. kiwi will now produce only one service file called `kiwi_containers.service` It makes use of the one-shot service types which also allows for multiple execution of programs. Writing an extra script is not needed in this case. The implementation also takes care for eventual backend service requirements. So far we only have `docker.service` to run beforehand. If we combine docker, podman, containerd, etc backends in the future the service requirement for all of them has to be met, which I believe is also a wanted behavior. Also if one execution fails the complete service run is marked failed, that's also what we want because a mixed state of success and error is not good.

kiwi/system/setup.py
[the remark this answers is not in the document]
schaefi: I think the way it is now offers a proper systemd journal and status information via `systemctl status -l kiwi_containers` and/or `journalctl -u kiwi_containers` or for long running loads  `journalctl -u kiwi_containers -f`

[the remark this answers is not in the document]
Conan-Kudo: If we're going to have dependency information, it would have to be in the XML description anyway. Additionally, Podman already supports being able to [define groups of containers with dependencies](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html) too with its quadlet functionality.

Conan-Kudo:
Is there a reason we can't add registry sources as repositories with `oci` type and then be able to reference them using the `alias` attribute for `containers`? Also, why should registry be mandatory here, since you can pull container images with fully qualified paths (e.g. `registry.opensuse.org/git`)? I would prefer we more closely mimic what people do on the commandline here and require fully qualified paths instead.

doc/source/image_description/elements.rst
```diff
@@ -1233,6 +1233,78 @@ Used to customize the installation media images created for oem images
...
+   <registry source="registry.opensuse.org">
+       <containers backend="podman">
+           <container name="some"/>
+       </containers>
+   </registry>
+
+Supported `backend` values as of today are `docker` and `podman`.
+The `backend` attribute is mandatory and specifies for which
+container backend the image should be available in the system.
+The `containera` element has the following optional attributes:
```
rjschwei, line 1274: s/containera/container/

And shouldn't the `path` attribute be explained in the next section as part of `<registry><containers><container>` ?

doc/source/image_description/elements.rst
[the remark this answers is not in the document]
rjschwei: Hmm the diff in GitHub still shows `ore` but this could be a GitHub diff artifact. I am aware that GitHub has occasional issues with displaying an accurate diff.

doc/source/image_description/elements.rst
```diff
@@ -1233,6 +1233,78 @@ Used to customize the installation media images created for oem images
...
+
+.. code:: xml
+
+   <registry source="registry.opensuse.org">
+       <containers backend="podman">
+           <container name="some"/>
+       </containers>
+   </registry>
+
+The `name` attributes is mandatory and specifies
```
rjschwei, line 1293: s/attributes/attribute/

doc/source/image_description/elements.rst
```diff
@@ -1233,6 +1233,78 @@ Used to customize the installation media images created for oem images
...
+the name of the container as it exists in the registry.
+The `container` element has the following optional attributes
+
+fetch_only="true|false"
+  If set to `true` kiwi will only fetch the container but does not
+  setup the systemd unit for actually loading the container into
```
rjschwei, line 1299: s/actually//

schaefi:
> Is there a reason we can't add registry sources as repositories with `oci` type and then be able to reference them using the `alias` attribute for `containers`? Also, why should registry be mandatory here, since you can pull container images with fully qualified paths (e.g. `registry.opensuse.org/git`)? I would prefer we more closely mimic what people do on the commandline here and require fully qualified paths instead.

The reason is that a registry is not a repository in the meaning of kiwi. I asked you to make an example how you would express it. So I believe you had something like this in mind

```xml
<repository type="oci">
    <source path="registry.opensuse.org/home/marcus.schaefer/delta_containers/containers_tw"/>
</repository>

<packages>
    <container name="joe"/>
    <container name="foo"/>
</packages>
```

If the container `foo` should come from another registry how would you reference that ? There is no container package manager that would work on a number of repository entries of type oci and I personally find this completely confusing.

In addition if we intermingle the registry handling with the repositories we have to touch all package manager backends to ignore type="oci".

I'm sorry I see this as two different things. As for the commandline, we would need to add new commandline switches that says --add-registry, --set-registry if need be but also not intermix this with the existing --add-repo, --set-repo options

schaefi:
next we would have to add a backend information (podman/docker/containerd, etc...) to the repository element which is an information that does not exist for repositories but for registries. So we would add attributes to an element which is pointless for package repositories. And vice versa many repository attributes are pointless for registries. That doesn't sound like a good semantic to me

schaefi:
Also there might be additional changes on top of this feature which we currently don't know about. I would like to keep the repository and package management code free from any code change related to registry handling

Conan-Kudo:
So my confusion about the registry section is that it's reasonable to do `podman pull registry.opensuse.org/joe:latest`. In this sense, we don't _need_ a registry tag at all, since it's implied in the name of the container being fetched.

Now if we want to create some kind of construction mechanism instead of just passing it straight to the container tools, then we also need to handle the tag too, which I don't see a semantic for.

And also, how do we handle when the "registry" tag is used multiple times in different snippets?

rjschwei (approved):

schaefi:
> So my confusion about the registry section is that it's reasonable to do `podman pull registry.opensuse.org/joe:latest`. In this sense, we don't _need_ a registry tag at all, since it's implied in the name of the container being fetched.
> 
> Now if we want to create some kind of construction mechanism instead of just passing it straight to the container tools, then we also need to handle the tag too, which I don't see a semantic for.
> 
> And also, how do we handle when the "registry" tag is used multiple times in different snippets?

I can't follow you

```xml
<registry source="registry.opensuse.org">
    <containers backend="podman">
        <container name="joe"/>
        <container name="foo" tag="some"/>
        <container name="bar" fetch_only="true"/>
    </containers>
    <containers backend="docker" path=take/that>
        <container name="some"/>
    </containers>
</registry>
<registry source="registry.fedora.org">
    <containers backend="podman">
        <container name="xxx"/>
    </containers>
</registry>
```

translates to:

* podman pull registry.opensuse.org/joe:latest
* podman pull registry.opensuse.org/foo:some
* only fetching of registry.opensuse.org/bar:latest no loading into any backend
* docker pull registry.opensuse.org/take/that/some:latest
* podman pull registry.fedora.org/xxx:latest

Where do you see issues ? and as I said I consider this very different than the repository handling and don't want to mix it

Conan-Kudo:
What happens when we have two `registry="registry.fedoraproject.org"` tags? What behavior do we have there? Also, are registry and container tags able to be arched like packages? Not all containers are available on all arches. What about profiles?

But about the registry thing, why wouldn't we just do:

```xml
    <containers backend="podman">
        <container name="registry.opensuse.org/joe"/>
        <container name="registry.opensuse.org/foo" tag="some"/>
        <container name="registry.opensuse.org/bar" fetch_only="true"/>
    </containers>
    <containers backend="docker" path="take/that">
        <container name="docker.io/library/some"/>
    </containers>
```

My point is I don't think we need `registry` _per se_.

rjschwei:
> So my confusion about the registry section is that it's reasonable to do `podman pull registry.opensuse.org/joe:latest`. In this sense, we don't _need_ a registry tag at all, since it's implied in the name of the container being fetched.

Correct, but it gets a bit tiresome when there is more than 1 container and one has to type or generate the same registry path over an over again. Having a collector element that specifies the registry from which to pull a bunch of containers is a convenience feature for the user creting the kiwi description.

> 
> ~Now if we want to create some kind of construction mechanism instead of just passing it straight to the container tools, then we also need to handle the tag too, which I don't see a semantic for.~ EDIT: I see we have that already.
> 
> And also, how do we handle when the "registry" tag is used multiple times in different snippets? Is it merged? Does it conflict? What do we do here? 

I do not see a reason why merging logic should be implemented or the should be any conflict.

```
<registry source="registry.opensuse.org">
   <containers backend="podman">
      <container name=foo"/>
  </containers>
</registry>
<registry source="registry.opensuse.org">
   <containers backend="docker">
      <container name=foo"/>
  </containers>
</registry>
"""
Is perfectly fine IMHO. 

> And what about non-OCI things like Incus? And non-OCI Flatpaks?

Should this really be handled with the same mechanism?

**Comment by Conan-Kudo** [comment]
@rjschwei I don't want us to be trapped with awkward semantics just because we didn't think of it like we are with `systemdisk` and some others. I'm thinking about all the other things that matter with supporting "containers" in kiwi that have or will come up.

**Comment by Conan-Kudo** [comment]
And if we want to have the "convenience" of a registry tag, then I'd rather it be inside the `containers` top level tag rather than the `containers` tag being nested inside the `registry` tag.

**Comment by rjschwei** [comment]
> What happens when we have two `registry="registry.fedoraproject.org"` tags? What behavior do we have there? Also, are registry and container tags able to be arched like packages? Not all containers are available on all arches. What about profiles?
> 
> But about the registry thing, why wouldn't we just do:
> 
> ```
>     <containers backend="podman">
>         <container name="registry.opensuse.org/joe"/>
>         <container name="registry.opensuse.org/foo" tag="some"/>
>         <container name="registry.opensuse.org/bar" fetch_only="true"/>
>     </containers>
>     <containers backend="docker" path="take/that">
>         <container name="docker.io/library/some"/>
>     </containers>
> ```
> 
> My point is I don't think we need `registry` _per se_.

Why do you want to type, or make the user type `registry.opensuse.org` possibly 100s of times?

**Comment by Conan-Kudo** [comment]
Now that I think about it more, why wouldn't we just define the registry as an attribute of `containers`?

```xml
    <containers backend="podman" registry="registry.opensuse.org">
        <container name="joe"/>
        <container name="foo" tag="some"/>
        <container name="bar" fetch_only="true"/>
    </containers>
    <containers backend="docker" registry="docker.io/library" path="take/that">
        <container name="some"/>
    </containers>
```

Conan-Kudo:
>> And what about non-OCI things like Incus? And non-OCI Flatpaks?
> Should this really be handled with the same mechanism? 

Yes. We are not in the business of introducing custom things for every single thing. So maximizing reuse of our tags and attributes is beneficial.

rjschwei:
> And if we want to have the "convenience" of a registry tag, then I'd rather it be inside the `containers` top level tag rather than the `containers` tag being nested inside the `registry` tag.

I have no strong preference between

```
<containers backend="podman">
    <registry source="registry.opensuse.org">
        <container name="foo"/>
        <container name="bar"/>
    </registry>
    <registry source="registry.fedoraproject.org">
       <container name="blue"/>
       <container name="green"/>
    </registry>
</containers>
```
which is what I think your proposing or the order in the PR. That said, it does look at bit odd to have the `container` element be a child of `registry` rather than `containers` but that's fine by me.

rjschwei:
> Now that I think about it more, why wouldn't we just define the registry as an attribute of `containers`?
> 
> ```
>     <containers backend="podman" source="registry.opensuse.org">
>         <container name="joe"/>
>         <container name="foo" tag="some"/>
>         <container name="bar" fetch_only="true"/>
>     </containers>
>     <containers backend="docker" source="docker.io/library" path="take/that">
>         <container name="some"/>
>     </containers>
> ```

Shouldn't the path go wit the `container`? Then I can pick up different containers that are in different locations of the same registry.

schaefi:
> Now that I think about it more, why wouldn't we just define the registry as an attribute of `containers`?
> 
> ```
>     <containers backend="podman" source="registry.opensuse.org">
>         <container name="joe"/>
>         <container name="foo" tag="some"/>
>         <container name="bar" fetch_only="true"/>
>     </containers>
>     <containers backend="docker" source="docker.io/library" path="take/that">
>         <container name="some"/>
>     </containers>
> ```

I would be fine with this change

Conan-Kudo:
> Shouldn't the path go wit the container? Then I can pick up different containers that are in different locations of the same registry.

You would be able to define multiple containers sections with different registries, and then set them up as appropriate. This is just more compact than the current representation in the PR.

schaefi:
I'm fine with the more compact definition. I'll change it ... after some rest, today evening or by tomorrow when I'm back at work

Conan-Kudo:
We also need the `containers` to be able to be profile tagged like others too. That way things like Fedora's setup for their kiwi descriptions will be able to work with description fragments. And containers are arch-specific, so individual containers need to be able to be tagged with both profile and arch as needed.

rjschwei:
> > Shouldn't the path go wit the container? Then I can pick up different containers that are in different locations of the same registry.
> 
> You would be able to define multiple containers sections with different registries, and then set them up as appropriate. This is just more compact than the current representation in the PR.

Well I think it's a matter of deciding where we want the repetition. I prefer to have all containers that come from the same registry in one block. Then on visual inspection it makes the cognitive load a lot easier to handle. If the path attribute stays at the `<containers>`  level then I have to repeat the use of the same registry multiple times to pick out containers that exist in the same registry but ind different location. If the `path` attribute is on the `container` element then I might have to repeat the `path` setting but all the containers the come from the same registry are inside the same `<containers>` definition.

schaefi:
> We also need the `containers` to be able to be profile tagged like others too. That way things like Fedora's setup for their kiwi descriptions will be able to work with description fragments. And containers are arch-specific, so individual containers need to be able to be tagged with arch as needed.

yes that and also the arch selector will be available

schaefi:
> Well I think it's a matter of deciding where we want the repetition. I prefer to have all containers that come from the same registry in one block. Then on visual inspection it makes the cognitive load a lot easier to handle. If the path attribute stays at the `<containers>` level then I have to repeat the use of the same registry multiple times to pick out containers that exist in the same registry but ind different location. If the `path` attribute is on the `container` element then I might have to repeat the `path` setting but all the containers the come from the same registry are inside the same `<containers>` definition.

yes correct. So one or the other and we should make a decision:

```xml
<registry source="registry.opensuse.org">
    <containers backend="podman" path="take/that">
        <container name="a"/>
        <container name="b"/>
        <container name="c"/>
    </containers>
    <containers backend="docker" path="take/that">
        <container name="d"/>
    </containers>
</registry>
```

or

```xml
<containers backend="podman" source="registry.opensuse.org" path="take/that">
    <container name="a"/>
    <container name="b"/>
    <container name="c"/>
</containers>
<containers backend="docker" source="registry.opensuse.org" path="take/that">
    <container name="d"/>
</containers>
```

To be honest I have no strong opinion.

currently in place is the first one

Conan-Kudo:
I would prefer the second one.

schaefi:
I like @rjschwei point that the current definition allows to spot easily what containers are taken from a registry source and he will also be the first user of it. @Conan-Kudo also consider there might be more values to set in `<registry>` e.g authentication it might lead to a lot of repetition when we loose this layer. My vote goes to the first one

Conan-Kudo:
I _really_ don't want registry as a top level element. Either it is collapsed inside of `containers` or it's an element underneath it.

schaefi:
> I _really_ don't want registry as a top level element. Either it is collapsed inside of `containers` or it's an element underneath it.

I do not understand why a `<registry>` top level element is an issue whereas a `containers` top level element isn't ?

Conan-Kudo:
> > I _really_ don't want registry as a top level element. Either it is collapsed inside of `containers` or it's an element underneath it.
> 
> I do not understand why a `<registry>` top level element is an issue whereas a `containers` top level element isn't ?

Because registries are specific to particular containers and container things, they aren't a global resource. So containers -> registry -> container matches the expectation.

schaefi:
> Because registries are specific to particular containers and container things, they aren't a global resource. So containers -> registry -> container matches the expectation.

Still I have a hard time why `repository` is good enough as global resource and `registry` isn't but what about the following then

```xml
<containers source="registry.opensuse.org">
    <backend name="podman" path="take/that">
        <container name="a"/>
        <container name="b"/>
        <container name="c"/>
    </backend>
    <backend name="docker" path="take/that">
        <container name="d"/>
    </backend>
</containers>
```

Conan-Kudo:
That would be fine too.

I actually agree about `repository`, but we don't nest package selection inside of `repository` either. The problem here is that the source _matters_ for fetching containers, but it doesn't matter for packages, because containers are always fully qualified references.

doc/source/image_description/elements.rst
schaefi: fixed

doc/source/image_description/elements.rst
schaefi: fixed

schaefi:
> That would be fine too.

hooray :) @rjschwei your vote on this please

Conan-Kudo:
Would the normal rules about merging elements apply here? Say we have portions in different profiles like we do for packages and repositories, would that merge properly?

schaefi:
> Would the normal rules about merging elements apply here? Say we have portions in different profiles like we do for packages and repositories, would that merge properly?

yes I took care for profiles and arch like we do for others in the same way. So you could specify for example:

```xml
<containers source="registry.opensuse.org" profiles="SUSE">
    <backend name="podman" path="take/that">
        <container name="a"/>
     </backend>
    <backend name="docker" path="take/that">
        <container name="b"/>
    </backend>
</containers>

<containers source="registry.fedora.org" profiles="Fedora" arch="x86_64">
    <backend name="podman" path="foo/bar">
        <container name="some"/>
    </backend>
</containers>
```

schaefi:
unless there are any objections I will adapt the PR to the above schema

Conan-Kudo:
I think that works. It should be possible to extend this to OCI-delivered Flatpaks afterward.

Conan-Kudo:
Note that individual `container` lines need to support the `arch` attribute too, since that can be per container.

rjschwei:
> > Well I think it's a matter of deciding where we want the repetition. I prefer to have all containers that come from the same registry in one block. Then on visual inspection it makes the cognitive load a lot easier to handle. If the path attribute stays at the `<containers>` level then I have to repeat the use of the same registry multiple times to pick out containers that exist in the same registry but ind different location. If the `path` attribute is on the `container` element then I might have to repeat the `path` setting but all the containers the come from the same registry are inside the same `<containers>` definition.
> 
> yes correct. So one or the other and we should make a decision:
> 
> ```
> <registry source="registry.opensuse.org">
>     <containers backend="podman" path="take/that">
>         <container name="a"/>
>         <container name="b"/>
>         <container name="c"/>
>     </containers>
>     <containers backend="docker" path="take/that">
>         <container name="d"/>
>     </containers>
> </registry>
> ```
> 
> or
> 
> ```
> <containers backend="podman" source="registry.opensuse.org" path="take/that">
>     <container name="a"/>
>     <container name="b"/>
>     <container name="c"/>
> </containers>
> <containers backend="docker" source="registry.opensuse.org" path="take/that">
>     <container name="d"/>
> </containers>
> ```
> 
> To be honest I have no strong opinion.
> 
> currently in place is the first one

Or option 3, which is what I was trying to express:

```
<containers backend="podman" source="registry.opensuse.org">
     <container name="a" path="take/that"/>
     <container name="b" path="take/that"/>
    <container name="c" path="take/andthisoverhere"/>
</containers>
```
This keeps all the containers used from one registry for the same backend in one block, which I, as a reviewer of many image descriptions prefer.

rjschwei:
> > That would be fine too.
> 
> hooray :) @rjschwei your vote on this please

I vote for option 3 ;)

```
<containers backend="podman" source="registry.opensuse.org">
     <container name="a" path="take/that"/>
     <container name="b" path="take/that"/>
    <container name="c" path="take/andthisoverhere"/>
</containers>
```

 but if there is a strong feeling that the path should stay with the location where the registry source is set instead of being on the container itself then I can live with that.

Conan-Kudo:
I'm fine with this option too.

schaefi:
> Note that individual `container` lines need to support the `arch` attribute too, since that can be per container.

like with the packages, yes I will add that

travier:
Is there any specific reason that we are not using the URL format from https://github.com/containers/image/blob/main/docs/containers-transports.5.md for referencing the container images? Not all containers come from registries so making this a required element feels weird.

What's the reason behind the podman/docker option? Both should handle the same container images fine.

travier:
Overall, I like this option. I think I should be able to use it to make things simpler for the ostree/bootable container path.

travier:
Thinking about this more, I think we should not focus on how the import is made but more on how the container image is stored in the image.

The idea would be to specify where the image should go and under which format and then it would be up to the user doing the image to handle how to import it. This would let us pre-setup the containers for the backends that support it.

Example:

```
<containers>
     <!-- Pull the image from a registry and store it as an OCI archive at the given path -->
     <container source="docker://docker-reference" dest="oci:path[:reference]"/>
     <!-- Pull the image directly from the local container storage and store it in the image destination storage -->
     <container source="containers-storage:[[storage-specifier]]{image-id|docker-reference[@image-id]}" dest="containers-storage:[[storage-specifier]]{image-id|docker-reference[@image-id]}"/>
     <!-- Pull the image from a local OCI archive and store it at the given path -->
     <container source="oci:path[:reference]" dest="oci:path[:reference]"/>
</containers>

```

schaefi:
In reply to @travier idea

* a direct pull of the container into the local container storage backend is possible only in a very limited way when building the image because the storage backend is offline. For podman I could make this work only with the vfs storage backend driver and in many other cases I stumbled over issues like: conflicts with overlay because podman thinks the real-root is not the system, conflicts with podman which switches to fuse-overlayfs when it should use overlay, conflicts with docker which requires dockerd to run for importing containers, conflicts with nested workloads when you want to build an image in a container and you run nested imports, etc etc etc.... the list is endless. That's the reason why we don't support the direct import and offered a two stage system where we store the container archive locally and load it into the container storage backend on first boot via a systemd service

* With the support for containers that we store intermediate in the image we made two static settings:
   1. The transport protocol to store the container image is `oci-archive` and connected to the configured `backend` name. I don't see a benefit to let a user make this decision because this is not data the user handles, we handle it for the purpose of loading the container to the local registry. So the tooling to use and the transport protocol is encoded in kiwi according to the selected backend. We currently support docker and podman and for both we copy docker:// to oci-archive:target:source
   2. The location to store those images is set to `/var/tmp/kiwi_containers`. It's intentionally not configurable at the moment as it is considered a temporary space to store the container archives until they get loaded into the container storage backend. The systemd service which we write deletes the container archive as soon as it is loaded into the local storage unless the `fetch_only` attribute is set

We can have a conversation if it would be beneficial to add additional attributes to influence some of the fixed settings, but I would do this on real demand in followup pull requests.

Thanks

schaefi:
> Is there any specific reason that we are not using the URL format from https://github.com/containers/image/blob/main/docs/containers-transports.5.md for referencing the container images? Not all containers come from registries so making this a required element feels weird.

I can't follow you on this. We use the official supported transport URL. The final URL is constructed out of the
`source`, `name`, `tag` and `path` information. `source` specifies an endpoint and how that endpoint should be reached is
determined by the specified backend 

 > What's the reason behind the podman/docker option? Both should handle the same container images fine. Edit: Ah, it's to specify how to generate the systemd unit that will load the container into the storage? It feels like that should be part of something else and not managed by kiwi as this is modifying the content of the final system.

With the backend name we determine several settings, how to reach the endpoint, which tools to use, how to load the
image. The backend name gives us also the opportunity to evaluate the given container data also for other containers
types, e.g flatpacks, or whatever else comes in the future

> It feels like that should be part of something else and not managed by kiwi as this is modifying the content of the final system.

I'm afraid as I tried to explain that a direct import while building is very limited and is only causing trouble I only saw the way via a one-shot firstboot service on boot. If you can be more explicit what you have in mind by `something else` we could adapt the concept :)

kiwi/system/setup.py
```diff
@@ -87,6 +88,48 @@ def __init__(self, xml_state: XMLState, root_dir: str):
...
+
+        if container_files_to_load:
+            log.info('--> Setup kiwi_containers.service import unit')
+            service = BuilderTemplateSystemdUnit()
+            unit_template = service.get_container_import_template(
+                container_files_to_load, container_execs_to_load,
+                list(after_services)
+            )
+            unit = unit_template.substitute()
+            unit_file = '{0}/usr/lib/systemd/system/{1}.service'.format(
```
Conan-Kudo, line 121: Why are we writing this in `/usr` instead of `/etc`? Even if we're writing into `/usr`, we should put it in `/usr/local` rather than `/usr`, since this is not packaged content.

Conan-Kudo (changes requested):

kiwi/system/setup.py
schaefi: good catch, yes this needs to change

schaefi: fixed

travier:
> I'm afraid as I tried to explain that a direct import while building is very limited and is only causing trouble I only saw the way via a one-shot firstboot service on boot. If you can be more explicit what you have in mind by `something else` we could adapt the concept :)

What I meant by `something else` is that the user would write this unit and control its behavior. But I can understand how that could fit into the current behavior. Are there any other occurrences in Kiwi where it directly writes systemd units to the system?

> > Is there any specific reason that we are not using the URL format from [containers/image@`main`/docs/containers-transports.5.md](https://github.com/containers/image/blob/main/docs/containers-transports.5.md) for referencing the container images? Not all containers come from registries so making this a required element feels weird.
> 
> I can't follow you on this. We use the official supported transport URL. The final URL is constructed out of the `source`, `name`, `tag` and `path` information. `source` specifies an endpoint and how that endpoint should be reached is determined by the specified backend
>
> We currently support docker and podman and for both we copy docker:// to oci-archive:target:source

So if I understand correctly, this will always generate a "docker://" URL and copy to an OCI archive.

If you can not specify that you want to take it from the current local storage for example then this either means pulling it every time (to make sure it is the latest version) or caching it locally (and then needing a way to force a refresh).

This is part of why I suggested that we don't split those and use the full transport URL (at lest for the source) so that the user can decide and manage that on their own.

travier:
Note that I'm only sharing thoughts here and you should feel free to ignore anything I say. I don't think we'll be able to use that directly for the bootable container cases in this current form.

schaefi:
>What I meant by something else is that the user would write this unit and control its behavior. But I can understand how that could fit into the current behavior. Are there any other occurrences in Kiwi where it directly writes systemd units to the system?

Nope this is the first time that we do this and I also think it smells a bit, but as I said I don't see another way except for letting the user alone with it. btw if you specify all containers with `fetch_only="true"` we only fetch the containers and do not create any systemd unit

schaefi:
> So if I understand correctly, this will always generate a "docker://" URL and copy to an OCI archive.

If the backend is set to either `backend="podman"` or `backend="docker"` yes

schaefi:
> If you can not specify that you want to take it from the current local storage for example then this either means pulling it every time (to make sure it is the latest version) or caching it locally (and then needing a way to force a refresh).

We currently fetch the container at build time of the image and store it there. On first boot it gets loaded. Assuming you build the image on a Monday and deploy it on a Friday, the container could have been updated in the registry yes, but the same applies to packages taken from a a repository.

I was not quite clear what you mean by `specify that you want to take it from the current local storage` ? which local storage ? the one of the build host you build the image on ?

schaefi:
> To clarify, I'm only sharing thoughts here and you should feel free to ignore anything I say :). I don't think we'll be able to use that directly for the bootable container cases in this current form.

I take this serious as it covers use cases which are the important part. A feature of no use is useless right :) But I agree this is not working towards the `bootc` idea. With the `containers` element we want to make containers available to the local storage of the image such that you can pre-populate a container storage backend with container images fetched at build time of the image from a trusted source

Conan-Kudo:
We still need it even for `bootc` because it's entirely reasonable to expect a base image seeded with any and all content in an offline context. From my perspective, `bootc` is just saving things as an image layer that's pushed and pulled through OCI tooling. What is in the image layer is still much more like what we do with any other appliance image build.

Conan-Kudo:
But I did not think about the fact that there's multiple input source types supported by podman and docker. Since the `source=` attribute is optional, we should be able to allow fully qualified references when it's not set. But if it's used, we probably should also add something to specify the protocol for the source. Or just say it needs to be specified in `source`?

doc/source/image_description/elements.rst
```diff
@@ -1233,6 +1233,61 @@ Used to customize the installation media images created for oem images
...
+   <containers source="registry.opensuse.org" backend="podman">
+       <container name="some"/>
+   </containers>
+
+The optional containers element specifies the location of one ore
+more containers on a registry `source` server. {kiwi} will take
+this information and fetch the containers as OCI archives to
+the image. On first boot those container archives will be loaded
+into the local container backend store for the selected
+backend and the archive files gets deleted.
```
rjschwei, line 1255: s/gets/get/

doc/source/image_description/elements.rst
```diff
@@ -1233,6 +1233,61 @@ Used to customize the installation media images created for oem images
...
+
+Supported `backend` values as of today are `docker` and `podman`.
```
rjschwei, line 1257: s/as of today//

doc/source/image_description/elements.rst
rjschwei: GitHub diff still shows "attributes"

doc/source/image_description/elements.rst
```diff
@@ -1233,6 +1233,61 @@ Used to customize the installation media images created for oem images
...
+  If set to `true` kiwi will only fetch the container but does not
+  setup the systemd unit for loading the container into
+  the local registry. In this mode the container archive file stays
+  in the system and can be handled in a custom way. By default
+  `fetch_only` is set to `false`.
+
+tag="tagname"
+  Specifies the container tag to fetch. If not set the tag name
+  defaults to `latest`
+
```
rjschwei, line 1290: `arch="archname"` missing

doc/source/image_description/elements.rst
```diff
@@ -1233,6 +1233,61 @@ Used to customize the installation media images created for oem images
...
+more containers on a registry `source` server. {kiwi} will take
+this information and fetch the containers as OCI archives to
+the image. On first boot those container archives will be loaded
+into the local container backend store for the selected
+backend and the archive files gets deleted.
+
+Supported `backend` values as of today are `docker` and `podman`.
+The `backend` attribute is mandatory and specifies for which
+container backend the image should be available in the system.
+
```
rjschwei, line 1260: Something about arch and profile setting missing.

kiwi/system/setup.py
```diff
@@ -87,6 +88,48 @@ def __init__(self, xml_state: XMLState, root_dir: str):
         self._preferences_lookup()
         self._oemconfig_lookup()
 
+    def setup_registry_import(self) -> None:
+        """
+        Fetch container(s) and activate systemd unit to load
+        fetched containers during boot
```
rjschwei, line 94: s/fetched containers/containers from local tarball/

schaefi:
> But I did not think about the fact that there's multiple input source types supported by podman and docker. Since the `source=` attribute is optional, we should be able to allow fully qualified references when it's not set. But if it's used, we probably should also add something to specify the protocol for the source. Or just say it needs to be specified in `source`?

`source` is not optional, you have to give an endpoint to contact. How you specify this endpoint is up to you, I have not made any restrictions there. So it can be `source="obs:..."` or `source="https://..."` or `source="registry.opensuse.org"` or `source="local:..."` ... point is if the source is understood depends on the given backend name. This gives us some flexibility to add more

travier:
> I was not quite clear what you mean by `specify that you want to take it from the current local storage` ? which local storage ? the one of the build host you build the image on ?

Yes, that would be the build host docker or podman image store. If I know that I'm going to build X images with a specific container in it, instead of pulling it X times into each of the images, I would pull it once in my local storage and then point the XML config to it instead.

rjschwei (commented):
A few more wording and description suggestions. LGTM

schaefi:
> Yes, that would be the build host's docker or podman image store. If I know that I'm going to build X images with a specific container in it, instead of pulling it X times into each of the images, I would pull it once in my local storage and then point the XML config to it instead.

ok this means your local system needs a registry server e.g the CNCF distribution registry configured as a proxy. You would pull your container once and in your image build you could reference via `source="localhost:5000"` as an example for a registry server on port 5000. That would work this way without changes on the kiwi side once this PR is merged
 
> Or that could be for images that are not pushed to a registry at all.

right which means `fetch_only="true"` 

So I guess this use case would be covered. There is just not much influence where in the system the container would be stored and the current selection of backends would allow for the format to be oci-archive only. But as I said we can add more granular settings later when required

travier:
> > Yes, that would be the build host's docker or podman image store. If I know that I'm going to build X images with a specific container in it, instead of pulling it X times into each of the images, I would pull it once in my local storage and then point the XML config to it instead.
> 
> ok this means your local system needs a registry server e.g the CNCF distribution registry configured as a proxy. You would pull your container once and in your image build you could reference via `source="localhost:5000"` as an example for a registry server on port 5000. That would work this way without changes on the kiwi side once this PR is merged

Sure, that could work, but it's super inefficient. If I already have the OCI archive on the build host, copying it into the rootfs of the destination image is "free" for filesystems that do copy-on-write for example. Note that all those transports are supported by skopeo for example so it's not code in kiwi.

> > Or that could be for images that are not pushed to a registry at all.
> 
> right which means `fetch_only="true"`

I was more thinking about images that you've just built locally and are on the build host but not a in registry. It's independent of how you would import them in the final system.

schaefi:
> I was more thinking about images that you've just built locally and are on the build host but not a in registry. It's independent of how you would import them in the final system.

sure and to add this data into an image you don't need anything from what we are discussing here. I'm not sure if we are following the same line of thoughts. The feature provided in this PR is all about making container(s) from registrie(s) available in the local storage of an image such that a backend system be it podman, docker, containerd, etc etc can consume it without the need to pull them in first. If you have everything available on your local machine and just want to share data with an image build you can do this in various ways but it will also only work on this machine and nowhere else. Thus there is no reason to specify a source definition in a kiwi description, you could just go with a kiwi script that runs some cp, skopeo or podman image save or some other command to get the data into the image.

Maybe we can take this topic offline ? It confuses me to be honest and I'm not sure if it helps with the review of the PR to be merged

For me the most important part is, if we all agree that the current description schema serves the purpose for the feature for which @rjschwei has opened a request for and that it can be used/adapted to other scenarios that @Conan-Kudo and @travier have in mind ? I mean it's not all that of a rocket science, we fetch containers and load them and the description elements to describe this should be flexible enough to fetch from everywhere and to load into anything, which I believe is possible the way we describe it.

Thoughts ?

doc/source/image_description/elements.rst
schaefi: fixed

doc/source/image_description/elements.rst
schaefi: fixed

doc/source/image_description/elements.rst
schaefi: added

doc/source/image_description/elements.rst
schaefi: added with reference to profile chapter

kiwi/system/setup.py
schaefi: fixed, I use oci-archive instead of local tarball

rjschwei:
The implementation in this PR works for me and covers #2663 as far as I am concerned. We have some immediate use cases and a couple in the not too distant future, as such getting this over the finish line soon would be great. Thanks @schaefi for the effort an dthe quick implementation. @Conan-Kudo any other concerns?

Conan-Kudo (approved):

PR #2665, opened 2024-10-16
Turn DiskFormat into an ordinary class

dcermak:
- it does not need to be an abstract base class
- use f-strings where applicable instead of `format()`
- change return type of `_custom_args_for_format` from list to tuple


schaefi (approved):
ok we never registered subclasses from the abstract base, thus this concept can let go

PR #2672, opened 2024-10-30
Add pytest-container as optional dependency

schaefi:
The pyproject.toml listed pytest-container as dependency but it is used only to run the container based integration tests for the shell helper methods. For building the package this dependency should not be pulled in

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,6 +66,7 @@ xmltodict = { version = ">=0.12.0", optional = true }
 
 [tool.poetry.extras]
 markup = [ "anymarkup-core", "xmltodict" ]
+integration_tests = ["pytest-container"]
 
 [tool.poetry.plugins]
 [tool.poetry.plugins."kiwi.tasks"]
@@ -87,7 +88,8 @@ kiwi-ng = "kiwi.kiwi:main"
 pytest = ">=6.2.0"
 pytest-cov = "*"
 pytest-xdist = "*"
-pytest-container = "*"
+# Optional dependencies for shell functions integration tests
+pytest-container = {version = "*", optional = true}
 # type checking
 mypy = ">=0.971"
 types-requests = "*"


pyproject.toml
```diff
@@ -63,9 +63,11 @@ simplejson = ">=3.17.0"
 # Optional dependencies for markup extra
 anymarkup-core = { version = ">=0.8.0", optional = true }
 xmltodict = { version = ">=0.12.0", optional = true }
+pytest-container = {version = "*", optional = true}
```
dcermak, line 66: maybe add another comment indicating that this is not a dependency for "markup extra"?

schaefi: yes makes sense and fixed

pyproject.toml
```diff
@@ -63,9 +63,12 @@ simplejson = ">=3.17.0"
 # Optional dependencies for markup extra
 anymarkup-core = { version = ">=0.8.0", optional = true }
 xmltodict = { version = ">=0.12.0", optional = true }
+# Optional dependencies for shell functions integration tests
+pytest-container = {version = "*", optional = true}
```
Conan-Kudo, line 67: This is not how to do it. You need to define a new dependency group and move it to that.

Conan-Kudo (changes requested):

pyproject.toml
schaefi: ok moved it into a new dep group

schaefi: data.group.test must not contain {'optional_dependencies'} properties

pyproject.toml
```diff
@@ -66,6 +66,7 @@ xmltodict = { version = ">=0.12.0", optional = true }
 
 [tool.poetry.extras]
 markup = [ "anymarkup-core", "xmltodict" ]
+testing = ["pytest-container"]
```
Conan-Kudo, line 69: Drop this.

pyproject.toml
```diff
@@ -87,14 +88,17 @@ kiwi-ng = "kiwi.kiwi:main"
...
 # type checking
 mypy = ">=0.971"
 types-requests = "*"
 types-PyYAML = "*"
 types-mock = "*"
 types-simplejson = "*"
 
+[tool.poetry.group.test.optional_dependencies]
+# Optional dependencies for shell functions integration tests
+pytest-container = {version = "*", optional = true}
```
Conan-Kudo, line 100:
```toml
[tool.poetry.group.integration_test]
[tool.poetry.group.integration_test.dependencies]
pytest-container = "*"
```

Conan-Kudo (changes requested):

pyproject.toml
schaefi: ok done, thanks

pyproject.toml
```diff
@@ -87,14 +88,18 @@ kiwi-ng = "kiwi.kiwi:main"
...
 types-requests = "*"
 types-PyYAML = "*"
 types-mock = "*"
 types-simplejson = "*"
 
+[tool.poetry.group.integration_test]
+[tool.poetry.group.integration_test.dependencies]
+# Optional dependencies for shell functions integration tests
+pytest-container = {version = "*", optional = true}
+
```
Conan-Kudo, line 102: On second thought, put this back in the regular test dependencies, like before.

pyproject.toml
Conan-Kudo, line 69: Let's change this to `integration_tests`

Conan-Kudo (changes requested):

pyproject.toml
schaefi: done

pyproject.toml
Conan-Kudo, line 67: I meant that this should still be in the `tool.poetry.group.test.dependencies` section.

schaefi: done

Conan-Kudo (approved):

glaubitz:
I just gave it a try with 10.1.17 and it still tries to import `pytest_container`:

```
I: pybuild base:311: cd /<<PKGBUILDDIR>>/.pybuild/cpython3_3.12/build; python3.12 -m pytest test
============================= test session starts ==============================
platform linux -- Python 3.12.7, pytest-8.3.3, pluggy-1.5.0
rootdir: /<<PKGBUILDDIR>>
configfile: setup.cfg
plugins: testinfra-10.1.1, typeguard-4.4.0
collected 1721 items / 1 error

==================================== ERRORS ====================================
__________ ERROR collecting .pybuild/cpython3_3.12/build/test/scripts __________
/usr/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1387: in _gcd_import
    ???
<frozen importlib._bootstrap>:1360: in _find_and_load
    ???
<frozen importlib._bootstrap>:1331: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:935: in _load_unlocked
    ???
/usr/lib/python3/dist-packages/_pytest/assertion/rewrite.py:184: in exec_module
    exec(co, module.__dict__)
test/scripts/conftest.py:1: in <module>
    from pytest_container import auto_container_parametrize, DerivedContainer
E   ModuleNotFoundError: No module named 'pytest_container'
=========================== short test summary info ============================
ERROR test/scripts - ModuleNotFoundError: No module named 'pytest_container'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 2.66s ===============================
```

glaubitz:
Could you change the testsuite such that it can also be run without having to use containers?

glaubitz:
Is there a way to exclude `conftest.py` and `test_baseVagrantSetup.py`?

I tried the following but that didn't help:

```
export PYBUILD_TEST_ARGS=-k 'not conftest and not test_baseVagrantSetup'
```

schaefi:
@glaubitz I tested the setup of the venv via poetry like this

```
git clone https://github.com/OSInside/kiwi.git
poetry install
poetry shell
pip list

... no pytest_container

```

So the code added to handle this requirement in poetry works. However, I believe this is irrelevant for the way how you package it. So how do you create the venv in the packaging ? with tox ? If yes the place to look at is the `tox.ini` file which does the following

```
deps =
    -r.virtualenv.dev-requirements.txt

```

In that `virtualenv.dev-requirements.txt` we still have the entry

```
# for scripts testing
pytest-container
```

But this entry exists there for a long time and I wonder why it is an issue now ?

@dcermak Can we delete the entry from .virtualenv.dev-requirements.txt or will this cause any negative side effects. It's the only place that I can imagine to be the problem for Adrian

Thanks

glaubitz:
I finally figured it out. I will upload an updated kiwi package tomorrow to Debian unstable.

PR #2673, opened 2024-11-03
Remove tox dependency

schaefi:
tox was used as sort of a make target to run unit
tests and more in a venv. However, since we switched everything to poetry it's no longer needed to let tox create the python virtual environments. This commit moves the tox targets into the Makefile and adapts the github workflow files accordingly


Conan-Kudo:
Why'd we drop the CentOS tests? And what happened to the Fedora tests?

schaefi:
> Why'd we drop the CentOS tests? And what happened to the Fedora tests?

I reduced the scope of the custom script tests for the following reasons:

* running them in docker is unstable because every now and then the tagging fails, running the same tests in one container engine is more than enough, podman it is
* The script code runs for test_suseSetupProduct.py are suse specific and it doesn't make sense to run them in non suse containers
* The test_baseVagrantSetup.py tests the vagrant setup functions which are using core tools like cp, useradd, and so one which are expected to work in the same way on any linux distribution. It doesn't make sense to run this shell code more than once.
* I dropped all targets except for TW and Leap because of the suse specific tests. There are no fedora or any other linux specific tests in this test/scripts/ implementation

Conan-Kudo:
> I dropped all targets except for TW and Leap because of the suse specific product name tests. There are no fedora or any other linux specific tests in this test/scripts/ implementation

Can we add at least one Fedora or CentOS one to make sure the non-product stuff works?

schaefi:
> Can we add at least one Fedora or CentOS one to make sure the non-product stuff works?

sure, can you tell me a stable registry URL to fetch a Fedora container from ? I'll add that then

Conan-Kudo:
`registry.fedoraproject.org/fedora:latest`

Conan-Kudo (approved):

PR #2674, opened 2024-11-06
Added containers integration with OBS

schaefi:
When building in the Open Build Service (OBS) there is no way to create outgoing connections from the build workers. To allow the <containers> section to fetch containers from the SUSE registry we need to apply the OBS uri translation in the same way as it was done for the derived_container feature. The actual OCI container image is expected to be provided by the obs backend on the worker. Along with this commit also an integration test named test-image-disk-containers is provided.
This Fixes jira#OBS-351

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-disk-containers/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-disk-containers/appliance.kiwi
@@ -1,0 +1,86 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<image schemaversion="7.5" name="kiwi-test-image-disk-containers">
+    <description type="system">
+        <author>Marcus Schäfer</author>
+        <contact>marcus.schaefer@suse.com</contact>
+        <specification>Disk test build with container import</specification>
+    </description>
+    <preferences>
+        <version>1.42.1</version>
+        <packagemanager>zypper</packagemanager>
+        <locale>en_US</locale>
+        <keytable>us</keytable>
+        <timezone>Europe/Berlin</timezone>
+        <rpm-excludedocs>true</rpm-excludedocs>
+        <rpm-check-signatures>false</rpm-check-signatures>
+        <bootsplash-theme>breeze</bootsplash-theme>
+        <bootloader-theme>openSUSE</bootloader-theme>
+        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="efi" installiso="true" bootpartition="false" btrfs_root_is_snapshot="true" installboot="install">
+            <bootloader name="grub2" console="serial" timeout="10"/>
+            <oemconfig>
+                <oem-unattended>true</oem-unattended>
+                <oem-swapsize>1024</oem-swapsize>
+                <oem-multipath-scan>false</oem-multipath-scan>
+            </oemconfig>
+            <systemdisk>
+                <volume name="home" quota="5G"/>
+            </systemdisk>
+        </type>
+    </preferences>
+    <containers source="registry.opensuse.org" backend="podman">
+        <container name="tumbleweed" tag="latest" path="opensuse"/>
+    </containers>
+    <users>
+        <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root" groups="root"/>
+    </users>
+    <repository type="rpm-md">
+        <source path="obsrepositories:/"/>
+    </repository>
+    <packages type="image">
+        <package name="patterns-openSUSE-base"/>
+        <package name="skopeo"/>
+        <package name="podman"/>
+        <package name="procps"/>
+        <package name="systemd"/>
+        <package name="plymouth-theme-breeze"/>
+        <package name="plymouth-plugin-script"/>
+        <package name="grub2-branding-openSUSE"/>
+        <package name="iputils"/>
+        <package name="vim"/>
+        <package name="grub2"/>
+        <package name="grub2-x86_64-efi" arch="x86_64"/>
+        <package name="grub2-i386-pc"/>
+        <package name="lvm2"/>
+        <package name="plymouth"/>
+        <package name="fontconfig"/>
+        <package name="fonts-config"/>
+        <package name="tar"/>
+        <package name="parted"/>
+        <package name="openssh"/>
+        <package name="iproute2"/>
+        <package name="less"/>
+        <package name="bash-completion"/>
+        <package name="bind-utils"/>
+        <package name="dhcp-client"/>
+        <package name="which"/>
+        <package name="kernel-default"/>
+        <package name="timezone"/>
+        <package name="dracut-kiwi-oem-repart"/>
+        <package name="dracut-kiwi-oem-dump"/>
+    </packages>
+    <packages type="bootstrap">
+        <package name="gawk"/>
+        <package name="grep"/>
+        <package name="gzip"/>
+        <package name="udev"/>
+        <package name="xz"/>
+        <package name="shadow"/>
+        <package name="filesystem"/>
+        <package name="glibc-locale"/>
+        <package name="cracklib-dict-full"/>
+        <package name="ca-certificates"/>
+        <package name="ca-certificates-mozilla"/>
+        <package name="openSUSE-release"/>
+    </packages>
+</image>
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -101,9 +101,7 @@ class SystemSetup:
             pathlib.Path(f'{self.root_dir}/{defaults.LOCAL_CONTAINERS}').mkdir(
                 parents=True, exist_ok=True
             )
-            Command.run(
-                ['chroot', self.root_dir] + container.fetch_command
-            )
+            container.fetch_command(self.root_dir)
             if container.load_command:
                 container_files_to_load.append(container.container_file)
                 container_execs_to_load.append(container.load_command)
--- a/kiwi/xml_state.py
+++ b/kiwi/xml_state.py
@@ -17,7 +17,7 @@
 #
 import os
 from typing import (
-    List, Optional, Any, Dict, NamedTuple
+    List, Optional, Any, Dict, NamedTuple, Callable
 )
 import re
 import logging
@@ -32,6 +32,7 @@ from kiwi.storage.disk import ptable_entry_type
 from kiwi.system.uri import Uri
 from kiwi.defaults import Defaults
 from kiwi.utils.size import StringToSize
+from kiwi.command import Command
 
 from kiwi.exceptions import (
     KiwiProfileNotFound,
@@ -90,7 +91,7 @@ class ContainerT(NamedTuple):
     backend: str
     container_file: str
     fetch_only: bool
-    fetch_command: List[str]
+    fetch_command: Callable
     load_command: List[str]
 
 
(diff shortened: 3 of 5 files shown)


schaefi:
For this PR to be merged an implementation in the OBS backend is required too. Once the backend code supports providing the requested containers on the worker the following integration test will start to succeed

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-disk-containers

Currently this ends with the expected error

```
[   61s] [ ERROR   ]: 09:21:55 | KiwiCommandError: chroot: stderr: cp: cannot stat '/usr/src/packages/SOURCES/repos/openSUSE:Containers:Tumbleweed/containers/opensuse/tumbleweed:latest': No such file or directory
```

from the containers definition in the integration test as

```xml
<containers source="obs://openSUSE:Containers:Tumbleweed" backend="podman">
    <container name="tumbleweed" tag="latest" path="/containers/opensuse"/>
</containers>
```

rjschwei (approved):

mlschroe:
Can you please change your testcase so that the repository is part of the obs:// url? i.e.
```
<containers source="obs://openSUSE:Containers:Tumbleweed/containers" backend="podman">
  <container name="tumbleweed" tag="latest" path="opensuse"/>
</containers>
```

I also think it would be a good idea to have a suffix for the containers, i.e.:
```
                            fetch_command = [
                                'cp', Uri(container_endpoint).translate() + ".ociarchive",
                                container_file_name
                            ]
```

mlschroe:
Also, please add `repo_type='container'` to your Uri() call so that the path uses `containers` instead of `repos`.

mlschroe:
Somewhat unrelated: can you shed some light on commit 9487a189ed3a7a1cd8b91abd7daed18a8e78b0bc? Why was this done? OBS never needed this, as it uses the `--set-container-derived-from` to change the url point to a local file.

schaefi:
@mlschroe 

> Can you please change your testcase so that the repository is part of the obs:// url?

done

> I also think it would be a good idea to have a suffix for the containers,

ok, done. kiwi looks up including the suggested `.ociarchive` suffix

> Also, please add repo_type='container' to your Uri() call so that the path uses containers instead of repos

yeah good catch, thanks and fixed

schaefi:
> Somewhat unrelated: can you shed some light on commit [9487a18](https://github.com/OSInside/kiwi/commit/9487a189ed3a7a1cd8b91abd7daed18a8e78b0bc)? Why was this done? OBS never needed this, as it uses the `--set-container-derived-from` to change the url point to a local file.

If I remember correctly this was the first implementation and we agreed on this naming schema. Then I believe Adrian suggested that he wants to add the full path name to the derived container on the command line and we added the option. Up to then OBS  passes path names to this option but they do not correlate with the naming schema that we implemented in the first place. As you said it doesn't hurt as long as the option is used. I realize that the naming convention we implemented if `--set-container-derived-from` is `not` set will lead to a path that OBS doesn't serve the containers  to. We might want to change this such that the code is at least compatible with the current values passed to `--set-container-derived-from`, but in an extra PR

schaefi:
@mlschroe 

JFI: The test after the proposed changes now reports as:

```
[   60s] [ ERROR   ]: 15:23:58 | KiwiCommandError: chroot: stderr: cp: cannot stat '/usr/src/packages/SOURCES/containers/openSUSE:Containers:Tumbleweed/containers/opensuse/tumbleweed:latest.ociarchive': No such file or directory
```

mlschroe:
I ran into a problem we have: when the customer runs the created image we want the containers to be available with the right names, i.e. `registry.suse.com/some/thing:latest`. I.e. we can't have `obs:` or `obsrepositories:` as `source`, but need the correct registry name.
We can still use `obsrepositories:` internally in OBS to get the container via download on demand.

Would that work for you?

schaefi:
> I ran into a problem we have: when the customer runs the created image we want the containers to be available with the right names, i.e. `registry.suse.com/some/thing:latest`

Yes I also want this and ensure it with the `skopeo copy` in the non OBS case.

>. I.e. we can't have `obs:` or `obsrepositories:` as `source`, but need the correct registry name. We can still use `obsrepositories:` internally in OBS to get the container via download on demand.
> Would that work for you ?

I'm not sure if I can follow you on the issue. The information about the repository is usually stored in the container metadata itself. The name of the file should be immaterial. If we fetch the container (in the non OBS case) it is done like in the following example:

```
skopeo copy \
    docker://registry.opensuse.org/home/marcus.schaefer/delta_containers/images_leap/aws-cli-transparent:latest \
    oci-archive:foo.ociarchive:registry.opensuse.org/home/marcus.schaefer/delta_containers/images_leap/aws-cli-transparent:latest
```

The file in this example is named `foo.ociarchive` and actually doesn't matter. On first boot of the system the locally stored container files gets imported via

```
podman load -i ~ms/foo.ociarchive
```

And because the container metadata contains the correct repository information it will list correctly as

```
podman images

REPOSITORY                                                                                   TAG         IMAGE ID      CREATED      SIZE
registry.opensuse.org/home/marcus.schaefer/delta_containers/images_leap/aws-cli-transparent  latest      cb1de5660f28  5 hours ago  335 MB
```

So when OBS provides us the container file it actually doesn't matter how it's named or which source we have to specify to find it on the worker as long as the metadata inside of the container file is correct

> Would that work for you ?

For me it doesn't matter how we have to specify the source, path, tag attribute to meet the needs of the OBS backend. I choose the OBS vs non OBS path via the Defaults.is_buildservice_worker() method and that detection is based on the presence of a `/.buildenv` file. So yes you can just tell me how you would like us to specify the data but I would prefer if the provided container files just comes with correct metadata.

Does this makes sense ?

mlschroe:
My point is that your `container_endpoint` does not work if the source contains a url schema. Basically you will then do
```
  skopeo copy docker:obs:/... obs:/prj/repo/xxx/xx
```
which does not work.

mlschroe:
The question is if we want to make the `obs:` and `obsrepository:` schema work for the source or not. For now, we could also just insist on the source being a registry hostname. You would need to change the code to:
```
                        if Defaults.is_buildservice_worker():
                            container_uri = Uri('obsrepositories/:{0}'.format(container_endpoint), 'container')
```

mlschroe:
But it can't do that because it does not have the information. What is the use case of all this? My impression was that this feature was added to include containers from registry.suse.com in an appliance. This means, the final tag must be `registry.suse.com/yada/yada`. I don't really get the point of directly referencing a project here with `obs:/`.

Implementation wise I also do not have the origin information of the container at build time, so it is not easy to put the container in the correct directory with the current implementation. It's much easier to use the desired end tag as file name, i.e. `registry.suse.com/yada/yada` instead of `openSUSE:Containers:Tumbleweed/containers/yada/yada`.

mlschroe:
(My suggestion for making obs:/ and obsrepositories:/ work is to not use the source for the final tag, i.e. your example would result in the container getting tagged `opensuse/tumbleweed:latest`. This is consistent with how OBS internally handles containers, e.g. when expanding the `derived_from` element.)

schaefi:
@mlschroe I'm confused ??

Let me first answer your question

>  My impression was that this feature was added to include containers from registry.suse.com in an appliance.

Not exclusively from registry.suse.com. From any registry, and of course when building in obs only from the registries that can be accessed in the scope of SUSE.

For more clarity it is maybe helpful when we compare the two situations for building outside of OBS vs inside of OBS. If I build the example image outside of OBS the specification in kiwi is simple:

```xml
<containers source="registry.opensuse.org" backend="podman">
    <container name="tumbleweed" tag="latest" path="opensuse"/>
</containers>
```

Now as we all know inside of OBS we are not allowed to access registry.opensuse.org. Consequently you told me the OBS backend will provide us with the container and we will find it at a certain place on the worker. I was under the assumption that it should be done in a consistent way like it is already implemented with the derived_container attribute. From the conversation it seems that this is not the case. The information I now need is:

* how should the above endpoint be expressed for the obs use case ?
* how should the expression that you can live with be translated into a local path such that we can find the container at build time ?

As we are turning a bit in circles maybe you can just paste the XML you have in mind because I'm confused :)

>But it can't do that because it does not have the information

So you are saying it's not possible that the backend knows that the project location `obs://openSUSE:Containers:Tumbleweed/containers/opensuse/tumbleweed#latest` originates from the OCI reference: `registry.opensuse.org/opensuse/tumbleweed:latest` and that an OCI container file copy with that metadata cannot be provided as a file `/usr/src/packages/SOURCES/containers/openSUSE:Containers:Tumbleweed/containers/opensuse/tumbleweed:latest.ociarchive` on the worker ? Isn't all that information required to make the derived_container feature to work ? I was under the assumption it's the exact same code that just needs to be applied to another section of the kiwi description

Anyway I guess I'm lacking information about the backend. 

So how to continue here ? I'm lost :)

Thanks

mlschroe:
With the derived_from implementation, `obs://openSUSE:Containers:Tumbleweed/containers/opensuse/tumbleweed#latest` describes an _unpublished_ container, i.e. a container that it is currently not on a registry.

I think the new `containers` element is (at least for now) designed to fetch containers from a registry. You should not use the an `obs:` url for this.

So, my suggestion is to make
```
<containers source="registry.opensuse.org" backend="podman">
    <container name="tumbleweed" tag="latest" path="opensuse"/>
</containers>
```
work in OBS as well. To make this work we just need to change kiwi so that it fetches the container from the local filesystem. My suggestion was to use
```
if Defaults.is_buildservice_worker():
    container_uri = Uri('obsrepositories/:{0}'.format(container_endpoint), 'container')
```
so that it copies it from `/usr/src/packages/SOURCES/containers/_obsrepositories/registry.opensuse.org/opensuse/tumbleweed:latest.ociarchive`

We can later extend the code so that unpublished containers are also supported.

(We may also want to revisit the way derived_from is implemented so that it can also fetch containers from a registry. Or does that already work?)

schaefi:
> With the derived_from implementation, obs://openSUSE:Containers:Tumbleweed/containers/opensuse/tumbleweed#latest describes an unpublished container, i.e. a container that it is currently not on a registry.

ok that makes sense now I understand what you meant

> (We may also want to revisit the way derived_from is implemented so that it can also fetch containers from a registry. Or does that already work?)

I'm not sure, I myself use this feature only with project references. I would say we change this on demand and in an extra PR

Ok thanks for the info, I update the PR here such that the suggested location is searched and let you know when I'm done

schaefi:
@mlschroe I applied the suggested change and rolled out to Staging. The integration test now specifies

```xml
<containers source="registry.opensuse.org" backend="podman">
    <container name="tumbleweed" tag="latest" path="opensuse"/>
</containers>
```

In the same way was we would do it for a non obs build. When building in obs kiwi tries to fetch the container as follows

```
[   61s] [ ERROR   ]: 09:57:56 | KiwiCommandError: chroot: stderr: cp: cannot stat '/usr/src/packages/SOURCES/containers/_obsrepositories/registry.opensuse.org/opensuse/tumbleweed:latest.ociarchive': No such file or directory
```

Is this file path as you would expect it ?

Thanks

schaefi:
There is still a mistake on my end, the cp should not be a chroot call, but just to double check if the path is correct ? Thanks

schaefi:
For reference the commit in obs-build to add the feature on the backend: https://github.com/openSUSE/obs-build/commit/b375fb452ad1925e08dc07be4e8d2103e45a8654

PR #2682, opened 2024-11-25
Allow to derive from multiple containers

### LINKED ISSUE #2680: Support multiple containers as derived_from
## Problem description

When creating delta containers it is very useful to support multiple layers of delta_containers. This allows the creation of thins layers that can be combined in various ways to create a complete stack. For example it is useful to have a base-container that supplied the runtime requirements for the aws-cli (basic python modules only) and then have a delta container that isolates the SDK bits (boto3, botocore, and s3transfer) and then have the aws-cli container be a layer on top of these 2 containers.

## Expected behaviour

Allow specification of multiple `derived_from` containers via a defined separator, possible "," in the value string.

At present the `dervived_from` attribute can only have a single container specification.

* KIWI version: 
Latest

Any

* Operating system target version: 
openSUSE & SLE

Latest

schaefi:
Add support for multi inheritance to the derived_from attribute In the order of a comma separated list of docker source URI's a base tree is created. This was possible only with one container so far and Fixes #2680 as well as jira#OBS-354

### DIFF
--- a/kiwi/builder/container.py
+++ b/kiwi/builder/container.py
@@ -65,9 +65,9 @@ class ContainerBuilder:
             xml_state.build_type.get_metadata_path()
 
         if xml_state.get_derived_from_image_uri() and not self.delta_root:
-            # The base image is expected to be unpacked by the kiwi
-            # prepare step and stored inside of the root_dir/image directory.
-            # In addition a md5 file of the image is expected too
+            # The base image(all derived imports) is expected to be unpacked
+            # by the kiwi prepare step and stored inside of the root_dir/image
+            # directory. In addition a md5 file of the image is expected too
             self.base_image = Defaults.get_imported_root_image(
                 self.root_dir
             )
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -82,12 +82,12 @@ class SystemPrepare:
             root_dir, allow_existing
         )
         root.create()
-        image_uri = xml_state.get_derived_from_image_uri()
+        image_uris = xml_state.get_derived_from_image_uri()
         delta_root = xml_state.build_type.get_delta_root()
 
-        if image_uri:
+        if image_uris:
             self.root_import = RootImport.new(
-                root_dir, image_uri, xml_state.build_type.get_image()
+                root_dir, image_uris, xml_state.build_type.get_image()
             )
             if delta_root:
                 self.root_import.overlay_data()
--- a/kiwi/system/root_import/__init__.py
+++ b/kiwi/system/root_import/__init__.py
@@ -17,6 +17,7 @@
 #
 import logging
 import importlib
+from typing import List
 from abc import (
     ABCMeta,
     abstractmethod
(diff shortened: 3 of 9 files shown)


schaefi:
@rjschwei example image description for you to test 

```xml
<image schemaversion="7.5" name="kiwi-test-image-delta-joe">
    <description type="system">
        <author>Marcus Schäfer</author>
        <contact>marcus.schaefer@gmail.com</contact>
        <specification>Delta container providing the joe editor</specification>
    </description>
    <preferences>
        <type image="docker" derived_from="docker://registry.opensuse.org/home/marcus.schaefer/delta_containers/containers_tw/basesystem:latest,docker://registry.opensuse.org/home/marcus.schaefer/delta_containers/containers_tw/mc:latest" delta_root="true">
            <containerconfig name="joe"/>
        </type>
        <version>1.15.4</version>
    </preferences>
    <repository type="rpm-md">
        <source path="obsrepositories:/"/>
    </repository>
    <packages type="image">
        <package name="joe"/>
    </packages>
</image>
```

Conan-Kudo:
I'm confused, what is this supposed to do?

schaefi:
> I'm confused, what is this supposed to do?

This connects to the delta container support in kiwi and the flake-pilot project as a provision tool for apps provided as containers. It is a little bit comparable to Ubuntu's snaps but we use existing standards with OCI. The delta_root feature in kiwi allows you to create a container that only includes a delta compared to a given base container. In the past you could only specify one base and with this change you can specify more than one. Of course the resulting container is on its own no longer usable, it requires a provision of the data it was derived from. This task is done by the flake-pilot project I wrote recently.

At SUSE we want to use this concept to provide application containers faster and with less problems when these applications are third party, poor packaged or having dependencies the distribution does not/not yet provide. The container build in kiwi can then derive from "good" containers build from well maintained packages and only add the "evil" application and build a delta container from it. With flake-pilot you then register an application and bring things together. The application runs as an isolated instance on the system and we can offer customers an application without harming the integrity of the rest of the system.

There are more use cases to flake-pilot and I wrote all that up here

* https://github.com/OSInside/flake-pilot?tab=readme-ov-file#flake-pilot

The idea is around for a bit, but it seems at SUSE there is now a real use case and while we are working on it some issues and missing features will be added to both projects.

@rjschwei is one of the lead architects for what is called a component architecture and that's where most of the requests comes from and also the reason why I put him as a reviewer. 

Hope this clarifies a bit of the story.

schaefi:
The original idea why I started flake-pilot was to run automotive applications on linux hosts in cars. I hate to say that but I did not see a single automotive application that was not a bunch of security critical software. Everybody was concerned running this thing on the OS and sometimes we even only got a binary and no sources. As I'm no longer working in automotive I was using the project to improve my rust coding :smile: but it seems there are also other use cases for which it might find its niche

schaefi:
Ah and the above example is really not producing anything useful, it just serves the purpose for a test

rjschwei (approved):

PR #2688, opened 2024-12-02
Fix legacy_bios_mode detection

schaefi:
The code in this method does not work correctly if the firmware is set to 'bios'. In bios only mode the method returned a false value which is incorrect as it should return a true value in this case. Without this patch ISO images will fail to boot because no loader gets configured.


schaefi:
Found this while testing several ISO builds. The PR must not be merged until all integration tests has rebuild. I'll submit a staging build with this patch to let them run

schaefi:
Integration test builds looks good

PR #2691, opened 2024-12-02
Add support for loongarch64

### LINKED ISSUE #2684: Please add support for LoongArch
openSUSE is currently adding support for the LoongArch architecture, a new 64-bit architecture designed by the Chinese company Loongson, based on the MIPS architecture with custom refinements and RISC-V-like enhancements.

The architecture uses standard UEFI for booting which is fully supported by current versions of GRUB. Thus, adding support for LoongArch should be a matter of adding the EFI architecture name of LoongArch everywhere in KIWI where Arm64 support is enabled for UEFI systems.

For more information on LoongArch in openSUSE, see [this mailing list post](https://lists.opensuse.org/archives/list/factory@lists.opensuse.org/thread/6B37BCN7RSEXKK3JNY32IS6B4DIMII5G/).

glaubitz:
This Fixes #2684

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -75,6 +75,9 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         if arch == 'x86_64':
             # grub2 support for bios and efi systems
             self.arch = arch
+        elif arch == 'loongarch64':
+            # grub2 support for efi systems
+            self.arch = arch
         elif arch.startswith('ppc64'):
             # grub2 support for ofw and opal systems
             self.arch = arch
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -1307,6 +1307,7 @@ class Defaults:
             'armv7hl': ['efi', 'uefi'],
             'armv7l': ['efi', 'uefi'],
             'armv8l': ['efi', 'uefi'],
+            'loongarch64': ['efi', 'uefi'],
             'ppc': ['ofw'],
             'ppc64': ['ofw', 'opal'],
             'ppc64le': ['ofw', 'opal'],
@@ -1331,6 +1332,7 @@ class Defaults:
             'i586': 'bios',
             'i686': 'bios',
             'ix86': 'bios',
+            'loongarch64': 'efi',
             'ppc': 'ofw',
             'ppc64': 'ofw',
             'ppc64le': 'ofw',
@@ -1401,6 +1403,7 @@ class Defaults:
             'armv6l': 'arm-efi',
             'armv7l': 'arm-efi',
             'armv8l': 'arm-efi',
+            'loongarch64': 'loongarch64-efi',
             'riscv64': 'riscv64-efi'
         }
         if arch in default_module_directory_names:
@@ -1437,6 +1440,7 @@ class Defaults:
             'armv6l': 'bootarm.efi',
             'armv7l': 'bootarm.efi',
             'armv8l': 'bootarm.efi',
+            'loongarch64': 'bootloongarch64.efi',
             'riscv64': 'bootriscv64.efi'
         }
         if arch in default_efi_image_names:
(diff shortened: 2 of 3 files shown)


schaefi:
Thanks for digging into this topic. For the code itself can you please also adapt the tests accordingly ? Thanks much.
For merging a new architecture support at least one integration test build is needed. I assume you do this primarily for Debian ? Our debian/ubuntu integration tests can be found here:

* https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_$ARCH:debian
* https://build.opensuse.org/project/show/Virtualization:Appliances:Images:Testing_$ARCH:ubuntu

The descriptions you can find in the code here

* https://github.com/OSInside/kiwi/tree/main/build-tests

For adding loongarch64 we need a test at this point. I have to admit I never played with that architecture nor do I have access to any board that implements it, nor do I know which distribution builds for loongarch64

If this is all still in flux and providing a test build is not so easy we at least need a clean unit checklist for the changed code paths.

Thanks much

schaefi (changes requested):
see my former comment

schaefi:
@glaubitz also looked at the information from your hackweek project. So it seems like you are already able to build something for loongarch64 based on tumbleweed ?

glaubitz:
> @glaubitz also looked at the information from your hackweek project. So it seems like you are already able to build something for loongarch64 based on tumbleweed ?

Yes, openSUSE for loongarch64 is currently being bootstrapped [here](https://build.opensuse.org/project/show/devel:LoongArch:Factory).

The reason I want to add loongarch64 support to KIWI is to be able to build openSUSE installation images for loongarch64.

glaubitz:
I just had a look at the tests and I think implementing these for loongarch64 will be too much.

Is there a reason why we need those for loongarch64 but not riscv64?

Conan-Kudo:
We need them for RISCV64 too, the fact that they are missing is not good either.

schaefi:
> I just had a look at the tests and I think implementing these for loongarch64 will be too much.

I'm talking about covering your code changes with proper unit tests and providing at least one integration test build which you probably do anyway because I assume you have built something with your current set of bootstrap packages. I'm afraid we cannot merge code that is not covered by some sort of testing. If the current state of the distro does not allow you to run a build test in the buildservice (which is the service used for the integration test builds) then we can let that go for the moment, but covering code changes by a unit test is a must have in kiwi to keep full freedom to do changes in all parts of the code without regressions. I believe it is not too much of a burden to add that with regards to the code changes done in this PR.

Thanks

schaefi:
> > We need them for RISCV64 too, the fact that they are missing is not good either.
> 
> nobody ever used kiwi to build anything for riscv64. This is basically an unsupported arch for kiwi as we have not a single build for any riscv64 board. I'm not surprised that this is missing.

And to be more precise here; There was one guy in the past (Andreas Schwab) who added a patch in a similar fashion than Adrian does it now for loongarch64. The changes to kiwi regarding riscv were approved by me without further asking and we now have code for this architecture without any tests and without any integration test. This is basically dead code and nobody knows if it still works. I would say it doesn't. 

This is the reason why I'm picky on adding new architecture support to come with an image build that actually proves it also works. There is also a tendency that people who added some-arch support vanishes and leave others with arch specific code alone. You cannot read a test you cannot test the board because it was only done half-way. I myself know about x86, aarch64 and s390. I have no idea about ppc, riscv and loongarch64 and that means other people need to stay responsible for the work they do. Those who are willing to write tests and come up with integration tests are usually also there to fix stuff when things starts to break... and they always break.

Just in case you are asking yourself why I'm so picky on tests when it comes to new architectures, this is the reason and I hope for your understanding

schaefi:
@glaubitz I fixed the unit test in your fork and pushed a commit. I hope this is ok ?

schaefi:
I had  a short call with Adrian, he will try to add a test build to our build-tests/ area and once this exists I'm going to add a new project at `Virtualization:Appliances:Images:Testing_loongarch64` such that we have some test build to refer to. Once this is done we are ready for a merge :) Thanks

schaefi:
@glaubitz any news here ?

PR #2693, opened 2024-12-09
Use GPT table for EFI capable ISO images

### LINKED ISSUE #2685: Add support for building GPT-labelled ISOs with xorriso
## Problem description
Kiwi's current implementation of xorriso support does not allow you to produce GPT-labelled ISOs (that is, ISOs with a GPT disk label / partition table as opposed to an MBR one). This is a problem for Fedora, as our ISOs were hybrid MBR/GPT from Fedora 17 to Fedora 36, and have been pure GPT since Fedora 37. Switching to Kiwi for existing ISOs (as we're starting to experiment with) thus makes them suddenly MBR-labelled, which can cause problems - see https://bugzilla.redhat.com/show_bug.cgi?id=2328590 , where we discovered this.

I wrote a pretty simple patch which makes Kiwi *always* use a GPT disk label when producing an EFI-capable ISO with xorriso:
```
commit d1e6a14037cb7f5862c3a0b108448b7e230c7427 (HEAD -> live-gpt-test)
Author: Adam Williamson <awilliam@redhat.com>
Date:   Fri Nov 29 11:12:38 2024 -0800

    TEST DO NOT MERGE: build hybrid live as GPT not MBR
    
    Signed-off-by: Adam Williamson <awilliam@redhat.com>

diff --git a/kiwi/iso_tools/xorriso.py b/kiwi/iso_tools/xorriso.py
index 39eab4f32..f33cd11e5 100644
--- a/kiwi/iso_tools/xorriso.py
+++ b/kiwi/iso_tools/xorriso.py
@@ -157,6 +157,7 @@ class IsoToolsXorrIso(IsoToolsBase):
         bootloader/config/grub2.py
         """
         self.iso_loaders += [
+            '-boot_image', 'any', 'appended_part_as=gpt',
             '-append_partition', '2', '0xef', loader_file,
             '-boot_image', 'any', 'next',
             '-boot_image', 'any',
```
But I suspect it wouldn't be wanted as-is as we might not want to suddenly flip the disk label on existing Kiwi-built ISOs. We probably want this to be configurable. But I kinda gave up on figuring out the best way to do that, Kiwi is fairly big and this code is a long way from the config stuff, I feel like an experienced Kiwi dev can figure out the best approach to implementing a config toggle for this much more easily than I can.

## Expected behaviour
It should be possible to specify a Kiwi template for an ISO image such that the resulting image is GPT-labelled.

Build a live image with Kiwi. Note it has an MBR disk label (`gdisk -l` and `fdisk -l` will tell you this). Note there's no way to make it use GPT instead.

* KIWI version: 10.2.0
* Operating system target version: Fedora Rawhide
* Koji version (N/A if not using Koji): irrelevant

schaefi:
If the ISO should be able to boot via EFI make sure the partition table is GPT. In addition to that setting also make sure the CSM boot flag for the embedded MBR table is provided through an extra bearer partition in the same way as it's done for EFI capable disk images. For better debugging also add the report of the system area as debug information to the log file. This Fixes #2685

### DIFF
--- a/kiwi/iso_tools/xorriso.py
+++ b/kiwi/iso_tools/xorriso.py
@@ -103,6 +103,10 @@ class IsoToolsXorrIso(IsoToolsBase):
             self.iso_parameters += [
                 '-compliance', 'untranslated_names'
             ]
+        else:
+            self.iso_parameters += [
+                '-compliance', 'no_emul_toc'
+            ]
 
         if Defaults.is_x86_arch(self.arch) and legacy_bios_mode:
             mbr_file = os.sep.join(
@@ -127,7 +131,8 @@ class IsoToolsXorrIso(IsoToolsBase):
             else:
                 log.warning(f'No hybrid MBR file found: {mbr_file}: skipped')
             self.iso_loaders += [
-                '-boot_image', 'grub', 'grub2_boot_info=on'
+                '-boot_image', 'grub', 'grub2_boot_info=on',
+                '-boot_image', 'grub', 'partition_table=on'
             ]
 
         if Defaults.is_ppc64_arch(self.arch):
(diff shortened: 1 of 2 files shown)


schaefi:
We want to be sure the results are really better than what we had before. Test images available here:

*  https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-live

AdamWill:
https://github.com/OSInside/kiwi/pull/2695 is my alternative here.

PR #2695, opened 2024-12-12
xorriso: respect efiparttable and gpt_hybrid_mbr

### LINKED ISSUE #2685: Add support for building GPT-labelled ISOs with xorriso

AdamWill:
This should make the xorriso-based ISO build path respect the 'efiparttable' and 'gpt_hybrid_mbr' settings when building a UEFI-compatible image, making it write a GPT disk label by default instead of an MBR (msdos) one. If it's building an image that is not UEFI-compatible it will always write an MBR label, regardless of this setting.

If 'gpt_hybrid_mbr' is set, xorriso will write an Ubuntu-style MBR/GPT hybrid partition table, where the MBR partition table includes a partition with type 00 and the bootable flag, as well as the partition with type ee required by the UEFI spec. This mildly violates the UEFI spec but may make the image bootable on native BIOS or CSM firmwares which refuse to boot from a disk with no partition marked 'bootable' in the MBR. If 'gpt_hybrid_mbr' is not set, xorriso will write a strictly UEFI-spec compliant label, with just the 'protective MBR' required by the UEFI spec (no bootable partition) and the correct GPT partition table. Note this is somewhat different from what gpt_hybrid_mbr does for disk images.

Also, we now pass -compliance no_emul_toc when building ISOs, as recommended by upstream in
https://lists.gnu.org/archive/html/bug-xorriso/2024-11/msg00012.html This tool is generally always going to be building ISOs intended for write-once use, not multi-session use (and which are rarely, these days, written to physical discs at all anyway).

Fixes #2685 .

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -402,7 +402,7 @@ efifatimagesize="nonNegativeInteger":
 efiparttable="msdos|gpt":
   For images with an EFI firmware specifies the partition
   table type to use. If not set defaults to the GPT partition
-  table type
+  table type for disk images, MBR (msdos) for ISO images.
 
 dosparttable_extended_layout="true|false":
   For oem disk images, specifies to make use of logical partitions
@@ -889,7 +889,14 @@ force_mbr="true|false":
   partitions
 
 gpt_hybrid_mbr="true|false":
-  For GPT disk types only: Create a hybrid GPT/MBR partition table
+  For disk types, create a hybrid GPT/MBR partition table with an
+  'accurate' MBR table that will have no 'bootable' flagged partition
+  For ISO types, create a hybrid GPT/MBR partition table where the
+  MBR partition table contains a whole-disk 'protective' partition
+  and a second bootable-flagged partition (intended to make the image
+  bootable in both UEFI and BIOS modes on as much hardware as possible)
+  In both cases, only has any effect if the EFI partition table
+  type is GPT
 
 hybridpersistent="true|false":
   For the live ISO type, triggers the creation of a partition for
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -144,6 +144,8 @@ class LiveImageBuilder:
                 'mbr_id': self.mbrid.get_id(),
                 'application_id': self.application_id,
                 'efi_mode': self.firmware.efi_mode(),
+                'efi_partition_table': self.firmware.get_partition_table_type(),
+                'gpt_hybrid_mbr': self.firmware.gpt_hybrid_mbr,
                 'legacy_bios_mode': self.firmware.legacy_bios_mode()
             }
         }
--- a/kiwi/builder/install.py
+++ b/kiwi/builder/install.py
@@ -161,6 +161,8 @@ class InstallImageBuilder:
                 'mbr_id': self.mbrid.get_id(),
                 'application_id': self.xml_state.build_type.get_application_id(),
                 'efi_mode': self.firmware.efi_mode(),
+                'efi_partition_table': self.firmware.get_partition_table_type(),
+                'gpt_hybrid_mbr': self.firmware.gpt_hybrid_mbr,
                 'ofw_mode': self.firmware.ofw_mode(),
                 'legacy_bios_mode': self.firmware.legacy_bios_mode()
             }
--- a/kiwi/filesystem/isofs.py
+++ b/kiwi/filesystem/isofs.py
@@ -50,6 +50,6 @@ class FileSystemIsoFs(FileSystemBase):
         iso_tool.init_iso_creation_parameters(meta_data)
 
         if efi_loader:
-            iso_tool.add_efi_loader_parameters(efi_loader)
+            iso_tool.add_efi_loader_parameters(efi_loader, meta_data)
 
         iso_tool.create_iso(self.filename)
--- a/kiwi/firmware.py
+++ b/kiwi/firmware.py
@@ -42,6 +42,7 @@ class FirmWare:
         self.firmware = xml_state.build_type.get_firmware()
         self.efipart_mbytes = xml_state.build_type.get_efipartsize()
         self.efi_partition_table = xml_state.build_type.get_efiparttable()
+        self.gpt_hybrid_mbr = xml_state.build_type.get_gpt_hybrid_mbr()
         self.efi_csm = True if xml_state.build_type.get_eficsm() is None \
             else xml_state.build_type.get_eficsm()
 
--- a/kiwi/iso_tools/base.py
+++ b/kiwi/iso_tools/base.py
@@ -19,7 +19,7 @@ import os
 import shutil
 import logging
 from typing import (
-    Dict, List, Union
+    Dict, List, Optional, Union
 )
 
 # project
@@ -72,7 +72,9 @@ class IsoToolsBase:
         """
         raise NotImplementedError
 
-    def add_efi_loader_parameters(self, loader_file: str) -> None:
+    def add_efi_loader_parameters(
+        self, loader_file: str, custom_args: Optional[Dict[str, Union[str, bool]]] = None
+    ) -> None:
         """
         Add ISO creation parameters to embed the EFI loader
 
--- a/kiwi/iso_tools/xorriso.py
+++ b/kiwi/iso_tools/xorriso.py
@@ -103,6 +103,11 @@ class IsoToolsXorrIso(IsoToolsBase):
             self.iso_parameters += [
                 '-compliance', 'untranslated_names'
             ]
+        else:
+            self.iso_parameters += [
+                # https://lists.gnu.org/archive/html/bug-xorriso/2024-11/msg00012.html
+                '-compliance', 'no_emul_toc'
+            ]
 
         if Defaults.is_x86_arch(self.arch) and legacy_bios_mode:
             mbr_file = os.sep.join(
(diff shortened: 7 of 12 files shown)


AdamWill:
Compare with @schaefi 's https://github.com/OSInside/kiwi/pull/2693 . I need to update the docs for gpt_hybrid_mbr but I wanted to get this pushed. Note @schaefi also added `'-boot_image', 'grub', 'partition_table=on'` and `'-boot_image', 'any', 'partition_cyl_align=off'`, but I'm not clear on why, and his commit message doesn't say, and I can't see anywhere Thomas recommended them in the bug or the mailing list. If there's a good reason to add these, I can go ahead and add them.

I did adapt the `report_call` bit from @schaefi 's PR and adapted the test changes from it too, thanks for those.

Conan-Kudo:
@AdamWill do you have a sample image build using this change to test with?

AdamWill:
no, I didn't get to that yet. Going to do it soon.

AdamWill:
https://openqa.stg.fedoraproject.org/tests/4524655/asset/iso/04524655-Fedora-KDE-Live-x86_64-126774230.iso , I'll upload it to my FP space in a minute as it'll get garbage collected. That's built with the GPT and hybrid options set.

AdamWill:
Here's the relevant log bit:
```
[ DEBUG   ]: 21:46:00 | EXEC: [/usr/bin/xorriso -application_id KDE-Desktop-Live-126774230 -publisher Fedora Project -preparer_id KIWI - https://github.com/OSInside/kiwi -volid KIWI-126774230 -joliet on -padding 0 -compliance no_emul_toc -outdev /builddir/result/image/Fedora.x86_64-Rawhide.iso -map /builddir/result/image/live-media.b80yanuy / -chmod 0755 / -- -boot_image grub bin_path=boot/x86_64/loader/eltorito.img -boot_image grub grub2_mbr=/builddir/result/image/live-media.b80yanuy/boot/x86_64/loader/boot_hybrid.img -boot_image grub grub2_boot_info=on -boot_image any partition_offset=16 -boot_image any cat_path=boot/x86_64/boot.catalog -boot_image any cat_hidden=on -boot_image any boot_info_table=on -boot_image any platform_id=0x00 -boot_image any emul_type=no_emulation -boot_image any load_size=2048 -boot_image any appended_part_as=gpt -boot_image any mbr_force_bootable=on -append_partition 2 0xef /builddir/result/image/efi-loader.skcf_7kg -boot_image any next -boot_image any efi_path=--interval:appended_partition_2:all:: -boot_image any platform_id=0xef -boot_image any emul_type=no_emulation]
[ DEBUG   ]: 21:46:12 | Looking for xorriso in /usr/bin:/bin:/usr/sbin:/sbin
[ DEBUG   ]: 21:46:12 | EXEC: [/usr/bin/xorriso -indev /builddir/result/image/Fedora.x86_64-Rawhide.iso -report_system_area plain]
[ DEBUG   ]: 21:46:12 | System area options: 0x00004201
System area summary: MBR protective-msdos-label grub2-mbr cyl-align-off GPT
ISO image size/512 : 5613184
Partition offset   : 16
MBR heads per cyl  : 0
MBR secs per head  : 0
MBR partition table:   N Status  Type        Start       Blocks
MBR partition      :   1   0x00  0xee            1      5613183
MBR partition      :   2   0x80  0x00            0            1
GPT                :   N  Info
GPT disk GUID      :      88f0d44d807ab24b945351d60c3b0b6f
GPT entry array    :      2  248  separated
GPT lba range      :      64  5613120  5613183
GPT partition name :   1  490053004f003900360036003000
GPT partname local :   1  ISO9660
GPT partition GUID :   1  88f0d44d807ab24b945251d60c3b0b6f
GPT type GUID      :   1  a2a0d0ebe5b9334487c068b6b72699c7
GPT partition flags:   1  0x1000000000000001
GPT start and size :   1  64  5551616
GPT partition name :   2  41007000700065006e006400650064003200
GPT partname local :   2  Appended2
GPT partition GUID :   2  88f0d44d807ab24b945151d60c3b0b6f
GPT type GUID      :   2  28732ac11ff8d211ba4b00a0c93ec93b
GPT partition flags:   2  0x0000000000000000
GPT start and size :   2  5551680  61440
```

schaefi (approved):
Thanks much, looks good to me and I can see all relevant changes from #2693 to be part of this PR :+1:

kiwi/builder/live.py
```diff
@@ -144,6 +144,8 @@ def create(self) -> Result:
                 'mbr_id': self.mbrid.get_id(),
                 'application_id': self.application_id,
                 'efi_mode': self.firmware.efi_mode(),
+                'efi_partition_table': self.firmware.get_partition_table_type(),
+                'gpt_hybrid_mbr': self.firmware.gpt_hybrid_mbr,
```
schaefi, line 148: @AdamWill can you add these settings also to `kiwi/builder/install.py` at self.custom_iso_args in the same way ? Otherwise it does not apply when building install ISO's for oem types. Thanks

schaefi:
@AdamWill

> Note @schaefi also added '-boot_image', 'grub', 'partition_table=on' and '-boot_image', 'any', 'partition_cyl_align=off', but I'm not clear on why, 

I added those because in my mail conversation with Thomas he used them too. But frankly speaking I cannot give a good answer to the why

AdamWill:
> @AdamWill
> 
> > Note @schaefi also added '-boot_image', 'grub', 'partition_table=on' and '-boot_image', 'any', 'partition_cyl_align=off', but I'm not clear on why,
> 
> I added those because in my mail conversation with Thomas he used them too. But frankly speaking I cannot give a good answer to the why

Let's ask him! I'll ask over in the rh bug.

doc/source/image_description/elements.rst
```diff
@@ -889,7 +889,15 @@ force_mbr="true|false":
   partitions
 
 gpt_hybrid_mbr="true|false":
-  For GPT disk types only: Create a hybrid GPT/MBR partition table
+  For disk types, create a hybrid GPT/MBR partition table with an
+  'accurate' MBR table that will have no 'bootable' flagged partition
+  For the live ISO type, create a hybrid GPT/MBR partition table
```
Conan-Kudo, line 894: `live ISO types` -> `ISO types`. This now applies to any ISO created through kiwi.

doc/source/image_description/elements.rst
```diff
@@ -889,7 +889,15 @@ force_mbr="true|false":
...
+  where the MBR partition table contains a whole-disk 'protective'
+  partition and a second bootable-flagged partition (intended to
+  make the image bootable in both UEFI and BIOS modes on as much
+  hardware as possible)
+  In both cases, only has any effect if the EFI partition table
+  type is GPT
```
Conan-Kudo, line 900:
```suggestion
  In both cases, this only has any effect if the EFI partition 
  table type is GPT.
```

Conan-Kudo (changes requested):
Two minor documentation fixes and then I think this is good to go.

doc/source/image_description/elements.rst
AdamWill: I don't think this is at all necessary, the "this" is implied, the line length is fine, and the file is inconsistent about whether entries end with a period or not. The ones immediately before and after this one don't end with a period either, which is why I did the same for this one.

doc/source/image_description/elements.rst
AdamWill: Done.

doc/source/image_description/elements.rst
Conan-Kudo: Okay. We really need to take a grammar hammer to the docs... 😅

Conan-Kudo (approved):

Issue #1931, opened 2021-09-20
grub.cfg path incorrect when using BTRFS for root filesystem


luisgf:
## Problem description

Creating an ubuntu 20.04 image (may be applicable to other ubuntu versions) creates a wrong path in grub.cfg that leads to a image that doesn't boot properly when cloned to disk. 

Grub doesn't honor the default subvolume directive for a btrfs volume, so get_boot_path() should be take care of that and return a path that include the default volume @.  

The grub.cfg generated is this:

    set btrfs_relative_path="yes"
    search --fs-uuid --set=root ee00dc7e-b0c9-4b8e-ab7d-363af93f28b4
    set prefix=($root)/boot/grub
    configfile ($root)/boot/grub/grub.cfg

And the correct entry should be that:

    set btrfs_relative_path="yes"
    search --fs-uuid --set=root ee00dc7e-b0c9-4b8e-ab7d-363af93f28b4
    set prefix=($root)/@/boot/grub
    configfile ($root)/@/boot/grub/grub.cfg

## Expected behaviour
Image cloned to disk boots properly.

## Steps to reproduce the behaviour

Create a kiwi config.file with this configuration for storage:

    <preferences profiles="Disk">
          <type image="oem" filesystem="btrfs" firmware="uefi" installiso="true" btrfs_root_is_snapshot="false" installboot="install" installprovidefailsafe="false">
              <oemconfig>
                  <oem-swap>false</oem-swap>
                  <oem-boot-title>MyDevice</oem-boot-title>
                  <oem-unattended>false</oem-unattended>
                  <oem-swapsize>1024</oem-swapsize>
                  <oem-device-filter>/dev/ram</oem-device-filter>
                  <oem-multipath-scan>false</oem-multipath-scan>
              </oemconfig>
              <systemdisk name="vgsys">
                  <volume name="@root" size="10G"/>
                  <volume name="@home" mountpoint="/home" size="5G" />
                  <volume name="@tmp" mountpoint="/tmp" size="1G" />
                  <volume name="@varlog" mountpoint="/var/log" size="5G" copy_on_write="false" />
              </systemdisk>
          </type>
      </preferences>

## OS and Software information
+ KIWI version: 9.23.58
+ Operating system: Ubuntu 20.04
+ OBS version:

luisgf:
Thanks @schaefi 

[...]

PR #2697, opened 2024-12-18
WIP: Adapt to latest snapper

davidcassany:
This PR is a reaction to the latest snapper features to better support image building and installation processes, see openSUSE/snapper#944

The fundamental idea is to prevent KIWI from running logic and configurations which are silently coupled to snapper internals. KIWI does not require snapper to build images in which the root is a snapshot, however this assumes snapper within the image to be meaningful. As a consequence KIWI manually creates the `.snapshots/1/snapshot` subvolume structure and its `.snapshots/1/info.xml` metadata and it also creates the root snapper configuration. All these are snapper specific details being managed without using snapper assuming there will be snapper in the image to actually be meaningful.

With the latest changes in snapper on the mentioned issue snapper provides an `installation-helper` utility to help us creating the first empty snapshot (so we can easily dump in the root-tree) and also generate the appropriate setup. The frees KIWI to know about snapper internals and it explicitly couples the `btrfs_root_is_snapshot` flag to snapper usage. The counter part is requiring snapper as build time dependency.

This PR is not complete as it does not address:
* backward compatibility
* snapper requirement in spec
* unit tests

However this is enough for rough testing and discuss weather the change is worth or not. The code in this PR is just droping some btrfs calls and manual snapper configuration setup in favor of the `installation-helper`.


kiwi/volume_manager/btrfs.py
```diff
@@ -438,6 +426,7 @@ def set_property_readonly_root(self):
         root_is_readonly_snapshot = \
             self.custom_args['root_is_readonly_snapshot']
         if root_is_snapshot and root_is_readonly_snapshot:
+            # TODO we could call snapper modify here instead
             sync_target = self.get_mountpoint()
             Command.run(
                 ['btrfs', 'property', 'set', sync_target, 'ro', 'true']
```
davidcassany, line 432: Here we could run something like `snapper --dbus --root <image_root> modify --read-only 1` if we require snapper as part of the build environment.

schaefi:
> and it explicitly couples the btrfs_root_is_snapshot flag to snapper usage. 
I really like this. My only concern is that we would still need the "old" code somehow for backward compatibility. Do you know a stable way to identify that we need to run the fallback code path other than "try with snapper first, if it fails fallback" ?

>The counter part is requiring snapper as build time dependency
I see the current code runs `/usr/lib/snapper/installation-helper` and also `snapper` from the build host. Do you think it would be possible to call the tools from the new root (`chroot`) instead. This way we could also say we explicitly run the fallback code if no `snapper` tool was found in the root tree

Overall I really like this initiative and we should do the proposed change. Thanks much

davidcassany:
> > and it explicitly couples the btrfs_root_is_snapshot flag to snapper usage.
> 
> I really like this. My only concern is that we would still need the "old" code somehow for backward compatibility. Do you know a stable way to identify that we need to run the fallback code path other than "try with snapper first, if it fails fallback" ?

Well, for the time being this new snapper logic is merged upstream but not yet included in a specific release. My idea was something like checking the snapper version once a new release including the change is available at the very least in TW. I don't think there is a better way to handle backward compatibility other than keeping around the old logic in case we detect the given snapper does not include the new logic. Here I dropped the code in this PR to make explicit which are the bits that could be eventually removed at some point, just to make the change obvious for discussion.

> I see the current code runs /usr/lib/snapper/installation-helper and also snapper from the build host. Do you think it would be possible to call the tools from the new root (chroot) instead.

That is a good point and it should be possible, then `btrfs_root_is_snapshot` would require snapper in the image to actually build, which is good IMHO (note this concept could be extended to almost any tool used after `system create` step... 🤔 ). The implementation might be a bit more invasive as this chroot env I am not sure how complex would it be to bring it up there.

> Overall I really like this initiative and we should do the proposed change. Thanks much

Great, count on me to bring this to a mergeable state on early January 😉

Conan-Kudo:
I don't like this. There are a multitude of snapshot management tools for Btrfs. Snapper just happens to be the oldest one. I would actually prefer if we're going to have a snapper-specific thing that we _explicitly_ create a snapper attribute instead of making something that looks tool-agnostic be snapper-specific.

schaefi:
> I don't like this. There are a multitude of snapshot management tools for Btrfs. Snapper just happens to be the oldest one. I would actually prefer if we're going to have a snapper-specific thing that we _explicitly_ create a snapper attribute instead of making something that looks tool-agnostic be snapper-specific.

ok so we have the following in our schema 

```xml
k.type.btrfs_root_is_snapshot.attribute =
        ## Tell kiwi to install the system into a btrfs snapshot
        ## The snapshot layout is compatible with the snapper management
        ## toolkit. By default no snapshots are used
```

This attribute is responsible for creating the snapshot layout suitable for snapper and also suitable for SUSE. It's been used on SUSE btrfs based images only at the moment and I suggest that we rename this attribute e.g `btrfs_root_is_snapper_snapshot`. Along with the change update the documentation and also mention that SUSE is the primary user of it and that using it will require snapper.

We would need an XSLT stylesheet to auto-rename the attribute and a schema version bump but I would say it's worth it regarding the snapper dependency that is been introduced here.

In any way it's a good idea from David to get rid of the "clone code" in kiwi to produce a snapper compatible layout

davidcassany:
@schaefi @Conan-Kudo I agree with you, all snapshots management in KIWI is already highly coupled with snapper and I agree this should be explicit. I also doubt we can built a generic concept here as snapper snapshots is something we must support (it's being used and it has been there for years) and I doubt we can consider it a generic approach, meaning the same snapper snapshots being managed by a different tool.

So to wrap up, and according to the discussion, to push this forward this is what we would need:
* A new snapper release
* Adapt the XML to rename `btrfs_root_is_snapshot` attribute to explicitly related it to snapper (note this is an independent change)
* Refactor the PoC of this PR to actually not require snapper in the host (use chroot), so no changes in spec would be required.
* Keep backward compatibility so keep old logic with path selector based on the detected version of snapper in the root-tree.
* Add and adapt tests

Does it make sense to you?

schaefi:
@davidcassany makes perfect sense to me :+1: I volunteer to implement

- Adapt the XML to rename btrfs_root_is_snapshot attribute to explicitly related it to snapper (note this is an independent change)

davidcassany:
@schaefi I am adapting the proof of concept to make use of chroots instead of calling the snapper, I will also paste here the config.XML I use for testing.

After that I'd like to acknowledge here in general terms that the implementation I am doing is actually matching expectations, basically before getting into the mess of adding backward compatibility and reworking the tests I'd like to make sure the implementation itself does not require a significant rework.

davidcassany:
I adapted the PoC so that snapper is only called from the image root-tree, to do that chrooting at the `<target-dir>/build/image-root/` is required together with access to the btrfs partition mountpoints. Since this requires either new mountpoints for subvolumes under the root tree or either bind mounts to the mounted subvolumes to the actual root three I though having a chroot context manager would be handy and eventually reusable. This is why a `ChrootManager` class is created. It could allow to do something like:

```python
which ChrootManager(root_dir, binds=['/sys', '/proc', '/dev']) as chroot:
    chroot.run(['snapper', '--no-dbus', ...])
```

in `__enter__` is mounts the bind mounts and in `__exit__` they are umounted. I realized this does not follow exactly the same critera of other context managers in KIWI, as it actually runs some setup logic in `__enter__`. Other KIWI context managers don't do that and I suppose there is a reason for that, is it due to having to handle exception in `__enter__`?

I am not fully convinced about the ChrootManager as it feels and looks similar to ImageSystem class, however ImageSystem assumes there is a fully functional root-tree on the device which is not the case when the involved btrfs and snapper operations of this PR are called. Probably we could simple have helper functions inside VolumeManagerBtrfs, but this prevents any possibility of later reuse. All in all I kept the ChrootManager class as it is small and simple should be easy to rework of someone has better ideas or concerns.,

The reason I think reuse might be interesting because I think that using the image root-tree as a chroot environment to operate over the devices is a powerful concept that could be potentially extended to other use cases. I wonder if LVM tools and filesystem tools couldn't also be leveraged from the image tree as well, preventing its requirement in the host itself. Wouldn't it be an option? Obviously this would be a significant refactor probably something we don't want, but I think it is worth a separate discussion.

davidcassany:
To test the current PR I used a simplified version of openSUSE-MicroOS image, I only set a single profile:

```xml
<image schemaversion="7.4" name="openSUSE-MicroOS" displayname="openSUSE MicroOS">
    <description type="system">
        <author>openSUSE Project</author>
        <contact>opensuse-factory@opensuse.org</contact>
        <specification>openSUSE MicroOS</specification>
    </description>
    <profiles>
        <profile name="kvm-and-xen" description="kvm-and-xen" arch="x86_64" import="true"/>
    </profiles>
     <preferences profiles="kvm-and-xen" arch="x86_64">
        <version>16.0.0</version>
        <bootloader-theme>openSUSE</bootloader-theme>
        <locale>en_US</locale>
        <type image="oem" filesystem="btrfs" format="qcow2" firmware="uefi" bootpartition="false"
            bootkernel="custom" devicepersistency="by-uuid" btrfs_root_is_snapshot="true"
            btrfs_root_is_readonly_snapshot="true" trfs_quota_groups="true">
            <bootloader name="grub2" console="gfxterm" />
            <systemdisk>
                <volume name="home"/>
                <volume name="root"/>
                <volume name="opt"/>
                <volume name="srv"/>
                <volume name="boot/writable"/>
                <volume name="usr/local"/>
                <volume name="var" copy_on_write="false"/>
            </systemdisk>
            <size unit="G">20</size>
        </type>
    </preferences>
    <repository type="rpm-md" alias="repo-oss">
        <source path="http://download.opensuse.org/tumbleweed/repo/oss/"/>
    </repository>
    <packages type="image">
        <package name="live-add-yast-repos"/>
        <package name="patterns-microos-basesystem"/>
        <package name="patterns-microos-base-zypper"/>
        <package name="patterns-microos-defaults"/>
        <package name="patterns-microos-selinux"/>
        <package name="kernel-default"/>
        <package name="device-mapper"/>
        <package name="cryptsetup"/>
    </packages>
    <packages type="image" profiles="kvm-and-xen">
        <package name="ignition-dracut"/>
        <package name="combustion &gt;= 1.2"/> <!-- New firstboot mechanism -->
        <package name="jeos-firstboot"/>
        <package name="growpart-generator"/>
        <package name="patterns-base-bootloader"/>
    </packages>
    <packages type="image" profiles="kvm-and-xen">
        <!-- KVM and Xen specific packages -->
        <package name="xen-tools-domU" arch="x86_64"/>
        <package name="qemu-guest-agent"/>
    </packages>
    <packages type="bootstrap">
        <package name="coreutils"/>
        <package name="gawk"/>
        <package name="gzip"/>
        <package name="hostname"/>
        <package name="openssl"/>
        <package name="filesystem"/>
        <package name="MicroOS-release-dvd"/>
    </packages>
</image>
```

schaefi:
> @davidcassany   Other KIWI context managers don't do that and I suppose there is a reason for that, is it due to having to handle exception in __enter__?

Remember the move into context managers was a refactoring from existing classes. Therefore the required __enter__ and __exit__ methods where added to the existing implementation to not break the interface. Also if you design your context managers in this way

```python
class Foo:
    def __init__(self):
        print('some foo')

    def __enter__(self):
        return self

    def cleanup(self):
        print('done')

    def __exit__(self, exc_type, exc_value, traceback) -> None:
        self.cleanup()
```

You can create an object instance that runs the constructor code in both ways the same

```python
with Foo() as foo:
    pass

foo = Foo()
foo.cleanup
```

In your case this does not apply as you are introducing a new class and you can decide how people should use it. Context Manager only is fine with me

schaefi:
> I am not fully convinced about the ChrootManager as it feels and looks similar to ImageSystem 

You are right when you say ImageSystem expects a full system to be mountable. It was the reason why I added that class in the past. I like the additional ChrootManager class to invoke commands chrooted. I was thinking instead of a new class you could also turn the Command class into a context manager such that we can call:

```python
with Command(root_dir, binds=['/sys', '/proc', '/dev']) as chroot:
    chroot.run(['snapper', '--no-dbus', ...])
```

Thoughts ?

schaefi:
> I wonder if LVM tools and filesystem tools couldn't also be leveraged from the image tree as well, preventing its requirement in the host itself. Wouldn't it be an option? 

yes it would be an option to place more commands that we currently call from the host as chroot calls. I just can't estimate after effects of this move. I'm open to do this one after the other in separate PRs and we can see if the integration tests that makes use of the commands that we moved still works. Would that be a plan ?

schaefi:
> @schaefi I am adapting the proof of concept to make use of chroots instead of calling the snapper, I will also paste here the config.XML I use for testing.
> 
> After that I'd like to acknowledge here in general terms that the implementation I am doing is actually matching expectations, basically before getting into the mess of adding backward compatibility and reworking the tests I'd like to make sure the implementation itself does not require a significant rework.

Implementation wise this is exactly what I had expected. A lot code path removed and replaced by snapper calls. Thanks much for doing this. Regarding ChrootManager, see my idea from the former comment. iirc all code changed is also only active if `btrfs_root_is_snapper_snapshot` is set. Given you plan a backward compatibility there should also be no problem for systems that has no snapper or an older version of it.

Overall you have my vote on this. We should wait for @Conan-Kudo though

kiwi/volume_manager/btrfs.py
```diff
@@ -413,19 +407,24 @@ def sync_data(self, exclude=None):
...
-                    )
-                )
             data = DataSync(self.root_dir, sync_target)
             data.sync_data(
                 options=Defaults.get_sync_options(), exclude=exclude
             )
+            if self.custom_args['root_is_snapper_snapshot']:
+                root_prefix = os.path.join(
+                    self.mountpoint,
+                    f'{self.root_volume_name}/.snapshots/1/snapshot'
```
Conan-Kudo, line 417: Can we have snapper tell us what this is? This is still an assumption of how snapper works rather than snapper telling us stuff.

kiwi/volume_manager/btrfs.py
```diff
@@ -503,76 +502,22 @@ def _set_default_volume(self, default_volume):
...
             [
                 self.mountpoint,
                 f'{self.root_volume_name}/.snapshots/1/snapshot'
             ]
         )
-        snapper_default_conf = Defaults.get_snapper_config_template_file(
-            root_path
+        # snapper requires an extra parent qgroup to operate with quotas
+        Command.run(
+            ['btrfs', 'qgroup', 'create', '1/0', self.mountpoint]
```
Conan-Kudo, line 514: This is again something that snapper should do rather than us.

Conan-Kudo (changes requested):
Some notes about implementation that should be resolved.

Conan-Kudo:
> > @schaefi I am adapting the proof of concept to make use of chroots instead of calling the snapper, I will also paste here the config.XML I use for testing.
> > After that I'd like to acknowledge here in general terms that the implementation I am doing is actually matching expectations, basically before getting into the mess of adding backward compatibility and reworking the tests I'd like to make sure the implementation itself does not require a significant rework.
> 
> Implementation wise this is exactly what I had expected. A lot code path removed and replaced by snapper calls. Thanks much for doing this. Regarding ChrootManager, see my idea from the former comment. iirc all code changed is also only active if `btrfs_root_is_snapper_snapshot` is set. Given you plan a backward compatibility there should also be no problem for systems that has no snapper or an older version of it.
> 
> Overall you have my vote on this. We should wait for @Conan-Kudo though

This makes sense to me too. The key thing though is that we shouldn't be doing _any_ configuration ourselves of the volume if we expect snapper to manage it. We need to go fully to "we don't know wtf is going on" in this case.

kiwi/volume_manager/btrfs.py
davidcassany: Not so sure about this, I don´t know enough about this use case for a proper opinion, but at a glance I am not convinced this is on snapper responsibilities. Snapper _only_ manages subvolumes in a CRUD fashion and it expects a btrfs filesystem there ready to be used. We are just telling to snapper which quota group to use on the existing filesystem.

davidcassany:
> yes it would be an option to place more commands that we currently call from the host as chroot calls. I just can't estimate after effects of this move. I'm open to do this one after the other in separate PRs and we can see if the integration tests that makes use of the commands that we moved still works. Would that be a plan ?

Sure, something like that should be explored in small steps. For instance, lets try to run LVM support this way and focus only on LVM tools and see how good or painful it turns to be.

davidcassany:
Finally found some time to rework this PR to include all the unit tests and arrange the code to keep backward compatibility. Obviously, keeping the logic for older snapper around prevents cleaning up the associated code, so this PR ends up adding a bunch of logic not removing.

I tried to keep clean the dual paths, hence the actual logic changes happens only it two methods which are actually  starting with a path fork based an snapper version:

```python
if CommandCapabilities.check_version('snapper', (0, 12, 1)):
   ...
else:
   ...
```

I also kept the new class ChrootManager instead of expanding Command class, the reason for that is simple, my current implementation of the ChrootManager makes use of the MountManager class to handle bind mounts, which in turn imports the Command class. So, to prevent circular dependencies I kept it as a separate class. Moreover, an implementation like the suggestion from https://github.com/OSInside/kiwi/pull/2697#issuecomment-2599807465 would have been tricky too, as Command.run is a static method and I don't think it can be overloaded nicely with a non static method, that leads to having to rename the method which IMHO would also lead to a confusing implementation. All in all I believe ChrootManager logic should not be part of Command class.

There are still a couple of issues to be covered:

* @Conan-Kudo's concerns regarding the logic still being coupled to snapper internals. He is right and I am not sure how to proceed in this front. Shall we create a snapper specific class? Hoping the concept will also clear the path to snapper alternatives some day? I think this would be the right way of coding it, but not sure to what extend is worth the effort or if we have well defined picture enough to get into that.
* Regarding the ChrootManager class I have a doubt, should we support creating bind mounts at a custom location? For instance mounting `/path/to/loop/device/part` to `/some/root/home`, where `/some/root` is the chroot path. Current implementations assumes the mount point will be the same in the inner root, just pre-appending the new root. This works fine for the classic and common `/dev`, `/proc` and `/sys` and probably for that vast majority of cases, but probably not for all. Is this something we should support? I am considering it now is this is likely to change the class API.

Conan-Kudo:
> @Conan-Kudo's concerns regarding the logic still being coupled to snapper internals. He is right and I am not sure how to proceed in this front. Shall we create a snapper specific class? Hoping the concept will also clear the path to snapper alternatives some day? I think this would be the right way of coding it, but not sure to what extend is worth the effort or if we have well defined picture enough to get into that.

I would prefer this since eventually we'll want to support @bmr-cymru's snapm and boom tooling that Fedora/Alma/CentOS/RHEL has.

schaefi:
> Shall we create a snapper specific class? 

yes actually we decided to move forward in this direction.

The biggest problem I see is not technology but time. I'm not sure if @davidcassany will have the time to work on this topic in the near future

Conan-Kudo:
We'll also need to deal with the fact that we have random snapper assumptions to pull into this rework, like #1931.

davidcassany:
@Conan-Kudo @schaefi I have updated the PR, finally managed to find out some time. I moved the logic in to a SnapshotManagerSnapper class. Lets see from there what needs to be changed and improved.

Regarding #1931 I am a bit skeptical, I did not check it in full detail, but this doesn't seams to be an snapper specific thing...  hence this might not fall into this specific refactor. In fact snapper does not require a `@` root subvolume, I'd say this is a SUSE thing rather than a snapper thing. See the snapper constructor gets the root subvolume (`@`) to operate on. I'd say that #1931 is essentially caused by deviations between upstream grub2 and SUSE's grub2.

Happy to discuss and find areas of improvement.

davidcassany:
I prepared a couple of tests building factory's [openSUSE MicroOS](https://build.opensuse.org/project/show/home:dcassany:branches:openSUSE:Factory) (this one uses latest snapper helper methods) and another one building [openSUSE Leap Micro 6.0](https://build.opensuse.org/project/show/home:dcassany:branches:openSUSE:Leap:Micro:6.0) (this one manually crafts snapper snapshot without snapper intervention, this the old logic).  Both images are bootable and transactional-updates is functional on them.

Conan-Kudo:
Does the snapper installation helper need to be on the host or the install root?

kiwi/snapshot_manager/base.py
```diff
@@ -0,0 +1,62 @@
...
+    :param dict custom_args: dictionary of custom snapshot manager arguments
+    """
+    def __init__(
+        self, device: str, root_dir: str, mountpoint: str,
+        root_volume_name: str, custom_args: Optional[Dict[str, Any]] = None
+    ) -> None:
+        self.device = device
+        self.root_dir = root_dir
+        self.mountpoint = mountpoint
+        self.root_volume_name = root_volume_name
```
davidcassany, line 44: This is `@` for a SUSE system, having a root `@` subvolume  is not a snapper choice.

davidcassany:
> Does the snapper installation helper need to be on the host or the install root?

In the install root, it does not run snapper from the host.

Conan-Kudo:
Then I think this is probably fine.

schaefi:
> @Conan-Kudo @schaefi I have updated the PR, finally managed to find out some time. I moved the logic in to a SnapshotManagerSnapper class. Lets see from there what needs to be changed and improved.

@davidcassany Thanks, much appreciated

> Regarding #1931 I am a bit skeptical, I did not check it in full detail, but this doesn't seams to be an snapper specific thing... hence this might not fall into this specific refactor.

I agree. If this is an issue we need to care for we can take a look after this refactor

> Happy to discuss and find areas of improvement.

I have to admit that I lost a bit track on this one because it is open for so long. We have several integration tests that uses btrfs in different subvolume and snapshot settings. I suggest that we squash commits as you see fit and I prepare a Staging python-kiwi package which will cause all integration tests to run. From there we can check the btrfs based ones to still look correct.

If that is the case I would say we can move forward.

I also saw this PR is still in draft mode, on purpose ?

Thanks much

davidcassany:
> I have to admit that I lost a bit track on this one because it is open for so long. We have several integration tests that uses btrfs in different subvolume and snapshot settings. 

My comment was here was pointing from a design and abstraction perspective, not from a functional PoV. In order to not be super intrusive the snapshotter class I created I believe is likely to be too specialized for Btrfs snapshots. Since there is no other implementation is also hard to figure out what would be good and solid abstraction. So far the suggested interface here is just three methods:
* `create_first_snapshot`, which creates the subvolumes structure for snapper (`<root>/.snapshots` and `<root>/.snapshots/1/snapshot`, but not the parent `@`, it has to be there already) so we can sync to the new empty "snapshot", so that it also mounts the `.snapshots` into `.snapshots/1/snapshot/.snapshots`.
* `get_default_snapshot_name`, returns the default snapshot name, aka returns `<root>/.snapshots/1/snapshot` for snapper implementation. This is how the volume manage gets the path to just created new snapshot.
* `setup_first_snapshot`, after data sync to snapshot the new root is ready to be configured from snapper PoV. This method in snapper's implementation sets snapper into the new root.

The new snapshotter base class requires a parameter called `root_volume_name` which is essentially meant for `@` subvolume. I am not really satisfied with this, but so far since having this subvolume is not a snapper choice I opted to include it as an extra parameter. Probably there could not some other way to handle the this subvolume, so far I did not make it part of the snapper class implementation. Also because this subvolume is always created regardless of setting the root as an snapshot or not.

> I suggest that we squash commits as you see fit and I prepare a Staging python-kiwi package which will cause all integration tests to run. From there we can check the btrfs based ones to still look correct.

Sure, I am on it, gonna squash so you can prepare a built for the tests.

davidcassany:
@schaefi I squashed a couple of commits, I'd leave the rest as they are for the sake of a keeping a meaningful history in case someone needs to review it in the future. But I can squash them all if we prefer having a single commit to facilitate backports, revert or similar operations.

schaefi:
@davidcassany I really like the refactor done so far. I checked out your fork and branch from the current commit situation, rebased (without push) to the current main and created a Staging kiwi package from this state. You can find the package here

* Virtualization:Appliances:Staging/python-kiwi

Our integration tests are now building with your refactor effective. I'm waiting for the results.

We need to double check at least the following build results
* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-MicroOS

SUSE style and Fedora style

they should both work and look correct.

I would say if this is the case let's merge this PR

schaefi:
@davidcassany @Conan-Kudo All tests are passing. I took a closer look to the btrfs layouts of SUSE and Fedora and both integration tests are function like before. From my perspective this is good to go but I would prefer if someone else than me also gives the two integration tests a short boot and especially test the snapper layout on the SUSE side. Thanks

PR #2698, opened 2024-12-19
Fix config functions action

schaefi:
The action failed on the setup of the runtime because the upgrade of pip failed.

### DIFF
--- a/.github/workflows/ci-config-functions.yml
+++ b/.github/workflows/ci-config-functions.yml
@@ -33,9 +33,6 @@ jobs:
           path: |
             ~/.cache/pip
             ~/.cache/pypoetry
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-config-functions.yml
```diff
@@ -33,17 +33,13 @@ jobs:
           path: |
             ~/.cache/pip
             ~/.cache/pypoetry
-          key: pip-${{ hashFiles('.virtualenv.dev-requirements.txt') }}-${{ hashFiles('.virtualenv.requirements.txt') }}
-          restore-keys: |
-            -pip-
```
dcermak, line 38: Why did you remove this? It ensures that the action restores the correct cache

schaefi: These files don't exist anymore since we moved to poetry.

Conan-Kudo (approved):

PR #2699, opened 2024-12-19
Drop insecure and unsupported md5 digest

### LINKED ISSUE #2696: cannot build appliances on a FIPS enforcing host
## Problem description

`kiwi-ng system build --description kiwi-main/build-tests/x86/leap/test-image-disk --set-repo obs://openSUSE:Leap:15.6/standard --target-dir /tmp/myimage`

gives

```
[ INFO    ]: 12:45:27 | Creating disk image checksum
[ ERROR   ]: 12:45:27 | Unexpected error:
  File "/usr/bin/kiwi-ng", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/lib/python3.11/site-packages/kiwi/kiwi.py", line 68, in main
    App()
  File "/usr/lib/python3.11/site-packages/kiwi/app.py", line 35, in __init__
    task_class().process()
  File "/usr/lib/python3.11/site-packages/kiwi/tasks/system_build.py", line 352, in process
    result = image_builder.create()
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/kiwi/builder/disk.py", line 277, in create
    result = self.create_install_media(result)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/kiwi/builder/disk.py", line 510, in create_install_media
    install_image.create_install_iso()
  File "/usr/lib/python3.11/site-packages/kiwi/builder/install.py", line 175, in create_install_iso
    checksum.md5(self.squashed_contents.name + '/' + self.md5name)
  File "/usr/lib/python3.11/site-packages/kiwi/utils/checksum.py", line 84, in md5
    hashlib.md5(), self.source_filename
    ^^^^^^^^^^^^^
_hashlib.UnsupportedDigestmodError: [digital envelope routines] unsupported
```

can you please add `usedforsecurity=False` in the appropriate places if python is new enough? see https://docs.python.org/3/library/hashlib.html#hashlib.new

## Expected behaviour

building works and succeeds

see above

* KIWI version: KIWI (next generation) version 10.1.16
* Operating system host version: SUSE Linux Enterprise Server 16.0 (Beta1)
* Operating system target version: openSUSE 15.6

schaefi:
Decommission the Checksum.md5() method and move all places in code to sha256(). The md5 digest is considered insecure and has also been removed from hashlib as a supported digest. This Fixes #2696

### DIFF
--- a/doc/source/building_images/build_expandable_disk.rst
+++ b/doc/source/building_images/build_expandable_disk.rst
@@ -183,18 +183,18 @@ a target system:
           scp pxeboot.{exc_image_base_name_disk}.x86_64-{exc_image_version}.initrd PXE_SERVER_IP:/srv/tftpboot/boot/initrd
           scp pxeboot.{exc_image_base_name_disk}.x86_64-{exc_image_version}.kernel PXE_SERVER_IP:/srv/tftpboot/boot/linux
 
-3. Copy the disk image, MD5 file, system kernel, initrd and bootoptions to
+3. Copy the disk image, SHA256 file, system kernel, initrd and bootoptions to
    the PXE boot server.
 
    Activation of the deployed system is done via `kexec` of the kernel
    and initrd provided here.
 
-   a) Copy system image and MD5 checksum:
+   a) Copy system image and SHA256 checksum:
 
       .. code:: bash
 
           scp {exc_image_base_name_disk}.x86_64-{exc_image_version}.xz PXE_SERVER_IP:/srv/tftpboot/image/
-          scp {exc_image_base_name_disk}.x86_64-{exc_image_version}.md5 PXE_SERVER_IP:/srv/tftpboot/image/
+          scp {exc_image_base_name_disk}.x86_64-{exc_image_version}.sha256 PXE_SERVER_IP:/srv/tftpboot/image/
 
    b) Copy kernel, initrd and bootoptions used for booting the system via kexec:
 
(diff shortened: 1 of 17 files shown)


Conan-Kudo:
Note that [`isomd5sum`](https://github.com/rhinstaller/isomd5sum) only supports md5 (hence the name). This is the only supported media check implementation available outside of SUSE distributions, so that can be a problem if the kiwi-live dracut module doesn't support it. This pull request doesn't _seem_ to touch that part, but can you please check to make sure we don't accidentally break this?

Conan-Kudo (approved):

schaefi:
Hi, the issue is the same as in #2726 

If you build the image with a newer version of kiwi but still using the older dracut modules you run into this problem.
You can resolve this issue if you add the Builder repository to your image description:

https://build.opensuse.org/project/repository_state/Virtualization:Appliances:Builder/openSUSE_Tumbleweed

It should then take the kiwi dracut module packages from this repo and that will then be compatible with the kiwi builder you used.

In general we build all of kiwi in the Virt:App:Builder project. If you take kiwi and the dracut module packages from the same source it will be compatible. We also submit all of kiwi to the distribution. Meaning if you take all packages from the distribution it will also be compatible.

In this particular case a problem exists if you have a mix of builder and kiwi dracut module packages

I'm sorry for introducing this issue but I did not see a better way. md5 is unsupported and had to go away and there is no tight coupling between the kiwi builder on your host and the repository setup used to build the image. Also for dracut it's not a fatal error if it can't find md5 anymore. All you get is maybe a warning message from dracut in the log file of your kiwi build

I'm open for ideas

svenbeisiegel:
Hi @schaefi 
I did not spot the issue #2726 raised. But thank you very much for your response, it helped clarifying things.
I am now dracut and kiwi from the same source and this fixed it for me.

In general, I can fully understand getting rid of md5, so no worries.
Taking the recommandation into account to use all packages from the same distribution to avoid such issues in the future, is probably good enough of an idea.

PR #2700, opened 2024-12-20
Fix zypper repo and package gpgchecks setup

### LINKED ISSUE #2273: Gpg checks in zypper repo config disabled by default
## Problem description
As per the documentation [here](https://osinside.github.io/kiwi/concept_and_workflow/repository_setup.html) and [here](https://osinside.github.io/kiwi/image_description/elements.html#repository), you can set `repository_gpgcheck` and `package_gpgcheck`, and if they are omitted, the package manager's default is used instead. The second link also mentions: `If not set, no value is appended into the repository configuration file.`.

The code for zypper though does set these values to 0 if they are omitted, I'm not familiar with the codebase but these lines look to be the reason: https://github.com/OSInside/kiwi/blob/53f1c72f9f1514c990ab668d8d8efe4ec46470e4/kiwi/repository/zypper.py#L326

I end up with these lines in the repo files in `/etc/zypp/repos.d/...` in the created image:
```
repo_gpgcheck = 0
pkg_gpgcheck = 0
```
If I do `zypper lr` in the built image, it still shows it as having repo GPG checks enabled though, which I don't really understand, but those 2 lines in the repo config definitely look wrong.

Side note: `package_gpgcheck` is missing from the image description documentation.
## Expected behaviour
That kiwi-ng doesn't add any extra lines to the repo files.
I tried to make a minimal setup for this, here's the config.xml:
```
<image schemaversion="7.4" name="test">
	<description type="system"><author>test</author><contact>test</contact><specification>test</specification></description>
	<preferences><version>1</version><packagemanager>zypper</packagemanager><type image="kis"/></preferences>
	<repository type="rpm-md" alias="leap153" imageinclude="true"><source path="http://download.opensuse.org/distribution/leap/15.3/repo/oss/"/></repository>
	<packages type="bootstrap"><package name="openSUSE-release"/></packages>
	<packages type="image"><package name="dracut"/><package name="kernel-default"/></packages>
</image>
```
And then using kiwi-ng:
```shell
$ kiwi-ng system build --description tmp/ --target-dir tmp2
$ cat tmp2/build/image-root/etc/zypp/repos.d/leap153.repo
```

* KIWI version: 9.24.58
* Operating system host version: leap 15.2
* Operating system target version: leap 15.3

schaefi:
Only set a value for repo_gpgcheck, pkg_gpgcheck in the zypper repo config file if the respective kiwi attributes repository_gpgcheck and/or package_gpgcheck are set in the kiwi image description. If not provided do not set any value. This Fixes #2273

### DIFF
--- a/doc/source/concept_and_workflow/repository_setup.rst
+++ b/doc/source/concept_and_workflow/repository_setup.rst
@@ -58,13 +58,13 @@ following optional attributes:
   not specified, {kiwi} generates an alias name using hex representation from
   uuid4.
 
-- `repository_gpgcheck`: Specifies whether the repository is
-  configured to perform repository signature validation. If not set, the
-  package manager's default is used.
+- `repository_gpgcheck`: Specify whether or not this specific repository
+  values the result of the repository signature validation.
+  The default value is `false`.
 
-- `package_gpgcheck`: Boolean value that specifies whether each package's
-  GPG signature is verified. If omitted, the package manager's default
-  is used.
+- `package_gpgcheck`: Specify whether or not this specific repository
+  values the result of the package signature validation for each
+  package taken from this repository. The default value is `false`.
 
 - `components`: Distribution components used for `deb` repositories. Default is `main`.
 
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -1413,13 +1413,19 @@ imageonly="true|false"
   the value is set to false
 
 repository_gpgcheck="true|false"
-  Specifies whether or not this specific repository is configured to
-  to run repository signature validation. If not set, no value is
-  appended into the repository configuration file. If set the
-  relevant key information needs to be provided on the {kiwi}
-  commandline using the `--signing-key` option or via the `<signing>`
-  element as part of the `<repository><source>` setting in the
-  image description.
+  Specifies whether or not this specific repository values the result
+  of the repository signature validation. If not set, the default is
+  `false`. If set the relevant key information needs to be provided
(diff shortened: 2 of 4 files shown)


schaefi:
@Conan-Kudo please don't merge until the integration tests has run with this change applied. I'm still waiting for them to finish

Conan-Kudo:
I don't think we want this behavior. That would mean that we'd inherit a behavior from how the distribution packages Zypper. Instead, we should probably propagate default values from kiwi into Zypper configs (and probably the same for DNF too).

Conan-Kudo:
For example, if we did what this PR does on Fedora with DNF, all image builds would be broken because GPG checks are on by default, not off. It's a seriously bad idea to not express a default setting.

schaefi:
> I don't think we want this behavior. That would mean that we'd inherit a behavior from how the distribution packages Zypper. Instead, we should probably propagate default values from kiwi into Zypper configs (and probably the same for DNF too).

This is what we do today, kiwi decides for the default value

schaefi:
>  It's a seriously bad idea to not express a default setting.

ok thinking twice this might be true, but then the docs about this setting must be corrected

schaefi:
@Conan-Kudo I updated the docs to match the implementation. I'm sure what I wrote can be improved... Thanks

Conan-Kudo:
Yeah, this makes sense. Can you drop the commits related to code so that we only have the documentation commit?

Conan-Kudo:
There's no point in merging a code commit that is reverted in the same PR. 😉

Conan-Kudo (approved):

PR #2703, opened 2025-01-09
Added LUKS reencryption support

schaefi:
Added `rd.kiwi.oem.luks.reencrypt` boot option consumed by the kiwi-repart dracut module. For OEM LUKS2 encrypted disk images. If set, reencrypts the disk prior an eventual resize and therefore creates a new key pool and master key. The reencryption is advisable if the image binary is not protected. With access to the image binary it's possible to extract the luks header which then allows to decrypt the data unless it was reencrypted. The reencryption process only runs if the checksum of the luks header still matches the one from the original disk image. Be aware that the reencryption will ask for the passphrase if the image has been built with an initial luks passphrase.

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-luks/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-luks/appliance.kiwi
@@ -1,11 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
-
+<!-- OBS-Profiles: @BUILD_FLAVOR@ -->
 <image schemaversion="7.5" name="kiwi-test-image-luks">
     <description type="system">
         <author>Marcus Schäfer</author>
         <contact>ms@suse.com</contact>
         <specification>Disk full encryption disk test build</specification>
     </description>
+    <profiles>
+        <profile name="Insecure" description="Encrypted no reencryption"/>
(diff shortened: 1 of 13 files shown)


schaefi:
Integration test build for testing the reencryption is available here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-luks

davidcassany (approved):
LGTM 


Nit: Just wondering if `sed` shouldn't be required as part of the `kiwi-lib`. I know is is only used as part of the kiwi-repart through the luks library code, but from a dependency management I find the current approach obscure... I wonder if some of this logic should be organized differently in terms of dracut modules.

If I'd have to maintain this dracut module I'd grep around `kiwi-repart` module to check the tools in requires and remove anything that is not called within the module, assuming each module requires the tools it uses.

I know there is no easy solution and dracut modules proliferation is also a maintenance burden.

Probably I'd add a comment in code the explain this odd dependency if reorganizing the logic in a devoted module is too complex.

schaefi:
> Nit: Just wondering if sed shouldn't be required as part of the kiwi-lib

yeah you are absolutely right. I put them to the wrong place. Thanks and fixed

PR #2705, opened 2025-01-13
Rename btrfs root is snapshot

schaefi:
Rename btrfs_root_is_snapshot to btrfs_root_is_snapper_snapshot. This happens in preparation for the changes suggested in #2697  where we want to get rid of snapper specific btrfs code which  will be available in snapper natively soon. To make sure a btrfs  layout specific to snapper(and SUSE), the implicitly used attribute  named btrfs_root_is_snapshot now becomes explicit and its new name will indicate that snapper sits behind it. Along with the rename a XSLT stylesheet to automatically convert the old name  into the new name for schema v8.3 will be performed.


davidcassany (approved):
LGTM :+1:

PR #2706, opened 2025-01-13
Add support for reading optional package manager env file

schaefi:
If there is a file .zypp.env in the root of the image tree it will be read and put into the caller environment for the zypper package and repository manager. There are features in zypper which can only be used via env variables. This Fixes bsc#1235448


Conan-Kudo:
This is also a thing with DNF, so maybe we should genericize this.

schaefi:
> This is also a thing with DNF and APT, so maybe we should genericize this.

yeah we could. Then we need to agree on a generic name for the environment file. How about `.package_manager.env` ?
And I realized I forgot to document it

Conan-Kudo:
Let's prefix it with `.kiwi` though, so `.kiwi.package_manager.env`.

schaefi:
@Conan-Kudo ok done

Conan-Kudo (approved):

PR #2713, opened 2025-01-20
Added disk validation for duplicate installs

schaefi:
Installing the same image to different storage disks on the same machine creates device conflicts with unexpected side effects. This commit adds a validation based on the PTUUID of the disk image to check if another device on the system has the same ID and if yes, does not allow to install the image again including a message which device takes the same identifier. This references bsc#1228741


davidcassany (approved):
LGTM

I am wondering why this check is only executed when `rd.kiwi.oem.disk.consistency` flag is present. I assume this is a matter of not introducing a new behavior on already existing descriptions, which I fully agree from that angle, but I am wondering and curious to know if there is some use case I am not seeing. I struggle to see a use case where multiple disks are based on different KIWI OEM image deployments, how this could be a sane setup?

PR #2720, opened 2025-01-27
Use multipath child instead of parent device

aaannz:
On multipath systems we need to find underlying child device instead of using parent device.
This prevents listing all parent devices for a multipath device.


schaefi (approved):
Thanks, to me this looks good :+1: 

@davidcassany can you also have a look at this one ? another pair of eyes on dracut code is much appreciated :) Thanks

davidcassany (approved):
LGTM :+1:

schaefi:
@aaannz Thanks, I'm going to merge this one and wait for the integration tests to build. After a quick check of the installation mode I'll go and look into your sle15 port PR.

PR #2721, opened 2025-01-28
Add custom CA runtime config option

aaannz:
This commit adds custom_certificates:ca option to the kiwi.yml runtime configuration.

Setting this configuration option will then inject CA certificate to the chroot after bootstrap phase. System phase then can connect to repositories requiring custom CA.

### Draft note

I open this as a draft PR to discuss if this is viable approach.
Our use case is with SUSE Multi-Linux Manager build hosts. We usually have custom PKI in place so to access repositories provided by SMLM we need to have custom CA certificate. Previously we accomplished this by using bootstrap package added by `--add-bootstrap-package` option, however we are moving away from building RPMs with our certificate and need another way to pass plain certificate file.

We cannot use usual means like specifying file in template `root` directory as we need to call update-ca-certificates to load the certificate and also do not want to modify our users templates.

What do you think about this? If this is ok, I will add unit tests to this PR.

### DIFF
--- a/kiwi/runtime_config.py
+++ b/kiwi/runtime_config.py
@@ -416,6 +416,37 @@ class RuntimeConfig:
         )
         return disabled_checks or ''
 
+    def get_custom_ca_cert(self):
+        """
+        Returns a custom CA certificate to be injected to the chroot.
+        The value is certificate itself:
+
+        custom_certificates:
+          - ca: |
+            -----BEGIN CERTIFICATE-----
+            MIICGzCCAaGgAwIBAgIQQdKd0XLq7qeAwSxs6S+HUjAKBggqhkjOPQQDAzBPMQsw
+            CQYDVQQGEwJVUzEpMCcGA1UEChMgSW50ZXJuZXQgU2VjdXJpdHkgUmVzZWFyY2gg
+            R3JvdXAxFTATBgNVBAMTDElTUkcgUm9vdCBYMjAeFw0yMDA5MDQwMDAwMDBaFw00
+            MDA5MTcxNjAwMDBaME8xCzAJBgNVBAYTAlVTMSkwJwYDVQQKEyBJbnRlcm5ldCBT
+            ZWN1cml0eSBSZXNlYXJjaCBHcm91cDEVMBMGA1UEAxMMSVNSRyBSb290IFgyMHYw
+            EAYHKoZIzj0CAQYFK4EEACIDYgAEzZvVn4CDCuwJSvMWSj5cz3es3mcFDR0HttwW
+            +1qLFNvicWDEukWVEYmO6gbf9yoWHKS5xcUy4APgHoIYOIvXRdgKam7mAHf7AlF9
+            ItgKbppbd9/w+kHsOdx1ymgHDB/qo0IwQDAOBgNVHQ8BAf8EBAMCAQYwDwYDVR0T
+            AQH/BAUwAwEB/zAdBgNVHQ4EFgQUfEKWrt5LSDv6kviejM9ti6lyN5UwCgYIKoZI
+            zj0EAwMDaAAwZQIwe3lORlCEwkSHRhtFcP9Ymd70/aTSVaYgLXTWNLxBo1BfASdW
+            tL4ndQavEi51mI38AjEAi/V3bNTIZargCyzuFJ0nN6T5U6VR5CmD1/iQMVtCnwr1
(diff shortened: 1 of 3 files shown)


schaefi:
Hi @aaannz, thanks for your effort. I do believe we already have some options to achieve what you are looking for.
So you want to inject a CA certificate to the new root system after bootstrap. To do this you already have the following options:

> You can provide your CA in a package and add that to the `<packages type="bootstrap">` section

Not an option because you don't want to package it

> You can add a `post_bootstrap.sh` script to your image description which copies the CA to the root. See: https://osinside.github.io/kiwi/concept_and_workflow/shell_scripts.html

Not an option because you don't want to change the image description

I don't have a strong opinion with adding a third method via the runtime configuration file but I believe it looks more like abusing the runtime config :) That config file exists to influence the behavior of the builder and not to modify the root tree.

From that perspective I'd like to seek input from @Conan-Kudo and @davidcassany too.

If we support this it needs to be handled in its own namespace in the config file, but that you already do with `custom_certificates`

So let's see what others have to say

Thanks

aaannz:
> I don't have a strong opinion with adding a third method via the runtime configuration file but I believe it looks more like abusing the runtime config :) That config file exists to influence the behavior of the builder and not to modify the root tree.

Right. I have no problem adding this as a command line option instead of runtime config, similarly like we have `--add-bootstrap-package` I can add `--add-custom-ca` option.

Or I can add removal of this runtime config added certificate at the end of the prepare phase.

Conan-Kudo:
Why wouldn't we just expose [the yum repo settings for certificates](https://www.mankier.com/5/dnf5.conf#Options_for_Both_%5BMain%5D_and_Repo-sslcacert)?

aaannz:
Interesting idea. So
1) it would remain to be configurable in runtime config
2) if set, kiwi would inject cert file to some temporary file
3) if set, zypper/dnf would use this file for `sslcacert` option

Correct?

Conan-Kudo:
Yes. This would also be a nice hack to work around the lack of support for `rhsm://` protocol in kiwi too.

Conan-Kudo:
> Interesting idea. So
> 
> 1. it would remain to be configurable in runtime config

It would be in the kiwi description XML, not `kiwi.yml`.

> 2. if set, kiwi would inject cert file to some temporary file
> 3. if set, zypper/dnf would use this file for `sslcacert` option
> 

You would set it in repository section and it would make kiwi copy it into a temporary file and configs would be generated as appropriate.

schaefi:
@aaannz I wondered what is wrong with the packaging solution ? Providing the CA as a package including the needed post install actions seems like the cleanest and best maintainable solution to this issue. In addition you can pass a package to be installed as a commandline parameter to kiwi as you seem to do it right now. I wondered why you want to move away from this concept ?

> I don't have too much of an issue with the runtime config for CA bundles, but there's no way I'd want this without it also being properly possible in the kiwi description XML. It is common for SSL certs to be needed for repos anyway.

That's true. Currently we have a declarative way to handle signing certificates for repositories. kiwi has no declarative way to import system certificates to the key pool. This is done via packages and tools of the distribution (update-ca-certificates). Therefore if we add a generic way to import certificates it should be done declarative in the image description first e.g.

```xml
<certificates>
    <certificate name="foo"/>
</certificates>
```

along with feature we allow this setting also on the commandline. In the end you can then say

```
kiwi-ng ... --ca-certificate foo ...
```

and it will do the same as with the section in the XML description. Same concept as we have for packages and other elements.

That should provide a solution acceptable to @aaannz, please correct me when wrong. It will also keep the builder configuration free from code that it should imho not have and it is compatible with our rules to describe an image.

Thoughts ?

Conan-Kudo:
That makes sense to me. This pattern is much more palatable than putting stuff in the "mystery meat" `kiwi.yml` config file.

aaannz:
Thanks for all the feedback, I will get into that after our product freeze, approx in 2 months.

schaefi:
> Thanks for all the feedback, I will get into that after our product freeze, approx in 2 months.

@aaannz Can you share a quick update on plans regarding this one ? Thanks much

schaefi:
@aaannz I guess there is no further work happening on this topic ? Without a sign of live I'm going to close it

aaannz:
I agree, let's close this. I don't have time to work on this and cannot plan too much to the future. For now we'll continue using package based solution.

PR #2725, opened 2025-01-30
Make sure copy actions does not drop context

### LINKED ISSUE #2709: /usr/sbin/grub2-install is missing the correct SELinux label
Even with kiwi v10.2.5, the system isn't fully labelled: `system_u:object_r:unlabeled_t:s0 ./usr/sbin/grub2-install` remains.

Either `setfiles` needs to run after the `grub2-install` -> `grub2-install.orig` and back dance or the cp/mv invocations need to be SELinux aware, i.e. `cp --preserve=context`.

schaefi:
Use shutil.copy2 to copy files preserving their attributes in the grub BootLoader space.
This Fixes #2709

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -610,7 +610,7 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
                     config_file, grub_config_file_for_efi_boot
                 )
             )
-            shutil.copy(
+            shutil.copy2(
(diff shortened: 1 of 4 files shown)


schaefi:
@Vogtinator From the python documentation shutil.copy2 should also preserve the context. Moving a file via mv should also keep the context as I only see an option to reset the moved file to its default context which I believe is unwanted

Conan-Kudo (approved):

PR #2727, opened 2025-01-30
Add rd.kiwi.dialog.timeout option

### LINKED ISSUE #2718: Disable timeout for checksum test failure
We have noticed that if the checksum test at the end of the installation from an ISO built using kiwi fails, the installation process will wait on the checksum test failure screen for user confirmation for a short period of time (less than 1 minute), but eventually times out and will continue booting into the OS. This means that unless someone is sitting through the whole installation process (which can take hours depending on the ISO size and hardware specs) and watches the checksum verification process and makes sure it passes, the checksum verification can actually fail because of some data corruption that is not big enough to break the OS boot, and by the time user gets back to the system, the checksum test failure screen could have timed out and the corrupted OS installation could have booted up, leaving the user confused as to why there are random failures happening across different parts of the OS.

Is there a way to disable the timeout of the checksum failure screen, and force the installation to stall on that screen until the failure is acknowledged through user input?

schaefi:
Allow to configure the timeout value for dialogs displayed by the kiwi dracut code. By default the timeout is set to 60 seconds. With the special value "off" the dialog will never timeout. This Fixes #2718

### DIFF
--- a/doc/source/concept_and_workflow/customize_the_boot_process.rst
+++ b/doc/source/concept_and_workflow/customize_the_boot_process.rst
@@ -194,6 +194,12 @@ the available kernel boot parameters for these modules:
   Activates the debug log file for the {kiwi} part of
   the boot process in :file:`/run/initramfs/log/boot.kiwi`.
 
+``rd.kiwi.dialog.timeout=seconds|off``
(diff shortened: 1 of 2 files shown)


davidcassany (approved):
LGTM :+1:

PR #2729, opened 2025-01-30
Classify missing chkstat as debug message

### LINKED ISSUE #2711: Be smarter about chkstat calls ?
Related to https://github.com/OSInside/kiwi/issues/895

Building a Debian image, you get:

> [ WARNING ]: 16:38:20 | chkstat not found in image. File Permissions Check skipped

I'm no expert, but Mr Google seems to suggest `chkstat` and the underlying `/etc/sysconfig/security` is a SuSe thing ?  A brief search of the Debian package repo doesn't reveal any package that contains `chkstat` either ?

Maybe you could check for the existence of `/etc/sysconfig/security` before attempting a call to chkstat ?

schaefi:
chkstat is a distribution specific tool. If it is
present we use it, if not we don't but it's not worth a warning. This Fixes #2711

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -356,7 +356,7 @@ class SystemSetup:
                 ['chroot', self.root_dir, 'chkstat', '--system', '--set']
             )
         else:
-            log.warning(
+            log.debug(
                 'chkstat not found in image. File Permissions Check skipped'
             )
 


Conan-Kudo:
Why is this tool in here at all? This is a SUSE specific quirk for the `permissions` package?

schaefi:
> Why is this tool in here at all? This is a SUSE specific quirk for the `permissions` package, and should be done from config.sh for SUSE distributions.

This is an artifact from the old days and I can't just get rid of it. Also running from config.sh would be too early in the process because permissions could be changed on the way to the final image. Thus the invocation happens as late as possible.
It does not hurt anyone as this exists actually since the port from the old(perl kiwi) to the new one.

I agree the warning message should go away which is what the PR does, but the rest needs to stay to avoid negative side effects for existing builds.

PR #2731, opened 2025-02-09
Print env in github workflow

schaefi:
Print the envrionment to see if SOURCE_DATE_EPOCH is set


schaefi:
I added this PR only to check the environment settings of the github runners

PR #2747, opened 2025-03-06
distutils sysconfig is deprecated

schaefi:
Move to sysconfig module

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -29,7 +29,7 @@
 %endif
 
 %if %{undefined python3_sitelib}
-%global python3_sitelib %(%{__python3} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
+%global python3_sitelib %(%{__python3} -c "import sysconfig; print(sysconfig.get_path('platlib'))")
 %endif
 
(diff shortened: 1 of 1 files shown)


package/python-kiwi-spec-template
```diff
@@ -29,7 +29,7 @@
 %endif
 
 %if %{undefined python3_sitelib}
-%global python3_sitelib %(%{__python3} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
+%global python3_sitelib %(%{__python3} -c "import sysconfig; print(sysconfig.get_path('platlib'))")
```
Conan-Kudo, line 32: Does this module exist on all Debian/Ubuntu versions we build against right now?

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: in standard python for 3.9 and older:
* https://docs.python.org/fr/3.9/library/sysconfig.html

as we set a constraint to python >= 3.9 I believe this change should be safe

Conan-Kudo (approved):

PR #2750, opened 2025-03-13
Add check for threading option in setfiles

### LINKED ISSUE #2749: setfiles: invalid option -- 'T'
[Threading has been added relatively recently to `setfiles`](https://github.com/SELinuxProject/selinux/commit/93902fc8340f8a6ee5ba69ccb150d47918aad226) so kiwi fails with `setfiles: invalid option -- 'T'` on EL8:

```
[ DEBUG   ]: 06:48:59 | Looking for setfiles in /data/img-boot/build/image-root/usr/local/sbin:/data/img-boot/build/image-root/usr/local/bin:/data/img-boot/build/image-root/usr/sbin:/data/img-boot/build/image-root/usr/bin:/data/img-boot/build/image-root/sbin:/data/img-boot/build/image-root/bin
[ INFO    ]: 06:48:59 | Processing SELinux file security contexts
[ DEBUG   ]: 06:48:59 | Looking for chroot in /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
[ DEBUG   ]: 06:48:59 | EXEC: [chroot /data/img-boot/build/image-root setfiles --help]
[ DEBUG   ]: 06:48:59 | Looking for chroot in /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
[ DEBUG   ]: 06:48:59 | EXEC: [chroot /data/img-boot/build/image-root setfiles -T0 -F -p -c /etc/selinux/targeted/policy/policy.31 -e /proc -e /sys -e /dev /etc/selinux/targeted/contexts/files/file_contexts /]
[ DEBUG   ]: 06:48:59 | EXEC: Failed with stderr: setfiles: invalid option -- 'T'
usage:  setfiles [-diIDlmnpqvEFW] [-e excludedir] [-r alt_root_path] [-c policyfile] spec_file pathname...
usage:  setfiles [-diIDlmnpqvEFW] [-e excludedir] [-r alt_root_path] [-c policyfile] spec_file -f filename
usage:  setfiles -s [-diIDlmnpqvFW] spec_file
```

schaefi:
Not all versions of setfiles in the supported distribution matrix supports the -T option. This commit adds a capability check prior using the option. This Fixes #2749

### DIFF
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -557,6 +557,14 @@ class SystemSetup:
         for devname in Defaults.get_exclude_list_for_non_physical_devices():
             exclude.append('-e')
             exclude.append(f'/{devname}')
+        # check for -T in setfiles
+        setfiles_supports_threads = False
+        if CommandCapabilities.has_option_in_help(
+            'setfiles', 'setfiles [-diIDlmnpqvCEFWT]', ['--help'],
+            root=self.root_dir, raise_on_error=False, silent=True
+        ):
+            setfiles_supports_threads = True
+
         # setfiles doesn't come with a stable command API and older versions
         # needs a different invocation syntax. On older versions the usage
         # information explicitly lists "setfiles -c policyfile" which is not
@@ -585,16 +593,19 @@ class SystemSetup:
                 ]
             )
         else:
-            Command.run(
-                [
-                    'chroot', self.root_dir, 'setfiles',
-                    '-T0', '-F', '-p', '-c', self._find_selinux_policy_file(
-                        self.xml_state.build_type.get_selinux_policy() or 'targeted'
-                    )
-                ] + exclude + [
-                    security_context_file, '/'
-                ]
-            )
+            setfiles = [
+                'chroot', self.root_dir, 'setfiles'
+            ]
+            if setfiles_supports_threads:
+                setfiles.append('-T0')
+            setfiles += [
+                '-F', '-p', '-c', self._find_selinux_policy_file(
+                    self.xml_state.build_type.get_selinux_policy() or 'targeted'
+                )
+            ] + exclude + [
+                security_context_file, '/'
+            ]
+            Command.run(setfiles)
 
     def setup_selinux_file_contexts(self) -> None:
         """
--- a/test/unit/system/setup_test.py
+++ b/test/unit/system/setup_test.py
@@ -1646,7 +1646,14 @@ class TestSystemSetup:
     def test_set_selinux_file_contexts_new_version(
         self, mock_os_scandir, mock_command, mock_has_option_in_help
     ):
-        mock_has_option_in_help.return_value = False
+        def has_option_in_help(
+            call, flag, help_flags, root, raise_on_error, silent
+        ):
+            if flag == 'setfiles [-diIDlmnpqvCEFWT]':
+                return True
+            return False
+
+        mock_has_option_in_help.side_effect = has_option_in_help
         mock_os_scandir.return_value = self.selinux_policies
         self.setup.set_selinux_file_contexts('security_context_file')
         mock_command.assert_called_once_with(


Conan-Kudo:
Who doesn't support this still?

Conan-Kudo:
The bug references RHEL 8, which is not a supported target for kiwi 10. I don't even ship kiwi 10 to EPEL 8.

schaefi:
> Who doesn't support this still?

See the associated issue report: #2749

schaefi:
> The bug references RHEL 8, which is not a supported target for kiwi 10. I don't even ship kiwi 10 to EPEL 8.

Hmm, good point. From the associated issue request it seems like v10 was shipped to this target

Conan-Kudo:
We really should make it more public what our supported distributions are somewhere...

Currently the tested/supported options are:

* Fedora Linux 40+
* openSUSE Tumbleweed (after June 2023)+
* SLE/Leap 15.5+
* CentOS Stream/RHEL 9+
* Ubuntu 24.04+
* Debian 12+
* Arch Linux (after January 2024)+

Conan-Kudo:
I'm wary of this code mostly because if setfiles changes again we have to handle that too.

schaefi:
> I'm wary of this code mostly because if setfiles changes again we have to handle that too.

I agree. I added the patch because I thought there seems to be some delivery of kiwi 10 to EL8.

schaefi:
> We really should make it more public what our supported distributions are somewhere...
> 
> Currently the tested/supported options are:
> 
> * Fedora Linux 40+
> * openSUSE Tumbleweed (after June 2023)+
> * SLE/Leap 15.5+
> * CentOS Stream/RHEL 9+
> * Ubuntu 24.04+
> * Debian 12+
> * Arch Linux (after January 2024)+

It would be great if you can add that as a chapter to the documentation. That way its exposed in the docs and we can point users to it. Thoughts ?

Conan-Kudo:
I'm not delivering it in EPEL, are we delivering it in the OBS project somehow?

Conan-Kudo:
> > We really should make it more public what our supported distributions are somewhere...
> > Currently the tested/supported options are:
> > 
> > * Fedora Linux 40+
> > * openSUSE Tumbleweed (after June 2023)+
> > * SLE/Leap 15.5+
> > * CentOS Stream/RHEL 9+
> > * Ubuntu 24.04+
> > * Debian 12+
> > * Arch Linux (after January 2024)+
> 
> It would be great if you can add that as a chapter to the documentation. That way its exposed in the docs and we can point users to it. Thoughts ?

Sure, that makes sense to me.

schaefi:
> I'm not delivering it in EPEL, are we delivering it in the OBS project somehow?

Nope, the OBS project only builds for publicly available distributions

schaefi:
> > The bug references RHEL 8, which is not a supported target for kiwi 10. I don't even ship kiwi 10 to EPEL 8.
> 
> AFAICT `setfiles` runs chrooted inside the image, so it depends on the distro you're building. I can reproduce this building a Rocky 8.10 image on Fedora 41.

Yes this is true and would probably also apply to centos8 and other older distributions outside the kiwi v10.x.x support matrix. In general I have no objections to add patches that allows building for older targets. I'm currently just thinking how harmful this one could be with regards to further changes on `setfiles` I must admit the story on setfiles is not a good one. The tool didn't had a `--version` information which we usually use to declare capabilities, so we had to base some tests on the usage output. I believe all that is not following a stable API and as such yet another help output based capability check could become an issue on newer versions of setfiles that maybe has changed the help output or something.

How important is this change for you ? Could you also live with a custom patch that you carry for old targets ?

Thanks

zzzoom:
> How important is this change for you ? Could you also live with a custom patch that you carry for old targets ?

We'll keep building our Rocky 8.10 images with kiwi 9, but who knows what else will break between now and 2029 when it finally goes EOL. We'd probably switch to whatever warewulf is doing if that happens.

schaefi:
> > How important is this change for you ? Could you also live with a custom patch that you carry for old targets ?
> 
> We'll keep building our Rocky 8.10 images with kiwi 9, but who knows what else will break between now and 2029 when it finally goes EOL. We'd probably switch to whatever warewulf is doing if that happens.

ok if I understand correctly that's a workaround you could live with

> but who knows what else will break between now and 2029
right, nobody can tell I guess ;) If it's something that can be fixed in the v9 code base we could patch it there.

PR #2751, opened 2025-03-19
Move commandline interface to typer

schaefi:
This commit moves the kiwi commandline and plugin interface to `typer` and drops the use of `docopt`. Typer is based on python type hints and allows for a more modern way to define Cli interfaces and auto completion, even for plugins. Also docopt is split into an old and a next generation variant and allows us to get rid of spec file hacks as well as documentation rendering hacks.


schaefi:
There are still open tasks before we can consider this change to go in 

- [x] stackbuild plugin needs cli.py: TBD by @schaefi 
- [x] box plugin needs cli.py: TBD by @schaefi 
- [x] is python-typer package available in all distros ?

schaefi:
@Conan-Kudo I'm mostly interested in getting the open questions from the checklist answered, so no need to do an in-depth review of the code changes as it's time consuming. Thanks much

Conan-Kudo:
We can't drop Python 3.9 support since that's required for CentOS/RHEL 9.

schaefi:
> We can't drop Python 3.9 support since that's required for CentOS/RHEL 9.

I found a fix to make it work on 3.9 too

schaefi:
@Conan-Kudo I finished the Cli change and also updated the plugins. The plugins continues to support both docopt and typer such that they can continue to work also with older versions of kiwi. The upstream kiwi should change to typer and the interface is a drop-in-replacement such that there should be no difference in the former caller semantic.

The only open question is if the typer python module is available in all distros we support. I did a short test on Virtualization:Appliances:Staging/python-kiwi and it would build the package for all targets that exists there except for ALP which is an issue that I need to resolve prior merge

Also I appreciate your feedback with regards to the other OS'es

Thanks

pyproject.toml
```diff
@@ -51,7 +52,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = "^3.9"
-docopt-ng = ">=0.9.0"
+typer = ">=0.15.2"
```
Conan-Kudo, line 55: Our existing matrix of distributions means we need to support as far back as `0.7.0` because that's what Debian 12 shipped with.

Conan-Kudo (changes requested):

pyproject.toml
schaefi: ok, thanks and done

schaefi:
I'll give it a test on Staging

schaefi:
I also figured typer 0.9.0 only works for python version < 3.8 because it imports `typing_extensions` which no longer exists starting with py3.8. This is especially a problem for Leap who added the python 3.11 stack and we only provide kiwi for Leap for 3.11 (the other python version there is 3.6 which is too old for kiwi v10.x.x) So typer 0.9.0 doesn't play with python 3.11 but fortunately the backports project offers the latest typer, so I don't see an issue here

schaefi:
However, as we said kiwi works with python >= 3.9 we actually cannot stand the requirement for typer >= 0.9.0 and should move to a typer that supports python >= 3.9. Which version of typer that is I don't know though

schaefi:
Overall such a change is a pita but I had hoped with the kiwi v10.x.x constraints it could be done... maybe a mistake ?

Conan-Kudo:
We could consider this for kiwi 11 and decide we do a kiwi 11 next year?

schaefi:
> We could consider this for kiwi 11 and decide we do a kiwi 11 next year?

ok, I did that on this PR and the plugin PRs. btw are Fedora 38 and xUbuntu_23.04 still supported targets ?

Conan-Kudo:
No. Our minimum for Fedora is currently Fedora 40. And Ubuntu is at 24.04.

schaefi:
> No. Our minimum for Fedora is currently Fedora 40. And Ubuntu is at 24.04.

ok I dropped those. For the Debian 12 part I added an updated typer to the Staging project which allows me to finally run all integration tests with a changed kiwi that contains this new cli interface. All tests built which is good.

So I will keep this and the plugin PRs open with a kiwi-11 label set to be merged in 2026

schaefi:
- [x] port --ca-cert option 
- [x] port --ca-target-distribution

schaefi:
- [x] port --add-repo
- [x] port --add-repo-credentials

from the `image info` task

PR #2753, opened 2025-03-24
Allow stderr data in CommandProcess

### LINKED ISSUE #2748: Capture stderr output for logged command output
We apparently do not capture stderr output, which resulted in https://github.com/rpm-software-management/dnf5/issues/2116.

We should change this because having information hidden from us in logs can make diagnosing issues much more difficult.

schaefi:
Enhance poll_show_progress() method to allow polling on stderr data too. The new parameter with_stderr is used together with the dnf5 package manager. dnf5 has changed in a way that a lot of useful information during the install of packages is printed to stderr. From my perspective a clear regression to former behavior but we can fix this in kiwi to poll on both channels. This Fixes #2748

### DIFF
--- a/kiwi/command_process.py
+++ b/kiwi/command_process.py
@@ -18,6 +18,9 @@
 import logging
 from collections import namedtuple
 from kiwi.command import CommandCallT
+from typing import (
+    NamedTuple, List, Callable
+)
 
 # project
 from kiwi.utils.codec import Codec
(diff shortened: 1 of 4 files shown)


schaefi:
@Conan-Kudo the tests are failing outside kiwi some complains regarding travis-sphinx :(

Conan-Kudo (approved):

PR #2756, opened 2025-03-28
Introduce SELinux policy module for kiwi

Conan-Kudo:
This policy module ensures that kiwi is set up to function properly in SELinux enforcing mode.

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -55,6 +55,12 @@
 %global develsuffix devel
 %endif
 
+%if 0%{?fedora} || 0%{?rhel} || 0%{?suse_version} >= 1600
+%bcond_without selinux
+%else
+%bcond_with selinux
+%endif
+
 Name:           python-kiwi
 Version:        %%VERSION
 Provides:       kiwi-schema = 8.1
@@ -456,6 +462,9 @@ Requires:       python%{python3_pkgversion}-docopt-ng
 Requires:       python%{python3_pkgversion}-lxml
 Requires:       python%{python3_pkgversion}-requests
 Requires:       python%{python3_pkgversion}-xmltodict
+%if %{with selinux}
+Requires:       (kiwi-selinux = %{version}-%{release} if selinux-policy)
+%endif
 %if ! (0%{?rhel} && 0%{?rhel} < 8)
 Recommends:     kiwi-man-pages
 %endif
@@ -695,6 +704,21 @@ BuildArch:      noarch
 Bash command line completion support for python-kiwi - completion
 of subcommands, parameters and keywords for the kiwi-ng command.
 
+%if %{with selinux}
+%package -n kiwi-selinux
+Summary:        SELinux module for kiwi-ng
+License:        GPL-2.0-or-later
+BuildArch:      noarch
+BuildRequires:  selinux-policy
+BuildRequires:  selinux-policy-devel
+BuildRequires:  make
+%{?selinux_requires}
+
+%description -n kiwi-selinux
+This package provides the SELinux policy module to ensure kiwi-ng
+runs properly under an environment with SELinux enabled.
+%endif
+
 %prep
 %setup -q -n kiwi-%{version}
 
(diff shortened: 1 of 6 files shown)


Conan-Kudo:
This is a fuller alternative to #2757 that lets us easily add more things in the future.

schaefi:
Really nice, thanks for this work :+1: I'm pretty dumb regarding selinux policies. Would be great if we get some reviewer with knowledge in that area. Otherwise I'm perfectly fine adding it the way you did it because I believe you tested it to work :)

ca-hu:
question: why add this as separate module here instead of in the fedora policy that is used by basically everyone? usually this only makes sense when the maintainers of the project are well-versed in selinux and can maintain this module long-term, otherwise you will run into more issues

also it is a bit odd that the module is reusing an existing file context, usually modules define their own types and rules around their type

also: what exactly is breaking?

Conan-Kudo:
> question: why add this as separate module here instead of in the fedora policy that is used by basically everyone? usually this only makes sense when the maintainers of the project are well-versed in selinux and can maintain this module long-term, otherwise you will run into more issues

Well, I *am* well-versed in SELinux. I've done a fair bit of SELinux policy work over the years in both Fedora and openSUSE. And as a team, we do need to develop more skill with SELinux anyway, since not knowing how to deal with it is bad for us as an image build tool.

But there are two big reasons:
* We need this to work across all distributions and releases using SELinux. And some of those would not get a policy update with this for a long time (if ever).
* We need a "fast path" since kiwi releases much more frequently than the policy package.

This does not preclude contributing it into fedora-selinux, and that will probably happen down the road as this is firmed up.

> also it is a bit odd that the module is reusing an existing type, usually modules define their own types and rules around their type

Yes, I will probably change to this approach, but I need to spend more time to write it.

Conan-Kudo:
> also: what exactly is breaking?

There have been requests from Fedora and CentOS to make kiwi work properly in SELinux enforcing mode as all the legacy image build tools in Fedora already do. After soliciting some advice from the OSBuild folks, I had identified an approach to resolve that problem. It also neatly resolves a problem nobody has yet noticed in openSUSE, in which you cannot build an image in SELinux enforcing mode.

schaefi:
@Conan-Kudo it would be so great to see this merged, it feels we are so close

Conan-Kudo:
Yeah, I want to resume this work asap.

schaefi:
> Yeah, I want to resume this work asap.

I think we are desperately waiting for this to come true ;)

Conan-Kudo:
Coming back to it now, don't worry. 👍🏾

schaefi:
> Coming back to it now, don't worry. 👍🏾

@Conan-Kudo Are you sure ;)

Conan-Kudo:
Yes, yes! :)

schaefi:
I guess we still want this, dropping stale label

Vogtinator:
> I guess we still want this

IMO no, this really belongs in the main upstream policy.

Conan-Kudo:
I've already talked to the fedora-selinux people. We will need it here regardless because when we change things that require new knobs, we need a way to backport it.

Conan-Kudo:
I've been asked to rewrite this to use its own labels and policy tunables, which is why this is taking so long...

Conan-Kudo:
@schaefi I've revised this based on what I've seen for osbuild and other image build tools. Please test and merge if it works for you in SELinux enforcing mode.

schaefi:
Thanks, I created a new Staging build and will test on a selinux enabled system in the AWS cloud. Give me some time to get there... other work related stuff required my attention

schaefi:
@Conan-Kudo can you please double check on the package build from here

* https://build.opensuse.org/package/show/Virtualization:Appliances:Staging/python-kiwi

It does fail on several targets, e.g Rawhide and Tumbleweed

Thanks

schaefi:
I think problem is that `selinux` is not in the included source files in pyproject.toml
I suggest


```
include = [
   { path = "selinux", format = "sdist" },  
```

Conan-Kudo:
Whoops, you're right! Fixed.

schaefi:
This worked for me, tested on a Fedora cloud instance in AWS with selinux enabled.

Thanks

schaefi (approved):
LGTM

PR #2759, opened 2025-03-29
schema: Allow C as a valid locale

Conan-Kudo:
It should be permitted to set the "C.UTF-8" locale for minimal images that are not preloaded with locales. The "C.UTF-8" locale has been supported in Linux distributions for many years.


schaefi (approved):
LGTM

PR #2762, opened 2025-04-01
Add support for Alpine

schaefi:
Add apk repository and package manager support and provide an integration test build for the Alpine distribution

### DIFF
--- a/build-tests.sh
+++ b/build-tests.sh
@@ -20,6 +20,7 @@ ARGUMENT_LIST=(
     "test-dir:"
     "test-name:"
     "box-name:"
+    "vm"
 )
 
 function usage() {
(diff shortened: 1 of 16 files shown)


schaefi:
There is no support for apk in OBS at the moment. Thus the integration test is not provided there. To build the integration test checkout the kiwi git and call

```
pip install --upgrade kiwi-boxed-plugin
./build-tests.sh --test-dir build-tests/x86/alpine --box-name staging
```

This will fetch the staging box which contains a kiwi including this PR and runs the build in the kiwi box plugin.

Alternatively if your host system has apk-tools installed also a local build as follows is possible from the kiwi git checkout

```
poetry install --all-extras
poetry shell
sudo kiwi-ng system build --description build-tests/x86/alpine/test-image-disk --target-dir /tmp/my-alpine
```

schaefi:
This codacy stuff, we should turn it off. It was helpful but nowadays it is just a pita. I can tell the interface what I want and it says "reanalyzing" ... ending in the same complain as before even when I said, no this is not something we complain about. If it will not start to function properly in the next hour I will switch it off forever

Conan-Kudo (approved):

PR #2765, opened 2025-04-02
package: Add kiwi-image:oci Provides to -systemdeps-containers

Conan-Kudo:
This allows the Open Build Service to correctly resolve dependencies when building OCI images.


schaefi (approved):
LGTM

PR #2766, opened 2025-04-03
Restrict keyfile permissions

schaefi:
For reencrypt in combination with rd.kiwi.oem.luks.reencrypt_randompass make sure that the temporary random pass keyfile has 0400 root owned access permissions set


jsrain:
LGTM

Conan-Kudo (approved):

PR #2770, opened 2025-04-08
reinstall bootstrap packs in image phase for apt

### LINKED ISSUE #2768: Missing package metadata in built Ubuntu-based image
Hi,

I’m building a disk image based on Ubuntu 24.04 LTS using KIWI NG version 10.2.16. The generated image boots and appears functional, but the package management system (APT/dpkg) is in an inconsistent state.

Specifically, many binaries are present on the system, but dpkg does not list their corresponding packages as installed. For example:
/usr/bin/bash exists, but dpkg -l does not show the bash package
/usr/bin/find exists, but dpkg -l does not list findutils
There are many other similar cases.

It seems that these binaries are part of packages that KIWI installs during the bootstrap phase, but since they are not declared in the `<packages type="image">` section, they do not appear in the resulting dpkg database of the image.

Please let me know if you'd like me to share an example config.xml to help reproduce the issue.

Thanks!

schaefi:
Due to the special bootstrap process, the packages unpacked during bootstrap are not properly listed in the apt index. Therefore the bootstrap packages are added to the install phase which causes an install of this packages again to fix the apt index and provide a consistent system from an apt perspective. This Fixes #2768

### DIFF
--- a/kiwi/package_manager/apt.py
+++ b/kiwi/package_manager/apt.py
@@ -72,6 +72,7 @@ class PackageManagerApt(PackageManagerBase):
         self.command_env = runtime_config['command_env']
         self.distribution = runtime_config['distribution']
         self.distribution_path = runtime_config['distribution_path']
+        self.bootstrap_packages: List[str] = []
 
     def request_package(self, name: str) -> None:
         """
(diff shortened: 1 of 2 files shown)


schaefi:
Setting don't merge label because I want to zip a Staging build to check if all integration tests like this change

Conan-Kudo:
When it passes Staging tests, mark this PR ready for review and I'll look at it again.

schaefi:
All Debian based distro tests passed 

```
Virtualization:Appliances:Images:Testing_x86:ubuntu
images x86_64 (published)
.  test-image-docker
.  test-image-live-disk:Disk
.  test-image-live-disk:Live
.  test-image-live-disk:Virtual
0

Virtualization:Appliances:Images:Testing_arm:ubuntu
images aarch64 (published)
.  test-image-rpi
0

Virtualization:Appliances:Images:Testing_x86:debian
images x86_64 (published)
.  test-image-live-disk:Disk
.  test-image-live-disk:Live
.  test-image-live-disk:Virtual
0
```

schaefi:
I did a quick check on one of the integration tests 

```
root@localhost:~# dpkg -S /usr/bin/bash
bash: /usr/bin/bash
root@localhost:~# dpkg -S /usr/bin/find
findutils: /usr/bin/find
```

So this looks more consistent now

Conan-Kudo (approved):

PR #2771, opened 2025-04-09
Added <oem-ramdisk-size> element

schaefi:
So far it was only possible to specify the size of the ramdisk via the kernel commandline option: ramdisk_size. In a remote deployment it was therefore required to carry this size as a mandatory information to the deployment server. With this commit we allow to specify the size for the ramdisk to be configured as part of the image configuration which makes this information also available inside of the initrd. If provided the ramdisk_size kernel commandline option still takes precedence over the <oem-ramdisk-size> setting to avoid any behavior change and to still allow dynamic overrides of the ramdisk size.

### DIFF
--- a/doc/source/working_with_images/disk_ramdisk_deployment.rst
+++ b/doc/source/working_with_images/disk_ramdisk_deployment.rst
@@ -20,7 +20,7 @@ oem type definition:
 
 .. code:: xml
 
-    <type image="oem" filesystem="ext4" installiso="true" initrd_system="dracut" installboot="install" kernelcmdline="rd.kiwi.ramdisk ramdisk_size=2048000">
+    <type image="oem" filesystem="ext4" installiso="true" initrd_system="dracut" installboot="install" kernelcmdline="rd.kiwi.ramdisk">
         <bootloader name="grub2" timeout="1"/>
         <oemconfig>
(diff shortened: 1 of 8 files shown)


doc/source/working_with_images/disk_ramdisk_deployment.rst
```diff
@@ -39,12 +40,12 @@ without asking any questions. In a ramdisk deployment the
 optional target verification, swap space and multipath targets
 are out of scope and therefore disabled.
 
-The configured size of the ramdisk specifies the size of the
-OS disk and must be at least of the size of the System Image.
-The disk size can be configured with the following value in
-the kernelcmdline attribute:
+The configured size of the ramdisk via `oem-ramdisk-size` specifies
+the a kB size of the OS disk and must be at least of the size of the
```
mtomaschewski, line 44:
```suggestion
the kB size of the OS disk and must be at least of the size of the
```

mtomaschewski (commented):
Great, thanks for this improvement!

doc/source/working_with_images/disk_ramdisk_deployment.rst
schaefi: Thanks and fixed

Issue #2773, opened 2025-04-10
/etc/sysconfig/bootloader can be empty


aplanas:
When the bootloader is configured by the user, or `grub2-bls` is used, the file `/etc/sysconfig/bootloader` can be present in the final image, but empty.

A workaround is fill it in `disk.sh`. For example:

```
loader_type="grub2-bls"
rpm -q systemd-boot && loader_type="systemd-boot"
if [ -s /etc/sysconfig/bootloader ]; then
    sed -i "s/^LOADER_TYPE=.*$/LOADER_TYPE=\"$loader_type\"/g" /etc/sysconfig/bootloader
else
    echo "LOADER_TYPE=\"${loader_type}\"" > /etc/sysconfig/bootloader
fi
```

schaefi:
Sorry I can't follow you. Looking at the code and if I setup `<bootloader name="grub2" bls="true"/> it leads to
`GRUB_ENABLE_BLSCFG=true` in /etc/default/grub and the code calls `_setup_sysconfig_bootloader` leading to a log message saying `Writing sysconfig bootloader file` 

At least LOADER_TYPE and LOADER_LOCATION gets added or updated

So how can the file be empty ?

The provided workaround should not be needed at all. Can you be more specific or point us to an image build to reproduce it ?

Vogtinator:
> Looking at the code and if I setup `<bootloader name="grub2" bls="true"/>`

What about `<bootloader name="custom"/>`?

aplanas:
> Sorry I can't follow you. Looking at the code and if I setup `<bootloader name="grub2" bls="true"/>` it leads to `GRUB_ENABLE_BLSCFG=true` in /etc/default/grub and the code calls `_setup_sysconfig_bootloader` leading to a log message saying `Writing sysconfig bootloader file`
> 
> At least LOADER_TYPE and LOADER_LOCATION gets added or updated

Uhm I do not think that LOADER_TYPE nor LOADER_LOCATION are populated in this case ...

schaefi:
>Uhm I do not think that LOADER_TYPE nor LOADER_LOCATION are populated in this case ...

I did a local TW based build with the following type setup 

```xml
<type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="efi" installiso="true" bootpartition="false" btrfs_root_is_snapshot="true" installboot="install">
    <bootloader name="grub2" bls="true" console="serial" timeout="10"/>
    <oemconfig>
[...]

PR #2774, opened 2025-04-11
Update LOADER_TYPE setup for grub

schaefi:
If the bootloader attribute: bls is set to true, make sure the LOADER_TYPE changes to grub-bls. This is related to Issue #2773

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -624,8 +624,10 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
 
     def _setup_sysconfig_bootloader(self):
         """
-        Create or update etc/sysconfig/bootloader by the following
-        parameters required according to the grub2 bootloader setup
+        Create or update etc/sysconfig/bootloader for the following
+        parameters. Please note this file is consumed by perl-bootloader
+        only and marked for deletion with the next major release of
+        kiwi because not a mainline grub config file.
 
         * LOADER_TYPE
         * LOADER_LOCATION
@@ -640,6 +642,8 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
             'LOADER_LOCATION':
                 'none' if self.firmware.efi_mode() else 'mbr'
         }
+        if self.bls:
+            sysconfig_bootloader_entries['LOADER_TYPE'] = 'grub2-bls'
         if '--set-trusted-boot' in self.config_options:
             sysconfig_bootloader_entries['TRUSTED_BOOT'] = 'yes'
         if self.firmware.efi_mode() == 'uefi':
--- a/test/unit/bootloader/config/grub2_test.py
+++ b/test/unit/bootloader/config/grub2_test.py
@@ -874,6 +874,7 @@ class TestBootLoaderConfigGrub2:
         sysconfig_bootloader = MagicMock()
         mock_sysconfig.return_value = sysconfig_bootloader
         mock_exists.return_value = True
+        self.bootloader.bls = False
         self.bootloader._setup_sysconfig_bootloader()
         mock_sysconfig.assert_called_once_with(
             'root_dir/etc/sysconfig/bootloader'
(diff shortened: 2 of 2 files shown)


schaefi:
@Conan-Kudo I know things are different on Fedora and RHEL regarding BLS and I believe it is just working there. At least that's what my integration tests tell me. The same applies to Ubuntu and the implementation in kiwi for grub just makes use of the available grub tooling. There is a conversation about the setting of the LOADER_TYPE in the referenced Issue #2773. I would be very much interested how this is handled in Fedora and RHEL and if we maybe need to adapt more kiwi code.

The change here is tightly coupled to the `bls` attribute. Thus users have a chance to control it. That's why I think the change here is safe and useful but to be honest all about BLS is so confusing and I thought with the spec things should be easy and not so messy

Alternatively it would also be possible to implement a complete new bootloader class only for BLS. But I'm only open for this if it can be done in a generic way for all OS'es

Thanks

aplanas:
> If the bootloader attribute: bls is set to true, make sure the LOADER_TYPE changes to grub-bls

Should be `LOADER_TYPE="grub2-bls"`, as is recognized by `update-bootloader`

(Moving from https://github.com/OSInside/kiwi/issues/2773#issuecomment-2796305503)

> Look, many other OS'es uses BLS as the default and we don't set the LOADER_TYPE to grub2-bls on any of them because it's a setting for the system boot architecture and it's working as such on all of them.

Are they using `update-bootloader`? This file in `/etc/sysconfig/bootloader` does now have an owner, and it is generated by YaST, but I thought that main consumer as always `pbl`.  Is this a cross-distro file?

> From a plain boot technology it makes a huge difference if you need boot code at some offset that you jump to (BIOS) or if you boot through a firmware and its layout requirements (EFI). Therefore the loader type setting to grub2 or grub2-efi made sense to me.

Ah, but this is not only about the UEFI/BIOS separation, is about the kind of installed bootloader. Another possible value for `LOADER_TYPE` is `systemd-boot`

>  Well if the system is EFI it stays EFI and a lot of actions must be the same because EFI is EFI

No, they are totally different. The ESP (an EFI artifact) is managed very different in the case of grub2-efi or systemd-boot.  In this case, grub2-efi is almost identical to systemd-boot.

> Don't get me wrong but everyone is cooking its own soup here and all this is making our lives unbelievable complicated. I'm very concerned that I won't be able to map what SUSE is doing with grub BLS into kiwi in a generic way.

I understand. I would be OK if this is not implemented at all.  At least not until there is an official BLS support from GRUB2.

With grub2-bls we decided to move to a different path, more close to systemd-boot, with the hope to unify all the BLS under bootctl and the systemd tools.

In this case `GRUB_ENABLE_BLSCFG` is just wrong.  For example, with grub2-bls there is no place to put this variable (there is no grub.conf not in ESP nor in /etc).  grub2-bls is just a file that you copy into the ESP, and all the theme is embedded there (grubenv is used as the load.conf from systemd-boot)

Conan-Kudo:
We need to put the `/etc/sysconfig/bootloader` logic on the chopping block for kiwi 11. There's no reason that this shouldn't just be configured in `config.sh` like every other distribution's weird config files.

schaefi:
>Should be LOADER_TYPE="grub2-bls", as is recognized by update-bootloader

typo in my commit text, the code was doing it right :) fixed

Conan-Kudo:
This commit needs to make it clear that it's about the SUSE-only `/etc/sysconfig/bootloader` file used by `perl-Bootloader`.

Conan-Kudo:
`/etc/default/grub` is mainline GRUB configuration file. `/etc/sysconfig/bootloader` is for perl-bootloader. We don't have anything for `grubby` in kiwi either.

Conan-Kudo:
If we're okay with dropping it now in kiwi 10, we _could_, but all the SUSE kiwi descriptions need to be updated for such an incompatibility...

schaefi:
I created #2783 to drop writing /etc/sysconfig/bootloader beginning with kiwi-11

* I don't want to add incompatible changes in the v10 code base.
* I added a comment to the method in grub that it will be deleted with the next major version of kiwi because not a mainline grub config file

schaefi:
@Conan-Kudo can you redo the review please ? Thanks

Conan-Kudo (approved):

Issue #2777, opened 2025-04-14
Too much stuff installed into the ISO ESP, causing weird issues and maxing out the ESP


Conan-Kudo:
## Problem description

Currently, kiwi semi-blindly syncs all the ESP content for ISO builds into the ISO ESP. This has a couple of unfortunate consequences:

* The ISO ESP has a lot of stuff that is never used, which wastes space and [forces the ESP size to be maxed out](https://gitlab.com/CentOS/Hyperscale/releng/kiwi-descriptions/-/commit/ab378d7e8915a5401384c00e84d014f99720387f).
* In some circumstances, the extra stuff gets activated and causes weird results (like spurious extra boot entries, like in [RH#2358785](https://bugzilla.redhat.com/2358785))

Compare what Lorax (a Fedora image creation tool) ships for an ISO ESP to what kiwi ships:

Lorax:
```
Fedora-SB-ostree-x86_64-42/EFI$ tree
.
└── BOOT
    ├── BOOT.conf
    ├── BOOTIA32.EFI
    ├── BOOTX64.EFI
    ├── fonts
    │   └── unicode.pf2
    ├── grub.cfg
    ├── grubia32.efi
    ├── grubx64.efi
    ├── mmia32.efi
    └── mmx64.efi

3 directories, 9 files
```

kiwi:
```
Fedora-WS-Live-42/EFI$ tree
.
├── BOOT
│   ├── BOOTIA32.EFI
│   ├── bootx64.efi
│   ├── BOOTX64.EFI
│   ├── fbia32.efi
│   ├── fbx64.efi
│   ├── grub.cfg
│   ├── grubx64.efi
│   └── mmx64.efi
└── fedora
    ├── BOOTIA32.CSV
    ├── BOOTX64.CSV
    ├── gcdia32.efi
    ├── gcdx64.efi
    ├── grub.cfg
    ├── grubia32.efi
    ├── grubx64.efi
    ├── mmia32.efi
    ├── mmx64.efi
    ├── shim.efi
    ├── shimia32.efi
    └── shimx64.efi

3 directories, 20 files
```

There's clearly too much stuff here and we need a way to reduce this without going full "custom".

## Proposed solution

A mechanism similar to `exclude-files.yaml` used for images needs to be made available for filtering out ESP content. 

## OS and Software information

* KIWI version: 10.2.16
* Operating system host version: Fedora Linux 42
* Operating system target version: Fedora Linux 42
[...]

Issue #2008, opened 2021-12-28
Fedora UEFI Live images (re)create boot entry


ignatenkobrain:
## Problem description

Building Fedora image with following image type/bootloader, every time I boot it on a real system, it (re)creates a boot entry in EFI leading to very useless entries in the UEFI.

```xml
<type image="iso" primary="true" flags="overlay" firmware="uefi" hybridpersistent="false"/>
```

```
❯ sudo efibootmgr -v | grep Fedora
Boot0000* Fedora	HD(1,GPT,4d16e6d9-001c-460f-be90-59cdb1ef74fc,0x800,0x12c000)/File(\EFI\fedora\shimx64.efi)
Boot0002* Fedora	PciRoot(0x0)/Pci(0x14,0x0)/USB(13,0)/CDROM(1,0x207fc8,0x28000)/File(\EFI\fedora\shimx64.efi)
Boot0003* Fedora	PciRoot(0x0)/Pci(0x14,0x0)/USB(13,0)/CDROM(1,0x267210,0x28000)/File(\EFI\fedora\shimx64.efi)
```

I believe shim invokes fallback as described here https://github.com/rhboot/shim/blob/main/README.fallback.

If I understand the documentation correctly, Live systems are not supposed to have /boot/fedora at all and use /boot/EFI only but that's not what kiwi is producing.

For comparison, official Fedora Live image has:

```
media/brain/Fedora-WS-Live-35-1-2 
❯ find EFI -type f
EFI/BOOT/BOOT.conf
EFI/BOOT/BOOTIA32.EFI
EFI/BOOT/BOOTX64.EFI
EFI/BOOT/fonts/unicode.pf2
EFI/BOOT/grub.cfg
EFI/BOOT/grubia32.efi
EFI/BOOT/grubx64.efi
EFI/BOOT/mmia32.efi
EFI/BOOT/mmx64.efi
```

while what kiwi is producing:

```
[root@powerful-box tmp]# find EFI -type f
EFI/BOOT/BOOTX64.EFI
EFI/BOOT/bootx64.efi
EFI/BOOT/fbx64.efi
EFI/BOOT/grub.cfg
EFI/BOOT/grubx64.efi
EFI/BOOT/mmx64.efi
EFI/fedora/BOOTX64.CSV
EFI/fedora/grub.cfg
EFI/fedora/grubx64.efi
EFI/fedora/mmx64.efi
EFI/fedora/shim.efi
EFI/fedora/shimx64.efi
```
```

[...]

PR #2778, opened 2025-04-15
Add support for filtering out files from the ESP image for GRUB

Conan-Kudo:
Prior to this change, KIWI blindly synced the ESP directory into the embedded ESP image. Depending on the distribution and packages included for the created image, this can have undesirable side-effects.

For image builds that need some more fine-grained control over the creation of the embedded ESP image (particularly for ISO images), this change introduces the ability to inject an exclusion list similar to what is used to filter out files for the root filesystem.

Fixes: https://github.com/OSInside/kiwi/issues/2008
Fixes: https://github.com/OSInside/kiwi/issues/2777

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -443,19 +443,30 @@ class Defaults:
             return yaml.safe_load(meta)
 
     @staticmethod
-    def get_exclude_list_from_custom_exclude_files(root_dir: str) -> List:
+    def _parse_exclude_file(root_dir: str, exclude_filename: str) -> List:
         """
-        Provides the list of folders that are excluded by the
-        optional metadata file image/exclude_files.yaml
+        Retrieves an exclusion list from the provided metadata file
 
-        :return: list of file and directory names
+        The file should contain a YAML dictionary with a top-level key
+        named 'exclude' and the list of exclusions as its value.
+
+        The list of exclusions may include:
+            * file paths
+            * folder paths
+            * glob patterns
+
+        Paths and patterns should be relative to the filesystem or
+        directory that they're being excluded from.
+
+        :return: list of paths and glob patterns
 
         :param string root_dir: image root directory
+        :param string exclude_filename: file exclusion YAML metadata file
 
         :rtype: list
         """
         exclude_file = os.sep.join(
-            [root_dir, 'image', 'exclude_files.yaml']
+            [root_dir, 'image', exclude_filename]
         )
         exclude_list = []
         if os.path.isfile(exclude_file):
@@ -472,6 +483,36 @@ class Defaults:
                         f'invalid yaml structure in {exclude_file}, ignored'
                     )
         return exclude_list
+
+    @staticmethod
+    def get_exclude_list_from_custom_exclude_files(root_dir: str) -> List:
+        """
+        Gets the list of excluded items for the root filesystem from
+        the optional metadata file image/exclude_files.yaml
+
+        :return: list of paths and glob patterns
+
+        :param string root_dir: image root directory
+
+        :rtype: list
+        """
+        return Defaults._parse_exclude_file(root_dir, 'exclude_files.yaml')
+
+    @staticmethod
+    def get_exclude_list_from_custom_exclude_files_for_efifatimage(root_dir: str) -> List:
+        """
+        Gets the list of excluded items for the ESP's EFI folder from
+        the optional metadata file image/exclude_files_efifatimage.yaml
+
+        Excluded items must be relative to the ESP's /EFI directory.
+
+        :return: list of paths and glob patterns
+
+        :param string root_dir: EFI root directory
+
+        :rtype: list
+        """
+        return Defaults._parse_exclude_file(root_dir, 'exclude_files_efifatimage.yaml')
 
     @staticmethod
     def get_exclude_list_for_non_physical_devices():
--- a/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
+++ b/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
@@ -61,6 +61,8 @@
         <source path="obsrepositories:/"/>
     </repository>
     <packages type="image">
+        <!-- Filter out unwanted EFI files from the embedded ESP -->
+        <file name="iso-esp-excludes.yaml" target="image/exclude_files_efifatimage.yaml"/>
         <package name="grub2"/>
         <package name="grubby"/>
         <package name="kernel"/>
--- a/build-tests/x86/fedora/test-image-live-disk/iso-esp-excludes.yaml
+++ b/build-tests/x86/fedora/test-image-live-disk/iso-esp-excludes.yaml
@@ -1,0 +1,3 @@
+exclude:
+- BOOT/fb*.efi
+- fedora
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -35,6 +35,7 @@ from kiwi.path import Path
 from kiwi.system.identifier import SystemIdentifier
 from kiwi.utils.sync import DataSync
 from kiwi.utils.sysconfig import SysConfig
+from kiwi.utils.temporary import Temporary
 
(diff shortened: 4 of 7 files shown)


Conan-Kudo:
This is still in draft because this needs an integration test, but I think it should work now.

AdamWill:
does this allow for wildcards? it'd be nice to be able to exclude `fb*.efi` or something like that.

Conan-Kudo:
It is passed into `rsync` so whatever rsync supports?

Conan-Kudo:
Hmm, looks like it's not hooked up right?

```
[ DEBUG   ]: 13:02:41 | Looking for mcopy in /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/var/lib/snapd/snap/bin
[ DEBUG   ]: 13:02:41 | EXEC: [mcopy -Do -s -i /home/ngompa/Sources/pagure.io/fedora-kiwi-descriptions/outdir-build/efi-loader.ffrlgtci /var/tmp/efi_folderbkwzwc9q ::EFI]
```

It doesn't seem to have the impact I expected.

Conan-Kudo:
Okay some minor tweaks and here we go, it works!

@AdamWill yes globs work, tested with the following snippet:

```yaml
exclude:
- BOOT/fb*.efi
- fedora
```

Conan-Kudo:
Here's the fedora-kiwi-descriptions pull request that leverages this new code: https://pagure.io/fedora-kiwi-descriptions/pull-request/177

Conan-Kudo:
I've updated our integration test to include testing this logic.

supakeen:
I've tested this PR by taking Fedora's `fedora-kiwi-descriptions` and building an ISO without this patch applied, and then one with in combination with the following patch(es) to the definitions:

```
diff --git a/components/liveinstall.xml b/components/liveinstall.xml
index ad15aa7..a1b3e0a 100644
--- a/components/liveinstall.xml
+++ b/components/liveinstall.xml
@@ -47,6 +47,7 @@
        </preferences>
        <packages type="iso" patternType="plusRecommended" profiles="LiveInstall">
                <!-- The point of a live image is to install -->
+               <file name="iso-esp-excludes.yaml" target="image/exclude_files_efifatimage.yaml"/>
                <namedCollection name="anaconda-tools"/>
                <package name="anaconda"/>
                <package name="anaconda-install-env-deps"/>
```

And this exclude file:

```
exclude:
  - BOOT/fb*.efi
  - fedora
```

Before the patch the ISO ESP looks like this:

```
EFI/
├── BOOT
│   ├── BOOTIA32.EFI
│   ├── BOOTX64.EFI
│   ├── fbia32.efi
│   ├── fbx64.efi
│   ├── grub.cfg
│   ├── grubx64.efi
│   └── mmia32.efi
└── fedora
    ├── BOOTIA32.CSV
    ├── BOOTX64.CSV
    ├── gcdia32.efi
    ├── gcdx64.efi
    ├── grub.cfg
    ├── grubia32.efi
    ├── grubx64.efi
    ├── mmia32.efi
    ├── mmx64.efi
    ├── shim.efi
    ├── shimia32.efi
    └── shimx64.efi
```

and after:

```
EFI/
└── BOOT
    ├── BOOTIA32.EFI
    ├── BOOTX64.EFI
    ├── grub.cfg
    ├── grubx64.efi
    └── mmia32.efi
```

Conan-Kudo:
@supakeen Were you able to verify this fixes the issue in boot testing?

supakeen:
> @supakeen Were you able to verify this fixes the issue in boot testing?

Not yet, only statically that the offending files are gone. I'll try a boot test after dinner.

AdamWill:
OK, this does look like it fixes the Fedora issue to me in local testing.

kiwi/defaults.py
```diff
@@ -443,19 +443,20 @@ def get_runtime_checker_metadata() -> Dict:
             return yaml.safe_load(meta)
 
     @staticmethod
-    def get_exclude_list_from_custom_exclude_files(root_dir: str) -> List:
+    def _parse_exclude_file(root_dir: str, exclude_filename: str) -> List:
         """
         Provides the list of folders that are excluded by the
-        optional metadata file image/exclude_files.yaml
+        optional metadata file
```
iammattcoleman, line 449: This method's description should focus on the exclude list format. Also, the file is not optional within the context of this method.

```suggestion
        Retrieves an exclusion list from the provided metadata file

        The file should contain a YAML dictionary with a top-level key
        named 'exclude' and the list of exclusions as its value.

        The list of exclusions may include:
            * file paths
            * folder paths
            * glob patterns

        Paths and patterns should be relative to the root of the
        filesystem that they're being excluded from.
```

kiwi/defaults.py
```diff
@@ -443,19 +443,20 @@ def get_runtime_checker_metadata() -> Dict:
...
 
         :return: list of file and directory names
```
iammattcoleman, line 451:
```suggestion
        :return: list of paths and glob patterns
```

kiwi/defaults.py
```diff
@@ -443,19 +443,20 @@ def get_runtime_checker_metadata() -> Dict:
...
 
         :param string root_dir: image root directory
+        :param string exclude_filename: file exclusion yaml metadata file
```
iammattcoleman, line 454:
```suggestion
        :param string exclude_filename: file exclusion YAML metadata file
```

kiwi/defaults.py
```diff
@@ -473,6 +474,35 @@ def get_exclude_list_from_custom_exclude_files(root_dir: str) -> List:
                     )
         return exclude_list
 
+    @staticmethod
+    def get_exclude_list_from_custom_exclude_files(root_dir: str) -> List:
+        """
+        Provides the list of folders that are excluded by the
+        optional metadata file image/exclude_files.yaml
```
iammattcoleman, line 481:
```suggestion
        Gets the list of excluded items for the root filesystem from
        the optional metadata file image/exclude_files.yaml
```

kiwi/defaults.py
```diff
@@ -473,6 +474,35 @@ def get_exclude_list_from_custom_exclude_files(root_dir: str) -> List:
...
+
+        :return: list of file and directory names
```
iammattcoleman, line 483:
```suggestion
        :return: list of paths and glob patterns
```

kiwi/defaults.py
```diff
@@ -473,6 +474,35 @@ def get_exclude_list_from_custom_exclude_files(root_dir: str) -> List:
...
+        :rtype: list
+        """
+        return Defaults._parse_exclude_file(root_dir, 'exclude_files.yaml')
+
+    @staticmethod
+    def get_exclude_list_from_custom_exclude_files_for_efifatimage(root_dir: str) -> List:
+        """
+        Provides the list of folders that are excluded by the
+        optional metadata file image/exclude_files_efifatimage.yaml
+        Note that file list must be relative to /EFI directory in ESP.
```
iammattcoleman, line 496:
```suggestion
        Gets the list of excluded items for the ESP's EFI folder from
        the optional metadata file image/exclude_files_efifatimage.yaml

        Excluded items must be relative to the ESP's /EFI directory.
```

kiwi/defaults.py
```diff
@@ -473,6 +474,35 @@ def get_exclude_list_from_custom_exclude_files(root_dir: str) -> List:
...
+
+        :return: list of file and directory names
```
iammattcoleman, line 498:
```suggestion
        :return: list of paths and glob patterns
```

kiwi/defaults.py
```diff
@@ -473,6 +474,35 @@ def get_exclude_list_from_custom_exclude_files(root_dir: str) -> List:
...
+
+        :param string root_dir: efi root directory
```
iammattcoleman, line 500:
```suggestion
        :param string root_dir: EFI root directory
```

iammattcoleman (changes requested):
I've added a few suggestions to improve the documentation for these methods.

iammattcoleman (approved):

PR #2783, opened 2025-04-24
Drop writing /etc/sysconfig/bootloader

schaefi:
There are several mainline grub configurations which kiwi cares for but /etc/sysconfig/bootloader is for perl-bootloader only and pretty distribution specific which makes it hard to maintain the correct content. We don't have anything for other distro specific grub tools like grubby in kiwi either.

### DIFF
--- a/kiwi/bootloader/config/grub2.py
+++ b/kiwi/bootloader/config/grub2.py
@@ -240,7 +240,6 @@ class BootLoaderConfigGrub2(BootLoaderConfigBase):
         )
 
         self._setup_default_grub()
-        self._setup_sysconfig_bootloader()
         if self.arch.startswith('s390'):
             self._setup_zipl2grub_conf()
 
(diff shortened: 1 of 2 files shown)


Conan-Kudo (approved):

Conan-Kudo:
This will need rebasing.

schaefi:
> This will need rebasing.

done

Vogtinator:
This will break *all* openSUSE and SLES images, possibly even silently (only breaks after installing updates). This really needs an announcement or should be opt-in for migration.

schaefi:
>This will break all openSUSE and SLES images,

kiwi11 will not be part of SLES or Leap, kiwi for SLES and Leap in SUSE has its own repo
I plan a submission to TW and also only if no breakage occur. I will collect the issues for openSUSE TW
and send a topic list on the kiwi slack channel for maybe needed adaptions on image builds
for TW.

/etc/sysconfig/bootloader is consumed by perl-bootloader and not a mainline grub config file. The changes still in flux with regards to bls on suse would cause further trouble and we have announced the drop in code 
```
    Please note this file is consumed by perl-bootloader
    only and marked for deletion with the next major release of  
    kiwi because not a mainline grub config file.
```

on the approved PR and labeled as kiwi-11

and in the community meeting

As I said I will collect issues from the integration test builds from Virt:App:Builder first and no submission into TW takes place prior clarification on open issue topics.

Vogtinator:
> I plan a submission to TW and also only if no breakage occur.

The issue is, it may be completely silent: The image boots and appears to work properly, but updates of the bootloader are never actually installed because perl-Bootloader thinks that none is selected.

Conan-Kudo:
Then someone should go and add the `/etc/sysconfig/bootloader` stuff to the `config.sh` or the overlay tarball. It's been slated for removal for over a year. There's no real excuse to not prepare for it.

Vogtinator:
Sure, but this needs to be announced and coordinated properly, not just as a breaking change but a silently breaking one that likely requires action.

Conan-Kudo:
It was, though. @schaefi told everyone over a year ago, because _I_ asked to defer it for kiwi-11.

Vogtinator:
> It was, though.

In that case that's fine. I don't remember seeing that announcement though, but possibly it was just too long ago.

Conan-Kudo:
Yeah, I asked him to tell folks when we discussed this during the community meeting after this PR was made. Unfortunately, it's not so easy to know whether everyone knows because openSUSE kiwi descriptions are not centralized like Fedora ones are.

Vogtinator:
> Unfortunately, it's not so easy to know whether everyone knows

I'm pretty sure the answer to that is unfortunately "no": There are an unknown number of users of kiwi that build custom openSUSE images that may or may not be subscribed to mailing lists or GitHub releases.

The only way to really make everyone aware of it is to cause intentional build fails (like forced schemaversion upgrades), which are annoying on their own...

PR #2785, opened 2025-04-25
Set security context after root sync

schaefi:
On selinux enabled image builds we call setfiles initially after the root tree is complete and after each script invocation that might change the system. However the security context also applies to mount points e.g volumes which only exists at the time when the root tree gets synced to the actual image binary. Thus this commit also calls setfiles on the mounted root tree after data sync.

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -848,10 +848,8 @@ class DiskBuilder:
                 integrity_root
             )
 
-            # run post sync script hook
-            if system and self.system_setup.script_exists(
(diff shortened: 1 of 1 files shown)


schaefi:
@Conan-Kudo I believe this is referencing the bug you reported in the Fedora bugzilla. Maybe you can add the bug url here for reference

schaefi:
> You can add `Fixes: rh#2333743` to the commit message (similar to how you do for `bsc#` and `boo#` bug IDs).

done

Conan-Kudo (approved):

PR #2793, opened 2025-04-30
Fixed default bls value setup

schaefi:
Fixed get_build_type_bootloader_bls behavior in case the bls attribute is not set. In this case get_bls() returns a None value which was returned. However in this case the attribute value should not be taken into account and the method defined default value for bls should be returned. This Fixes #2542

### DIFF
--- a/package/kiwi-revert-bls-default-for-suse.patch
+++ b/package/kiwi-revert-bls-default-for-suse.patch
@@ -1,25 +1,15 @@
-From b0d850842b0c8d4c98f0486f66ffd7c6e8f1a685 Mon Sep 17 00:00:00 2001
-From: Neal Gompa <ngompa@opensuse.org>
-Date: Thu, 2 May 2024 06:23:45 -0400
-Subject: [PATCH] Revert "kiwi/bootloader: restore backward compatibility for
- grub2 with bls"
-
-For the time being, SUSE distributions cannot handle KIWI's default to use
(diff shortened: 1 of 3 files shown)


package/kiwi-revert-bls-default-for-suse.patch
```diff
@@ -1,67 +1,57 @@
-From b0d850842b0c8d4c98f0486f66ffd7c6e8f1a685 Mon Sep 17 00:00:00 2001
-From: Neal Gompa <ngompa@opensuse.org>
-Date: Thu, 2 May 2024 06:23:45 -0400
-Subject: [PATCH] Revert "kiwi/bootloader: restore backward compatibility for
- grub2 with bls"
-
-For the time being, SUSE distributions cannot handle KIWI's default to use
-BLS with GRUB2. Until they catch up, revert this.
-
-This reverts commit 155d969d7752d92c90c69044bdd69be70f0a1c24.
```
Conan-Kudo, line 10: Can we preserve the metadata here so that people know why this patch exists?

Conan-Kudo (changes requested):

package/kiwi-revert-bls-default-for-suse.patch
schaefi: yes done

Conan-Kudo (approved):

PR #2802, opened 2025-05-12
Add support for new tarball-based WSL format

### LINKED ISSUE #2678: Support new tarball-based WSL format
## Problem description
Microsoft announced today a brand new way for WSL images to work. Instead of creating appx packages, now regular tarballs with some extra configuration can be used to create `.wsl` archives. [Microsoft documents the full definition of this new format](https://learn.microsoft.com/en-us/windows/wsl/build-custom-distro).

We should add a new "wsl" image type that handles this new format, derived from our tarball format with some extra declarative metadata.

schaefi:
With the new image="wsl" type one can build a WSL container image that uses the new tarball format. This Fixes #2678

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-wsl-ng/root/etc/wsl-distribution.conf
+++ b/build-tests/x86/tumbleweed/test-image-wsl-ng/root/etc/wsl-distribution.conf
@@ -1,0 +1,3 @@
+[oobe]
+defaultUid = 1000
+defaultName = SUSE
--- a/build-tests/x86/tumbleweed/test-image-wsl-ng/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-wsl-ng/appliance.kiwi
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<image schemaversion="7.5" name="kiwi-test-image-wsl-ng">
+    <description type="system">
+        <author>Marcus Schäfer</author>
+        <contact>marcus.schaefer@suse.com</contact>
+        <specification>WSL test build for WSL >= 2.4.4 new tar spec</specification>
+    </description>
+    <preferences>
+        <version>2.4.4</version>
+        <packagemanager>zypper</packagemanager>
+        <rpm-excludedocs>true</rpm-excludedocs>
+        <rpm-check-signatures>false</rpm-check-signatures>
+        <locale>en_US</locale>
+        <keytable>us</keytable>
+        <type image="wsl"/>
+    </preferences>
+    <users>
+        <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root" groups="root"/>
+    </users>
+    <repository type="rpm-md">
+        <source path="obsrepositories:/"/>
+    </repository>
+    <packages type="image">
+        <package name="patterns-openSUSE-base"/>
+        <package name="systemd"/>
+        <package name="bash"/>
+        <package name="acl"/>
+        <package name="coreutils"/>
+        <package name="cpio"/>
+        <package name="findutils"/>
+        <package name="dracut"/>
+        <package name="pigz"/>
+        <package name="iputils"/>
+        <package name="tar"/>
+        <package name="psmisc"/>
+        <package name="lvm2"/>
+        <package name="fontconfig"/>
+        <package name="fonts-config"/>
+        <package name="tar"/>
+        <package name="sudo"/>
+        <package name="rsync"/>
+    </packages>
+    <packages type="bootstrap">
+        <package name="gawk"/>
+        <package name="grep"/>
+        <package name="gzip"/>
+        <package name="udev"/>
+        <package name="xz"/>
+        <package name="shadow"/>
+        <package name="filesystem"/>
+        <package name="glibc-locale"/>
+        <package name="cracklib-dict-full"/>
+        <package name="ca-certificates"/>
+        <package name="ca-certificates-mozilla"/>
+        <package name="openSUSE-release"/>
+    </packages>
+</image>
--- a/build-tests/x86/tumbleweed/test-image-wsl-ng/root/etc/wsl.conf
+++ b/build-tests/x86/tumbleweed/test-image-wsl-ng/root/etc/wsl.conf
@@ -1,0 +1,2 @@
+[boot]
+systemd=true
--- a/doc/source/concept_and_workflow/systemdeps.rst
+++ b/doc/source/concept_and_workflow/systemdeps.rst
@@ -29,7 +29,8 @@ following systemdeps packages:
     compliant container images.
 
 `kiwi-systemdeps-containers-wsl`:
-  * Supports building `appx` image types.
+  * Supports building `appx` and `wsl` image types. The `wsl` type
+    is the successor for the former `appx` type
   * Installs the distribution specific tool chain to build
     WSL compliant container images on Windows systems.
 
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -364,6 +364,14 @@ image="appx"
   container engine. The image can be loaded From a Windows System
   that has support for WSL activated.
 
+image="wsl"
+  An archive image suitable for the Windows Subsystem For Linux
+  container engine >= v2.4.4. The image represents a gzip compressed
+  tar file as described in
+  https://learn.microsoft.com/en-us/windows/wsl/build-custom-distro
+  and can be loaded From a Windows System
+  that has support for WSL in that version activated.
+
 image="kis"
   An optional root filesystem image associated with a kernel and initrd.
   The use case for this component image type is highly customizable.
--- a/doc/source/image_types_and_results.rst
+++ b/doc/source/image_types_and_results.rst
@@ -173,6 +173,14 @@ image="appx"
   - **container**:
     :file:`{exc_image_base_name}.x86_64-{exc_image_version}.appx`
 
+image="wsl"
+  An archive image suitable for the Windows Subsystem For Linux
+  container engine >= v2.4.4. The result is a gzip compressed tar
+  archive with the `.wsl` extension:
+
+  - **container**:
+    :file:`{exc_image_base_name}.x86_64-{exc_image_version}.wsl`
+
 image="kis"
   An optional root filesystem image associated with a kernel and initrd.
   All three binaries are packed in a tarball, see :ref:`kis` for further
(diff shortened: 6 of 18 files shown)


schaefi:
An integration test to build for this new type will be available here

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-wsl-ng

schaefi:
Things to clarify. According to the docs two metadata files are needed in the rootfs. `etc/wsl.conf` and `etc/wsl-distribution.conf`. The implementation from here does not create those files as I did not see this to be a job of the builder. Instead the setup routine checks for the presence of these files and fails the build if they are not there. Thoughts about this ?

tdawson:
> Things to clarify. According to the docs two metadata files are needed in the rootfs. `etc/wsl.conf` and `etc/wsl-distribution.conf`. The implementation from here does not create those files as I did not see this to be a job of the builder. Instead the setup routine checks for the presence of these files and fails the build if they are not there. Thoughts about this ?

That sounds good to me.  I don't think it should be kiwi's job to create the files.  If someone is going to use the .wsl extension, they need those files, and they should supply them on their own.  Either through a package, manually in the config.sh file, or someway else.

tdawson:
> I don't have time to actually test the WSL container. I'm asking for help on testing the integration test image once it has built. Thanks much

I should be able to test locally.  I have things setup for testing wsl images.
But I'm having a hard time finding the images to download and test.
Is this the correct area to start looking?
https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-wsl-ng

Conan-Kudo:
I would prefer optional fields to generate this data, since it's WSL-specific metadata.

Conan-Kudo:
> Things to clarify. According to the docs two metadata files are needed in the rootfs. `etc/wsl.conf` and `etc/wsl-distribution.conf`. The implementation from here does not create those files as I did not see this to be a job of the builder. Instead the setup routine checks for the presence of these files and fails the build if they are not there. Thoughts about this ?

I think this should be optionally definable in the XML. Not everyone wants to do it externally or has the ability to do so.

Conan-Kudo:
Additionally, we should probably fail the image build if the files don't exist.

schaefi:
> > I don't have time to actually test the WSL container. I'm asking for help on testing the integration test image once it has built. Thanks much
> 
> I should be able to test locally. I have things setup for testing wsl images. But I'm having a hard time finding the images to download and test. Is this the correct area to start looking? https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-wsl-ng

yes, but the repos are currently being created. Give the system a bit more time to settle. I will provide you with a download link soon...

Thanks much for helping

schaefi:
> Additionally, we should probably fail the image build if the files don't exist.

This is what this PR does

schaefi:
> > Things to clarify. According to the docs two metadata files are needed in the rootfs. `etc/wsl.conf` and `etc/wsl-distribution.conf`. The implementation from here does not create those files as I did not see this to be a job of the builder. Instead the setup routine checks for the presence of these files and fails the build if they are not there. Thoughts about this ?
> 
> I think this should be optionally definable in the XML. Not everyone wants to do it externally or has the ability to do so.

I did not quite get that. What exactly should be defined in the XML regarding wsl.conf and wsl-distribution.conf ?

schaefi:
@tdawson this should be a download link that works if you login to build.opensuse.org

* https://build.opensuse.org/build/Virtualization:Appliances:Images:Testing_x86:tumbleweed/images/x86_64/test-image-wsl-ng/kiwi-test-image-wsl-ng.x86_64-2.4.4-Build1.1.wsl

unfortunately there is no public facing download link from the publisher because obs currently has no handle for the `.wsl` extension and ignores those files from being published. I need to clarify this with the OBS team.

Let me know if you were able to download. You need an obs account... sorry for the inconvenience

Conan-Kudo:
> > > Things to clarify. According to the docs two metadata files are needed in the rootfs. `etc/wsl.conf` and `etc/wsl-distribution.conf`. The implementation from here does not create those files as I did not see this to be a job of the builder. Instead the setup routine checks for the presence of these files and fails the build if they are not there. Thoughts about this ?
> > 
> > 
> > I think this should be optionally definable in the XML. Not everyone wants to do it externally or has the ability to do so.
> 
> I did not quite get that. What exactly should be defined in the XML regarding wsl.conf and wsl-distribution.conf ?

The data that would get written into `wsl.conf` and `wsl-distribution.conf`.

schaefi:
> > I did not quite get that. What exactly should be defined in the XML regarding wsl.conf and wsl-distribution.conf ?
> 
> The data that would get written into `wsl.conf` and `wsl-distribution.conf`.

Hmm, so to bring in that data you have the following options

1. overlay files (so done by the integration test here)
2. some arbitrary archive and `<archive name="my-wsl-meta"/>
3. some package and `<package name="my-wsl-meta"/>

why do we need to add more options ? And if you suggest that we come up with a schema to describe the format of wsl.conf and wsl-distribution.conf in our own words, then I object against that plan simply because we have zero control what Microsoft does at this point and I already suffered a lot from this when we tried to add a schema for the VMware guest config... it's basically broken all the time. I simply don't trust the information from https://learn.microsoft.com/en-us/windows/wsl/build-custom-distro to be a standard that they don't change in an incompatible way whenever they want. Sorry too much experience with Microsoft Azure APIs to trust them ;)

With the appx support I added, code to modify data according to the schema from `http://schemas.microsoft.com/appx` exists. That is something different, but with this tar specification I don't see any schema that you can validate content against and as such I will not add code that writes data in a format that can be changed into anything anytime with a closed source company in the backend that usually doesn't care. It will just hurt on the long run

jeremycline:
In my opinion, we don't need more ways to write out the WSL-specific stuff.

Either folks hack together that stuff in the `config.sh`, which is how I got things up and running for Fedora's image initially (and worked fine for a quick setup), or they produce a package to install stuff. Unless someone comes forward with a clear reason they have to have it in XML and cannot achieve the same thing with the config script or a package, I don't think we should build additional ways to do it (which increases ways to test it, is confusing to users, etc).

Conan-Kudo:
We don't need to do anything for now. I'm perfectly fine merging this as-is and figuring that out later if need be.

build-tests/x86/tumbleweed/test-image-wsl-ng/root/etc/wsl-distribution.conf
```diff
@@ -0,0 +1,3 @@
+[oobe]
+defaultUid = 1000
+defaultName = SUSE
```
jeremycline, line 3: I believe you'll need to specify a [setup script to run on first use](https://src.fedoraproject.org/rpms/wsl-setup/blame/wsl-distribution.conf?identifier=rawhide#_13), like [this one](https://src.fedoraproject.org/rpms/wsl-setup/blob/rawhide/f/wsl-oobe.sh).

Conan-Kudo:
> In my opinion, we don't need more ways to write out the WSL-specific stuff.
> 
> Either folks hack together that stuff in the `config.sh`, which is how I got things up and running for Fedora's image initially (and worked fine for a quick setup), or they produce a package to install stuff. Unless someone comes forward with a clear reason they have to have it in XML and cannot achieve the same thing with the config script or a package, I don't think we should build additional ways to do it (which increases ways to test it, is confusing to users, etc).

Philosophically, kiwi operates the other way unless there's a good reason not to. In this case, the reason is that Microsoft has not proven that the configuration format is stable enough to depend on.

tdawson:
> @tdawson this should be a download link that works if you login to build.opensuse.org
> 
>     * https://build.opensuse.org/build/Virtualization:Appliances:Images:Testing_x86:tumbleweed/images/x86_64/test-image-wsl-ng/kiwi-test-image-wsl-ng.x86_64-2.4.4-Build1.1.wsl
> 
> 
> unfortunately there is no public facing download link from the publisher because obs currently has no handle for the `.wsl` extension and ignores those files from being published. I need to clarify this with the OBS team.
> 
> Let me know if you were able to download. You need an obs account... sorry for the inconvenience

I have been able to download and run that image.
I did the "double click" method.  (double click on the icon once it's downloaded).  It ran, and I got the SUSE entry in my start menu.
So, from me, it get's a thumbs up.
Thank you for working on this.

Conan-Kudo (approved):

schaefi:
> We don't need to do anything for now. I'm perfectly fine merging this as-is and figuring that out later if need be.

Sounds good to me. If those manifest files settles or get covered by some Microsoft validation I'm open to produce them from our schema at least to the minimum required data

schaefi:
> I have been able to download and run that image. I did the "double click" method. (double click on the icon once it's downloaded). It ran, and I got the SUSE entry in my start menu. So, from me, it get's a thumbs up. Thank you for working on this.

@tdawson Great, thanks a ton for testing much appreciated :)

schaefi:
I also informed the autobuild team at SUSE and asked for support of the new `.wsl` output binaries such that they also land on the download server

schaefi:
> I also informed the autobuild team at SUSE and asked for support of the new `.wsl` output binaries such that they also land on the download server

The OBS team has solved this. The new .wsl binaries are now also published properly

* https://download.opensuse.org/repositories/Virtualization:/Appliances:/Images:/Testing_x86:/tumbleweed/images/kiwi-test-image-wsl-ng.x86_64.wsl

```
kiwi-test-image-wsl-ng.x86_64-2.4.4-Build3.2.cdx.json 5/14/2025, 4:32 AM 0.1 MB
kiwi-test-image-wsl-ng.x86_64-2.4.4-Build3.2.spdx.json 5/14/2025, 4:32 AM 12.0 MB
kiwi-test-image-wsl-ng.x86_64-2.4.4-Build3.2.wsl 5/14/2025, 4:32 AM 0.1 GB
kiwi-test-image-wsl-ng.x86_64-2.4.4-Build3.2.wsl.sha256 5/14/2025, 4:32 AM 115
kiwi-test-image-wsl-ng.x86_64-2.4.4-Build3.2.wsl.sha256.asc 5/14/2025, 4:32 AM 481
kiwi-test-image-wsl-ng.x86_64.json 5/14/2025, 4:32 AM 0.1 MB
kiwi-test-image-wsl-ng.x86_64.wsl 5/14/2025, 4:32 AM 0.1 GB
kiwi-test-image-wsl-ng.x86_64.wsl.sha256 5/14/2025, 4:43 AM 100
kiwi-test-image-wsl-ng.x86_64.wsl.sha256.asc 5/14/2025, 4:43 AM 481
```

PR #2812, opened 2025-05-16
Add support for <initrd> section as part of <type>

### LINKED ISSUE #2408: RFE: Add support for direct UKI boot
## Problem description
As part of the effort to support [UKIs in Fedora Cloud](https://fedoraproject.org/wiki/Changes/Unified_Kernel_Support_Phase_2), we need to be able to create images that either generate UKIs or use pre-generated UKIs to boot the system (UEFI->shim->UKI or UEFI->UKI).

I suspect the way to solve this would be to add a `uki` "bootloader" which would force UEFI only mode and do the minimal effort required to configure the image to boot. We'd want to detect if there's already a UKI built as `/boot/efi/EFI/*/vmlinuz-$(uname -r).efi` to use or `/lib/modules/$(uname -r)/*.efi` to copy and install. If one doesn't exist, we should use `dracut --uefi` to generate one.

An example of a `dracut --uefi` call (ripped from the Fedora `kernel` package that produces `kernel-virt-uki`):

```shell
dracut --conf=/path/to/dracut/uki-config.conf \
       --confdir="$(mktemp -d)" \
       --verbose \
       --kver "$(uname -r)" \
       --kmoddir "/lib/modules/$(uname -r)/" \
       --logfile="$(mktemp)" \
       --uefi \
       --kernel-image "/lib/modules/$(uname -r)/vmlinuz" \
       --kernel-cmdline 'console=tty0 console=ttyS0' \
       "/boot/efi/EFI/Linux/vmlinuz-$(uname -r).efi"
```

Sample dracut uki config:

```shell
# generic + compressed please
hostonly="no"
compress="xz"

# VMs can't update microcode anyway
early_microcode="no"

# modules: basics
dracutmodules+=" base systemd systemd-initrd dracut-systemd dbus dbus-broker usrmount shutdown "

# modules: storage support
dracutmodules+=" dm lvm rootfs-block fs-lib "

# modules: tpm and crypto
dracutmodules+=" crypt crypt-loop tpm2-tss "

# drivers: virtual buses, pci
drivers+=" virtio-pci virtio-mmio "      # qemu-kvm
drivers+=" hv-vmbus pci-hyperv "         # hyperv
drivers+=" xen-pcifront "                # xen

# drivers: storage
drivers+=" ahci nvme sd_mod sr_mod "     # generic
drivers+=" virtio-blk virtio-scsi "      # qemu-kvm
drivers+=" hv-storvsc "                  # hyperv
drivers+=" xen-blkfront "                # xen

# root encryption
drivers+=" dm_crypt "

# filesystems
filesystems+=" vfat btrfs ext4 xfs overlay "

# systemd-pcrphase
install_items+=" /lib/systemd/system/systemd-pcrphase-initrd.service /usr/lib/systemd/systemd-pcrphase /usr/lib/systemd/system/initrd.target.wants/systemd-pcrphase-initrd.service "
```

Unlike with the shim+grub situation, we will want a way to force generating UKIs because a different cmdline or module bundle may be desired, since there's no way to pass kernel cmdlines from UEFI to boot the OS.

* KIWI version: 9.25.16
* Operating system target version: Fedora Linux 40
* Koji version (N/A if not using Koji): 1.33.1

### LINKED ISSUE #2809: Add basic UKI generation support to dracut module
### Overview

This issue derives from discussion in https://github.com/OSInside/kiwi/issues/2801 and relates to https://github.com/OSInside/kiwi/issues/2408.

We need to add support in KIWI-NG to generate Unified Kernel Images (UKIs) using dracut's `--uefi` flag. This will allow users to create UKIs as part of their normal KIWI build process, rather than requiring manual steps or external tools.

In our case, we need UKIs for attestation purposes, where having a single verifiable boot image is crucial. Other users might find this useful for secure boot scenarios, as UKIs simplify the signing and verification process by combining kernel, initramfs, and cmdline into a single EFI executable.

We've experimented with manual UKI creation using dracut --uefi and are willing to implement this enhancement as a PR if there's interest.

### Current State

Currently there's no direct way to generate a basic UKI using KIWI-NG's dracut integration.

### Proposed Enhancement:

Add basic UKI generation support by enabling dracut's --uefi flag through KIWI configuration.

### Example configuration:
```
<bootloader name="systemd_boot" console="serial">
    <dracut>
        <uefi enabled="true"/>
    </dracut>
</bootloader>
```

When `<uefi enabled="true"/>` is present:

* Add `—uefi` flag to dracut command

This provides a minimal implementation focused on basic UKI generation, which can be extended later if needed.

schaefi:
Extend scope and content of the ```<initrd>``` section to be allowed as part of the ```<type>``` section. This allows to specify custom call options and modules for the dracut tool. In particular this commit implementes support for passing the uefi option to dracut to enable building an UKI EFI binary as follows:

```xml
<initrd action="setup">
    <dracut uefi="true"/>
</initrd>
```

This Fixes #2809 and Fixes #2408

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-overlayroot/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-overlayroot/appliance.kiwi
@@ -9,6 +9,7 @@
     <profiles>
         <profile name="sdboot_erofs" description="systemd boot overlay disk using erofs"/>
         <profile name="sdboot_verity_erofs" description="systemd boot verity baked overlay disk using erofs"/>
+        <profile name="sdboot_uki_verity_erofs" description="systemd UKI boot verity baked overlay disk using erofs"/>
         <profile name="grub_verity_erofs" description="grub verity baked overlay disk using erofs"/>
     </profiles>
     <preferences>
@@ -29,7 +30,7 @@
             format="vmdk"
             overlayroot="true"
             overlayroot_readonly_filesystem="erofs"
-            overlayroot_readonly_partsize="1000"
+            overlayroot_readonly_partsize="1500"
             erofscompression="zstd,level=9"
             eficsm="false"
             bootpartition="false"
@@ -51,7 +52,7 @@
             format="vmdk"
             overlayroot="true"
             overlayroot_readonly_filesystem="erofs"
-            overlayroot_readonly_partsize="1000"
+            overlayroot_readonly_partsize="1500"
             erofscompression="zstd,level=9"
             eficsm="false"
             verity_blocks="all"
@@ -65,6 +66,32 @@
             <size unit="G">4</size>
         </type>
     </preferences>
+    <preferences profiles="sdboot_uki_verity_erofs">
+        <type
+            image="oem"
+            filesystem="xfs"
+            kernelcmdline="console=ttyS0 rd.systemd.verity=1 security=selinux selinux=1 enforcing=1"
+            firmware="efi"
+            format="vmdk"
+            overlayroot="true"
+            overlayroot_readonly_filesystem="erofs"
+            overlayroot_readonly_partsize="1500"
+            erofscompression="zstd,level=9"
+            eficsm="false"
+            verity_blocks="all"
+            bootpartition="false"
+            efipartsize="200"
+        >
+            <oemconfig>
+                <oem-resize>false</oem-resize>
+            </oemconfig>
+            <bootloader name="systemd_boot" timeout="10"/>
+            <initrd action="setup">
+                <dracut uefi="true"/>
+            </initrd>
+            <size unit="G">4</size>
+        </type>
+    </preferences>
     <preferences profiles="grub_verity_erofs">
         <type
             image="oem"
@@ -74,7 +101,7 @@
             format="vmdk"
             overlayroot="true"
             overlayroot_readonly_filesystem="erofs"
-            overlayroot_readonly_partsize="1000"
+            overlayroot_readonly_partsize="1500"
             erofscompression="zstd,level=9"
             eficsm="false"
             verity_blocks="all"
@@ -92,8 +119,9 @@
     <repository type="rpm-md">
         <source path="obsrepositories:/"/>
     </repository>
-    <packages type="image" profiles="sdboot_verity_erofs,grub_verity_erofs">
+    <packages type="image" profiles="sdboot_verity_erofs,sdboot_uki_verity_erofs,grub_verity_erofs">
         <package name="cryptsetup"/>
+        <package name="systemd-boot"/>
         <package name="dracut-kiwi-verity"/>
         <package name="restorecond"/>
         <package name="policycoreutils"/>
@@ -102,12 +130,19 @@
         <package name="selinux-policy-devel"/>
         <package name="selinux-autorelabel"/>
     </packages>
-    <packages type="image">
-        <package name="patterns-base-minimal_base"/>
+    <packages type="image" profiles="sdboot_uki_verity_erofs">
+        <package name="binutils"/>
+    </packages>
+    <packages type="image" profiles="sdboot_erofs">
         <package name="systemd-boot"/>
+    </packages>
+    <packages type="image" profiles="grub_verity_erofs">
         <package name="grub2"/>
         <package name="grub2-x86_64-efi" arch="x86_64"/>
         <package name="shim"/>
+    </packages>
+    <packages type="image">
+        <package name="patterns-base-minimal_base"/>
         <package name="procps"/>
         <package name="bind-utils"/>
         <package name="systemd"/>
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -1319,6 +1319,27 @@ Used to customize the installation media images created for oem images
 deployment.
 For details see: :ref:`installmedia_customize`
 
+<preferences><type><initrd>
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Used to specify custom arguments for the initrd tooling e.g. dracut
+
+.. code:: xml
+
+   <initrd action="setup">
+       <dracut uefi="true"/>
+   </initrd>
+   <initrd action="add|omit">
+       <dracut module="some_a"/>
+       <dracut module="some_b"/>
+   </initrd>
+
+dracut.uefi.true:
+  As part of the `setup` action, enables creating a UKI EFI binary
+
+dracut.module.NAME:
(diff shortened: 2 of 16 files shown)


schaefi:
> This doesn't fully fix #2408 since for that case we need to have a bootloader mode that sets dracut up that way permanently on the final image.

Right, sorry my fault. #2408 brings up more than just building an efi binary via dracut. There is also the conversation that says we should not offer the creation of an UKI at all because it should come from the distro. So what is our take on this ?

Conan-Kudo:
I think we should offer a way to do this, it's up to the image creator how to take advantage of it.

schaefi:
> In particular this commit implementes support for passing the uefi option to dracut to enable building an UKI EFI binary as follows:

I think we are lacking clarity on UKI's and my suggestion is that I drop the code part that covers this from this PR, such that we only add the basic functionality to allow modifications to the way dracut gets called. From there we can then step by step add missing functionality. Just passing `--uefi` is by far not enough, more options are required to produce a fully working EFI binary. In addition that binary has to be placed to the right place and correctly named on the ESP. 

Managing pre-built EFI binaries has been added to kiwi for e.g shim and also for grub. It would be possible to add an UKI search  provided by the distro and triggered by a setting in the `<initrd>` section. This would then follow the approach Kraxel suggested in #2408

AmazonLinux as a requester of the feature would then require to provide a pre-build UKI via a package at a place such that kiwi can find it and we are not creating them but expecting a distro to provide them

schaefi:
> I think we should offer a way to do this, it's up to the image creator how to take advantage of it.

But this is not quite correct. If we produce an UKI initially the image will boot. On the first kernel update the system becomes inconsistent unless the UKI is recreated. If we expect the distro to provide this EFI binary, we can also expect the distro to provide a concept to recreate it on kernel updates. There is no way the image creator can take advantage of kiwi creating an UKI if the distribution is not prepared for this concept

Conan-Kudo:
Well we _could_ if we install kernel-install hook along with a dracut config setting, I think.

schaefi:
I added some rough support for the UKI stuff. It works for standard and read-only disks but together with verity there is a very late update of the initrd to include the veritytab which contains offset addresses and other data known after final syncing. kiwi updates the initrd on the ESP to achieve but with UKI support also the UKI has to be rebuilt to integrate the new initrd into the PE image. I could not find a good way to do it right now... Thus still experimenting

Tests are failing, this is all WIP

schaefi:
ok I finally managed to get this done. This PR covers

- [ ] ```<initrd>``` section usable for all oem disk types
- [ ] implementation for UKI support in a way that kiwi builds one through dracut.
- [ ] A new integration test profile to test UKI image build

During implementation I stumbled over the following questions that we can either discuss here or add later

Q: Is the distribution expected to provide a UKI image signed off by Microsoft such that secure boot becomes possible ? With this implementation the generated UKI works but is not signed. Consequently bootctl says

```
System:
      Firmware: UEFI 2.70 (EDK II 1.00)
 Firmware Arch: x64
   Secure Boot: disabled (setup)
  TPM2 Support: no
  Measured UKI: no
  Boot into FW: supported
```

Q: If the distro provides prebuilt signed UKIs, is there a standard search path across distributions ?

Q: In case of prebuilt signed UKI binaries, those also embedds the kernel commandline, as well as the initrd and the kernel itself and require to be rebuild on any kernel update. In contrast to shim where the signed code allows reading of data from some other place, how can the UKI rebuild and still be signed for secure boot ?

As I said I probably miss some details on the PE signing and I also think they are outside of the scope for this initial UKI support. But all this needs clarification imho.

Thanks

schaefi:
Find integration test build here

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-overlayroot

Conan-Kudo:
Yeah, I do not think UKIs will be popular once people realize how restrictive they are.

Conan-Kudo (approved):

Conan-Kudo:
> Q: Is the distribution expected to provide a UKI image signed off by Microsoft such that secure boot becomes possible ? With this implementation the generated UKI works but is not signed.

There is no expectation that it is signed by Microsoft. Generally, shim with grub2 or sd-boot would be used so that a chain to a distro-controlled certificate remains possible.

> Q: If the distro provides prebuilt signed UKIs, is there a standard search path across distributions ?

Fedora, to this date, is the only distribution providing distro-built UKIs, so technically yes and no? 😅 

Fedora-built UKIs are installed into `/lib/modules/$(uname -r)/vmlinuz-$flavor.efi` with addon EFI binaries in `/lib/modules/$(uname -r)/vmlinuz-$flavor.efi.extra.d/` I assume "default" flavor would actually be `vmlinuz.efi` (with no "default" flavor) since Fedora uses the "default" flavor to not have a flavor suffix. We currently only build a UKI for virtual machines with a "virt" flavor.

Fedora's `kernel-install` scriptlet will run after they are installed to copy them to `/boot/efi/EFI/Linux/` automatically. So that would probably be the place to check for pre-built UKI EFI binaries.

> Q: In case of prebuilt signed UKI binaries, those also embedds the kernel commandline, as well as the initrd and the kernel itself and require to be rebuild on any kernel update. In contrast to shim where the signed code allows reading of data from some other place, how can the UKI rebuild and still be signed for secure boot ?

shim and sd-boot/grub2 would be used and the bootloader would be used so that signing by Microsoft would not be required. Though if you are signing yourself, then you could drop these and do whatever you'd like here. I guess presumably we'd need a way to expose signing keys so that during the UKI build they would get used.

schaefi:
Ok, thanks. These details we can revisit for follow up changes with regards to UKI and secure boot

> Fedora-built UKIs are installed into /lib/modules/$(uname -r)/vmlinuz-$flavor.efi

ok, I believe none of them would work for the special image types, like overlay, verity baked, etc etc. As those require additional kernel parameters that you only happen to know when building the image. As I don't see a way to modify the kernel cmdline of a prebuilt UKI we would need to recreate them under conditions. Knowing these conditions is not straight forward and I'm unsure how to keep the secure boot capability

Since Fedora is the only distro that comes up with all this, I appreciate help for the secure boot topic regarding UKIs for future adaptions

Conan-Kudo:
I think with prebuilt UKIs all that stuff would need to be skipped. We'd probably need a way to tell kiwi which UKI flavor to pick and install, and it would follow on from that. The extent of configuration would be setting up the bootloader to chain to the UKI.

Conan-Kudo:
(And we might not need to do much beyond making sure grub2 is in BLS mode)

PR #2816, opened 2025-05-21
Drop use of setuptools

schaefi:
Since we moved to poetry and no code using setuptools anymore, this requirement can be dropped. The commit also updates the plugin documentation which was still based on setup.py


schaefi:
I'm waiting for the integration test builds to pass...

Issue #2815, opened 2025-05-20
Targeted Overlays for the root filesystem


xcawolfe-amzn:
Hello all, Caleb here from Amazon Linux

## Proposal

We'd like to propose an enhancement to KIWI's existing overlay functionality that would provide more fine-grained control over how overlays are applied to the root file system. This would support use cases where customers want to:

- Ensure that workloads on a given system cannot alter data on the root filesystem except in specific directories
- Maintain consistency in the root filesystem for the lifetime of an instance, regardless of workload activity
- Control the size of tmpfs allocated for each overlaid directory

## What it might look like?

The overlay would be customizable by allowing users to specify which directories should be writable to memory and with what size constraints:

```xml
    <preferences profiles="custom">
        <type
            image="oem"
            filesystem="xfs"
            kernelcmdline="console=ttyS0"
            firmware="uefi"
            format="vmdk"
            overlayroot="true"
            overlayroot_readonly_filesystem="erofs"
            overlayroot_readonly_partsize="2048"
            erofscompression="zstd,level=9"
            eficsm="false"
            bootpartition="false"
            efipartsize="200"
        >
            <oemconfig>
                <oem-resize>false</oem-resize>
            </oemconfig>
            <overlayroot targeted="true">
               <!-- Allow users to control the tmpfs for the each respctive overlay -->
               <path name="/etc" size="512MB">
               <path name="/home" size="2048MB">
            </overlayroot>
            <bootloader name="systemd-boot"/>
        </type>
    </preferences>
```

### Alternative Considered

We initially considered adding a custom dracut module that would mimic the overlay feature KIWI currently provides. However, since the implementation would be so similar to what KIWI already does, it seems better to extend KIWI's existing overlay capabilities.

We're flexible on the exact implementation details, but we believe this feature would be valuable for KIWI users who need more granular control over their overlay configurations.

Would the KIWI team be interested in adding a targeted overlay feature like this? We'd appreciate your thoughts on the feasibility and we are happy to help with a PR and/or discussion.

schaefi:
Actually this is a nice idea. The current overlay dracut code overlay the entire rootfs with a write space that can be either temporary or persistent. We could extend this feature to allow partial overlayfs mounts for customized parts of the rootfs.

schaefi:
I did some experiments with this and together with the patch in #2817 you can achieve the partial write access as follows.

In the kiwi image main description add the following kernel option

```
[...]

PR #2818, opened 2025-05-25
Update overlay integration test for partial write

schaefi:
Update the sdboot_uki_verity_erofs profile of the
test-image-overlayroot integration test with a custom fstab example to overlay only parts of the system
for writing. This is related to Issue #2815


schaefi:
According to the conversation on the associated issue it's ok to go with the example description provided here as a first step

PR #2819, opened 2025-05-26
Cleanup build metadata

### LINKED ISSUE #2668: Left over /image directory in images
## Problem description

I noticed that openSUSE MicroOS images have an empty `/images` directory in the root filesystem.

## Expected behaviour

The empty directory should be deleted by kiwi.

Apparently this is handled by [SystemSetup.cleanup](https://github.com/OSInside/kiwi/blob/8f4dc32124190344dd79203a2c5a4b9fcce9a089/kiwi/system/setup.py#L122), but it's not called anywhere.

https://build.opensuse.org/package/show/devel:microos:images/openSUSE-MicroOS is affected, but I guess any image with a `disk.sh` script triggers the same behaviour.

* KIWI version: 10.1.11
* Operating system host version: Tumbleweed
* Operating system target version: Tumbleweed

schaefi:
Make sure the final image rootfs does not contain unneeded metadata files used during build time. The respective cleanup call is performed after the root sync and after all initrd/boot processing has been done. This is because up to that point it's still possible that the information is required. This means when building images with a read-only rootfs, it might not be possible that the metadata can be deleted due to a chicken&egg situation. Furthermore the cleanup is applied to the disk builder only as other builders do not really suffer from this data and for the container builder the metadata can also be used for the stackbuild feature when building images derived from containers. This Fixes #2668

### DIFF
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -926,6 +926,10 @@ class DiskBuilder:
                             ]
                         )
 
+                    # cleanup build metadata
+                    if os.access(f'{root}/', os.W_OK):
+                        disk_system.cleanup()
+
                     # set root filesystem properties
                     if system:
                         self._setup_property_root_is_readonly_snapshot(system)
--- a/kiwi/system/setup.py
+++ b/kiwi/system/setup.py
@@ -168,9 +168,20 @@ class SystemSetup:
         Command.run(
             [
                 'chroot', self.root_dir,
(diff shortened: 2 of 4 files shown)


schaefi:
I'll create a Staging build to check for any regressions. No merge until then please

schaefi:
Integration test build

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-MicroOS

looks good to me now

Issue #2445, opened 2024-01-31
Add loop-less disk builder (systemd-repart)


schaefi:
## Description

The current implementation for the disk builder makes use kernel interfaces, meaning the produced devices are created through loop devices as well as the produced file-systems gets mounted and more. Under certain conditions the dependency into the kernel could cause a problem e.g if the calling user has not enough permissions or the environment does not expose access to these interfaces. In this case it would be handy if kiwi could offer a disk builder which does not require kernel interfaces.

The systemd project offers a tool called ```systemd-repart``` which implements many parts of the build process without the need for the kernel interfaces. This leads to the idea to write a new builder class that utilizes  ```systemd-repart``` to build the disk image. systemd-repart is also used in the mkosi image builder from which the idea originates. For reference see make_image():

* https://github.com/systemd/mkosi/blob/main/mkosi/__init__.py#L2621

kiwi needs to implement an interface to ```systemd-repart``` and its required input files

schaefi:
on suse ```systemd-repart``` is provided with the ```systemd-experimental``` package

Conan-Kudo:
This might become more important to resolve issues around running kiwi in container environments.

schaefi:
yes this is a known problem in containers when no udev is running inside. You can fix this as follows

Make sure your container comes with a file named `/etc/kiwi.yml` inside of the file set:

```
mapper:
  - part_mapper: kpartx
```

rebuild the image. kiwi will now use kpartx which doesn't require udev to create the device nodes.

Your container still needs to be privileged though to allow creation of device nodes

bnicolae86:
> yes this is a known problem in containers when no udev is running inside. You can fix this as follows
> 
> Make sure your container comes with a file named `/etc/kiwi.yml` inside of the file set:
> 
> ```
> mapper:
>   - part_mapper: kpartx
> ```
> 
> rebuild the image. kiwi will now use kpartx which doesn't require udev to create the device nodes.
> 
> Your container still needs to be privileged though to allow creation of device nodes

thanks, I'll try it

bnicolae86:
> > yes this is a known problem in containers when no udev is running inside. You can fix this as follows
> > Make sure your container comes with a file named `/etc/kiwi.yml` inside of the file set:
> > ```
> > mapper:
> >   - part_mapper: kpartx
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
[...]

Issue #2293, opened 2023-05-04
While building in an nspawn container loop device partitions are not found


PhilipSnell:
## Problem description
I am trying to build in a nspawn container and it is failing to find the loop partitions

```
[ DEBUG   ]: 22:40:03 | Initialize gpt disk
[ DEBUG   ]: 22:40:03 | EXEC: [sgdisk --zap-all /dev/loop0]
[ INFO    ]: 22:40:04 | --> creating EFI CSM(legacy bios) partition
[ DEBUG   ]: 22:40:04 | EXEC: [sgdisk -n 1:2048:+2M -c 1:p.legacy /dev/loop0]
[ DEBUG   ]: 22:40:05 | EXEC: [sgdisk -t 1:EF02 /dev/loop0]
[ INFO    ]: 22:40:07 | --> creating EFI partition
[ DEBUG   ]: 22:40:07 | EXEC: [sgdisk -n 2:0:+20M -c 2:p.UEFI /dev/loop0]
[ DEBUG   ]: 22:40:08 | EXEC: [sgdisk -t 2:EF00 /dev/loop0]
[ INFO    ]: 22:40:09 | --> creating SWAP partition
[ DEBUG   ]: 22:40:09 | EXEC: [sgdisk -n 3:0:+128M -c 3:p.swap /dev/loop0]
[ DEBUG   ]: 22:40:10 | EXEC: [sgdisk -t 3:8200 /dev/loop0]
[ INFO    ]: 22:40:11 | --> using all_freeMB for the root(rw) partition if present
[ INFO    ]: 22:40:11 | --> creating root partition [with 0 clone(s)]
[ DEBUG   ]: 22:40:11 | EXEC: [sgdisk -n 4:0:0 -c 4:p.lxroot /dev/loop0]
[ DEBUG   ]: 22:40:12 | EXEC: [sgdisk -t 4:8300 /dev/loop0]
[ DEBUG   ]: 22:40:13 | EXEC: [partx --add /dev/loop0]
[ ERROR   ]: 22:40:13 | KiwiMappedDeviceError: Device /dev/loop0p1 does not exist
```

## Expected behaviour
Loop partitions should be created and formatted properly during the build process

## Steps to reproduce the behaviour
- Create an nspawn container (I am using Debian Bullseye)
- install kiwi, rsync, qemu-utils and gdisk
- create loop device 
`mknod /dev/loop0 b 7 0`
- build debian buster from kiwi descriptions
[...]

PR #2826, opened 2025-06-02
Add overlayfs  as supported

dirkmueller:
Currently kiwi can no longer be run inside a container as it rejects overlayfs. Is there a real reason for this? extended attributes are working there.

* add overlayfs
* refactor slightly

### DIFF
--- a/kiwi/runtime_checker.py
+++ b/kiwi/runtime_checker.py
@@ -91,25 +91,21 @@ class RuntimeChecker:
             ACLs or xattrs. The image build may fail or the resulting
             image misbehave.
         ''')
-        support_target_filesystem = [
-            'btrfs',
-            'ext2',
-            'ext2/ext3',
-            'ext3',
-            'ext4',
-            'overlayfs',
-            'tmpfs',
-            'xfs',
-        ]
-        stat = Command.run(
-            ['stat', '-f', '-c', '%T', target_dir]
-        )
+        stat = Command.run(['stat', '-f', '-c', '%T', target_dir])
         if stat:
             target_fs = stat.output.strip()
-            if target_fs not in support_target_filesystem:
-                raise KiwiRuntimeError(
-                    message.format(target_fs, target_dir)
-                )
+            supported_target_filesystem = (
+                'btrfs',
+                'ext2',
+                'ext2/ext3',
+                'ext3',
+                'ext4',
+                'overlayfs',
+                'tmpfs',
+                'xfs',
+            )
+            if target_fs not in supported_target_filesystem:
+                raise KiwiRuntimeError(message.format(target_fs, target_dir))
 
     def check_include_references_unresolvable(self) -> None:
         """
--- a/kiwi/runtime_checker.py
+++ b/kiwi/runtime_checker.py
@@ -92,7 +92,14 @@ class RuntimeChecker:
             image misbehave.
         ''')
         support_target_filesystem = [
-            'tmpfs', 'btrfs', 'xfs', 'ext2', 'ext3', 'ext4', 'ext2/ext3'
+            'btrfs',
+            'ext2',
+            'ext2/ext3',
+            'ext3',
+            'ext4',
+            'overlayfs',
+            'tmpfs',
+            'xfs',
         ]
         stat = Command.run(
             ['stat', '-f', '-c', '%T', target_dir]


Conan-Kudo:
There are other problems too. We still have issues with loop devices inside of containers. And device mapper is kind of broken in there too.

dirkmueller:
what issues? how to reproduce? we're currently shipping this and haven't received a bugreport yet (from QA or users)

Conan-Kudo:
It's the reason #2445 exists. We have issues with Docker/Podman based environments (which are now limited to using LVM/DM/LUKS thanks to #2208), we have issues with Nspawn based environments (because partitions in loop devices cannot be created, see #2293).

As a project, we recommend using microVMs instead of containers because it's extremely problematic to support all the different image types properly in containers.

schaefi:
>Currently kiwi can no longer be run inside a container as it rejects overlayfs. Is there a real reason for this?

yes reason was #2806 

Can you guarantee that a host on overlayfs (e.g container env) supports creating data with required extended attributes ? Just to mention this is only one out of a number of other problems you can run into when building inside of a container. If you build your image inside of a container you should really investigate first what type of image is built and if the produced result is correct. We already spent hours and hours of debugging strange bugs and after effects all caused by building images in limited runtime environments.

The check can easily be disabled

* /etc/kiwi.yml

```
runtime_checks:
  - disable:
      - check_target_dir_on_unsupported_filesystem
``` 

I'm happy to merge this PR if you can confirm that overlayfs by default supports ACLs and xattrs

dirkmueller:
> > Currently kiwi can no longer be run inside a container as it rejects overlayfs. Is there a real reason for this?
> yes reason was #2806

I've seen the pull request, but it didn't have a rationale or a reproducer on what the issue was. 

> Can you guarantee that a host on overlayfs (e.g container env) supports creating data with required extended attributes

Well, overlayfs is by itself implemented using xattrs, so I would think it always works? https://docs.kernel.org/filesystems/overlayfs.html#whiteouts-and-opaque-directories 

I have never seen it not being available. obviously I'm not a kiwi maintainer which is why I am searching for information for the check being the way it is. the only issue I've seen so far is selinux or fips being enforcing on the host, but kiwi can not deal with that even without overlayfs so the current check seems to not be able to prevent that either.

> If you build your image inside of a container you should really investigate first what type of image is built and if the produced result is correct.

We're doing that, that's why I found this change because it now refuses to work while it worked just fine before. 

> We already spent hours and hours of debugging strange bugs and after effects all caused by building images in limited runtime environments.

But how about we simply test for the issue that we want to not have in the code rather than maintaining an allowlist of filesystems?  we could be creating a file and set a xattr and a fACL on it and see if that succeeds ? the only challenge I see is that kiwi doesn't depend on any python module that would bring that functionality, so I would potentially have to call some CLI tools (that might or might not be installed)?

e.g. if selinux prevents setting xattr on an overlayfs, we could simply set an xattr and see if that succeeds ?

> The check can easily be disabled
> runtime_checks:
>   - disable:
>       - check_target_dir_on_unsupported_filesystem

I understand. But why did you implement the check if you want me to ship a container that disables the check? :slightly_smiling_face: 

> I'm happy to merge this PR if you can confirm that overlayfs by default supports ACLs and xattrs

I'm happy to close this PR if we can somehow confirm that overlayfs is problematic. but we are currently shipping this as the recommended solution to customers ( https://registry.suse.com/repositories/bci-kiwi10 ) and now this is no longer working.

schaefi:
> I've seen the pull request, but it didn't have a rationale or a reproducer on what the issue was.

#2806  was the result of a long conversation on the help-kiwi slack channel in suse. We couldn't add a reference to this conversation into the PR as it was company internal. The description of the PR is short, I agree. However, I had hoped that the code change and the provided runtime check message explains the reasoning well enough.

> Well, overlayfs is by itself implemented using xattrs

yes, it uses extended attributes to manage itself for upper and lower, especially when files gets removed. I did a bit of a research and actually we will run into extended file permission issues with overlayfs only if the upper is a filesystem that doesn't support them. e.g. a tmpfs.

As tmpfs is in the allowlist I have no objections to add your change. 

> But how about we simply test for the issue that we want to not have in the code rather than maintaining an allowlist of filesystems?

Because the number of issues caused by running on a limited target filesystem might not be complete and the potential to still build something that is broken without knowing it still persists. Someone build on a 9p filesystem and everything works until the level of symbolic links was too deep. So you really want to explicitly check for all weirdness that can arise ? It's way cleaner to say we know it works on a,b,c and if you use something different you can disable our check and hope for the best. So this is not only about xattrs and acls, it's about a certain level of confidence that it will work on the whitelist

> I understand. But why did you implement the check if you want me to ship a container that disables the check?

To make you aware that you explicitly disabled a check that exists for good reasons and if things goes wrong you might be reminded that it could be because of this reason. To be clear.  I don't want you to disable the check. I want you to keep it. But I can understand that you might have the need to disable it according to your use case.

If there are no objections @Conan-Kudo I would be ok with this PR

@dirkmueller the tests needs to pass though. Let me know if you want me to help with that part

Thanks

schaefi:
@dirkmueller 

```
kiwi/builder/disk.py:1136: error: Argument 1 to "create_boot_partition" of "Disk" has incompatible type "int"; expected "str"  [arg-type]
```
just rebase to latest main, this will fix it

Conan-Kudo:
I'm okay with relaxing the check, but I'd like a warning to be printed just so we know when people give us logs. And the tests need to pass.

dirkmueller:
> yes, it uses extended attributes to manage itself for upper and lower, especially when files gets removed. I did a bit of a research and actually we will run into extended file permission issues with overlayfs only if the upper is a filesystem that doesn't support them. e.g. a tmpfs.

I believe this is actually the case for container runs by default? can you share what exactly the issue is with tmpfs or how to replicate this?

> Because the number of issues caused by running on a limited target filesystem might not be complete and the potential to still build something that is broken without knowing it still persists. Someone build on a 9p filesystem and everything works until the level of symbolic links was too deep. So you really want to explicitly check for all weirdness that can arise ?

no, I don't think there are too many users of 9p :-) 

> @dirkmueller the tests needs to pass though. Let me know if you want me to help with that part

Tests are passing after rebase.

schaefi:
>but I'd like a warning to be printed

It's a runtime check that prints a message and exits the build. If you skip the check a warning is placed in the log that the `check_target_dir_on_unsupported_filesystem` check was skipped. That should be imho enough to give you a debugging pointer

schaefi:
I'm uncertain about `tmpfs` from the list. It does not support xattr and it says the following for ACLs

```
Setting POSIX Access Control Lists (ACLs) on a tmpfs mount point might not always be supported or may behave unexpectedly
```

whereas

* https://docs.kernel.org/filesystems/tmpfs.html

says

```
tmpfs also supports POSIX ACLs, and extended attributes for the trusted.*, security.* and user.* namespaces.
```

thoughts ?

Issue #2808, opened 2025-05-14
Enhanced Dracut Configuration Options in Image Description


aale24:
### Overview

Hello, this ticket derives from https://github.com/OSInside/kiwi/issues/2801

This issue proposes exposing dracut configuration capabilities directly in the kiwi image description XML, allowing users to configure dracut options without requiring manual overlay files.

Note: This proposal aims to provide flexibility while supporting distribution limitations like inexistent modules.

We've experimented with manual dracut and are willing to implement this enhancement as a PR if there's interest.

### Current State:

Dracut configuration requires manual overlay files (e.g., etc/dracut.conf.d/some.conf) to set options.

### Proposed Enhancement:

Add support for the following dracut configuration options as XML elements under `<bootloader><dracut>`:

1. Installation of additional tools: 
    ```
    <install paths="/usr/bin/veritysetup /path/to/debug_tool"/>
    ```
4. Kernel module configuration:
   ```
    <add_drivers names="driver1 driver2"/>
    <force_drivers names="driver3 driver4"/>
    <omit-drivers names="driver5 driver6"/>
    ```
3. Additional dracut modules:
    ```
    <add modules="module1 module2"/>
    <omit modules="module3 module4"/>
    <force-add modules="module5 module6"/>
   ```
4. Filesystem support:
    ```
    <filesystems names="ext4 xfs btrfs"/>
    ```
5. Compression method:
    ```
    <compress method="gzip|bzip2|lzma|xz|lzo|lz4|zstd"/>
    ```
6. Kernel command line options:
    ```
    <kernel_cmdline>root=UUID=xxx dm-mod.create=".." rd.debug</kernel_cmdline>
    ```
    _Note: This option already exists as an attribute in the image type element. We need to determine the best placement for this option and how it should interact with the existing attribute._ 
7. Persistent policy:
   ```
   <persistent-policy value="by-uuid|by-label|mapper"/>
   ```

### Example Configuration:
```
<bootloader name="systemd_boot" console="serial">
    <dracut>
       <kernel_cmdline>root=/dev/mapper/root dm-mod.create=".." rd.debug</kernel_cmdline>
        <install paths="/usr/bin/veritysetup ..."/>
        <filesystems names="erofs ext4 ..."/>
        <add_drivers names= "erofs ext4 ..." />
        <force_drivers names="erofs ext4 ..."/>
        <add modules="dm-verity ..."/>
       ...
    </dracut>
</bootloader>
```

### Benefits:

[...]

PR #2832, opened 2025-06-08
Add driver configuration support for dracut initrd

aale24:
Add support for specifying kernel drivers to be included or omitted in the dracut initrd configuration. This extends the existing dracut configuration capabilities by allowing users to:

- Configure specific drivers through the <dracut> element: <dracut driver="driver_name"/>

- Control driver inclusion/omission in both system and installation initrds through:
   - add_drivers/omit_drivers in dracut.conf
   - --drivers/--omit-drivers command line options

Key changes:
- Add driver attribute to dracut schema definition
- Implement driver handling in BootImageDracut
- Add driver support in DiskBuilder and InstallImageBuilder
- Update XMLState to process driver configurations
- Add comprehensive test coverage for driver handling

Example usage:
<initrd action="add">
    <dracut driver="erofs"/>
</initrd>

Partially Fixes [#2808](https://github.com/OSInside/kiwi/issues/2808)


kiwi/boot/image/dracut.py
```diff
@@ -340,13 +394,43 @@ def _get_modules(self) -> List[str]:
...
+            for line in cmd.output.splitlines():
+                if line:
+                    driver_path = line.split(':')[0].strip()
+                    driver_name = os.path.basename(driver_path).split('.ko')[0]
+                    drivers.append(driver_name)
+            return drivers
+        except Exception as e:
+            log.warning(f"Error reading drivers: {str(e)}")
+            return []
+
```
schaefi: We offer a `Kernel` class that simplifies retrieving data like the kernel version. Can we make use of this code instead of sort of duplicating the code. e.g.

```python3
from kiwi.system.kernel import Kernel
kernel = Kernel(root_dir)
kernel_info = kernel.get_kernel()
print(kernel_info.version)
```

feel free to extend the Kernel interface if need be. Thanks much

kiwi/xml_parse.py
```diff
@@ -9007,9 +9007,10 @@ class dracut(GeneratedsSuper):
     """A dracut module"""
     subclass = None
     superclass = None
-    def __init__(self, module=None, uefi=None):
+    def __init__(self, module=None, driver=None, uefi=None):
```
schaefi, line 9010: Maybe a stupid question and just to put me right. The `xml_parse.py` is like the `schema/kiwi.rng` an auto generated file. So am I correct that you only modified `schema/kiwi.rnc` and then called `make kiwi/schema/kiwi.rng` to auto generated the API ? If yes, all is good, if not and you have manually modified  `xml_parse.py` or `schema/kiwi.rng` then please let it create through the make target. Thanks much

schaefi (changes requested):
This looks great, thanks much for your work. Just one remark of a Kernel class that we should re-use and a question regarding the creation of auto generated data. Thanks much

schaefi:
Oh and I forgot to mention we need at least some words about the new elements in the documentation. Please find it here:

* doc/source/image_description/elements.rst

There is already documentation for the existing elements, just add the new ones as you see fit. Thanks

aale24:
Thanks for the comments, I will be working on them.

kiwi/xml_parse.py
aale24: In the latest commit, I ran `make kiwi/schema/kiwi.rng`, but I had to manually adjust the output in `kiwi/xml_parse.py` to use f-string formatting because generateDS.py tool doesn't generate the output using f-string format

kiwi/boot/image/dracut.py
aale24: Now using `kernel_info.version` to obtain the kernel version. Thank you for suggesting using the Kernel class.

kiwi/xml_parse.py
schaefi: @aale24 I can keep this for now but please have in mind that the next schema changes might revert this back to the way how the `generatedDS` module handles it. Did you had a programmatic way to move from format sequences into `f` strings ? If yes we should add that to the Makefile as a step after generateDS.

In addition I could not find an issue in the auto generated API with regards to format sequences vs. f-strings. So did you had a specific reason why this change was required ?

Thanks

aale24: @schaefi. You're correct that this situation requires some clarification:
1. The change to f-strings was initially made in this commit: https://github.com/OSInside/kiwi/commit/28ed67e45674f1e06c7db882d89d7cac5438258e
1.  When I ran make `kiwi/schema/kiwi.rng`, the generateDS tool reverted the f-strings back to format sequences in `kiwi/xml_parse.py`.
1. To maintain consistency with the earlier commit and to avoid introducing unnecessary changes in the PR, I manually adjusted the output in `kiwi/xml_parse.py` back to f-strings after running the make command.

You're right that this manual adjustment isn't ideal, as future schema changes might revert it again. Currently, I don't have a programmatic way to convert format sequences to f-strings automatically but I could create an issue and work on that. Let me know if you want to use the f-string format in `kiwi/xml_parse.py`

PR #2833, opened 2025-06-08
Add dependency for isomd5sum for iso images and set in kiwi-settings

Conan-Kudo:
This ensures that isomd5sum is pulled into the environment for ISO image builds, and the updated settings makes it so that kiwi boxes will use it.


schaefi (approved):
Thanks, looks good :+1:

PR #2840, opened 2025-06-25
Add support for container-snap as a container-image engine

dcermak:
With this commit, we can now pre-load images using container-snap directly during the kiwi image build

*
*


schaefi (approved):
Looks great to me :+1: 

any reason why it's in draft mode ?

dcermak:
> any reason why it's in draft mode ?

I need to test whether it actually works first :sweat_smile:

dcermak:
Yep, this works for me locally

schaefi:
> Yep, this works for me locally

great, let's add it :) Thanks

Issue #2358, opened 2023-08-22
support reproducible builds


JanZerebecki:
## Problem description

When building the same image twice the output is different. Fix this by making the kiwi output reproducible. See https://reproducible-builds.org/ from more info on the general idea.

## Expected behaviour

Output of kiwi is reproducible.

## Steps to reproduce the behaviour

Branch from the Factory with: osc branch --nodevelproject openSUSE:Factory/openSUSE-Tumbleweed-Yomi
Download image after it is done: osc getbinaries home:jzerebecki:branches:openSUSE:Factory/openSUSE-Tumbleweed-Yomi:livecd images/x86_64
Rebuild: osc rebuild home:jzerebecki:branches:openSUSE:Factory/openSUSE-Tumbleweed-Yomi:livecd images x86_64
Download again, build result is different.
```
> diff -sr binaries.kiwicd.1/openSUSE-Tumbleweed-Yomi.x86_64-1.0.0-livecd-Build21.1.iso binaries.kiwicd.2/openSUSE-Tumbleweed-Yomi.x86_64-1.0.0-livecd-Build21.2.iso
Binary files binaries.kiwicd.1/openSUSE-Tumbleweed-Yomi.x86_64-1.0.0-livecd-Build21.1.iso and binaries.kiwicd.2/openSUSE-Tumbleweed-Yomi.x86_64-1.0.0-livecd-Build21.2.iso differ
```

The following images show this:
iso: https://build.opensuse.org/package/show/openSUSE:Factory/openSUSE-Tumbleweed-Yomi livecd

Not tested, but possibly differently affected:
docker: https://build.opensuse.org/package/show/openSUSE:Factory/minimal-image
vmx oem btrfs xfs vhdx vmdk qcow2: https://build.opensuse.org/package/show/openSUSE:Factory/kiwi-templates-Minimal

Perhaps some PRs I made to make the product-builder reproducible are useful:
https://github.com/openSUSE/product-builder/pull/26
https://github.com/openSUSE/product-builder/pull/27
https://github.com/openSUSE/product-builder-plugins/pull/16

## OS and Software information

* KIWI version: python3-kiwi-9.25.12-1.1
* Operating system host version: obs Factory
* Operating system target version: obs Factory
* Open Build Service version (N/A if not using OBS): 2.11~alpha.20230822T165508.31116c20b
* Koji version (N/A if not using Koji): N/A

lslezak:
[...]

PR #2845, opened 2025-07-01
Ensure dracut initrd is reproducible

bmwiedemann:
This helps a bit with issue #2358

* call `dracut` with `--reproducible` option to create reproducible `initrd` files.


schaefi:
@bmwiedemann do we know what exactly this flag does behind the scene ?

If this is a safe change I have no objections. Can you please update the corresponding unit test too ?

Thanks much

bmwiedemann:
From looking at
https://github.com/dracutdevs/dracut/blob/5d2bda46f4e75e85445ee4d3bd3f68bf966287b9/dracut.sh#L2301
it touches files+dirs to ensure fixed mtimes and calls `cpio` with `--reproducible` to not embed inode numbers into cpio headers.

It should be safe.

schaefi:
looks good, let's clarify if the change is compatible with the supported distro matrix. I will zip a Staging build and wait for the integration test runs. If all is good we are safe to merge

schaefi:
All tests pass

schaefi (approved):

PR #2846, opened 2025-07-01
Make mbr-id deterministic

bmwiedemann:
helps #2358

* use `SOURCE_DATE_EPOCH` value to seed the random number generator to ensure mbrid is deterministic

### DIFF
--- a/kiwi/system/identifier.py
+++ b/kiwi/system/identifier.py
@@ -15,11 +15,15 @@
 # You should have received a copy of the GNU General Public License
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
+import logging
+import os
 import random
 import struct
 
 # project
 from kiwi.storage.device_provider import DeviceProvider
+
+log = logging.getLogger('kiwi')
 
 
 class SystemIdentifier:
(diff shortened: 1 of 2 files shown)


schaefi:
Thanks, nice idea. I will fix the tests as discussed

kiwi/system/identifier.py
```diff
@@ -32,6 +36,10 @@ class SystemIdentifier:
     """
     def __init__(self):
         self.image_id = None
+        sde = os.environ.get('SOURCE_DATE_EPOCH')
+        if sde:
+            log.info('Using SOURCE_DATE_EPOCH as random seed.')
```
schaefi: can we change this to

```python3
log.info(f'Using SOURCE_DATE_EPOCH as random seed: {sde}')
```

Conan-Kudo: Yes please, we need to know what value is used.

bmwiedemann: Done

schaefi:
@bmwiedemann for me this change looks good. You marked it as a WIP, so unless you plan further changes I believe we can merge this one and see if it causes any trouble on the Staging builds. Thoughts ?

schaefi (approved):

PR #2847, opened 2025-07-02
Seed filesystem UUIDs with label info

### LINKED ISSUE #2761: Allow defining filesystem UUIDs in `<partition>`
From a quick skim of [the documentation](https://osinside.github.io/kiwi/working_with_images/custom_partitions.html), and a chat with @Conan-Kudo, it's currently not possible in `kiwi` to control the UUIDs used for filesystems.

It'd be nice if it could so higher level tools can pin the UUIDs they want to use. This would allow to make builds a tiny bit more reproducible (for some value of it).

schaefi:
For reproducible builds the calculation of the filesystem UUID should be persistent with each rebuild of the image. To achieve this the UUID is calculated using the label name as seed. Only in case the filesystem has no label assigned the UUID stays random and changes with each rebuild. In addition to this change a new label attribute has been added to the <partition> section. This Fixes #2761

### DIFF
--- a/doc/source/working_with_images/custom_partitions.rst
+++ b/doc/source/working_with_images/custom_partitions.rst
@@ -65,11 +65,16 @@ size="size_string"
   specification mega-Byte is used.
 
 mountpoint="path"
-  Mandatory mountpoint to mount the partition in the system.
+  Optional mountpoint to mount the partition in the system.
 
 filesystem="btrfs|ext2|ext3|ext4|squashfs|xfs
-  Mandatory filesystem configuration to create one of the supported
+  Optional filesystem configuration to create one of the supported
   filesystems on the partition.
+
+label="string"
+  Optional filesystem label if a filesystem is provided. If no
+  label is specified the `name` identifier is set as filesystem
+  label
 
 clone="number"
   Optional setting to indicate that this partition should be
--- a/kiwi/filesystem/base.py
+++ b/kiwi/filesystem/base.py
@@ -16,8 +16,11 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
+import uuid
+import random
 import logging
 import copy
+from functools import reduce
 from typing import (
     Dict, List, Optional
 )
@@ -315,6 +318,19 @@ class FileSystemBase:
         Raises
         """
         raise NotImplementedError
+
+    def _generate_seed_uuid(self, label: str, random_bits: int = 128) -> str:
+        """
+        Create random UUID from the specified label name as seed
+
+        This ensures the same UUID for the same seed
+        """
+        label_seed = reduce(lambda x, y: x + y, map(ord, label))
+        rd = random.Random()
+        rd.seed(label_seed)
+        return format(
+            uuid.UUID(int=rd.getrandbits(random_bits))
+        )
 
     def _map_size(self, size: float, from_unit: str, to_unit: str) -> float:
         """
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -1008,6 +1008,7 @@ class DiskBuilder:
                             squashed_root_file = Temporary().new_file()
                             filesystem.create_on_file(
                                 filename=squashed_root_file.name,
+                                label=ptable_entry.label,
                                 exclude=[Defaults.get_shared_cache_location()]
                             )
                             readonly_target = device_map[map_name].get_device()
@@ -1034,7 +1035,7 @@ class DiskBuilder:
                             )
                         else:
                             filesystem.create_on_device(
-                                label=map_name.upper()
+                                label=ptable_entry.label or map_name.upper()
                             )
                         filesystem_dict[map_name] = filesystem
         return filesystem_dict
--- a/kiwi/filesystem/btrfs.py
+++ b/kiwi/filesystem/btrfs.py
@@ -42,15 +42,18 @@ class FileSystemBtrfs(FileSystemBase):
         :param str uuid: UUID name
         """
         device = self.device_provider.get_device()
+        call_args = self.custom_args['create_options'].copy()
+        if not uuid and label:
+            uuid = self._generate_seed_uuid(label)
         if label:
-            self.custom_args['create_options'].append('-L')
-            self.custom_args['create_options'].append(label)
+            call_args.append('-L')
+            call_args.append(label)
         if uuid:
-            self.custom_args['create_options'].append('-U')
-            self.custom_args['create_options'].append(uuid)
+            call_args.append('-U')
+            call_args.append(uuid)
         if size:
-            self.custom_args['create_options'].append('--byte-count')
-            self.custom_args['create_options'].append(
+            call_args.append('--byte-count')
+            call_args.append(
                 self._fs_size(
                     size=self._map_size(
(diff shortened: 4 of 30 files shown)


schaefi:
waiting for integration tests to complete

Conan-Kudo:
I'm confused? I think the idea was the control the PRNG, not use the label itself as a seed.

This creates some weird effects, particularly when images aren't intentionally doing this. For example, Fedora Asahi Remix systems all have the partition label "fedora" (as do Fedora Cloud images). Having a fixed UUID automatically derived from the label means that it's now impossible to produce disk images on those systems since the UUIDs _and_ labels match, creating mount confusion.

schaefi:
> I'm confused? I think the idea was the control the PRNG, not use the label itself as a seed.
> 
> This creates some weird effects, particularly when images aren't intentionally doing this. For example, Fedora Asahi Remix systems all have the partition label "fedora" (as do Fedora Cloud images). Having a fixed UUID automatically derived from the label means that it's now impossible to produce disk images on those systems since the UUIDs _and_ labels match, creating mount confusion.

Hmm, I can't follow you on this. label and uuids are not the same. For the calculation of the UUID the "randomness" is disabled by providing a seed, such that a rebuild of the image comes with the same UUID. The systemd provided standard UUID definitions for partitions are the same UUIDs for the same type of partition. This code now "standardize" the filesystem UUID according to its label. Example:

kiwi applies the label name 'ROOT' for the rootfs. This will then result in the following UUID for the rootfs

```python3
import random
import uuid
from functools import reduce
label='ROOT'
label_seed = reduce(lambda x, y: x + y, map(ord, label))
rd = random.Random()
rd.seed(label_seed)
print(uuid.UUID(int=rd.getrandbits(128)))
```

* dbbfa6df-b766-9114-2b15-e2da971029da

And it will be the same as long as the label is the same. For the custom partitions (those you specify yourself) you can specify a custom label which is used as the seed for the UUID. That way you create a connection between your label and your UUID but I don't see any conflict

If no label exists, we stay random

?

schaefi:
All tests passing, but we need to clarify what we really want to achieve. Maybe also a topic for tomorrows community meeting

schaefi:
I can imagine a device conflict for the following condition:

* you build two different images
* kiwi uses systemd defined partition UUIDs and with this change kiwi uses label defined filesystem UUIDs
* the two images are deployed on two disk of the same machine
* on boot it wouldn't be possible to differentiate the device from one or the other disk because /dev/disk/by-partuuid and /dev/disk/by-uuid would offer the same for both disks and by default by-uuid is used for booting.

How likely is this ? systemd is aware of this conflict when systems are setup to boot using the standard IDs

I believe we should at least add an opportunity to configure the seed for those filesystems where it is currently not possible (root, boot, swap)

Conan-Kudo:
I think it's likely enough that I'd be concerned. The idea that @davide125 and @supakeen had was to allow setting a time-value seed for UUIDs (e.g. `SOURCE_DATE_EPOCH` variable if available).

schaefi:
So then let's update the code to make use of `SOURCE_DATE_EPOCH` if available

schaefi:
The problem here is that you can't use `SOURCE_DATE_EPOCH` for all filesystems because that would create UUID conflicts. So how do we want to differentiate them ?

SOURCE_DATE_EPOCH + 1
SOURCE_DATE_EPOCH + 2 
...

and so on ?

Conan-Kudo:
I would use `-1` instead of `+1` (it is a time signature), but otherwise yeah.

davide125:
Yeah, I agree with @Conan-Kudo, it's not at all uncommon to deploy multiple images on the same system. Using SOURCE_DATE_EPOCH - offset seems reasonable to me.

schaefi:
all right, so let's take SOURCE_DATE_EPOCH into account

doc/source/working_with_images/custom_partitions.rst
```diff
@@ -65,12 +65,17 @@ size="size_string"
...
 filesystem="btrfs|ext2|ext3|ext4|squashfs|xfs
-  Mandatory filesystem configuration to create one of the supported
+  Optional filesystem configuration to create one of the supported
   filesystems on the partition.
 
+label="string"
+  Optional filesystem label if a filesystem is provided. If no
+  label is specified the `name` identifier is set as filesystem
+  label
+
```
Conan-Kudo, line 78: Is this still valid documentation?

kiwi/filesystem/base.py
```diff
@@ -316,6 +319,28 @@ def set_property_readonly_root(self) -> None:
         """
         raise NotImplementedError
 
+    def _generate_seed_uuid(self, label: str, random_bits: int = 128) -> str:
+        """
+        Create random UUID. If SOURCE_DATE_EPOCH is present use
+        SOURCE_DATE_EPOCH + label name as seed
+        """
+        sde = os.environ.get('SOURCE_DATE_EPOCH')
```
Conan-Kudo, line 327: I think it'd be a good idea to allow a command-line option to set this too. If `SOURCE_DATE_EPOCH` doesn't exist (or even if it does!), I think it's worth exposing a CLI knob to provide a user-set timestamp seed.

Conan-Kudo (changes requested):

doc/source/working_with_images/custom_partitions.rst
schaefi: ah no, I overlooked that part, thanks for pointing it out

kiwi/filesystem/base.py
schaefi: yeah let's allow that

doc/source/working_with_images/custom_partitions.rst
schaefi: yes it is. It's not directly related to the uuid and seed initiative but while implementing I found the documentation about the label handling was wrong and incomplete. I keep this change as an extra commit

kiwi/filesystem/base.py
schaefi: done, with the last commit you can influence the caller environment and run kiwi like this 

```
kiwi-ng --setenv SOURCE_DATE_EPOCH=42 system build ...
```

Conan-Kudo:
We currently cannot set the fliesystem label for the ESP (or even mount options), which is why we have this in Fedora's kiwi descriptions: https://pagure.io/fedora-kiwi-descriptions/blob/rawhide/f/root/etc/fstab.script

schaefi:
> We currently cannot set the fliesystem label for the ESP (or even mount options), which is why we have this in Fedora's kiwi descriptions: https://pagure.io/fedora-kiwi-descriptions/blob/rawhide/f/root/etc/fstab.script

Hmm, I don't get that part. So with this change here, the EFI filesystem will be created like this

```
[ INFO    ]: 09:39:20 | Using UUID seed SOURCE_DATE_EPOCH:42 + LABEL:EFI=212
...
[ DEBUG   ]: 09:39:20 | EXEC: [mkdosfs -F16 -I -n EFI -i 68FAD047 /dev/loop0p1]
```

So you are saying we have a problem here ?

Conan-Kudo:
We need a way to set the shortname behavior and mount settings for the ESP because they affect the reproducibility of image. This should be a separate pull request, though.

PR #2849, opened 2025-07-06
Fix regression in get_partition_node_name

uecasm:
Fixes part of [this issue](https://groups.google.com/g/kiwi-images/c/-hcVSbMSWaw/m/3906dsdSAwAJ)

* Fixes regression in get_partition_node_name for targets that do not support the `START` column in `lsblk`.
    * This restores prior behaviour, which may reintroduce [the bug that was intended to fix](https://github.com/OSInside/kiwi/commit/246a478d490e85187366ca5717627fc7601c737a), but that's better than breaking existing functionality.

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -191,6 +191,7 @@ function get_partition_node_name {
     local index=1
     local part
     udev_pending
+    # backwards compat for lsblk before 2.38: if START column not supported, fall back to default sort
     for partnode in $(
         { lsblk -p -l -o NAME,TYPE,START -x START "${disk}" 2>/dev/null ||\
         lsblk -p -l -o NAME,TYPE "${disk}"; } |\
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -192,7 +192,8 @@ function get_partition_node_name {
     local part
     udev_pending
     for partnode in $(
-        lsblk -p -l -o NAME,TYPE,START -x START "${disk}" |\
+        { lsblk -p -l -o NAME,TYPE,START -x START "${disk}" 2>/dev/null ||\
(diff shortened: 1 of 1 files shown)


schaefi:
@uecasm Uhh, good catch thanks.

Can you give me a short background explanation for which devices we hit the regression ? The places in code where we call it targets the storage disk which is expected to hold a partition table and thus proper start addresses. Thanks

schaefi:
Ah forget my last question, you referenced an issue link.. just reading

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -192,7 +192,8 @@ function get_partition_node_name {
     local part
     udev_pending
     for partnode in $(
-        lsblk -p -l -o NAME,TYPE,START -x START "${disk}" |\
+        { lsblk -p -l -o NAME,TYPE,START -x START "${disk}" 2>/dev/null ||\
```
schaefi, line 196: Can we add a comment to the method providing a bit of a reason for the two stage concept ?

schaefi:
A reproducer to the issue would be great. I believe if you can send

```
sfdisk -d /dev/nvme0n1 > table
```

From an affected system, I should be able to reproduce this also on a loop device

Thoughts ?

uecasm:
This isn't device-specific, it's an issue with the version of `lsblk`.  As noted, I'm testing on a Leap 15.5 target and the version in there does not support this column.  The START column was not added until util-linux-systemd 2.38, and 15.5 has 2.37.4.

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
uecasm: Done.

schaefi:
> This isn't device-specific, it's an issue with the version of `lsblk`. As noted, I'm testing on a Leap 15.5 target and the version in there does not support this column. The START column was not added until util-linux-systemd 2.38, and 15.5 has 2.37.4.

I see, thanks and the comment makes that part clear :+1:

schaefi (approved):

PR #2850, opened 2025-07-07
Add rd.kiwi.install.devicepersistency

schaefi:
Allow to specify which type of persistent device name should be used to build up the list of installation disk devices. For example rd.kiwi.install.devicepersistency=by-path would use the by-path representations for the available disk devices. The default (by-id) stays untouched. In case an invalid or not present device representation is selected, kiwi falls back to the non persistent unix node names.


dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
```diff
@@ -39,9 +84,13 @@ function get_disk_list {
         disk_id=${kiwi_devicepersistency}
```
davidcassany, line 84: I see this disk_id is also defined by `kiwi_install_devicepersistency` overwriting this value. I could not understand at a glance where this value is coming from and if this is actually interacting somehow with the the `kiwi_install_devicepersistenty`. I have the feeling the two respond to two different execution paths, so no issue, but to be honest, this is blurry in my head just right now.

davidcassany (approved):
LGTM

I am just a bit lost on how `disk_id` variable is defined, I see gets defined by two different inputs and that `rd.kiwi.install.devicepersistency` flag has priority over the other setup, which is not clear for me where it is coming from. I don't see any issue with it, just a behavior I could not double check.

dracut/modules.d/90kiwi-dump/kiwi-dump-image.sh
schaefi: your observations are correct. In the dracut code any assignment that says `variable=${kiwi_something}` originates from a value that can be set in the kiwi image description. In this case it would be the value of the attribute 

```
<type ... devicepersistency="..."/>
```

The new boot option `rd.kiwi.install.devicepersistency` allows to overwrite this value. Boot options by definition starts with `rd.`. I tried to consistently apply this concept across all the initrd code. 

So the implementation from here has:

```
1 disk_id="by-id"
# our default to begin with

2 disk_id="${kiwi_devicepersistency}"
# the value from the kiwi image description if specified
```

and now  new

```
3 disk_id="$(getarg rd.kiwi.install.devicepersistency=)"
# the value from the kernel cmdline if specified
```

schaefi: @davidcassany I guess that clarifies your question and still keeps the approve statement right ? :)

PR #2851, opened 2025-07-09
Fixed check for unallocated space on disk

schaefi:
So far the check for unallocated space was only working for GPT and there it was also not really stable. The check was based on verifying if the backup GPT table is really at the end of the disk. Depending on which tool was used to dump the image on the target this "mistake" often got corrected by the tools that dumped the image. In this case the check no longer worked. This commit improves the check by another test which looks for the real free bytes on disk compared to the current partition geometry.

### DIFF
--- a/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
+++ b/dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
@@ -328,7 +328,9 @@ function relocate_gpt_at_end_of_disk {
 
 function disk_has_unallocated_space {
     local disk_device=$1
+    local free_mbytes
     local pt_table_type
+    local flag="GPT PMBR size mismatch"
     pt_table_type=$(get_partition_table_type "${disk_device}")
     if [ "${pt_table_type}" = "gpt" ];then
         # GPT disks store a backup table at the end of the disk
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

dracut/modules.d/99kiwi-lib/kiwi-partitions-lib.sh
```diff
@@ -337,13 +339,20 @@ function disk_has_unallocated_space {
         # checked and used to detect that there is space
         # unallocated due to a geometry change of the underlying
         # block device layer
-        sgdisk --verify "${disk_device}" 2>&1 | grep -q "end of the disk"
-    else
-        # There is currently no method we could come up with
-        # to detect a geometry change for non GPT based disks.
-        # Thus we assume it's not fully allocated and allow
-        # for resize
+        if sfdisk --list-free "${disk_device}" 2>&1 | grep -q "${flag}"; then
```
uecasm, line 342: Reposting here in case you didn't see my comment in the forum.  I don't think this will behave as you expect, since this warning does not disappear after successful resize by kiwi-repart.

schaefi: @uecasm I also think I did not well thought through all cases. In my test that resizes to the full capacity, sfdisk did no longer show any warning and just displayed 

```
Unpartitioned space /dev/sda: 0 B, 0 bytes, 0 sectors
```

But I found other issues and more changes will be needed

PR #2853, opened 2025-07-14
Bind mount /dev/pts during image build time

schaefi:
Make sure to bind mount /dev/pts into the build environment. Tools which allocates a pseudo terminal won't work otherwise. Also apply a small refactor for the bind mounts which will create the target mount point directory if it does not exist

### DIFF
--- a/kiwi/system/root_bind.py
+++ b/kiwi/system/root_bind.py
@@ -17,6 +17,7 @@
 #
 import os
 import logging
+import pathlib
 import shutil
 import textwrap
 from typing import List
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
What requires a PTY?

Conan-Kudo:
We already tried this and had to revert it in https://github.com/OSInside/kiwi/pull/2692.

schaefi:
> We already tried this and had to revert it in #2692.

Oh darn :facepalm: I forgot that we already had this problem. Thanks for pointing it out

PR #2856, opened 2025-07-14
Cleanup integration tests config.sh script code

schaefi:
Add script code to shellcheck and fix all reported issues. Get rid of suseXX and baseXX methods as much as possible. Add set -ex for all script code. Do not allow any script code to fail.


schaefi:
In preparation to deprecate more of the shell code in `functions.sh` let's make sure our integration tests are clean first

build-tests/x86/tumbleweed/test-image-vagrant/config.sh
```diff
@@ -1,38 +1,7 @@
...
-#======================================
-# Greeting...
-#--------------------------------------
-echo "Configure image: [$kiwi_iname]..."
-
-#======================================
-# Setup baseproduct link
-#--------------------------------------
-suseSetupProduct
+declare kiwi_profiles=${kiwi_profiles}
```
Conan-Kudo, line 4: Where does this come from? Don't we need the sourcing lines for this?

Conan-Kudo (changes requested):

build-tests/arm/fedora/test-image-live/config.sh
```diff
@@ -1,38 +1,8 @@
...
-#======================================
-# Functions...
-#--------------------------------------
-test -f /.kconfig && . /.kconfig
-test -f /.profile && . /.profile
-
-#======================================
-# Greeting...
-#--------------------------------------
-echo "Configure image: [$kiwi_iname]..."
```
Conan-Kudo, line 27: Can we keep these? I think it's useful for us to at least verify the variables work.

build-tests/x86/tumbleweed/test-image-vagrant/config.sh
schaefi: nope because in kiwi executed scripts we automatically populate the profile environment into the execution environment. Therefore it's since a long time no longer required to source .profile. To make shellcheck happy a declare statement is required however

build-tests/arm/fedora/test-image-live/config.sh
schaefi: any variables that are really used in the context of the script got a `declare` statement which produces a log line including the content of the variable. For example

```
set -x
declare foo=${HOME}
```

would produce in the kiwi log

```
+ declare foo=/home/ms
```

I don't see any reason why greeting information for a variable that is never used in any other context of the script code should be be kept or has a value.

In addition the complete profile env is added in the kiwi log as well. Thus if you are only interested in what variables exists and their values, you will find in the kiwi log in a section that lists them all

For debugging issues of script executions only profile variables that are used for a significant action should be there. That greeting line is basically pointless :)

PR #2861, opened 2025-07-17
dracut kiwi-repart fixes

schaefi:
This change is many fold

**extend integration test for testing more of the resize code**

Use the lvm integration test as an example for a bit more complex resize testing

**apply proper udev locking**

Several commands during repart, resize and other actions require a proper lock to be set for udev such that other  events knows about the locked state of a device and do not mess with it until the command for which the lock persists has completed. This commit applies proper udev locks to all commands that requires it. In addition incorrect code that was expected to prevent such race conditions got dropped from the implementation.
This is related to https://bugzilla.suse.com/show_bug.cgi?id=1242987#c124

**relocate GPT at the end of disk using sfdisk**

Using sfdisk for relocation and verification makes this part more consistent. We also want to move away from gdisk. This is related to #2851

**Do not strictly require config.partids in repart**

The kiwi-repart implementation requires a metadata file named config.partids which holds information about  partition ids and more stored at the time the image was built. Depending on the complexity of the image and the resize request some of the information can be rebuilt in case the metadata file is missing. This commit adds  the rebuild of the minimum required information to run a standard resize and therefore allows the kiwi-repart dracut module to work also without config.partids to be present in the system

**Do not drop /config.partids**

The partition id metadata file is used in the kiwi-repart module. If a user wants to use the kiwi repart module permanently, this metadata file needs to stay in the system. Therefore it should not be automatically deleted by the cleanup. A disk.sh hook script can be used to force the deletion of the file though. This is related #2851

### DIFF
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -226,38 +226,12 @@ function check_repart_possible {
     return 0
 }
 
-function mask_fsck_root_service {
-    info "disable systemd-fsck-root.service"
-    systemctl mask systemd-fsck-root.service
-}
-
-function unmask_fsck_root_service {
-    info "enabling systemd-fsck-root.service"
-    systemctl unmask systemd-fsck-root.service
-}
-
 #======================================
 # Perform repart/resize operations
 #--------------------------------------
 PATH=/usr/sbin:/usr/bin:/sbin:/bin
 
 setup_debug
-
-# when repartitioning disks, the used tools can trigger re-reads of
-# the partition table, in turn triggering systemd-fsck-root.service
-# repeatedly via udev events, which finally can cause booting to fail with
-# * start request repeated too quickly for systemd-fsck-root.service
-# * Failed to start File System Check on /dev/disk/by-uuid...
-# * Dependency failed for /sysroot.
-# To avoid this, disable the root fsck (is finished at this point anyway
-# *and* the filesystem is brand new ;) by masking it.
-# "systemctl disable" does not work here, because it is event driven
-# More details: https://github.com/SUSE/kiwi/issues/1034
-
-# make sure we unmask the fsck service
-trap unmask_fsck_root_service EXIT
-
-mask_fsck_root_service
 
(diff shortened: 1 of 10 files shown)


schaefi:
@uecasm some followup changes to the topic we also discussed on the forum

I was not able to reproduce the issue you explained that after relocation of the backup header the warning still persists. Maybe you can give it a try too ?

Thanks

schaefi:
@Conan-Kudo please let us wait for feedback from @uecasm prior merge

schaefi:
There are more issues, we need to udevadm lock certain operations to prevent clashes with udev. Expect another commit soon

schaefi:
Did a Staging build with these changes and run some tests, for me things are working.

dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
```diff
@@ -35,6 +32,15 @@ function initialize {
     disk=$(lookup_disk_device_from_root)
     export disk
 
+    if ! test -f ${partition_ids}; then
```
davidcassany, line 35: I assume the partition_ids is missing for the modified integration test, so this logic is triggered there. However I am wondering which is the reason of not having the partition_ids file defined, couldn't be something pointing to a different issue/bug?

schaefi: I wanted to allow the repart code to also work in its core functions if there is no ```/config.partids``` in the system. If you look to the ```dracut/modules.d/90kiwi-repart/module-setup.sh``` we had the partids as mandatory file and if it was not present dracut skipped the inclusion of the module.

However, we don't have to be that strict because for the core functionality of the repart code we only need the id of the partition to resize. If this information is not present it can be taken as `the last partition on disk`. This information can be read from the current table and so we can rebuild at least the core part of the partids file and let the code work with that

There are some cases why ```/config.partids``` does not exist:
* some disk.sh script deleted it because users don't want this file in their image
* someone could just delete it
* kiwi prior this PR drops it as part of the cleanup (I consider this a bug and fixed it here)

So overall I just wanted to be a bit less strict on our repart code

Thoughts ?

PR #2865, opened 2025-07-25
Catch potential exceptions from pathlib.Path.mkdir

schaefi:
Creating a directory can fail, we should catch this error instead of ending up in a stack trace


schaefi:
Example for an exception would be if the given --target-dir points to a file or to a symlink pointing to a file or something that doesn't exist

PR #2866, opened 2025-07-29
Fix repartitioning with parted

Vogtinator:
parted does locking itself already. Wrapping it in udevadm lock results in a deadlock, breaking boot.


schaefi (approved):
Great catch, thanks :+1: 

I didn't know that parted does that

PR #2870, opened 2025-08-05
Skip repart when booting install/live iso

uecasm:
Fixes [error](https://groups.google.com/g/kiwi-images/c/SkpfZcRVTTk/m/hxvMTckqBgAJ).

* Skips doing anything in the kiwi-repart module if it's installed into the install or live iso images, since repartitioning makes no sense there and the format of "$root" is not supported by later code.

----

Note that this is sufficient to get the install iso booting again, but there is still some blkid errors being logged just prior to the install menu popup which suggests that another module might be unhappy, but I wasn't able to work out which.  I do have only these installed though:

```
        <package name="dracut-kiwi-live"/>
```

### DIFF
--- a/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/90kiwi-repart/kiwi-repart-disk.sh
@@ -234,6 +234,13 @@ PATH=/usr/sbin:/usr/bin:/sbin:/bin
 
 setup_debug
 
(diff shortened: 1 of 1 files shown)


schaefi:
Thanks for the patch. I'll continued the conversation on the mailing list and will come back here once we have clarity

schaefi (approved):
Looks good :+1:

PR #2871, opened 2025-08-05
Skip kiwi-repart module in install ISOs

schaefi:
In case the kiwi-repart module is explicitly requested in a dracut.conf file and the image is also configured to build an install ISO image this leads the install ISO to contain the kiwi-repart module as well which is unwanted. This commit explicitly omits the kiwi-repart when creating the initrd for the install image


schaefi:
needs verification from @uecasm

Conan-Kudo (approved):

schaefi:
@uecasm Added a commit to also fix the name conflict. Can you please check if that makes sense ? Thanks much

PR #2877, opened 2025-08-31
Follow the recommended dracut module ordering for out-of-tree modules

jozzsi:
In dracut release v108 or later the recommended ordering for out out of tree modules is 50-59 range. The following is a section from dracut documentation:

> Not using the 50-59 range for out of tree dracut modules will likely
> lead to unintended errors in the initramfs generation process as your
> dracut module will either run too early or too late in the generation process.
> You have been warned.


schaefi:
@jozzsi Thanks for catching this. Let's hope it has no other unintended side effects ;) 

I'll merge when tests are green and push this to the integration test builds

schaefi:
differential shellcheck complains about SC1091 (follow include file) which is not meaningful in the context of dracut module code. I'm going to ignore this warning. Unfortunately redhat-plumbers-in-action/differential-shellcheck@v3 cannot use the the sc_disable setting we use in in kiwi's self call of shellcheck

PR #2879, opened 2025-09-01
Add test for profiled overlays

schaefi:
kiwi supports overlay files per profile, but we didn't had a proper integration test for it. This commit adds one


schaefi:
See
* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-tbz

PR #2891, opened 2025-09-28
Add handling for stale issues and PRs

schaefi:
As discussed in the community meeting stale issues and PRs will be warned and stale labeled after 60 days. Without any further action the auto close happens after 30 days of the stale notification


Conan-Kudo:
I don't think we should auto-close issues. Many people consider them rude, and they really don't have a negative impact on us.

schaefi:
> I don't think we should auto-close issues. Many people consider them rude, and they really don't have a negative impact on us.

agreed and updated as such

Conan-Kudo:
Why do we want to close PRs? We have long-running PRs for good reason, and I'm not sure what we gain from auto-closing them other than to create a door slam on a contributor...

schaefi:
>Why do we want to close PRs? We have long-running PRs for good reason

We set a stale label after 60 days and would close a PR after another 30 days. To prevent this a sign of live (a comment or label change) is requested from the contributor. That is a 90 day period for some sign of live and if it does not come I believe the contributor has lost interest. In that case and only then the close doesn't seem to be too harsh

>We have long-running PRs for good reason

yes but I'm talking about orphan PRs. I can only speak for myself but when I create a PR to someones project I don't leave it untouched for 90 days

I intentionally added all the three of us to agree on a path forward. I believe you were not part of the last community meeting. So your voice was not heard, unfortunately. 

Let us agree to a procedure we all agree with

Thoughts ?

dcermak:
As long as this is not marking issues as stale and the PR submitter can dismiss the label, I'm fine with this

Conan-Kudo:
I'm fine with marking a PR as stale, I don't particularly like auto-closing things. If _we_ close something, then at least we're acknowledging it.

davidcassany:
I agree with @dcermak , I think the most important is that the stale label can be removed by the submitter and that the label is applied with a meaningful message, which is already the case.

Regarding autoclosing I do not have a strong opinion, I am fine with it. Given there some concerns around it I am wondering about another alternative. I don't think there won't be an unmanageable mount of abandoned PRs, so I think a potential idea could be relabeling it again as `abandoned` or something similar.  So a maintainer can easily filter by abandoned PRs (the ones that had 90 days without activity) and close them without having to think twice about it. So we are still having a final manual step giving room to scape the rules on weird or exceptional circumstances and also giving some sort of a human touch, not bot involved in the final close action.

Conan-Kudo:
I would prefer we didn't do auto-closing, and instead added it to our review at community meetings.

schaefi:
Ok, let's not auto close stuff

But make it clear in the message that orphan PRs will be closed by us as an action item in the community meeting

schaefi:
I'm seeking approval from you all prior merge :) 

@dcermak @davidcassany @Conan-Kudo 

Thanks

.github/workflows/ci-stale.yml
```diff
@@ -0,0 +1,16 @@
+name: 'Close stale issues and PRs'
```
dcermak:
```suggestion
name: 'Mark stale PRs with the stale label'
```

schaefi: oh yes good catch.

davidcassany (approved):
Thanks @schaefi , looks good :+1:

PR #2892, opened 2025-10-01
Support SOURCE_DATE_EPOCH for OCI containers

alexandrevicenzi:
If SOURCE_DATE_EPOCH is available, use it for the container creation time and history.


schaefi (approved):
Looks good to me. Let's wait for the tests to pass then I merge

PR #2894, opened 2025-10-17
Change appx util from required to recommended

schaefi:
Recently we added support for the new tarball-based WSL format. To my understanding this is the successor for the "old" format produced by the appx utility. Our current kiwi-systemdeps-containers-wsl package has appx as a hard requirement. I believe this can change into a recommended package


Conan-Kudo:
We can't until the Microsoft Store based WSL images go away.

schaefi:
Ah ok, makes sense

PR #2899, opened 2025-11-10
Add support for s390 UV-attestation

schaefi:
To run pvattest perform an attestation request binary as well as the corresponding attestation key and the secure execution header is required. This data should be created as part of the image build. In this commit we are adding support for creating the attestation data when building a secure execution enabled image

### DIFF
--- a/.github/workflows/ci-code-style.yml
+++ b/.github/workflows/ci-code-style.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci-units-types.yml
+++ b/.github/workflows/ci-units-types.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.11", "3.12", "3.13"]
 
     steps:
       - uses: actions/checkout@v3


schaefi:
Integration test available here

* https://build.opensuse.org/projects/Virtualization:Appliances:Images:Testing_s390:tumbleweed/packages/test-image-MicroOS:SUSE-Infra/repositories/images/binaries

Conan-Kudo:
What is UV-attestation?

schaefi:
@Conan-Kudo sorry I was not finished providing all data. So this is all about the following:
* https://www.ibm.com/docs/en/linux-on-systems?topic=commands-pvattest

schaefi:
I'm not sure how much insight you have into s390 and added you mostly for eventual side effects you might see with this change. I believe there are none.

I'm also adding an s390 pro to the review list :)

mfriesenegger:
@Conan-Kudo UV stands for the [ultravisor](https://www.ibm.com/docs/en/linux-on-systems?topic=execution-components#se_components__title__5) which is trusted firmware that uses memory-protection hardware to enforce memory protection.

kiwi/bootloader/config/zipl.py
```diff
@@ -135,6 +137,73 @@ def setup_loader(self, target: str) -> None:
...
+                        ]
+                    )
+                )
+
+                # generate attestation request binary and response key
+                attestation_request = 'attestation_request.bin'
+                attestation_response_key = 'attestation_response.key'
+                pvattest = [
+                    'chroot', root_dir, 'pvattest', 'create',
+                    '--no-verify', '--verbose',
```
mfriesenegger, line 169: IBM [documentation](https://www.ibm.com/docs/en/linux-on-systems?topic=commands-pvattest#pvattest_cmd__pvattest_create_cmd__title__1) warns of man-in-the-middle when not verifying the HKD.  I believe the HKD and certs are required to build an image. So the HKD will be available for lines 173-176.  It seems better to remove `--no-verify` and include `--cert` options.  @schaefi what are you thinking with using `--no-verify`?

schaefi: @mfriesenegger I had that before but this fails because the verification only succeeds on a trusted host. As such I was reaching out to IBM asking why pvattest has this restriction and that I believe it's not needed as also genprotimg supports offline verification. I got the response from Dirk Herrendörfer that there is the --no-verify option and even IBM themselves are using it that way

That's why I changed it accordingly and I also think it's safe because the way we build images does not allow for any man-in-the-middle attack as long as you consider the build service worker an isolated build env... which it is.

Thoughts ?

PR #2900, opened 2025-11-14
Fix repo type setup when set from cmdline

schaefi:
Using --add-repo/--set-repo options on the commandline allows to skip the repotype setting by passing in an empty value. That empty value however caused the type attribute in a repository section to be set to an empty string which is not allowed by the schema. Instead an empty type should lead to no type attribute set at all. This commit fixes it

### DIFF
--- a/kiwi/tasks/system_build.py
+++ b/kiwi/tasks/system_build.py
@@ -371,6 +371,13 @@ class SystemBuildTask(CliTask):
         parameters = self.eleventuple_token(tokens)
         signing_keys_index = 6
         repo_source_index = 0
+        repo_type_index = 1
+        if not parameters[repo_type_index]:
+            # make sure to pass a None value if an empty string
+            # is provided as repo type. This will cause no type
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
Why wouldn't we just have it fail? We don't support type-less repositories.

schaefi:
> Why wouldn't we just have it fail? We don't support type-less repositories.

yes we do, see for example: https://github.com/OSInside/kiwi/pull/2505

If no type is set we leave it up to the package manager what to do with the repo. At worst it fails when the package manager doesn't like it.

The code before this commit was surely broken because it added the repo with `type=""` which is not allowed by the schema saying

```
attribute type {
    "apk" | "apt-deb" | "apt-rpm" | "deb-dir" | "mirrors" | "rpm-md"
}
```

As such two options are correct:
* no type attribute at all
* a correct type attribute (and an empty value is not correct)

Therefore this fix here makes complete sense imho and fixing a bug

PR #2901, opened 2025-11-17
Fixed spec file requires

schaefi:
The package requirement for binutils was set to TW (>=1650) only but is also required for SLES16/Leap16 which is 1600 This commit fixes the condition to match with all required distributions and fixes bsc#1253637

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -354,7 +354,7 @@ Provides:       kiwi-image:vmx
 Requires:       kiwi-systemdeps-filesystems = %{version}-%{release}
 Requires:       kiwi-systemdeps-bootloaders = %{version}-%{release}
 Requires:       kiwi-systemdeps-iso-media = %{version}-%{release}
-%if 0%{?suse_version} >= 1650
+%if 0%{?suse_version} >= 1600
 Requires:       binutils
 Requires:       glibc-gconv-modules-extra
 %endif


package/python-kiwi-spec-template
```diff
@@ -141,21 +141,20 @@ Recommends:     dpkg
...
-%if 0%{?suse_version} >= 1650
+%if 0%{?suse_version} >= 1600
 Recommends:     dnf
 %endif
 %endif
 # package managers required by distro
-%if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version} >= 1650
+%if 0%{?fedora} || 0%{?rhel} >= 8 || 0%{?suse_version} >= 1600
 Provides:       kiwi-packagemanager:microdnf
-Requires:       microdnf
```
Conan-Kudo, line 151: You cannot delete this until we decide to drop the microdnf backend.

Conan-Kudo:
Also, microdnf is still shipped in RHEL.

package/python-kiwi-spec-template
schaefi: So microdnf as a package still exists ? I thought all this was "merged" to dnf ?

schaefi: Also see

```
commit 4fe1c918d9d97fe8c487ef8db06bf2f70f0d0048
Author: Marcus Schäfer <marcus.schaefer@gmail.com>
Date:   Fri Jun 16 18:27:07 2023 +0200

    Fix repository setup for microdnf
    
    An instance of the Repository class for microdnf has
    to use RepositoryDnf4

```

schaefi:
> Also, microdnf is still shipped in RHEL.

We are not calling it, we map all of it to dnf ... as I was told to do so

package/python-kiwi-spec-template
Conan-Kudo: Yes, the repository code is the same, the package manager code isn't.

schaefi: Ah because of PackageManagerMicroDnf, thanks

schaefi: but requiring microdnf on suse seems wrong, it doesn't exist there. I adapted without dropping the microdnf requirement for the others

Conan-Kudo: It does though? It's in Tumbleweed at least.

schaefi: > It's in Tumbleweed at least.

yes but I don't know how to differentiate TW from Leap16 from SLE16 properly. According to the docs `%if 0%{?suse_version} > 1600` matches TW and `%if 0%{?suse_version} == 1600 && 0%{?is_opensuse}` matches Leap16 but then `SLES 16 currently doesn't use the sle_version macro.` I got tired looking for something that doesn't cause problems and decided that most probably nobody on the suse side will use `microdnf` to build anything. They would rather go with `dnf5` 

Yes this is an assumption but I guess it's mostly correct. And even if someone wants to build on suse with microdnf, this user could also manually install the package instead of expecting some `kiwi-systemdeps-...` package that pulls it in.

I was thinking in simplifying the requirement to prevent at least one of this strange sle16 after effects.

Thoughts ?

schaefi: ok I believe I found the correct setting

Conan-Kudo: Well, this is why we used `1650`... It's higher than SLE 16 on purpose.

schaefi: but 1650 doesn't exist at all, see the docs. I changed that to "> 1600" which is TW, that should be ok

schaefi: see the table in  https://en.opensuse.org/openSUSE:Build_Service_cross_distribution_howto

Conan-Kudo: Well, it used to be set to 1550, then 1650, and now it's 1699. But 1650 is still higher than 1600.

https://build.opensuse.org/projects/openSUSE:Factory/prjconf

schaefi: I can't follow what you are after. I'm looking at the docs and I believe the condition ">= 1650" should match TW *only* So shouldn't we follow the docs and say "> 1600" ?

If you are completely against it I can revert that part. But when I look at the spec I'm searching for the conditions as documented and not for some arbitrary number > 1600

Thoughts ?

schaefi: ok silence means you don't like it :-)) I reverted this part and only left the relevant condition change

schaefi:
@Conan-Kudo could you have a look again ? Thanks much

PR #2905, opened 2025-11-19
Add support for encryped live images

schaefi:
So far setting the luks= attribute on a live image build had no impact to the generated ISO image. This commit adds the encryption capability also for live ISO images. The read-only part of the rootfs gets encrypted using the provided luks passphrase. An eventual persistent storage area gets encrypted at boot time if the rd.live.encrypt kernel cmdline parameters is passed. encryption/decryption requires to interactively set/provide passhphrase information at boot time. Please note due to the read-only restrictions of an ISO image there is no way to apply the standard re-encryption process as it is usually performed by kiwi encrypted systems. As such the specified luks passphrase in the kiwi image descriptions becomes sensitive information that needs to be protected

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-live/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-live/appliance.kiwi
@@ -11,6 +11,7 @@
         <specification>live iso test build</specification>
     </description>
     <profiles>
+        <profile name="Encrypted" description="Encrypted EFI/BIOS Live Boot"/>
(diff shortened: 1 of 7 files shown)


schaefi:
Please find an integration test build for this feature here:
* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-live

The `Encrypted` profile build

Conan-Kudo:
You should consider also adding this functionality to dmsquash-live too.

PR #2909, opened 2025-11-22
Improve grub console setup

### LINKED ISSUE #2886: Allow specifying multiple output consoles for grub
We'd like to be able to specify multiple output consoles for grub in our Kiwi builds, specifically console and serial.

Grub itself supports specifying multiple output terminals separated by spaces:
```
‘GRUB_TERMINAL_OUTPUT’

    Select the terminal output device. You may select multiple devices here, separated by spaces. 
```
(from https://www.gnu.org/software/grub/manual/grub/html_node/Simple-configuration.html)

In Kiwi, space separation has been "abused" to differentiate output and input consoles, so I guess the next best thing would be comma separated values:

```
<bootloader name="grub2" bls="true" console="console,serial console" timeout="5"/>
```

This could get kind of confusing, so it might be better to separate the consoles like this, but that would be a bigger change:
```
<bootloader name="grub2" bls="true" output_console="console serial" input_console="console" timeout="5"/>
```

schaefi:
The console attribute in the bootloader section was used to target both, output and input console setup using a space separator. However, there can be multiple consoles specified for output and input consoles and that turns the idea to handle both in one attribute into a mess. This commit drops the console attribute and offers two new attributes output_console and input_console which can define multiple consoles each using space as a separator. For example:

```xml
<bootloader ... output_console="console serial" input_console="console"/>
```

This change comes with a schema version bump to v8.4 An automatic XSL transformation template for the old console attribute into the new output_console and input_console attributes is also part of this commit.

This Fixes #2886

### DIFF
--- a/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
+++ b/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
@@ -1,9 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 
 <!-- The line below is required in order to use the multibuild OBS features -->
+
 <!-- OBS-Profiles: @BUILD_FLAVOR@ -->
 
-<image schemaversion="7.5" name="kiwi-test-image-live-disk">
+<image schemaversion="8.4" name="kiwi-test-image-live-disk">
     <description type="system">
         <author>Marcus Schaefer</author>
         <contact>ms@suse.com</contact>
@@ -16,7 +17,7 @@
     </profiles>
     <preferences>
         <version>2.0.0</version>
-        <packagemanager>dnf</packagemanager>
+        <packagemanager>dnf4</packagemanager>
         <bootsplash-theme>charge</bootsplash-theme>
         <bootloader-theme>breeze</bootloader-theme>
         <locale>en_US</locale>
@@ -25,12 +26,12 @@
         <rpm-check-signatures>false</rpm-check-signatures>
     </preferences>
     <preferences profiles="Live">
-        <type image="iso" flags="overlay" firmware="uefi" hybridpersistent_filesystem="ext4" hybridpersistent="true" kernelcmdline="console=ttyS0">
-            <bootloader name="grub2" console="serial" timeout="10"/>
+        <type image="iso" flags="overlay" firmware="uefi" hybridpersistent_filesystem="ext4" hybridpersistent="true" kernelcmdline="console=tty1 console=ttyS0">
+            <bootloader name="grub2" timeout="10" output_console="console serial" input_console="serial"/>
         </type>
     </preferences>
     <preferences profiles="Virtual">
-        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="uefi" format="qcow2" btrfs_root_is_subvolume="true" btrfs_set_default_volume="false" rootfs_label="fedora" bootpartition="false">
+        <type image="oem" filesystem="btrfs" kernelcmdline="console=tty1 console=ttyS0" firmware="uefi" format="qcow2" btrfs_root_is_subvolume="true" btrfs_set_default_volume="false" rootfs_label="fedora" bootpartition="false">
             <systemdisk name="fedora">
                 <volume name="@root=root"/>
                 <volume name="boot" parent="/"/>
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
So with this schema update, how is this hooked up? I don't see any wiring to use the new properties...

schaefi:
> So with this schema update, how is this hooked up? I don't see any wiring to use the new properties...

Not sure what you are specifically looking at. The XSL transformation automatically moves you into the new attribute situation. Some examples:

* console="gfxterm" turns into "output_console="gfxterm"
* console="gfxterm serial" turns into "output_console="gfxterm" input_console="serial"

This is for schema <= 8.3

If you want to make use of the use case described in the referenced issue you need to write an 8.4 compliant description (or let kiwi convert it and then adapt). With 8.4 you can then also write a console setup like this

```
output_console="console serial" input_console="console"
```

The schema allows to specify multiple console separated by a space for input and output individually as requested.

I haven't changed a runtime build to make use of the feature as all runtime integration tests uses the serial console and I did not yet see a good setup for a test. We certainly can add one. I'm open for suggestions

schaefi:
ah and maybe something you didn't know. In code meaning in the grub bootloader code in kiwi we have the separation for input and output consoles for a long time. As such you don't see a change in this implementation as it was already there. With the schema so far it was only possible to specify just one console name for output and one for input but not multiple ones. This is what changes now. The value from the new attributes output_console and input_console is taken 1:1 into the grub config. grub supports it in the same way as we allow it in the attributes

Conan-Kudo:
Ah, okay! So we already had the split and so this just propagates it up properly in the schema.

Conan-Kudo:
I think we could have a Fedora test image that has two output consoles and one input console defined. That's useful for various hypervisor environments anyway.

schaefi:
> I think we could have a Fedora test image that has two output consoles and one input console defined. That's useful for various hypervisor environments anyway.

I will adapt one of the existing tests to match this setup

schaefi:
ok done and also updated the docs. Can you review this one and maybe also the other open PR #2905 ? I'd like to zip a staging build and check all the integration tests for no regressions

Thanks

schaefi:
I will do some testing locally ...

schaefi:
> I will do some testing locally ...

test build worked for me now

doc/source/image_description/elements.rst
```diff
@@ -1119,24 +1119,17 @@ bls="true|false":
...
-
-  * Two values separated by a space are provided. In this case the
-    first value configures the output console and the second value
-    configures the input console. The `none` value can be used to skip
-    one or the other console setup. More than two space separated
-    values will be ignored.
-
+output_console="none|console|gfxterm|serial|vga_text|mda_text|morse|spkmodem":
+  Specifies the bootloader output console. The attribute is available for the
+  `grub` bootloader type. multiple console types can be provided separated
```
alexiri:
```suggestion
  `grub` bootloader type. Multiple console types can be provided separated
```

doc/source/image_description/elements.rst
```diff
@@ -1119,24 +1119,17 @@ bls="true|false":
...
+  by a space
+
+input_console="none|console|serial|at_keyboard|usb_keyboard":
+  Specifies the bootloader input console. The attribute is available for the
+  `grub` bootloader type. multiple console types can be provided separated
```
alexiri:
```suggestion
  `grub` bootloader type. Multiple console types can be provided separated
```

kiwi/xml_state.py
```diff
@@ -1197,14 +1197,13 @@ def get_build_type_bootloader_console(self) -> List[str]:
...
-            if combined_console:
-                console_out, *console_in = combined_console.split(' ')[:2]
-                console_in = console_out if not console_in else console_in[0]
-                result = [
-                    console_out if console_out != 'none' else '',
-                    console_in if console_in != 'none' else ''
-                ]
+            console_out = bootloader.get_output_console()
+            console_in = bootloader.get_input_console()
+            console_in = console_out if not console_in else console_in
```
alexiri:
```suggestion
            console_in = console_in if console_in else console_out
```

Easier to read, I think, but up to you.

Conan-Kudo:
Those changes make sense to make, yes.

Conan-Kudo (changes requested):
Please address @alexiri's comments.

kiwi/xml_state.py
schaefi: yeah :) depending on the weekday I'm reading it. I'll use your suggestion. Thanks

doc/source/image_description/elements.rst
schaefi: fixed, thanks

kiwi/xml_state.py
schaefi: done

schaefi:
@alexiri I believe I covered your change requests. Appreciate if you can take a look. @Conan-Kudo your approval provided, this can be merged. Thanks

doc/source/image_description/elements.rst
Conan-Kudo: It doesn't look like this fix got pushed? I don't see the sentence case fix.

schaefi: it got pushed, but there are two sentences that have the problem and I fixed only the one commented by Alex. Will complete this now

schaefi: done at all places now

PR #2913, opened 2025-12-02
Adding support for Custom CA Certificates during system build

rdoxenham:
This commit adds support for providing custom CA certificates during the build process. It allows a user to specify a path to a local directory, either by updating `kiwi.yml` with `ca_certificates:path`, or by passing `--ca-certificates=<directory>` via the command line.

The code looks for all `*.pem`, `*.crt`, and `*.crt` files in the specified directory and imports them immediately post bootstrap (where the required CA update tools are available), but before any further packages are retrieved, solving for situations where the chroot environment needs certificates, e.g. when there's a proxy server in the build environment.

Reference: https://github.com/OSInside/kiwi/pull/2721

### DIFF
--- a/doc/source/commands/system_build.rst
+++ b/doc/source/commands/system_build.rst
@@ -24,6 +24,8 @@ SYNOPSIS
        [--add-repo-credentials=<user:pass_or_filename>...]
        [--add-package=<name>...]
        [--add-bootstrap-package=<name>...]
+       [--ca-cert=<cert-file>...]
+       [--ca-target-distribution=<suse|rhel|debian|archlinux>]
        [--delete-package=<name>...]
        [--set-container-derived-from=<uri>]
        [--set-container-tag=<name>]
@@ -97,6 +99,24 @@ OPTIONS
   cache data dedicated to one build. In case of {kiwi}, the cache
   is shared between multiple image builds on that host for performance
   reasons.
+
+--ca-cert=<cert-file>
+
+  Add a cert-file to the directory storing additional local CA certificates.
+  The import will occur immediately after the bootstrap process, where
+  the required CA update tooling is expected to be installed. This
+  option is useful for situations where certificates are not packaged,
+  or the certificates are required during the build process, e.g. due
+  to proxy servers in the build environment that need certificates
+  in chroot. The option can be specified multiple times.
+
+--ca-target-distribution=<suse|rhel|debian|archlinux>
+
+  Specify target distribution for the import of certificates
+  via the --ca-cert options(s) and/or the provided <certificates>
+  from the image description. The selected distribution is used
+  in KIWI to map the distribution specific CA storage path and
+  update tool for the import process.
 
 --delete-package=<name>
 
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -120,6 +120,41 @@ The following optional sub sections can be inserted below the description tag:
 license
   Specifies the license name which applies to this image description.
 
+.. _sec.certificates:
+
+<certificates>
+--------------
+
+Add a cert-file to the directory storing additional local CA certificates.
+The import will occur immediately after the bootstrap process, where
+the required CA update tooling is expected to be installed. This
+setting is useful for situations where certificates are not packaged,
+or the certificates are required during the build process, e.g. due
+to proxy servers in the build environment that need certificates
+in chroot. The required `target_distribution` attribute must be set
+to allow kiwi a correct matching for the CA store path and the update
+tool with regards to the image target distribution. The following
+settings apply:
+
++--------------+-------------------------------------------+------------------------+
+| Distributor  | CA Store                                  | Update Tool            |
++==============+===========================================+========================+
+| SUSE         | /etc/pki/trust/anchors                    | update-ca-certificates |
++--------------+-------------------------------------------+------------------------+
+| Red Hat      | /etc/pki/ca-trust/source/anchors          | update-ca-certificates |
++--------------+-------------------------------------------+------------------------+
+| Debian Based | /usr/local/share/ca-certificates          | update-ca-certificates |
++--------------+-------------------------------------------+------------------------+
+| Arch Linux   | /etc/ca-certificates/trust-source/anchors | update-ca-trust        |
++--------------+-------------------------------------------+------------------------+
+
+.. code:: xml
+
+   <certificates target_distribution="suse|rhel|debian|archlinux">
+     <certificate name="/some/ca/filename1"/>
+     <certificate name="/some/ca/filename2"/>
+   </certificates>
+
 .. _sec.preferences:
 
 <preferences>
@@ -170,7 +205,7 @@ table shows which package manager is connected to which distributor:
 +==============+=================+
 | SUSE         | zypper          |
 +--------------+-----------------+
-| RedHat       | dnf4 / dnf5     |
+| Red Hat      | dnf4 / dnf5     |
 +--------------+-----------------+
 | Debian Based | apt             |
 +--------------+-----------------+ 
@@ -1706,7 +1741,7 @@ The namedCollection element is used to install a number of packages
 grouped together under a name. This is a feature of the individual
 distribution and used in the implementation of the {kiwi} package
 manager backend. At the moment collections are only supported for
-SUSE and RedHat based distributions. The optional `patternType` attribute
+SUSE and Red Hat based distributions. The optional `patternType` attribute
 is used to control the behavior of the dependency resolution of
 the package collection. `onlyRequired` installs only the collection
 and its required packages. `plusRecommended` installs the collection,
@@ -1724,9 +1759,9 @@ any of its required packages and any recommended packages.
    `$ zypper search patterns`. By convention all packages that starts
    with the name "patterns-" are representing a pattern package.
 
-.. note:: Collections on RedHat
+.. note:: Collections on Red Hat
 
-   On RedHat based distributions collections are called `groups` and are
+   On Red Hat based distributions collections are called `groups` and are
    extra metadata. To get the names of these groups type the following
    command: `$ dnf group list -v`. Please note that since {kiwi} v9.23.39,
    group IDs are allowed only, e.g.: 
--- a/kiwi/system/prepare.py
+++ b/kiwi/system/prepare.py
@@ -36,6 +36,9 @@ from kiwi.package_manager.base import PackageManagerBase
 from kiwi.command_process import CommandProcess
 from kiwi.system.uri import Uri
 from kiwi.archive.tar import ArchiveTar
+from kiwi.utils.sync import DataSync
+from kiwi.path import Path
+from kiwi.defaults import Defaults
 
 from kiwi.exceptions import (
     KiwiBootStrapPhaseFailed,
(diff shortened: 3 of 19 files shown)


schaefi:
Hey @rdoxenham this is great work thank you very much :+1: 

I hope you don't mind I pushed a commit to fix the tests

Other than that there is a bit of "metadata" to complete the change. Can you please do the following

* There is a kiwi.yml file which contains all possible options as comments. Can you please add your new runtime config setup there too ?
* Your patch adds a new commandline option. Can you please add documentation for it to the man pages in doc/source/commands

Would be great thanks in advance

Conan-Kudo:
This needs a stanza in the XML too.

schaefi:
@rdoxenham ah sorry I saw you already added the docs topics... nevermind forget what I said :)

schaefi:
> This needs a stanza in the XML too.

I agree, it could also be added in a followup PR though

rdoxenham:
> > This needs a stanza in the XML too.
> 
> I agree, it could also be added in a followup PR though

I actually have some questions about this from #2721... do you expect for it to be something like:

```
<certificates>
    <certificate name="/path/to/certificate1.pem"/>
    <certificate name="/path/to/certificate2.pem"/>
</certificates>
```

Otherwise, I'm curious as to what would be placed into each entry other than a path?

schaefi:
> I actually have some questions about this from #2721... do you expect for it to be something like:
> 
> ```
> <certificates>
>     <certificate name="/path/to/certificate1.pem"/>
>     <certificate name="/path/to/certificate2.pem"/>
> </certificates>
> ```
> 
> Otherwise, I'm curious as to what would be placed into each entry other than a path?

Yes it would be good to add this new `<certificates>` section. As you implemented in a way that you take all certs from a path I'm fine if you add something like this

```xml
<certificates>
    <certificate path="/some/path"/>
</certificates>
```

We should allow for more than one entry though. In the commandline implementation we allow `--ca-certificates` only once. you can consider to allow this option multiple times. This would then return a list and you would be compatible with the XML stanza

kiwi/system/prepare.py
```diff
@@ -560,6 +562,71 @@ def clean_package_manager_leftovers(self) -> None:
...
+            package installation, yet before the system packages are
+            installed, therefore it makes sense to exit with a bootstrap
+            failure if we catch any exception here.
+        """
+        if not ca_certs_path:
+            return
+
+        if not os.path.isdir(ca_certs_path):
+            log.warning(
+                'Custom CA certificates path %s not found, skipping',
```
schaefi: can we use `f` strings please

kiwi/system/prepare.py
```diff
@@ -560,6 +562,71 @@ def clean_package_manager_leftovers(self) -> None:
...
+
+        try:
+            cert_found = False
+            cert_extensions = ['.pem', '.crt', '.cer']
+            for ext in cert_extensions:
+                for cert_file in glob.glob(os.path.join(ca_certs_path, f'*{ext}')):
+                    if os.path.isfile(cert_file):
+                        if not cert_found:
+                            Path.create(ca_chroot_path)
+                        log.info('--> Installing certificate: %s', os.path.basename(cert_file))
```
schaefi: ditto

kiwi/system/prepare.py
```diff
@@ -560,6 +562,71 @@ def clean_package_manager_leftovers(self) -> None:
...
+                        DataSync(cert_file, ca_chroot_path).sync_data()
+                        cert_found = True
+
+            if cert_found:
+                log.info('--> Updating CA certificate store with: %s', ca_update_info['tool'])
```
schaefi: ditto

rdoxenham:
> Yes it would be good to add this new `<certificates>` section. As you implemented in a way that you take all certs from a path I'm fine if you add something like this
> 
> ```
> <certificates>
>     <certificate path="/some/path"/>
> </certificates>
> ```
> 
> We should allow for more than one entry though. In the commandline implementation we allow `--ca-certificates` only once. you can consider to allow this option multiple times. This would then return a list and you would be compatible with the XML stanza

Are you ok with this being a follow-up change?

kiwi/system/prepare.py
rdoxenham: Yep, no worries, forgive the oversight. Fixed.

Conan-Kudo:
I would prefer it to be part of this change.

rdoxenham:
> I would prefer it to be part of this change.

Understood. Can you and @schaefi help me with expected behaviour, please? We're introducing three different ways of specifying additional CA certificates to import:

1. Via `--ca-certificates=/path` entries in the CLI
2. Via `ca_certificates:path[]` in `kiwi.yml` runtime configuration
3. Via `<certificate path=".."/>` elements in the XML state

In the current PR, (1) is favoured over (2). How do you envisage this happening with (3)? The way I'm thinking about it is still one or the other, in that if either (1) or (3) are specified, they override (2), yet if both (1) and (3) are specified then they're combined into a single list. Written more easily:

* Entries only in `kiwi.yml` --> Only these are used
* Entries in `kiwi.yml` *and* either `--ca-certificates` *or* `<certificate path=".."/>` --> `kiwi.yml` is ignored
* As above, but if `--ca-certificates` *and* `<certificate path=".."/>` --> one longer list is used, `kiwi.yml` is still ignored

Thoughts? Is the original plan of ignoring `kiwi.yml` even a good idea, or should it *always* be considered, regardless of others being specified?

Conan-Kudo:
None of them should be ignored. It's combinatorial.

schaefi:
@rdoxenham Thinking more about it I believe the implementation to read certificates from the the kiwi runtime config might be an uncommon place. The `kiwi.yml` runtime config actually only contains settings that affects kiwi as a builder. For example compression options or how to bundle results, etc etc. As of today there are no settings in the runtime config that affects the image to build. Therefore you could also think of dropping the reading of certificates from the runtime config and stick to only two places:

* commandline
* image description

The two source of information should be combined.

We have similar implementation already in kiwi for example for the handling of the repositories. You can specify them on the commandline and also in the image description. From an implementation perspective the way how we approach this is as follows:

1. Create the needed modifications to the schema file `kiwi/schema/kiwi.rnc`
2. make sure to call `poetry shell` now which drops you into an env such that the following works
3. When you are done with the schema changes call `make kiwi/schema/kiwi.rng` This call auto generates the schema and the data structures such that they can be consumed by python. (In a future project I plan to add some AI here to make this even easier by just expressing a high level structure change and this performs the needed rnc adaptions :)) But for today you still need to work with the kiwi.rnc file.
4. When this is done you have access to a new data structure API to read the certificates (get_certificates() , get_certificate(), get_path()) Use this API to read the data as you like to work with it in the kiwi code base. I usually create a method in `kiwi/xml_state.py` to provide a statefull method for this data.
5. When you are done with this part you can modify the way how you approach the commandline implementation. The code that handles `--ca-certificates` should now also use the generated API and transfer the commandline data into the internal data structure by using (set_certificates(), set_certificate(), set_path()). You can use the existing implementation for the repos to get a better understanding if you need to

When doing it that way you also automatically set the precedence for the data which is that the value from the command line may overwrite the data read from the image description

I hope this helps to make you started.

Thanks much

schaefi:
@rdoxenham I hope you don't mind, I have added new commits that implements the suggested change from my last comment. Would be great if you can review the code changes and share your thoughts about it.

With the latest changes we can now specify an optional `<certificates>` section as discussed. The command-line option I renamed to `ca-cert-path` and allow it to be specified multiple times. The command-line option handling operates as follows:
* no `<certificates>` section present, the first `ca-cert-path` option would create one using the image description data structure. Any subsequent option would add to it
* existing  `<certificates>` section,  `ca-cert-path` options are added

Thanks

schaefi:
I we agree this is good to go, feel free to squash commits into one

doc/source/image_description/elements.rst
```diff
@@ -120,6 +120,26 @@ The following optional sub sections can be inserted below the description tag:
...
+or the certificates are required during the build process, e.g. due to
+proxy servers in the build environment that need certificates in chroot.
+Supported certificate extensions are .pem, .crt, and .cer. All other
+file types in the specified directory will be ignored.
+
+.. code:: xml
+
+   <certificates>
+     <certificate path="/some/path"/>
+   </certificates>
```
Conan-Kudo: Wouldn't it make sense to list files instead of directories? I don't like the idea of just wholesale importing random files.

Conan-Kudo (changes requested):

doc/source/image_description/elements.rst
schaefi: It's not importing random files. @rdoxenham has implemented it in a way that the files from the directory are checked against their prefix to match as a certificate file. I'd like to get feedback from Rhys first to see if that directory import is based on some design pattern.

schaefi: If there is no strong design pattern involved here I agree with you that specifying single files is better and more specific

Conan-Kudo: The problem is that even if there was, that pattern is localized to Rhys' setup. As this would be available to everyone, that is a risky proposition.

Plus, usually people import cert bundle files or specialty certs (like a local intranet certificate).

schaefi: ok I believe this is enough risk analysis to add a commit that moves this implementation to file based cert setup. But now let's wait for @rdoxenham as I have touched his code a lot :)

Thanks

rdoxenham:
Hey @schaefi! Happy New Year :) my apologies for the slow reply to this. Thanks so much for taking the action to follow-up on this; I had planned on getting to it at some point in the first few weeks of 2026, but I'm very glad to see that you've made the progress, so I certainly do not mind you doing this 👍 

As @Conan-Kudo suggests, the pattern that I was originally targeting was to more easily support our containerised Kiwi workflow; in this instance, we'd expect the user to simply bind mount a directory containing their certificates to the container, and we'd iterate over all certificate files found (I specifically chose crt/pem/cer files for this). However, I can see that this may introduce some risks.

I think the path proposed here with *files* rather than *directories* can also work; it means a little more effort required for the containerised workflow, but I also see this as a fairly rare use-case, so I don't think it should become a blocker as far as I'm concerned.

Thanks again!

schaefi:
Hi @rdoxenham Happy new year to you too :)

Thanks much for your feedback and understanding.

> I think the path proposed here with files rather than directories can also work; it means a little more effort required for the containerised workflow, but I also see this as a fairly rare use-case, so I don't think it should become a blocker as far as I'm concerned.

That's great :+1:  As we have to look at this as a generic feature, we need to make sure it gets accepted from a community perspective. According to your feedback I'm going to rebase the branch such that this PR can be reviewed and merged.

After merge I will provide a Staging kiwi build which you can use for your container based workflow. I guess with the changes done on top of your initial implementation you probably need some glue code that reads the directories with the certs in question and provides a list of files to kiwi either via the commandline or added to the image description. Please note that kiwi also supports `includes` which would allow you to create only the `<certificates>` section as an XML snippet that can be
included with an include statement in the main image description. On the Fedora side we make use of this feature a lot. OBS does not support it though as they applied their own kiwi image description reading which is not using the kiwi API and is also not aligned with upstream kiwi development.

doc/source/image_description/elements.rst

```xml
<certificates>
    <certificate name="/some/ca/file1"/>
    <certificate name="/some/ca/file2"/>
</certificates>
```

schaefi (approved):
approving my own changes 😄

schaefi:
we probably want an integration test to test this

- [x] integration test

kiwi/xml_state.py
```diff
@@ -2976,6 +3013,44 @@ def get_system_archives_target_dirs(self) -> Dict:
...
+            return {
+                'tool': 'update-ca-certificates',
+                'destination_path': '/usr/local/share/ca-certificates/'
+            }
+        elif package_manager == 'pacman':
+            # ArchLinux family
+            return {
+                'tool': 'update-ca-trust',
+                'destination_path': '/etc/ca-certificates/trust-source/anchors/'
+            }
```
Conan-Kudo: This is pretty much not going to work. For example, what happens when DNF is used on openSUSE (like I do)?

You're going to have to check the os-release in the rootfs to identify the family. Maybe you could get away with some kind of priority order to identify paths, if we can assume only one of these paths exist in a rootfs at a time.

But we definitely cannot use the package manager as the key.

rdoxenham: Good point. However, checking the `/etc/os-release` file is also not really an option for the use-case I'm concerned about, which is ensuring that the chroot environment *already* has the certificate before downloading the required packages (which I suspect would provide `/etc/os-release`). For the use-case I'm concerned about, the certificates need to be in place to accommodate proxies in the build environment. Let me have a think about this one.

schaefi: I agree and I vote for dropping the distribution detection completely and add this information as explicit setting to the image description. Something like this

```xml
<certificates target="/etc/pki/trust/anchors">
    <certificate name="foo"/>
</certificates>
```

Thoughts ?

schaefi:
note for myself. The changes here needs a porting effort of the new `--ca-cert` option to #2751

kiwi/xml_state.py
rdoxenham: If we go down this path we'd also need to consider the update tool, as they're different across distributions, too, and we'd need to update the CLI args to add a `--ca-tool` and `--ca-location` (or similar). The more I think about it, that may be the cleanest way of achieving it.

schaefi: We could also consider a mapping table like this

```xml
<certificates target_distribution="suse">
    <certificate name="foo"/>
</certificates>
```

Then we maintain a mapping table to the target path and the tool. Going forward we can adapt the mapping table when new distributions comes up or special handling is needed for example for `sles` or `rhel` ... etc

I did a similar implementation to handle the different paths and styles regarding EFI secure boot between the distros. This worked well and is not much of a maintenance effort

consequently we also add `--ca-target-distribution` or similar

Thoughts ?

rdoxenham: I think that's a good outcome @schaefi! Then, the choice of package manager is irrelevant to this. Thanks!

schaefi: ok I'll change it

Conan-Kudo: That makes sense to me. We'll want to make sure this is described and it's easy to add more as it comes up.

schaefi: @rdoxenham @Conan-Kudo I changed the way how we setup the CA store path and tool as discussed.

doc/source/image_description/elements.rst
```diff
@@ -120,6 +120,41 @@ The following optional sub sections can be inserted below the description tag:
...
+to allow kiwi a correct matching for the CA store path and the update
+tool with regards to the image target distribution. The following
+settings apply:
+
++--------------+-------------------------------------------+------------------------+
+| Distributor  | CA Store                                  | Update Tool            |
++==============+===========================================+========================+
+| SUSE         | /etc/pki/trust/anchors                    | update-ca-certificates |
++--------------+-------------------------------------------+------------------------+
+| RedHat       | /etc/pki/ca-trust/source/anchors          | update-ca-certificates |
```
rdoxenham: Old habits die hard...

```suggestion
| Red Hat      | /etc/pki/ca-trust/source/anchors          | update-ca-certificates |
```

schaefi: LOL :) I'm going to change it

doc/source/commands/system_build.rst
```diff
@@ -24,6 +24,8 @@ SYNOPSIS
        [--add-repo-credentials=<user:pass_or_filename>...]
        [--add-package=<name>...]
        [--add-bootstrap-package=<name>...]
+       [--ca-cert=<cert-file>...]
+       [--ca-target-distribution=<suse|rhel|debian|archlinux>]
```
Conan-Kudo, line 28: This should have been `redhat` instead of `rhel`.

PR #2914, opened 2025-12-05
Run disk.sh script hook with / being a mountpoint

schaefi:
Add optional root_is_mountpoint parameter to the _call_script method. When set the provided root path is turned into a bind mount such that it appears as / being a mountpoint in the chrooted call. Apply the argument to the call_disk_script method such that is becomes effective for the disk.sh script hook.

### DIFF
--- a/kiwi/volume_manager/btrfs.py
+++ b/kiwi/volume_manager/btrfs.py
@@ -94,6 +94,7 @@ class VolumeManagerBtrfs(VolumeManagerBase):
 
         self.subvol_mount_list = []
         self.toplevel_mount = None
+        self.root_volume_mount = None
         self.snapshots_root_mount = None
 
     def setup(self, name=None):
@@ -169,6 +170,14 @@ class VolumeManagerBtrfs(VolumeManagerBase):
             )
         elif self.root_volume_name != '/':
             self._set_default_volume(self.root_volume_name)
+            # make sure the root volume appears as proper (/) in the chroot
+            root_dir = os.path.normpath(
+                os.sep.join([self.mountpoint, self.get_root_volume_name()])
+            )
+            self.root_volume_mount = MountManager(
+                device=root_dir, mountpoint=root_dir
+            )
+            self.root_volume_mount.bind_mount()
 
     def get_root_volume_name(self) -> str:
         """
@@ -390,6 +399,9 @@ class VolumeManagerBtrfs(VolumeManagerBase):
         if self.snapshots_root_mount and self.snapshots_root_mount.is_mounted():
             self.snapshots_root_mount.umount()
 
+        if self.root_volume_mount:
+            self.root_volume_mount.umount()
+
         if self.toplevel_mount.is_mounted():
             self.toplevel_mount.umount()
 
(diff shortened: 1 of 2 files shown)


kiwi/system/setup.py
```diff
@@ -1278,38 +1280,50 @@ def _import_custom_scripts(self):
...
-            caller_environment.update(profile.get_settings())
-            config_script = Command.call(command, caller_environment)
-            process = CommandProcess(
-                command=config_script, log_topic='Calling ' + name + ' script'
-            )
-            result = process.poll_and_watch()
-            if result.returncode != 0:
-                raise KiwiScriptFailed(
-                    '{0} failed: {1}'.format(name, result.stderr)
+        with ExitStack() as stack:
```
dcermak, line 1289: Shouldn't we first check for the script and then setup bind mounts? In the current form, we will always bind mount root even if disk.sh is not there

kiwi/system/setup.py
```diff
@@ -1278,38 +1280,50 @@ def _import_custom_scripts(self):
...
+                caller_environment.update(profile.get_settings())
+                config_script = Command.call(command, caller_environment)
+                process = CommandProcess(
+                    command=config_script,
+                    log_topic='Calling ' + name + ' script'
+                )
+                result = process.poll_and_watch()
+                if result.returncode != 0:
+                    raise KiwiScriptFailed(
+                        '{0} failed: {1}'.format(name, result.stderr)
```
dcermak: f-strings are nice :-)
```suggestion
                        f"{name} failed: {result.stderr}"
```

kiwi/system/setup.py
schaefi: definitely yes :facepalm: I will adapt it

schaefi: done

kiwi/system/setup.py
schaefi: yes this is a burden of the old days and we adapt any place when it pops up. Thanks for pointing it out... and fixed :)

Conan-Kudo:
My first instinct is that this should be a new stage rather than messing with the existing one. I need to think about this and review it more.

schaefi:
> My first instinct is that this should be a new stage rather than messing with the existing one. I need to think about this and review it more.

Actually the `disk.sh` script hook was added for exactly that "to run a script as close to the actual root system when booted" I was also thinking if we want to add this change in the existing hook invocation for `disk.sh` but it really fits into its purpose

> that this should be a new stage 

That I would object to, as I said we added `disk.sh` for this purpose. As such not offering `/` as a mountpoint in this execution env could also be seen as a bug ;)

Conan-Kudo (approved):

Conan-Kudo:
I'm good with this, but let's get @dcermak's signoff before I merge it.

dcermak:
Neal Gompa (ニール・ゴンパ) ***@***.***> writes:

> Conan-Kudo left a comment (OSInside/kiwi#2914)
>
> I'm good with this, but let's get @dcermak's signoff before I merge it.

The code looks fine, but I'd need the staging image to be able to test
this 😅

schaefi:
I will push a staging build

@dcermak I let you know when the build is done

schaefi:
@dcermak I found some after effects, fixed them here and also increased the integration tests to make use of a disk.sh script that prints the actual root mountpoint via findmnt for the main disk builds using standard fs, lvm and filesystem_with_volume_capabilities(btrfs)

Feel free to check your use case once the `staging` box has built again

PR #2917, opened 2025-12-08
Add custom_part_control attribute for explicit partition layout control

cwolfe007:
Add custom_part_control attribute to <type> element enabling explicit
  partition numbering and layout control via <partitions>. 

## Why?
This is for allowing customers who have non standard layouts for their partitions to define their layouts as they need. For example Amazon linux has the root partition last in the order but it is numbered as 1, and the efi and boot paritions have the numbers 127 and 128. Other users may also have non standard partition requirements that having more refined control would be desired for legacy needs or other use cases. 

## Important points

- This is "opt-in" so if users do not have to use this feature if they do not want to
- Uses the existing custom partitions logic
- Maintains backwards compatibility

 

 ## Changes proposed in this pull request:

  - Add custom_part_control="true" attribute on <type> to enable explicit
  partition layout control
  - Add partition_number attribute on <partition> for explicit partition
  numbering (supports non-sequential numbers like 1, 127, 128)
  - Ordering of paritions reads top to bottom, allows users to intuit parition layout 
  - Add boot_flag attribute on <partition> to mark bootable partition
  - Allow reserved partition names (root, boot, efi) in <partition> when
  custom_part_control="true"
  - Block auto-partition attributes (bootpartition, bootpartsize,
  efipartsize, spare_part) when custom_part_control="true" with descriptive
  error messages
  - Update GPT and MSDOS partitioners to create partitions in
  partition_number order
  - Update bootloader config to correctly map custom partition devices
  - Set storage_map['system'] from custom partitions for bootloader
  installation
  - Add schema validation for new attributes and conditional rules
- Add TW and Fedora integration test image definition files

 ##  Test Plan

  - All 1844 unit tests pass
  - 100% code coverage maintained
  - GRUB2 image builds and boots in QEMU with new partition scheme
  - systemd-boot image builds and boots in QEMU with new partition scheme
  - Validated F37 image description still builds and is bootable in QEMU, seems to confirm backwards compatibility
    - https://github.com/OSInside/kiwi-descriptions/tree/main/fedora/x86_64/fedora-37.0 
  - Schema validation works for new attributes
  - Auto-partition attribute rejection works with clear error messages

Example file I tested with: 

```xml

<image schemaversion="8.4" name="Fedora-37.0_custom_partition">
    <description type="system">
        <author>Marcus Schaefer</author>
        <contact>ms@suse.de</contact>
        <specification>Fedora 37 Appliance</specification>
    </description>
    <profiles>
        <profile name="Virtual" description="Simple Disk image" import="true"/>
        <profile name="Live" description="Live image"/>
    </profiles>
    <preferences>
        <version>1.0.0</version>
        <packagemanager>dnf4</packagemanager>
        <bootsplash-theme>charge</bootsplash-theme>
        <bootloader-theme>breeze</bootloader-theme>
        <locale>en_US</locale>
        <keytable>us</keytable>
        <timezone>UTC</timezone>
        <rpm-check-signatures>false</rpm-check-signatures>
    </preferences>
    <preferences profiles="Live">
        <type image="iso" flags="dmsquash" firmware="uefi" hybridpersistent_filesystem="ext4" hybridpersistent="true"/>
    </preferences>
        <type image="oem" kernelcmdline="selinux=targeted console=ttyS0,115200" eficsm="false" firmware="uefi" custom_part_control="true">
            <oemconfig>
            </oemconfig>
            <bootloader name="grub2"/>
            <size unit="G">15</size>
            <partitions>
                <partition name="efi" size="100" filesystem="fat32" mountpoint="/boot/efi" partition_number="128" partition_type="t.efi" boot_flag="false" label="EFI"/>
                <partition name="boot" size="1024" filesystem="ext4" mountpoint="/boot" partition_number="127" partition_type="t.linux" boot_flag="true" label="Boot"/>
                <partition name="root" size="10000" filesystem="xfs" mountpoint="/" partition_number="1" partition_type="t.linux" label="Root"/>
            </partitions>
        </type>
    </preferences>
    <users>
        <user password="$1$wYJUgpM5$RXMMeASDc035eX.NbYWFl0" home="/root" name="root" groups="root"/>
    </users>
    <repository type="rpm-md" alias="local" sourcetype="baseurl" priority="3">
	    <source path="file:///root/local-rpm"/>
    </repository>
    <repository type="rpm-md" alias="Fedora37" sourcetype="metalink" priority="1">
        <source path="https://mirrors.fedoraproject.org/metalink?repo=fedora-37&amp;arch=x86_64"/>
    </repository>
    <repository type="rpm-md" alias="Fedora37-updates" sourcetype="metalink" priority="2">
        <source path="https://mirrors.fedoraproject.org/metalink?repo=updates-released-f37&amp;arch=x86_64"/>
    </repository>
    <packages type="image">
        <namedCollection name="network-server"/>
        <package name="grub2"/>
        <package name="grubby"/>
        <package name="kernel"/>
        <package name="plymouth-theme-charge"/>
        <package name="grub2-breeze-theme"/>
        <package name="selinux-policy-targeted"/>
        <package name="dhclient"/>
        <package name="NetworkManager"/>
        <package name="vim"/>
        <package name="tzdata"/>
        <package name="python-kiwi"/>
        <package name="git"/>
    </packages>
    <packages type="iso">
        <package name="grub2-efi-x64-cdboot"/>
        <package name="dracut-live"/>
    </packages>
    <packages type="bootstrap">
        <package name="glibc-all-langpacks"/>
        <package name="filesystem"/>
        <package name="basesystem"/>
        <package name="grub2-efi-x64"/>
        <package name="grub2-efi-x64-modules"/>
        <package name="shim-x64"/>
        <package name="syslinux"/>
        <package name="syslinux-efi64"/>
    </packages>
</image>

```

### DIFF
--- a/kiwi/bootloader/config/base.py
+++ b/kiwi/bootloader/config/base.py
@@ -63,6 +63,7 @@ class BootLoaderConfigBase(ABC):
         self.etc_kernel_mount = None
 
         self.root_filesystem_is_overlay = xml_state.build_type.get_overlayroot()
+        self.bootpartition = False
         self.post_init(custom_args)
 
     def __enter__(self):
@@ -522,8 +523,17 @@ class BootLoaderConfigBase(ABC):
             custom_root_mount_args += [f'subvol={root_volume_name}']
         self.root_mount.mount(options=custom_root_mount_args)
 
+        log.debug(f'_mount_system: root_mount.device={self.root_mount.device}, boot_mount.device={self.boot_mount.device}')
         if not self.root_mount.device == self.boot_mount.device:
+            log.debug(f'_mount_system: Mounting boot partition at {self.boot_mount.mountpoint}')
             self.boot_mount.mount()
+            # Update bootpartition flag when boot is on separate partition
+            # This handles custom_part_control where bootpartition attribute
+            # may not be set but a separate boot partition exists
+            log.debug(f'_mount_system: Setting bootpartition=True (was {self.bootpartition})')
+            self.bootpartition = True
+        else:
+            log.debug('_mount_system: Boot is on root partition (same device)')
 
         if volumes:
             for volume_path in Path.sort_by_hierarchy(
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -271,6 +271,66 @@ class DiskBuilder:
         if not self.boot_image.has_initrd_support():
             log.warning('Building without initrd support !')
 
+    def _has_custom_partition_control(self) -> bool:
+        """
+        Check if custom partition control is enabled
+
+        Custom partition control is enabled when the <type> element has
+        the custom_part_control="true" attribute.
+
+        :return: True if custom partition control is enabled
+
+        :rtype: bool
+        """
+        return self.xml_state.build_type.get_custom_part_control() is not None and \
+            str(self.xml_state.build_type.get_custom_part_control()).lower() == 'true'
+
+    def _validate_custom_partition_control_config(self) -> None:
+        """
+        Validate that auto-partition attributes are not mixed with custom control
+
+        KIWI supports two partition control modes:
+
+        1. Auto-partition mode (default): KIWI automatically creates partitions
+           based on attributes like bootpartition, bootpartsize, efipartsize, etc.
+
+        2. Custom partition control mode: When custom_part_control="true" is set,
+           users explicitly define all partitions via <partition> elements with
+           partition_number attributes.
+
+        These modes are mutually exclusive. When custom_part_control="true",
+        auto-partition attributes are not allowed as they would conflict with
+        the explicit partition definitions.
+
+        :raises KiwiDiskConfigError: If auto-partition attributes are found
+
+        :rtype: None
+        """
+        from kiwi.exceptions import KiwiDiskConfigError
+
+        build_type = self.xml_state.build_type
+
+        # Auto-partition attributes that conflict with custom_part_control
+        # These attributes instruct KIWI to automatically create partitions,
+        # which conflicts with explicit partition control via <partition> elements
+        auto_partition_attributes = {
+            'bootpartition': build_type.get_bootpartition(),
+            'bootpartsize': build_type.get_bootpartsize(),
+            'efipartsize': build_type.get_efipartsize(),
+            'spare_part': build_type.get_spare_part()
+        }
+
+        # Check for any auto-partition attributes that are set
+        for attr_name, attr_value in auto_partition_attributes.items():
+            if attr_value is not None:
+                raise KiwiDiskConfigError(
+                    f'custom_part_control="true" requires all partitions to be '
+                    f'defined via <partition> elements. The attribute "{attr_name}" '
+                    f'cannot be used with custom_part_control as it instructs KIWI '
+                    f'to auto-create partitions. Please define this partition '
+                    f'explicitly in <partitions> with a partition_number attribute.'
+                )
+
     def create(self) -> Result:
         """
         Build a bootable disk image and optional installation image
@@ -667,8 +727,14 @@ class DiskBuilder:
         )
 
         # create filesystems on boot partition(s) if any
-        self.storage_map['system_boot'], self.storage_map['system_efi'] = \
-            self._build_boot_filesystems(device_map)
+        # Skip when custom_part_control is enabled - filesystems already created per-partition
+        if not self._has_custom_partition_control():
+            self.storage_map['system_boot'], self.storage_map['system_efi'] = \
+                self._build_boot_filesystems(device_map)
+        else:
+            # Custom partitions already have filesystems created
+            self.storage_map['system_boot'] = None
+            self.storage_map['system_efi'] = None
 
         if self.volume_manager_name:
             volume_manager = VolumeManager.new(
@@ -681,7 +747,8 @@ class DiskBuilder:
             device_map = self._map_root_volume_manager(
                 device_map, volume_manager
             )
-        elif self.need_root_filesystem:
+        elif self.need_root_filesystem and not self._has_custom_partition_control():
+            # Skip when custom_part_control is enabled - root filesystem already created per-partition
             filesystem = FileSystem.new(
                 self.requested_filesystem, device_map['root'],
                 self.root_dir + '/',
@@ -689,6 +756,12 @@ class DiskBuilder:
             )
             stack.push(filesystem)
             self._map_root_filesystem(device_map, filesystem)
+        elif self._has_custom_partition_control():
+            # For custom partition control, get root filesystem from custom parts
+            # This is required for bootloader installation which checks 'if system:'
+            if 'root' in self.storage_map.get('system_custom_parts', {}):
+                self.storage_map['system'] = \
+                    self.storage_map['system_custom_parts']['root']
 
         return device_map
 
@@ -1168,6 +1241,31 @@ class DiskBuilder:
     ) -> Dict:
         disk.wipe()
         disksize_used_mbytes = 0
+
+        # Check if custom partition control is enabled
+        if self._has_custom_partition_control():
+            # Validate that auto-partition attributes are not mixed with custom control
+            self._validate_custom_partition_control_config()
+
+            # Use custom partition control
+            log.info('--> using custom partition control')
+            log.info(
+                '--> creating custom partition(s): {0}'.format(
+                    list(self.custom_partitions.keys())
+                )
+            )
+            disk.create_custom_partitions(
+                self.custom_partitions, custom_part_control=True
+            )
+            disk.map_partitions()
+
+            device_map = disk.get_device()
+            device_map['origin_root'] = \
+                device_map.get('readonly') or device_map['root']
+
+            return device_map
+
+        # Standard partition creation flow (backwards compatible)
         if self.firmware.get_legacy_bios_partition_size():
             log.info('--> creating EFI CSM(legacy bios) partition')
             partition_mbsize = self.firmware.get_legacy_bios_partition_size()
@@ -1918,8 +2016,12 @@ class DiskBuilder:
             readonly_device = device_map['readonly']
             boot_device = readonly_device
 
+        log.debug(f'_install_bootloader: device_map keys={list(device_map.keys())}')
         if 'boot' in device_map:
             boot_device = device_map['boot']
+            log.debug(f'_install_bootloader: Using separate boot device: {boot_device.get_device()}')
+        else:
+            log.debug(f'_install_bootloader: boot not in device_map, using root: {boot_device.get_device()}')
 
         custom_install_arguments = {
             'boot_image': boot_image,
--- a/kiwi/exceptions.py
+++ b/kiwi/exceptions.py
@@ -251,6 +251,14 @@ class KiwiDiskGeometryError(KiwiError):
     """
 
 
+class KiwiDiskConfigError(KiwiError):
+    """
+    Exception raised if the disk configuration is invalid,
+    such as when auto-partition control attributes are mixed
+    with custom_part_control mode.
+    """
+
+
 class KiwiDistributionNameError(KiwiError):
     """
     Exception raised if the distribution name could not be found.
--- a/kiwi/partitioner/base.py
+++ b/kiwi/partitioner/base.py
@@ -16,7 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 from typing import (
-    List, Dict, Union
+    List, Dict, Union, Optional
 )
 
 # project
@@ -73,7 +73,8 @@ class PartitionerBase:
         return self.partition_id
 
     def create(
-        self, name: str, mbsize: int, type_name: str, flags: List[str] = []
+        self, name: str, mbsize: int, type_name: str, flags: List[str] = [],
+        explicit_partition_id: Optional[int] = None
     ):
         """
         Create partition
@@ -84,6 +85,10 @@ class PartitionerBase:
         :param int mbsize: unused
         :param string type_name: unused
         :param list flags: unused
+        :param int explicit_partition_id:
+            If provided, use this exact partition ID instead of
+            auto-incrementing. Used for custom partition control.
+            When None, auto-increment is used (backwards compatible).
         """
         raise NotImplementedError
 
--- a/kiwi/partitioner/gpt.py
+++ b/kiwi/partitioner/gpt.py
@@ -16,7 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import logging
-from typing import List
+from typing import List, Optional
 
 # project
 from kiwi.command import Command
@@ -51,7 +51,8 @@ class PartitionerGpt(PartitionerBase):
         }
 
     def create(
-        self, name: str, mbsize: int, type_name: str, flags: List[str] = None
+        self, name: str, mbsize: int, type_name: str, flags: List[str] = None,
+        explicit_partition_id: Optional[int] = None
     ) -> None:
         """
         Create GPT partition
@@ -60,16 +61,27 @@ class PartitionerGpt(PartitionerBase):
         :param int mbsize: partition size
         :param string type_name: partition type
         :param list flags: additional flags
+        :param int explicit_partition_id:
+            If provided, use this exact partition ID instead of auto-incrementing
         """
-        self.partition_id += 1
+        # Remember if this is the first partition (partition_id starts at 0)
+        is_first_partition = (self.partition_id == 0)
+
+        if explicit_partition_id is not None:
+            self.partition_id = explicit_partition_id
+        else:
+            self.partition_id += 1
         if mbsize == 'all_free':
             partition_end = '0'
         else:
             partition_end = '+' + format(mbsize) + 'M'
-        if self.partition_id > 1 or not self.start_sector:
-            # A start  sector value of 0 specifies the default value
-            # defined in sgdisk
+
+        # Set start_sector=0 to let sgdisk find the next available sector
+        # For the first partition, honor any custom start_sector passed to constructor
+        # For subsequent partitions, always use 0 (sgdisk finds next available space)
+        if not is_first_partition or not self.start_sector:
             self.start_sector = 0
+
         Command.run(
             [
                 'sgdisk', '-n', ':'.join(
--- a/kiwi/partitioner/msdos.py
+++ b/kiwi/partitioner/msdos.py
@@ -16,7 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import logging
-from typing import List
+from typing import List, Optional
 
 # project
 from kiwi.utils.temporary import Temporary
@@ -56,7 +56,8 @@ class PartitionerMsDos(PartitionerBase):
         }
 
     def create(
-        self, name: str, mbsize: int, type_name: str, flags: List[str] = []
+        self, name: str, mbsize: int, type_name: str, flags: List[str] = [],
+        explicit_partition_id: Optional[int] = None
     ) -> None:
         """
         Create msdos partition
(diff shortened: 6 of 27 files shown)


kiwi/exceptions.py
```diff
@@ -251,6 +251,14 @@ class KiwiDiskGeometryError(KiwiError):
     """
 
 
+class KiwiDiskConfigError(KiwiError):
+    """
+    Exception raised if the disk configuration is invalid,
+    such as when legacy partition control attributes are mixed
```
schaefi: Can we rephrase `legacy partition control` to something like `classic` or `default` or `standard` ? I don't think the numbering of partitions without explicit values is a legacy mode :)

kiwi/partitioner/base.py
```diff
@@ -84,6 +85,10 @@ def create(
         :param int mbsize: unused
         :param string type_name: unused
         :param list flags: unused
+        :param int explicit_partition_id: If provided, use this exact partition ID
+                                         instead of auto-incrementing. Used for
+                                         custom partition control. When None,
+                                         auto-increment is used (backwards compatible).
```
schaefi: nitpick: Can we keep the 4 space indentation style ?

```
:param int explicit_partition_id:
    If provided, ...
```

Thanks

kiwi/partitioner/msdos.py
```diff
@@ -65,8 +66,26 @@ def create(
         :param int mbsize: partition size
         :param string type_name: partition type
         :param list flags: additional flags
+        :param int explicit_partition_id: If provided, use this exact partition ID
+                                         instead of auto-incrementing. Presence of
+                                         explicit_partition_id indicates custom_part_control
+                                         is enabled, which means user has explicit control
+                                         over partition layout. In this case, extended_layout
+                                         logic is bypassed.
```
schaefi: ditto: indentation

kiwi/storage/disk.py
```diff
@@ -150,31 +152,65 @@ def create_custom_partitions(
                map_name: ptable_entry_type
            }
 
-        :param dict table: partition table spec
+        :param dict table_entries: partition table spec
+        :param bool custom_part_control: If True, allow reserved partition names
+                                        and use explicit partition_number values
+                                        from partition_number attribute
```
schaefi: ditto: indentation

kiwi/partitioner/gpt.py
```diff
@@ -60,16 +61,27 @@ def create(
         :param int mbsize: partition size
         :param string type_name: partition type
         :param list flags: additional flags
+        :param int explicit_partition_id: If provided, use this exact partition ID
+                                         instead of auto-incrementing
```
schaefi: ditto: indentation

kiwi/storage/disk.py
```diff
@@ -150,31 +152,65 @@ def create_custom_partitions(
...
+            if custom_part_control and entry.boot_flag is True:
+                self.partitioner.set_flag(
+                    self.partition_id_map[map_name], 'f.active'
+                )
+
+            # Create canonical aliases for custom_part_control partitions
+            # This allows bootloader code to find partitions by their function
+            # (root, boot, efi) even when using custom partition names
+            if custom_part_control:
+                self._create_canonical_partition_alias(map_name, entry)
```
schaefi, line 212: I was wrapping my head around this method and under which conditions it is needed. You are actually setting `root`, `boot`, `efi` map entries if they are not present. But the conditional checks to add them e.g `t.efi` only exists when that partition has been added (create_efi_partition). In fact `root`, `boot` and `efi` are partitions from the protected space and not allowed to be an element of the custom partitions section.

I did not understand the logic in this case, can you explain this ? Thanks

kiwi/bootloader/config/base.py
```diff
@@ -522,8 +522,18 @@ def _mount_system(
...
         self.root_mount.mount(options=custom_root_mount_args)
 
+        log.debug(f'_mount_system: root_mount.device={self.root_mount.device}, boot_mount.device={self.boot_mount.device}')
         if not self.root_mount.device == self.boot_mount.device:
+            log.debug(f'_mount_system: Mounting boot partition at {self.boot_mount.mountpoint}')
             self.boot_mount.mount()
+            # Update bootpartition flag when boot is on separate partition
+            # This handles custom_part_control where bootpartition attribute
+            # may not be set but a separate boot partition exists
+            if hasattr(self, 'bootpartition'):
```
schaefi: Instead of `hasattr` can we make sure that the base class initializes this attribute

```
class BootLoaderConfigBase(ABC):
    def __init__(self, xml_state, root_dir, boot_dir=None, custom_args={}):
        ...
        self.bootpartition = False
        ...
```

schaefi: If it gets initialized as a False value in the base class, it becomes eventually overwritten by the respective bootloader implementation e.g. grub2. In _mount_system() there should really be no need to set this value as it's out of context and responsibility in this method

kiwi/builder/disk.py
```diff
@@ -385,7 +445,6 @@ def create_disk(self) -> Result:
                         device_map = self._map_luks(
                             device_map, luks_root
                         )
-
```
schaefi, line 388: this line removal seems unneeded ?

kiwi/storage/disk.py
schaefi: Oh I think I understand it, `custom_part_control` circumvents the protected check and also prevents the creation of the respective boot and efi partitions. As such a user has to setup these partitions and the build would fail if they are not specified.

Hmm, maybe it would be less intrusive if you add attributes like `bootpartition_number="42"` and `efipartition_number="43"` instead of handling them in the `<partitions>` section ?

As we have several attributes for boot and efi partitions in the type we considered `root`, `boot` and `efi` as protected partitions not covered by the custom partitions section. I believe it's also documented that way. Your new `custom_part_control` disables this sort of ;) 

I believe when you handle the partition number for the protected partitions as part of the `<type>` element you can get rid of `custom_part_control` and we would keep the logic and only need to enhance the docs by the new feature to be able to set partition numbers.

Thoughts ?

schaefi: There are also attributes like `efipartsize` which allows to specify the size of the EFI partition and with the new `custom_part_control` all of them would become pointless and moves into the `<partitions>` section. This can be pretty misleading and confusing imho

schaefi: So I thought your type section could be changed to:

```xml
<type
    image="oem"
    kernelcmdline="selinux=targeted console=ttyS0,115200" eficsm="false"
    firmware="uefi"
    filesystem="xfs"
    rootfs_label="Root"
    rootpartition_number="1"
    efipartition_number="128"
    bootpartition_number="127"
    bootpartition="true"
    bootpartsize="1024"
    bootfilesystem="ext4"
    efipartsize="100" 
>
        <oemconfig>
                <oem-systemsize>10000</oem-systemsize>
        </oemconfig>
        <bootloader name="grub2"/>
        <size unit="G">15</size>            
</type>
```

The problem is really that `custom_part_control` disables many existing type attributes for protected partitions boot,efi and root which where not subject of the custom `<partitions>` section

In the above layout your implementation would only need to add the attributes
* rootpartition_number
* efipartition_number
* bootpartition_number

and handle them in the in the disk builder such that it gets passed to the implementation you did for disk and storage classes.

All other already present type options would then still work in the exact same way as before. The size of the root partition is specified by the `oem-systemsize` which is a value applied in the target disk image if resize is switched off or applied on first boot if resize is active.

For the boot flag setup you might need another attribute.

I'm afraid if this violates a bit of the idea to manage everything in the `<partitions>` element and I hope you see my point ?

Thanks

kiwi/exceptions.py
cwolfe007: Yes, apologies here, I meant to update this before review. I agree that "legacy" has the wrong connotation and sounds as if the current way of partitioning is out of date, which it is not.

 I had updated other comments to use the phrase `auto-partition` instead of `legacy` as it is clearer that a user could either use the `cust_part_control` to have more control (and riskier) partition control or use the kiwi-ng way to auto build the paritions (which has inherit protections built in).

How do you feel about the term `auto-partition` to replace the term `legacy` here?

kiwi/storage/disk.py
cwolfe007: Taking a look at this feedback, you have some interesting points and I will get back to you later on after I have taken some time to reflect on this. Thank you for the thorough review
!

cwolfe007: Thanks again for the review, here are my thoughts. 

First I want to acknowldege d the concern that `custom_part_control` disables many existing type attributes, which could be confusing for users. I do see your point about maintaining consistency in the user experience.

That said, the spirit of this PR is to provide an opt-in "power user mode" for users with unusual partition requirements to manage ordering and numbering,. i.e  large systems migrating to kiwi-ng, or advanced users exploring non-standard partition layouts. I don't imagine most users would need this, but for those few, not having this level of control could be a blocker for adopting KIWI. I would not expect users to use this unless they really know what they are doing. 

One benefit of the `<partitions>` approach is that users who opt into this feature get a single, consolidated view of their entire disk layout - root, boot, efi, and any custom partitions all defined in one place and they can easily order the partitions however they would like. For users managing complex or non-standard partition schemes, having everything together improves clarity and makes it easier to reason about the full disk structure. The declarative, top-to-bottom ordering in XML also makes it intuitive to understand how partitions will be laid out on disk.

If the simpler `*partition_number` approach can meet these needs, that's the better path. If there are edge cases it can't cover (like for some users putting the partitions in an arbitrary order for reasons not yet known) , perhaps we could consider renaming this attribute to something  like `manual_part_control` with clear documentation that it's "use at your own risk" - signaling that users are taking full responsibility for their partition layout and lose some of the protections kiwi-ng builds-in inherently.

What do you think?

cwolfe007:
>A lot of work, thanks much for taking the time and for this effort 👍

haha, it was fun and np!

>I have a few questions especially on the alias setup. An integration test build to test this feature would also be great in kiwi/build-tests. Just one build for e.g. TW or another distro you prefer and for which we have integration tests would be great.

Ah I see, yes I can add in the integration tests.

kiwi/bootloader/config/base.py
cwolfe007: Agreed, this was unnecessary. I have added your suggestion

kiwi/exceptions.py
cwolfe007: replaced `legacy` with `auto-partition`

cwolfe007:
@schaefi I added Fedora and TumbleWeed image definition files in the the `build-test/` and tested both with qemu and they boot as expected. I believe I have addressed you're comments and continued the discussions in the various threads. 

How do you feel about this change?

Thanks!

Conan-Kudo:
So how does this work if you define a btrfs volume and want to have subvolumes in it, but _also_ want ordered partitions?

schaefi:
> So how does this work if you define a btrfs volume and want to have subvolumes in it, but _also_ want ordered partitions?

volumes are defined in the `<systemdisk>` section, partitions are on top of volumes in the optional `<partitions>` section. So it was always possible to combine both. Volumes are a feature of the filesystem (e.g. btrfs) or an extra indirection (LVM) but partitions always sits on top of it. As such there is no conflict

schaefi:
I'm still strugging with adding an attribute (manual_part_control) that allows and then also requires a manual custom partition layout which has to contain root, boot, efi and other partitions. Such a feature would allow for many side effects and failed builds with probably obscure error messages. e.g. if someone doesn't add a root partition. The fact that you had to implement extra validation methods in code showed that we now circumvent a concept that was not meant to be circumvented :) 

I still believe that it would be way easier to just add new attributes (xxx_number) for the partitions that are not meant to be a part of the optional `<partitions>` section. As a user you wold then only specify a partition number per attribute for the partitions kiwi creates mandatory according to the rest of the `<type>` spec and if you need further partitions those are listed in `<partitions>` and can also receive a custom partition number per entry.

> If there are edge cases it can't cover (like for some users putting the partitions in an arbitrary order for reasons not yet known) 

We had good reasons to pre-define the order in which the partitions are created such that resize capabilities are met, such that the layout is a consistent setting users can rely on, such that we can make sure the boot procedure for architecture 'X' is correct. I'm struggling with the capability that a user would be able to place the root partition first, followed by maybe swap and then boot or some other strange layout which is just source for trouble :) From my experience it also doesn't help if you provide fantastic documentation about a powerful feature and its many restrictions, people will do what you don't want them to do if you open this gate. In addition it's even today possible to mess with the partitions and ordering as you wish via existing script hooks. There is one example here:

https://github.com/OSInside/kiwi/tree/main/build-tests/x86/tumbleweed/test-image-custom-partitions

It's ugly it's messy I would never go with that for a production image but it's possible :) This way you could also move root somewhere else by preserving data temporarily then partition the disk as you wish and place the data back. So all of the super edge cases could be covered by ugly script code.

My point is I believe we should not allow everything and also aim for a consistent layout as we think it makes sense. Within this design the description can evolve. That's why I think adding the feature to allow custom partition numbering is great but we should not allow any arbitrary order of partitions with the schema definition for the reason to maybe cover an unknown edge case.

I hope this somewhat also makes sense to you ?

Conan-Kudo:
> > So how does this work if you define a btrfs volume and want to have subvolumes in it, but _also_ want ordered partitions?
> 
> volumes are defined in the `<systemdisk>` section, partitions are on top of volumes in the optional `<partitions>` section. So it was always possible to combine both. Volumes are a feature of the filesystem (e.g. btrfs) or an extra indirection (LVM) but partitions always sits on top of it. As such there is no conflict

I would not be comfortable with merging this without the test-case containing the combination to verify it.

xcawolfe-amzn:
> > >So how does this work if you define a btrfs volume and want to have subvolumes in it, but also want ordered partitions?

> > volumes are defined in the <systemdisk> section, partitions are on top of volumes in the optional <partitions> section. So it was always possible to combine both. Volumes are a feature of the filesystem (e.g. btrfs) or an extra indirection (LVM) but partitions always sits on top of it. As such there is no conflict

> I would not be comfortable with merging this without the test-case containing the combination to verify it.

@Conan-Kudo  Hmm yeah I agree with both points, considering that the partitioning has essentially been moved entirely to the custom partition logic, the existing subvolume functionality should work as expected. That said, we should validate. Ill add an example XML to `build_tests/` using the subvolume features 

cc @schaefi

xcawolfe-amzn:
> I still believe that it would be way easier to just add new attributes (xxx_number) for the partitions that are not meant to be a part of the optional <partitions> section.

@schaefi I certainly agree with this approach, and I think whether or not we go with the `manual_part_control` feature, this  `xxx_number` attribute would be good to add for those cases where people just want to number the reserved partitions and don't care about ordering. 

 As for the rest of your feedback, Ill get back to you later on when I take a deeper look at the examples you provided and reflect a bit more.

ozbenh:
Just a bit more background as to why this is done. The root reason for this is that we (Amazon Linux) want/need Kiwi to be able to generate layouts identical to what we generate today for Amazon Linux AMIs. We use a "funny" scheme in which partitions are numbered 1,127,128 but on disk are actually 127,128,1 ... 

There are historical reasons for this. There is too much legacy infrastructure and existing customer scripts assuming root is 1 and it needs to be physically at the end of the disk so it can be grown on first launch.

As for why 128,127 for grub and EFI , well ... why not ? :-) Those are fixed things, we wanted to keep the number space "in between" for whatever users might want to customize further.

Now as for your concerns about "too much freedom to shoot yourself in the foot", I would frankly not worry about it. If somebody is going to go down the path of using that "manual control" method, they know what they are doing and can deal with odd errors. I wouldn't expect most users to. But our own recipes would have been setup appropriately. I don't expect this feature to be used outside of case like ours where Kiwi is used to repelace existing tooling in a context were layouts are already "special" or people who really know what they are doing wanting to experiment with odd cases/scenarios.

schaefi:
@ozbenh @xcawolfe-amzn Thanks much for your feedback. I had another closer look to the code and I believe some code changes when done a bit differently could be added today without conflicts and reduces the amount of code to review. In order to implement the basic functionality to allow custom partition IDs the two API classes `Disk` and `Partitioner` must be touched. This API feature change can be done in a generic way and I pushed a new PR that does this based on your work. I hope you are ok with this approach as I believe it is better to split up the work, add what we can already add and I also don't want to loose any of the effort you have done so far.

Going forward I think we need to clarify on the following topics:

* how do we want to express the settings from an XML perspective ?
* Once we agreed on the schema changes, let's add them and effectively use them as part of integration tests

The final decision on the schema has quite some impact on the current implementation. For example, kiwi always sets the boot flag to the boot partition but not if this is skipped and the logic moves to the partitions section for which a new attribute got added. I also have concerns that we need validations of the XML data in python code because of `custom_part_control` requiring it. If we come to the conclusion that this is needed all extra validations needs to be implemented in `kiwi/runtime_checker.py`
Last but not least I took this code built a package and submitted it to staging, several integration builds started to fail. Yours worked but others doesn't. I haven't done a deeper inspection but we cannot afford a new feature to break existing stuff. You can look at the integration test matrix by calling `helper/build_status.sh` from the git. I reverted the Staging build to get back to a working state.

> Now as for your concerns about "too much freedom to shoot yourself in the foot", I would frankly not worry about it.

I agree with you, if it does not impact existing implementations and if the validation of the image description in this context is still clean and does not require extra code like `_validate_custom_partition_control_config`. My point here is, if you are sort of forced to write such a validation in code something seems broken or questionable designed in the data semantic and representation. Please also take into account that it feels like we are adding all this to serve a legacy path that you (Amazon) can revert/change/redesign at any time but we (kiwi) have to maintain it until the next major release

> We use a "funny" scheme in which partitions are numbered 1,127,128 but on disk are actually 127,128,1 ...
> There are historical reasons for this. There is too much legacy infrastructure and existing customer scripts assuming root is 1 > and it needs to be physically at the end of the disk so it can be grown on first launch.

Thanks, I can fully understand where you come from and that the numbers here really counts. I also read that the order of partitions is not the problem but just the numbering. As you mentioned root must be 1 but is in fact last as you cannot resize it otherwise, we shouldn't have any ordering issue (from a geometry point of view) and only have to make sure that we can set the partition IDs in a custom way.

From that perspective I still think the suggested XML description as proposed in a former comment should be sufficient

```xml
<type
    image="oem"
    kernelcmdline="selinux=targeted console=ttyS0,115200" eficsm="false"
    firmware="uefi"
    filesystem="xfs"
    rootfs_label="Root"
    rootpartition_number="1"
    efipartition_number="127"
    bootpartition_number="128"
    bootpartition="true"
    bootpartsize="1024"
    bootfilesystem="ext4"
    efipartsize="100" 
>
        <oemconfig>
                <oem-systemsize>10000</oem-systemsize>
        </oemconfig>
        <bootloader name="grub2"/>
        <size unit="G">15</size>            
</type>
```

This would place the partitions in geometry order (EFI, boot, root) and in partition ID order (127, 128, 1)

The implementation of the above would be straight forward, does not require any extra work with an unneeded partitions section for mandatory partitions that was not meant to be handled by the optional partitions sections. We also don't need any extra validations in code and we don't circumvent existing type attributes applied to mandatory partitions.

In addition I would consider the new attributes as useful for more than your "legacy"/"traditional" use case and it does not break with any of the existing code or integrations as we would only add a feature and not circumvent behavior in code depending on a `custom_partition_control` type attribute

Let me know what you think and I would appreciate a review of the PR I created to get parts of the implementation merged before we are complete.

Thanks much and wish you all Merry Christmas and a good start into 2026 :)

xcawolfe-amzn:
@schaefi Thanks for the additional context and info. Apologies for my delay in responding. Ill will take a look later on and get back to you about your feedback and new PR (which at a glance, adding part id to the api sounds great). 

If I don't get a chance to respond before the holiday, merry Christmas, happy new year, and happy holidays!

schaefi:
@xcawolfe-amzn I expressed how I invision the feature in code. See #2927 for details and testing. Thanks in advance :)

cwolfe007:
> Last but not least I took this code built a package and submitted it to staging, several integration builds started to fail. Yours worked but others doesn't. I haven't done a deeper inspection but we cannot afford a new feature to break existing stuff. You can look at the integration test matrix by calling helper/build_status.sh from the git. I reverted the Staging build to get back to a working state.

Hmm that is concerning, I was aiming to maintain backwards compatibility. Would you mind sharing the cases where you saw errors?

schaefi:
> Hmm that is concerning, I was aiming to maintain backwards compatibility.

Yes I saw in the review that you worked hard to keep the backward compatibility. Thanks much for this effort and it was also the reason why I built a Staging package such that I could build your integration test image. Next I saw a few failed builds but only had a very short look, one failed on a python error and another on the bootloader call.

As I believe we are still discussing the final solution I did not dig into errors and rather spent the time to use parts of your implementation and added a similar proposal according to my comments here in code.  I hope this is ok for you ? So we can now also discuss on the two PR's #2925 and #2927

I believe this way we can meet your needs and also prevent the extra attribute that allows to circumvent the kiwi design concept to not force customers to specify mandatory partitions. As of today and without any optional `<partitions>` section, kiwi was able to produce a disk layout suitable to build a working image depending on its architecture. That part I don't want to let go or open the gate for it with an extra attribute.

> Would you mind sharing the cases where you saw errors?

I'm sorry I reverted that staging build already to let the integration test based on #2925 and #2927 build. Would you mind to have a look at this test build. Details are in #2927

Hope this makes sense to you too

Conan-Kudo:
I think at this point, this has been superseded with the just-merged #2927.

PR #2921, opened 2025-12-12
Fix initrd outputfile format detection

### LINKED ISSUE #2918: `_get_boot_image_output_file_format_from_dracut_code()` not compatible with dracut-ng
A few years ago I reported something similar (#2149), but this time the fix is not as easy as updating the regex pattern.

Since https://github.com/dracut-ng/dracut-ng/commit/28820e20532890b9b0cf816ce1b527c582a1f3b3, the format of the initrd output file is set via configuration, so the code in `dracut.sh` changed from:

```bash
outfile="$dracutsysrootdir/boot/initramfs-${kernel}.img"
```

to:

```bash
outfile="${dracutsysrootdir-}/boot/$initrdname"
```

This `$initrdname` variable is defined in, for example:
- openSUSE: `/usr/lib/dracut/dracut.conf.d/01-dist.conf` - https://github.com/dracut-ng/dracut-ng/blob/30d121b1bd4ad2fb26ea72e1b0d5581663179882/dracut.conf.d/opensuse/01-dist.conf#L10
- Debian: ? - https://github.com/dracut-ng/dracut-ng/blob/30d121b1bd4ad2fb26ea72e1b0d5581663179882/dracut.conf.d/debian/01-debian.conf#L3

I came across this issue while integrating dracut-ng in openSUSE, since `_get_boot_image_output_file_format_from_dracut_code()` returns `None`, all openQA tests involving MicroOS images had the initrd named `initramfs-<kver>.img`, that is, it's taking the content of the `default_outfile_format` variable:

https://github.com/OSInside/kiwi/blob/24e884dca3d81dd2b7062a49147536fa5e6aa467/kiwi/boot/image/base.py#L484

While the initial image boots fine because kiwi also set grub entries using that initrd name, dracut generates new initrds using the configured `$initrdname`.

This issue is blocking the switch to dracut-ng in openSUSE.

schaefi:
With dracut-ng the output file format is a configuration option in the system wide dracut dist config file. The detection mechanics must be extended by also looking at this dist config. This Fixes #2918

### DIFF
--- a/kiwi/boot/image/base.py
+++ b/kiwi/boot/image/base.py
@@ -30,7 +30,11 @@ from kiwi.system.identifier import SystemIdentifier
 from kiwi.xml_description import XMLDescription
 from kiwi.xml_state import XMLState
 from kiwi.path import Path
+from kiwi.utils.sysconfig import SysConfig
 from kiwi.system.kernel import Kernel
+from kiwi.utils.command_capabilities import CommandCapabilities
+from kiwi.utils.temporary import Temporary
+from kiwi.command import Command
 
 from kiwi.exceptions import (
     KiwiTargetDirectoryNotFound,
@@ -491,6 +495,12 @@ class BootImageBase:
             outfile_format = \
                 self._get_boot_image_output_file_format_from_dracut_code()
 
+        if not outfile_format:
+            outfile_format = \
+                self._get_boot_image_output_file_format_from_config(
+                    kernel_version
+                )
+
         if outfile_format:
             return outfile_format
         else:
@@ -526,6 +536,27 @@ class BootImageBase:
         for initrd_file in glob.iglob(self.boot_root_directory + '/boot/init*'):
             if not os.path.islink(initrd_file):
                 return os.path.basename(initrd_file).replace(
+                    kernel_version, '{kernel_version}'
+                )
+        return None
+
+    def _get_boot_image_output_file_format_from_config(
+        self, kernel_version: str
+    ) -> Optional[str]:
+        if CommandCapabilities.has_option_in_help(
+            'dracut', '--printconfig', ['--help'],
+            root=self.boot_root_directory, raise_on_error=True
+        ):
+            dracut_call = Command.run(
+                ['chroot', self.boot_root_directory, 'dracut', '--printconfig']
+            )
+            temp_file = Temporary().new_file()
+            with open(temp_file.name, 'w') as config:
+                config.write(dracut_call.output.replace(' ', ''))
+            dracut_config = SysConfig(temp_file.name)
+            initrdname = dracut_config.get('dracutconfig:initrdname')
+            if initrdname:
+                return initrdname.replace(
                     kernel_version, '{kernel_version}'
                 )
         return None
--- a/test/data/dracut_config
+++ b/test/data/dracut_config
@@ -1,0 +1,8 @@
+dracutconfig: dracutmodules=all
+dracutconfig: fileloglvl=0
+dracutconfig: fw_dir=/lib/firmware/updates/6.18.0-2-default /lib/firmware/updates /lib/firmware/6.18.0-2-default /lib/firmware
+dracutconfig: hostonly_cmdline=yes
+dracutconfig: initrdname=initrd-kernel_version
+dracutconfig: kmsgloglvl=0
+dracutconfig: libdirs= /lib64 /usr/lib64 /lib /usr/lib
+dracutconfig: omit_drivers=^i2o_scsi$
--- a/test/unit/boot/image/base_test.py
+++ b/test/unit/boot/image/base_test.py
@@ -130,9 +130,16 @@ class TestBootImageBase:
     @patch('kiwi.boot.image.base.Path.which')
     @patch('kiwi.boot.image.base.log.warning')
     @patch('glob.iglob')
+    @patch('kiwi.boot.image.base.CommandCapabilities.has_option_in_help')
     def test_get_boot_names_default(
-        self, mock_iglob, mock_warning, mock_Path_which, mock_Kernel
+        self,
+        mock_CommandCapabilities_has_option_in_help,
+        mock_iglob,
+        mock_warning,
+        mock_Path_which,
+        mock_Kernel
     ):
+        mock_CommandCapabilities_has_option_in_help.return_value = False
         mock_iglob.return_value = []
         mock_Path_which.return_value = None
         mock_Kernel.return_value = self.kernel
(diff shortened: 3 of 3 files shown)


Conan-Kudo:
This should use `dracut --printconfig` and only attempt this logic when the option exists in dracut.

schaefi:
> This should use `dracut --printconfig` and only attempt this logic when the option exists in dracut.

I don't see it that strict. All distros I checked (Fedora, Rawhide, Leap, TW, Ubuntu) provides a system wide dracut dist config for dracut if dracut-ng is in use. They follow the same principals than systemd does. There are system wide settings and there are custom user settings.

In addition that magic `--printconfig` option does not exist in any of the dracut versions I have tested, however the switch to the dist config is already present. As such I don't see why reading the file when it exists is a problem

Add support for `--printconfig` can be added any time as yet another way to check it

This code doesn't hurt, is compatible with the other checks and can be extended to any new fancy feature of dracut

> This should use dracut --printconfig and only attempt this logic when the option exists in dracut.

This option also only reads the dist config ;)

Conan-Kudo:
No, that option reads the entire configuration across all layers. Remember that dracut has layered configs in /usr and /etc, so it's actually important to be able to deal with config layered in by the root overlay or config.sh.

Conan-Kudo:
The `--printconfig` option was added in Dracut 109: https://github.com/dracut-ng/dracut-ng/commit/8e24c4bf06850da1bc89022081b7e793548d9b0b

schaefi:
> No, that option reads the entire configuration across all layers. Remember that dracut has layered configs in /usr and /etc, so it's actually important to be able to deal with config layered in by the root overlay or config.sh.

I know, and you want to allow that `initrdname` can be overwritten by a custom user setting ? Not sure if that makes any sense

schaefi:
> The `--printconfig` option was added in Dracut 109: [dracut-ng/dracut-ng@8e24c4b](https://github.com/dracut-ng/dracut-ng/commit/8e24c4bf06850da1bc89022081b7e793548d9b0b)

If this is the decision a fix for the defect has to wait until all major distros are using >= 109

Fine with me

schaefi:
- [x] wait for `--printconfig`, moving code when distros can be tested with it

Conan-Kudo:
_I_ don't want to (I didn't even want the initrd name to be runtime configurable, but here we are), but incoherence with whatever the user expects is bad for us.

aafeijoo-suse:
> If this is the decision a fix for the defect has to wait until all major distros are using >= 109

I don't think this problem affects distributions other than openSUSE derivatives, it is present since dracut-102, so others use the default `initramfs-<kver>.img` name, or don't use kiwi at all.

Currently dracut-109 is available in the Factory devel project (https://build.opensuse.org/package/show/Base:System/dracut), it would only miss an easy patch on top providing `$initrdname` with `--printconfig`. If you implement this solution in kiwi, I'll add it so you can test it. Otherwise, I'd have to add an ugly patch with the expected `initrd-<kver`, ignoring the upstream `$initrdname` so we can switch to dracut-ng... I also don't like this being configurable by a user, but it's what we have.

schaefi:
@aafeijoo-suse 

> If you implement this solution in kiwi, I'll add it so you can test it

Thanks, I will implement the check via `--printconfig` and give your package a test

schaefi:
Hmm, I installed the latest dracut package

```
localhost:~ # dracut --version
dracut 109+suse.23.g0ac0b875-1

localhost:~ # dracut --printconfig
dracut[F]: Will not override existing initramfs (/boot/initrd-6.18.0-2-default) without --force

# Hmm, this smells

localhost:~ # dracut --printconfig --force
dracutconfig: dracutmodules=all
dracutconfig: fileloglvl=0
dracutconfig: fw_dir=/lib/firmware/updates/6.18.0-2-default /lib/firmware/updates /lib/firmware/6.18.0-2-default /lib/firmware
dracutconfig: hostonly_cmdline=yes
dracutconfig: kmsgloglvl=0
dracutconfig: libdirs= /lib64 /usr/lib64 /lib /usr/lib 
dracutconfig: omit_drivers=^i2o_scsi$

# Hmm, no information about the initrd name 
```

aafeijoo-suse:
> @aafeijoo-suse
> 
> > If you implement this solution in kiwi, I'll add it so you can test it
> 
> Thanks, I will implement the check via `--printconfig` and give your package a test

Thanks, I was waiting for your feedback before pushing that to Base:System. The package with the fix is here for now: https://build.opensuse.org/package/show/home:afeijoo:branches:openSUSE:Factory:dracut-ng/dracut

schaefi:
And please just because it feels a bit like it. I'm not willing to implement anything distribution specific. As such, is it correct that `--printconfig` is an upstream accepted and consistent implementation ?

schaefi:
> Thanks, I was waiting for your feedback before pushing that to Base:System. The package with the fix is here for now: https://build.opensuse.org/package/show/home:afeijoo:branches:openSUSE:Factory:dracut-ng/dracut

Ah ok, in your home. All right going to test this one

aafeijoo-suse:
> And please just because it feels a bit like it. I'm not willing to implement anything distribution specific. As such, is it correct that `--printconfig` is an upstream accepted and consistent implementation ?

See https://github.com/dracut-ng/dracut-ng/pull/1928

schaefi:
I added support for `--printconfig` This needs real world testing though and we should wait when dracut has landed in a distro such that we can test it with the existing integration tests.

Probably after my vacation

Thanks

schaefi:
I will rebase the commits once we have a successful integration test running

schaefi:
@Conan-Kudo rebase is done, I submitted a version of kiwi to Staging. Testing has been done with dracut-ng from the mentioned home project here and any other distro that already has it will make use of the new --printconfig. Rawhide is also still at v108 of dracut. As such we can wait or we merge this. I vote for merging and wait with a release until the rolling distros TW and rawhide have it

Conan-Kudo (approved):

AdamWill:
Looks to me like this was confused by the misleadingly-named `raise_on_error` argument for `has_option_in_help`. That actually raises an exception if the check *fails* - the flag is not present - rather than raising an exception if the check somehow *errors*. So that made `dracut --printconfig` mandatory, effectively, in any case where `_get_boot_image_output_file_format_from_existing_file` or `_get_boot_image_output_file_format_from_dracut_code` didn't do the job first.

I've sent https://github.com/OSInside/kiwi/pull/2944 to fix it up.

AdamWill:
Hmm, I also note this neglected to update the docstring of `_get_boot_image_output_file_format`, which still only describes *three* stages, when now there's a fourth.

PR #2925, opened 2025-12-17
Support custom partition IDs in Partitioner and Disk API

schaefi:
This change is two fold

**Add support for custom partition ID to Disk**
    
Allow to specify a custom partition ID in the Disk class interface and API. As of today the new parameter partition_id in the ptable_entry_type record is not in use by the current kiwi image description schema. This change will be a followup to the API change and needs further conversation upstream first.
    
**Add support for custom partition ID to Partitioner**
    
Allow to specify a custom partition ID in the Partitioner class interface and API.

This code changes are based on the work from:

Co-authored-by: Caleb Wolfe <cwolfe651@gmail.com>
Co-authored-by: Caleb Wolfe (Amazon) <xcawolfe@amazon.com>
Assisted-by: Claude Opus 4.5 <noreply@anthropic.com>

This is related to #2917

PR #2927, opened 2025-12-23
Add custom partition id to schema

schaefi:
Based on the changes proposed in #2917 and my comments there I'd like to suggest an alternative solution based on the API update in #2925. On top of this PR the required changes to meet your needs and an integration test build has been created with this PR.

Please find the integration test here:
* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:fedora/test-image-amazon-linux-layout
 
You can test boot the image as follows:

```
wget wget https://download.opensuse.org/repositories/Virtualization:/Appliances:/Images:/Testing_x86:/fedora/images/kiwi-test-image-amazon-linux-layout.x86_64.raw.xz
xz -d kiwi-test-image-amazon-linux-layout.x86_64.raw.xz
kvm -bios /usr/share/qemu/ovmf-x86_64.bin -hda kiwi-test-image-amazon-linux-layout.x86_64.raw
```

I believe the disk layout meets the needs according to our conversation. I also think that the feature in the way I invision it, is less intrusive and does not circumvent existing implementations. The part that counts the most for me is that we don't force users to provide their own idea of a partition layout in all aspects and can still keep on the constraints that we have set for kiwi in this regard. With the above build the layout of the disk is as follows

```
[root@fedora ~]# fdisk -l
Disk /dev/sda: 10 GiB, 10737418240 bytes, 20971520 sectors
Disk model: QEMU HARDDISK
Disklabel type: gpt
Disk identifier: 35995CCD-F15E-40E5-9AAF-7C6E6CC4340A

Device        Start      End  Sectors  Size Type
/dev/sda1   6400000 20971486 14571487  6.9G Linux root (x86-64)
/dev/sda13  2304000  6399999  4096000    2G Linux filesystem
/dev/sda55     2048   206847   204800  100M EFI System
/dev/sda101  206848  2303999  2097152    1G Linux extended boot

Partition table entries are not in disk order.
```

Please check it out yourself and let's discuss what we are still missing. Thanks in advance

Still missing:

- [x] Documentation update

### DIFF
--- a/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
+++ b/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
@@ -13,6 +13,7 @@
     <profiles>
         <profile name="Live" description="Live image of Fedora"/>
         <profile name="Virtual" description="Virtual image of Fedora"/>
+        <profile name="Virtual_bootpart" description="Virtual image of Fedora with extra boot partition"/>
         <profile name="Disk" description="OEM image of Fedora"/>
     </profiles>
     <preferences>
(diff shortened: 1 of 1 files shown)


xcawolfe-amzn:
I am good with this change, it covers the basic case for amazon linux and I tested with an amazon linux image definition and it seems to work as expected. For most users I think this is the right way to go. Thanks for this PR @schaefi! 

I will follow up on #2917 for the other edge cases we are thinking about there

Conan-Kudo:
What does this look like if we also have the Fedora Btrfs layout mixed in? I'd like a test case that includes that too.

schaefi:
> What does this look like if we also have the Fedora Btrfs layout mixed in? I'd like a test case that includes that too.

We can add that to one of the fedora integration tests. Which partition would you see to make use of a custom ID ?
As I wrote in some former comment the btrfs layer and its subvolumes/snapshots is on top of the partition table. As such nothing will change for the rootfs if the main partition is /dev/sda3 or /dev/sda42 :)

Please give me some details how you want this to look like

Thanks

Conan-Kudo:
I'd like two variants, one that follows the F43 layout (with ESP and /boot partitions) and one that follows the F44+ layout (with only an ESP partition). The numbers can be the same as the Amazon test.

F43 layout: https://pagure.io/fedora-kiwi-descriptions/blob/f43/f/teams/cloud/cloud.xml#_71-90

F44+ layout: https://pagure.io/fedora-kiwi-descriptions/blob/rawhide/f/teams/cloud/cloud.xml#_73-93

schaefi:
ok, going to add that next year :smile:

build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
```diff
@@ -30,8 +31,47 @@
...
+            btrfs_set_default_volume="false"
+            rootfs_label="fedora"
+            bootpartition="true"
+            efipart_id="55"
+            rootpart_id="1"
+            bootpart_id="101"
+        >
+            <systemdisk name="fedora">
+                <volume name="@root=root"/>
+                <volume name="boot" parent="/"/>
```
Conan-Kudo, line 51: Since you have a separate boot partition, this subvol goes away.

schaefi:
Please find additional integration test builds for the Fedora btrfs layout using custom partition IDs here

The profile `Virtual` and `Virtual_bootpart` provides image builds based on btrfs using custom partition numbers in the same way as they are set for the amazon test

Conan-Kudo (approved):

PR #2930, opened 2026-01-09
Fixed ramdisk sysroot generator

schaefi:
Do not use a custom _dev name and stick with the UUID representation of the disk image in RAM after deployment. Former versions of udev did not create a by-uuid device representation which now seems to have changed. This then leads to the device name RamDisk_rootfs not being created and the respective .device unit times out. In addition the timer unit for the standard device representation changed to infinity. This fixes bsc#1254116


schaefi:
yep I have done local testing but let's wait for the bug feedback

PR #2933, opened 2026-01-20
Move OVA support to open-vmdk

### LINKED ISSUE #2292: Switch from ovatool for building OVAs to open-vmdk
## Problem description

kiwi currently uses ovatool for building vmware OVA images. there is an Apache-2.0 licensed project by VMware for this instead:

https://github.com/vmware/open-vmdk

I don't know how feature complete it is, but it would be nice to be able use it with kiwi and figure out any incompatibilities. OVA format is largely preferred by VMware users over the other option of vmdk. For users that can not build their images using the proprietary ovatool, open-vmdk is the only option.

## Expected behaviour

Stream-optimized OVA images can be built without dependency on proprietary tooling with kiwi.

Not relevant

schaefi:
Finally this commit drops the use of the VMware ovftool and moves to the real opensource alternative open-vmdk This Fixes #2292

### DIFF
--- a/kiwi/storage/subformat/ova.py
+++ b/kiwi/storage/subformat/ova.py
@@ -16,20 +16,24 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
-import stat
+import pathlib
 from textwrap import dedent
 
 # project
+from kiwi.firmware import FirmWare
 from kiwi.storage.subformat.vmdk import DiskFormatVmdk
 from kiwi.storage.subformat.base import DiskFormatBase
 from kiwi.command import Command
-from kiwi.utils.command_capabilities import CommandCapabilities
 from kiwi.path import Path
 from kiwi.system.result import Result
+from kiwi.storage.subformat.template.ova_compose import (
+    OvaComposeTemplate
+)
 
 from kiwi.exceptions import (
     KiwiFormatSetupError,
-    KiwiCommandNotFound
+    KiwiCommandNotFound,
+    KiwiTemplateError
 )
 
 
@@ -45,11 +49,20 @@ class DiskFormatOva(DiskFormatBase):
 
         :param dict custom_args: custom qemu arguments dictionary
         """
-        ovftype = self.xml_state.get_build_type_machine_section().get_ovftype()
+        ovftype = 'vmware'  # default OVA target infrastructure
+        self.machine_setup = self.xml_state.get_build_type_machine_section()
+        if self.machine_setup and self.machine_setup.get_ovftype():
+            ovftype = self.machine_setup.get_ovftype()
         if ovftype != 'vmware':
             raise KiwiFormatSetupError(f'Unsupported ovftype {ovftype}')
         self.image_format = 'ova'
-        self.options = self.get_qemu_option_list(custom_args)
+        # ensure streamOptimized format is enabled
+        custom_args.update(
+            {'subformat=streamOptimized': None}
+        )
+        self.firmware = FirmWare(
+            self.xml_state
+        )
         self.vmdk = DiskFormatVmdk(
             self.xml_state, self.root_dir, self.target_dir, custom_args
         )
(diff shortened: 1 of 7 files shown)


package/python-kiwi-spec-template
```diff
@@ -369,6 +369,7 @@ Requires:       kpartx
 Requires:       cryptsetup
 Requires:       mdadm
 Requires:       util-linux
+Requires:       open-vmdk
```
schaefi, line 373: @Conan-Kudo I need your help to do this correctly. Thanks much

Conan-Kudo: I don't see anything wrong? It seems right to me.

Conan-Kudo: Ah, it's only available in Fedora and openSUSE, so this needs to be guarded at the moment. I'll get it added to EPEL for CentOS/RHEL.

So for now, this needs to be:

```suggestion
%if 0%{?fedora} || 0%{?rhel} || 0%{?suse_version}
Requires:       open-vmdk
%endif
```

Vogtinator: If in doubt, file requires? (`Requires: /usr/bin/mkova.sh`)

kiwi/storage/subformat/ova.py
```diff
@@ -59,16 +65,14 @@ def create_image_format(self) -> None:
         Create ova disk format using ovftool from
         https://www.vmware.com/support/developer/ovf
         """
-        # Check for required ovftool
-        ovftool = Path.which(filename='ovftool', access_mode=os.X_OK)
-        if not ovftool:
+        # Check for required mkova tool
+        mkova = Path.which(filename='mkova.sh', access_mode=os.X_OK)
```
Vogtinator: Apparently mkova.sh is legacy and something else should be used? https://github.com/vmware/open-vmdk/blob/3508f07162a3538429544a155dfb19f1c02a47c2/README.md?plain=1#L18

package/python-kiwi-spec-template
Conan-Kudo: We'd still need the package requires once Debian has the package.

kiwi/storage/subformat/ova.py
Conan-Kudo: Yeah, `ova-compose` should be used instead.

schaefi: Thanks, so this needs to change and we need to create that input file from the machine description. I will take
https://github.com/vmware/open-vmdk/blob/master/ova-compose/example.yaml as example

package/python-kiwi-spec-template
schaefi: done

kiwi/storage/subformat/ova.py
schaefi: code updated... tests are still missing

schaefi: done

schaefi:
@Conan-Kudo @Vogtinator I appreciate a review of the template data to be solid as default. I have no opportunity to test the OVA in something like VMware Vsphere

Conan-Kudo (approved):
The code seems reasonable. As a followup, we probably want to eventually allow users to define their own templates that kiwi uses as input (so things like OS type and whatnot can be set).

schaefi:
> The code seems reasonable. As a followup, we probably want to eventually allow users to define their own templates that kiwi uses as input (so things like OS type and whatnot can be set).

Or we extend the existing `<machine>` section with eventually missing settings. To be honest the current `<machine>` section already allows to specify quite some details as I added most of the guest settings that you can manage with VMware vmplayer. In addition there is also a `custom_entry` field which you can use to add arbitrary information. It's basically a matter of making the template stable enough to serve as a default. At the moment it seems that `ova-compose` itself is lacking some possible OVA settings for the `vmware` provider. For example on the network setup side I could not see that it handles different drivers, static mac assignment and things like that. The disk driver seems to be only validating on nvme devices etc etc. 

I believe further change will be needed in the future but with this PR we set the stage. So I'm ok merging it in this state

Conan-Kudo:
Yes, we can figure out what to do in future PRs. I like your idea about extending the machine stanza.

PR #2935, opened 2026-01-23
Add plymouth_msg

uecasm:
* Adds `plymouth_msg` to kiwi-repart

----
This is mostly a minimal proof of concept; it can be merged as is but I expect you might want to move `plymouth_msg` to the library and then use it in other places too.

The core idea is that I noticed on large disks the first-boot repartitioning could take long enough that people might wonder what's going on, especially on systems running plymouth.  This adds a simple message that displays on plymouth to let people know what it's doing (and implicitly that subsequent boots should be faster).

One caveat is that at least on current versions of OpenSUSE (tested in 15.6 but at time of writing it hasn't been fixed in later versions) there's a bug in `plymouth-dracut` which means that `label-pango.so` is not installed to the initrd and thus no text will display.  This can either be fixed by patching `/usr/lib/plymouth/plymouth-populate-initrd` to install `label-pango.so` instead of `label.so`, or by adding an equivalent entry to any dracut module, e.g. `install_optional_items+=" /usr/lib64/plymouth/label-pango.so "`.

### DIFF
--- a/dracut/modules.d/55kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/55kiwi-repart/kiwi-repart-disk.sh
@@ -117,6 +117,12 @@ function get_target_rootpart_size {
     echo "${kiwi_oemrootMB}"
 }
 
+function plymouth_msg {
+    if command -v plymouth >/dev/null 2>&1; then
+        plymouth display-message --text="$1"
+    fi
+}
+
 function repart_standard_disk {
     # """
     # repartition disk with read/write root filesystem
@@ -322,6 +328,8 @@ if ! resize_wanted "${last_device}" "${disk}"; then
     return
 fi
 
+plymouth_msg "Expanding disk..."
+
 # prepare disk for repartition
 if [ "$(get_partition_table_type "${disk}")" = 'gpt' ];then
     relocate_gpt_at_end_of_disk "${disk}"
@@ -357,6 +365,8 @@ if lvm_system; then
 else
     resize_filesystem "$(get_root_map)"
 fi
+
+plymouth_msg ""
 
 # wait for the root device to appear
 wait_for_storage_device "${last_device}"
--- a/dracut/modules.d/55kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/55kiwi-repart/kiwi-repart-disk.sh
@@ -118,6 +118,7 @@ function get_target_rootpart_size {
 }
 
 function plymouth_msg {
+    [ -z "$1" ] || info "$1"
     if command -v plymouth &>/dev/null; then
         plymouth display-message --text="$1"
     fi
--- a/dracut/modules.d/55kiwi-repart/kiwi-repart-disk.sh
+++ b/dracut/modules.d/55kiwi-repart/kiwi-repart-disk.sh
@@ -118,7 +118,7 @@ function get_target_rootpart_size {
 }
 
 function plymouth_msg {
-    if command -v plymouth >/dev/null 2>&1; then
+    if command -v plymouth &>/dev/null; then
         plymouth display-message --text="$1"
     fi
 }


dracut/modules.d/55kiwi-repart/kiwi-repart-disk.sh
```diff
@@ -117,6 +117,12 @@ function get_target_rootpart_size {
     echo "${kiwi_oemrootMB}"
 }
 
+function plymouth_msg {
+    if command -v plymouth >/dev/null 2>&1; then
```
schaefi: can we change this to

```bash
if command -v plymouth &>/dev/null; then
    :
fi
```

or

```bash
if type plymouth &>/dev/null; then
    :
fi
```

as you prefer. Thanks

dracut/modules.d/55kiwi-repart/kiwi-repart-disk.sh
```diff
@@ -322,6 +328,8 @@ if ! resize_wanted "${last_device}" "${disk}"; then
     return
 fi
 
+plymouth_msg "Expanding disk..."
```
schaefi, line 332: I like that you want to expose the messages also to the framebuffer if plymouth is enabled. However, I believe we should not duplicate different messages depending on where they are displayed. 

In the kiwi dracut code I'm using the dracut provided messaging system. You can see this when looking for

```
warn "..."
info "..."
```

kind a like messages. I believe it would be good if we check if dracut itself is not already capable to show this as plymouth message and if not I would propose that we create our own `warn` , `info` functions which wraps the dracut provided ones and also calls your code to display the same message also through plymouth.

This way we still have the same messages but just stream them to different channels.

Thoughts ?

schaefi (changes requested):
Thanks, a good idea 👍  Please see my comments

dracut/modules.d/55kiwi-repart/kiwi-repart-disk.sh
schaefi: You find the implementation of the messaging system in dracut at `./99base/dracut-lib.sh`

uecasm: I'm not sure this entirely makes sense; `info` and `warn` are often used multi-line, while `plymouth_msg` can only display a single line at a time.

`plymouth_msg` also requires explicitly being called with an empty string to indicate "previous message no longer applicable".

In general I think the sort of information that would be logged by each of these would likely differ; `plymouth_msg` would likely only be useful for lengthy operations, and most of the boot scripts are trying to avoid those.

I could see a good argument to make `plymouth_msg` _also_ call `info`, rather than the reverse, though.

uecasm: FWIW, it looks like there are a few dracut modules that do isolated plymouth commands, but nothing centralised.

schaefi: ok got it and thanks for clarification

dracut/modules.d/55kiwi-repart/kiwi-repart-disk.sh
```diff
@@ -117,6 +117,13 @@ function get_target_rootpart_size {
     echo "${kiwi_oemrootMB}"
 }
 
+function plymouth_msg {
+    [ -z "$1" ] || info "$1"
+    if command -v plymouth &>/dev/null; then
+        plymouth display-message --text="$1"
+    fi
+}
```
schaefi, line 125: So if the given message is empty the method will not log via `info` but will still call `plymouth display-message` with --text being empty. Is this a wanted behavior ?

If not how about 

```bash
local message=$1
if [ -n "${message}" ];then
    info "${message}"
    if command -v plymouth &>/dev/null; then
        plymouth display-message --text="${message}"
    fi
fi
```

schaefi: I also think it would be good to add the method in `dracut/modules.d/59kiwi-lib/kiwi-dialog-lib.sh` for consistency and that other code can also use it.

Thoughts ?

uecasm: > So if the given message is empty the method will not log via info but will still call plymouth display-message with --text being empty. Is this a wanted behavior ?

Yes, calling it with an empty string (and not logging that) is the intended behaviour.  That's how you clear the message from Plymouth to indicate that it's no longer relevant.  See:

https://github.com/uecasm/kiwi/blob/eccf4a70f9ce1033a5d72691e8f275dbc2f39495/dracut/modules.d/55kiwi-repart/kiwi-repart-disk.sh#L370

(You can, of course, just call again with another non-empty string if you're going from one reported section to another; you don't have to explicitly clear in between, just at the end.)

Ideally it'd be set as an EXIT trap or something so that it always fires but I wasn't sure if it was safe to mess with that, and if anything does fail in here there's a decent chance you'd get dumped to an emergency shell anyway.

> I also think it would be good to add the method in dracut/modules.d/59kiwi-lib/kiwi-dialog-lib.sh for consistency and that other code can also use it.

That's fair, and I did suggest that in the description, but I wanted it to be a fairly self-contained change to gauge interest.

schaefi: > Yes, calling it with an empty string (and not logging that) is the intended behaviour. That's how you clear the message from Plymouth to indicate that it's no longer relevant.

Ah ok, got it, thanks. Can we make this behavior somehow visible by a comment at call time. I think it's not clear without one.

> That's fair, and I did suggest that in the description, but I wanted it to be a fairly self-contained change to gauge interest.

I think it's nice to have this feature available and I agree with your observations. personally I do not stumble over the problem because most of the deployments I'm connected with have no opportunity to display something graphically and the images are also not that big.

So if you can move the method to the library I'm happy to merge the PR 

Thanks much

schaefi: or maybe instead of 

```
plymouth_msg ""
```

we add another method that does the wanted action, e.g. 

```
plymouth_done
```

or something that makes the action more clear

Thoughts ?

schaefi: @uecasm ping ?

schaefi: @uecasm ping ?

schaefi:
Hmm, unfortunately no further information

PR #2936, opened 2026-01-23
Identify the correct binary for DNF4

Conan-Kudo:
This is pretty much the same solution we used back when we supported YUM and had to deal with yum vs yum-deprecated.

### DIFF
--- a/build-tests/x86/centos/test-image-live-disk-v10/appliance.kiwi
+++ b/build-tests/x86/centos/test-image-live-disk-v10/appliance.kiwi
@@ -92,5 +92,6 @@
         <package name="grub2-efi-x64" arch="x86_64"/>
         <package name="shim-x64" arch="x86_64"/>
         <package name="libdb-utils"/>
+        <package name="dnf"/>
     </packages>
 </image>
--- a/build-tests/x86/centos/test-image-live-disk-v9/appliance.kiwi
+++ b/build-tests/x86/centos/test-image-live-disk-v9/appliance.kiwi
@@ -96,5 +96,6 @@
         <package name="grub2-efi-x64" arch="x86_64"/>
         <package name="shim-x64" arch="x86_64"/>
         <package name="libdb-utils"/>
+        <package name="dnf"/>
     </packages>
 </image>


kiwi/package_manager/dnf4.py
```diff
@@ -88,6 +89,31 @@ def request_package_exclusion(self, name: str) -> None:
...
+            'PATH': os.sep.join([root, 'usr', 'bin'])
+        } if root else None
+
+        # Python interpreter specific path
+        if Path.which(
+            filename='dnf4',
+            custom_env=dnf4_search_env,
+            access_mode=os.X_OK
+        ):
+            dnf4_binary = 'dnf4'
```
schaefi, line 113: I think you can change this to

```python3
dnf_binary = 'dnf-4' if Path.which(
    filename='dnf4', access_mode=os.X_OK, root_dir=root
) else 'dnf-3'

return dnf_binary
```

kiwi/package_manager/dnf4.py
```diff
@@ -88,6 +89,31 @@ def request_package_exclusion(self, name: str) -> None:
         """
         self.exclude_requests.append(name)
 
+    def _get_dnf4_binary_name(self, root=None):
```
schaefi, line 92: The function name is a bit misleading as it could return `dnf-3` or `dnf-4`. You are in the PackageManagerDnf4 namespace, as such you can also simply name it just `_get_dnf_binary_name`. Thoughts ?

schaefi: It stays misleading though because I would not expect dnf-3 from a PackageManagerDnf4 instance. But I think we can live with that as a technical debt

kiwi/package_manager/dnf4.py
schaefi: It actually means that we are using `dnf-4` if that exists in the root tree of the image and else we are using `dnf-3` unconditionally. Doesn't this open some room for issues ? What if `dnf-3` doesn't exist ? Also when checking inside the root tree then checking outside of it would also be needed as the bootstrap phase will use the package manager from the host. The decision logic here is not quite clear to me

schaefi: If the idea was to lookup PATH (host) and in $root then the code needs to run two lookup calls, because the first call would be happy if there is just one hit in the given PATH

kiwi/package_manager/dnf4.py
Conan-Kudo: Yeah, I don't love it, but it's fine. I'd rather leave the function name alone since it makes it straightforward if we're looking at it in a live debugger.

kiwi/package_manager/dnf4.py
Conan-Kudo: This is the exact implementation we used in the old yum one: https://github.com/OSInside/kiwi/blob/v9.16.36/kiwi/package_manager/yum.py#L89-L110

Conan-Kudo: At least with all the current supported distributions, the `dnf-3` binary exists unless the per-Python version suffixed binaries don't exist, and those have `dnf4` instead.

Conan-Kudo: We just can't check for `dnf` because it's an ambiguous binary now.

schaefi: I zip a Staging build and give it a test

schaefi: build succeeding now

PR #2939, opened 2026-01-28
Fix booting non-encrypted live images

Vogtinator:
activate_luks returned true if the passed blockdev was not luks.

Bug report: https://bugzilla.opensuse.org/show_bug.cgi?id=1257389


schaefi:
you were faster I was coding a fix for that at the same time and wondered why there is a conflict :)
Thanks for fixing, I throw away mine and zip a new Staging build

PR #2940, opened 2026-01-28
Fix doc issues

schaefi:
This change it two fold

**Fix doc: Booting a Live ISO Images from Grub2**
    
The chapter described the setup of a menu entry for grub  to loop boot a kiwi live ISO. The information was outdated and did not longer apply to the current state of the art. This commit updates the documentation to be correct again.

**Fix doc: Setting Up YaST at First Boot**
    
The firstboot wizzard in YaST which allowed to run AutoYaST from a specific trigger setup no longer exists and now executes the yast-firstboot service. However, firstboot.xml  is something completely different than AutoYaST.xml which obsoletes the documentation we wrote in kiwi about it and  also confuses our users. I was not able to find a proper utility which allows an easy setup of firstboot.xml in  the same way as it existed for "yast autoyast". As SUSE  also obsoleted and dropped YaST from their products the  only stable move is to drop this chapter from the documentation.


doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
```diff
@@ -15,52 +15,84 @@ In {kiwi}, all generated ISO images are created to be hybrid. This means,
...
 The dracut initrd system used by {kiwi} provides this
 feature upstream called as "iso-scan/filename". Therefore all {kiwi} generated
 live ISO images supports this deployment mode.
 
-The following procedure shows how to setup Grub2 on your hard drive:
+The following procedure expects an existing Grub2 installation on your
+hard drive or USB stick for the x86_64 architecture.
 
 1. Copy the ISO image to a folder of your choice on your hard drive.
+   For example, you can create a folder called "iso" in the
```
tomschr: Remove filler words:

```suggestion
   For example, create a folder called "iso" in the
```

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
```diff
@@ -15,52 +15,84 @@ In {kiwi}, all generated ISO images are created to be hybrid. This means,
...
+   .. code:: bash
+
+      sudo mkdir /iso
+      sudo cp some-kiwi-live.iso /iso/
 
-   Be sure to set the path to the ISO image, you can set your own menu name.
-   The drive identification for Grub2 can be checked at boot time 
-   by pressing the 'c' key and typing 'ls'.
+2. Lookup the root filesystem UUID of your current operating system.
+   You can get this information by running the command:
```
tomschr: Remove filler words:

```suggestion
   Get this information by running the command:
```

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
```diff
@@ -15,52 +15,84 @@ In {kiwi}, all generated ISO images are created to be hybrid. This means,
...
-      submenu "Boot from openSUSE ISO" {
-	   iso_path="(hd0,gpt2)/path/to/openSUSE.iso"
-	   export iso_path
-	   loopback loop "$iso_path"
-	   root=(loop)
-	   source /boot/grub2/loopback.cfg
-	   loopback --delete loop
+      findmnt -n -o UUID /
+
+2. Add the following submenu setup to the `grub.cfg` file:
```
tomschr: If I'm not mistaken, you should use the `:file:` macro here:

```suggestion
2. Add the following submenu setup to the :file:`grub.cfg` file:
```

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
```diff
@@ -15,52 +15,84 @@ In {kiwi}, all generated ISO images are created to be hybrid. This means,
...
+
+.. note::
 
-For USB sticks, the procedure is identical. You would then install 
-Grub2 on the USB drive and follow the steps above. 
-The use of scripts such as "MultiOS-USB" is strongly recommended.
+   For USB sticks, the procedure is identical. You would then install
+   Grub2 on the USB drive and follow the steps above. However, it is
+   uncommon to have Grub2 already installed on a USB stick and there are
+   good helper tools such as `MultiOS-USB` which simplifies the setup.
```
tomschr: I would slightly change the order:

```suggestion
   For USB sticks, the procedure is identical. You would install
   Grub2 on the USB drive and follow the steps above. However, it is
   uncommon to have Grub2 already installed on a USB stick. To simplify
   the setup, there are good helper tools such as `MultiOS-USB`.
```

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
```diff
@@ -15,52 +15,84 @@ In {kiwi}, all generated ISO images are created to be hybrid. This means,
...
-	   source /boot/grub2/loopback.cfg
-	   loopback --delete loop
+      findmnt -n -o UUID /
+
+2. Add the following submenu setup to the `grub.cfg` file:
+
+   .. note::
+
+      Be sure to set the correct path to the ISO image, you can set
+      your own menu name. The drive identification for Grub2 can be
```
tomschr: I would replace the "Be sure" and avoid the "can" to formulate it more strict:


```suggestion
      Verify that the path to the ISO image is correct.
      Set your own menu name. The drive identification for Grub2 can be
```

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
```diff
@@ -15,52 +15,84 @@ In {kiwi}, all generated ISO images are created to be hybrid. This means,
...
 the ISO image also has a partition table embedded. With more and more
 computers delivered without a CD/DVD drive this becomes important.
-The deployment of such an image onto a disk like an USB stick normally 
-destroys all existing data on this device. It is also not possible to use 
-USB stick as a data storage device. Most USB sticks are pre-formatted 
-with a FAT32 or exFAT Windows File System and to keep the existing data 
+The deployment of such an image onto a disk like an USB stick normally
+destroys all existing data on this device. It is also not possible to use
+USB stick as a data storage device. Most USB sticks are pre-formatted
+with a FAT32 or exFAT Windows File System and to keep the existing data
```
tomschr: I see some issues with this sentence:

* "Deployment" is a bit heavy for an ISO image. "Writing" is more common for ISOs.
* "Normally" is a dangerous word here. 😉 If you are writing an ISO image to a drive, it **will** overwrite the partition table and data. "Normally" implies there is a chance the data might survive, which gives the user false hope.
* Since you just said the data will be erased, you want to signal that the drive remains unusable for storage *after* the process.

Combining all of them, here is my suggestion:

```suggestion
Writing this image to a USB stick will permanently erase all existing data
on the device. Additionally, the stick will no longer be usable for
general data storage. Most USB sticks are pre-formatted
with a FAT32 or exFAT Windows file system and to keep the existing data
```

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
```diff
@@ -15,52 +15,84 @@ In {kiwi}, all generated ISO images are created to be hybrid. This means,
...
+The following deployment process copies the ISO image as an additional file
+to the USB stick or hard drive. The ability to boot from the disk is configured
 through a Grub2 feature which allows to loopback mount an ISO file and boot the
 kernel and initrd directly from the ISO file.
 
 The initrd loaded in this process must also be able to loopback
 mount the ISO file to access the root filesystem and boot the
-live system. Almost every Linux distribution supports fat32, 
-and more and more of them also support exFAT. For hard drives, 
+live system. Almost every Linux distribution supports fat32,
```
tomschr: Typo:

```suggestion
live system. Almost every Linux distribution supports FAT32,
```

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
```diff
@@ -15,52 +15,84 @@ In {kiwi}, all generated ISO images are created to be hybrid. This means,
...
+and more and more of them also support exFAT. For hard drives,
 Linux filesystems are also supported.
 
 The dracut initrd system used by {kiwi} provides this
 feature upstream called as "iso-scan/filename". Therefore all {kiwi} generated
 live ISO images supports this deployment mode.
 
-The following procedure shows how to setup Grub2 on your hard drive:
+The following procedure expects an existing Grub2 installation on your
+hard drive or USB stick for the x86_64 architecture.
```
tomschr: Hmn, if you want to combine standard Intel or AMD processors, you could replace "x86_64" with "64-bit". Otherwise you would need to add ARM64, AArch64 too, right?


```suggestion
hard drive or USB stick for the 64-bit architecture.
```

If you want to be *specific* for Intel only, then keep it as is.

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
schaefi: done

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
schaefi: is has no impact on the rendered results but yes :) Thanks and done

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
schaefi: done

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
schaefi: Thanks, I like it

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
schaefi: Yeah sounds much better

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
schaefi: done

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
schaefi: I rephrased that

doc/source/working_with_images/iso_to_usb_stick_grub2_boot_from_iso.rst
schaefi: I rephrased that

schaefi:
@tomschr Thanks much for your review time. much appreciated :+1:

PR #2943, opened 2026-02-02
Make error logged/raised from has_option_in_help more accurate

AdamWill:
If I'm following this code correctly, we're not raising here because we "could not parse" the output. We're raising because we parsed it and it did not have the flag we were looking for.

### DIFF
--- a/kiwi/utils/command_capabilities.py
+++ b/kiwi/utils/command_capabilities.py
@@ -69,7 +69,7 @@ class CommandCapabilities:
         for line in command.error.splitlines():
             if flag in line:
                 return True
-        message = 'Could not parse {} output'.format(call)
+        message = 'Could not find flag {} in {} output'.format(flag, call)
         if raise_on_error:
             raise KiwiCommandCapabilitiesError(message)
         if not silent:


AdamWill:
Hmm, so there's history here. Before https://github.com/OSInside/kiwi/pull/2159 , this did behave as the message implies - the exception was only raised if the command exited non-zero. However, in #2159 @schaefi changed it, on the basis that some commands might return non-zero when printing a usage message. He kept the exception raising behavior around but changed it to "raise an exception if the argument isn't present", not "raise an exception if the command exited non-zero"; he updated the docstring to describe this accurately, but not the exception message itself.

I'm not sure if the behavior is actually useful as it currently exists? Maybe we should just get rid of it? Only one thing sets this to True, and it appears to be a bug - https://github.com/OSInside/kiwi/pull/2944 changes it to False.

schaefi:
@AdamWill Thanks I like the new message :+1:

schaefi:
> changed it, on the basis that some commands might return non-zero when printing a usage message. He kept the exception raising behavior around but changed it to "raise an exception if the argument isn't present", not "raise an exception if the command exited non-zero";

right I remember that `setfiles` behaved this way. As kiwi's Command instance raises an exception if the command returns with a non zero exit code we only had the chance to let it go and parse whatever we get on any channel. This lousy style allowed us to parse some output but also does not improve the message on real errors. In addition to your change I also added #2945

PR #2944, opened 2026-02-03
initrd format detection: make dracut --printconfig optional

AdamWill:
In https://github.com/OSInside/kiwi/pull/2921 , based on the commit message and the fact that the code uses an `if`, it's pretty clear the use of `dracut --printconfig` is meant to be optional. But setting the misleadingly-named
`raise_on_error` to `True` means it isn't. `raise_on_error` makes `has_option_in_help` raise an exception if it *fails* - doesn't find the specified flag - not if it *errors*. Since it does not handle the exception, this code crashes if dracut does not have the argument:
https://koji.fedoraproject.org/koji/taskinfo?taskID=141808928

DEBUG util.py:459:  [ ERROR   ]: 06:03:28 | KiwiCommandCapabilitiesError: Could not parse dracut output

I've sent https://github.com/OSInside/kiwi/pull/2943 to make the message less of a lie, and this makes us stop crashing if the arg isn't available.

### DIFF
--- a/kiwi/boot/image/base.py
+++ b/kiwi/boot/image/base.py
@@ -545,7 +545,7 @@ class BootImageBase:
     ) -> Optional[str]:
         if CommandCapabilities.has_option_in_help(
             'dracut', '--printconfig', ['--help'],
-            root=self.boot_root_directory, raise_on_error=True
+            root=self.boot_root_directory, raise_on_error=False
         ):
             dracut_call = Command.run(
                 ['chroot', self.boot_root_directory, 'dracut', '--printconfig']
--- a/test/unit/boot/image/base_test.py
+++ b/test/unit/boot/image/base_test.py
@@ -247,7 +247,7 @@ class TestBootImageBase:
             )
             mock_CommandCapabilities_has_option_in_help.assert_called_once_with(
                 'dracut', '--printconfig', ['--help'],
-                root='system-directory', raise_on_error=True
+                root='system-directory', raise_on_error=False
             )
             mock_Command_run.assert_called_once_with(
                 ['chroot', 'system-directory', 'dracut', '--printconfig']


Conan-Kudo (approved):

schaefi:
@AdamWill  @Conan-Kudo There was a reason why `raise_on_error=True` was set. As you can see from the code we try to detect the initrd output file format in three stages

* _get_boot_image_output_file_format_from_existing_file  (from an existing initrd file)
* _get_boot_image_output_file_format_from_dracut_code (by parsing the dracut implementation)
* _get_boot_image_output_file_format_from_config (by reading the config)

The reason why we wanted to raise on the last check is because if the first two has failed we expect a dracut version that supports --printconfig. If this is not the case we fail and that was intentional. If you are not using a dracut version that has --printconfig it is expected that either check 1 or check 2 is successful.

> KiwiCommandCapabilitiesError: Could not parse dracut output

In my opinion you can only see this message if dracut is not installed. I agree the message is poor and that should be fixed

However, shadowing the problem is imho not the right solution unless I misunderstood that we want to raise if none of the detection mechanics found the format

?

AdamWill:
> @AdamWill @Conan-Kudo There was a reason why `raise_on_error=True` was set. As you can see from the code we try to detect the initrd output file format in three stages
> 
>     * _get_boot_image_output_file_format_from_existing_file  (from an existing initrd file)
> 
>     * _get_boot_image_output_file_format_from_dracut_code (by parsing the dracut implementation)
> 
>     * _get_boot_image_output_file_format_from_config (by reading the config)
> 
> 
> The reason why we wanted to raise on the last check is because if the first two has failed we expect a dracut version that supports --printconfig. If this is not the case we fail and that was intentional. If you are not using a dracut version that has --printconfig it is expected that either check 1 or check 2 is successful.

This is definitely not the case, though. I did check that the previous day's successful build printed the warning, indicating that neither check 1 nor 2 had worked.

> > KiwiCommandCapabilitiesError: Could not parse dracut output
> 
> In my opinion you can only see this message if dracut is not installed. I agree the message is poor and that should be fixed

That's not the case. Currently in Rawhide:
```
[adamw@omnibook ~]$ rpm -q dracut
dracut-108-5.fc44.x86_64
[adamw@omnibook ~]$ dracut --help | grep print
  --print-cmdline       Print the kernel command line for the given disk layout.
  --printsize           Print out the module install size.
```

> However, shadowing the problem is imho not the right solution unless I misunderstood that we want to raise if none of the detection mechanics found the format

If this was your intention, why have the `if` statement? In fact, why call `has_option_in_help` at all? If your expectation was that, if we're in `_get_boot_image_output_file_format_from_config` then dracut *must* have that option or things are going to fail, why not just skip the `if`, skip the `has_option_in_help` statement, and go right to calling dracut?

Also, if that was your intention, why keep the `log.warning('Could not detect dracut output file format')` block that will never be encountered any more?

PR #2945, opened 2026-02-03
Improve error message from CommandCapabilities

schaefi:
In case CommandCapabilities is called from a command that does not exist and raise_on_error is requested make sure to provide a better error message pointing out that the command does not exist instead of the misleading "Could not parse ... output" error message


schaefi:
This is related to my findings when e.g. building an image that has no dracut installed but the image requires it

PR #2979, opened 2026-04-02
Refactor use of kiwi settings file

schaefi:
KIWI supports an optional runtime configuration file for settings to control the behavior of the tools used by    KIWI on the build host. So far this config file could be specified via --config or is searched in the user's HOME    or looked up as /etc/kiwi.yml. With this commit the  following changes to this heuristic are made:
    
1. Support reading of `/etc/kiwi.yml.d/*.yml`
2. Support reading of `/usr/share/kiwi/kiwi.yml` and `/usr/share/kiwi/kiwi.yml.d/*.yml`
3. Install default settings file as `/usr/share/kiwi/kiwi.yml.example` from the main package and drop it from kiwi-man-pages which  was considered a weird place.

### DIFF
--- a/kiwi/defaults.py
+++ b/kiwi/defaults.py
@@ -94,6 +94,10 @@ RUNTIME_CHECKER_METADATA = '{}/runtime_checker_metadata.yml'.format(
 TEMP_DIR = '/var/tmp'
 LOCAL_CONTAINERS = '/var/tmp/kiwi_containers'
 CUSTOM_RUNTIME_CONFIG_FILE = None
+ETC_RUNTIME_CONFIG_FILE = '/etc/kiwi.yml'
+ETC_RUNTIME_CONFIG_DIR = '/etc/kiwi.yml.d'
+USR_RUNTIME_CONFIG_FILE = '/usr/share/kiwi/kiwi.yml'
+USR_RUNTIME_CONFIG_DIR = '/usr/share/kiwi/kiwi.yml.d'
 PLATFORM_MACHINE = platform.machine()
 EFI_FAT_IMAGE_SIZE = 20
 
--- a/kiwi/runtime_config.py
+++ b/kiwi/runtime_config.py
@@ -43,11 +43,12 @@ class RuntimeConfig:
 
     1. Check for --config provided from the CLI
     2. ~/.config/kiwi/config.yml
-    3. /etc/kiwi.yml
+    3. /etc/kiwi.yml + /etc/kiwi.yml.d/*.yml
+    4. /usr/share/kiwi/kiwi.yml + /usr/share/kiwi/kiwi.yml.d/*.yml
 
     The KIWI runtime configuration file is a yaml formatted file
     containing information to control the behavior of the tools
-    used by KIWI.
+    used by KIWI on the build host.
 
     :param bool reread: reread runtime config
     """
@@ -56,6 +57,7 @@ class RuntimeConfig:
 
         if RUNTIME_CONFIG is None or reread:
             config_file = None
+            config_dir = None
             custom_config_file = defaults.CUSTOM_RUNTIME_CONFIG_FILE
 
             if custom_config_file:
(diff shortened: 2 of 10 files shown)


schaefi:
Uhm, I was wrong in the precedence of usr/etc vs. etc and will fix that...

kiwi/runtime_config.py
```diff
@@ -69,13 +70,29 @@ def __init__(self, reread: bool = False):
                     [self._home_path(), '.config', 'kiwi', 'config.yml']
                 )
             if not config_file or not os.path.exists(config_file):
-                config_file = '/etc/kiwi.yml'
+                config_file = defaults.USR_ETC_RUNTIME_CONFIG_FILE
```
jsrain: The order here is wrong. Distribution-provided configuration should go to /usr/etc and not be changed by the users, who - if needed - put their configuration to /etc (which then only contains what users configured). Therefore, if /etc/kiwi.conf exists, it should be used, otherwise find the bistro-provided default in /usr/etc/.

Conan-Kudo:
Please move `/usr/etc` to `/usr/share/kiwi`. Remember that only SUSE distributions use `/usr/etc` and everyone else forbids it.

So, I propose:

1. `/usr/share/kiwi/kiwi.conf.yml`
2. `/usr/share/kiwi/conf.d/*.yml`
3. Example file as `/usr/share/kiwi/kiwi.conf.yml.example`

kiwi/runtime_config.py
schaefi: order fixed

schaefi:
@Conan-Kudo interesting, yet another aspect I didn't know. Only suse is using `/usr/etc` ... why ? strange

I'm fine with your `usr/share` suggestion

Conan-Kudo:
> @Conan-Kudo interesting, yet another aspect I didn't know. Only suse is using `/usr/etc` ... why ? strange
> 

It's always been forbidden, but at least in Fedora, it's technically infeasible to consider because rpm-ostree reserves the directory for storing pristine copies of `/etc` content it gets from RPMs. It does a dpkg-style divert for all of `/etc` into `/usr/etc` and then clones it into `/etc` for three-way merges.

kiwi/defaults.py
```diff
@@ -94,6 +94,10 @@
 TEMP_DIR = '/var/tmp'
 LOCAL_CONTAINERS = '/var/tmp/kiwi_containers'
 CUSTOM_RUNTIME_CONFIG_FILE = None
+ETC_RUNTIME_CONFIG_FILE = '/etc/kiwi.yml'
+ETC_RUNTIME_CONFIG_DIR = '/etc/kiwi.d'
+USR_RUNTIME_CONFIG_FILE = '/usr/share/kiwi/kiwi.yml'
+USR_RUNTIME_CONFIG_DIR = '/usr/share/kiwi/kiwi.d'
```
Conan-Kudo: Let's use `/usr/share/kiwi/kiwi.yml.d` so that the directory name tells you what the config format is. This also opens the door to us switching away from YAML to something more reasonable in the future (like TOML).

Conan-Kudo (changes requested):

kiwi/defaults.py
schaefi: ok, done

package/python-kiwi-spec-template
```diff
@@ -803,6 +803,9 @@ fi
 %files -n python%{python3_pkgversion}-kiwi
 %dir %{_defaultdocdir}/python-kiwi
 %dir %{_usr}/share/kiwi
+%dir %{_usr}/share/kiwi/kiwi.yml.d
+%dir %_sysconfdir/kiwi.yml.d
+%config %{_usr}/share/kiwi/kiwi.yml.example
```
Conan-Kudo: Instead of `%config`, it should be `%doc`.

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: :facepalm: oh man yes, thanks for finding

schaefi: fixed

Conan-Kudo (approved):

Vogtinator:
> @Conan-Kudo interesting, yet another aspect I didn't know. Only suse is using /usr/etc ... why ? strange

To elaborate: It's the downstream way of dealing with projects that don't support /usr/share/ or /usr/lib/... on their own.

Conan-Kudo:
Well, it was done despite people saying that path was a bad idea.

PR #2989, opened 2026-04-24
Update documentation for devicepersistency

schaefi:
Setting the devicepersistency also impacts an eventually created install image. This information was missing in the documentation.


jsrain (approved):
Thanks for this update, clarifies the behaviour I was experiencing a lot!

PR #2990, opened 2026-05-06
Do not hardcode dracut omit module in live builder

schaefi:
This is a relict from the old days when the pure presence of multipath in a live ISO caused issues at boot time. kiwi should not maintain a hardcoded list of dracut modules except for those that are mandatory for live boot, if there should be any special setting needed it should come from an overlay setup in /etc/dracut.conf.d/*.conf as part of the image description. This is related to bsc#1260340

### DIFF
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -473,11 +473,9 @@ class LiveImageBuilder:
         ) + modules
         for dracut_module in live_dracut_modules:
             self.boot_image.include_module(dracut_module)
-        self.boot_image.omit_module('multipath')
         self.boot_image.write_system_config_file(
             config={
-                'modules': live_dracut_modules,
-                'omit_modules': ['multipath']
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
What's the bug about? I can't access it.

schaefi:
> What's the bug about? I can't access it.

@Conan-Kudo Ah sorry forgot that. The bug is with regards to the DMS - Distribution Migration System (https://github.com/SUSE/suse-migration-services). That's a suse specific upgrade system from one major release to another. A live system to drive this process is used for this purpose. During testing the people found out that multipath based systems can't upgrade because dracut's isoscan did not find any root device. Reason was that the kiwi live builder has a hard coded omit line for multipath. This should not be done in my eyes because as a user you have no chance to influence this now. This PR drops the hard coded omit and leaves it in the hands of the user how to design the live system.

Conan-Kudo (approved):

PR #2991, opened 2026-05-06
Relax check_checkmedia_used_with_msdos_table

schaefi:
Relax this runtime check to only apply for tagmedia/checkmedia versions < v6.6. Newer versions also supports GPT. This Fixes bsc#1262229.


kiwi/runtime_checker.py
Vogtinator: Comment is outdated now

schaefi: yeah thanks, forgot to update the comment. Done

Issue #2996, opened 2026-05-14
StringToSize.to_bytes() returns float instead of int when size has m/g unit


officialasishkumar:
## Problem description

`kiwi.utils.size.StringToSize.to_bytes()` is documented to return an `int`, and downstream callers (e.g. `XMLState.get_build_type_unpartitioned_bytes`) declare a `-> int` return type, but the function actually returns a `float` whenever the size string ends in an `m`/`M` or `g`/`G` suffix.

The cause is the use of `math.pow(1024, n)`, which always returns a `float`:

```python
return size_base * math.pow(1024, size_unit) if size_unit \
    else size_base
```

Reproducer:

```python
>>> from kiwi.utils.size import StringToSize
>>> StringToSize.to_bytes('100m')
104857600.0
>>> type(StringToSize.to_bytes('100m'))
<class 'float'>
>>> StringToSize.to_bytes('1234')
1234
>>> type(StringToSize.to_bytes('1234'))
<class 'int'>
```

This float value then leaks into places that format it as a shell argument or log line. For example, `DiskBuilder.append_unpartitioned_space()` passes the value (originating from `<size unpartitioned="100" unit="M">`) into `DiskFormatBase.resize_raw_disk(..., append=True)`, which builds the `qemu-img resize` argument with:

```python
'+{0}'.format(size_bytes) if append else format(size_bytes)
```

producing a `+104857600.0` argument (with a trailing `.0`). The same happens in `image_resize` log output:

```
Resizing raw disk to 1073741824.0 bytes
```

The non-append branch of `resize_raw_disk` happens to call `int(size_bytes)` before formatting, so it works by accident, but the type contract is still violated and the value is mistyped end-to-end.

## Expected behaviour

`StringToSize.to_bytes()` should return an `int` for every supported input, matching its documented `:rtype: int`. Downstream callers annotated `-> int` should not silently return a `float`.

## Steps to reproduce the behaviour

```python
>>> from kiwi.utils.size import StringToSize
>>> isinstance(StringToSize.to_bytes('1g'), int)
False
```

## OS and Software information

* KIWI version: 10.3.4 (current `main`)
* Operating system host version: N/A (logic bug, reproducible on any host)
* Operating system target version: N/A
* Open Build Service version (N/A if not using OBS): N/A
* Koji version (N/A if not using Koji): N/A

PR #2997, opened 2026-05-14
Return integer byte values from StringToSize.to_bytes()

officialasishkumar:
## Summary

- `kiwi.utils.size.StringToSize.to_bytes()` was documented to return an `int`, but used `math.pow(1024, n)` to scale the unit suffix, which always returns a `float`. Every call with an `m`/`g` suffix returned a `float` and that float propagated to callers whose return type was annotated `-> int` (e.g. `XMLState.get_build_type_unpartitioned_bytes`).
- The float value reached `DiskFormatBase.resize_raw_disk(..., append=True)`, which formats the value directly into the `qemu-img resize` argument (`'+{0}'.format(size_bytes)`), producing arguments with a trailing `.0` (e.g. `+104857600.0`). The log output of `image_resize` had the same issue.
- Replaces `math.pow(1024, n)` with the integer expression `1024 ** n`, adds proper type annotations (`size_value: str` -> `int`), and asserts `isinstance(..., int)` in the unit tests for both unitless and unit-suffixed inputs.

Fixes Issue#2996

## Test plan

- [x] `pytest test/unit/utils/size_test.py` &rarr; 3 passed
- [x] `pytest test/unit/xml_state_test.py test/unit/tasks/image_resize_test.py test/unit/builder/disk_test.py test/unit/runtime_config_test.py test/unit/filesystem/setup_test.py` &rarr; 205 passed
- [x] `mypy kiwi` &rarr; no issues found in 213 source files
- [x] Manual sanity check confirms `StringToSize.to_bytes('100m')` now returns `104857600` (`int`) and the formatted append argument is `+104857600` rather than `+104857600.0`

### DIFF
--- a/kiwi/utils/size.py
+++ b/kiwi/utils/size.py
@@ -16,7 +16,6 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import re
-import math
 
 from kiwi.exceptions import KiwiSizeError
 
@@ -26,7 +25,7 @@ class StringToSize:
     **Performs size convertions from strings to numbers**
     """
     @staticmethod
-    def to_bytes(size_value):
+    def to_bytes(size_value: str) -> int:
         """
         Convert the given string representig a size into the appropriate
         number of bytes.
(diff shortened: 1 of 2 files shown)


schaefi (approved):
Thanks much, good catch 👍

schaefi:
This Fixes #2996 correct ? 

@officialasishkumar if so, maybe you can add that to the commit message, so the associated issue gets auto closed when we merge this one. Thank you

schaefi:
Hmm, just saw you have a reference in the commit message, forget what I wrote :facepalm:

PR #2999, opened 2026-05-16
Refactor bootloader video mode setting

### LINKED ISSUE #2998: Add support for 1920x1080x32 resolution in defaults.py
I am reporting this issue on behalf of my development team. This bug report was filed by Gemini 3.1 at my request.

It appears that kiwi does not currently support the 1920x1080x32 resolution in its configuration defaults, which is problematic for modern high-resolution displays (e.g., 4K).

The following files were identified as likely needing updates to incorporate this resolution:
- kiwi/defaults.py (Likely location for core defaults)
- grub.cfg.disk.iso-template
- grub.cfg.iso-template
- root/etc/grub.d/01_nortk_theme
- scripts/fix-efi-stub.sh

Proposed fix:
1. Update kiwi/defaults.py to recognize or include 1920x1080x32 in its supported resolutions.
2. Update the identified grub template files to include '1920x1080x32' in the gfxmode strings, ensuring it is a preferred fallback for modern hardware.

schaefi:
The bootloader interface supported a generic video mode setting based on VESA VBE hex mode numbers. This value could be set via the `vga=` kernel parameter and is obsolete since a long time.

kiwi did not set the vga value since it is deprecated but it still used an internal table mapping video mode settings per bootloader using the VBE hex numbers as matching key. This commit now finally drops all of this mapping table and the obsolete vga attribute from the type section and adds the new attribute:

```xml
<bootloader ... video_mode="loader_specific_value"/>
```

alongside with all other bootloader specific settings. As of today the video_mode setting is evaluated by the grub bootloader only but can be extended to other loaders in the future.

Please note there is by intention no automatic translation from `<type vga="hex_value">` into `<bootloader video_mode="loader_specific_value">` because I would like to prevent to maintain custom mapping tables for bootloader settings. This concept did not pay off in the past for other bootloader settings. An entry in the documentation has been created to support users with the proper value according to the used bootloader.

This Fixes #2998


schaefi:
The setting will be integration build tested. Please wait with a merge for
* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-agama

which is the test build where I applied it. The expected video mode should be 1024x768

schaefi:
Integration test showed the expected video mode for the ISO and the Disk

PR #3001, opened 2026-05-20
Fixed partition name validation

schaefi:
The method get_partition_name returns a default name value in case a partition name cannot be retrieved, for example if the partition table does not support this value. In this case the default name was 'unsupported'. However there is a naming policy in kiwi for partition names to match 'p:name' This policy is expected to be true and applied at some places for example in create_dasd_partitions. If the default name is in use which is the case on DASD devices because no partition names exists, then the non policy conform default name causes the method create_dasd_partitions to create an invalid partition setup instructions file. This results in the partition resize to fail. This commit fixes the default name to match the policy as 'p:unsupported'


Bischoff:
Tested to work on ECKD image.

Issue #2804, opened 2025-05-13
Replace usage of gdisk with sfdisk


Conan-Kudo:
We currently use `gdisk` for GPT disks due to a legacy in which GPT was only supported by `gdisk`, but `fdisk` and `sfdisk` from util-linux has supported GPT in addition to MBR since 2013 with util-linux 2.23. We no longer support any distribution that doesn't have a version of fdisk that doesn't support GPT, so let's move to `sfdisk` and drop the `gdisk`/`gptfdisk` dependency.

Conan-Kudo:
Also note, `gdisk` not available in RHEL 10 either, and RHEL users are expected to use `fdisk`/`sfdisk` instead.

Conan-Kudo:
We may want to also migrate our `fdisk` usage to `sfdisk` too, to simplify how we interact with partitioning.

Conan-Kudo:
This is still planned to be done.

Conan-Kudo:
This is still planned.

PR #3002, opened 2026-05-31
Drop parted requirement for msdos partitioner

schaefi:
parted was used in kiwi at one place to set the active flag on a partition in the DOS table. This action can also be done via sfdisk and drops the parted requirement from the kiwi builder code

### DIFF
--- a/kiwi/partitioner/msdos.py
+++ b/kiwi/partitioner/msdos.py
@@ -108,12 +108,11 @@ class PartitionerMsDos(PartitionerBase):
             if flag_name == 'f.active':
                 Command.run(
                     [
-                        'parted', self.disk_device,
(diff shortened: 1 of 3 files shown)


schaefi:
@Conan-Kudo as a result of the ongoing conversation #3000 I still believe we should free ourselves from parted. At least in the main python builder code this can be done in my opinion. That's why I opened this PR.

I would love if we could get rid of parted and partprobe on s390 too, but I need to think more about it because there is coded added to parted by suse only to circumvent issues in `fdasd` which should be the only tool needed for s390 DASD operations. As such it's hard

Conan-Kudo:
We also have a ticket to get rid of gdisk for sfdisk too (#2804), so yes in principle I think we should consolidate on sfdisk or parted for everything.

PR #3005, opened 2026-06-12
Fixed grub vendor directory lookup

### LINKED ISSUE #3003: Bootloader EFI discovery fails on RHEL 10.2 due to multiline string output
**Description:**
When building a RHEL 10.2 image, the build fails during the bootloader configuration step with an [Errno 22] Invalid argument error when Kiwi attempts to create the EFI vendor directory.

**To Reproduce:**
Build a RHEL 10.2 image using grub2 as the bootloader.
The build crashes during the bootloader setup phase.

**Expected behavior:**
The EFI vendor directory should be correctly identified and created without throwing a path error.

**Error Log:**
`[ ERROR   ]: 13:36:38 | KiwiFileAccessError: Cannot create directory: /var/tmp/kiwi_mount_manager.5s3bx8fw/boot/efi/EFI/Linux`
`/EFI/redhat: [Errno 22] Invalid argument: '/var/tmp/kiwi_mount_manager.5s3bx8fw/boot/efi/EFI/Linux\n'`

**Root Cause Analysis:**
In RHEL 10.2, the compiled grubx64.efi binary contains multiple EFI paths (e.g., EFI/Linux for UKI/systemd-boot and EFI/redhat for classic grub).

In kiwi/bootloader/config/grub2.py, within the _copy_grub_config_to_efi_path method, Kiwi uses the strings command and grep EFI/ to dynamically find the vendor path. Because the binary contains multiple matches, the command returns a multiline string:
EFI/Linux\nEFI/redhat

The current code applies .strip(), but this only removes leading/trailing whitespaces, leaving the newline character \n in the middle of the string. When os.path.normpath tries to construct the path, it results in an invalid directory path.

**Workaround / Proposed Fix:**
I successfully bypassed this issue locally by implementing a workaround that splits the multiline string and grabs the last element before joining the path. In grub2.py (_copy_grub_config_to_efi_path), I applied the following:

                    efi_boot_path = Command.run(
                        ['bash', '-c', ' '.join(bash_command)],
                        raise_on_error=False, raise_on_command_not_found=True
                    ).output
                    if efi_boot_path:
                        # Extract the final EFI path if multiple lines are returned
                        if '\n' in efi_boot_path:
                            efi_boot_path = efi_boot_path.strip().split('\n')[-1]

                        efi_boot_path = os.path.normpath(
                            os.sep.join([root_path, efi_boot_path.strip()])
                        )

This workaround successfully resolved the build error for me. However, I am not entirely sure if this is the cleanest or most robust way to solve this upstream, so I would appreciate your input on how to properly handle this scenario in the codebase.

**Environment:**
Host OS: Red Hat Enterprise Linux 10.2 (Coughlan)
Target OS: RHEL 10.2
KIWI version: 10.3.0

schaefi:
If the grub vendor directory cannot be found by the static list of directories used to lookup the shim binary, kiwi also checks the encoded boot paths from the distribution provided grub binary. This information can be a multiline string holding more than one path. With this commit we support multiple vendor directories. This Fixes #3003

### DIFF
--- a/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
+++ b/build-tests/x86/fedora/test-image-live-disk/appliance.kiwi
@@ -13,7 +13,7 @@
     <profiles>
         <profile name="Live" description="Live image of Fedora"/>
         <profile name="Virtual" description="Virtual image of Fedora"/>
-        <profile name="Virtual_bootpart" description="Virtual image of Fedora with extra boot partition"/>
+        <profile name="Virtual_bootpart" description="Virtual image of Fedora with extra boot partition and no shim"/>
         <profile name="Disk" description="OEM image of Fedora"/>
     </profiles>
     <preferences>
@@ -126,12 +126,14 @@
         <package name="dracut-kiwi-oem-repart"/>
         <package name="dracut-kiwi-oem-dump"/>
     </packages>
+    <packages type="bootstrap" profiles="Live,Disk,Virtual">
+        <package name="shim" arch="x86_64"/>
+    </packages>
     <packages type="bootstrap">
         <package name="filesystem"/>
         <package name="basesystem"/>
         <package name="grub2-efi-x64-modules"/>
         <package name="grub2-efi-x64"/>
-        <package name="shim" arch="x86_64"/>
         <package name="fedora-release"/>
         <package name="shadow"/>
     </packages>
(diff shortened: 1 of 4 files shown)


Conan-Kudo:
What do we do with multiple vendor directories?

schaefi:
> What do we do with multiple vendor directories?

We copy the grub config there. There is an ongoing conversation that I started on the respective issue report. Please feel free to share your thoughts. Thanks

PR #3007, opened 2026-06-19
Add support for partition based installation

schaefi:
Instead of dumping the entire disk also support a mode that dumps each partition individually. Reason for adding this mode is to test features of btrfs like the seed/sprout process which requires access to the partition with the btrfs filesystem. More details here:

*  https://btrfs.readthedocs.io/en/latest/Seeding-device.html

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-disk/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-disk/appliance.kiwi
@@ -22,7 +22,7 @@
         <bootloader-theme>openSUSE</bootloader-theme>
     </preferences>
     <preferences profiles="Standard">
-        <type image="oem" filesystem="btrfs" kernelcmdline="console=ttyS0" firmware="efi" installiso="true" bootpartition="false" btrfs_root_is_snapshot="true" installboot="install">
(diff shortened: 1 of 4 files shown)


Conan-Kudo:
sproud->sprout

schaefi:
@Conan-Kudo 

I completed the support for btrfs seed/sprout feature and ran several tests to see if this brings a significant performance gain. So far I could not see a significant performance gain. If you look at the code you see that I modified the disk builder to automatically enable seeding on the btrfs device if an install image is requested. The respective installer code then just maps the partitions (kpartx) runs seed/sprout procedure for the btrfs partition and the standard dump procedure for all others (EFI, etc ...). This works but is not significantly faster than dumping the entire disk image as one blob to the disk device.

I also added support for remote seeding. This is done in a way that the disk image first gets fetched as file (/image.raw) into the initramfs (ram space). From there the same procedure as described below is applied. In this mode I even see a performance loss due to the extra local store prior target dump. In addition this mode needs more resources (preserve RAM space in the initrd to store the disk image from the remote location)

Overall things are working but I don't see an improvement that legitimates the extra code so far.

Review from the experts is very much welcome.

To be honest I haven't expected a big performance win, somehow the blocks needs to be transferred to the target block device which is empty to begin with :) 

Thanks this was a nice learning opportunity though.

I keep this open as a draft

schaefi:
We are also loosing target verification against the installed image blob because the seed/sprout transfer is not bit identical with the image

PR #3010, opened 2026-06-29
Fix qemu-tools requirement

schaefi:
Disable building 32-bit QEMU support (bsc#1263564)

### DIFF
--- a/package/python-kiwi-spec-template
+++ b/package/python-kiwi-spec-template
@@ -326,6 +326,7 @@ Requires:       squashfs
 %else
 Requires:       squashfs-tools
 %endif
+%ifnarch %{ix86} %{arm}
 %if "%{_vendor}" == "debbuild"
 Requires:       qemu-utils
 %else
@@ -333,6 +334,7 @@ Requires:       qemu-utils
 Requires:       qemu-tools
 %else
 Requires:       qemu-img
+%endif
 %endif
 %endif
 Requires:       kiwi-systemdeps-core = %{version}-%{release}


package/python-kiwi-spec-template
```diff
@@ -330,7 +330,10 @@ Requires:       squashfs-tools
 Requires:       qemu-utils
 %else
 %if 0%{?suse_version}
+#Disable building 32-bit QEMU support (bsc#1263564)
+%ifnarch %{ix86} %{arm}
```
Conan-Kudo: We need to do this for all the variants of the qemu package.

Conan-Kudo (changes requested):

package/python-kiwi-spec-template
schaefi: @Conan-Kudo , Ana is currently on vacation, so I respond to this one. You mean the arch condition should also apply for `qemu-img` and `qemu-utils` as used on non suse and debian ?

schaefi: @Conan-Kudo I think I covered it, can you please double check? Thanks

Conan-Kudo (approved):

package/python-kiwi-spec-template
Conan-Kudo: Yep that's what I mean.

ana:
Thank you, can we this submitted to openSUSE Factory ASAP? see https://bugzilla.opensuse.org/show_bug.cgi?id=1263564

PR #3011, opened 2026-06-29
Fix dnf5 handling for patternType attribute

### LINKED ISSUE #3008: No way to only include "mandatory" packages from `namedCollection`s with dnf5
Hi, I was messing with `kiwi-ng` to get custom Fedora live images with various environments, for testing. Since I just need to run some stuff, I figured I could probably trim the thing enough to fit on a DVD5.

I started with this (which works):

```xml
	<packages type="image" patternType="plusRecommended" profiles="Test-Desktop">
		<namedCollection name="kde-desktop-environment"/>
		<namedCollection name="workstation-product-environment"/>
		<namedCollection name="lxqt-desktop-environment"/>
		<namedCollection name="cinnamon-desktop-environment"/>
		<namedCollection name="mate-desktop-environment"/>
		<namedCollection name="budgie-desktop-environment"/>
		<namedCollection name="cosmic-desktop-environment"/>
		<namedCollection name="miraclewm-desktop-environment"/>

		<!-- This stuff conflicts with sddm-wayland-sway. I prefer to keep that one,
		if it even matters. -->
		<ignore name="sddm-wayland-miraclewm"/>
		<ignore name="sddm-wayland-miriway"/>

		<namedCollection name="sway-desktop-environment"/>
	</packages>
```

And replaced `plusRecommended` with `onlyRequired`. Unfortunately this resulted in basically no difference. There was a slight (~1MB) difference, but I'm not sure why.

Turns out that groups (and environments which are just groups AFAICT) have a more complicated way to sort packages. Usually there's "mandatory", "default", and "optional" groups. The right flag to pass is `--setopt=group_package_types="mandatory"` (see https://dnf.readthedocs.io/en/latest/conf_ref.html ). I think it could be forced globally but I think that having a way to set the preference on a per-`packages` basis would be nifty. Right now the only work around I can think of is just declaring all mandatory groups directly.

I am still learning how to use this nice tool, so please don't hesitate to point out any mistakes I did. I also tried to contact you all on Matrix but unfortunately the link in the docs is broken :(

schaefi:
The patternType attribute is used to control the mandatory vs. optional installation for the concept of patterns,groups aka collections for the respective package manager. In dnf5 this is also controlled by the group_package_types option which is used now in combination with install_weak_deps. This Fixes #3008

### DIFF
--- a/kiwi/package_manager/dnf5.py
+++ b/kiwi/package_manager/dnf5.py
@@ -280,7 +280,7 @@ class PackageManagerDnf5(PackageManagerBase):
         """
         self.__clean_pattern_type()
         self.custom_args.append('--setopt=install_weak_deps=False')
-        self.custom_args.append('--setopt=group_package_types=default,mandatory')
+        self.custom_args.append('--setopt=group_package_types=mandatory')
 
     def process_plus_recommended(self) -> None:
         """
@@ -288,7 +288,7 @@ class PackageManagerDnf5(PackageManagerBase):
         """
         self.__clean_pattern_type()
         self.custom_args.append('--setopt=install_weak_deps=True')
-        self.custom_args.append('--setopt=group_package_types=default,mandatory,optional')
+        self.custom_args.append('--setopt=group_package_types=default,mandatory,conditional')
 
     def match_package_installed(
         self, package_name: str, package_manager_output: str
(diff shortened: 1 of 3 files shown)


schaefi:
@Conan-Kudo it would be nice to get someone from the DNF team to review this as well

kiwi/package_manager/dnf5.py
```diff
@@ -278,15 +278,17 @@ def process_only_required(self) -> None:
...
 
     def process_plus_recommended(self) -> None:
         """
         Setup package processing to also include recommended dependencies.
         """
-        if '--setopt=install_weak_deps=False' in self.custom_args:
-            self.custom_args.remove('--setopt=install_weak_deps=False')
+        self.__clean_pattern_type()
+        self.custom_args.append('--setopt=install_weak_deps=True')
+        self.custom_args.append('--setopt=group_package_types=default,mandatory,optional')
```
deralmas: FTR, [the docs](https://dnf5.readthedocs.io/en/latest/dnf5.conf.5.html) say that the default is `default,mandatory,conditional`. Looks like the `optional` part is not installed by default.

Take this with a grain of salt, but I think that the "onlyRequired" pattern should just be `mandatory`, while the "plusRecommended" pattern would have `default,mandatory,conditional` as per default.

For example, `kde-desktop-environment` puts `kde-apps` in the `default` tier, with the compositor and everything essential in the `mandatory` tier.

schaefi: yeah makes sense. I updated as you suggested and also updated one of our integration tests. I'd like to see this in action before we merge and will let it build through Staging once we get an approval on the changes

Conan-Kudo (approved):

PR #3012, opened 2026-07-01
Add support for sha512sum

### LINKED ISSUE #3004: Use sha512 as the new default for image checksums
Hello

product-composer 0.9.X has changed its defaults to produce openSUSE install images with sha512 checksum files.
openSUSE Leap 16.1 has currently online + offline images with sha512 checksum files and all appliances images built directly by kiwi use 256 sum.

I belive it would be good opportunity to update default checkum files for images built by kiwi from sha256 to sha512 as well.

schaefi:
Allow to select between the default shasum size 256 and 512 via a new runtime config section named: shasum
This Fixes #3004

### DIFF
--- a/build-tests/x86/fedora/kiwi-settings/kiwi-settings/kiwi.yml
+++ b/build-tests/x86/fedora/kiwi-settings/kiwi-settings/kiwi.yml
@@ -2,3 +2,5 @@ iso:
   - media_tag_tool: isomd5sum
 oci:
   - archive_tool: buildah
+bundle:
+  - shasum_size: "512"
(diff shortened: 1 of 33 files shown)


schaefi:
Needs testing which is not yet completed

- [x] Also one open question if the setting for the shasum size should be individual for the bundler and the builder or can stay as a global to all option

schaefi:
Summarizing the changes from a user perspective:

In the kiwi runtime config file e.g. `/etc/kiwi.yml` it is now possible to specify a system wide shasum size as follows

```yaml
shasum:
  - size: "512"
```

This will change the calculation of any shasum in any procedure in kiwi that needs it to `sha512sum`. In the initrd code a new env variable `kiwi_shasum_suffix` provides information to let the initrd code chose the right checksum tool.

If only the result bundler should use a different shasum size than the default or an overwrite for a system wide setting is needed, the following new option in the `bundle` section can be set, either exclusively or in addition to the above

```yaml
bundle:
  - shasum_size: "512"
```

Conan-Kudo (approved):
We should switch the default to 512 soon, but otherwise, LGTM.

PR #3013, opened 2026-07-01
Fix system resize

schaefi:
When resizing the partition table via sfdisk --relocate, the resize failed when the device was locked beforehand. In addition the resize of the actual filesystem failed the prior filesystem check if the btrfs filesystem is used. btrfs reports the free space cache to be outdated/corrupt after kiwi has resized the partition table. This is an expected condition prior resize of the filesystem and should not prevent the following resize. As such the filesystem check is not applied for btrfs prior resize.

### DIFF
--- a/dracut/modules.d/59kiwi-lib/kiwi-filesystem-lib.sh
+++ b/dracut/modules.d/59kiwi-lib/kiwi-filesystem-lib.sh
@@ -62,7 +62,14 @@ function resize_filesystem {
     ;;
     esac
     if ! _is_ramdisk_device "${device}"; then
-        check_filesystem "${device}"
+        if [ ! "${fstype}" = "btrfs" ];then
+            # btrfs reports the free space cache to be outdated/corrupt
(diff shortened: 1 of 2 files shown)


Conan-Kudo (changes requested):
Why aren't we just clearing the space cache instead of leaving it broken?

schaefi:
> Why aren't we just clearing the space cache instead of leaving it broken?

@Conan-Kudo Because this would be a superfluous action. The check operation happens before the resize because some filesystems requires a check before they actually allow a resize (extX for example). This is not the case for btrfs. After we have changed the partition geometry the subsequent `btrfs filesystem resize max` does a good job and of course also fixes the space cache. That's why it's really not needed to mess with a filesystem prior the resize.

Also this is a behavior change in btrfs, former versions did not report free space in the cache as an error.

Conan-Kudo:
Yes, the change is because of the space cache v2 transition.

Conan-Kudo (approved):

PR #3015, opened 2026-07-01
Fix reading of runtime config files

### LINKED ISSUE #3014: Reading of kiwi.yml.d is incompatible with systemd/uapi spec
While 9dc954178a3eccee1c506a7b9d199a42f642f894 introduced support for a kiwi.yml.d directory for drop-ins, it has different behaviour than expected from other applications, including systemd, zypper, util-linux, anything using libeconf, etc. and the current implementation makes it much less useful:

The main difference is that drop-in files must always be read, no matter if the main file exists.

e.g. if you have

`/usr/share/kiwi.yml` (vendor-provided)
`/etc/kiwi.yml.d/foo.yml` (admin-provided),

both need to be read. This is the main reason for the `/etc` and `/usr` separation. Currently kiwi will only read the latter if `/etc/kiwi.yml` exists, but then it will no longer read the vendor-provided drop-ins.

Ideally kiwi should also support to read drop-ins in `yml.d` without the main file needing to be present (or ship `/usr/share/kiwi.yml` itself), otherwise dropins provided by packages would either get ignored (no `kiwi.yml`) or cause conflicts (both have `kiwi.yml`).

See https://uapi-group.org/specifications/specs/configuration_files_specification/#drop-ins

schaefi:
Make sure the reading of the runtime config files in the different locations is not an exclusive procedure. According to the systemd/uapi spec both vendor(/usr) and admin(/etc) places needs to be taken into account and not only one. In addition the drop-ins directories should also be taken into account if their main file is not present.
This Fixes #3014

### DIFF
--- a/kiwi/runtime_config.py
+++ b/kiwi/runtime_config.py
@@ -19,7 +19,7 @@ import os
 import logging
 import functools
 from typing import (
-    Literal, List, Optional, NamedTuple, Callable
+    Literal, List, Optional, NamedTuple, Callable, Dict, Any
 )
 import yaml
 
@@ -36,7 +36,7 @@ from kiwi.exceptions import (
 
 log = logging.getLogger('kiwi')
 
-RUNTIME_CONFIG = None
+RUNTIME_CONFIG: Optional[Dict[str, Any]] = None
 
 
 class ShasumT(NamedTuple):
@@ -46,12 +46,12 @@ class ShasumT(NamedTuple):
 
 class RuntimeConfig:
     """
-    **Implements reading of runtime configuration file:**
+    **Implements reading of runtime configuration files:**
 
-    1. Check for --config provided from the CLI
-    2. ~/.config/kiwi/config.yml
-    3. /etc/kiwi.yml + /etc/kiwi.yml.d/*.yml
-    4. /usr/share/kiwi/kiwi.yml + /usr/share/kiwi/kiwi.yml.d/*.yml
+    1. vendor: /usr/share/kiwi/kiwi.yml + /usr/share/kiwi/kiwi.yml.d/*.yml
+    2. admin: /etc/kiwi.yml + /etc/kiwi.yml.d/*.yml
+    3. ~/.config/kiwi/config.yml
+    4. Check for --config provided from the CLI
 
     The KIWI runtime configuration file is a yaml formatted file
     containing information to control the behavior of the tools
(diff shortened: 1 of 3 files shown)


Conan-Kudo (approved):

Conan-Kudo (changes requested):
Rebase to deal with conflicts, please.

schaefi:
> Rebase to deal with conflicts, please.

done

Conan-Kudo (approved):

Vogtinator:
This is still not entirely according to the UAPI spec, where the various .d dropins have the same priority and files override complete other files, i.e. if `/etc/kiwi.yml` exists, `/usr/share/kiwi/kiwi.yml` is not read at all or if `/etc/kiwi.yml.d/foo.conf` exists, `/usr/share/kiwi/kiwi.yml.d/foo.conf` is not read.

In practice this is usable though, it just needs to be documented as such to avoid surprises.

schaefi:
@Vogtinator Hmm, not sure what you mean by not read at all. The current  implementation collects all files when they exist in the following order:

1. vendor (usr)
2. vendor drop-ins (usr)
3. admin (etc)
4. admin drop-ins (etc)
5. home
6. cmdline

I have tested this by adding files in all possible locations and kiwi reports this as follows

```
kiwi-ng image info --description build-tests/x86/tumbleweed/test-image-disk
[ INFO    ]: 17:53:09 | Reading runtime config file: '/usr/share/kiwi/kiwi.yml'
[ INFO    ]: 17:53:09 | Reading runtime config file: '/usr/share/kiwi/kiwi.yml.d/a.yml'
[ INFO    ]: 17:53:09 | Reading runtime config file: '/etc/kiwi.yml'
[ INFO    ]: 17:53:09 | Reading runtime config file: '/etc/kiwi.yml.d/b.yml'
[ INFO    ]: 17:53:09 | Reading runtime config file: '/home/ms/.config/kiwi/config.yml'
[ INFO    ]: 17:53:09 | Loading XML description: build-tests/x86/tumbleweed/test-image-disk/appliance.kiwi
```

This results in a list of files which are now read in. The read in python dicts are "merged" into one via dict.update().

However, you still have a point. The dict merging strategy has its limits and the way the kiwi runtime config is structured does not allow easy extending of existing configuration files

I will create a follow up PR to make this more flexible

Vogtinator:
> Hmm, not sure what you mean by not read at all.

kiwi reads even the files from `/usr` which should be masked by their `/etc` equivalent.

This creates issues as follows:

`/usr/share/kiwi/kiwi.yml` has

```
foo: bar
asdf: jkl
```

while

`/etc/kiwi.yml` has

```
foo: baz
```

The end result must be

```
foo: baz
```

but currently it will be

```
foo: baz
asdf: jkl
```

This means that some vendor config changes will affect kiwi even if they have been overwritten.

> This results in a list of files which are now read in. The read in python dicts are "merged" into one via dict.update().

IMO that's fine for now, being able to override top-level items separately is already useful, though this could be improved for the future if e.g. the `container` branch gets more keys.

PR #3018, opened 2026-07-03
Add compatibility symlink in live images

schaefi:
The live image layout places the kernel and initrd files for loading from ISO below the directory structure `boot/ARCH/loader/...` Some PXE infrastructure code also reads from `boot/ARCH/...`. To support this infrastructure code we add a compat symlink such that both locations can be used to lookup the kernel and initrd on the ISO. This Fixes bsc#1240633

### DIFF
--- a/kiwi/builder/live.py
+++ b/kiwi/builder/live.py
@@ -562,6 +562,10 @@ class LiveImageBuilder:
         kernel = Kernel(self.boot_image.boot_root_directory)
         if kernel.get_kernel():
             kernel.copy_kernel(boot_path, '/linux')
+            os.symlink(
+                'loader/linux',
+                f'{self.media_dir.name}/boot/{self.arch}/linux'
+            )
         else:
             raise KiwiLiveBootImageError(
                 'No kernel in boot image tree {0} found'.format(
@@ -583,6 +587,10 @@ class LiveImageBuilder:
             os.chmod(self.boot_image.initrd_filename, 0o644)
             shutil.move(
                 self.boot_image.initrd_filename, boot_path + '/initrd'
+            )
+            os.symlink(
+                'loader/initrd',
+                f'{self.media_dir.name}/boot/{self.arch}/initrd'
             )
         else:
             raise KiwiLiveBootImageError(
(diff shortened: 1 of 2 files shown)


hramrach:
Sounds reasonable, thanks.

That said, I am fine with exclusively the new layout, and not sure which tool should be tested that would require the old one.

Conan-Kudo:
Hold up, what? The bug is private so I don't understand what problem we're dealing with here.

Conan-Kudo:
> That said, I am fine with exclusively the new layout, and not sure which tool should be tested that would require the old one.

Yes, I would also prefer not to introduce this if we don't need to.

hramrach:
Historical media (eg. Leap 15.6, corresponding SLE, early 16.0 revisions, ..) have the `linux` and `initrd` outside of the `loader` directory for some architectures (but not all).

Some people probably have some tooling that relies on this bespoke layout but not aware of any specific example.

> SUSE boot media have a long established layout for the relevant boot files:
> 
> x86 has /boot/x86_64/loader/linux|initrd
> not-x86 has /boot/aarch64/linux|initrd (for ARM64 serving as an example)
> 
> As I just had to learn, all SLE16 media broke that established layout:
> 
> files are in /boot/aarch64/loader/linux|initrd

Conan-Kudo:
That breakage was intentional, since people often unpack multiple arch ISOs into the same hierarchy for netboot and expect things to work.

hramrach:
Also this can be handled in the customization script, all media I have seen have one anyway.

hramrach:
https://src.opensuse.org/products/SLES/src/branch/16.1/agama-installer-SLES/fix_bootconfig

schaefi:
@Conan-Kudo sorry for not providing enough details on the background, I always forget that the bugzilla is private :facepalm: @hramrach Thanks for providing the relevant details to have this conversation. I actually opened the PR because the reporter in bugzilla sounded like that addressing the change in the customization script was not an option. 

> That breakage was intentional, since people often unpack multiple arch ISOs into the same hierarchy for netboot and expect things to work.

right which is the reason why we have it arch specific and consistent from the beginning on

> Also this can be handled in the customization script, all media I have seen have one anyway.

This sounds like the better place for me too. Thanks much

So are we in agreement that we don't need the compat link upstream ? If so I would address it as such in the bug and hope for agreement with the reporter on the suse side.

Thoughts ?

schaefi:
ok, seems we agree that this should not be handled by a compat link

PR #3019, opened 2026-07-06
Allow to build kis image without archiving

### LINKED ISSUE #3016: Option to disable building tarball in KIS image type
The KIS builder always creates a tarball containing the built kernel+initrd+image. https://github.com/OSInside/kiwi/blob/c994a8c1a2e0fd015b7990a5d2855296efa57536/kiwi/builder/kis.py#L229

 In some cases this is just a waste of disk space, and also time with the default XZ compression.

schaefi:
By default the files (kernel, initrd, system) plus the metadata information (shasum, append) are all placed into a tar archive and served as one entity. In some cases this archived bundle is unwanted and users want to just handle the files individually. To allow this a new type attribute named: archive="true|false" is introduced which is evaluated for the kis type only at the moment. If set to false no tar archive is produced and the kiwi bundler handles all result files individually. In addition the documentation around the topic has been improved/corrected. This Fixes #3016

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-kis/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-kis/appliance.kiwi
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<image schemaversion="7.5" name="kiwi-test-image-kis">
+    <description type="system">
+        <author>Marcus Schäfer</author>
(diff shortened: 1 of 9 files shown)


schaefi:
An integration test to test the no-archive mode is available here:

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-kis

PR #3029, opened 2026-08-09
Align docs with code

dcermak:
Fix a plethora of inconsistencies in the docs


schaefi (approved):
Thanks much, lots of good changes 👍 

I walked through most of the changes but due to the size of the doc patch I might have missed one ore the other context change. Should there be an "invalid" doc change we will find them individually. I'm convinced this PR fixes a lot of inconsistencies that existed for quite some time. Thanks again for looking into this

PR #3032, opened 2026-08-10
Tiny fixes

dcermak:


kiwi/container/oci.py
```diff
@@ -17,12 +17,7 @@
 #
 import os
 import logging
-import sys
-from typing import Dict, List, Optional
-if sys.version_info >= (3, 8):
-    from typing import TypedDict  # pragma: no cover
-else:  # pragma: no cover
-    from typing_extensions import TypedDict  # pragma: no cover
+from typing import Dict, List, Optional, TypedDict
```
schaefi, line 20: If you want to remove use of typing_extensions can you do it also at the other places:
* bootloader/config/__init__.py
* builder/disk.py
* command.py

Thanks

schaefi (changes requested):
nice conftest, thanks

kiwi/container/oci.py
schaefi: @dcermak ?

dcermak: done

schaefi (approved):
Looks good thanks 👍

PR #3033, opened 2026-08-11
doc: contributing: Rework AI policy

Conan-Kudo:
This is more in line with our intent and legal frameworks we operate in.


rjschwei (commented):
LGTM, it is more explicit than the previous statement. Being more explicit is helpful, form my perspective. `Assisted-by` vs. `Co-authored-by` falls into the "personal preference" category and I have no strong opinion about using one or the other.

PR #3034, opened 2026-08-11
add settings for claude

jesusbv:
add claude settings.json file

add google-cloud-cli to get the token

add vim in case we want to edit files


schaefi:
Thanks, looks good. I'll merge and check the container rebuilds

PR #3035, opened 2026-08-12
add claude settings.json content

jesusbv:
set a placeholder for the project id

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-claude/config.sh
+++ b/build-tests/x86/tumbleweed/test-image-claude/config.sh
@@ -21,7 +21,6 @@ cat > /etc/claude-code/managed-settings.json <<'EOF'
   "env": {
     "CLAUDE_CODE_USE_VERTEX": "1",
     "CLOUD_ML_REGION": "global",
-    "ANTHROPIC_VERTEX_PROJECT_ID": "your_project_id",
     "ANTHROPIC_DEFAULT_SONNET_MODEL": "claude-sonnet-5",
     "ANTHROPIC_DEFAULT_OPUS_MODEL": "claude-opus-5",
     "ANTHROPIC_DEFAULT_HAIKU_MODEL": "claude-haiku-4-5@20251001"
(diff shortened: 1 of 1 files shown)


build-tests/x86/tumbleweed/test-image-claude/config.sh
```diff
@@ -13,4 +13,19 @@ zypper in google-cloud-cli
 
 mkdir -p ~/.claude
 
-touch ~/.claude/settings.json
+cat > ~/.claude/settings.json <<'EOF'
```
schaefi: Can we move this to 

```/etc/claude-code/managed-settings.json```

The reason is while building the container `~` would match root but when running the container as a flake the user setting up the flake config can decide under which identity and workdir the AI workload should run. As such we cannot pre-define a user config, it might not be found

File: build-tests/x86/tumbleweed/test-image-claude/config.sh [hunk 2]
```diff
@@ -13,4 +13,19 @@ zypper in google-cloud-cli
...
+{
+  "permissions": {
+    "defaultMode": "default"
+  },
+  "env": {
+    "CLAUDE_CODE_USE_VERTEX": "1",
+    "CLOUD_ML_REGION": "global",
+    "ANTHROPIC_VERTEX_PROJECT_ID": "your_project_id",
```
schaefi: the placeholder can then also be dropped, since someone would provide this in a user config

build-tests/x86/tumbleweed/test-image-claude/config.sh
```diff
@@ -13,4 +13,19 @@ zypper in google-cloud-cli
...
+    "ANTHROPIC_DEFAULT_SONNET_MODEL": "claude-sonnet-5",
+    "ANTHROPIC_DEFAULT_OPUS_MODEL": "claude-opus-5",
+    "ANTHROPIC_DEFAULT_HAIKU_MODEL": "claude-haiku-4-5@20251001"
+  }
+}
+EOF
+chmod 0644 ~/.claude/settings.json
```
schaefi: perms should go to `/etc/claude-code/managed-settings.json`

PR #3037, opened 2026-08-20
Fix setup of partition IDs

### LINKED ISSUE #3036: MBR disk images with LVM don't mark the partitions as type LVM
An issue that's been present for a while is that kiwi seems to indiscriminately mark partitions as "Linux", which seems to mess with LVM-based disk images (where it is expected that the partition is marked as an LVM member partition instead). This seems to screw up the LVM tooling too and cause boot failures.

Downstream ticket: https://forge.fedoraproject.org/server/tickets/issues/193

schaefi:
When setting the partition type code (gUID, UUID) the partition id hex code gets resetted to Linux(83) when it should stay at the value it had before. This causes for example on LVM type(8e) partitions a reset to Linux(83) which is unwanted. With this commit we make sure the partition type hex code gets reset to the correct value. This Fixes #3036

### DIFF
--- a/build-tests/x86/tumbleweed/test-image-lvm/appliance.kiwi
+++ b/build-tests/x86/tumbleweed/test-image-lvm/appliance.kiwi
@@ -1,11 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
-
+<!-- OBS-Profiles: @BUILD_FLAVOR@ -->
 <image schemaversion="7.5" name="kiwi-test-image-lvm">
     <description type="system">
         <author>Marcus Schäfer</author>
         <contact>ms@suse.com</contact>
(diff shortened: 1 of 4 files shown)


schaefi:
@Conan-Kudo please wait with merging after my tests are complete

Conan-Kudo:
Do you have a test of an MBR-only image to verify?

schaefi:
> Do you have a test of an MBR-only image to verify?

We can add that. I added it here do the PR... still waiting for the tests to finish

schaefi:
Watching

* https://build.opensuse.org/package/show/Virtualization:Appliances:Images:Testing_x86:tumbleweed/test-image-lvm

schaefi:
@Conan-Kudo two images were built, please double check

PR #3042, opened 2026-08-25
Erofs overlay patches

martinp26:
# disk: honor erofscompression and allow create options for the overlay read-only root fs

Issue#3038
Issue#3039

## What
Two commits fixing/extending compression handling for the `overlayroot`
read-only filesystem:

1. **disk: honor erofscompression for overlay read-only root filesystem**
   The overlay read-only fs can be `squashfs` or `erofs`
   (`overlayroot_readonly_filesystem`), but both overlay code paths in
   `DiskBuilder` hard-coded `get_squashfscompression()`, so
   `erofscompression` was ignored and an erofs overlay base was written
   uncompressed. Routes the read-only compression through a small helper
   (`_get_root_readonly_compression`) that returns `get_erofscompression()`
   for an erofs read-only type and `get_squashfscompression()` otherwise.
   Complements #2648 (which fixed only the standalone erofs builder).

2. **disk: allow extra create options for the overlay read-only root fs**
   Adds a new free-form `overlayroot_readonly_createoptions` attribute
   (whitespace-split, analogous to `fscreateoptions`) passed as
   `create_options` to both overlay read-only `FileSystem` builds. This
   lets a description request e.g.
   `overlayroot_readonly_createoptions="-C 1048576"` for a larger EROFS
   pcluster. Updates schema (rnc + rng), the generated `xml_parse.py`,
   `XMLState.get_fs_readonly_create_option_list()`, `disk.py`, docs and
   tests.

## Why
Building an AL2023 `overlayroot` EROFS image, `erofscompression` had no
effect (uncompressed ~1.4 GB base). With these two changes the recipe can
use `erofscompression="lz4hc,level=12"` +
`overlayroot_readonly_createoptions="-C 1048576"` directly.

## Testing
- `make check` — flake8 clean (shell unchanged)
- `make test` — `mypy kiwi` clean; full unit suite passes at **100%
  coverage** (new code covered; existing overlay test updated)
- `make docs` — sphinx html builds, new attribute documented
- schema regeneration verified: `trang` reproduces `kiwi.rng`
  byte-for-byte from the edited `kiwi.rnc`; the generated `xml_parse.py`
  additions match the generateDS output pattern

Assisted-by: Kiro:claude-opus-4.8

### DIFF
--- a/doc/source/image_description/elements.rst
+++ b/doc/source/image_description/elements.rst
@@ -993,6 +993,15 @@ overlayroot_readonly_filesystem="squashfs|erofs":
   as a read-only filesystem in an `overlayroot` setup. By default,
   `squashfs` is used.
 
+overlayroot_readonly_createoptions="string":
+  For the `oem` type only, specifies additional creation options
+  passed to the tool that builds the read-only filesystem of an
+  `overlayroot` setup (`mkfs.erofs` or `mksquashfs`). The value is a
+  whitespace separated list of options, analogous to `fscreateoptions`
+  for the read-write root filesystem. A typical use is to select a
+  larger EROFS physical cluster for a better compression ratio, for
+  example `overlayroot_readonly_createoptions="-C 1048576"`.
+
 overlayroot_write_partition="true|false":
   For the `oem` type only, this allows you to specify if the extra read-write
   partition in an `overlayroot` setup should be created or not.
--- a/kiwi/builder/disk.py
+++ b/kiwi/builder/disk.py
@@ -283,6 +283,22 @@ class DiskBuilder:
         self.append_unpartitioned_space()
         return self.create_disk_format(result)
 
+    def _get_root_readonly_compression(self):
+        """
+        Compression setting for the overlay read-only root filesystem
+
+        The read-only root filesystem type of an overlayroot image is
+        selectable via the overlayroot_readonly_filesystem attribute
+        (squashfs or erofs). Return the compression value that matches
+        that type, so that both squashfscompression and erofscompression
+        are honored. Before this, the overlay code path always read
+        squashfscompression, which silently ignored erofscompression for
+        an erofs overlay root.
+        """
+        if self.root_filesystem_read_only_type == 'erofs':
+            return self.xml_state.build_type.get_erofscompression()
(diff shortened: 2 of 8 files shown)


schaefi (changes requested):
Thanks much for looking into this.

I don't think we need a new attribute to differentiate compression options for erofs between different image layouts. The schema already provides `k.type.erofscompression.attribute`. So it should be possible to utilize this in the same way as it was done for `builder/live.py` and `builder/filesystem.py`. Just add the missing part to `builder/disk.py` as you you did but without the need for a new attribute in the schema

Thoughts ?

schaefi:
> use erofscompression="lz4hc,level=12" + overlayroot_readonly_createoptions="-C 1048576" directly.

under which condition you would need to specify both for one image type ?

schaefi:
From my perspective you only need to apply the code we have in the other builders 

```
filesystem_custom_parameters.update( 
    {
        'compression':
    } if root_filesystem == 'squashfs' else {
        'compression':
            self.xml_state.build_type.get_erofscompression()
    }
)
```

to the `builder/disk.py` where we only have

```
custom_args={
    'compression':
}
```

at several places

martinp26:
> > use erofscompression="lz4hc,level=12" + overlayroot_readonly_createoptions="-C 1048576" directly.
> 
> under which condition you would need to specify both for one image type ?

mkfs.erofs carves out `-C max-pcluster-size` as a distinct setting from `-z compression-algorithm[,#][:...]` and there are potentially more options that one may want to pass to mkfs.*.  `fscreateoptions` is used for the RW layer, `overlayroot_readonly_createoptions` would be the equivalent for the RO layer.

I am not seeing a way to express that with today's schema, am I missing it?

schaefi:
@martinp26 ok, sorry I was on the wrong track. Yes makes perfect sense as we are using `erofscompression` specifically for compression settings and there is nothing that handles filesystem creation options for the read-only filesystem in an overlay build.

All good, sorry for the noise

schaefi (approved):
Looks good 👍

schaefi:
I will build Staging packages at

https://build.opensuse.org/project/show/Virtualization:Appliances:Staging

martinp26:
Thanks!

PR #3046, opened 2026-09-10
Use typer shell completion

schaefi:
Fix typer standalone mode and allow to use the typer completion mechanics. The standalone completion we provide with the kiwi-bash-completion has changed to the variant provided by typer. Even without a completion installed the new --install-completion option provided by typer allows to install a completion based on the running shell in the user home directory.


schaefi:
@Conan-Kudo This also works with fish ;)

Conan-Kudo (approved):

PR #3048, opened 2026-09-10
Make the luks device mapper name unique per build

pulchart:
## Problem

`LuksDevice.luks_name` is the fixed string `luksRoot`. Device mapper names live in the host kernel and are not namespaced by the build chroot, so two kiwi builds that use `luks=` and run concurrently on one host race on that name. The second one to reach `luksOpen` fails:

```
[ DEBUG   ]: EXEC: [cryptsetup --key-file /dev/zero --keyfile-size 32 luksOpen /dev/mapper/loop2p3 luksRoot]
[ DEBUG   ]: EXEC: Failed with stderr: Device luksRoot already exists.
[ ERROR   ]: KiwiCommandError: cryptsetup: stderr: Device luksRoot already exists.
```

Observed with kiwi 10.3.0 building two disk images from the same description (different distro majors, both with `<type image="oem" luks="" luks_version="luks2">`) in parallel on one build host. The two builds started 7s apart; the loser died at `luksOpen` while the winner still held the map:

| build | start | end |
|---|---|---|
| A | 17:31:22 | 17:39:07 (ok) |
| B | 17:31:29 | 17:36:55 (failed at `luksOpen`) |
| B, retried alone | 17:39:43 | 17:44:02 (ok) |

The window is short, at the very end of the build, so this shows up as an intermittent failure that passes on retry, which makes it easy to misattribute.

## Fix

Generate the map name per instance: `luksRoot_<uuid4 hex>`.

The name is only used to open and close the map during the build (`luksOpen`, `/dev/mapper/<name>`, `luksClose`). It is not written into the produced image: `create_crypttab` emits the mapping name `luks` together with `UUID=` independently of `luks_name`, and `grep -rn luksRoot` over the tree matches only the one assignment. So the built image is unaffected.

kiwi already handles this class of collision the same way in `clone_device.py`, where the LUKS UUID, the LVM volume group name and the RAID superblock UUID are each made unique to avoid conflicts.

## Tests

Added `test_luks_name_is_unique_per_instance`. The existing tests that asserted the literal `luksRoot` passed to cryptsetup now assert against the generated name.

`pytest test/unit/storage/luks_device_test.py` -> 11 passed. The wider `test/unit/storage` suite is unchanged relative to `main` in this environment: 1 pre-existing failure (`disk_test.py::test_get_discoverable_partition_ids`) and 27 pre-existing errors before and after, with one additional passing test from this change.

Happy to open a tracking issue and reference it here if you would prefer that per CONTRIBUTING.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kiwi/storage/luks_device.py
+++ b/kiwi/storage/luks_device.py
@@ -16,6 +16,7 @@
 # along with kiwi.  If not, see <http://www.gnu.org/licenses/>
 #
 import os
+import uuid
 import logging
 import binascii
 from typing import Optional
@@ -49,7 +50,11 @@ class LuksDevice(DeviceProvider):
         self.luks_device: Optional[str] = None
         self.luks_keyfile: str = ''
         self.passphrase: str = ''
-        self.luks_name = 'luksRoot'
+        # The device mapper name must be unique per build. Device mapper
+        # names are global to the host kernel and are not namespaced by
+        # the build chroot, so a fixed name makes concurrent builds on
+        # one host fail with 'Device luksRoot already exists.'
+        self.luks_name = 'luksRoot_{0}'.format(uuid.uuid4().hex)
 
         self.option_map = {
             'sle12': [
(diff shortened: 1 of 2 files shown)


schaefi (approved):
Thanks and makes sense 👍 We would have that issue even in container isolated builds when the host kernel is shared, as such a good catch. Mr claude did a good job 😄 Thanks

PR #3051, opened 2026-09-14
Compare oem dump target blocksize against the logical block size

saty-nexthop:
Fixes Issue#3050

`target_blocksize` is documented as the logical blocksize of the target device (the `SSZ` column of `blockdev --report`), but the sanity check added in #3022 compares it against `physical_block_size`. On 512e drives (512 logical, 4096 physical) that mismatches on every default image and the oem dump reboot-loops.

Read `logical_block_size` instead. A native 4Kn disk reports 4096 for both, so the case #3022 was written for is still caught.

No test change: the dracut scripts have no unit coverage and this only changes which sysfs attribute is read. Verified on a 512e NVMe (logical 512 / physical 4096) where the unpatched check aborts and the patched one dumps and boots.


schaefi:
Uhm, thanks good catch and a mistake I made when adding the sanity check. 

> A native 4Kn disk reports 4096 for both

yes and I tested only on such devices, my bad

Thanks

PR #3053, opened 2026-09-18
Fix LUKS header checksum reference for reencryption

fcrozat:
There was a regression when implementing sha512 support: header checksum is computed too late and end-up being null.

This is breaking SL Micro 6.2 Encrypted image when booting with FIPS. It is only visible in FIPS environment due to other bugs in fde-tools causing re-encryption to fail if kiwi re-encryption fails in the first place ( https://github.com/openSUSE/fde-tools/pull/54 and https://github.com/openSUSE/fde-tools/pull/57 ).


schaefi (approved):
Great catch, thanks much 👍

PR #3054, opened 2026-09-21
Fix handling of Path annotated commandline options

schaefi:
With the move to typer all options that takes a path argument are validated through typer as PosixPath. After the successful validation the value must be transported as a string into the application because that's what the implementation expected from the former docopt based option information. This commit adds a method that makes sure all Path annotated commandline options are properly provided as a string into kiwi.

This Fixes bsc#1280097


Conan-Kudo (approved):

garloff:
Thanks!

